PR #147, opened 2012-05-09
I have extended the limnology and oceanography style to include technica...

chogg:
...l report names and to deal with conference papers more simply.

In addition, I've taken the doi reference out. This is optional in the author's instructions so people might want to put this back in at line 209.

This is the first time I've used github, so help would be appreciated if I've done this wrong. Thanks.


jensdmueller:
Hi all,
I currently write an article for L&O. I used Mendeley + this style to write citations.
I noticed two disagreements with the guidelines of L&O:
1. The doi does not appear in the bibliography.
2. In the text, two citations from the same year should appear in alphatical order, but do not.
Is it possible for you guys to change this?
I'am very new to the system, so I fear to break more than I fix, if I try it myself.
Thanks for helping.
Cheers
Jens

adam3smith:
you can't actually break anything, so feel free to try, but we should also be able to do this in the next couple of days.

jensdmueller:
Sounds great! I'll give you guys a try first :-) Thanks Jens

adam3smith:
OK, @jensdmueller  Fix for 1. is up (together with some other touch-ups). 2. should already have been correct. I made a small change that might address this, but I doubt it. I recall this coming up before in Mendeley (with different styles) so report to them. 
You'll want to mention that you already talked to me, otherwise they'll just send you to their style error report tool, which in turn gets forward to us.

jensdmueller:
Hi adam3smith,
thanks a lot for fixing 1.
I updated the references today and it works perfectly fine.
2. (the alphabetical order) is still not working. I will report to Mendeley as you suggested.
Thanks again
Jens

PR #702, opened 2013-09-15
French-canadian styles : update default-locale to fr-CA

f-mb:


rmzelle:
@f-mb, do you know anything about this?

https://forums.zotero.org/discussion/63923/make-a-non-listed-csl-1-0-1-valid

PR #837, opened 2014-02-09
Create journal-of-computational-biology.csl

delavefm:


sjackman:
I would like this CSL as well.

adam3smith:
You can find the file here: https://github.com/citation-style-language/styles/pull/837/files feel free to fix it up as per Rintze's comments (and double-check for remaining issues) and submit in a new pull request.

rmzelle:
(the direct download link is at https://raw.githubusercontent.com/delavefm/styles/patch-1/journal-of-computational-biology.csl )

PR #1092, opened 2014-09-03
Update mis-quarterly.csl

guoxh:
I don't see that DOI is required in the journal's guidelines (http://www.misq.org/manuscript-guidelines), so I suggest that it can be removed. Anyway, if DOI is still included, there should be a space before it, as I have proposed in this change.

### DIFF
--- a/mis-quarterly.csl
+++ b/mis-quarterly.csl
@@ -349,7 +349,7 @@
           <label variable="page" form="short"/>
           <text variable="page"/>
         </group>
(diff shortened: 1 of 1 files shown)


rmzelle:
@jameshowison, did you ever hear back?

rmzelle:

jameshowison:
Not meaning to re-open, just following up. I didn't hear back from the editorial coordinator.

PR #1114, opened 2014-09-14
Update american-journal-of-archaeology.csl

charlieh1006:
1) Removes comma which appears after "edited by" but before first editor's name.
2) Removes whitespace between semicolon and page numbers for journal articles.


rmzelle:
@charlieh1006, I'm closing this pull request for now because of inactivity. (if you get back to us we can re-open it)

PR #1224, opened 2014-11-04
Delete journal-of-chiropractic-medicine.csl

grnathan2000:
Deleted as per the request.


rmzelle:
I added this style in #1314.

Thanks @grnathan2000!

PR #1227, opened 2014-11-05
Create modern-language-association-annotated-bib-with-url.csl

atbenson:
This was developed for the International Writing Centers Association's Antiracism SIG annotation project. It's based on the latest MLA .csl file with the addition of URLs in opening and closing brackets and the use of the abstract field for annotations. It aligns with MLA guidelines for annotated bibliographies, so it could be used by students for annotated bibliographies.


adam3smith:
I'll close this. You will be able to point people to the style and they can still install it, but for the time being we won't take it for the repository. Thanks for putting it up, though!

PR #1229, opened 2014-11-07
Create iso690-numeric-pl.csl

milekpl:
This is another standard style in Polish bibliography, available on CC license.

### DIFF
--- a/iso690-numeric-pl.csl
+++ b/iso690-numeric-pl.csl
@@ -1,0 +1,723 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" default-locale="pl-PL">
+  <info>
+    <title>Polish Standard ISO 690-1999 and 690-2-2002 (numeric, Polish)</title>
+    <id>http://www.zotero.org/styles/iso690-numeric-pl</id>
+    <link href="http://www.zotero.org/styles/iso690-numeric-pl" rel="self"/>
+    <link href="http://www.zotero.org/styles/iso690-numeric-en" rel="template"/>
+    <author>
+      <name>Krzysztof Saja</name>
+      <email>krzysztofsaja@gmail.com</email>
+    </author>
+    <category citation-format="numeric"/>
+    <category field="generic-base"/>
+    <updated>2012-11-15T14:21:45+00:00</updated>
+    <summary>Polish Standard ISO 690-1999 and 690-2-2002. Polish translation of style based on ISO 690:2010(E), V1</summary>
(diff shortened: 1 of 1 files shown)


rmzelle:
Coming back to this, @milekpl, where did you find this style? Is it currently available on a different website?

milekpl:
Huh, it's on a different website. Sorry, I was flooded with work. I will email Krzysztof right away.

cogosakrth:
Hi, I made that style by coping Polish locale file into style file because it had to work also in Mendeley. And as far as I know, this was the only way to do it for Zotero and Mendeley.

rmzelle:
@cogosakrth, maybe back when you made this style, Zotero and Mendeley didn't yet have the Polish locale file?

Ideally the style should only redefine those terms that are different from the ones in the Polish locale file at https://github.com/citation-style-language/locales/blob/master/locales-pl-PL.xml.

adam3smith:
(@rmzelle - no, this still doesn't work in Mendeley without the embedded locale file, as you'll recall from other conversations)

rmzelle:
> (@rmzelle - no, this still doesn't work in Mendeley without the embedded locale file, as you'll recall from other conversations)

Even if the user changes the locale to Polish as per http://support.mendeley.com/customer/portal/articles/1367232-how-do-i-change-the-bibliography-s-language- ?

adam3smith:
oh, I assume that works, but that's not a real solution, is it? If you want a style to "just work" rather than require elaborate instructions to everyone who installs it, you still (though hopefully not much longer) have to embed the locale.

rmzelle:
Last I heard is that Mendely is finally, finally, looking at improving this in their software.

@cogosakrth, do you remember if you made any other changes to the style, other than changing the "default-locale" and pasting in the locale file?

rmzelle:
Mendeley's next release should make it easier to select a locale. They also say "By default citations will use the language associated with the selected citation style", so it really shouldn't be necessary to include a complete locale file within the style. If it doesn't work, it should be reported as a bug to Mendeley.

See http://www.mendeley.com/release-notes/v1_13_4-dev1/

milekpl:
@rmzelle: There is a problem here, as Mendeley >1.11 has no support for Mac Word 2008, and a lot of people use the outdated version now.

rmzelle:
Uhm, sorry for letting this pull request linger. On second thought, I'm willing to accept this, but the title needs to be changed to just `<title>ISO-690 (numeric, Polish)</title>`. @milekpl can make the change in the current pull request by clicking the pencil icon at https://github.com/milekpl/styles/blob/patch-2/iso690-numeric-pl.csl

rmzelle:
Thanks. Is there an online description on this format, so that we can add a "documentation" link? (point 10 at https://github.com/citation-style-language/styles/wiki/Style-Requirements )

rmzelle:

PR #1257, opened 2014-11-25
Update acta-neuropathologica.csl

robertb-r:
This style requires the references to be sorted alphabetically and then numbered. 
I have added a sort function that fixes this error.
new to csl and github and do not know how to move this from a dependent as it is now modified from the parent style.


adam3smith:
added in https://github.com/citation-style-language/styles/commit/0066ae3432e68e940f9231fa1a09b9c43fd88ec7

PR #1276, opened 2014-12-05
Create Wiley-book

maksrules:
This is a newly created style for references in a chapter of a book published by Wiley. Field - research. Template used to create this style - IEEE.

### DIFF
--- a/wiley-vch-books.csl
+++ b/wiley-vch-books.csl
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
   <info>
-    <title>Wiley-VCH</title>
-    <id>http://www.zotero.org/styles/wiley-vch</id>
-    <link href="http://www.zotero.org/styles/wiley-vch" rel="self"/>
+    <title>Wiley-VCH books</title>
+    <id>http://www.zotero.org/styles/wiley-vch-books</id>
+    <link href="http://www.zotero.org/styles/wiley-vch-books" rel="self"/>
     <link href="http://www.wiley-vch.de/publish/en/authors/auguidelines" rel="documentation"/>
     <author>
       <name>Maksim Ivanov</name>


adam3smith:
@maksrules -- we've received an error report that states:
> "For Wiley-VCH books, the year of publication should come after the publication title, before the name of the journal and in brackets. Not at the end of the reference as is currently the case. 

The example given is

> F. Manoni, C. Cornaggia, J. Murray, S. Tallon, and S. J. Connon, “Catalytic, enantio- and diastereoselective synthesis of γ-butyrolactones incorporating quaternary stereocentres.,” (2012), Chem. Commun. (Camb)., 48, 52, 6502–4. 

Could you check that with the instructions you received from Wiley? The online style is, unfortunately, no help at all here, since that's for the Harvard version.

maksrules:
@adam3smith 

I checked the instructions. I'm not sure if it is OK to make the instructions public (you never know), but here is an excerpt (for the Vancouver version, I don't really have Harvard version instructions):
>**An article in a journal:**
>Onodera, A., Inoue, K., Yoshihara, H. et al. (1990) Synthesis of cubic boron nitride from hombohedral under high static pressure. *Journal of Materials Science*, **25** (10), 429–484.
>
>**A chapter in an edited book:**
>Stewart, S. and Kam, K.C. (1990) Radiotherapy techniques, planning and equipment, in *Treatment of Cancer*, 2nd edn (eds K. Sikora and K.E. Halnan), John Wiley & Sons, Ltd, Chichester, pp. 827–851.
>
>Sissons, H.A. (1971) The growth of bone, in *The Biochemistry and Physiology of Bone*, vol. 3 (ed. GH. Bourne), Academic Press, New York, pp. 145–180.
>
>**Conference proceedings:**
>*1. Published*
>Reinhardt, H.W. and Naaman, A.E. (eds) (1992) *High Performance Fiber Reinforced Cement Composites*. Proceedings of the International RILEM/ACI Workshop, June 23–26, 1991, Mainz, Germany. John Wiley & Sons, Ltd, Chichester.
>
>*2. Unpublished*
>Johnson, W.T., Phelps, R.W. and Beatson, P.J. (1997) Wastewater reuse using membranes. IDA World Congress on Desalination and Water Science, Madrid.




Is there more information from where the style in the report is taken from? Perhaps the problem is deeper, i.e. the style depends on the science field, or something like that? That would be a not-so-good situation...

adam3smith:
OK, that's what I found online, too, but this is different from the style you submitted, which has the date all the way at the end?

maksrules:
I have no idea how, but I uploaded an interim version of the style, and replaced the final one everywhere I had it... :( The worst ever.

Anyway, I remade the style from scratch. As it is not based on any template now, it supports articles, books and chapters well, everything else uses fallback style, which is based in article. But I have no examples to check how it works.

I will upload the new version.

PR #1284, opened 2014-12-10
New style for journal "Computer und Recht" (German)

rm2342:


rmzelle:
@rm2342, are there Instructions to Authors for this journal that describe this journal's citation format? I couldn't find them at http://www.computerundrecht.de/cr.htm.

rm2342:
Not that I know of. What I did was look at the style they use in
articles. I write for this journal now and then and also followed what I
used - which was not corrected by the lector.

rmzelle:
Okay, thanks.

I just discovered that your style didn't have a documentation link, so I added one pointing to the journal's website. That'll have to do for now. (see https://github.com/citation-style-language/styles/commit/66790e2ecdabb81e15c2ee007244a648c7607a64#diff-0 )

PR #1297, opened 2014-12-24
Create FLOSS4INGO

vmalep:
Style created for a project in the framework of a Master Degree in Spain

### DIFF
--- a/FLOSS4INGO
+++ b/FLOSS4INGO
@@ -1,0 +1,327 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
+  <info>
+    <title>vmalep</title>
+    <author>
+      <name>Pierre van Male</name>
+      <email>vmalep@gmail.com</email>
+    </author>
+    <id>http://www.zotero.org/styles/vmalep</id>
+    <link href="http://www.zotero.org/styles/vmalep" rel="self"/>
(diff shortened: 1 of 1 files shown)


rmzelle:
> I am know copiing the document on a wiki website (http://www.floss4ingo.xyz) and I would like to use the same style, but for this, it must be available from zotero database.

Which software are you using that requires this?

Since the styles from our repository are offered to hundreds of thousands of users, we're very hesitant to accept styles for individuals or individual projects.

vmalep:
Dear Rintze,
    
    Thanks for your reply. The website is done with TikiWiki and it has
    the feature to create a footnote with the Zotero reference according
    to the style specified in the parameter, but this style must be
    available from your website.
    
    In any case, I decided to proceed differently (with a small tag
    defined as an hyperlink to the Zotero page directly).
    
    Thanks for your support and have a good continuation!
    Pierre
    
    On 04/01/15 14:11, Rintze M. Zelle
      wrote:
    
    
      
        I am know copiing the document on a wiki website (http://www.floss4ingo.xyz)
          and I would like to use the same style, but for this, it must
          be available from zotero database.
      
      Which software are you using that requires this?
      Since the styles from our repository are offered to hundreds of
        thousands of users, we're very hesitant to accept styles for
        individuals or individual projects.
      —
        Reply to this email directly or view
          it on GitHub.

rmzelle:
Does TikiWiki support Zotero referencing by itself, or were you planning to use some TikiWiki plugin?

rmzelle:
Closing this as we're unlikely to accept this style.

PR #1298, opened 2014-12-26
Update us-geological-survey.csl

kknierim:
USGS Reports need to have the report type and number in the bibliography.  This edits changes the USGS style for reports to:

> Moix, M.W., Barks, C.S., and Funkhouser, J.E., 2003, Water quality and streamflow gains and losses of Osage and Prairie Creeks, Benton County, Arkansas, July 2001: U.S. Geological Survey Water-Resources Investigations Report 03-4187, p. 35.

Note that the report type ("Water-Resources Investigations Report") and number ("03-4187") are pulled into the citation.

### DIFF
--- a/us-geological-survey.csl
+++ b/us-geological-survey.csl
@@ -88,11 +88,18 @@
       </group>
       <choose>
         <if type="report">
-          <text variable="title" prefix=" " suffix=":"/>
-          <text variable="publisher" prefix=" "/>
-          <text variable="genre" prefix=" "/>
-          <text variable="number" prefix =" " suffix=","/>
-          <text variable="page" prefix=" p. " suffix="."/>
+           <group prefix=" " delimiter=", " suffix=".">
+            <group delimiter=" ">
+              <text variable="title" suffix=":"/>
+              <text variable="publisher"/>
+              <text variable="genre"/>
+              <text variable="number"/>
+            </group>
+            <group>
+              <label variable="page" form="short" plural="never" suffix=" "/>
+              <text variable="page"/>
+            </group>
+          </group>
         </if>
         <else-if type="bill book graphic legal_case legislation motion_picture report song" match="any">
           <group suffix=":">
--- a/us-geological-survey.csl
+++ b/us-geological-survey.csl
@@ -89,8 +89,10 @@
       <choose>
         <if type="report">
           <text variable="title" prefix=" " suffix=":"/>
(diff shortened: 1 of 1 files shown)


rmzelle:
@adam3smith, could you check whether

```
        <group prefix=" " delimiter=", " suffix=".">
          <group delimiter=" ">
            <text variable="title" suffix=":"/>
            <text variable="publisher"/>
            <text variable="genre"/>
            <text variable="number"/>
          </group>
          <group>
            <label variable="page" form="short" suffix=" "/>
            <text variable="page"/>
          </group>
        </group>
```

would be preferable over

```
          <text variable="title" prefix=" " suffix=":"/>
          <text variable="publisher" prefix=" "/>
          <text variable="genre" prefix=" "/>
          <text variable="number" prefix =" " suffix=","/>
          <text variable="page" prefix=" p. " suffix="."/>
```

adam3smith:
yes, definitely, though the label should be plural="never" to be safe and also to make this more robust for when this is finally switched to number-of-pages. 
But otherwise, as it is now the comma after number, especially, is very fragile: it doesn't show up when there is no report number and it ends the citation when there is no page number. In other words, we definitely want to handle this with groups/delimiters.

rmzelle:
@kknierim, can you replace the five lines I highlighted with:

```
        <group prefix=" " delimiter=", " suffix=".">
          <group delimiter=" ">
            <text variable="title" suffix=":"/>
            <text variable="publisher"/>
            <text variable="genre"/>
            <text variable="number"/>
          </group>
          <group>
            <label variable="page" form="short" plural="never" suffix=" "/>
            <text variable="page"/>
          </group>
        </group>
```

As @adam3smith explained, the use of prefixes, delimiters, and suffixes on `<group/>` is usually preferable over putting them directly on `<text/>` elements.

You can edit your style by using the edit (pencil) icon at https://github.com/kknierim/styles/blob/patch-1/us-geological-survey.csl

kknierim:
Thanks for the help all! I'll get this fixed up next week (traveling now).
Appreciate the help.

Kathy

kknierim:
@rmzelle , I added your "group" changes in.  I tried to follow the tabs correctly (I am new to coding, so apologies if I didn't paste and tab correctly).  Thanks for the help.  I see how the group elements would be better than including punctuation in the <text/> elements.

rmzelle:
Great, thanks!

PR #1299, opened 2014-12-27
Create revista-de-filologia-espannola.csl

jmfr:
Citational software does not have proper style for Spanish language journal on Spanish Language, Literature & Linguistics topic. The Revista de Filología Española is one of the most valued journals and its style is widely used outside in other publications.

### DIFF
--- a/revista-de-filologia-espannola.csl
+++ b/revista-de-filologia-espannola.csl
@@ -1,0 +1,182 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" et-al-min="10" initialize="false" demote-non-dropping-particle="sort-only" default-locale="es-ES">
+<!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>Revista de Filología Española (Spanish)</title>
+    <title-short>RFE parentético</title-short>
+    <id>http://www.zotero.org/styles/revista-de-filologia-espannola</id>
+    <link href="http://www.zotero.org/styles/revista-de-filología-española" rel="self">http://www.zotero.org/styles/revista-de-filologia-espannola</link>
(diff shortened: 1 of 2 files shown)


rmzelle:
@jmfr, can you fix the following issues?

http://validator.citationstyles.org/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fjmfr%2Fstyles%2Fa9d473aad299d6306be4e78c470320912bb1e031%2Frevista-de-filologia-espanola.csl&version=1.0.1

1. You have two `<updated/>` elements.
2. You use ` term="en"`, but "en" is not a CSL term.

Finally, please change `<link href="http://www.zotero.org/styles/revista-de-filologia-espanola" rel="self">http://www.zotero.org/styles/revista-de-filologia-espanola</link>` to `<link href="http://www.zotero.org/styles/revista-de-filologia-espanola" rel="self"/>`.

jmfr:
Done as indicated. However, I'm dubious about the fixing I've done on line 138.  But validator says it is fine.

José Manuel Fradejas Rueda
Dpto. Lengua Española
Facultad de Filosofía y Letras
Universidad de Valladolid
España

rmzelle:
That doesn't seem right.

You call the "title" macro on line 165 (https://github.com/jmfr/styles/blob/a2da6014a71163aa75d2861e4139c56988c74323/revista-de-filologia-espanola.csl#L165), and then the "container-prefix" macro on line 167, and now both print the "title" variable.

Shouldn't you just be using the "in" term, as in your original submission? (see https://github.com/jmfr/styles/commit/e9eeee7b94b0dff0ea2d42c03014aa6e7e98cb88#diff-1233e50558908aaacdc4fa131116416dR135 )

jmfr:
Fixed. I originally used "en" because it is the Spanish equivalent to "in".

rmzelle:
Great, thanks! (and with the style set to Spanish, the "in" term will automatically translate to "en", as defined in https://github.com/citation-style-language/locales/blob/master/locales-es-ES.xml )

PR #1300, opened 2014-12-28
Fix double space before "in" ("dans") in lluelles.csl and lluelles-lexisnexis.csl

f-mb:


adam3smith:
thanks! (on a side-note: simple pull requests should never take us more than a week to review. Please feel free to send a reminder if we take longer).

PR #1302, opened 2014-12-30
european-journal-of-vascular-and-endovascular-surgery.csl

Sankar-GEO-ELS:
Style for Elsevier nonstandard journal (European Journal of Vascular and Endovascular Surgery) with slight changes to the Vancouver style.


rmzelle:
Continued in #1303.

PR #1305, opened 2014-12-30
Update vietnam-ministry-of-education-and-training-en.csl

Tungphamfsh:
Add support for citing a thesis. I forget to add it in the first place


rmzelle:
Continued by #1306.

PR #1307, opened 2014-12-30
Update vietnam-ministry-of-education-and-training-vi.csl

Tungphamfsh:
Add support for citing thesis


Tungphamfsh:
Dear adam3smith

Is there any problem with the style. I saw that the request is not closed yet

PR #1309, opened 2014-12-31
european-journal-of-paediatric-neurology.csl

Sankar-GEO-ELS:

### DIFF
--- a/european-journal-of-paediatric-neurology.csl
+++ b/european-journal-of-paediatric-neurology.csl
@@ -1,0 +1,200 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" page-range-format="minimal" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>European Journal of Paediatric Neurology</title>
+    <title-short>EJPN</title-short>
+    <id>http://www.zotero.org/styles/european-journal-of-paediatric-neurology</id>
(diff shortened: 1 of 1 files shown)


european-journal-of-paediatric-neurology.csl
```diff
@@ -0,0 +1,200 @@
...
+    <link href="http://www.zotero.org/styles/elsevier-vancouver" rel="template"/>
+    <link href="http://www.elsevier.com/journals/european-journal-of-paediatric-neurology/1090-3798/guide-for-authors" rel="documentation"/>
+    <author>
+      <name>Sankar</name>
+      <email>s.murugesan@elsevier.com</email>
+    </author>
+    <category citation-format="numeric"/>
+    <category field="medicine"/>
+    <issn>1090-3798</issn>
+	<issn>1532-2130</issn>
```
adam3smith: we're not super picky with new submissions, but since I believe you're working in Oxygen anyway (is that right?) please do auto-indent the styles (at two spaces per level as you do in general) before submission. Thanks & Happy New Year.

rmzelle: @Sankar-GEO-ELS, for future submissions, could you use the `<eissn/>` element for the electronic ISSN? (in this case, 1532-2130 is the eISSN, see http://www.ncbi.nlm.nih.gov/nlmcatalog?term=1090-3798%5BISSN%5D )

Thanks!

Changed in https://github.com/citation-style-language/styles/commit/e0918a75a0b7840282cdefc764d0b6b42c024842

Issue #1301, opened 2014-12-29
Adjacent spaces from suffix and prefix


zuphilip:
Sometimes there is a double space coded in citation styles:
 * one space from an suffix attribute
 * and a following space from the next element's prefix attribute.
One example is the pull request #1300. Other examples I have spotted:
```
Search "<[^/>]*suffix="[^"/>]* "[^/>]*/?>\r\n\s*<[^/>]*prefix=" [^"/>]*"[^/>]*/?>" (291 hits in 226 files)
  .\styles\acta-anaesthesiologica-scandinavica.csl (1 hit)
	Line 124:           <text macro="edition" prefix=" " suffix=" "/>
  .\styles\acta-societatis-botanicorum-poloniae.csl (1 hit)
	Line 117:           <text macro="edition" prefix=" " suffix=" "/>
  .\styles\ambio.csl (1 hit)
	Line 272:             <text variable="edition" suffix=" "/>
  .\styles\american-journal-of-archaeology.csl (1 hit)
	Line 229:             <text variable="edition" suffix=" "/>
  .\styles\american-physiological-society.csl (1 hit)
	Line 218:           <text variable="page" prefix=": " suffix=", "/>
  .\styles\american-sociological-association.csl (1 hit)
	Line 64:           <date variable="accessed" form="text" suffix=" "/>
  .\styles\annalen-des-naturhistorischen-museums-in-wien.csl (1 hit)
	Line 196:           <text term="in" suffix=": " text-case="capitalize-first"/>
  .\styles\annals-of-neurology.csl (1 hit)
	Line 125:           <text macro="edition" prefix=" " suffix=" "/>
  .\styles\annals-of-oncology.csl (2 hits)
	Line 96:               <text macro="edition" suffix=" "/>
	Line 102:               <text macro="edition" suffix=". "/>
  .\styles\antiquity.csl (1 hit)
	Line 251:             <text variable="edition" suffix=" "/>
  .\styles\apa-tr.csl (2 hits)
	Line 178:                   <text variable="URL" suffix=" "/>
	Line 185:                   <text variable="URL" suffix=" "/>
  .\styles\archives-of-physical-medicine-and-rehabilitation.csl (1 hit)
	Line 124:           <text macro="edition" prefix=" " suffix=" "/>
  .\styles\arzneimitteltherapie.csl (1 hit)
	Line 133:           <text macro="edition" prefix=" " suffix=" "/>
  .\styles\asian-studies-review.csl (1 hit)
[...]

PR #1310, opened 2015-01-01
Delete adjacent spaces in edition/publisher pattern

zuphilip:
Replacing the prefix/suffix adjacent spaces with group and one delimiter space, as described generally in #1301 . Search for this pattern here (57 hits in 56 files):

```
<[^/>]*"edition"[^/>]*suffix="[^"/>]* "[^/>]*/?>\s*\r\n\s*<[^/>]*"publisher"[^/>]*prefix=" [^"/>]*"[^/>]*/?>
```

@adam3smith et al: Happy New Year to you!


adam3smith:
looking good, too.

zuphilip:
Thanks. Then, I can continue on the remaining classes of this pattern.

PR #1311, opened 2015-01-01
Delete adjacent spaces in edition/edition pattern

zuphilip:
Replacing the prefix/suffix adjacent spaces with group and one delimiter space, as described generally in #1301 . Search for this pattern here (70 hits in 70 files):

```
<[^/>]*"edition"[^/>]*suffix="[^"/>]* "[^/>]*/?>\s*\r\n\s*<[^/>]*"edition"[^/>]*prefix=" [^"/>]*"[^/>]*/?>
```

Every occurence here looks actually the same:

```
          <group>
            <text variable="edition" suffix=" "/>
            <text term="edition" prefix=" "/>
          </group>
```
which will then replaced by
```
          <group delimiter=" ">
            <text variable="edition"/>
            <text term="edition"/>
          </group>
```

### DIFF
--- a/ambio.csl
+++ b/ambio.csl
@@ -268,9 +268,9 @@
     <choose>
       <if type="article-newspaper">
         <group prefix=", " delimiter=", ">
-          <group>
-            <text variable="edition" suffix=" "/>
-            <text term="edition" prefix=" "/>
+          <group delimiter=" ">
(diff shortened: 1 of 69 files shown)


adam3smith:
great, thanks. Sorry, these had slipped from my list of things to look at.

PR #1316, opened 2015-01-06
GBE: Remove parentheses around the year

sjackman:
I'm using `genome-biology-and-evolution.csl` with Pandoc. A reference such as `(@foo)` expands to `(Foo et al. (2015))`, whereas it should expand to `(Foo et al. 2015)`. This patch removes the parentheses around the year.


adam3smith:
The parentheses are correct the way they are. You should just be using [@foo] for a regular parenthetical reference and Foo [-@foo] for Author (Year).

sjackman:
Ah! Thanks for the help.

PR #1324, opened 2015-01-08
orthopedic-clinics-of-north-america.csl

Sankar-GEO-ELS:
Elsevier nonstandard journal with slight changes to the Vancouver style


rmzelle:
Thanks!

PR #1326, opened 2015-01-12
Update Academy of Management Style

bwiernik:
Updated according to current style guide.

1. Fix title case of journal titles
2. Consistently format titles of items and containers across item types
3. Fix formatting of genre, publisher, and publisher place as best as
CSL can handle
4. Adjusted display of URL to only display for non-print references and to prefer issue date over access date

### DIFF
--- a/academy-of-management-review.csl
+++ b/academy-of-management-review.csl
@@ -146,9 +146,19 @@
   <macro name="event">
     <choose>
       <if variable="event">
-        <text variable="genre" text-case="capitalize-first" suffix=" "/>
-        <text term="presented at" text-case="capitalize-first" suffix=" "/>
-        <text variable="event"/>
+        <choose>
+		  <if variable="genre" match="none">
+		    <text term="presented at" text-case="capitalize-first" suffix=" "/>
+            <text variable="event"/>
+		  </if>
+		  <else>
+		    <group delimiter=" ">
+			  <text variable="genre" text-case="capitalize-first"/>
+              <text term="presented at"/>
+              <text variable="event"/>
+			</group>
+		  </else>
+		</choose>
       </if>
     </choose>
   </macro>
--- a/academy-of-management-review.csl
+++ b/academy-of-management-review.csl
@@ -5,7 +5,7 @@
     <title-short>AMR</title-short>
     <id>http://www.zotero.org/styles/academy-of-management-review</id>
     <link href="http://www.zotero.org/styles/academy-of-management-review" rel="self"/>
-    <link href="http://journals.aomonline.org/amr/AMRstyleguide.pdf" rel="documentation"/>
+    <link href="http://aom.org/uploadedFiles/Publications/AMR/AMR%20Style%20Guide%202014.pdf" rel="documentation"/>
     <author>
       <name>Sebastian Karcher</name>
     </author>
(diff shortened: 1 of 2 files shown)


academy-of-management-review.csl
```diff
@@ -234,16 +252,23 @@
           <group>
             <text macro="container-contributors"/>
             <text macro="secondary-contributors"/>
-            <text variable="container-title" font-style="italic" font-weight="bold"/>
+            <choose>
+              <if type="article-journal article-magazine article-newspaper paper-conference" match="any">
```
adam3smith: are you basing the title case for conference proceedings on more than the one example in the guide for authors? That seems like a very odd rule to me.

adam3smith:
could you also update the documentation link to http://aom.org/uploadedFiles/Publications/AMR/AMR%20Style%20Guide%202014.pdf

academy-of-management-review.csl
bwiernik: For this journal, most conference proceedings that are cited will be published in journal-like format (specifically the Academy of Management Proceedings), so the proceedings title should be capitalized like a journal title. I'm not sure the best way to handle both formats of proceedings correctly (as book-like proceedings should be in Title case). Would it work to test for an editor (as most journal-like proceedings don't have editors that are cited)?

adam3smith: hmmm. And you'd actually expect something like Academy of Management Proceedings to appear in databases in sentence case? I'm fine testing for editor if you want to go that way, it just seems like a lot of complexity for few cases and I always worry that those types of rules may confuse end users. But as I said, if you want to put it in, I won't object. 
Please add a comment in the code explaining that, though, so later authors can make sense of it.

rmzelle:
@bwiernik, can you make the requested changes?

bwiernik:
Yes, I will get to them in the next few days

academy-of-management-review.csl
```diff
@@ -82,30 +86,31 @@
...
+            <text variable="URL"/>
+            <choose>
+              <if variable="issued" match="none">
+                <group prefix=", ">
+                  <date variable="accessed">
+                    <date-part name="month" suffix=" "/>
+                    <date-part name="day" suffix=", "/>
+                    <date-part name="year"/>
+                  </date>
+                </group>
```
zuphilip: You don't need the `group` here, the `prefix` can go into the `date`, cf. http://citationstyles.org/downloads/specification.html#affixes .

bwiernik:
I've made the requested changes. Also fixed the formatting of book volumes and presented-at. Updated the documentation for Academy of Management Journal. 

I dropped the inclusion of conference papers in the bibliography title-casing on the assumption that the proceedings title will always be in Title Case in databases and that no styles require it to be in sentence case.

academy-of-management-review.csl
```diff
@@ -131,8 +146,19 @@
   <macro name="event">
     <choose>
       <if variable="event">
-        <text term="presented at" text-case="capitalize-first" suffix=" "/>
-        <text variable="event"/>
+        <choose>
```
adam3smith: would you mind fixing the indenting here? Everything else looks good.

bwiernik: Fixed. Sorry about that.

adam3smith:
thanks!

PR #1330, opened 2015-01-14
Create journal-of-international-business-studies.csl

zuphilip:
Style requested in https://forums.zotero.org/discussion/45591/style-request-journal-of-international-business-studies/ . I just adapted as it was suggested in the forum:
<blockquote>Basically the only difference with Academy of Management Review is that the titles of books and journals should not be bold, only italicized.</blockquote>

### DIFF
--- a/journal-of-international-business-studies.csl
+++ b/journal-of-international-business-studies.csl
@@ -1,0 +1,237 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" default-locale="en-US">
+  <info>
+    <title>Journal of International Business Studies</title>
+    <title-short>JIBS</title-short>
+    <id>http://www.zotero.org/styles/journal-of-international-business-studies</id>
+    <link href="http://www.zotero.org/styles/journal-of-international-business-studies" rel="self"/>
+    <link href="http://www.zotero.org/styles/academy-of-management-review" rel="template"/>
+    <link rel="documentation" href="http://www.palgrave-journals.com/jibs/style_guide.html#references"/>
+    <author>
+      <name>Philipp Zumstein</name>
+      <uri>https://github.com/zuphilip</uri>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="social_science"/>
+    <issn>0047-2506</issn>
+    <eissn>1478-6990</eissn>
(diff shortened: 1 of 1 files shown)


journal-of-international-business-studies.csl
```diff
@@ -0,0 +1,237 @@
...
+    </sort>
+    <layout prefix="(" suffix=")" delimiter="; ">
+      <group delimiter=", ">
+        <text macro="author-short"/>
+        <text macro="issued-year"/>
+      </group>
+      <text macro="citation-locator" prefix=":"/>
+    </layout>
+  </citation>
+  <bibliography hanging-indent="true" et-al-min="6" et-al-use-first="6" entry-spacing="0" line-spacing="2">
```
adam3smith: are the et al settings in the template? because 6/6 doesn't make sense. While citeproc-js will treat this like 6/5, technically speaking these settings mean that for items with six authors, all authors should be rendered, followed by et al.

journal-of-international-business-studies.csl
```diff
@@ -0,0 +1,237 @@
...
+        <text variable="page" prefix=": "/>
+      </else-if>
+    </choose>
+  </macro>
+  <macro name="citation-locator">
+    <group>
+      <text variable="locator" prefix=" "/>
+    </group>
+  </macro>
+  <citation et-al-min="6" et-al-use-first="1" et-al-subsequent-min="3" et-al-subsequent-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year">
```
adam3smith: this should be 7/1: "For three through six authors, give all names the first time the work is cited"

journal-of-international-business-studies.csl
zuphilip: I didn't change that part. I have seen an example for 5 names, but cannot confirm the 6/5 rule for sure. However, 6/6 does not make sense, as you are saying. I will change both here and in the template to 6/5.

adam3smith: go with 7/6 - that makes more sense given the in-text rules, at least here.

journal-of-international-business-studies.csl
zuphilip: Hm... then 6/5 does not make sense, I guess.

adam3smith: yup, see my comment below.

zuphilip:
The AMR style should be a unlimited number of authors and 3/1 for subsequent citations. Should I change it in a seperate pull request?

adam3smith:
yes, please. Thanks for checking!

PR #1331, opened 2015-01-14
Update academy-of-management-review.csl

zuphilip:
Ah, that is more complicated to read:

<blockquote>
If a work has two authors, cite both names
every time the work is cited in the text. If the
work has more than two authors, cite all authors
the first time the reference occurs; in subsequent
citations of the same work, include only the sur-
name of the first author followed by “et al.” (not
underlined) and the year.

However, for works with six or more authors, use
only the surname of the first author followed by
et al. whenever the work is cited.
</blockquote>

Thus, the citations are okay. However, there is no sentence about the et-al rule in bibliography. Therefore, I just changed it to the 6/5 setting (which should be consistent with the previous version, but be more logical csl). Okay?


adam3smith:
yeah that's great. They're not getting an award for clarity of instructions...

PR #1334, opened 2015-01-17
Fix several patterns of adjacent spaces

zuphilip:
See #1301 for more information. I will make one commit for every pattern.


zuphilip:
Ready for review

adam3smith:
thanks, went through them one-by-one, all looks safe.

PR #1336, opened 2015-01-17
Create springer-fachzeitschriften-medizin-psychologie.csl

zuphilip:
This is a general new German Springer style, see https://forums.zotero.org/discussion/45511/style-for-springer-journal-zeitschrift-fuer-rheumatologie/ . Differences to `http://www.zotero.org/styles/springer-basic-brackets-no-et-al-alphabetical` are:
 * It uses German terms like "Hrsg" and "S".
 * It follows the 5/3 et al rule.
 * It uses "et al" without a period.

I would like to connect this afterwards to ~51 German journals which I have identified.

### DIFF
--- a/springer-fachzeitschriften-medizin-psychologie.csl
+++ b/springer-fachzeitschriften-medizin-psychologie.csl
@@ -14,8 +14,8 @@
     <category citation-format="numeric"/>
     <category field="medicine"/>
     <category field="psychology"/>
-    <summary>Springer Basic style (German, numeric, brackets, alphabetical) for Springer journals in medicine and psychology (linked to EndNote_Literaturstyle_SpringerFachzeitschriftenMedizinPsychologie_StandDez2012.zip).</summary>
-    <updated>2015-01-12T21:55:01+00:00</updated>
+    <summary>Springer Basic style (German, numeric, brackets, alphabetical) for Springer journals in medicine and psychology.</summary>
+    <updated>2015-01-17T17:55:52+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="de">
@@ -151,10 +151,12 @@
           </group>
         </else-if>
         <else>
-          <group>
-            <text variable="container-title" form="short" suffix=" "/>
-            <text variable="volume" suffix=":"/>
-            <text variable="page" suffix="."/>
+          <group delimiter=": ">
+            <group delimiter=" ">
+              <text variable="container-title" form="short"/>
+              <text variable="volume"/>
+            </group>
+            <text variable="page"/>
           </group>
         </else>
       </choose>
--- a/springer-fachzeitschriften-medizin-psychologie.csl
+++ b/springer-fachzeitschriften-medizin-psychologie.csl
@@ -1,0 +1,163 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="de-DE" xmlns="http://purl.org/net/xbiblio/csl">
+  <info>
+    <title>Springer Fachzeitschriften Medizin Psychologie (German)</title>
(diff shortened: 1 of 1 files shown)


springer-fachzeitschriften-medizin-psychologie.csl
```diff
@@ -0,0 +1,163 @@
...
+    <link href="http://static.springer.com/sgw/documents/135255/application/pdf/Musterbeitrag_Uebersichten_Leitthema_Schwerpunkt.pdf" rel="documentation"/>
+    <link href="http://static.springer.com/sgw/documents/69331/application/pdf/Leitfaden_Uebersichten_Leitthemen_Schwerpunkt.pdf" rel="documentation"/>
+    <author>
+      <name>Philipp Zumstein</name>
+      <uri>https://github.com/zuphilip</uri>
+    </author>
+    <category citation-format="numeric"/>
+    <category field="medicine"/>
+    <category field="psychology"/>
+    <summary>Springer Basic style (German, numeric, brackets, alphabetical) for Springer journals in medicine and psychology (linked to EndNote_Literaturstyle_SpringerFachzeitschriftenMedizinPsychologie_StandDez2012.zip).</summary>
```
adam3smith: please remove the EndNote reference. I understand why it's there, but given the legal history, I'd rather not refer to their styles at all.

springer-fachzeitschriften-medizin-psychologie.csl
```diff
@@ -0,0 +1,163 @@
...
+          <group delimiter=", ">
+            <text variable="genre"/>
+            <text variable="publisher"/>
+          </group>
+        </else-if>
+        <else>
+          <group>
+            <text variable="container-title" form="short" suffix=" "/>
+            <text variable="volume" suffix=":"/>
+            <text variable="page" suffix="."/>
```
adam3smith: it doesn't look like we'll end up here much, but just in case, remove the period in the suffix--it seems like bibliographies don't have periods after entries, right?

adam3smith:
Looks great otherwise, thanks!

zuphilip:
Done. I styled the `else` case a little more.

zuphilip:
I created a <a href="https://gist.github.com/zuphilip/ea63fa0b350bd02f10f0">springer_journals_fchzss.tab</a> which list all corresponding journals to this citation style. I haven't found any of them in the repo already. Can someone make the necessary steps in the utils to create this dependent styles?

Please note I add additional columns: Language, Document which can be used if handy and Springer id which is maybe useful for further maintainance of the styles.

zuphilip:
Update: There are some in the repo already:
 * dependent/herzschrittmachertherapie-elektrophysiologie.csl:    <issn>0938-7412</issn>  -->  springer-basic-brackets
 * dependent/obere-extremitat.csl:    <issn>1862-6599</issn>  --> springer-basic-brackets
 * dependent/pravention-und-gesundheitsforderung.csl:    <issn>1861-6755</issn>  --> springer-basic-brackets

adam3smith:
are we thinking your list is better?

zuphilip:
Yes, I would think that these three styles are better reconnected to the new German Springer style. I checked the first one: http://www.springer.com/medicine/cardiology/journal/399?detailsPage=pltci_1381409 and an article: http://link.springer.com/article/10.1007/s00399-012-0230-x . If you want, I can check the other two also in more detail (but some time later tomorrow).

zuphilip:
I checked the other two journals as well:
 * All guidelines and information for authors suggest to use the reference style "Springer Fachzeitschriften Medizing und Psychologie".
 * The <a href="http://www.springer.com/de/unser-angebot-fuer-bibliothekare/katalog-und-preisliste">journal lists 2015 by Springer</a> declares the main language with German for these journals.
 * I checked some articles: all of them uses a Springer style with numbers in brackets sorted alphabetical. Sometimes the authors don't use et al or maybe a different rule (but this is alwas in contradiction with the specific style guide). Sometimes I only see journal articles cited without any language specific term.

Please note that the remaining 48 journals are not yet mentioned in the repo. I would vote to include dependent styles for these 48 journals as well as to change to parent of the 3 journals mentioned above. This would at least correct the order of the references for them.

PR #1337, opened 2015-01-18
Delete adjacent space in remaining cases - Part 1

zuphilip:
I don't have seen any pattern in the remaing cases. Thus, I will work through all of them and create maybe 2 or 3 pull requests as this one. I have encountered 3 cases which I feel we should discuss seperately  and I will make individual pull requests for them.

( Reference to  #1301 )

### DIFF
--- a/american-physiological-society.csl
+++ b/american-physiological-society.csl
@@ -212,14 +212,16 @@
           </choose>
         </else-if>
         <else>
-          <group delimiter=" ">
-            <text variable="container-title" form="short" font-style="italic"/>
-            <text variable="volume"/>
-            <text variable="publisher"/>
-          </group>
-          <group delimiter=", " prefix=": " suffix=".">
-            <text variable="page"/>
-            <text macro="date"/>
+          <group delimiter=": " suffix=".">
+            <group delimiter=" ">
+              <text variable="container-title" form="short" font-style="italic"/>
+              <text variable="volume"/>
+              <text variable="publisher"/>
+            </group>
+            <group delimiter=", ">
+              <text variable="page"/>
+              <text macro="date"/>
+            </group>
           </group>
         </else>
       </choose>
--- a/american-physiological-society.csl
+++ b/american-physiological-society.csl
@@ -212,11 +212,15 @@
           </choose>
         </else-if>
         <else>
-          <text variable="container-title" suffix=" " form="short" font-style="italic"/>
-          <text variable="volume"/>
-          <text variable="publisher" suffix="."/>
-          <text variable="page" prefix=": " suffix=", "/>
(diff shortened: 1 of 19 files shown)


american-physiological-society.csl
```diff
@@ -212,11 +212,15 @@
           </choose>
         </else-if>
         <else>
-          <text variable="container-title" suffix=" " form="short" font-style="italic"/>
-          <text variable="volume"/>
-          <text variable="publisher" suffix="."/>
-          <text variable="page" prefix=": " suffix=", "/>
-          <text macro="date" prefix=" " suffix="."/>
+          <group delimiter=" ">
```
adam3smith: could we wrap these two groups in another group with delimiter=": " and suffix="."? My concern is that you'd not get the "." if there are neither date nor pages (or is that too hypothetical?)

body-and-society.csl
```diff
@@ -100,8 +100,8 @@
           </if>
           <else-if variable="URL">
             <text value="URL"/>
```
adam3smith: while you're at it, why not remove the two groups with a single item, wrap everything from line 102 to 111 into a group with delimiter=" " and remove all spaces in affixes. Does that make sense?

french3.csl
```diff
@@ -208,9 +208,11 @@
       <text macro="title" suffix=". "/>
       <choose>
         <if type="thesis">
-          <text variable="genre" suffix=". "/>
-          <text macro="publisher" prefix=" "/>
-          <text variable="number-of-pages" suffix=" p."/>
+          <group delimiter=" ">
+            <text variable="genre" suffix="."/>
+            <text macro="publisher"/>
```
adam3smith: technically you're adding a space here--but if you look up at the publisher macro, it's a huge mess anyway--so unless you're super motivated, I'd just leave this as you have it here. If you want to go all out, the publisher macro would have to all go into a group with delimiters and have trailing spaces removed from affixes.

zuphilip: I hoped you will not see this ;-) I will leave it as it as now here then.

adam3smith:
OK, I'm done -- are you OK to fix the two other issues I note above?

american-physiological-society.csl
zuphilip: No problem. I can add that group. I think we are here in the last `else` case, i.e. we don't even know what strange itemType we are dealing. Thus, maybe there will also be cases as you described.

body-and-society.csl
zuphilip: Yes, I think that make sense.

zuphilip:
I fixed both issues. Thank you

adam3smith:
great, thanks!

PR #1338, opened 2015-01-18
Update currents-in-biblical-research.csl

zuphilip:
(special case from #1337)
My changes should make sense on a replacement view, but I am skeptical that the style really wants a `suffix=" ; "`. We should check/discuss carefully...

### DIFF
--- a/currents-in-biblical-research.csl
+++ b/currents-in-biblical-research.csl
@@ -43,7 +43,7 @@
             <if variable="author">
               <names variable="editor">
                 <label form="verb-short" text-case="capitalize-first" suffix=" "/>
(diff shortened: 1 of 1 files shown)


adam3smith:
yes, this should definitely always be "; " (there's a specific example in the detailed stylesheet for this case:
Gath, C.E.
        1986	A Dictionary of Biblical Heroes (4 vols; trans. W.W. Wordfinder; Chicago: Plushbottom Press).
 - ideally let's solve this with a group and a delimiter for all of these elements, but otherwise just take out the space in this suffix.

zuphilip:
Done

adam3smith:
Thanks!

PR #1339, opened 2015-01-18
Update iica-catie.csl

zuphilip:
(special case from #1337) We should especially check the handling for articles. The macro issue is called twice which seems suspicious. But I don't understand Spanish.

### DIFF
--- a/iica-catie.csl
+++ b/iica-catie.csl
@@ -319,29 +319,14 @@
             </else>
           </choose>
         </else-if>
-        <else-if type="article">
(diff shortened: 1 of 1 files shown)


adam3smith:
I'd just remove the case for articles altogether and add ```article``` as another type with article-journal in the macro below. 
In Zotero we never use article. Mendeley uses it as a fallback. The way the macro is currently written makes no sense at all, so this makes the most sense.

zuphilip:
Done

adam3smith:
Thanks!

PR #1340, opened 2015-01-18
Update methods-in-ecology-and-evolution.csl

zuphilip:
(special case from #1337 ) It is somehow hard to make the changes based on just some replacement logic here. Maybe, we should just have a look a the citation style guide and then make implement this part new.

### DIFF
--- a/methods-in-ecology-and-evolution.csl
+++ b/methods-in-ecology-and-evolution.csl
@@ -146,20 +146,18 @@
         <else-if type="chapter paper-conference" match="any">
           <text macro="title" prefix=" "/>
           <group prefix=". ">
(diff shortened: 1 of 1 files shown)


zuphilip:
I rewrote the part of chapter according to the <a href="http://www.methodsinecologyandevolution.org/view/0/authorGuidelines.html#Formatting">guidelines</a>, where there is the following example:

McArthur, W.M. (1993) History of landscape development. <i>Reintegrating Fragmented Landscapes</i> (eds R.J. Hobbs & D.A. Saunders), pp. 10-22. Springer Verlag, Berlin.

adam3smith:
looks good, thanks

PR #1344, opened 2015-01-19
fix some wrong dependents for Springer. Small tweak for Thesis

adam3smith:


zuphilip:
I guess you will update these changes in the `utils` lists as well?

PR #1346, opened 2015-01-21
Fix extra space between citation number and author in pnas.csl

simonster:

PR #1349, opened 2015-01-21
Delete prefix=" " in second field if second-field-align="flush"

zuphilip:
(reference to #1346)

Searched for
```
<bibliography[^>]*second-field-align="flush"[^>]*>.*<layout[^>]*>\r\n\s*<text variable="citation-number"[^>]*/>\r\n\s*<text[^>]*prefix="
```

### DIFF
--- a/acta-anaesthesiologica-taiwanica.csl
+++ b/acta-anaesthesiologica-taiwanica.csl
@@ -183,7 +183,7 @@
   <bibliography entry-spacing="0" second-field-align="flush" et-al-min="7" et-al-use-first="6">
     <layout suffix=".">
       <text variable="citation-number"/>
-      <text macro="author" prefix=" " suffix=". "/>
+      <text macro="author" suffix=". "/>
       <choose>
         <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
           <group delimiter=". ">
--- a/advanced-engineering-materials.csl
+++ b/advanced-engineering-materials.csl
@@ -67,7 +67,7 @@
   <bibliography entry-spacing="0" second-field-align="flush">
     <layout suffix=".">
       <text variable="citation-number" prefix="[" suffix="]"/>
-      <text macro="author" prefix=" " suffix=", "/>
+      <text macro="author" suffix=", "/>
       <choose>
         <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
           <group delimiter=", ">
--- a/advanced-materials.csl
+++ b/advanced-materials.csl
@@ -86,7 +86,7 @@
   <bibliography entry-spacing="0" second-field-align="flush">
     <layout suffix=".">
       <text variable="citation-number" prefix="[" suffix="]"/>
-      <text macro="author" prefix=" " suffix=", "/>
+      <text macro="author" suffix=", "/>
       <choose>
         <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
           <group delimiter=", ">
--- a/advances-in-complex-systems.csl
+++ b/advances-in-complex-systems.csl
@@ -77,7 +77,7 @@
     </sort>
     <layout suffix=".">
       <text variable="citation-number" prefix="[" suffix="]"/>
-      <text macro="author" prefix=" " suffix=", "/>
+      <text macro="author" suffix=", "/>
       <choose>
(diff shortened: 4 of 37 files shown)


adam3smith:
excellent, thanks!

rmzelle:
It might be good to keep track of all these problematic patterns we identified, e.g. in the wiki or so, with a short description, the regex to find the pattern, and the fixes.

We do these clean-ups once in a while, but these patterns often slowly sneak back in again.

zuphilip:
Yeah, documentation is always a good idea. Can you prepare an appropriate place in the wiki? Maybe, for a start, we can make a list of the corresponding pull requests...

rmzelle:
https://github.com/citation-style-language/styles/wiki/Quality-Control ?

adam3smith:
I like it

On Thu, Jan 22, 2015 at 9:59 AM, Rintze M. Zelle <notifications@github.com>
wrote:

> https://github.com/citation-style-language/styles/wiki/Quality-Control ?
>
> —
> <https://github.com/citation-style-language/styles/pull/1349#issuecomment-71043255>
> .
>



-- 
Sebastian Karcher, PhD
Northwestern University

zuphilip:
+1

zuphilip:
I cannot edit wikipages in the csl-styles repo. At least I don't see any "edit" button...

rmzelle:
I invited you to become a team member for the "styles" and "locales" repos, which also means you should now have direct push access to those.

If you push to the repos, please first run the Travis CI tests locally, and be careful in general. You can of course also just keep using pull requests.

adam3smith:
also, if possible enable two-factor authentication for github to be on the safe side (since this is going out directly to clients)

rmzelle:
(the alternative is to open up the wikis to everybody, but I never found a good way to get notifications when people make edits to those)

adam3smith:
FWIW, my rule of thumb is to never push anything I did in bulk, I always use pull requests from a separate branch for those. And if you don't want to set up travis locally, I'd stick with pull requests--you can merge them yourself, especially for new styles & smaller edits.

zuphilip:
Thank you for the trust in me! I happily accepted the invitation. However, I would like to continue to work as before. I don't see currently a reason for me to create direct pushs. 2Way authentification is at the moment not a good option for me, thus I will leave that as it is now.

rmzelle:
Note that you can use 2-factor authentication without text-messages. You just need a smartphone: https://help.github.com/articles/configuring-two-factor-authentication-via-a-totp-mobile-app/

zuphilip:
Yeah, but I don't have a smartphone...

rmzelle:
That's harder, then ;).

zuphilip:
I documented this pattern on the new wiki page. Hope it is close to what you had in mind.

rmzelle:
Did you fix the problem with a regex-based find and replace, or by hand? If the former, you could document those expressions as well.

zuphilip:
I did look at them manually and correct it. But actually it would have been possible to automatically replace them (I added a sentence)

rmzelle:
Some other changes that we could document:

Add legislation to bill in conditional tests
https://github.com/citation-style-language/styles/pull/126

Add match="any" to conditionals that test for multiple item types.
https://github.com/citation-style-language/styles/commit/b55a9c1704b74a35e9aa522dc4e67dc632146781

remove secondary contributors from journal-article
https://github.com/citation-style-language/styles/pull/638

remove empty prefixes, suffixes, and group delimiters
https://github.com/citation-style-language/styles/commit/65ca21b750e041d92633e21dec63fe80088c1a0e

zuphilip:
Delete unnecessary et-al-subsequent statements
#1112

rmzelle:
Hi @zuphilip, I revoked your team status again. It's not that we don't trust you, but any account without 2-factor authentication is rather sensitive to hacking (see e.g. https://github.com/blog/1698-weak-passwords-brute-forced ).

I made the wiki of the "styles" repo editable for members of the public, so you can still edit it.

If, in the future, you get a smartphone and can set up 2-factor authentication, we'd be happy to give you write access again, if you'd like that.

zuphilip:
Okay. I am still on the honor roll of CSL, that sounds good for me. I will let you know if other write access would become handy and I have the technical requirements for them.

PR #1350, opened 2015-01-22
Update iso690-numeric-en.csl

lauremelli:
Deleting the <text macro="archive_location"/> in the layout because it's already contained in the macro "archive". There was a repetition of the archive location in the final output.


rmzelle:
Thanks!

PR #1357, opened 2015-01-24
Delete adjacent space in remaining cases - Part 3

zuphilip:
( Reference to #1301 )

### DIFF
--- a/documents-d-archeologie-francaise.csl
+++ b/documents-d-archeologie-francaise.csl
@@ -595,22 +595,26 @@
           </group>
         </else-if>
         <else>
-          <group>
-            <text macro="responsability" suffix=", "/>
-            <text macro="year-date" suffix=". "/>
-            <text macro="title" suffix=". "/>
-            <text macro="medium" suffix=". "/>
-            <text macro="genre" suffix=". "/>
-            <text macro="date" suffix=". "/>
-            <text macro="edition" suffix=". "/>
-            <text macro="publisher-place" suffix="&#160;: "/>
-            <text macro="publisher" suffix=". "/>
-            <text macro="accessed" suffix=". "/>
-            <text macro="collection" suffix=". "/>
-            <text macro="page" suffix=". "/>
-            <text macro="isbn" suffix=". "/>
-            <text macro="url" suffix=". "/>
-          <text macro="archive" prefix=" (" suffix=")."/>
+          <group suffix="." delimiter=", ">
+            <text macro="responsability"/>
+            <group delimiter=". ">
+              <text macro="year-date"/>
+              <text macro="title"/>
+              <text macro="medium"/>
+              <text macro="genre"/>
+              <text macro="date"/>
+              <text macro="edition"/>
+              <group delimiter="&#160;: ">
+                <text macro="publisher-place"/>
+                <text macro="publisher"/>
+              </group>
+              <text macro="accessed"/>
+              <text macro="collection"/>
+              <text macro="page"/>
+              <text macro="isbn"/>
+              <text macro="url"/>
+              <text macro="archive" prefix=" (" suffix=")"/>
+            </group>
           </group>
         </else>
       </choose>
--- a/documents-d-archeologie-francaise.csl
+++ b/documents-d-archeologie-francaise.csl
@@ -613,7 +613,7 @@
               <text macro="page"/>
               <text macro="isbn"/>
               <text macro="url"/>
-              <text macro="archive" prefix=" (" suffix=")"/>
+              <text macro="archive" prefix="(" suffix=")"/>
             </group>
           </group>
         </else>
--- a/harvard-kings-college-london.csl
+++ b/harvard-kings-college-london.csl
@@ -251,7 +251,7 @@
         </else-if>
         <else-if type="chapter paper-conference" match="any">
           <group prefix=" " delimiter=" ">
-            <text macro="title" prefix="'" suffix="',"/>
+            <text macro="title" quotes="true" suffix=","/>
             <text term="in"/>
             <text macro="editor"/>
             <text variable="container-title" font-style="italic" suffix="."/>
--- a/harvard-kings-college-london.csl
+++ b/harvard-kings-college-london.csl
@@ -250,8 +250,8 @@
           </group>
         </else-if>
         <else-if type="chapter paper-conference" match="any">
-          <text macro="title" prefix=" '" suffix="', "/>
           <group prefix=" " delimiter=" ">
+            <text macro="title" prefix="'" suffix="',"/>
             <text term="in"/>
             <text macro="editor"/>
             <text variable="container-title" font-style="italic" suffix="."/>
@@ -265,19 +265,19 @@
           </group>
         </else-if>
         <else>
-          <group suffix=".">
-            <text macro="title" prefix=" "/>
-            <text macro="editor" prefix=" "/>
+          <group suffix=". " prefix=" " delimiter=" ">
+            <text macro="title"/>
+            <text macro="editor"/>
           </group>
-          <text variable="container-title" prefix=" "/>
-          <text macro="access" prefix=" " suffix=". "/>
-          <group prefix=" " suffix=".">
-            <group prefix=" ">
+          <text variable="container-title" suffix=" "/>
+          <text macro="access" suffix=". "/>
+          <group suffix="." delimiter=" ">
+            <group>
               <text variable="volume"/>
               <text variable="issue" prefix=" (" suffix=")"/>
               <text macro="issued"/>
             </group>
-            <group prefix=" ">
+            <group>
               <label variable="locator" form="short"/>
               <text macro="pages"/>
             </group>
--- a/norma-portuguesa-405.csl
+++ b/norma-portuguesa-405.csl
@@ -309,8 +309,8 @@
           <date-part name="month" form="short" suffix=". " text-case="lowercase"/>
           <date-part name="year" suffix="). "/>
         </date>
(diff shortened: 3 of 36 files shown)


documents-d-archeologie-francaise.csl
```diff
@@ -595,22 +595,26 @@
...
+              <group delimiter="&#160;: ">
+                <text macro="publisher-place"/>
+                <text macro="publisher"/>
+              </group>
+              <text macro="accessed"/>
+              <text macro="collection"/>
+              <text macro="page"/>
+              <text macro="isbn"/>
+              <text macro="url"/>
+              <text macro="archive" prefix=" (" suffix=")"/>
```
adam3smith: this is an extra space, too, right? remove.

harvard-kings-college-london.csl
```diff
@@ -250,8 +250,8 @@
           </group>
         </else-if>
         <else-if type="chapter paper-conference" match="any">
-          <text macro="title" prefix=" '" suffix="', "/>
           <group prefix=" " delimiter=" ">
+            <text macro="title" prefix="'" suffix="',"/>
```
adam3smith: could you change this to quotes="true" (the style is already en-GB, so they'll be single quotes with comma outside)

norma-portuguesa-405.csl
```diff
@@ -309,8 +309,8 @@
           <date-part name="month" form="short" suffix=". " text-case="lowercase"/>
```
adam3smith: while we're at it, the parentheses should really be set as a prefix/suffix on the ```date``` in case any of the ```date-part``` are missing.

harvard-kings-college-london.csl
zuphilip: If I understand correctly that would then be
```
<text macro="title" quotes="true" suffix=","/>
```
Is this correct?

adam3smith: yes

philosophia-scientiae.csl
```diff
@@ -280,9 +280,9 @@
             <text term="volume" form="short" suffix=" "/>
             <number variable="volume" form="numeric"/>
           </group>
-          <group suffix="; ">
-            <number variable="number-of-volumes" form="numeric" suffix=" "/>
-            <text term="volume" form="short" prefix=" " suffix=" " plural="true"/>
+          <group suffix=" ; " delimiter=" ">
```
adam3smith: That space there isn't actually right given the styleguide:
     

> Boltzmann, Ludwig [1909], Wissenschaftliche Abhandlungen von Ludwig Boltzmann, 3 volumes; New York: Chelsea Publishing Company, 1968.

adam3smith: also, delete the form="short" as per the same example.

proceedings-of-the-royal-society-b.csl
```diff
@@ -49,10 +49,8 @@
   </macro>
   <macro name="publisher">
```
adam3smith: given how popular this style is, we should also do the publisher/publisher place thing right, i.e. with a group with delimiter=": "

revue-archeologique-du-centre-de-la-france.csl
```diff
@@ -277,7 +277,7 @@
         <choose>
           <if type="thesis">
             <text variable="genre" suffix=". "/>
-            <text macro="publisher" prefix=" "/>
+            <text macro="publisher"/>
```
adam3smith: while I was looking at this style -- we should run a sed script to remove all empty affixes (i.e. prefix="" and suffix="") Since those never do anything, that can just be scripted and separate. @rmzelle might even consider including it in his general maintenance script?

risk-analysis.csl
```diff
@@ -102,17 +102,17 @@
...
             <text term="in" text-case="lowercase" suffix=" "/>
             <text macro="editor" suffix=". "/>
-            <text variable="container-title" text-case="title" prefix=" " suffix="."/>
-            <text variable="volume" prefix="Vol " suffix="."/>
-            <text macro="edition" prefix=" "/>
-            <text variable="collection-title" prefix=" " suffix="."/>
-            <group suffix=".">
-              <text macro="publisher" prefix=" "/>
-              <group suffix="." prefix=", ">
+            <text variable="container-title" text-case="title" suffix="."/>
```
adam3smith: this must be suffix=". "--no way the volume (or publisher or whatever) starts after a period without a space.

standards-in-genomic-sciences.csl
```diff
@@ -182,7 +182,7 @@
           <text macro="editor" prefix=" " suffix="."/>
           <group prefix=" " suffix=".">
             <text variable="container-title" font-style="italic" form="short" suffix=" "/>
-            <group delimiter=";" prefix=" ">
+            <group delimiter=";">
               <date variable="issued">
                 <date-part name="year"/>
               </date>
```
adam3smith: another thing we could consider removing via script: groups with a single element. Sed probably isn't a good choice here, but python should work nicely. cc @rmzelle

the-british-journal-of-sociology.csl
```diff
@@ -167,9 +167,9 @@
           </group>
         </else-if>
         <else>
-          <group suffix="', ">
-            <text macro="title" prefix=" '"/>
-            <text macro="editor" prefix=" "/>
+          <group prefix=" '" suffix="', " delimiter=" ">
```
adam3smith: this won't typically actually occur, but obviously we don't want quotes around the editor.
I'd like the quotes="true" to be handled to the title macro (for the else case in l. 83) and then removed from the rest of the style.

wheaton-college-phd-in-biblical-and-theological-studies.csl
```diff
@@ -331,8 +333,10 @@
         <text term="in" text-case="capitalize-first" suffix=" "/>
         <choose>
           <if variable="volume" match="any">
-            <text term="volume" form="short" suffix=" "/>
-            <text variable="volume" prefix=" " suffix=" of "/>
+            <group delimiter=" ">
+              <text term="volume" form="short"/>
+              <text variable="volume" suffix=" of "/>
```
adam3smith: @dowens76 -- could you take a quick look at this? This looks like number-of-volumes is missing here?

adam3smith:
OK, sorry that most of these notes are actually fixes to things that were wrong before you did this, but I figured while I was reviewing this I might as well catch obvious errors. Let me know if that got too much, I can take some of them.

standards-in-genomic-sciences.csl
zuphilip: Yeah, I had the same thought. But this can become more tricky if the only child node has children itself. Moreover, AFAIK one can use `group` also for a conditional existence statement.

adam3smith: > But this can become more tricky if the only child node has children itself.

Does it though? I can't think of a case where that'd make a difference. Obviously if the group has affixes, those need to be passed down to the child.

> Moreover, AFAIK one can use group also for a conditional existence statement.

But only in a meaningful way if there are more than one child items, right? The conditional is that if you have a term and a variable in a group, the term only gets printed when the variable does. Or what did you have in mind?

zuphilip: 1) Finding the relevant cases might be a different. I would guess that for that you need to go trough the XML structure and looking for group nodes with only one child. (Sure, this is doable, but it is different from find&replace of a regexp).
2) You are right. That was a false thought from my side...
3) One would have to merge all attributes from the group and its child node. I am not sure if this is possible automatically.

adam3smith: right, that's why I was thinking python, which has a nice XML parser, which Rintze uses for the re-indent/reorder script anyway. I agree a regex would probably not be feasible, at least not reasonably so.

zuphilip:
I updated the commit. This should contain all your comments and I added another style: theory-culture-and-society.csl , for which I had first do look at the documentation. I appreciate the possibility to improve on some styles and I think such search pattern can lead us to the more problematic structured implementations. It is worth to spend some time to improve on them. So, no problem.

adam3smith:
man, those new Travis builds are fast @rmzelle @inukshuk --thanks again for this.
Looks great now, thanks.

wheaton-college-phd-in-biblical-and-theological-studies.csl
dowens76: I am no longer studying there, but based on the current style guide at http://www.wheaton.edu/~/media/Files/Academics/Departments/BTS/PhD%20Program/Style_Guide.pdf and my memory, the only place you include the number of volumes is when you are citing a work as a whole or in the bibliography. But just by looking at the commit I cannot tell if this creates a problem or not. Here is an example of how things should be: 

Note: G. E. Ladd, “Kingdom of God,” in vol 3 of *The International Standard Bible Encyclopedia*, ed. Geoffrey W. Bromiley, rev. ed. (Grand Rapids: Eerdmans, 1986), 29.
Bibliography: G. E. Ladd, “Kingdom of God.” In vol 3 of *The International Standard Bible Encyclopedia*, edited by Geoffrey W. Bromiley, 23–39. Rev. ed. Grand Rapids: Eerdmans, 1986. 

But this commit does not change that behavior, right?

zuphilip: No, the commit didn't change that behaviour and it seems to conform with your examples: vol 3 of <i>Title</i>. I think @adam3smith has tought it could also be e.g. vol 3 of 6 (then the 6 = number of volumes would have been missing).

adam3smith: yup, exactly. This is fine as is then.

PR #1358, opened 2015-01-24
Update university-of-helsinki-faculty-of-theology.csl

zuphilip:
(ref to #1301)
I don't know how I should change the other cases in the `names`

### DIFF
--- a/university-of-helsinki-faculty-of-theology.csl
+++ b/university-of-helsinki-faculty-of-theology.csl
@@ -279,34 +279,34 @@
   <macro name="publisher">
     <choose>
       <if type="bill graphic legal_case legislation motion_picture report song" match="any">
-        <group prefix=" ">
-          <group suffix="; ">
+        <group prefix=" " suffix=" " delimiter="; ">
+          <group delimiter=" ">
             <text variable="collection-title"/>
-            <text variable="collection-number" prefix=" "/>
+            <text variable="collection-number"/>
           </group>
-          <group suffix="; ">
-            <text term="volume" form="short" suffix=" "/>
+          <group delimiter=" ">
+            <text term="volume" form="short"/>
             <number variable="volume" form="numeric"/>
           </group>
-          <group suffix="; ">
-            <number variable="number-of-volumes" form="numeric" suffix=" "/>
-            <text term="volume" form="short" prefix=" " suffix=" " plural="true"/>
+          <group delimiter=" ">
+            <number variable="number-of-volumes" form="numeric"/>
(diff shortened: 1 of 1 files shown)


zuphilip:
```
 <if variable="author">
<names variable="editor">
<label form="verb-short" text-case="capitalize-first" suffix=" "/>
<name prefix=" " and="text" initialize-with=".."/>
</names>
</if>
```

rmzelle:
You don't know what to do because you can't use a delimiter here?

I would keep the suffix on `<label/>` and get rid of the prefix on `<name/>`. A name label will only print if there is a name, but a label term might be empty/non-defined.

adam3smith:
That's a really odd initialization think, but unfortunately we don't have a style guide...

university-of-helsinki-faculty-of-theology.csl
```diff
@@ -279,34 +279,34 @@
   <macro name="publisher">
     <choose>
       <if type="bill graphic legal_case legislation motion_picture report song" match="any">
-        <group prefix=" ">
-          <group suffix="; ">
+        <group prefix=" " suffix=" " delimiter="; ">
```
adam3smith: this needs a semicolon in the suffix, since in the original there was one as part of the last group.

zuphilip:
I think I mostly would appreciate if we can speack the changes trough together. Okay? Let us maybe start with no. 3
```
                <if variable="container-author">
                  <group>
                    <names variable="container-author">
                      <label form="verb-short" prefix=" " suffix=" " strip-periods="true"/>
                      <name and="text" prefix=" " delimiter=", " initialize-with=". "/>
                    </names>
                  </group>
                </if>
```
 * delete the prefix and let the suffix as it is
 * the group is not needed, thus I can delete it
 * the if is also not needed, right?
 * anything else?

adam3smith:
that's all correct.

rmzelle:
For no. 3, it would be better to move the prefix on `<label/>` to `<names/>` (assuming it's needed).

zuphilip:
Okay, I updated the other case in a similar spirit. Please have a look.

BTW I think the only case left for adjacent spaces is https://github.com/citation-style-language/styles/blob/master/spip-cite.csl#L67 . I am not sure what exactly this style is for. Is it okay as it is? Or should I delete the `suffix=" "` on this line?

adam3smith:
it's for an online publishing system. Leave this as is.

zuphilip:
Okay. Travis is happy here. Let me know what you think about the last one. Yes, I think this change here would close #1301 for now. :satisfied:

PR #1359, opened 2015-01-27
Updates to two styles

myurkin:
Mainly related to handling the short form of container-title in combination with automatic abbreviation in Word processor.


rmzelle:
@adam3smith, do you agree with this solution? Or should we ask Zotero to limit automatic abbreviation to journal titles?

adam3smith:
I think this is the better solution. We want to at least keep the option open for other titles to be abbreviated via other lists. So yes, I agree, this is what we're doing in other styles, too.

myurkin:
The behavior of abbreviations should be partly controlled by styles, i.e. some style standards may specify that book titles should be abbreviated (when a chapter is referenced) or not.

But it would also help if the Word plugin has more flexibility with regards to using abbreviation lists. 
For instance, a separate option to use MEDLINE list only for journal titles (as is written in its description) will help a lot of users right away. Or alternatively it can be made to work always only with journal titles (until there will be cases, when abbreviating other container-title is useful).

rmzelle:
I guess we can merge this, then.

@myurkin, any reason you changed the indentation on the XML comments? Or is that just something your text editor did? See https://github.com/citation-style-language/styles/pull/1359/files

myurkin:
Concerning the xml comments, yes, I have restyled them on purpose. But it was not about indentation but rather removing manual line breaks. So now this comments are "one paragraph - one line". That makes it easier to make further changes in them.

rmzelle:
Thanks!

PR #1360, opened 2015-01-27
Update Revue française de sociologie, add English version

nalimilan:


rmzelle:
I'll look at this later, but generally we prefer to have separate pull requests for style updates and new style submissions.

It's usually also handier to submit from a new branch, instead of using your "master" branch. That way you can easier keep your "master" branch up to date.

nalimilan:
OK, I've made separate PR for two reviews.

But I have a problem with the RFS English style: I cannot find an ISSN different from the one for the French version, and Travis doesn't like that. What should I do?

adam3smith:
we can take care of that part--there's an exception list for Travis for such cases. Just to confirm--the English and French style differ by other things than language, correct?

nalimilan:
Cool. The two styles are identical except for a few localization issues: translation of "Ph.D. thesis", different quotes, quotes outside of punctuation.

adam3smith:
I think that's better handled with a dependent style then, so we avoid duplicate code. It's possible to override the locale of the parent style in a dependent style and that'd be the way to go. I'll try to look at this tonight with more details, but it could take me a little longer, so @rmzelle if you have time, please have a quick look.

nalimilan:
Ah, would make more sense indeed. Thanks for looking at this!

nalimilan:
I'll leave this PR open for the translation issue, and open a new PR for the update to the French style.

rmzelle:
@nalimilan, ISSNs that may appear multiple times are listed at https://github.com/citation-style-language/styles/blob/master/spec/spec_helper.rb#L8, so we could add "0035-2969" to that list.

As for the localization, you would create a dependent style with its locale set to English. Note that dependent styles themselves can't define locale data such as term translations. If you find that you need more customization than automatically happens through the use of the CSL locale files, you can put additional English localization data in the French independent parent style, e.g. with:

```
<style ...>
  <locale xml:lang="en">
    <terms>
      <term name="retrieved">available</term>
    </terms>
  </locale>
  <locale xml:lang="fr">
    <terms>
      <term name="retrieved">disponible</term>
    </terms>
  </locale>
</style>
```

(example from vancouver.csl)

See https://forums.zotero.org/discussion/36595/defaultlocale-of-dependent-doesnt-override-that-of-its-parent/?Focus=207473#Comment_207473 for a few more examples of dependent styles that have a different default-locale from their parent.

nalimilan:
Thanks. One problem I had is that I wanted to show "Ph.D. thesis" and its translation, but there's no standard term for that. I tried `<term name="thesis">`, but it didn't pass the CSL validator. Is there a solution?

rmzelle:
I think the "genre" variable is usually used for that. (at least for Zotero)

adam3smith:
the style uses a generic term where genre is empty. That's not unreasonable and I do think we should localize all item types in the medium term. 
Until then, I would still tend to say I'd prefer to just have one version of this style and have wrong localization for theses without genre/type.

nalimilan:
Here's the new dependent style.

rmzelle:
Looks great, thanks!

PR #1364, opened 2015-01-28
add original date and status to APA style

adam3smith:
@bwiernik @nickbart1980 could you take a quick looks if this looks right to you? Seemed pretty straightforward for APA. If you think this is OK, I'll add this to all APA variants.
I'll do Chicago, which will be a little trickier, next. 
@rmzelle - I'm hoping to add these to a couple of styles: for pandoc users, for people who want to use the Zotero/citeproc workarounds, and for other reference managers that are interested in implementing this. I assume there are no objections?

### DIFF
--- a/apa.csl
+++ b/apa.csl
@@ -281,6 +281,12 @@
               </choose>
             </group>
           </if>
+          <else-if variable="status">
+            <group prefix=" (" suffix=")">
+              <text variable="status"/>
+              <text variable="year-suffix" prefix="-"/>
+            </group>
+          </else-if>
           <else>
             <group prefix=" (" suffix=")">
               <text term="no date" form="short"/>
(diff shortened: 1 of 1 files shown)


apa.csl
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
-  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <!-- This style was edited with the Visual CSL Editor (http://editor.citatinstyles.org/visualEditor/) -->
```
rmzelle: Typo?

rmzelle:
Sure.

apa.csl
adam3smith: yeah, happened while saving. I'll take the line out given that this has now gone many iterations w/o the editor.

adam3smith:
OK, advanced online is added. CSL automatically sorts items with no date last (no matter the sort order), so that's already working correctly automatically.
I'm happy with this now, unless there are further objections, I'm going to accept his later today.

bwiernik:
Looks good to me.

adam3smith:
thanks everyone

PR #1367, opened 2015-01-29
Chicago ad orig date

adam3smith:
add original date and status to chicago-author-date. @nickbart1980 FYI.


adam3smith:
no objections, looks like this should work.

PR #1373, opened 2015-01-30
Update exercer.csl

paulfrappe:
Some users reported a bug in the hyphen delimiting pages, wich appears as a square.  
Insertion in line 26: <term name="page-range-delimiter">-</term>

### DIFF
--- a/exercer.csl
+++ b/exercer.csl
@@ -23,6 +23,7 @@
       <date-part name="year"/>
     </date>
     <terms>
+      <term name="page-range-delimiter">-</term>
       <term name="presented at">présenté à</term>
       <term name="retrieved">disponible</term>
       <term name="from">sur</term>


adam3smith:
can you replicate this issue? Because that shouldn't be the case and iI'f that's a general issue, I'd rather fix this properly. The French locale uses a non-breaking hyphen here, which is my understanding is preferable to the regular hyphen by French standards (it prevents page ranges to split over line breaks).

paulfrappe:
I don't have the bug on my PC. Two users sent me messages about this issue. The second one had Zotero for Firefox and the plugin Word for windows 3.1.8.

adam3smith:
I'll see if I can replicate this. 
One thing to check for would be non-standard fonts. It's possible that the character isn't defined in poorly designed fonts. The other issue is old Zotero versions--there was a bug that affected this several Zotero versions back, so if people haven't updated, they might be seeing that.

rmzelle:
@paulfrappe, so it would be helpful if you could ask your users to place the cursor in the bibliography where the problem occurs, and tell us which font is used for the bibliography. Version of Word and OS might help as well.

Are Zotero and the word plugin up to date? The latest Zotero Word plugin for Windows is 3.1.18, not 3.1.8. (see https://www.zotero.org/support/word_processor_plugin_installation)

paulfrappe:
Dear Mr Zelle, 
I sent a message to all my friends/users of exercer.csl. At this moment, I did not received any message about another user who had this bug. 
Be sure I will send you any message that come back to me. 
Yours sincereleyPaul Frappé

      De : Rintze M. Zelle <notifications@github.com>
 À : citation-style-language/styles <styles@noreply.github.com> 
Cc : paulfrappe <paul.frappe@yahoo.fr> 
 Envoyé le : Samedi 31 janvier 2015 18h09
 Objet : Re: [styles] Update exercer.csl (#1373)
   
@paulfrappe, so it would be helpful if you could ask your users to place the cursor in the bibliography where the problem occurs, and tell us which font is used for the bibliography. Version of Word and OS might help as well.Are Zotero and the word plugin up to date? The latest Zotero Word plugin for Windows is 3.1.18, not 3.1.8. (see https://www.zotero.org/support/word_processor_plugin_installation)—

adam3smith:
Since we haven't heard back, I'm going to close this. If you do get more confirmed errors of this, we can either re-open this or handle it in a separate bug report, just post here.

PR #1376, opened 2015-02-01

zuphilip:
There should be a point in citations in "et al." but no point in references in "et al". This can be found in the the guidelines and also in articles from the corresponding journals. Or do I miss something here? (The use of the other term "and others" defined to a variant of "et al" seems to be the only fix which CSL let me do...)


adam3smith:
looks good.

PR #1383, opened 2015-02-05
Create informationspraxis.csl

zuphilip:
This new Open Access journal from library, archive and information community chooses an already existing citation style. They will apply for an ISSN and when this happened, the information can be added here. Any other suggestion for the category?

### DIFF
--- a/dependent/informationspraxis.csl
+++ b/dependent/informationspraxis.csl
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" default-locale="de-DE">
+  <info>
+    <title>Informationspraxis (German)</title>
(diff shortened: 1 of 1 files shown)


rmzelle:
No, looks good!

rmzelle:
I assume you'll update the "documentation" link as well once the journal has a website?

zuphilip:
The first documentation link is fine: https://journals.ub.uni-heidelberg.de/index.php/ip/about/submissions#authorGuidelines and there the pdf from the second link is explicitely mentioned. Do you prefer any change?

There is the blog (with the open review): http://informationspraxis.de/ and there is the journal homepage https://journals.ub.uni-heidelberg.de/index.php/ip/index which collect the articles in issues etc. I see that the ISSN seems to be ready, I will update that later.

rmzelle:
Oh, never mind, then. I missed the first link.

PR #1387, opened 2015-02-06
Added dependent styles for Academy of Management

bwiernik:
Adding dependent styles for additional Academy of Management publications


rmzelle:
You checked that these all have the same citation format?

bwiernik:
Yes. Perspectives uses APA 5th. The others use the same style as Review/Journal. Some of the documentation is very sparse, but the examples look the same as the Review style.

rmzelle:
Okay, thanks!

PR #1391, opened 2015-02-08
Update invisu.csl

amendesd:
Because many people have asked me, I have introduced changes for pages in article-journal type, and other minor corrections for the rest of the style.
Thanks

### DIFF
--- a/invisu.csl
+++ b/invisu.csl
@@ -448,13 +448,14 @@
         </group>
       </else-if>
       <else-if type="graphic personal_communication interview webpage" match="any"/>
-       <choose>
+      <else-if type="article-journal article-magazine" match="any">
+        <choose>
           <if is-numeric="page">
             <group delimiter=" " font-style="normal" prefix=", ">
-              <group delimiter="&#160;" suffix=".">
-                <label variable="page" form="short"/>
-                <text variable="page"/>
-              </group>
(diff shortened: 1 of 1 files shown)


invisu.csl
```diff
@@ -448,13 +448,19 @@
...
       </else-if>
       <else-if type="graphic personal_communication interview webpage" match="any"/>
-      <else-if type="article-journal article-magazine" match="any">
-        <group delimiter=" " font-style="normal" prefix=", ">
-          <group delimiter="&#160;" suffix=".">
-            <label variable="page" form="short"/>
-            <text variable="page"/>
-          </group>
-        </group>
+       <choose>
```
adam3smith: not clear what you're trying to do here, but you need to define an ```<else-if>``` condition before you can put in that ```<choose>``` presumably one that corresponds to line 464 (the orphaned ```</else-if>```) below.

adam3smith:
Even after you fix the error, Travis is still going to complain about the unused (because commented out) ISSN and ISBN macros. The easiest solution would be to take those two lines out of the macros (and delete the macros) and write them into the bibliography directly (commented out, of course).

amendesd:
Thanks for your comments. I corrected the mistakes and I think it's ok now. I finally decided to remove ISSN and ISBN. Regards.

PR #1392, opened 2015-02-08

FrancoisC37:
It's corrections for editor name. I do it like for the authors.
And corrections for the use of "p." (1 page) or "pp." (two or more pages).
Thanks!
François


rmzelle:
Great, thanks!

PR #1393, opened 2015-02-09
Only display 'in' if editor variable is set

ms609:
Applies to books and equivalent contribution types

### DIFF
--- a/palaeontology.csl
+++ b/palaeontology.csl
@@ -12,14 +12,14 @@
     </author>
     <contributor>
       <name>Martin R. Smith</name>
-      <email>martins@gmail.com</email>
+      <email>ms609@cam.ac.uk</email>
       <uri>https://plus.google.com/108450310917386384868</uri>
     </contributor>
     <category citation-format="author-date"/>
     <category field="biology"/>
     <issn>0031-0239</issn>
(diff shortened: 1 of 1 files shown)


palaeontology.csl
```diff
@@ -106,8 +106,12 @@
...
                 <text variable="title" font-style="italic" text-case="capitalize-first" suffix=". "/>
-                <text value="In" font-style="italic" suffix=" "/>
-                <text macro="editor" suffix=" "/>
+                <choose>
+                  <if variable="editor">
+                    <text value="In" font-style="italic" suffix=" "/>
+                    <text macro="editor" suffix=" "/>
+                  </if>
+                </choose>
                 <group>
```
adam3smith: Let's do this properly:
```<group delimiter=" " suffix=" ">
  <text term="in" text-case="capitalize-first" font-style="italic"/>
  <text macro="editor">
</group>```
(if a group contains a variable, it only gets printed if that variable is present; also, terms are preferable to "text value" for versatility of styles down the road).

rmzelle: @adam3smith, I edited your comment to put the "`"s on separate lines for proper formatting.

@ms609, let us know if you need help with this.

PR #1394, opened 2015-02-09
Create ieee-with-url-no-access-date.csl

wbt:
Added style based on ieee-with-url which omits access dates, for where that is appropriate.

### DIFF
--- a/ieee-with-url-no-access-date.csl
+++ b/ieee-with-url-no-access-date.csl
@@ -1,0 +1,335 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
+  <info>
+    <title>IEEE (with URL, no access date)</title>
+    <id>http://www.zotero.org/styles/ieee-with-url-no-access-date</id>
+    <link href="http://www.zotero.org/styles/ieee-with-url-no-access-date" rel="self"/>
+    <link href="http://www.ieee.org/documents/style_manual.pdf" rel="documentation"/>
+    <link href="http://www.ieee.org/documents/auinfo07.pdf" rel="documentation"/>
+    <author>
+      <name>Michael Berkowitz</name>
+      <email>mberkowi@gmu.edu</email>
+    </author>
+    <contributor>
+      <name>Julian Onions</name>
+      <email>julian.onions@gmail.com</email>
+    </contributor>
+    <contributor>
+      <name>Rintze Zelle</name>
+      <uri>http://twitter.com/rintzezelle</uri>
+    </contributor>
+    <contributor>
+      <name>Stephen Frank</name>
+      <uri>http://www.zotero.org/sfrank</uri>
+    </contributor>
+    <contributor>
+      <name>W. Ben Towne</name>
+      <uri>https://www.zotero.org/wbt</uri>
+    </contributor>
+    <contributor>
+      <name>Sebastian Karcher</name>
+    </contributor>
+    <category citation-format="numeric"/>
+    <category field="engineering"/>
+    <category field="generic-base"/>
+    <updated>2014-09-03T18:33:58+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="en">
+    <terms>
+      <term name="chapter" form="short">ch.</term>
+      <term name="presented at">presented at the</term>
+      <term name="available at">available</term>
+    </terms>
+  </locale>
+  <!-- Macros -->
+  <macro name="edition">
+    <choose>
+      <if type="bill book chapter graphic legal_case legislation motion_picture paper-conference report song" match="any">
+        <choose>
+          <if is-numeric="edition">
+            <group delimiter=" ">
+              <number variable="edition" form="ordinal"/>
+              <text term="edition" form="short"/>
+            </group>
+          </if>
+          <else>
+            <text variable="edition" text-case="capitalize-first" suffix="."/>
+          </else>
+        </choose>
+      </if>
+    </choose>
+  </macro>
+  <macro name="issued">
+    <choose>
+      <if type="article-journal report" match="any">
+        <date variable="issued">
+          <date-part name="month" form="short" suffix=" "/>
+          <date-part name="year" form="long"/>
+        </date>
+      </if>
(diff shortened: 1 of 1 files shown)


rmzelle:
> for where that is appropriate.

Where is this appropriate? Is there a specific journal that has this requirement?

wbt:
Not a journal that I'm aware of, but it is sometimes a good option for reports, papers outside traditional peer-reviewed publications, or where character count is limited and the bibliography access dates give the reader the least amount of useful information, but the URL is helpful so the reader can quickly and easily access the original source.
  
Access dates are not always needed according to [APA style](http://www.apastyle.org/learn/faqs/when-include-retrieval-date.aspx) and it's better form to use permalinks, embedding access date information in the URL so that a reader could access the old content directly for any verification, instead of having to figure out how to go back in time to archived versions him/herself (e.g. using an oldid parameter with Wikipedia links).  Sometimes a numbered format is better than APA, and a URL is useful for readers, but access dates may be less useful.

rmzelle:
I agree that access dates are overrated, but we'll probably not accept this style.

Maintaining parallel versions of popular styles is a significant time drain for our volunteer maintainers, so we try to avoid create variants like this one unless absolutely necessary.

wbt:
Well, I personally find it more useful than going through the bibliography and manually resetting/deleting those access dates every time there's a small change in the document, and more reliable than the risk that I'll forget automatic bibliography updates are disabled by the edits. If you think that nobody else will have a similar opinion, then it makes sense to reject the contribution... I was just trying to help in case somebody else out there had a similar situation.

rmzelle:
I don't disagree that it might make sense for some people to use this variant, but there is a clear cost associated with adding style variants to our repository. Our time is arguably better spend on making it easier for users to make these kind of changes to their styles. And you're of course always welcome to distribute this style yourself, e.g. via your blog.

Sorry I couldn't be more helpful!

wbt:
You may want to state on your submissions instructions pages that you only want style files submitted if they are required for some archival-quality peer reviewed publication venue (or whatever the criteria actually is, if that's not it) to save folks like you and I our time and effort in submitting and reviewing what isn't acceptable.

zuphilip:
It is already there https://github.com/citation-style-language/styles/blob/master/CONTRIBUTING.md:

"We're particularly interested in styles for journals and published style guides."

wbt:
My reading of that was "if you're just thinking about writing a new style file and want some direction/pointers as to which ones would be most useful, here's a place to look." (I also expected to find a link to https://forums.zotero.org/11/ as described [here](https://github.com/citation-style-language/styles/wiki/Requesting-Styles)).
I did not at all think that meant "we only accept contributions of styles for official journals and published style guides."

rmzelle:
We were considering changing that to: "We're particularly interested in styles for journals 
and published style guides. We reserve the right to refuse styles that 
only appeal to a small audience, and styles for journals from 
[predatory publishers](http://en.wikipedia.org/wiki/Predatory_open_access_publishing)." (per http://xbiblio-devel.2463403.n2.nabble.com/predatory-publishers-tt7579203.html )

We often discuss style variants like yours on a case-by-case basis, so I don't want the text to be too discouraging towards new contributors.

adam3smith:
we'll tighten the language. I don't see why we'd be linking to the Zotero
forums, though; that's specifically for style requests.

On Thu, Feb 12, 2015 at 8:54 AM, wbt <notifications@github.com> wrote:

> My reading of that was "if you're just thinking about writing a new style
> file and want some direction/pointers as to which ones would be most
> useful, here's a place to look." (I also expected to find a link to
> https://forums.zotero.org/11/ as described here
> <https://github.com/citation-style-language/styles/wiki/Requesting-Styles>
> ).
> I did not at all think that meant "we only accept contributions of styles
> for official journals and published style guides."
>
> —
> <https://github.com/citation-style-language/styles/pull/1394#issuecomment-74083345>
> .
>



-- 
Sebastian Karcher, PhD
Northwestern University

wbt:
>I don't see why we'd be linking to the Zotero forums, though; that's specifically for style requests.
If somebody is considering becoming a new contributor because they like the project but don't have particular work in mind and want to know what contributions the community is explicitly asking for, it makes sense to point people to the place where folks make such requests.  

>We often discuss style variants like yours on a case-by-case basis, so I don't want the text to be too discouraging towards new contributors.

The most encouraging results would come from having the text that matches what rules or guidance you use for actual reviews and discussions.  My personal rule before investing the effort to make this contribution was seeing that it would have been useful to me on at least three very different projects, which to me hints that it's more useful than a one-off...it's impossible to know for sure whether or not it's something that would appeal to a broader audience without making it available and easily discoverable to that audience.

rmzelle:
> The most encouraging results would come from having the text that matches what rules or guidance you use for actual reviews and discussions.

Yeah, but our criteria are somewhat fluid, so I rather just say "no" a few times a year to submissions and leave our submissions guidelines as compact as there currently are. If we have too much text people won't read it anyway.

> It's impossible to know for sure whether or not it's something that would appeal to a broader audience without making it available and easily discoverable to that audience.

We don't have easy access to usage statistics for our styles, though. And we can't just make every submitted style variant available, which would make the repository unmanageable.

wbt:
So what are the criteria then, at least now?  Is "fluid" just a nice way of saying "arbitrary?"

rmzelle:
Pretty much.

Established journals (i.e., those with a registered ISSN) all deserve a style without question. Style guides like IEEE and Chicago Manual of Style often give authors some freedom in formatting their citations. In this case we decide case-by-case which variations deserve their own style. Here we're more lenient with the more popular style guides, and broad differences (like an "author-date" and "numeric" variant) are more likely to be considered deserving of separate styles than minor differences (like the presence or absence of access dates). It also depends on which variants we already have in the repository.

wbt:
If a potential new contributor is trying to decide whether or not to be a contributor in this community, seeing arbitrary decisions about acceptance/rejection is rather discouraging.  When that arbitrary decision comes down on the side of rejection, it can be a frustrating experience for everyone involved and demonstrate to the contributor that the community is opposite-of-welcoming.  When it is really arbitrary and a contributor can see other examples where the decision went the other way (because you have no basis for consistency in decisions), the illusion of a technical meritocracy goes out the window and a contributor might be left to wonder what other factors (personal, etc.) were behind the decision (personal factors would carry over to other potential contributions, so that matters), searching for some logical reasoning or decision rule which may (apparently does not?) actually exist. 

If you can figure out what guidelines you actually use to make those decisions and post those, you might find better self-selection for valuable contributions and give everybody a better experience.

If you want to reduce the amount of text in instructions, you could even put these guidelines on a separate page use link text like "How do we choose?" where submission instructions describe the review step and probably in the early part where you describe what you're looking for.

adam3smith:
Thanks. We'll take that into account.

PR #1398, opened 2015-02-12

roaldfrosig:
Use editor name in citations when no author is given.


rmzelle:
Thanks!

PR #1401, opened 2015-02-12
Update associacao-brasileira-de-normas-tecnicas-ufmg-face-full.csl

diogopereirario:
Correção de variável que não permitia aparecer o número de páginas na bibliografia de uma tese. 

(de 'number-of-pages' para 'page')

### DIFF
--- a/associacao-brasileira-de-normas-tecnicas-ufmg-face-full.csl
+++ b/associacao-brasileira-de-normas-tecnicas-ufmg-face-full.csl
@@ -511,7 +511,7 @@
             <text macro="author" suffix=". "/>
             <text macro="title" suffix=". "/>
             <text macro="issued-year" suffix=". "/>
-            <text variable="number-of-pages" suffix=" f. "/>
+            <text variable="page" suffix=" f. "/>
             <text variable="genre" suffix=" &#8211; "/>
             <text variable="publisher" suffix=", "/>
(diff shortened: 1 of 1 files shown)


adam3smith:
that looks wrong to me. Thesis should have number-of-pages, not page as a variable. Which software are you using this with?

diogopereirario:
Mendeley. According to a mendeley employee this is what should be done: http://support.mendeley.com/customer/portal/questions/11375497-page-number-is-not-displaying-in-the-bibliography

adam3smith:
yeah, but that's not a solution. I'll get in touch with Charlotte. That should be fixed in Mendeley. ```page``` is for a page range, ```number-of-pages``` is for the number of pages of work. Clearly you want the latter here and Mendeley's mapping is incorrect.

diogopereirario:
Agree. 
But i cannot see any other place to include this information either since there is no "number of pages" field. If you know something about it, pls let me know. :)

adam3smith:
Mendeley needs to fix that. I've e-mailed the people there. Your change would break this on other software.

adam3smith:
From Mendeley's developer on that:
> We will change, probably for the next stable one (with the other tiny
change that I need to do there). Thanks!

You'll have to wait (and/or work with a custom version of the style) until then. Thanks for bringing this up!

PR #1406, opened 2015-02-16
Style update for Books

Bibref:
The bibliography needed an update for books because of a punctuation problem. Adding a `<choose>` with two different outcomes will solve this problem

### DIFF
--- a/ets-ecole-de-technologie-superieure.csl
+++ b/ets-ecole-de-technologie-superieure.csl
@@ -163,22 +163,42 @@
           </group>
         </if>
         <else-if type="book" match="any">
-          <group delimiter=". " prefix=", ">
-            <text macro="edition"/>
-            <text macro="collection"/>
-            <group delimiter=", ">
-              <text macro="publisher"/>
-              <choose>
-                <if match="any" variable="number-of-pages">
-                  <text variable="number-of-pages" suffix=" p"/>
-                </if>
-                <else-if match="any" variable="page">
-                  <text variable="page" suffix=" p"/>
-                </else-if>
-              </choose>
-            </group>
-          </group>
-        </else-if>
+					<choose>
+            <if match="any" variable="edition">
+              <group delimiter=". " prefix=", ">
+                <text macro="edition"/>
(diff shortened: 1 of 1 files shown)


rmzelle:
You're missing a `</else-if>` at the end (just before `<else-if type="patent">`).

You can update your pull request by using the edit (pencil) icon at https://github.com/Bibref/styles/blob/patch-4/ets-ecole-de-technologie-superieure.csl

zuphilip:
@rmzelle : There is an update. @Bibref : There are no automatic notifications for commits, thus you should write a short note here as well next time.

ets-ecole-de-technologie-superieure.csl
```diff
@@ -163,21 +163,42 @@
...
+                    <else-if match="any" variable="page">
+                      <text variable="page" suffix=" p"/>
+                    </else-if>
+                  </choose>
+                </group>
+              </group>
+            </if>
+            <else>
+              <group delimiter=". " prefix=". ">
+                <text macro="edition"/>
```
rmzelle: @Bibref, why would you use the "edition" macro here, if the "edition" variable is empty here?

Bibref:
I deleted the macro "edition" as it was not necessary.

rmzelle:
It would be better to use:

```
        <else-if type="book" match="any">
          <text macro="edition" prefix=", "/>
          <group delimiter=". " prefix=". ">
            <text macro="collection"/>
            <group delimiter=", ">
              <text macro="publisher"/>
              <choose>
                <if match="any" variable="number-of-pages">
                  <text variable="number-of-pages" suffix=" p"/>
                </if>
                <else-if match="any" variable="page">
                  <text variable="page" suffix=" p"/>
                </else-if>
              </choose>
            </group>
          </group>
        </else-if>
```

by the way, since then you don't duplicate as much code. (not tested, but I think I got it right)

Bibref:
Very good suggestion. Changes were made!

rmzelle:
Thanks!

PR #1407, opened 2015-02-17
Disable incorrect disambiguate settings

kjloope:
Set disambiguate-add-names and disambiguate-add-givenname to false, as per journal style description here:
http://onlinelibrary.wiley.com/journal/10.1111/(ISSN)1420-9101/homepage/ForAuthors.html

### DIFF
--- a/journal-of-evolutionary-biology.csl
+++ b/journal-of-evolutionary-biology.csl
@@ -139,7 +139,7 @@
       <text variable="publisher-place"/>
     </group>
   </macro>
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks.

Note that while `disambiguate-add-names="false" disambiguate-add-givenname="false"` is not wrong, it's simpler to just delete the attributes (since "false" is their default value). Would you mind making that change? (use the pencil/edit icon at https://github.com/kjloope/styles/blob/patch-1/journal-of-evolutionary-biology.csl to update this pull request)

kjloope:
Thanks, for the advice, just made the change.  This is my first time editing, just trying to make it work!  Please let me know if there’s anything else I can do to fix it.

Cheers,
Kevin

rmzelle:
No, that's it. Thanks!

PR #1412, opened 2015-02-18
Implemented the style for conference papers.

bluechi:


adam3smith:
I don't think that's right. You're thinking of unpublished conference papers, which should be CSL ```speech``` and if you're using Zotero "Presentation" there. That's already correct in the style. ```paper-conference``` is for papers published in conference proceedings. IIRC OSCOLA says to treat those like regular book chapters.

bluechi:
It turns out "Presentation" in Zotero does in fact get me the citation I need, sorry about this!

adam3smith:
great, closing this then. Thanks!

PR #1414, opened 2015-02-19
Fix for patents, conference papers, and thesis.

ivyleavedtoadflax:
Corrected erroneous full stops in patents and conference paper.
Removed thesis variable in thesis (use genre instead).

Squash of three commits in [#1410](https://github.com/citation-style-language/styles/pull/1410).


rmzelle:
Thanks.

Note that if you're comfortable with git, you can squash commits in an existing pull request.

Also, I noticed some extra line-breaks (e.g. at https://github.com/citation-style-language/styles/pull/1414/files#diff-33431ac320636b05dd3e1ea3d9284c63R3). Not a big issue, but see if you can avoid those next time.

ivyleavedtoadflax:
Thanks yes - I'm getting more comfortable...the previous pull request was not on a branch, so it was more difficult to rebase. Will remove them in the next update. Many thanks!

rmzelle:
Yeah, using branches makes things a lot easier. You don't need to change the line breaks by hand; we'll take care of it, since we automatically reindent the styles every once in a while.

PR #1417, opened 2015-02-19
Create zeitschrift-fur-padagogik.csl

zuphilip:
Connected to #1416 and described in https://forums.zotero.org/discussion/39655?page=1#Item_12


rmzelle:
Thanks!

PR #1418, opened 2015-02-21

zuphilip:
Sorry, that was previously still the old version with the point after doi's etc. Moreover, there was a problem with the character encoding. This should fix both issues.


rmzelle:
Thanks again!

PR #1419, opened 2015-02-21
Remove text-case attribute from legislation titles

fbennett:
Responds to [this forum request](https://forums.zotero.org/discussion/46935/capitalization-in-oscola/#Item_0).

### DIFF
--- a/oscola.csl
+++ b/oscola.csl
@@ -187,7 +187,7 @@
         <text variable="title" font-style="italic" text-case="title"/>
       </if>
       <else-if type="bill legislation" match="any">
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks!

oscola.csl
```diff
@@ -202,8 +202,8 @@
       <if type="book legislation motion_picture manuscript" match="any">
         <text variable="title" font-style="italic" text-case="title" form="short"/>
```
zuphilip: @fbennett : The title-short for legislations will then be still title case (and italic). Is that okay?

PR #1420, opened 2015-02-22
OSCOLA: roman + no text-case on title-short legislation

fbennett:
Following up with a fix (thanks to Rintze for flagging the oversight in the previous OSCOLA commit).

Addresses the issue [here](https://forums.zotero.org/discussion/46935/capitalization-in-oscola/#Item_2).

### DIFF
--- a/oscola.csl
+++ b/oscola.csl
@@ -199,10 +199,10 @@
   </macro>
   <macro name="title-short">
     <choose>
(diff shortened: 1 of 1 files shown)


rmzelle:
That was @zuphilip, but I'd gladly share the credit :)

PR #1421, opened 2015-02-22

zuphilip:
Following the discussion https://forums.zotero.org/discussion/46953/zeitschrift-fuer-paedagogik-german/


rmzelle:
Thanks!

PR #1422, opened 2015-02-25
Create ie-Comunicaciones.csl

HendrysTobar:
CSL for the "IE Comunicaciones" Magazine

### DIFF
--- a/ie-comunicaciones.csl
+++ b/ie-comunicaciones.csl
@@ -3,7 +3,7 @@
   <info>
     <title>IE Comunicaciones</title>
     <title-short>IE Com</title-short>
-    <id>http://csl.mendeley.com/styles/10146035/ie-comunicaciones</id>
+    <id>http://www.zotero.org/styles/ie-comunicaciones</id>
     <link href="http://www.zotero.org/styles/ie-comunicaciones" rel="self"/>
     <link href="http://www.zotero.org/styles/harvard-european-archaeology" rel="template"/>
     <link href="http://dialnet.unirioja.es/servlet/revista?info=nor_estilo&amp;codigo=12765&amp;modo=popup" rel="documentation"/>
--- a/ie-comunicaciones.csl
+++ b/ie-comunicaciones.csl
@@ -45,14 +45,27 @@
     </names>
   </macro>
   <macro name="author-short">
-    <names variable="author">
-      <name form="short" and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="2" et-al-use-first="1" initialize-with=". "/>
-      <substitute>
-        <names variable="editor"/>
-        <names variable="translator"/>
+    <choose>
+      <if variable="author">
+        <names variable="author">
+          <name form="short" and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="2" et-al-use-first="1" initialize-with=". "/>
+        </names>
+      </if>
+      <else-if variable="editor">
+        <names variable="editor">
+          <name form="short" and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="2" et-al-use-first="1" initialize-with=". "/>
+        </names>
+      </else-if>
+      <else-if variable="translator">
+        <names variable="translator">
+          <name form="short" and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="2" et-al-use-first="1" initialize-with=". "/>
+        </names>
+       </else-if>
+      <else>
         <text macro="anon"/>
-      </substitute>
-    </names>
+      </else>
+    </choose>
+    
   </macro>
   <macro name="access">
     <choose>
--- a/ie-comunicaciones.csl
+++ b/ie-comunicaciones.csl
@@ -19,7 +19,7 @@
     <updated>2015-02-25T12:29:51+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
-  <locale xml:lang="en">
+  <locale>
     <terms>
       <term name="volume" form="symbol">V</term>
       <term name="issue" form="symbol">N</term>
--- a/ie-comunicaciones.csl
+++ b/ie-comunicaciones.csl
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="2" et-al-use-first="1" et-al-subsequent-min="2" et-al-subsequent-use-first="1" demote-non-dropping-particle="sort-only" default-locale="en-US">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never" demote-non-dropping-particle="sort-only" >
   <info>
     <title>IE Comunicaciones</title>
     <title-short>IE Com</title-short>
@@ -13,6 +13,7 @@
       <uri>http://www.mendeley.com/profiles/hendrys-tobar/</uri>
     </author>
     <category citation-format="author-date"/>
+    <category field="social_science"/>
     <issn>1699-4574</issn>
     <summary>Revista IE Comunicaciones. </summary>
     <updated>2015-02-25T12:29:51+00:00</updated>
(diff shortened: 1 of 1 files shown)


HendrysTobar:
I leave this comment to notify you I made some changes on the ID.

ie-comunicaciones.csl
```diff
@@ -0,0 +1,205 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="2" et-al-use-first="1" et-al-subsequent-min="2" et-al-subsequent-use-first="1" demote-non-dropping-particle="sort-only" default-locale="en-US">
```
adam3smith: given that you can publish in Spanish and Portuguese in the journal, we'll want the default-locale removed.
Also, please remove the et-al settings here, better to set them only once

ie-comunicaciones.csl
```diff
@@ -0,0 +1,205 @@
...
+    <id>http://www.zotero.org/styles/ie-comunicaciones</id>
+    <link href="http://www.zotero.org/styles/ie-comunicaciones" rel="self"/>
+    <link href="http://www.zotero.org/styles/harvard-european-archaeology" rel="template"/>
+    <link href="http://dialnet.unirioja.es/servlet/revista?info=nor_estilo&amp;codigo=12765&amp;modo=popup" rel="documentation"/>
+    <author>
+      <name>Hendrys Tobar</name>
+      <email>u1918137@campus.udg.edu</email>
+      <uri>http://www.mendeley.com/profiles/hendrys-tobar/</uri>
+    </author>
+    <category citation-format="author-date"/>
```
adam3smith: we'd want a field here, probably social_science? Has to be from this list: http://citationstyles.org/downloads/specification.html#appendix-i-categories
```<category field="social_science"/>

File: ie-comunicaciones.csl [hunk 3]
```diff
@@ -0,0 +1,205 @@
...
+    <updated>2015-02-25T12:29:51+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="en">
+    <terms>
+      <term name="volume" form="symbol">V</term>
+      <term name="issue" form="symbol">N</term>
+    </terms>
+  </locale>
+  <citation et-al-min="2" et-al-use-first="1" et-al-subsequent-min="2" et-al-subsequent-use-first="1" disambiguate-add-names="true" disambiguate-add-givenname="true" disambiguate-add-year-suffix="true">
```
adam3smith: while valid this way, we always place the citation section right above the bibliography section. please move it down.

ie-comunicaciones.csl
```diff
@@ -0,0 +1,205 @@
...
+      <email>u1918137@campus.udg.edu</email>
+      <uri>http://www.mendeley.com/profiles/hendrys-tobar/</uri>
+    </author>
+    <category citation-format="author-date"/>
+    <issn>1699-4574</issn>
+    <summary>Revista IE Comunicaciones. </summary>
+    <updated>2015-02-25T12:29:51+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="en">
```
adam3smith: that's not what I'm seeing in the journal? But in any case, once you remove the default locale at the top, you'll want to adjust this here, too.

ie-comunicaciones.csl
```diff
@@ -0,0 +1,205 @@
...
+      <name and="text" initialize-with=". "/>
+      <label form="short" prefix=" (" suffix=")" strip-periods="true"/>
+    </names>
+  </macro>
+  <macro name="anon">
+    <text term="anonymous" form="short" text-case="capitalize-first" strip-periods="true"/>
+  </macro>
+  <macro name="author">
+    <names variable="author">
+      <name and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="10" et-al-use-first="10" et-al-subsequent-min="10" et-al-subsequent-use-first="10" initialize-with="."/>
```
adam3smith: you always want the et-al-min to be at least one higher than the et-al-use-first: probably 11/10 here.

ie-comunicaciones.csl
adam3smith: there's no need to specify the est-al-subsequent conditions if they're the same as the regular et-al. Please remove them throughout the style.

ie-comunicaciones.csl
```diff
@@ -0,0 +1,205 @@
...
+                    <text term="volume" form="symbol" text-case="capitalize-first" suffix=" "/>
+                    <text variable="volume"/>
+                  </group>
+                  <group>
+                    <text term="issue" form="symbol" text-case="capitalize-first" prefix=" "/>
+                    <text variable="issue" prefix=" "/>
+                  </group>
+                </group>
+                <group prefix=" " suffix=".">
+                  <label plural="always" variable="locator" form="short"/>
```
adam3smith: you sure about that plural="always"? I'm seeing p. and pp. in the journal.

ie-comunicaciones.csl
```diff
@@ -0,0 +1,205 @@
...
+        <names variable="editor"/>
+        <text macro="anon"/>
+      </substitute>
+    </names>
+  </macro>
+  <macro name="author-short">
+    <names variable="author">
+      <name form="short" and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="2" et-al-use-first="1" et-al-subsequent-min="2" et-al-subsequent-use-first="1" initialize-with=". "/>
+      <substitute>
+        <names variable="editor"/>
```
adam3smith: you don't want to use substitute here if you're going to use this in the bibliography. Substitute will omit that variable from the rest of the reference, but you still want it to appear in long form, correct? So you'll have to run a series of if/else-ifs here

```
<choose>
<if variable="author">
</if>
<else-if variable="editor">
</else-if>
....
</choose>
```

adam3smith:
Thanks, see my comments, let us know if you have any questions.

ie-comunicaciones.csl
HendrysTobar: I've removed the locale. But:
In Mendeley's last version you can choose the language to use.
I've been invited to publish here an article in english. So I guess they accept english too.

ie-comunicaciones.csl
HendrysTobar: Done

ie-comunicaciones.csl
HendrysTobar: But I'm using the style in english. How do yoy define a term to be used by any language?

ie-comunicaciones.csl
HendrysTobar: Done

ie-comunicaciones.csl
HendrysTobar: Done

ie-comunicaciones.csl
HendrysTobar: Actually, I'm just guessing with the always plural. Because frankly their directions for authors are quite incomplete and in the example they use plural. 

However I'm changing it to contextual.

ie-comunicaciones.csl
HendrysTobar: I don't understand what yu're saying. The substitute element as I understand it (it was already in the template used) is there to replace the name is none is given. It replaces with the editor or the translator whichever is present.

I'm using it in the bibliography because it contains the same as in the inline citation.

Please clarify what youy are trying to tell me. thanks.

HendrysTobar:
I have updated the file. Please review some notes in the previous threas as I did not understand some suggestions.

ie-comunicaciones.csl
rmzelle: > How do yoy define a term to be used by any language?

For that, just remove the `xml:lang="en"`.

ie-comunicaciones.csl
adam3smith: the problem is that you're using both author-short and author in the bibliography. ```substitute``` actually does two things:
1. It replaces the original variable that's not present (as you say) and
2. It suppresses the swapped-in variable in the rest of the reference.  
See [here](http://citationstyles.org/downloads/specification.html#substitute) for the relevant portion of the specifications.

This normally makes a lot of sense, e.g. if you're using the title for an author-less work, you don't want it to appear twice. However, in your case, since you _do_ want editors of works with no author to appear twice (once in the label between [], once in the long form, it would break things. That's why you'll need to work with regular choose and if here, which doesn't have the 2nd effect.

ie-comunicaciones.csl
adam3smith: specify the locale without any xml:lang

HendrysTobar: Ok, done. Thanks

ie-comunicaciones.csl
HendrysTobar: I think I've done it. But I'm not sure, thank you for reviewing.

HendrysTobar:
I have updated the file with the <choose>

adam3smith:
Looks good, thanks!

HendrysTobar:
Thanks to you.

2015-02-25 18:18 GMT+01:00 Sebastian Karcher <notifications@github.com>:

> Looks good, thanks!
>
> —
> <https://github.com/citation-style-language/styles/pull/1422#issuecomment-76007750>
> .
>



-- 
Hendrys Fabián

HendrysTobar:
So, now I can share the style with the url http://www.zotero.org/styles/ie-comunicaciones ?

adam3smith:
will take about 30mins for that to become available.

adam3smith:
The style will also show up in Mendeley's style search within the next couple of days.

PR #1424, opened 2015-02-27

roaldfrosig:
Just correcting an embarrassing oversight:
The 'original-date' field should be displayed in the citation of chapters if it is provided. But since that field most often is empty, I need to substitute with the ordinary 'date' field.


rmzelle:
Thanks!

PR #1425, opened 2015-03-01
Create FASEB Journal style.

bovender:


adam3smith:
any particular reason you're closing this? If you need to make changes, we prefer to make them in the same pull request (simply by updating the style file in your branch of github). 
Let us know if you need help.

bovender:
Yes, in fact I had to fix a few things. Travis-CI testing failed due to a duplicate ISSN -- removed it. In addition, I fixed two minor typography issues. I'm posting the corrected version to this pull request now. Thanks for your patience.

rmzelle:
@bovender, thanks for submitting a style! (also, since I work in the life sciences, http://xltoolbox.sourceforge.net/ looks very cool; I will check it out!)

There are a few minor changes I'd like to request:
* Please add a "template" link so we know which CSL style you based this one on. (point 8 of https://github.com/citation-style-language/styles/wiki/Style-Requirements)
* Please add a "documentation" link that points to a webpage describing the citation format (probably http://www.fasebj.org/site/misc/rc.xhtml ?) (point 10)
* Please add "biology" as a category field. (`<category field="biology"/>`)
* Please rename the style to "The FASEB Journal" and "the-faseb-journal.csl", and update the style ID and "self" link accordingly, since that appears to be the journal's full title.

faseb-journal.csl
```diff
@@ -0,0 +1,145 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" initialize-with=", " names-delimiter=", " page-range-format="expanded" default-locale="en-US" xmlns="http://purl.org/net/xbiblio/csl">
+  <info>
+    <title>FASEB Journal</title>
+    <title-short>FASEB</title-short>
```
zuphilip: @rmzelle Should the subtitle "The Journal of the Federation of American Societies of Experimental Biology" may be added as `title-short` for retrieval?

adam3smith: let's not put that in title-short, no (because it's not short ;). It'd be nice in ```summary``` though, even though that's not searched on the repository

rmzelle: @zuphilip, no. This case is an argument for introduction `title-long`, but "The Journal of the Federation of American Societies of Experimental Biology" is definitely not the short version of the title.

zuphilip: Okay, sure.

bovender: I'd rather not use the very long name of the journal as "title", because "title" is what will appear e.g. on zotero.org/styles, but everybody would be searching for "FASEB", which is what this journal is commonly called, and nobody would find it if the full title was spelled out.

rmzelle: Right. I'm just asking you to use "The FASEB Journal" instead of "FASEB Journal".

adam3smith: the main title should clearly be The FASEB Journal (but with the "The"), I don't think there's any disagreement. @rmzelle was wondering whether we should have an additional field to store long titles of journals (e.g. containing association information or subtitles), but that's a discussion for another place.
Right now, if you'd add a summary field that contains the long title that's be cool. Also see Rintze's other requests.

the-faseb-journal.csl
```diff
@@ -0,0 +1,149 @@
...
+  </macro>
+  <citation collapse="citation-number">
+    <sort>
+      <key variable="citation-number"/>
+    </sort>
+    <layout delimiter=", " prefix=" (" suffix=")">
+      <text variable="citation-number" vertical-align="baseline" />
+    </layout>
+  </citation>
+  <bibliography hanging-indent="true" et-al-min="0" second-field-align="flush">
```
adam3smith: sorry I didn't see that earlier, but what do you mean to do with ```et-al-min=0``` ? It doesn't make a lot of sense CSL-wise. If you don't want any et al, ever, just remove that attribute everywhere in the style or set it to an arbitrary high number like 99 or 999

the-faseb-journal.csl
```diff
@@ -0,0 +1,149 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" initialize-with=", " names-delimiter=", " page-range-format="expanded" default-locale="en-US" xmlns="http://purl.org/net/xbiblio/csl">
```
adam3smith: could you remove the intialize-with and the names-delimiter here? I don't think you want that and it will make it harder to work with the style for others.

the-faseb-journal.csl
```diff
@@ -0,0 +1,149 @@
...
+    <category field="biology"/>
+    <issn>0892-6638</issn>
+    <eissn>1530-6860</eissn>
+    <updated>2015-03-01T11:08:41+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <macro name="editor">
+    <names variable="editor">
+      <name name-as-sort-order="all" sort-separator=" " initialize-with="" delimiter=", " delimiter-precedes-last="always"/>
+      <label form="short" prefix=", "/>
```
adam3smith: looking at the styleguide, this should have strip-periods="true" (so you get ed and eds rather than ed. and eds.)

the-faseb-journal.csl
```diff
@@ -0,0 +1,149 @@
...
+    <category field="medicine"/>
+    <category field="biology"/>
+    <issn>0892-6638</issn>
+    <eissn>1530-6860</eissn>
+    <updated>2015-03-01T11:08:41+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <macro name="editor">
+    <names variable="editor">
+      <name name-as-sort-order="all" sort-separator=" " initialize-with="" delimiter=", " delimiter-precedes-last="always"/>
```
adam3smith: according to the styleguide, this should be ```initialize-with=". "``` for editors, too.

the-faseb-journal.csl
```diff
@@ -0,0 +1,149 @@
...
+      <label form="short" prefix=", "/>
+    </names>
+  </macro>
+  <macro name="author">
+    <group suffix=".">
+      <names variable="author">
+        <name and="text" delimiter-precedes-last="always" et-al-min="0" et-al-use-last="true" initialize-with=". " name-as-sort-order="all"/>
```
adam3smith: I assume you want to remove the et-al-use-last="true" here? Or what is it supposed to do?

the-faseb-journal.csl
```diff
@@ -0,0 +1,149 @@
...
+        <label form="short" prefix=", "/>
```
adam3smith: probably add the strip-periods="true" here, too, for consistency.

the-faseb-journal.csl
```diff
@@ -0,0 +1,149 @@
...
+          <text macro="title"/>
+        </substitute>
+      </names>
+    </group>
+  </macro>
+  <macro name="title">
+    <text variable="title" suffix="."/>
+  </macro>
+  <macro name="publisher">
+    <group delimiter=": ">
```
adam3smith: styleguide would suggest a comma here?

the-faseb-journal.csl
```diff
@@ -0,0 +1,149 @@
...
+        <label form="short" prefix=", "/>
+        <substitute>
+          <names variable="editor"/>
+          <text macro="title"/>
+        </substitute>
+      </names>
+    </group>
+  </macro>
+  <macro name="title">
+    <text variable="title" suffix="."/>
```
adam3smith: you're already setting that period suffix below in the bibliography. Remove it here.

the-faseb-journal.csl
```diff
@@ -0,0 +1,149 @@
...
+    <layout delimiter=", " prefix=" (" suffix=")">
+      <text variable="citation-number" vertical-align="baseline" />
+    </layout>
+  </citation>
+  <bibliography hanging-indent="true" et-al-min="0" second-field-align="flush">
+    <layout>
+      <text variable="citation-number" suffix=". "/>
+      <text macro="author" suffix=". "/>
+      <date date-parts="year" form="text" variable="issued" prefix="(" suffix=")"/>
+      <text macro="title" prefix=" " suffix=". "/>
```
adam3smith: remove the period after the suffix here: you have spaces in all the prefixes -- best to not mix those and only put spaces in affixes in on one side to prevent problems when fields are missing.

the-faseb-journal.csl
```diff
@@ -0,0 +1,149 @@
...
+          <text prefix=" " macro="publisher"/>
+          <group suffix="." prefix=", ">
+            <text variable="page" prefix=":"/>
+          </group>
+        </if>
+        <else-if type="chapter paper-conference" match="any">
+          <group prefix=" ">
+            <text term="in" text-case="capitalize-first" suffix=": "/>
+            <text macro="editor"/>
+            <text variable="container-title" prefix=" " suffix="."/>
```
adam3smith: looks to me like this should be in italics? font-style="italics"

the-faseb-journal.csl
```diff
@@ -0,0 +1,149 @@
...
+          </group>
+          <text prefix=" " macro="publisher"/>
+          <group suffix="." prefix=", ">
+            <text variable="page" prefix=":"/>
+          </group>
+        </if>
+        <else-if type="chapter paper-conference" match="any">
+          <group prefix=" ">
+            <text term="in" text-case="capitalize-first" suffix=": "/>
+            <text macro="editor"/>
```
adam3smith: shouldn't this be in parentheses, too, just like for books?

the-faseb-journal.csl
```diff
@@ -0,0 +1,149 @@
...
+            <text variable="container-title" prefix=" " suffix="."/>
+            <text variable="volume" prefix="Vol " suffix="."/>
+            <text macro="edition" prefix=" "/>
+            <text variable="collection-title" prefix=" " suffix="."/>
+            <group suffix=".">
+              <text macro="publisher" prefix=" "/>
+              <group suffix="." prefix=", ">
+                <text variable="page" prefix=":"/>
```
adam3smith: looks like you should have ```<label variable="page"/>``` here. also, the colon in the prefix seems wrong for chapters.

adam3smith:
Sorry, I hadn't gotten a look at the actual contents of this before. See comments in line -- if there are reason you've done something I'm asking to be changed, please briefly explain and feel free to disregard my request. 
Let us know when you're done, we don't get notified on updates.
Thanks!

bovender:
@adam3smith Thanks for taking the time and scrutinizing the style. I think some (if not all) of the inconsistency comes from my inexperience with CSL, sorry. I fiddled with it until the citations looked the way I thought they should look... Will fix these issues shortly.

bovender:
Considering that the style guide at http://www.fasebj.org/site/misc/rc.xhtml defines bibliography styles only for journal articles and books, I am inclined to remove the definitions for all other item types from the style. Thus, there would be an `<if type="book">`, an `<if type="chapter">`, and an `<else>` element. I've looked at a couple of FASEB Journal's articles, and I don't know if any of the other possible item types are really ever cited.

bovender:
@adam3smith: I've cleaned up the style definition quite a bit. Let me know if more changes are needed.

the-faseb-journal.csl
```diff
@@ -0,0 +1,126 @@
...
+    <layout>
+      <text variable="citation-number" suffix=". "/>
+      <text macro="author" suffix=". "/>
+      <date variable="issued" date-parts="year" form="text" prefix="(" suffix=") "/>
+      <text macro="title" suffix=". "/>
+      <text macro="container" />
+    </layout>
+  </bibliography>
+</style>
+<!-- vim: set sw=2 sts=-1 et : -->
```
adam3smith: please delete the vim comment

the-faseb-journal.csl
```diff
@@ -0,0 +1,126 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0"
+    default-locale="en-US"
+    class="in-text" 
+    page-range-format="expanded"
+    demote-non-dropping-particle="sort-only"
+    initialize-with=". "
```
adam3smith: please delete lines 7 and 8, then put the style element into a single line. Don't separate the attributes into new lines.

the-faseb-journal.csl
```diff
@@ -0,0 +1,126 @@
...
+  <citation collapse="citation-number">
+    <sort>
+      <key variable="citation-number"/>
+    </sort>
+    <layout delimiter=", " prefix=" (" suffix=")">
+      <text variable="citation-number" vertical-align="baseline"/>
+    </layout>
+  </citation>
+  <bibliography
+      second-field-align="flush"
```
adam3smith: same here -- don't split the tag over multiple lines.

adam3smith:
thanks. 3 minor comments, everything else looks great.
Sebastian

the-faseb-journal.csl
bovender: @adam3smith I removed lines 7 and 8 (`initialize-with` and `name-as-sort-order` attributes), but this causes the names to be spelled out as 'first name [initial] last name'. The journal wants 'last name comma initials'. Is there a better way to achieve this than putting back lines 7 and 8?

adam3smith: yes, include them in lines 32 and 38 (of this version of the style) respectively, i.e. as part of the cs:name. 
There is nothing in the specifications that makes what you did wrong, but (almost?) all other styles have these attributes set on the name element and keeping this uniform will make it much easier to maintain styles.

bovender: @adam3smith ok, got it.

adam3smith:
great, thanks!

PR #1426, opened 2015-03-01
Update freie-universitat-berlin-geographische-wissenschaften.csl

Sidapo:
now follows the documentation more closely + some code clean up

### DIFF
--- a/freie-universitat-berlin-geographische-wissenschaften.csl
+++ b/freie-universitat-berlin-geographische-wissenschaften.csl
@@ -43,8 +43,10 @@
     </choose>
   </macro>
   <macro name="pagenumbers">
-    <number variable="number-of-pages" suffix=" "/>
-    <label variable="number-of-pages" form="short"/>
+    <group delimiter=" ">
+      <number variable="number-of-pages"/>
+      <label variable="number-of-pages" form="short"/>
+     </group>
   </macro>
   <macro name="thesis">
     <text variable="publisher" prefix="Dissertation, " suffix=". " />
@@ -114,13 +116,17 @@
   </choose>
   </macro>
   <macro name="page">
-    <label variable="page" form="short" suffix=" "/>
-    <text variable="page"/>
+    <group delimiter=" ">
+      <label variable="page" form="short"/>
+      <text variable="page"/>
+    </group>
   </macro>
   <macro name="journal">
     <group delimiter=". " suffix=".">
-      <text variable="container-title" suffix=" "/>
-      <number variable="volume" font-weight="bold"/>
+      <group delimiter=" "> 
+        <text variable="container-title" suffix=" "/>
+        <number variable="volume" font-weight="bold"/>
+      </group>
       <choose>
         <if variable="issue">
           <group prefix=" (" delimiter=" " suffix=")">
--- a/freie-universitat-berlin-geographische-wissenschaften.csl
+++ b/freie-universitat-berlin-geographische-wissenschaften.csl
@@ -46,11 +46,13 @@
     <group delimiter=" ">
       <number variable="number-of-pages"/>
       <label variable="number-of-pages" form="short"/>
-     </group>
+    </group>
   </macro>
   <macro name="thesis">
-    <text variable="publisher" prefix="Dissertation, " suffix=". " />
-    <text macro="pagenumbers" suffix="."/>
+    <group prefix="Dissertation, " delimiter=". " suffix=".">
+-     <text variable="publisher"/>
+-     <text macro="pagenumbers"/>
+-   </group>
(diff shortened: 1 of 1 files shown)


freie-universitat-berlin-geographische-wissenschaften.csl
```diff
@@ -17,35 +17,38 @@
...
     </choose>
-    <text variable="publisher-place"/>
   </macro>
   <macro name="pagenumbers">
-    <group delimiter=" ">
-      <number variable="number-of-pages"/>
-      <label variable="number-of-pages" form="short"/>
-    </group>
+    <number variable="number-of-pages" suffix=" "/>
+    <label variable="number-of-pages" form="short"/>
```
zuphilip: That doesn't look like an improvement. Why did you delete the group? Now it will display the label even the data for the number of pages is not present. There are more such cases following...

Sidapo:
oops, I didn't realize that would happen. I was under the impression my style had a lot of unnecessary groups... I'll undo the group changes then

freie-universitat-berlin-geographische-wissenschaften.csl
```diff
@@ -17,35 +17,40 @@
...
-    </group>
+     </group>
   </macro>
   <macro name="thesis">
-    <group prefix="Dissertation, " delimiter=". " suffix=".">
-      <text variable="publisher"/>
-      <text macro="pagenumbers"/>
-    </group>
+    <text variable="publisher" prefix="Dissertation, " suffix=". " />
+    <text macro="pagenumbers" suffix="."/>
```
zuphilip: Another destroyed group

freie-universitat-berlin-geographische-wissenschaften.csl
```diff
@@ -113,8 +123,8 @@
   </macro>
   <macro name="journal">
     <group delimiter=". " suffix=".">
-      <group delimiter=" ">
-        <text variable="container-title"/>
+      <group delimiter=" "> 
+        <text variable="container-title" suffix=" "/>
         <number variable="volume" font-weight="bold"/>
       </group>
```
zuphilip: The suffix is not needed here, because you already have it as a delimiter.

freie-universitat-berlin-geographische-wissenschaften.csl
Sidapo: ..but isn't this group actually unnecessary? Since the page numbers should still be displayed even if the publisher is not written.

freie-universitat-berlin-geographische-wissenschaften.csl
```diff
@@ -180,10 +188,8 @@
       <key macro="author"/>
     </sort>
     <layout prefix="(" suffix=")" delimiter="; ">
-      <group delimiter=" ">
-        <text macro="author-short"/>
-        <text macro="year-date"/>
-      </group>
+      <text macro="author-short" suffix=" "/>
+      <text macro="year-date"/>
```
zuphilip: Another destroyed group

freie-universitat-berlin-geographische-wissenschaften.csl
zuphilip: Right, but the suffixes are different. There are tree cases possible i) both variables are there, ii) only publisher is there, iii) only pagenumbers is there. The delimiter in the group will only be shown if both occur.

freie-universitat-berlin-geographische-wissenschaften.csl
Sidapo: here I thought the same as above. Since "author-short" always returns something, the group is superfluous..

freie-universitat-berlin-geographische-wissenschaften.csl
zuphilip: In iii) the pagenumbers will still be shown, but the prefix "Dissertation, " is missing.

zuphilip:
Here is a really good blog entry explaining groups in CSL: https://zoteromusings.wordpress.com/2013/10/28/writing-csl-features-and-best-practices/

freie-universitat-berlin-geographische-wissenschaften.csl
zuphilip: Assume "year-date" is missing, then we will still have the space from the suffix in your current code.

Sidapo:
okay great, thanks for your help and patience! I'll change the groups back then.. Classic example of "don't fix whats not broken" I guess

zuphilip:
Okay, you fixed it, thank you. Let us wait for one of the maintainer to review here.

adam3smith:
looks all good to me, thank you both!

PR #1428, opened 2015-03-02
Use title case for titles in MHRA author-date.

adunning:
I believe this was changed in the third edition of the guide. This style still needs a bit of work to bring it in line with recent modifications to the note version of MHRA.

Another thing I've noticed about the MHRA styles: because its method of indicating a DOI is to print the URL with the resolver, we're putting `http://dx.doi.org/` as a prefix to the DOI, but this causes problems for versions of citeproc that try to automatically link it, because it cannot not recognize the prefix as being part of the URL. Is there a better way of doing this?

### DIFF
--- a/modern-humanities-research-association-author-date.csl
+++ b/modern-humanities-research-association-author-date.csl
@@ -20,7 +20,7 @@
     <category citation-format="author-date"/>
     <category field="generic-base"/>
     <summary>MHRA author date format. This is rather poorly specified in the style guide, so it includes some (hopefully reasonable) assumptions: et al for 3+ author; "and" between two authors; no added names, but added given names for disambiguation.</summary>
-    <updated>2014-03-12T15:55:55+00:00</updated>
+    <updated>2015-03-02T23:37:43+00:00</updated>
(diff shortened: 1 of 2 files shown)


adam3smith:
Interesting Q about DOIs. @rmzelle @aurimasv @fbennett -- this would seem like a more general issue. How are we thinking about this? 
I assume the only processor that's auto-linking DOIs is pandoc-citeproc, correct? It's a nice option, though I'd very much like to keep it optional. 

(I'll review the proposed change to casing tonight).

adunning:
Linking is also theoretically possible with citeproc-js, and is proposed for Zotero at zotero/zotero#505 (though this implementation would hide the DOI/URL, skirting the problem entirely).

adam3smith:
thanks!

rmzelle:
@adam3smith, not sure. It doesn't seem to hard to include the (appropriate substring of the) prefix when defining the active URL, though.

adam3smith:
Yes, Rintze is right - as using the full link becomes more common in styles (APA recommends it, too), citeproc-pandoc should just recognize this.

adunning:
Thanks for the input; I will file a report with them.

PR #1429, opened 2015-03-02
Do not substitute title for journal articles in Chicago author-date.

adunning:
This is from citation-style-language/styles#1028, which implemented CMoS 14.207 and 14.217, referring specifically to 'Unsigned newspaper articles' and 'Unsigned reviews'. With journal articles, it gives some strange results, such as:

> *Sacris Erudiri*. 1948. ‘A Proposed New Edition of Early Christian Texts’ 1: 405–14. doi:10.1484/J.SE.2.305052.


adam3smith:
makes sense, thanks.

PR #1430, opened 2015-03-03
Fixes to Annual Reviews author-date

bwiernik:
1. Fix display of paper-conference (previously it had not displayed any container information)
2. Add support for status and original date variables

### DIFF
--- a/annual-reviews-author-date.csl
+++ b/annual-reviews-author-date.csl
@@ -84,6 +84,16 @@
       <date-part name="year"/>
     </date>
   </macro>
+  <macro name="date-original">
+    <group delimiter=" ">
+      <date variable="issued">
+        <date-part name="year"/>
+      </date>
+      <date variable="original-date" prefix="(" suffix=")">
+        <date-part name="year"/>
+      </date>
+    </group>
+  </macro>
   <macro name="date-m-d">
     <date variable="issued" delimiter=" ">
       <date-part name="month" form="short"/>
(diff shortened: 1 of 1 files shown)


annual-reviews-author-date.csl
```diff
@@ -224,16 +260,43 @@
...
+            <text macro="locator"/>            
+          </if>
+          <else>
+            <text macro="container-title"/>
+            <text macro="locator"/>
+            <choose>
+              <if variable="publisher" match="any">
+                <group delimiter=". " prefix=". ">
+                  <text macro="publisher"/>
+                  <text macro="status"/>
```
adam3smith: this looks like we could get status twice -- once from the locator macro, one here. Am I missing something?

adam3smith:
so, what do you see as being in the "status" variable? And does it really never replace the date, e.g. for "in press" when no date of publication is assigned?

annual-reviews-author-date.csl
bwiernik: There is a conditional to test for 'volume' in both the locator macro and the container macro that controls the placement of status. The point is to accommodate the different delimiters needed for journals/journal-like proceedings and books/book-like proceedings.

bwiernik:
For Annual Reviews, "status" should always be "In press". It is placed at the end of the reference, after as much of the rest of the reference information as possible. It never replaces the date (some date, usually the expected publication date, should always be included). It only appears in the bibliography, not in the citations. 

Pages 19 and 29 here:
http://www.annualreviews.org/userimages/ContentEditor/1343842689197/AuthorHandbook-GREEN.pdf

adam3smith:
OK, sounds good

PR #1431, opened 2015-03-04
Create Style Manual - Australian Government (author-date)

s002610:
Create style to comply with Snooks & co. Harvard style guide

### DIFF
--- a/SnooksHarvard.csl
+++ b/SnooksHarvard.csl
@@ -2,8 +2,8 @@
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
   <info>
     <title> Snooks Australia Harvard (author-date)</title>
-    <id>http://www.zotero.org/styles/snooksharvard</id>
-    <link href="http://www.zotero.org/styles/snooksharvard" rel="self"/>
+    <id>http://www.zotero.org/styles/SnooksHarvard</id>
+    <link href="/http[s]?:\/\/www.zotero.org\/styles\/SnooksHarvard/" rel="self"/>
     <link href="http://www.australia.gov.au/publications/style-manual" rel="documentation"/>
     <link href="http://www.zotero.org/styles/harvard1" rel="template"/>
     <author>
--- a/SnooksHarvard.csl
+++ b/SnooksHarvard.csl
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
   <info>
     <title> Snooks Australia Harvard (author-date)</title>
-    <id>http://www.zotero.org/styles/SnooksHarvard</id>
-    <link href="/http[s]?:\/\/www.zotero.org\/styles\/SnooksHarvard/" rel="self"/>
+    <id>http://www.zotero.org/styles/snooksharvard</id>
+    <link href="http://www.zotero.org/styles/snooksharvard" rel="self"/>
     <link href="http://www.australia.gov.au/publications/style-manual" rel="documentation"/>
     <link href="http://www.zotero.org/styles/harvard1" rel="template"/>
     <author>
@@ -24,19 +24,16 @@
  <macro name="editor"> 
     <names variable="editor" delimiter=" " prefix="">
       <name and="symbol" initialize-with="" delimiter=", "/>
-      <label form="short" prefix=" (" suffix=")" text-case="lowercase"/>
+      <label form="short" prefix=" (" suffix=")"/>
     </names>
-  </macro>
-  <macro name="anon">
-    <text macro="title"/>
   </macro>
   <macro name="author">
     <names variable="author">
       <name name-as-sort-order="all" and="symbol" sort-separator=", "  initialize-with="" delimiter-precedes-last="never" delimiter=", "/>
-      <label form="short" prefix=" (" suffix=")" text-case="lowercase"/>
+      <label form="short" prefix=" (" suffix=")"/>
       <substitute>
         <names variable="editor"/>
-        <text macro="anon"/>
+        <text macro="title"/>
       </substitute>
     </names>
   </macro>
@@ -46,15 +43,15 @@
       <substitute>
         <names variable="editor"/>
         <names variable="translator"/>
-        <text macro="anon"/>
+        <text macro="title"/>
       </substitute>
     </names>
   </macro>
   <macro name="access">
     <choose>
       <if variable="URL">
-	  <group suffix=" ">
-          <text value="viewed" text-case="lowercase" suffix=" "/>
+	  <group>
+          <text value="viewed" suffix=" "/>
           <date variable="accessed">
             <date-part name="month" suffix=" "/>
             <date-part name="day" suffix=" "/>
@@ -71,14 +68,14 @@
         <text variable="title" font-style="italic"/>
       </if>
       <else>
-        <text variable="title" text-case="title" prefix="'" suffix="'"/>
+        <text variable="title" text-case="title" quotes="true"/>
       </else>
     </choose>
   </macro>
   <macro name="publisher">
-    <group delimiter=", ">
-      <text variable="publisher-place" prefix=", "/>
+    <group delimiter=", " prefix=", ">
       <text variable="publisher"/>
+      <text variable="publisher-place"/>
     </group>
   </macro>
   <macro name="year-date">
@@ -97,7 +94,7 @@
     <choose>
       <if is-numeric="edition">
         <group delimiter=" ">
-          <number variable="edition" form="ordinal" prefix=", " />
+          <number variable="edition" form="ordinal" />
           <text term="edition" form="short"/>
         </group>
       </if>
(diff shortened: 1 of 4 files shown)


rmzelle:
Is Snooks & co. still involved in this edition? I don't see any mention of them at http://www.wiley.com/WileyCDA/WileyTitle/productCd-0701636483.html or http://www.australia.gov.au/publications/style-manual, and they don't seem to have their own website.

adam3smith:
It says "Revised by Snooks & Co at the bottom right of the title page. You can spot it if you zoom on Wiley. I'm not sure about naming--what are you thinking @rmzelle

rmzelle:
Depends on whether the style guide only defines an author-date format, or multiple formats. And on whether we should name it after Snooks & Co. or something else. Just "Style Manual" is too generic.

Maybe "Style Manual (Snooks & Co.)" and "style-manual-snooks-and-co.csl"?

adam3smith:
maybe Australian should be in the name? Doesn't appear to be part of the official name, but it's commissioned by the Australian gov't printing office.

SnooksHarvard.csl
```diff
@@ -0,0 +1,188 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
```
adam3smith: we'll want this with a default locale, I think. en-GB would make most sense given the similarities (and the fact that we don't have--nor seem to need--an Australia-specific locale.
So add default-locale="en-GB" to this line.

SnooksHarvard.csl
```diff
@@ -0,0 +1,188 @@
...
+  </info>
+  <locale xml:lang="en">
+         <terms>
+	<term name="edition" form="short">edn</term>
+</terms>  
+</locale>
+ <macro name="editor"> 
+    <names variable="editor" delimiter=" " prefix="">
+      <name and="symbol" initialize-with="" delimiter=", "/>
+      <label form="short" prefix=" (" suffix=")" text-case="lowercase"/>
```
adam3smith: all the labels and terms are already lowercase. Please remove all instances of text-case="lowercase" throughout the style. They make working with it down the road much harder.

SnooksHarvard.csl
```diff
@@ -0,0 +1,188 @@
...
+    </names>
+  </macro>
+  <macro name="anon">
```
adam3smith: this is confusing. If items with no author/editor should just print the title, put that into the author macros directly. (i.e. use ```<text macro="title"/>``` there instead of the anon macro. Then delete this one.

SnooksHarvard.csl
```diff
@@ -0,0 +1,188 @@
...
+      </if>
+    </choose>
+  </macro>
+  <macro name="title">
+    <choose>
+      <if type="bill book graphic legal_case legislation motion_picture report song thesis webpage" match="any">
+        <text variable="title" font-style="italic"/>
+      </if>
+      <else>
+        <text variable="title" text-case="title" prefix="'" suffix="'"/>
```
adam3smith: we much prefer using ```quotes="true"``` instead of hardcoding single quotes.

SnooksHarvard.csl
```diff
@@ -0,0 +1,188 @@
...
+      </else>
+    </choose>
+  </macro>
+  <macro name="publisher">
+    <group delimiter=", ">
+      <text variable="publisher-place" prefix=", "/>
```
adam3smith: you'll want to move that ```prefix=", "``` up to the group.

SnooksHarvard.csl
```diff
@@ -0,0 +1,188 @@
...
+        <group>
+          <label variable="locator" form="short"/>
+          <text variable="locator"/>
+        </group>
+      </group>
+    </layout>
+  </citation>
+  <bibliography hanging-indent="false" et-al-min="4" et-al-use-first="1">
+    <sort>
+      <key macro="author"/>
```
adam3smith: are you positive about the sort order? author-date style much more commonly sort by date as the second criterion.

SnooksHarvard.csl
```diff
@@ -0,0 +1,188 @@
...
+      <key variable="title"/>
+    </sort>
+    <layout suffix=".">
+      <text macro="author"/>
+      <text macro="year-date" prefix=" "/>
+      <choose>
+        <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
+          <group prefix="" delimiter="">
+            <text macro="title" prefix=", "/>
```
adam3smith: two things:
1) try to avoid having spaces in both prefixes and suffixes. Ideally, put most of them in delimiters, but where that's not an option, pick either prefix or suffix and keep all spaces in there
2) specifically, this looks wrong--there is no space between edition and title, for example.

SnooksHarvard.csl
```diff
@@ -0,0 +1,188 @@
...
+    <sort>
+      <key macro="author"/>
+      <key variable="title"/>
+    </sort>
+    <layout suffix=".">
+      <text macro="author"/>
+      <text macro="year-date" prefix=" "/>
+      <choose>
+        <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
+          <group prefix="" delimiter="">
```
adam3smith: remove empty affixes and delimiters throughout the style, please.

SnooksHarvard.csl
```diff
@@ -0,0 +1,188 @@
...
+        <else-if type="chapter paper-conference" match="any">
+          <text macro="title" prefix=", "/>
+          <group prefix=", " delimiter="">
+            <text term="in" text-case="lowercase" suffix=" "/>
+            <text macro="editor" suffix=", "/>
+            <text variable="container-title" font-style="italic"/>
+            <text variable="collection-title" prefix=", "/>
+            <text variable="event" prefix=", "/>
+            <group suffix="." delimiter=", ">
+              <text macro="publisher" prefix=", "/>
```
adam3smith: again, move the prefix up to the group.

SnooksHarvard.csl
```diff
@@ -0,0 +1,188 @@
...
+              <text macro="pages"/>
+            </group>
+          </group>
+        </else-if>
+        <else-if type="thesis">
+          <group>
+          <text macro="title" prefix=", "/>
+            <text variable="genre"/>
```
adam3smith: same issue with no space after title.

adam3smith:
we'll have a bit more to say on the info section and naming (and your input on naming is very much welcome), but if you want to get started on fixing the issues I note inline?

You can update your pull request by editing your style here (click the edit(pencil) button):
https://github.com/s002610/styles/blob/patch-4/SnooksHarvard.csl

SnooksHarvard.csl
s002610: I tried this, but it would only generate double quotes.

rmzelle: Adding `default-locale="en-GB"` (as @adam3smith suggested earlier) should automatically switch the style to using single quotes, as long as you're using `quotes="true"`.

adam3smith: once you have the default-locale set to en-GB the quotes will be correct, too.

s002610:
I'll get to work on those changes. 

The copyright for the style guide is Commonwealth of Australia, so that probably warrants a mention in the name. Also, the guide specifies its format for a couple of different styles, so I think it is important to mention Harvard in the name. Maybe Snooks & co Australia Harvard Style / snooks-and-co-australia-harvard-style.csl

SnooksHarvard.csl
s002610: Not sure what you mean here. There are no spaces in suffixes that I can see, and everything is rendering fine. I have tidied it up, though.

adam3smith: there's a space in the suffix for line 141 right below this. And I don't see how there could be a space (let alone punctuation) between title and edition the way you have this written. Might be you haven't tested with edition.

s002610:
Ok, I can't seem to get this past Travis CI and I can't work out why. There's apparently something wrong with the pages macro, which looks like this: 

  </macro>
   <macro name="pages">
    <group>
      <label variable="page" form="short"/>
      <text variable="page"/>
    </group>
  </macro>

rmzelle:
Travis complains because the macro "pages" (or "page") isn't used anywhere in the style. So just delete it.

And for the style title, how about: "Style Manual - Australian Government (author-date)" and "style-manual-australian-government.csl"? We try to avoid the term "Harvard" since it's a bit of a misnomer, and it doesn't really look like people refer to this manual as "Snooks".

s002610:
Oh, I didn't realise it wasn't used. That's easy fixed. No problem. I'll fix the name while I'm at it.

rmzelle:
@adam3smith, can you give this a final look-over and merge? Looks good to me.

style-manual-australian-government.csl
```diff
@@ -0,0 +1,181 @@
...
+            <text macro="editor" prefix=" "/>
+            <text variable="container-title" font-style="italic" prefix=", "/>
+            <text variable="collection-title" prefix=", "/>
+            <text variable="event" prefix=", "/>
+            <group>
+              <text macro="publisher"/>
+         </group>
+            <group prefix=", ">
+              <label variable="page" form="short"/>
+              <text variable="page" prefix=". "/>
```
adam3smith: remove the period in the prefix: it's already in the page label. Zotero strips the double period, other processor may not be as forgiving.

style-manual-australian-government.csl
```diff
@@ -0,0 +1,181 @@
...
+      </if>
+    </choose>
+  </macro>
+  <macro name="title">
+    <choose>
+      <if type="bill book graphic legal_case legislation motion_picture report song thesis webpage" match="any">
+        <text variable="title" font-style="italic"/>
+      </if>
+      <else>
+        <text variable="title" text-case="title" quotes="true"/>
```
adam3smith: the text-case="title" here, are you sure about that? Seems very odd that only article titles would be title-cased?

style-manual-australian-government.csl
```diff
@@ -0,0 +1,181 @@
...
+      <name>jstedman</name>
+      <email>jessica.stedman@gmail.com</email>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="generic-base"/>
+    <summary>Snooks style manual Australia Harvard author-date style</summary>
+    <updated>2014-09-10T06:18:55+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="en">
```
adam3smith: would you mind fixing the indenting here?

adam3smith:
three small comments, everything else looks great, thanks!

rmzelle:
Thanks! Sorry for letting this linger for a few days (we don't get email updates for commits).

PR #1432, opened 2015-03-06
Minor tweaks to apa.csl

bwiernik:
Several minor tweaks to the core APA style. The date for "speech" item is discussed here (https://forums.zotero.org/discussion/45252/apa-6th-conference-paper-error/#Item_9). Other notes are in commit messages.

Do you have a script to automatically update the various APA offshoots?

### DIFF
--- a/apa.csl
+++ b/apa.csl
@@ -196,9 +196,7 @@
     <text macro="format"/>
   </macro>
   <macro name="format">
-    <group delimiter=" " prefix=" [" suffix="]">
-      <text variable="medium" text-case="capitalize-first"/>
-    </group>
+    <text variable="medium" text-case="capitalize-first" prefix=" [" suffix="]"/>
   </macro>
   <macro name="publisher">
     <choose>
--- a/apa.csl
+++ b/apa.csl
@@ -129,7 +129,7 @@
       <else>
         <choose>
           <if variable="DOI">
-            <text variable="DOI" prefix="doi:"/>
+            <text variable="DOI" prefix="http://dx.doi.org/"/>
           </if>
           <else>
             <choose>
--- a/apa.csl
+++ b/apa.csl
@@ -193,6 +193,12 @@
         </group>
       </else-if>
     </choose>
+    <text macro="format"/>
+  </macro>
+  <macro name="format">
+    <group delimiter=" " prefix=" [" suffix="]">
+      <text variable="medium" text-case="capitalize-first"/>
+    </group>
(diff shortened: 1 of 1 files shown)


adam3smith:
no, I'm too worried that a script would get things wrong. I do that manually with a diff viewer, which is reasonably fast.

apa.csl
```diff
@@ -129,7 +129,7 @@
       <else>
         <choose>
           <if variable="DOI">
-            <text variable="DOI" prefix="doi:"/>
+            <text variable="DOI" prefix="http://dx.doi.org/"/>
```
adam3smith: I know they say dx.doi.org, but could we go with doi.org? It's both shorter and preferred by CrossRef

bwiernik: I asked the APA Style Experts. They basically said it didn't matter but that the technically correct version would be "dx." I think we can safely just do "doi.org".

apa.csl
```diff
@@ -378,14 +384,21 @@
...
+            <if variable="volume" match="any">
+              <group>
+                <text term="volume" form="short" text-case="capitalize-first" suffix=" "/>
+                <number variable="volume" form="numeric"/>
+              </group>
+            </if>
+            <else>
+              <group>
+                <text term="volume" form="short" plural="true" text-case="capitalize-first" suffix=" "/>
+                <number variable="number-of-volumes" form="numeric" prefix="1-"/>
```
adam3smith: it's been there for a long time, but this has got to be an en dash, not a hyphen, right?

adam3smith: yup, it should be: http://blog.apastyle.org/apastyle/2013/09/how-to-cite-an-anthology-or-collected-works.html
would you mind fixing this?

apa.csl
adam3smith: yeah, let's.

apa.csl
bwiernik: Do I enter it as prefix="1&#8211" or just type the en dash?

adam3smith:
for the en dash - both a literal and an escaped en dash work. We have a preference for escaping for hard to distinguish characters, so what you did is great.

adam3smith:
thanks. 10 commit messages seemed like a bit much, so I added this manually. I copied all relevant commit messages over so we have the history.

bwiernik:
Sorry for the mess.

adam3smith:
no, very helpful to have the details.

PR #1433, opened 2015-03-06
modern-language-association-6th-edition-note Editor label fixes

serpodrick:
The label editor should be preceded by a full-stop (not a comma), be abbreviated, and start with a capital letter.

Also, remove a group containing a stray text tag.


adam3smith:
I didn't try to dig up an old MLA manual to check, but I'm sure this is right. Thanks

PR #1436, opened 2015-03-09
american-chemical-society Misc. changes

serpodrick:
Fixes:
* Include issue number for journal articles.
* Format encyclopedia and dictionary articles correctly.

These, with the references in the commit messages, should be self-explanatory.

Change:
* Include translators and series editors.

Translators are mentioned in the "miscellaneous information" section in the manual as "additional information about a book that is important for the reader to know". Collection editors aren't specifically mentioned, but they would go here, too.

### DIFF
--- a/american-chemical-society-author-date.csl
+++ b/american-chemical-society-author-date.csl
@@ -23,11 +23,33 @@
     <updated>2012-09-27T22:06:38+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
+  <locale xml:lang="en">
+    <terms>
+      <term name="editortranslator" form="short">
+        <single>ed. and translator</single>
+        <multiple>eds. and translators</multiple>
+      </term>
+      <term name="translator" form="short">
+        <single>translator</single>
+        <multiple>translators</multiple>
+      </term>
+      <term name="collection-editor" form="short">
+        <single>series ed.</single>
+        <multiple>series eds.</multiple>
+      </term>
+    </terms>
+  </locale>
   <macro name="editor">
-    <names variable="editor">
-      <name sort-separator=", " initialize-with=". " name-as-sort-order="all" delimiter=", " delimiter-precedes-last="always"/>
-      <label form="short" prefix=", " text-case="capitalize-first"/>
-    </names>
+    <group delimiter="; ">
+      <names variable="editor translator" delimiter="; ">
+        <name sort-separator=", " initialize-with=". " name-as-sort-order="all" delimiter=", " delimiter-precedes-last="always"/>
+        <label form="short" prefix=", " text-case="title"/>
+      </names>
+      <names variable="collection-editor">
+        <name sort-separator=", " initialize-with=". " name-as-sort-order="all" delimiter=", " delimiter-precedes-last="always"/>
+        <label form="short" prefix=", " text-case="title"/>
+      </names>
+    </group>
   </macro>
   <macro name="author">
     <names variable="author" suffix=".">
--- a/american-chemical-society-author-date.csl
+++ b/american-chemical-society-author-date.csl
@@ -85,7 +85,11 @@
   <macro name="book-container">
     <group delimiter=" ">
       <text macro="title" suffix="."/>
-      <text term="in" text-case="capitalize-first"/>
+      <choose>
+        <if type="entry-dictionary entry-encyclopedia" match="none">
+          <text term="in" text-case="capitalize-first"/>
+        </if>
+      </choose>
       <text variable="container-title" font-style="italic"/>
     </group>
   </macro>
@@ -195,7 +199,7 @@
             <date variable="issued" form="text"/>
           </group>
         </else-if>
-        <else-if type="chapter paper-conference" match="any">
+        <else-if type="chapter paper-conference entry-dictionary entry-encyclopedia" match="any">
           <group delimiter="; ">
             <text macro="book-container"/>
             <text macro="editor"/>
--- a/american-chemical-society-author-date.csl
+++ b/american-chemical-society-author-date.csl
@@ -227,7 +227,20 @@
             <text variable="container-title" font-style="italic" form="short"/>
             <group delimiter=", ">
               <text macro="issued" font-weight="bold"/>
-              <text variable="volume" font-style="italic"/>
+             <choose>
+                <if variable="volume">
+                  <group delimiter=" ">
+                    <text variable="volume" font-style="italic"/>
+                    <text variable="issue" prefix="(" suffix=")"/>
+                  </group>
+                </if>
+                <else-if variable="issue">
+                  <group delimiter=" ">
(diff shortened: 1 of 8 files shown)


serpodrick:
No, that would not give the same result; the editortranslator term is not used if collection-editor is added to the variable attribute.

rmzelle:
Ah, you're right.

@adam3smith, while this follows the spec (http://citationstyles.org/downloads/specification.html#names) to the letter, maybe we should change 'when the selection *consists of* "editor" and "translator"' to 'when the selection *includes* "editor" and "translator"'? Thoughts? Or doesn't this particular case of having to duplicate `cs:names` happen often enough to be worth changing?

adam3smith:
series editors aren't cited very often in the first place, so this is quite rare; I say we leave this as is for simplicity's sake.

adam3smith:
@rmzelle - I'm fine with these. You did the last big round of ACS edits, so I'll let you merge this.

rmzelle:
Uhm, where did Travis go? (cc'ing @aurimasv)

aurimasv:
Not sure. Seems like Travis missed the web hook somehow (maybe temporary service outage). Try closing the PR and opening it back up again. If the S3 issue from yesterday is still not fixed, this should fail anyway though.

adam3smith:
@rmzelle - Travis is back, albeit still broken.

rmzelle:
I asked at https://twitter.com/rintzezelle/status/576778636902154241 and https://twitter.com/rintzezelle/status/576779085369761792

rmzelle:
We could get rid of our S3 caching, but that would slow down the builds as long as Travis doesn't have bundler caching enabling (see http://docs.travis-ci.com/user/caching/#Enabling-Bundler-caching and https://github.com/travis-ci/travis-build/pull/148). (also cc @inukshuk)

aurimasv:
I linked to an explanation for secure variable in the other PR. I don't think you ever hit the part of the script that relied on it before, since the bundle was always available on S3. Now that it's inaccessible, the script tries to upload it. I don't think this was ever going to work from a PR from another fork. Not sure what the actual issue with S3 is, but if the bundle is just missing, you can try creating a PR from this fork (it can be any small thing), trigger the build, the upload should succeed, and you can close the PR. Future builds may then succeed. The question would still remain as to what happened to the bundle on S3 though.

rmzelle:
Yeah, so this probably can be fixed on AWS. @dstillman, can you take a look?

For comparison, see https://travis-ci.org/citation-style-language/styles/builds/54088731 (pass) and https://travis-ci.org/citation-style-language/styles/builds/54331459 (fail).

dstillman:
Sorry, what am I fixing? The citationstyles-cache bucket is empty.

aurimasv:
Any clue how it became empty?

rmzelle:
Any clue why we get a 403 when trying to retrieve http://citationstyles-cache.s3.amazonaws.com/citation-style-language/styles/2.1-79219e2394b8c4bc4eb705eaabae71a0.tgz ?

inukshuk:
@aurimasv is right, the caching should not work for a PR from a forked repository (see here: http://docs.travis-ci.com/user/pull-requests/)

(Regardless of what's causing this issue, we could also start using Travis CI's bundler caching, by simply setting `sudo: false` and `cache: bundler` in the config file.)

dstillman:
> Any clue why we get a 403

Because it's an unauthenticated request for a file that doesn't exist, and the default bucket permissions don't allow public file access. It's possible for an upload to mark a file as publicly available, which I assume Travis does by default (unless they have documentation specifying that the bucket should be publicly available).

Are you sure the bucket hasn't always been empty? The passing test doesn't show the bundle being downloaded, just that it exists. Is there somewhere else that it could've been cached that was perhaps cleared manually during the S3 issues?

inukshuk:
We definitely tested this when we first set it up (I even deleted cached bundles from S3 manually).

You also see it in the output of bundle install: in the build where the download was successfull, you see that all the Gems are insalled as 'using...' which means it is already installed whereas in the failing build they are installed as 'installing...'. That means that in one case the gems are locally available because they have been downloaded from S3. No idea why the file was deleted though.. @rmzelle when did this happen?

rmzelle:
I remember we discussed to have the cache expire every three months or so, so the bucket wouldn't fill up with expired builds.

aurimasv:
@inukshuk #1450 was the first PR to fail

rmzelle:
@inukshuk, according to the Travis documentation, `cache: bundler` doesn't do anything yet for public repos.

dstillman:
Oh, yeah, there's a 90-day lifecycle rule on the bucket.

inukshuk:
@rmzelle I still think they allow some kind of caching — I enabled it in many repos and most of the Gems are being cached.  But in that case we better stcik with our setup for now.

@dstillman that's right, I forgot about that.

So does re-running a build from the repo istelf create a new cached bundle? Perhaps we should remove the 90-day lifecycle to make sure PR's don't fail because of it?

rmzelle:
So, pushing a new branch to the repo fixed things (https://travis-ci.org/citation-style-language/styles/builds/54385201). I restarted all the failing builds, and they all passed.

@dstillman, is it acceptable to keep the cache from expiring? There has been some recent activity for https://github.com/travis-ci/travis-build/pull/148, so hopefully we can use Travis' bundle caching soon anyway.

And @inukshuk, I thought that it was sufficient to use `sudo: false`, but I don't see (proper) caching e.g. at https://travis-ci.org/citation-style-language/locales/builds/52088646, but maybe I should add `cache: bundler`?

Sorry for the hubbub, everybody.

inukshuk:
@rmzelle looking at http://docs.travis-ci.com/user/caching/ I think that caching is available already. It says, at the top of the page: "The features described here are currently only available for private repositories on travis-ci.com and our new container-based infrastructure." -- I read this to mean, that you need to set `sudo: false` (which enables the docker build environment) in order to make use of it.

Caching is not enabled automatically yet, but we can enable it using `cache: bundler` -- (to answer your question, yes, it would not work currently without that setting).

rmzelle:
@inukshuk, ah, thanks, I misread that, then.

dstillman:
I've upped the lifecycle rule to 365 days. Let me know if that's not sufficient.

rmzelle:
@inukshuk, can you help me out here: https://github.com/citation-style-language/styles/commit/27ff7ab331c121cea74190be32ba671676736b00 and https://travis-ci.org/citation-style-language/styles/builds/54389806

Caching through Travis works, but I'm having trouble with making RSpec ignore the extra files.

inukshuk:
@rmzelle it should be sufficient to add 'vendor' to the list of extra files at: https://github.com/citation-style-language/styles/blob/master/spec/spec_helper.rb#L38

(You can test locally by adding a vendor directory with files in it)

rmzelle:
@dstillman, feel free to k̶i̶c̶k̶ delete the bucket. I just set up Travis's own caching, and we're getting comparable performance as we did with our own S3 bucket.

rmzelle:
Sorry, @serpodrick, but this looks good now!

PR #1437, opened 2015-03-09

zuphilip:
reported in https://forums.zotero.org/discussion/46953/zeitschrift-fuer-paedagogik-german/#Item_6


adam3smith:
thanks!

PR #1438, opened 2015-03-10
Update to harvard-cranfield-university.csl

ivyleavedtoadflax:
* Added support for [Blog] and [Wiki] as chapter using genre field.
* Added support for [Kindle] as book using genre field.
* Added support for Conference location outside of title using genre field.


rmzelle:
Thanks!

PR #1440, opened 2015-03-11
Create Journal-of-the-Japan-Medical-Library-Association.csl

hellboy84:
I create the citation style for "Journal of the Japan Medical Library Association".
It's new style.


rmzelle:
Note that it's quite easy to modify an existing pull request. Next time, no need to close it if the Travis build fails.

Continued in #1441.

PR #1441, opened 2015-03-11
Create journal-of-the-japan-medical-library-association.csl

hellboy84:
I create the citation style for Journal of the Japan Medical Library Association.

### DIFF
--- a/igaku-toshokan.csl
+++ b/igaku-toshokan.csl
@@ -29,23 +29,6 @@
       <term name="section" form="short">sect.</term>
     </terms>
   </locale>
-  <locale xml:lang="fr">
-    <date form="text" delimiter=" ">
-      <date-part name="day"/>
-      <date-part name="month" form="short" strip-periods="true"/>
-      <date-part name="year"/>
-    </date>
-    <terms>
-      <term name="retrieved">disponible</term>
-      <term name="from">sur</term>
-    </terms>
-  </locale>
-  <locale xml:lang="de">
-    <terms>
-      <term name="retrieved">verfügbar</term>
-      <term name="from">unter</term>
-    </terms>
-  </locale>
   <macro name="author">
     <names variable="author">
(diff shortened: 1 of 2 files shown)


rmzelle:
Do you expect people to use this CSL style primarily for English citations, or for Japanese ones?

hellboy84:
Thank you for your comment, I expect people to use this style primarily for Japanese citations.

rmzelle:
In that case, would it make sense to use the Japanese name ("Igaku Toshokan", or even "医学図書館") instead of "Journal of the Japan Medical Library Association" for style's title?

The style should probably also be set to Japanese (point 3 of https://github.com/citation-style-language/styles/wiki/Style-Requirements ). (the locale code for Japanese is "ja-JP")

hellboy84:
That's my mistake! I revised codes according to your proposal.
I have changed language code and title and Style ID, so file name has changed also. Would this be correct?

rmzelle:
Great.

Can you also delete

```
  <locale xml:lang="fr">
    <date form="text" delimiter=" ">
      <date-part name="day"/>
      <date-part name="month" form="short" strip-periods="true"/>
      <date-part name="year"/>
    </date>
    <terms>
      <term name="retrieved">disponible</term>
      <term name="from">sur</term>
    </terms>
  </locale>
  <locale xml:lang="de">
    <terms>
      <term name="retrieved">verfügbar</term>
      <term name="from">unter</term>
    </terms>
  </locale>
```

? That's for French and German localization, which isn't relevant for this style (the section for English might come in handy in the future, so you can leave that).

hellboy84:
Thanks a lot!
I deleted French and German section.

rmzelle:
Perfect, thanks!

hellboy84:
I appreciate your help!

rmzelle:
You're welcome! Thanks for contributing!

PR #1450, opened 2015-03-13
Minor ASM ch-ch-changes

aurimasv:
Based on MCB and JCM styles, but it looks like the style guide is copy-pasted for all ASM journals.

PR #1451, opened 2015-03-14
Echo bundle name that was not cached on S3

aurimasv:
Helpful to see why the bundle needs to be (re)built

Re #1450

### DIFF
--- a/script/cached-bundle
+++ b/script/cached-bundle
@@ -24,7 +24,7 @@ compute_md5() {
 }
 
 download() {
-  curl --tcp-nodelay -qsfL "$1" -o "$2"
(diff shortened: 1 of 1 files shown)


aurimasv:
Well, as you can see from the last build there, S3 server is returning 403 (Forbidden) for the bundle, so something has recently changed on the S3 side

aurimasv:
I guess it would be helpful to clarify that the 403 error is AccessDenied (when [loaded in a browser](http://citationstyles-cache.s3.amazonaws.com/citation-style-language/styles/2.1-79219e2394b8c4bc4eb705eaabae71a0.tgz))

PR #1452, opened 2015-03-14
Sort bibliography, as citations:  date descending

karnesky:
As reported in email from Carsten Allefeld.


rmzelle:
Confirmed. From http://www.elsevier.com/journals/biological-conservation/0006-3207/guide-for-authors#68000:

"Groups of references should be listed first alphabetically, then chronologically.
Examples: 'as demonstrated (Allan, 2000a, 2000b, 1999; Allan and Jones, 1999). Kramer et al. (2010) have recently shown ....' "

rmzelle:
Thanks!

PR #1455, opened 2015-03-16
Update society-for-general-microbiology.csl

zuphilip:
* use a group for container-title, collection-title to correct errornous comma when no container-title is present (see https://forums.zotero.org/discussion/47569/microbiology-journal-book-series-issue/ ) 
 * use another group to avoid to output "in " when containter-title and collection-title are missing


adam3smith:
Thanks!

PR #1456, opened 2015-03-17
complete style overhaul

scottcove:
Following an extensive review of the style over the (SH) summer, this updates the essay style to the current specs.

### DIFF
--- a/melbourne-school-of-theology.csl
+++ b/melbourne-school-of-theology.csl
@@ -1,13 +1,13 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" and="symbol" page-range-format="expanded" demote-non-dropping-particle="sort-only" default-locale="en-GB">
-  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" and="symbol" page-range-format="expanded" demote-non-dropping-particle="sort-only" default-locale="en-US">
+<!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
     <title>Melbourne School of Theology</title>
     <title-short>MST</title-short>
     <id>http://www.zotero.org/styles/melbourne-school-of-theology</id>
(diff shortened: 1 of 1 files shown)


melbourne-school-of-theology.csl
```diff
@@ -255,7 +294,7 @@
     <choose>
       <if type="bill book chapter graphic legal_case legislation motion_picture paper-conference report song" match="any">
         <choose>
-          <if is-numeric="edition">
+          <if is-numeric="edition note">
```
adam3smith: I'll have more tomorrow, but this is almost certainly not what you mean. What's the goal here of adding "note"?

melbourne-school-of-theology.csl
```diff
@@ -496,24 +574,32 @@
...
+                    <text value="of" prefix=" " suffix=" "/>
+                  </if>
+                </choose>
               </if>
             </choose>
             <group>
               <number variable="number-of-volumes" form="numeric"/>
-              <text term="volume" form="short" plural="true" strip-periods="true" prefix=" "/>
+              <choose>
+                <if match="none" variable="volume">
```
adam3smith: this can't be right - you only write "vols" if there is no volume number???. What you had before was probably right. If you have a term and a variable in a group, the term only gets printed if the variable is present.

scottcove: Yes, this is correct.  If there is a volume number, it is rendered "vol x of x".  If there is no volume number, because it is explicitly stated in the title, for example, but there is a number of volumes, then it is rendered "x vols".  It removes the redundancy of having whitespace in the form of "vol  of x" This is addressing a very specific, but very common case that is used in our university a lot.

melbourne-school-of-theology.csl
scottcove: That is an odd one.  I'm unsure how that snuck in there!  I must confess to using the Visual Editor almost exclusively, but do not remember slipping that in!  I shall remove it..

rmzelle:
@scottcove, our automatic checks fail because the style contains two unused macros ("container-contributors-short", "point-locators-sub"), which isn't allowed.

scottcove:
Unused macros removed.  Apologies for that.  I knew I had forgotten to remove point-locators-sub, but didn't even realise that container-contributors-short was unused..

rmzelle:
@adam3smith, will you review this further?

adam3smith:
looks good, thanks!

PR #1457, opened 2015-03-17
studia-bas.csl

gbalcerowski:
changes to "book" category:
- removed unnesessary comma beetween publisher place and a date


rmzelle:
Thanks!

PR #1459, opened 2015-03-18
Fix interpunctation in biological-psychiatry.csl

zuphilip:
This should now work smoothly w.r.t. to the interpunctations. However, the style guid is not covering every case and in the discipline normally only journal articles are cited. Therefore I tried to be consistent with everything I saw and add the rest as it seems to fit into the style. Moreover, I tried to take as the interpunctions outside the macros whenever possible.( Problems with interpuncations were reported: https://forums.zotero.org/discussion/47615/style-error-biological-psychiatry/#Item_2 )


rmzelle:
Thanks!

PR #1461, opened 2015-03-20

ckan9445:
Small corrections:
In macro 'access': underline URLs and suppress 'available at'-term when there's no URL.
In macro 'medium': add 'Online' term (if applicable) after genre.
In macro 'issued': remove repetition of issued year.


rmzelle:
Thanks!

PR #1464, opened 2015-03-20
Fixes to csl for FASTI-FOLDR

Jmuccigr:
I fixed a series of formatting errors, the main one being the switch from in-text to notes class, but also some minor ones in the citation and bibliography sections.

PS My first pull request!

### DIFF
--- a/fold-and-r.csl
+++ b/fold-and-r.csl
@@ -17,7 +17,7 @@
   </info>
   <macro name="author">
     <names variable="author">
-      <name initialize-with="." delimiter=", " and="text" name-as-sort-order="all" sort-separator=", "/>
+      <name initialize-with="." delimiter=", " and="text" name-as-sort-order="all" sort-separator=" "/>
       <label form="short" prefix=", " suffix="," text-case="capitalize-all" strip-periods="true"/>
       <substitute>
         <names variable="editor"/>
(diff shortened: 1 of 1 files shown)


fold-and-r.csl
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
```
adam3smith: you also need to change line 12 below https://github.com/citation-style-language/styles/pull/1464/files#diff-a03802aca1da523688424503e1158fcfR12 to note instead of author-date. That's why the automated test fails.

adam3smith:
everything else looks good. Looks like you know how to update your pull request. Let us know when you're done, we don't get notifications for updated pull requests.

fold-and-r.csl
Jmuccigr: Done.

Thanks for explaining. I couldn't figure it out from the test output and the github csl validation checked out.

adam3smith:
thanks!

PR #1465, opened 2015-03-21
Add "director" to "secondary-contributors" macro

nickbart1980:
For rendering the director element in, e.g., *Movie Title*. 2004. Directed by Mary Smith. Hollywood, CA: Bigshot Productions. 
See http://www.chicagomanualofstyle.org/qanda/data/faq/topics/Documentation/faq0020.html

### DIFF
--- a/chicago-author-date.csl
+++ b/chicago-author-date.csl
@@ -42,10 +42,16 @@
   <macro name="secondary-contributors">
     <choose>
       <if type="chapter paper-conference" match="none">
-        <names variable="editor translator director" delimiter=". ">
-          <label form="verb" text-case="capitalize-first" suffix=" "/>
-          <name and="text" delimiter=", "/>
-        </names>
+        <group delimiter=". ">
+          <names variable="editor translator" delimiter=". ">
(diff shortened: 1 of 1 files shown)


adam3smith:
the way citeproc-js interprets the specs for cs:names (rightly so, given the current wording):
http://docs.citationstyles.org/en/stable/specification.html#names
this would break the editortranslator functionality. 
Either add this as a separate element in a group, or we can think about revisiting this part of the specs; came up recently and @rmzelle suggested changing this: https://github.com/citation-style-language/styles/pull/1436#issuecomment-78958861 I said not worth it bc it's too rare, but I don't have any fundamental objections.

nickbart1980:
I see – I added director as a separate element in a group now and committed the change – is this ok or do you need another pull request?

adam3smith:
this is great, we're not super picky about minimizing the number of commits.

nickbart1980:
Same problem in apa.csl, by the way: https://github.com/citation-style-language/styles/blob/master/apa.csl#L48 and #L58. Revising the specs might make sense after all.

PR #1467, opened 2015-03-23
Fixed one last missed attribute double-space in harvard-leeds-beckett-university.csl

joejoinerr:


adam3smith:
thanks (btw. all implementations handle the update times themselves. There's no need to manually adjust that, though obviously it also doesn't hurt).

joejoinerr:
The more you know :)

rmzelle:
Actually, nowadays we have https://github.com/citation-style-language/styles-distribution, which mirrors this repository. It only gets updated when the Travis CI build passes, and only creates one commit per pull request (for this one, the changes are represented in https://github.com/citation-style-language/styles-distribution/commit/5fc46ffa69743b8e5d602cd68eca06090017809e ).

And when it updates, the style's time stamp is automatically updated as well!

joejoinerr:
@rmzelle Good to know for future reference, thanks.

PR #1468, opened 2015-03-23
Fix for editor translator in container-contributors & secondary-contributors macros

nickbart1980:
Similar issue to the one discussed in https://github.com/citation-style-language/styles/pull/1465.
Not quite sure about the format here, though. Common sense would suggest author(s) come before editor(s) and translator(s) – but how are authors separated then from editors/translators in the output? Couldn't find any example in the manual.

### DIFF
--- a/apa.csl
+++ b/apa.csl
@@ -45,20 +45,32 @@
   <macro name="container-contributors">
     <choose>
       <if type="chapter paper-conference entry-dictionary entry-encyclopedia" match="any">
-        <names variable="editor translator container-author" delimiter=", " suffix=", ">
-          <name and="symbol" initialize-with=". " delimiter=", "/>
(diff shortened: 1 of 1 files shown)


adam3smith:
thanks. @rmzelle -- so should we go ahead and suggest changing that after all?
@fbennett would it be hard to change the behavior in the processor so that editortranslator also works when there are other variables in the same cs:names?

rmzelle:
Sure, feel free to create a ticket (in the "documentation" repo?). We might want to restrict it to cases where "editor" is a direct neighbor of "translator", though, so that e.g.

`variable="container-author editor translator"`

invokes "editortranslator", whereas with

`variable="editor container-author translator"`

each name variable would be rendered individually.

adam3smith:
I'm going to wait for Frank, see what he says re: implementation.

PR #1469, opened 2015-03-24
Avoid capitalization of unneeded words.

ran81:
This PR removes the ``text-case="title"`` for the title container macro in order to avoid capitalizing unneeded words, for example "Journal of Political Economy" should not be "Journal Of Political Economy" (the "of" should be left as is.

### DIFF
--- a/chicago-author-date.csl
+++ b/chicago-author-date.csl
@@ -401,7 +401,7 @@
     </choose>
     <choose>
       <if type="legal_case" match="none">
-        <text variable="container-title" text-case="title" font-style="italic"/>
+        <text variable="container-title" font-style="italic"/>
       </if>
     </choose>
   </macro>


RoySegall:
@rbran100

adam3smith:
which reference manager are you testing this with? "of" should never get capitalized with text-case="title" see http://docs.citationstyles.org/en/stable/specification.html#title-case-conversion
Since some catalogs store journal data in sentence format (Journal of political economy), the attribute is correct here.

RoySegall:
We opened this PR due to this issue: https://github.com/openscholar/openscholar/issues/7038.

rbran100:
@RoySegall I think @adam3smith is right here, it looks like our biblio CSL processing is what is incorrect not the text-case tag.  I have reverted the change on our end and we should fix the processing. https://github.com/openscholar/openscholar/issues/7038

adam3smith:
Thanks for getting back so quickly. Closing this issue then.

rmzelle:
Does the current stable version of OpenScholar already use CSL? I like to keep tally of all projects that use CSL at http://citationstyles.org/.

rbran100:
Yes it does, along with quite a few other styles.  Thanks for your work.

rmzelle:
So you're relying on https://www.drupal.org/project/biblio ?

PR #1470, opened 2015-03-24
Update plant-ecology.csl

zuphilip:
* Temporary fix, while we are working on Springer styles.
 * Reminder: https://github.com/citation-style-language/utilities/issues/19
 * Reference https://forums.zotero.org/discussion/47789/plant-ecology-style/#Item_2


adam3smith:
thanks!

PR #1474, opened 2015-03-25
Update systematic-and-applied-microbiology.csl

Sankar-GEO-ELS:
removed the sorting node from under ‘Inline Citations’ and added a new Sort node under ‘Bibliography’ header specifying 'author'

### DIFF
--- a/systematic-and-applied-microbiology.csl
+++ b/systematic-and-applied-microbiology.csl
@@ -112,14 +112,14 @@
     </choose>
   </macro>
   <citation collapse="citation-number">
-    <sort>
-      <key variable="citation-number" sort="ascending"/>
-    </sort>
     <layout delimiter="," prefix="[" suffix="]">
       <text variable="citation-number"/>
     </layout>
   </citation>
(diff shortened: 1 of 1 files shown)


systematic-and-applied-microbiology.csl
```diff
@@ -112,14 +112,14 @@
     </choose>
   </macro>
   <citation collapse="citation-number">
-    <sort>
-      <key variable="citation-number" sort="ascending"/>
-    </sort>
     <layout delimiter="," prefix="[" suffix="]">
```
adam3smith: I don't think that's right. With this deleted, you could get citations out of order in a single citation in the text, things like, [14, 3, 7, 2]. That can't be right. You can leave this in and still get alphabetical sorting in the bibliography.

Sankar-GEO-ELS: OK Sebastian, I understood. Please ignore this pull request and I will create another pull request after fixing this. One author is waiting to use this style sheet.

adam3smith:
continued in https://github.com/citation-style-language/styles/pull/1478

PR #1476, opened 2015-03-25
Update harvard-cape-peninsula-university-of-technology.csl

LaraLibrarian:
moved the comma-space to the suffix of the in-text author, instead of a prefix to the year.

### DIFF
--- a/harvard-cape-peninsula-university-of-technology.csl
+++ b/harvard-cape-peninsula-university-of-technology.csl
@@ -109,8 +109,8 @@
   <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true">
     <layout prefix="(" suffix=")" delimiter="; ">
       <group>
-        <text macro="author-short"/>
-        <text macro="year-date" prefix=", "/>
+        <text macro="author-short" suffix=", "/>
+        <text macro="year-date"/>
       </group>
       <group>
         <choose>
--- a/harvard-cape-peninsula-university-of-technology.csl
+++ b/harvard-cape-peninsula-university-of-technology.csl
@@ -109,8 +109,10 @@
   <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true">
     <layout prefix="(" suffix=")" delimiter="; ">
       <group>
-        <text macro="author-short" suffix=", "/>
-        <text macro="year-date"/>
+        <group delimiter=", "/>
+          <text macro="author-short"/>
+          <text macro="year-date"/>
+        </group>
       </group>
       <group>
         <choose>


harvard-cape-peninsula-university-of-technology.csl
```diff
@@ -109,8 +109,8 @@
   <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true">
     <layout prefix="(" suffix=")" delimiter="; ">
       <group>
```
adam3smith: you really should just be using ```delimiter=", "``` for the group here instead of prefix/suffix.

rmzelle: I.e.,

```
      <group delimiter=", ">
        <text macro="author-short"/>
        <text macro="year-date"/>
      </group>
```

You can update the current pull request via the pencil (edit) icon at https://github.com/LaraLibrarian/styles/blob/patch-3/harvard-cape-peninsula-university-of-technology.csl

LaraLibrarian: I tested out the deliminator suggestion on csl.mendeley.com, but it doesn't give me the result that I want: it removed the comma completely from the citation. The change I suggested works though.

What am I doing wrong? I'm not nearly a developer and my only exposure to working with xml has been this.

adam3smith: If you're not getting a comma, then you're not trying the code as suggested by @rmzelle above. I just tested on csl.mendeley.com and it works as it should. Not sure how you're testing this -- one thing is that CSL (like all XML) is picky about details. Note it's ```delimiter``` not "deliminator". 
(The reason delimiter is better, btw., is that it's more robust. Should, e.g. a reference manager offer an option to just display an author and not the year, your code would produce odd results).

LaraLibrarian: Facepalm! Sorry, that was a totally dyslexic moment. Of course it 'delimiter'. It works like a charm in csl.mendeley.com now. I'll make the changes here.

And yes, you are absolutely right about the delimiter being more robust, I was wondering that.

harvard-cape-peninsula-university-of-technology.csl
```diff
@@ -109,8 +109,10 @@
...
-        <text macro="author-short"/>
-        <text macro="year-date" prefix=", "/>
+        <group delimiter=", "/>
```
adam3smith: you can't have the / at the end here. should be  ```<group delimiter=", ">```
It would also be great if you could delete ```<group>``` in line 111 and ```</group>``` in line 116: these don't do anything now. Same update instructions as by @rmzelle above.

rmzelle:
@LaraLibrarian, let us know if you need additional help to get this pull request in proper shape.

rmzelle:
I'm closing this pull request for now because of inactivity. If you wish to fix up your pull request in the future, just post a comment and we can reopen this.

PR #1479, opened 2015-03-25
Update gost-r-7-0-5-2008.csl

churikovaea:
Changes were made to sort sources in bibliography in alphabetical order.


rmzelle:
You changed the style to call the "language" macro, but that macro doesn't actually exist.

PR #1483, opened 2015-03-27

FrancoisC37:
Hello,
this is the correction of two minors errors :
-the call of two time the collection for a chapter
-i've forgot two put a space before the "archive location"
Thanks


rmzelle:
Thanks!

PR #1484, opened 2015-03-27
Create technische-universitat-dresden-medizin.csl

jmiba:


technische-universitat-dresden-medizin.csl
```diff
@@ -0,0 +1,205 @@
...
+    </author>
+    <category citation-format="author-date"/>
+    <category field="medicine"/>
+    <summary>Zitierstil entsprechend den Zitierrichtlinien für die Dissertation, Medizinische Fakultät Carl Gustav Carus, Technische Universität Dresden. Stil basiert auf dem Zitierstil der Zeitschrift Human Mutation</summary>
+    <updated>2015-03-27T18:16:26+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <macro name="editor">
+    <names variable="editor" delimiter=", ">
+      <name et-al-min="0" initialize-with="" name-as-sort-order="all" sort-separator=" "/>
```
adam3smith: what is the goal of ```et-al-min="0"``` here and below? It doesn't really make sense--if you don't want et al, you'd want to just remove this entirely.

technische-universitat-dresden-medizin.csl
```diff
@@ -0,0 +1,205 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
+<!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>Technische Universität Dresden - Medizin</title>
```
zuphilip: Is this a German (only) style? If yes, then you should include the `default-locale` attribute in the <a href="http://docs.citationstyles.org/en/stable/specification.html#the-root-element-cs-style">style node</a> and the title should contain a `(German)` at the end.

technische-universitat-dresden-medizin.csl
```diff
@@ -0,0 +1,205 @@
...
+    <link href="https://www.zotero.org/styles/harvard1" rel="template"/>
+    <link href="http://tu-dresden.de/die_tu_dresden/fakultaeten/medizinische_fakultaet/forschung/die_tu_dresden/fakultaeten/medizinische_fakultaet/forschung/dokumente/dok_akadgrad/documents_newpromordnung/zitierrichtinien_fuer_die_dissertation_formblatt9.0" rel="documentation" xml:lang="DE-de"/>
+    <author>
+      <name>Jens Mittelbach</name>
+      <email>jens.mittelbach@slub-dresden.de</email>
+      <uri>http://www.slub-dresden.de</uri>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="medicine"/>
+    <summary>Zitierstil entsprechend den Zitierrichtlinien für die Dissertation, Medizinische Fakultät Carl Gustav Carus, Technische Universität Dresden. Stil basiert auf dem Zitierstil der Zeitschrift Human Mutation</summary>
```
zuphilip: <blockquote>Stil basiert auf dem Zitierstil der Zeitschrift Human Mutation</blockquote>
What does this exactly mean? The style for this journal already exist: https://www.zotero.org/styles?q=human-mutation

jmiba: The existent journal style is not thoroughly programmed. Also, the new style differs slightly from Human Mutation. It is an exact rendering of the style guide of TU Dresden, Faculty of Medicine.

technische-universitat-dresden-medizin.csl
jmiba: No, it is language independent.

technische-universitat-dresden-medizin.csl
```diff
@@ -0,0 +1,204 @@
...
+    <names variable="editor" delimiter=", ">
+      <label form="short" strip-periods="true" prefix=" (" suffix=")"/>
+    </names>
+  </macro>
+  <macro name="anon">
+    <text term="anonymous" form="short" text-case="capitalize-first" strip-periods="true"/>
+  </macro>
+  <macro name="author">
+    <names variable="author">
+      <name delimiter-precedes-last="never" et-al-min="0" initialize-with="" name-as-sort-order="all" sort-separator=" "/>
```
adam3smith: same here

technische-universitat-dresden-medizin.csl
```diff
@@ -0,0 +1,204 @@
...
+        </if>
+        <else-if type="chapter paper-conference" match="any">
+          <text macro="title" prefix=" " suffix="."/>
+          <group prefix=" " delimiter=" ">
+            <text term="in" text-case="capitalize-first" suffix=":"/>
+            <text macro="editor"/>
+            <text variable="container-title" font-style="normal" suffix="."/>
+            <text variable="event" suffix="."/>
+            <group delimiter=", ">
+              <text macro="publisher" prefix=" "/>
```
adam3smith: you don't need the prefix=" " here -- the group delimiter gives you that.

technische-universitat-dresden-medizin.csl
jmiba: Done.

technische-universitat-dresden-medizin.csl
```diff
@@ -0,0 +1,204 @@
...
+              <text macro="pages"/>
+            </group>
+          </group>
+        </else-if>
+        <else-if type="thesis">
+          <group prefix=" ">
+            <text macro="title" suffix=". "/>
+            <text macro="publisher" suffix=", "/>
+            <text variable="genre"/>
```
adam3smith: I'd much rather see this with a group delimiter or at least prefix on genre -- this is too fragile as soon as "genre" is missing from a thesis (which I think it is in Mendeley).

technische-universitat-dresden-medizin.csl
```diff
@@ -0,0 +1,204 @@
...
+            <group prefix=":">
+              <text variable="page"/>
+            </group>
+          </group>
+        </else>
+      </choose>
+      <text prefix=" " macro="access" suffix="."/>
+    </layout>
+  </bibliography>
+  <locale xml:lang="de">
```
adam3smith: could you please move this up, to right below ```</info>``` that's where it usually is & I'd like to keep it uniform.

adam3smith:
looks good otherwise. Let us know when you're done.

technische-universitat-dresden-medizin.csl
zuphilip: Okay

technische-universitat-dresden-medizin.csl
```diff
@@ -0,0 +1,217 @@
...
+            <text variable="genre" prefix=", "/>
+          </group>
+        </else-if>
+        <else-if type="speech" match="any">
+          <group delimiter=" " prefix=" ">
+            <text macro="title" suffix="."/>
+            <text term="presented at" suffix=":"/>
+            <text variable="event" suffix=","/>
+            <text variable="event-place"/>
+          </group>
```
zuphilip: It is safer to use more `groups` and delimiter here (and less prefix/suffix), e.g.
```
  <text macro="title" suffix="."/>
  <group delimiter=": ">
    <text term="presented at"/>
    <group delimiter=", ">
      <text variable="event"/>
      <text variable="event-place"/>
    </group>
  </group>
</group>
```
This is safer for cases when any variable might be missing.

technische-universitat-dresden-medizin.csl
```diff
@@ -0,0 +1,217 @@
...
+        </else-if>
+        <else>
+          <group suffix=".">
+            <text macro="title" prefix=" "/>
+            <text macro="editor" prefix=" "/>
+          </group>
```
zuphilip: The `else` part is AFAIK mainly for journalArticles. Is this intentially that the editors are following the title and are just sperated by a space? I think you have in every other case that a point is followed after the title.

jmiba: You are right. I didn't think of the unusual (?) or improbable (?) or impossible (?) case of a journal (article) having a bibliographically relevant editor. I'll change this, though, even if I think it will never come to light. By the way, I would like to retain the conditional in the title macro. It makes the two cases look exactly the same, but if anyone would like to use the style as a template, they might want to italicise journal titles as opposed to book titles.

technische-universitat-dresden-medizin.csl
jmiba: It is by default, but you can easily display the field for this item type in the preferences. The "Genre" field in Mendeley is mapped to the CSL field genre. Cf. http://blog.mendeley.com/research-tutorials/howto-edit-citation-styles-for-use-in-mendeley/?utm_source=feedburner&utm_medium=feed&utm_campaign=Feed%3A+MendeleyBlog+%28Mendeley+Blog%29#comment-384552

technische-universitat-dresden-medizin.csl
jmiba: OK, this looks good. I'll change this.

adam3smith:
Looks good to me--thanks!

jmiba:
Thank you for your help!

rmzelle:
Sorry I'm late to the party, but is this style for German and English?

jmiba:
Yes, it is. Works both ways.

rmzelle:
For bilingual styles like this one, we sometimes fix the style to e.g. German, and then create a dependent style for English. See our "Population" styles for an example: https://www.zotero.org/styles?q=population%20(

If you're interested in that, we can help you set this up. It's also fine to just stay with a single style and have users switch the locale as necessary.

jmiba:
Sorry, I got you wrong. Actually it is a fully localized style with just one or two stylistically adjusted terms in German and English. Users can switch to any locale they like.

rmzelle:
But do you expect users to often use this style in languages other than German and English?

PR #1485, opened 2015-03-29
Same author, same year: sort by title

nickbart1980:
see CMS, 16e, 15.19, “Reference list entries with same author(s), same year”


adam3smith:
great, thanks!

PR #1487, opened 2015-03-30
Update protein-science.csl

holeung:
Remove "cited Year" text, which does not conform to journal requirements.


adam3smith:
thanks!

PR #1488, opened 2015-03-30
Create university-college-lillebaelt-apa.csl

roaldfrosig:
APA variation for use by students at University College Lillebælt, Denmark

### DIFF
--- a/university-college-lillebaelt-apa.csl
+++ b/university-college-lillebaelt-apa.csl
@@ -11,8 +11,6 @@
       <email>roaldfrosig@gmail.com</email>
     </author>
     <category citation-format="author-date"/>
-    <category field="psychology"/>
-    <category field="generic-base"/>
     <updated>2015-01-31T22:51:54+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
--- a/university-college-lillebaelt-apa.csl
+++ b/university-college-lillebaelt-apa.csl
@@ -1,0 +1,509 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" and="symbol" initialize-with=". " delimiter-precedes-last="never" default-locale="da-DK">
+  <info>
+    <title>University College Lillebælt - APA (Danish)</title>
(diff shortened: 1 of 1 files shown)


adam3smith:
this is distinct from the regular APA style with some Danish terms switched out?

roaldfrosig:
Yes, the main difference is how legislation and bills and bills are
rendered.

There's also some minor differences like underlining URLs.

On Tue, Mar 31, 2015 at 6:59 AM, Sebastian Karcher <notifications@github.com

rmzelle:
Are there any more variants than the Vancouver and Harvard ones?

roaldfrosig:
No. At University College Lillebælt the only reference styles allowed for
the students are Harvard, Vancouver and APA, so I won't be uploading more
styles.

I do hope that there will come some later corrections. The UCL variations
are in some cases inconsistent or unnecessary idiosyncratic, and I hope our
style requirements will be updated and bettered soon. But until then, all I
can do is to try to make Zotero consistent with the rules our students have
to follow.

rmzelle:
The use of field categories is a bit inconsistent. For this style, we have:

```
    <category field="psychology"/>
```

whereas the Harvard and Vancouver don't list any fields. Is that intentional? It would also be good to add a "template" link to the latter two, so we know which CSL style they're based on.

roaldfrosig:
That's because I made the Harvard and Vancouver styles from scratch - I
found that it was easier than using an existing style.
The APA-style didn't need so much tinkering, and I have a better
understanding of csl now, so I used the APA 6th ed. as template, and
haven't changed the categories.

I can see that having <category field="generic-base"/> is plain wrong. Is
it ok if I just delete both 'category fields' from the APA style, and leave
it at that?

rmzelle:
> I can see that having <category field="generic-base"/> is plain wrong. Is
it ok if I just delete both 'category fields' from the APA style, and leave
it at that?

Sure. For the full list, see
http://docs.citationstyles.org/en/stable/specification.html#appendix-i-categories

"generic-base" is actually acceptable for institution-wide CSL styles that are used for a wide range of disciplines. You can select all the styles with it at https://www.zotero.org/styles. So you can leave it in, too, but then you'd have to update the other two styles (and their dependents).

rmzelle:
Thanks. Will you create English dependents for the APA and Harvard versions as well?

roaldfrosig:
Thanks.
There are no plans to include english versions of the APA and Harvard
versions as of yet. But it might happen.

rmzelle:
Okay. Thanks again for contributing!

PR #1489, opened 2015-03-30

roaldfrosig:
Added more material types, and did numerous small corrections.


rmzelle:
Thanks!

PR #1490, opened 2015-03-31
Create university-college-lillebaelt-vancouver-english.csl

roaldfrosig:
English version of university-college-lillebaelt-vancouver, I have made it as a dependent style as suggested by Rintze: https://forums.zotero.org/discussion/47839/supporting-two-languages/#Item_2

### DIFF
--- a/dependent/university-college-lillebaelt-vancouver-en.csl
+++ b/dependent/university-college-lillebaelt-vancouver-en.csl
@@ -2,8 +2,9 @@
 <style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" class="in-text" default-locale="en-US">
   <info>
     <title>University College Lillebælt - Vancouver (English)</title>
-    <id>http://www.zotero.org/styles/university-college-lillebaelt-vancouver-english</id>
+    <id>http://www.zotero.org/styles/university-college-lillebaelt-vancouver-en</id>
     <link href="http://www.zotero.org/styles/university-college-lillebaelt-vancouver" rel="independent-parent"/>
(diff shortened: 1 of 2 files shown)


rmzelle:
Dependent styles need to be placed in the "dependent" subdirectory. I also prefer the file name "university-college-lillebaelt-vancouver-en.csl" (we usually just use the short-hand language code), with the style ID to match. Also add a "self" link, please.

You should be able to make all these changes by updating this pull request. Let us know if you need instructions.

zuphilip:
(@rmzelle A CSL style is still valid without the self link, why?)

rmzelle:
@zuphilip, the CSL specification doesn't require any links, except for "independent-parent" for dependent styles. It's also not really crucial for dependent styles, except for auto-updating.

We do have a stricter version of the CSL schema for the repository (https://github.com/citation-style-language/schema/blob/master/csl-repository.rnc), but it only works if the `<info/>` section is sorted correctly due to some limitations in RELAX NG. It's probably too much to ask from people submitting styles to validate against that. I run it manually once in a while.

rmzelle:
@roaldfrosig, will you set the other style to German once I merge this? ("de-DE")

roaldfrosig:
Thanks for the help rmzelle - I really appreciate it.
But I'm a bit confused about your last request: The independent style is already set to "da-DK", why should I change that to german?

rmzelle:
Uhm, never mind, then :). There is another pull request for a style that is also multilingual, and I was mixing things up.

rmzelle:
Looks good, thanks!

roaldfrosig:
Thanks!

PR #1493, opened 2015-04-01
reviews-of-modern-physics-with-titles fixes

serpodrick:
Miscellaneous changes for reviews-of-modern-physics-with-titles, as described in the commit messages. These changes are not explicitly defined in the style guide, but I can provide examples if needed.


rmzelle:
Thanks!

PR #1497, opened 2015-04-04
Update padagogische-hochschule-vorarlberg.csl

peboeck:
next try - thank you for your help!

### DIFF
--- a/padagogische-hochschule-vorarlberg.csl
+++ b/padagogische-hochschule-vorarlberg.csl
@@ -40,7 +40,7 @@
   </macro>
   <macro name="editor">
     <names variable="editor" delimiter="; ">
-      <name name-as-sort-order="all" sort-separator=", " delimiter="; " delimiter-precedes-last="always" form="long" suffix=""/>
+      <name name-as-sort-order="all" sort-separator=", " delimiter="; " delimiter-precedes-last="always" form="long" suffix="" prefix=".&#160;"/>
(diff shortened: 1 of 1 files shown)


rmzelle:
Did you try my suggestion from #1473? You seemed to have used the "publisher" macro as I wrote it, but you didn't use

```
      <group delimiter=" "/>
        <group delimiter=". "/>
          <text macro="publisher"/>
          <text macro="editor"/>
        </group>
        <text macro="published-date"/>
      </group>
```

rmzelle:
It's also much easier for us to keep track of things if you update your existing pull request, instead of creating a new one each time you wish to submit a change.

E.g. you can update this pull request by using the pencil icon at https://github.com/peboeck/styles/blob/patch-11/padagogische-hochschule-vorarlberg.csl

rmzelle:
And the point of

```
      <group delimiter=" "/>
        <group delimiter=". "/>
          <text macro="publisher"/>
          <text macro="editor"/>
        </group>
        <text macro="published-date"/>
      </group>
```

is to use ". " between the output of the "publisher" and "editor" macros, and " " between the output of those two macros and the "published-date" macro.

peboeck:
it works that way, I requested.

rmzelle:
> it works that way, I requested.

I'll just merge this, then. (also, if you're not totally comfortable writing in English, German is perfectly fine too)

PR #1499, opened 2015-04-06
Short titles for legislation

fbennett:
Related to https://forums.zotero.org/discussion/48076/shorten-subsequent-citations/#Item_1

### DIFF
--- a/oscola.csl
+++ b/oscola.csl
@@ -207,7 +207,14 @@
         <text variable="title" font-style="italic" text-case="title" form="short"/>
       </if>
       <else-if type="bill legislation" match="any">
-        <text variable="title"/>
+        <choose>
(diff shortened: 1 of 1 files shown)


oscola.csl
```diff
@@ -207,7 +207,14 @@
         <text variable="title" font-style="italic" text-case="title" form="short"/>
       </if>
       <else-if type="bill legislation" match="any">
-        <text variable="title"/>
+        <choose>
+          <if variable="title-short">
```
adam3smith: any reason you're not using the simpler
```<text variable="title" form="short"/>```
here?

fbennett: Habit from CSL-m coding. form="short" has a different meaning for legal types there.

PR #1500, opened 2015-04-06
Fix for items with both title and reviewed-title

nickbart1980:
Adapted from the following example for the notes style (CMoS, 16e, 14.215 – there seem to be no such examples in ch. 15):
2. David Kamp, “Deconstructing Dinner,” review of *The Omnivore’s Dilemma: A Natural History of Four Meals*, by Michael Pollan, *New York Times*, April 23, 2006, Sunday Book Review, http://www.nytimes.com/2006/04/23/books/review/23kamp.html.
Logic: If there's a reviewed-author and a title only, assume the title is the title of the reviewed work. If there's a reviewed-author, a title, and a reviewed-title, assume the title is the title of the review, and reviewed-title is the title of the reviewed work.
This would of course work better if Zotero had a proper review item type, and if CSL had reviewed-editor, reviewed-director, etc., and supported entering more than one reviewed work, too.

### DIFF
--- a/chicago-author-date.csl
+++ b/chicago-author-date.csl
@@ -24,7 +24,7 @@
     <category citation-format="author-date"/>
     <category field="generic-base"/>
     <summary>The author-date variant of the Chicago style</summary>
-    <updated>2015-03-29T10:53:00+00:00</updated>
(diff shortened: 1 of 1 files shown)


adam3smith:
looks good, thanks!

PR #1501, opened 2015-04-07
Add support for reviews

nickbart1980:
See APA, 6e, 7.06. I have used "Review of" rather than APA's "Review of the book"/"… DVD"/"… video game" since currently we have no reliable way of telling which it is.


adam3smith:
looks good, thanks!

PR #1502, opened 2015-04-07
Chroniques efr

NMonteix:


rmzelle:
@NMonteix, did you do this to get a non-breaking space? (for `prefix="§&#160;"`)

rmzelle:
(if you're only concerned about the symbol, it's probably better to redefine `<term name="paragraph" form="symbol">` ; see https://github.com/citation-style-language/locales/blob/master/locales-en-US.xml#L196 )

NMonteix:
Yes, in fact that was just about the symbol. Changes made. 
It seems that in French there is no proper symbol for sections, while § would be for paragraphs. Might need change in locales?

rmzelle:
The French CSL locale file already uses § for paragraphs (see https://github.com/citation-style-language/locales/blob/master/locales-fr-FR.xml#L201), so there should be no need to redefine the term in the style.

NMonteix:
Ok
Sorry, didn't check
Redefinition removed...

NMonteix:
As for me, no more corrections needed

adam3smith:
thanks!

NMonteix:
To you!

PR #1503, opened 2015-04-08
Associacao-Brasileira-de-Normas-Tecnicas-FMVZ-USP.csl 

saufmvz:
...usp.csl

- Inline: 
: citation: chronological order
: et-al: above 4 authors, mention all authors

- Bibliography: 
: chapter: added page number after year; (eds.) was changed to (ed.)
: thesis: added year after title; added page number after year;
: paper-conference: added proceeding issue; added publication year and locator-publisher; added issue year and publication year;

### DIFF
--- a/fmvz-usp.csl
+++ b/fmvz-usp.csl
@@ -1,50 +1,27 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" class="in-text" demote-non-dropping-particle="never" default-locale="pt-BR">
-  <!--Em caso de sobrenome composto, este nao sera desmontado para a montagem da citacao ou bibliografia como no caso de L. Van Der Pijl-->
   <info>
-    <title>Associação Brasileira de Normas Técnicas (Portuguese - Brazil)</title>
-    <title-short>ABNT</title-short>
-    <id>http://www.zotero.org/styles/associacao-brasileira-de-normas-tecnicas</id>
-    <link href="http://www.zotero.org/styles/associacao-brasileira-de-normas-tecnicas" rel="self"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
There's a couple of problems here, but let's start with what you actually want to do:
With this, you're actually replacing the existing ABNT style with a different style. We're certainly not going to accept that, but I suspect it's also not what you're actually trying to do.

AlbertoKawai:
I'm helping to write this style... in reality, we are trying to adapt to requested style from FMVZ-USP, a college from University of São Paulo - Brazil... it will be a new style, we don't want to replace the existing style.

rmzelle:
See https://github.com/citation-style-language/styles/blob/master/CONTRIBUTING.md for instructions on how to submit a new style.

saufmvz:
Dear sirs, we were submitted another style (#1). Please, disregard 1503. Thank you.

rmzelle:
Are you trying to submit https://github.com/saufmvz/fmvz-usp/commit/9ccfdf2fae011f32e6acd4b49023c3f96ad230e7 ?

AlbertoKawai:
Yes! We had verified the code in the indicated scripts before that too.

saufmvz:
Yes, please.  If all are correct, we want to submit. Thank you.

rmzelle:
Can you visit https://github.com/citation-style-language/styles, and then click the "+" icon, and then add your new style?
(at the right here: 
image )

The first two lines of the style at https://github.com/saufmvz/fmvz-usp/commit/9ccfdf2fae011f32e6acd4b49023c3f96ad230e7 should also be removed.

saufmvz:
Yes, thanks

rmzelle:
Continued in #1510.

PR #1504, opened 2015-04-09
Update australian-guide-to-legal-citation.csl

tabaxter:
Small update to allow definition of short titles for legislation, bills and cases in the first instance that they are used and use of the short title in subsequent refs in compliance with rule 2.1.14 in AGLC3.

For example:
(1) Conference of the Parties, UNFCCC, Decision 2/CP.15, UN Doc FCCC/CP/2009/11/Add.1 (30 March 2010) (‘<i>Copenhagen Accord</i>’).
(2) <i>Mabo v Queensland</i> [No 2] (1992) 175 CLR 1 (‘<i>Mabo [No 2]</i>’).
(3) <i>Copenhagen Accord</i>.
(4) <i>Mabo [No 2]</i> (1992) 175 CLR 1.

### DIFF
--- a/australian-guide-to-legal-citation.csl
+++ b/australian-guide-to-legal-citation.csl
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
-<!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
     <title>Australian Guide to Legal Citation</title>
     <title-short>AGLC</title-short>
@@ -29,7 +28,7 @@
     <category citation-format="note"/>
     <category field="law"/>
     <summary>A modification of the Bluebook legal citation style for Australian conditions.</summary>
-    <updated>2015-04-09T06:46:12+00:00</updated>
+    <updated>2015-04-15T02:43:49+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale>
@@ -37,7 +36,9 @@
       <term name="et-al">et al</term>
     </terms>
   </locale>
+  <!--Authors and Persons-->
   <macro name="author-note">
+    <!--for bills & hearing this should start with jurisdiction once available-->
     <choose>
       <if type="interview">
         <group delimiter=", ">
@@ -120,6 +121,7 @@
     </choose>
   </macro>
   <macro name="author">
+    <!--for bills & hearing this should start with jurisdiction once available-->
     <choose>
       <if type="interview">
         <group delimiter=", ">
@@ -174,6 +176,7 @@
       </names>
     </group>
   </macro>
+  <!-- Titles -->
   <macro name="title">
     <choose>
       <if type="book legislation webpage thesis motion_picture manuscript" match="any">
@@ -183,7 +186,7 @@
         <text variable="title" text-case="title"/>
       </else-if>
       <else-if type="legal_case">
-        <text variable="title" strip-periods="true" font-style="italic"/>
+        <text variable="title" font-style="italic" strip-periods="true" />
       </else-if>
       <else>
         <text variable="title" quotes="true" text-case="title"/>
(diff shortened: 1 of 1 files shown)


australian-guide-to-legal-citation.csl
```diff
@@ -199,11 +196,10 @@
         <text variable="title" font-style="italic" text-case="title" form="short"/>
       </if>
       <else>
-        <text variable="title" quotes="true" text-case="title" form="short"/>
+        <text variable="title" form="short" text-case="title" quotes="false"/>
```
adam3smith: I don't think that's right. E.g. short forms of article titles should still be in quotes. You'll need a separate else-if here.

adam3smith:
Also, as you see, this strips out all of the commenting because of the visual editor. That's not good. You'll have to make the changes manually.

australian-guide-to-legal-citation.csl
tabaxter: This is definitely how the style does it: http://www.law.unimelb.edu.au/files/dmfile/FinalOnlinePDF-2012Reprint.pdf#page=73 (rule 2.1.14)

You're absolutely right on the comments though. I'll fix that up when I get to work. Sorry. That was a dumb oversight.

Tim

adam3smith: no, you're misunderstanding my point: The line you're changing also affects subsequent citations for journal articles (and other item types) -- those should still be in quotes. (See the note 62 example in 1.4.2). You need to add a separate else-if for cases (and whichever else item type this applies to).

tabaxter: Yep. You're right. I broke it. Sorry. Very experienced with AGLC, not so experienced with coding. I think I've got it now, with comments, but I'll run a few documents through it tomorrow to check i haven't broken anything else before I resubmit.

rmzelle:
You can update the current pull request via the pencil icon at https://github.com/tabaxter/styles/blob/patch-1/australian-guide-to-legal-citation.csl

tabaxter:
Just to check in: Was editing the current pull request the best way to resubmit, or should I do a separate one?

T

rmzelle:
I'll let @adam3smith review this, but it looks good now. Sorry for not getting back to you, but we don't get email updates for commits, just for comments.

tabaxter:
All good. I didn't want to hassle. It's not a rush. Just wanted to make sure I'd done the right thing.

australian-guide-to-legal-citation.csl
```diff
@@ -195,11 +195,11 @@
...
   <macro name="title-short">
     <choose>
-      <if type="book legislation webpage thesis motion_picture manuscript" match="any">
-        <text variable="title" font-style="italic" text-case="title" form="short"/>
+      <if type="book legislation webpage thesis motion_picture manuscript legal_case article" match="any">
+        <text variable="title-short" font-style="italic" text-case="title" form="short"/>
       </if>
       <else>
-        <text variable="title" quotes="true" text-case="title" form="short"/>
+        <text variable="title" quotes="false" text-case="title" form="short"/>
```
adam3smith: but you're still doing quotes="false" here -- there should be quotes here, e.g. for journal articles, or am I misunderstanding something?

australian-guide-to-legal-citation.csl
```diff
@@ -195,11 +195,11 @@
   </macro>
   <macro name="title-short">
     <choose>
-      <if type="book legislation webpage thesis motion_picture manuscript" match="any">
-        <text variable="title" font-style="italic" text-case="title" form="short"/>
+      <if type="book legislation webpage thesis motion_picture manuscript legal_case article" match="any">
+        <text variable="title-short" font-style="italic" text-case="title" form="short"/>
```
adam3smith: remove the ```form="short"``` here since you're using title-short anyway.

adam3smith: also -- what are you intending ```article``` for here? From Zotero that's never used, but maybe you have a different application in mind?

adam3smith:
Yes, you did everything exactly right; we just ask that you leave a quick note when you're done since, as rmzelle says, we don't get a notification. See my requests and questions in line.

australian-guide-to-legal-citation.csl
tabaxter: Done. And nup. I'm just not careful enough.

Thanks for being much sharper than I am.

australian-guide-to-legal-citation.csl
tabaxter: As above. I'm not as careful checking my work as I should be. Sorry. 

I know it wastes your time if I don't. I'm not sure how I missed so much.

zuphilip:
Because there are no notifications for commits, I will let you know that a new version by @tabaxter is provided.

tabaxter:
Thanks @zuphilip, you beat me to it.

adam3smith:
Great, thanks, that should do it. (And no worries about missing some spots. It's a complex style and I've seen hundreds of these, so it'd be rather depressing if I didn't have some type of edge ;). Thanks for contributing!

PR #1505, opened 2015-04-09
Update national-science-foundation-grant-proposals.csl

dmkaplan2000:
Modifications to collapse numeric citations from [1,2,3,4] to [1-4] and to avoid a trailing comma after DOIs.

The format still deals very poorly with non-journal references (e.g., books) and is quite strange in terms of the placement of the year between journal volume and page numbers.

### DIFF
--- a/national-science-foundation-grant-proposals.csl
+++ b/national-science-foundation-grant-proposals.csl
@@ -15,7 +15,7 @@
     <updated>2012-09-14T21:22:32+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
-  <citation>
+  <citation collapse="citation-number">
     <sort>
       <key variable="citation-number"/>
     </sort>
@@ -37,8 +37,10 @@
         <date-part name="year" form="long"/>
       </date>
       <text variable="page" suffix=". "/>
-      <text variable="DOI" prefix="doi:" suffix=", "/>
-      <text variable="URL" prefix="Available at " suffix=" "/>
+      <group delimeter=", ">
+	      <text variable="DOI" prefix="doi:"/>
+	      <text variable="URL" prefix="Available at " suffix=" "/>
+      </group>
     </layout>
   </bibliography>
 </style>


national-science-foundation-grant-proposals.csl
```diff
@@ -37,8 +37,10 @@
         <date-part name="year" form="long"/>
       </date>
       <text variable="page" suffix=". "/>
-      <text variable="DOI" prefix="doi:" suffix=", "/>
-      <text variable="URL" prefix="Available at " suffix=" "/>
+      <group delimeter=", ">
```
rmzelle: @adam3smith, uhm...

adam3smith:
since NSF doesn't provide any reference format, I don't have any opinion on how these should look more generally (though clearly your two changes make sense). @rmzelle - the original commit is by you. Do you recall anything about this?

rmzelle:
It's from #216.

adam3smith:
ugh - I was sure this was greenlighted by Travis.

rmzelle:
No, I couldn't find this in Travis (https://travis-ci.org/citation-style-language/styles/pull_requests). Not sure why it skipped this pull request.

adam3smith:
@dmkaplan2000  sorry, I had to revert this since you misspelled "delimiter" - can you make another pull request with that bit corrected (and while you're at it - please indent with 2spaces, not with tab).

dmkaplan2000:
Will do.  I was trying to do this using the online interface, but that 
is a nightmare.  Will try using standard git fork.

Cheers,
David

dmkaplan2000:
By the way, I have one question.  When a reference has both a DOI and a 
URL (as many of my journal references have), zotero seems to magically 
know to only show the DOI.  How does this come about?

Thanks,
David

adam3smith:
that's not actually what's going on. Zotero by default treats URL=false when an article has a page range. You can toggle that behavior in the Cite/styles tab of the preferences. (Obviously the behavior you describe could be coded into the style and, in fact, that's done for a fair number of existing styles)

dmkaplan2000:
Yes, I have always done it in the style, which is why it surprised me 
when it just seemed to work.  At the moment, I am just doing minimal 
corrections to this style, but maybe next time I submit to NSF I will 
address these details and non-journal refs.

By the way, I created another pull request and then I realized I forgot 
to replace tabs with spaces.  I committed a new change to deal with 
this, but I am not sure how pull requests work and whether or not I need 
to create another one to deal with these latest changes...

Thanks,
David

PR #1508, opened 2015-04-12
Fixed formatting error in Chest style.

TimotheeJeannin:
With a friend in medicine field writing his thesis, we spotted a small problem in the Chest style. This journal requires italic titles for journal titles. 

As we can read here : http://journal.publications.chestnet.org/ss/forauthors.aspx#References

    Publication source (italicized), when referring to a journal, the journal name should be abbreviated according to Index Medicus


rmzelle:
Thanks!

PR #1512, opened 2015-04-13
Update vietnam-ministry-of-education-and-training-en.csl

Tungphamfsh:
Fix citation numbering error and remove abstract section from bibliography


rmzelle:
Thanks!

PR #1514, opened 2015-04-14
Update iso690-numeric-brackets-cs.csl

tomas4:
The list of references contained in some styles first author with family-name first, and subsequent authors in opposite order. Fixed that by changing the name-as-sort-order to "all" and putting <name-part name="family" text-case="uppercase"/> before <name-part name="given"/> everywhere. So the result is the order is family-name, given-name always.

### DIFF
--- a/iso690-numeric-brackets-cs.csl
+++ b/iso690-numeric-brackets-cs.csl
@@ -29,7 +29,6 @@
     <names variable="author">
       <name and="text" name-as-sort-order="all" sort-separator=", " delimiter=", " delimiter-precedes-last="never">
         <name-part name="family" text-case="uppercase"/>
-        <name-part name="given"/>
       </name>
     </names>
   </macro>
@@ -37,7 +36,6 @@
     <names variable="editor">
       <name and="text" name-as-sort-order="all" sort-separator=", " delimiter=", " delimiter-precedes-last="never">
         <name-part name="family" text-case="uppercase"/>
-        <name-part name="given"/>
       </name>
       <label prefix=", " form="short" suffix="."/>
     </names>
@@ -46,7 +44,6 @@
     <names variable="editor">
       <name and="text" delimiter=", " delimiter-precedes-last="never">
         <name-part name="family" text-case="uppercase"/>
-        <name-part name="given"/>
       </name>
       <label prefix=", " form="short" suffix=". "/>
     </names>
@@ -55,7 +52,6 @@
     <names variable="translator">
       <name and="text" delimiter=", " delimiter-precedes-last="never">
         <name-part name="family" text-case="uppercase"/>
-        <name-part name="given"/>
       </name>
       <label prefix=" (" form="short" suffix=".)"/>
     </names>
@@ -65,7 +61,6 @@
     <names variable="translator">
       <name and="text" delimiter=", " delimiter-precedes-last="never">
         <name-part name="family" text-case="uppercase"/>
-        <name-part name="given"/>
       </name>
     </names>
   </macro>
@@ -74,7 +69,6 @@
     <names variable="illustrator">
       <name and="text" delimiter=", " delimiter-precedes-last="never">
         <name-part name="family" text-case="uppercase"/>
-        <name-part name="given"/>
       </name>
     </names>
   </macro>
@@ -115,7 +109,6 @@
     <names variable="container-author">
       <name and="text" delimiter=", " delimiter-precedes-last="never">
         <name-part name="family" text-case="uppercase"/>
-        <name-part name="given"/>
       </name>
     </names>
   </macro>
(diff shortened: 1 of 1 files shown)


iso690-numeric-brackets-cs.csl
```diff
@@ -35,9 +35,9 @@
   </macro>
   <macro name="editor">
     <names variable="editor">
-      <name and="text" name-as-sort-order="first" sort-separator=", " delimiter=", " delimiter-precedes-last="never">
-        <name-part name="given"/>
+      <name and="text" name-as-sort-order="all" sort-separator=", " delimiter=", " delimiter-precedes-last="never">
         <name-part name="family" text-case="uppercase"/>
+        <name-part name="given"/>
```
adam3smith: could we actually take all the ```<name-part name="give"/>``` out of this style? They don't actually do anything. cs:name-part is only used for styling (e.g. the uppercase used for family names here). It has no effect on either the presence of name parts or their order.

tomas4: The family name has to be upper case all, and the given name has to be upper case first letter, lower case rest. So if it stays this way (and now I see it probably will), yes it can be removed. Should I do it?
You know, I do not understand how the markup works to such details, so I just did minimal edits of the parent style. I did thought the order of these tags matters.
Thanks for your time.

Dne 14. 4. 2015 15:44 napsal uživatel Sebastian Karcher <notifications@github.com>:In iso690-numeric-brackets-cs.csl:
>          <name-part name="family" text-case="uppercase"/>
> +        <name-part name="given"/>

could we actually take all the <name-part name="give"/> out of this style? They don't actually do anything. cs:name-part is only used for styling (e.g. the uppercase used for family names here). It has no effect on either the presence of name parts or their order.

rmzelle: > So if it stays this way (and now I see it probably will), yes it can be removed. Should I do it?

Yes, please.

tomas4: Dne 14.4.2015 v 19:45 Rintze M. Zelle napsal(a):
>
> In iso690-numeric-brackets-cs.csl 
> <https://github.com/citation-style-language/styles/pull/1514#discussion_r28350950>:
>
> >          <name-part name="family" text-case="uppercase"/>
> > +        <name-part name="given"/>
>
>     So if it stays this way (and now I see it probably will), yes it
>     can be removed. Should I do it?
>
> Yes, please.
>
> —
> <https://github.com/citation-style-language/styles/pull/1514/files#r28350950>.
>
Done.

-- 
Tomáš Brunclík
fotky: https://picasaweb.google.com/Tomas.Brunclik
mobil: (+420) 724 709 138
skype: tomas_brunclik, jabber: tomas_iv@jabber.cz

tomas4: Oh, wait a minute, it was the -en version..

tomas4: Done in the -en, now also in the -cs?

libora:
Were this changes consulted with Mr Kratochvíl (kratec@ukb.muni.cz) from Library of Masaryk University who is responsible for the correct behaviour of ČSN ISO 690:2011 style? ČSN ISO 690:2011 (and ISO 690:2010 too)  does not describe exact forms of names, but we need the same behaviour of all ČSN ISO 690:2011 styles (author-date, brackets, parentheses, note). If there are different requirements from publisher or university of tomas4, then I suggest creating new independent style for this publisher/university.

kratec:
As **libora **wrote the proposed amendment is inconsistent with ISO 690. ISO 690 orders to write first author with surname first and subsequent authors with surnames second. 

Correct form: SMITH, John, Michael NEWMAN, Richard HUGHES, and Paul HOPKINS.
Incorrect form: SMITH John, NEWMAN Michael,  HUGHES Richard, and HOPKINS Paul.

adam3smith:
Thanks @libora , @kratec for checking. Not taking this PR then. @tomas4 -- looked over the style, and this isn't actually inconsistent. The first-listed creator always has the last name first, all other authors are listed in their "natural" order.

tomas4:
Hello guys,
I created the style by modification of an existing one few years ago and 
yes, this variant was primarily created to satisfy local customs at the 
University of Pardubice and be still compatible with ISO 690. It was 
shared just in the hopes it can be useful where square brackets are used 
and I did not do any modifications to it since. I was also novice to 
editing the style definition, so maybe it contains unnecessary tags or 
plain mistakes. Just do with it what you feel is right.
Tomas B.


Dne 25.4.2016 v 18:20 Sebastian Karcher napsal(a):
>
> Thanks @libora <https://github.com/libora> , @kratec 
> <https://github.com/kratec> for checking. Not taking this PR then. 
> @tomas4 <https://github.com/tomas4> -- looked over the style, and this 
> isn't actually inconsistent. The first-listed creator always has the 
> last name first, all other authors are listed in their "natural" order.
>
> —
> <https://github.com/citation-style-language/styles/pull/1514#issuecomment-214426906>
>


-- 
Tomáš Brunclík
fotky: https://picasaweb.google.com/Tomas.Brunclik
mobil: (+420) 724 709 138
skype: tomas_brunclik, jabber: tomas_iv@jabber.cz

adam3smith:
Closing based on above feedback. Thanks everyone!

PR #1515, opened 2015-04-14
Create iso690-numeric-brackets-en.csl

tomas4:
It is just the iso690-numeric-brackets-cs.csl with the #1514 fix applied, Czech language additions removed and terms tag copied over from iso690-numeric-english.csl. Reason: there was square brackets variant of iso690 missing in English.

### DIFF
--- a/iso690-numeric-brackets-en.csl
+++ b/iso690-numeric-brackets-en.csl
@@ -30,7 +30,6 @@
     <names variable="author">
       <name and="text" name-as-sort-order="all" sort-separator=", " delimiter=", " delimiter-precedes-last="never">
         <name-part name="family" text-case="uppercase"/>
-        <name-part name="given"/>
       </name>
     </names>
   </macro>
@@ -38,7 +37,6 @@
     <names variable="editor">
       <name and="text" name-as-sort-order="all" sort-separator=", " delimiter=", " delimiter-precedes-last="never">
         <name-part name="family" text-case="uppercase"/>
-        <name-part name="given"/>
       </name>
       <label prefix=", " form="short" suffix="."/>
     </names>
@@ -47,7 +45,6 @@
     <names variable="editor">
       <name and="text" delimiter=", " delimiter-precedes-last="never">
         <name-part name="family" text-case="uppercase"/>
-        <name-part name="given"/>
       </name>
       <label prefix=", " form="short" suffix=". "/>
     </names>
@@ -56,7 +53,6 @@
     <names variable="translator">
       <name and="text" delimiter=", " delimiter-precedes-last="never">
         <name-part name="family" text-case="uppercase"/>
-        <name-part name="given"/>
       </name>
       <label prefix=" (" form="short" suffix=".)"/>
     </names>
@@ -66,7 +62,6 @@
     <names variable="translator">
       <name and="text" delimiter=", " delimiter-precedes-last="never">
         <name-part name="family" text-case="uppercase"/>
-        <name-part name="given"/>
       </name>
     </names>
   </macro>
@@ -75,7 +70,6 @@
     <names variable="illustrator">
       <name and="text" delimiter=", " delimiter-precedes-last="never">
         <name-part name="family" text-case="uppercase"/>
-        <name-part name="given"/>
       </name>
     </names>
   </macro>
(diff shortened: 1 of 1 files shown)


adam3smith:
wouldn't it make more sense to use the numeric-en version as a template? It seems like the parentheses and brackets version of the style in the same language should otherwise be identical.

And there's really a need for this, yes? I've gone back and did some research, and ISO-690 does not actually implement a citation style:

> ISO 690:2010 does not prescribe a particular style of reference or citation. The examples used in ISO 690:2010 are not prescriptive as to style and punctuation.

In particular, several standards institutes (including the British BSI and the German DIN) have implemented their own, national norms based on ISO-690, which should probably be listed separately. So particularly in English, I'm not quite sure what rules this would be following?

tomas4:
The square brackets are often used to be distinguished from formula numbers, traditionally numbered within parentheses in academic and research works, and referenced this way from the text. It is therefore required (or upper index,but not parentheses) at our University in combination with iso690 style, which is recommended for theses. And it is still compliant with the general iso690 rules AFAIK. We are Czech Uni, but of course we need to write both Czech and English. I am also sure similar rules or customs would  apply in many research organizations in different countries.
And why it was not based on general numeric-en is:
- it seems to me the list of sources looks more tidy with the text indented from the numbers.
- it is really handy to have visually identical version of czech and english style complying exactly with rules andeven customs at our Uni (yes this is personal reason)..

  

Dne 14. 4. 2015 15:55 napsal uživatel Sebastian Karcher <notifications@github.com>:

rmzelle:
@adam3smith, shall we make a decision here? See https://www.zotero.org/styles?q=iso-690 for our current ISO-690 collection. I'm thinking to decline, since these additional variants tend to boil down to personal preference in the absence of official guidance. If the university prescribed something specific, we would accept that, but

> It is therefore required (or upper index,but not parentheses) at our University in combination with iso690 style, which is recommended for theses

is just too vague for us. Agreed?

adam3smith:
not sure -- the ISO690 styles seem very popular in the Czech Republic, so if this is an English language version of the style that'd get use across universities there, it certainly seems worthwhile.

 @libora -- could you help us out here maybe? Would this be very commonly used and understood (and ideally, is it documented somewhere?), or is this particular to one specific university?

libora:
@adam3smith Yes you right, ČSN ISO 690:2011 is Czech national standard (it is the translation of English ISO 690:2010) and this style is required or recommended for Universities, for a lot of journals etc. On another site, I have to say, that ISO 690 is extremely problematic for implementation. This standard does not describe the exact form of a quote. This standard describes elements of the quote and the order of elements. But the delimiters are not prescribed, you can use volume and issue in short or in long form,  formatting attributes are not prescribed, form and sorting authors' names are not prescribed etc. Exist two main rules:
1. elements of quote (and their order) for each type of cited document
2. the formatting of quote must be the same in the whole document 
Usually, the Universities requires usage of ISO 690 but also do not describe the exact form of quotation. Each University (and sometimes each Faculty) has own interpretation of ISO 690, but in these documents, you can find different forms of quote very often. Here is the list of interpretation of ISO 690 (only selection of my preferred sources):
Masaryk University in Brno (author is Mr Kratochvíl - coauthor of Czech iso templates): http://is.muni.cz/do/rect/el/estud/prif/ps11/metodika/web/ebook_citace_2011.html#
University in Plzeň: https://sites.google.com/site/novaiso690/
Guidelines on citace.com portal: http://www.citace.com/download/CSN-ISO-690.pdf
Universitas Carolina: http://www1.cuni.cz/~brt/bibref/bibref.html
I think you have English ISO690:2010 available, if not, you can find it on the Internet (http://www.rtu.lv/writable/public_files/RTU_standarts_iso_6902010_english.pdf)

rmzelle:
> the ISO690 styles seem very popular in the Czech Republic, so if this is an English language version of the style that'd get use across universities there, it certainly seems worthwhile.

The whole naming scheme is still vague and problematic, though. E.g. say we create "iso690-numeric-brackets-en.csl" as a sibling for "iso690-numeric-brackets-cs.csl", and then somebody requests an English version of "iso690-numeric-lt" (also uses brackets). Would we then keep two "iso690-numeric-brackets-en.csl" versions? It will just become a mess.

Since ISO690 is so unspecific, I really want to prevent accumulating dozens of variations with no clear home. If a university requires a specific variant of ISO690, they should describe that variant in a style guide, and then we'll create university-specific styles for those.

adam3smith:
OK, based on Rintze's last post, closing this. Thanks everyone!

PR #1516, opened 2015-04-14
Revert "Update journal-of-chromatography-a.csl"

cpina:
This reverts commit dcd4c2a1c38521a511393a7add9c9e5611b6badf.

So now it matches the internal Elsevier guide and http://www.elsevier.com/journals/journal-of-chromatography-a/0021-9673/guide-for-authors

I suspect that when dcd4c2a1c38521a511393a7add9c9e5611b6badf was
introduced the public Guide for Authors was not updated yet.


adam3smith:
thanks!

PR #1517, opened 2015-04-14
Update peerj.csl

JackieThai:
Edited to include DOIs for all journal references.


adam3smith:
there's not actually an access macro. You'd have to add one that works for PeerJ, format-wise.

PR #1510, opened 2015-04-13
Create Associação Brasileira de Normas Técnicas - Faculdade de Medicina ...

saufmvz:
...Veterinária e Zootecnia - USP (Portuguese - Brazil)

PR #1519, opened 2015-04-15
associacao-brasileira-de-normas-tecnicas-usp-fmvz.csl

saufmvz:


rmzelle:
You used "fmvz-usp.csl" as the file name here, which isn't right either.

rmzelle:
Superseded by #1510.

PR #1520, opened 2015-04-15
Create harvard-fh-salzburg.csl

mmackh:
Citation for Fachhochschule Salzburg

### DIFF
--- a/harvard-fachhochschule-salzburg.csl
+++ b/harvard-fachhochschule-salzburg.csl
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" class="in-text" demote-non-dropping-particle="sort-only" default-locale="en-GB" page-range-format="expanded">
   <info>
-    <title>Harvard - FH Salzburg</title>
-    <id>http://www.zotero.org/styles/harvard-fh-salzburg</id>
+    <title>Harvard - Fachhochschule Salzburg</title>
+    <id>http://www.zotero.org/styles/harvard-fachhochschule-salzburg</id>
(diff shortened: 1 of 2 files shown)


rmzelle:
Is this style specific to Informationstechnik und Systemmanagement?

If so, in general, we would use "Harvard - Fachhochschule Salzburg - Informationstechnik und Systemmanagement" for the style title and "harvard-fachhochschule-salzburg-informationstechnik-und-systemmanagement.csl" for the file name (no abbreviations). If your university uses multiple style formats (e.g. both an author-date and a numeric style), we might want to use "Fachhochschule Salzburg - Informationstechnik und Systemmanagement (Harvard)" instead (and something similar for the file name), but that might not be the case here.

Which CSL did you base this on? Can you add a "template" link? (see point 8 at https://github.com/citation-style-language/styles/wiki/Style-Requirements )

The `<category field="medicine"/>` also needs to be updated. For a "Informationstechnik und Systemmanagement"-specific style, "communications" and "engineering" are probably reasonable choices (see http://docs.citationstyles.org/en/stable/specification.html#appendix-i-categories).

mmackh:
Hi, thank you for writing back so quickly - sorry it took a little time - exam season. FH Salzburg does not have any departments, so the style applies universally to every assignment, even from different subjects.

rmzelle:
Can you rename the style to ""Harvard - Fachhochschule Salzburg", then? Please use "harvard-fachhochschule-salzburg.csl" as the style file name, and update the style ID and "self" link accordingly.

"tourism" is not a valid CSL category, but maybe "science" works.

mmackh:
Ok, I've made the changes accordingly - please let me know if anything else needs to be changed?

rmzelle:
Looks good, thanks!

PR #1521, opened 2015-04-16
Add reviewed-title to american-chemical-society-* and reviews-of-modern-physics-with-titles

serpodrick:
Hi

This pull request adds reviewed-title to the american-chemical-society-* and reviews-of-modern-physics-with-titles styles.

Both style guides are entirely silent about about these kinds of publications.

(As an aside, the review and review-book types are confusing. I haven't figured out what review-book is supposed to mean!)

### DIFF
--- a/american-chemical-society-author-date.csl
+++ b/american-chemical-society-author-date.csl
@@ -82,14 +82,16 @@
     </group>
   </macro>
   <macro name="title">
-    <choose>
-      <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
-        <text variable="title" text-case="title" font-style="italic"/>
-      </if>
-      <else>
-        <text variable="title" text-case="title"/>
-      </else>
-    </choose>
+    <group delimiter=". ">
+      <choose>
+        <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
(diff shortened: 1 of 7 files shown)


adam3smith:
Not sure about this -- as it is, this would seem to add confusion if anything: It may look like the review title is part of the title or (in absence of a title) actually the title of the work. How are you envisioning this being used?

adam3smith:
(Travis fails because you're misspelling delimiter, btw.)

serpodrick:
I'm trying to add the reviewed-title after the title like council-of-science-editors does. The CSE style uses the type (review or review-book) to decide whether or not to display reviewed. I was trying to avoid that since if reviewed-title has a value, it's a review. (<ost of the things I'm trying to create citations for are actually journal articles.)  Adding it to the title macro was easier than adding a review macro.

If following the model in council-of-science-editors would be accepted, I can submit a new pull request. (I only use american-chemical-society-with-titles version of ACS and forgot to test the others, sorry!)

adam3smith:
What I'm wondering isn't how the style is going to tell this is a review (that makes sense) but how the reader would be able to tell from a citation that it is.

serpodrick:
I agree. I'll use a non-CSL solution for what I'm trying to accomplish.

PR #1524, opened 2015-04-17

zuphilip:
Error reported in https://forums.zotero.org/discussion/48393 also I agree with @adam3smith that the example given there is not ideal. Please look carefully at the change. Is it in general an improvement? Should I look at the other Springer variants as well?

### DIFF
--- a/springer-basic-author-date.csl
+++ b/springer-basic-author-date.csl
@@ -48,14 +48,6 @@
       </substitute>
     </names>
   </macro>
-  <macro name="editor">
-    <text term="in" text-case="capitalize-first" suffix=": "/>
-    <names variable="editor">
-      <name sort-separator=" " initialize-with="" name-as-sort-order="all" delimiter=", " delimiter-precedes-last="always"/>
-      <label form="short" strip-periods="true" prefix=" (" suffix=")"/>
-      <et-al term="and others"/>
-    </names>
-  </macro>
   <macro name="edition">
     <choose>
       <if is-numeric="edition">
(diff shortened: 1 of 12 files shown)


rmzelle:
@adam3smith, will you review this?

springer-basic-author-date.csl
```diff
@@ -121,10 +113,16 @@
...
-            <text variable="container-title"/>
-            <text macro="edition"/>
+          <text term="in" text-case="capitalize-first" suffix=": "/>
+          <group delimiter=" " suffix=". ">
+            <names variable="editor">
+              <name sort-separator=" " initialize-with="" name-as-sort-order="all" delimiter=", " delimiter-precedes-last="always"/>
+              <label form="short" strip-periods="true" prefix=" (" suffix=")"/>
+            </names>
+            <group delimiter=", ">
+              <text variable="container-title" form="short"/>
```
adam3smith: I don't think we want the form="short" here -- in Zotero/Mendeley that will force journal-style abbreviations on the book title, which is almost certainly wrong (it definitely is for the example above).

springer-basic-author-date.csl
```diff
@@ -121,10 +113,16 @@
              In: Smith J (ed) The rise of modern genomics, 3rd edn.
              Wiley, New York, pp 230-257 -->
         <if type="chapter paper-conference" match="any">
-          <text macro="editor"/>
-          <group prefix=" " delimiter=", " suffix=". ">
-            <text variable="container-title"/>
-            <text macro="edition"/>
+          <text term="in" text-case="capitalize-first" suffix=": "/>
```
adam3smith: It likely doesn't make a difference, but let's move the ```in``` term into the group (and them remove the space in the suffix, because you now have a delimiter). In the very unlikely case that we have neither a container title nor an editor, we likely don't want the "in", either.

adam3smith:
sorry for the delay, see comments inline. Yes, looking at the other Springer styles for chapters w/o editors would definitely be helpful.

zuphilip:
Please have a look at the new version. I add all other Springer styles which look similar or equally easy to adapt.

PR #1526, opened 2015-04-18
Update Springer styles

rmzelle:
Using the information from https://docs.google.com/spreadsheets/d/1O8IhzOk9tHV7jCbLFeD_9mQru69_EBUMm0mfpGCyGew/edit?usp=sharing

This is a spreadsheet prepared by @zuphilip (see also https://github.com/citation-style-language/utilities/issues/17). I added the "Results2" tab, which pulls out the metadata details required to generate the dependent styles. I also added language mapping to the "mapping-styles" tab.

Travis is rather unhappy, and there are still a bunch things to fix:

* some styles exist both as an independent and dependent style (like "journal-of-chemical-ecology").
* some styles incorrectly use the same value for ISSN and eISSN (like "biological-procedures-online"), which probably is a problem that originates with the Springer database (and should probably be fixed there).
* some styles are deleted, so those should be verified (e.g. to make sure a journal has been discontinued or moved to another publisher), and added to renamed-styles.json
* it looks like the new Springer list includes many if not all the BMC journals, which we have been generating separately previously.
* this metadata set also seems to include the styles from "springer-fachzeitschriften-medizin".
* we probably append the language to the title if it is not English.


dependent/bundesgesundheitsblatt-gesundheitsforschung-gesundheitsschutz.csl
```diff
@@ -1,15 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" default-locale="de-DE">
-  <!-- Generated with https://github.com/citation-style-language/utilities/tree/master/generate_dependent_styles/data/springer-fachzeitschriften-medizin -->
+  <!-- Generated with https://github.com/citation-style-language/utilities/tree/master/generate_dependent_styles/data/springer -->
   <info>
-    <title>Bundesgesundheitsblatt - Gesundheitsforschung - Gesundheitsschutz (German)</title>
-    <title-short>Bundesgesundheitsblatt Gesundheitsforschung Gesundheitsschutz</title-short>
+    <title>Bundesgesundheitsblatt - Gesundheitsforschung - Gesundheitsschutz</title>
```
zuphilip: The "(German)" as part otf the title is missing. And the title-short is missing.

zuphilip: title-short information is missing in Table2 but in Table1 it is there (as complete as we have it from our previous list).

dependent/academic-psychiatry.csl
```diff
@@ -2,17 +2,15 @@
 <style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" default-locale="en-US">
   <!-- Generated with https://github.com/citation-style-language/utilities/tree/master/generate_dependent_styles/data/springer -->
   <info>
-    <title>Modern Rheumatology</title>
-    <title-short>Mod Rheumatol</title-short>
-    <id>http://www.zotero.org/styles/modern-rheumatology</id>
-    <link href="http://www.zotero.org/styles/modern-rheumatology" rel="self"/>
+    <title>Academic Psychiatry</title>
+    <id>http://www.zotero.org/styles/academic-psychiatry</id>
+    <link href="http://www.zotero.org/styles/academic-psychiatry" rel="self"/>
```
zuphilip: From where is this renaming of the file coming from?

dependent/academic-psychiatry.csl
```diff
@@ -2,17 +2,15 @@
...
     <link href="http://www.zotero.org/styles/springer-vancouver-brackets" rel="independent-parent"/>
     <link href="http://www.springer.com/cda/content/document/cda_downloaddocument/Key_Style_Points_1.0.pdf" rel="documentation"/>
     <link href="http://www.springer.com/cda/content/document/cda_downloaddocument/manuscript-guidelines-1.0.pdf" rel="documentation"/>
     <category citation-format="numeric"/>
-    <category field="medicine"/>
-    <issn>1439-7595</issn>
-    <eissn>1439-7609</eissn>
+    <issn>1042-9670</issn>
+    <eissn>1545-7230</eissn>
```
zuphilip: The title seems to be different and the issn as well. Why should it then be the same file?

dependent/acta-veterinaria-scandinavica.csl
```diff
@@ -1,18 +1,16 @@
...
+    <link href="http://www.zotero.org/styles/springer-basic-brackets" rel="independent-parent"/>
+    <link href="http://www.springer.com/cda/content/document/cda_downloaddocument/Key_Style_Points_1.0.pdf" rel="documentation"/>
+    <link href="http://www.springer.com/cda/content/document/cda_downloaddocument/manuscript-guidelines-1.0.pdf" rel="documentation"/>
     <category citation-format="numeric"/>
-    <category field="biology"/>
-    <category field="medicine"/>
-    <issn>1751-0147</issn>
-    <updated>2014-05-09T12:00:00+00:00</updated>
+    <eissn>1751-0147</eissn>
+    <updated>2014-05-15T12:00:00+00:00</updated>
```
zuphilip: What date/time is this?

dependent/addiction-science-and-clinical-practice.csl
```diff
@@ -1,18 +1,16 @@
...
     <id>http://www.zotero.org/styles/addiction-science-and-clinical-practice</id>
     <link href="http://www.zotero.org/styles/addiction-science-and-clinical-practice" rel="self"/>
-    <link href="http://www.zotero.org/styles/biomed-central" rel="independent-parent"/>
-    <link href="http://www.ascpjournal.org" rel="documentation"/>
+    <link href="http://www.zotero.org/styles/springer-basic-brackets" rel="independent-parent"/>
+    <link href="http://www.springer.com/cda/content/document/cda_downloaddocument/Key_Style_Points_1.0.pdf" rel="documentation"/>
+    <link href="http://www.springer.com/cda/content/document/cda_downloaddocument/manuscript-guidelines-1.0.pdf" rel="documentation"/>
     <category citation-format="numeric"/>
-    <category field="biology"/>
-    <category field="medicine"/>
```
zuphilip: Why are the category fields vanishing?

dependent/academic-psychiatry.csl
rmzelle: GitHub or git mistakenly thinks that we renamed modern-rheumatology.csl to academic-psychiatry.csl, while in reality we deleted modern-rheumatology.csl and created academic-psychiatry.csl. I'm not sure if there is an easy solution to this.

dependent/acta-veterinaria-scandinavica.csl
rmzelle: It's the date and time stamp from the template file: https://github.com/citation-style-language/utilities/blob/master/generate_dependent_styles/data/springer/_template.csl

dependent/addiction-science-and-clinical-practice.csl
rmzelle: Because all BMC journals are annotated with the "biology" and "medicine" fields via the BMC template: https://github.com/citation-style-language/utilities/blob/master/generate_dependent_styles/data/bmc/_template.csl

dependent/bundesgesundheitsblatt-gesundheitsforschung-gesundheitsschutz.csl
rmzelle: Do we want a short-title if it's the same as the title, though? Seems redundant.

dependent/academic-psychiatry.csl
rmzelle: Best documentation on this I've found is http://stackoverflow.com/questions/24753281/git-thinks-ive-renamed-a-file-rather-than-deleting-one-and-creating-another. So it looks like the only option would be to separate the deletion and creation of styles into separate commits, which is unwieldy.

dependent/bundesgesundheitsblatt-gesundheitsforschung-gesundheitsschutz.csl
zuphilip: I think it is also missing everywhere else. I guess we can delete the redundant short-titles but may want to keep the rest.

dependent/acta-veterinaria-scandinavica.csl
zuphilip: Are you using that template file? I think it does not handle languages different from English...

dependent/bundesgesundheitsblatt-gesundheitsforschung-gesundheitsschutz.csl
rmzelle: I updated the first column "Results2" tab to include the language in the title for non-English styles.

dependent/acta-veterinaria-scandinavica.csl
rmzelle: I'm using an updated version that incorporates the desired default-locale, but I haven't pushed it to the utilities repo yet.

dependent/bundesgesundheitsblatt-gesundheitsforschung-gesundheitsschutz.csl
rmzelle: > I think it is also missing everywhere else. I guess we can delete the redundant short-titles but may want to keep the rest.

But that's an issue with the spreadsheet, no?

dependent/acta-veterinaria-scandinavica.csl
zuphilip: Okay. And which time do we want to have in the end in the `<updated>` field?

rmzelle: For large collections like this one, it's probably best not to touch it too often, since updating the time stamp will change **all** the files.

Plus people should retrieve their styles from https://github.com/citation-style-language/styles-distribution anyway, which always has correct time stamps.

dependent/bundesgesundheitsblatt-gesundheitsforschung-gesundheitsschutz.csl
rmzelle: > I think it is also missing everywhere else. I guess we can delete the redundant short-titles but may want to keep the rest.

I think most of the cases where we lose the abbreviated title are from (former) BMC styles. If we do end up retiring the https://github.com/citation-style-language/utilities/tree/master/generate_dependent_styles/data/bmc folder, we can first retrieve the old abbreviations from https://github.com/citation-style-language/utilities/blob/master/generate_dependent_styles/data/bmc/_journals.tab.

zuphilip: Ah, I see. In these examples the issn changed to an eissn and therefore the lookup is not working. But we could make the lookup for matching either issn or eissn in the excel sheet. Then we shouldn't loose the short-title data. Should we do that?

rmzelle: Sure. But you agree that in this case, we actually shouldn't be using "Bundesgesundheitsblatt Gesundheitsforschung Gesundheitsschutz" as the short title, right, since it's the same as the full title?

zuphilip: Yes, I agree.

It might really just miss the cases of bmc. Thus, maybe the other lookup is not needed...

rmzelle:
So, I overwrote the new Springer dependents with those from the data/bmc and data/springer-fachzeitschriften-medizin data sets.

I'm still wondering if there is anything we can do to avoid the renaming issue. The determination on whether a file is renamed seems to occur after the commit (e.g. via git-log or git-diff; see also http://stackoverflow.com/questions/15031576/how-can-i-prevent-git-from-thinking-i-did-a-rename/15031787#15031787 ), which isn't an option for how things look like on GitHub.

Maybe I should separate these big style updates in three commits? In order:

1. style additions
2. style deletions
3. style changes

Thoughts? It would be important to do the additions before the deletions, since styles might get renamed, and with this order the old and new styles would be correctly identified as duplicates (e.g. based on ISSN).

zuphilip:
I checked the first errors from Travis:
 * **Brazilian Journal of Botany** is <a href="http://ezb.uni-regensburg.de/searchres.phtml?bibid=AAAAA&colors=7&lang=de&jq_type1=QS&jq_term1=Brazilian+Journal+of+Botany">published since 2013 by Springer</a> (the Sociedade Botanica de Sao Paulo is still editor) and it is using a standard Springer style. <i>Thus, we should delete the independent style for this journal.</i>
 * **Journal of Chemical Ecology** seems to be a normal Springer journal with a normal Springer style. I cannot see why there exists an independent style for this. @adam3smith : Do you remember why <a href="https://github.com/citation-style-language/styles/commits/master/brazilian-journal-of-botany.csl">you created this 3 years ago</a>?
 * **Strahlentherapie und Onkologie** the independent style refers to a documentation from 2011 which does not exists anymore. This is a journal by Springer and <a href="http://www.springer.com/medicine/radiology/journal/66?detailsPage=pltci_2611870">refering to springer-basic-number on its webpage</a>. <i>Thus, we should delete the independent style for this journal.</i>

adam3smith:
I created the JCE style before we added automated Springer styles. We can delete the independent style.

dependent/european-journal-of-epidemiology.csl
```diff
@@ -11,7 +11,7 @@
     <link href="http://www.springer.com/cda/content/document/cda_downloaddocument/manuscript-guidelines-1.0.pdf" rel="documentation"/>
     <category citation-format="numeric"/>
     <category field="medicine"/>
-    <issn>0393-2990</issn>
+    <issn>1573-7284</issn>
     <eissn>1573-7284</eissn>
```
zuphilip: That is another problem Travis is complaining about: The issn and eissn is the same. Actually the one which is getting deleted is correct.

zuphilip: Okay, that is occuring several times and can be easily solved:
 * european-journal-of-epidemiology --> wrong replacement for issn (the old issn 0393-2990 <a href="http://www.springer.com/public+health/epidemiology/journal/10654">seems correct</a>)
 * aging-clinical-and-experimental-research --> delete issn (only eissn)
 * history-and-philosophy-of-the-life-sciences --> issn should be 0391-9714
 * springer-science-reviews --> delete issn (only eissn)

zuphilip:
I created a pull request #1527 which corrects these issues and make Travis happy again ;-) @rmzelle Several commits might help. However, I guess maybe we have to make anyway a review locally because the large number of files will stay. But I don't know (yet) how this can work most comfortable...

rmzelle:
I also found www.cristin.no/konsortieavtaler/avtaler/springer/cristin-2015-title-list.xlsx, which has a nice list of discontinued/renamed/merged titles. I wonder if we could get something similar for 2013 and 2014.

zuphilip:
@rmzelle : Do we still need such a list? I solved most of the opened cases (filterd out because strange citation style)...

rmzelle:
I asked for one, but we probably can manage without. It would make it easier to confirm all the added journals, though.

zuphilip:
Okay. How will you add the language to the title if it is not English? E.g. for the German (de-DE) style we want to add " (German)" as suffix to the title. Will this happen in the Ruby script?

rmzelle:
I already did this in the spreadsheet with an Excel formula, but I have to rerun the script with the new data.

zuphilip:
Okay. We should add the "no-et-al" to the two styles, where it was deleted, back, because it was changed at some day on purpose. Right?

rmzelle:
> We should add the "no-et-al" to the two styles, where it was deleted, back

Yes. We should fix that using a _extra.tab file, though, so we don't have to make any manual corrections to the overall metadata file.

rmzelle:
I just published https://github.com/citation-style-language/utilities/tree/springer-update-2015, which includes an update to separate perform deletions, additions, and modificiations (as per the new commits above). I again ran the bmc and springer-fachzeitschriften-medizin data sets prior to commit so we don't have to deal with those right now.

We should now update _skip.txt, _rename.tab, and _extra.tab. Might take me until tomorrow.

zuphilip:
IMO are the seperated commits nice for review, because the only too large commit are the additions and I don't see any reason to look there very carefully. The deletions and modifications are more or less discussed in the list and would let you deal with the details and how to change those things. Please let me know if I can help you there on something.

rmzelle:
I cleaned up the pull request a bit, and incorporated all the manual changes that were needed.

The only outstanding issue I'm aware of is that the generation script currently strips all content between parentheses in the titles. As a result, the language for non-English styles still doesn't show up. I'll try to fix up the script to deal with this.

rmzelle:
I fixed the language description issue. None of the _journals.tab files contained square brackets, so I reserved those for information that should be placed within parentheses in the title, and removed from the style ID. Seems to work well.

@adam3smith, @zuphilip, can you give the PR a quick lookover? Once we merge this, we can look at the BMC/SFM styles.

adam3smith:
I haven't followed this super closely, but looked at the PR and looks (apart from the small re-naming mess) fine to me.

rmzelle:
What renaming mess? :P

zuphilip:
Cool, thank you @rmzelle for all the complicated work! The citation count is now near a "magic number", i.e. we have now <a href="https://www.zotero.org/styles/">almost 7777 citation styles</a> in CSL! :sweat_smile:

PR #1528, opened 2015-04-22

agruber:
Update sorting of Bibliography. Now adheres more to the guidlines given here:
http://www.atmospheric-chemistry-and-physics.net/for_authors/manuscript_preparation.html

### DIFF
--- a/copernicus-publications.csl
+++ b/copernicus-publications.csl
@@ -152,8 +152,9 @@
   </citation>
   <bibliography et-al-min="100" et-al-use-first="99">
     <sort>
-      <key macro="author"/>
+      <key macro="author" names-min="3" names-use-first="1"/>
       <key macro="year-date"/>
+      <key macro="author"/>
       <key variable="title"/>
     </sort>
     <layout>


rmzelle:
This doesn't seem quite right yet. You could take a look at https://github.com/citation-style-language/styles/blob/master/american-geophysical-union.csl, which seems to have the same (complicated) sorting. So basically you would need something like:

```
      <key macro="author" names-min="1" names-use-first="1"/>
      <key macro="year-date"/>
```

So the sort order would be:
1. Sort the reference list by first author
2. Sort entries further by the number of authors (1, 2, or >= 3)
3. Sort entries with two authors further by second author (with `names-min="3" names-use-first="1"`, the sort key will consist of the just the first author name for 1 and >= 3 author papers, which wouldn't have any effect [since we already sorted by the first author], but consist of both names for 2 author papers).
4. Sort entries further by year.

Let us know if you want to give implementing this a try. See also http://docs.citationstyles.org/en/stable/specification.html#name for a description of getting the author "count".

agruber:
Hi,

Thanks for the hint. Now it is doing as intended! Slowly starting to 
grasp how this actually works ...
Filed another pull request with the updated version.

Best,
Andreas

rmzelle:
> Slowly starting to
grasp how this actually works ...

Well, you started with one of the most complicated bits, so good job! :)

PR #1529, opened 2015-04-23

agruber:
Second round of updating the style to adhere to the bibliography sorting rules given here: http://www.atmospheric-chemistry-and-physics.net/for_authors/manuscript_preparation.html


rmzelle:
Thanks! Followup of #1528.

@agruber, note for next time that it's easy (and better) to amend an existing pull request. When the time comes we'll give instructions on how to do this.

PR #1536, opened 2015-05-01
Small corrections

bwiernik:
- Fix placement of original date in citations
- Move all modifiers for reports (type, number, version, edition) into
same parentheses


adam3smith:
thanks!

PR #1537, opened 2015-05-01
Create NIH-ImagingPlatform

alisonkozol:


adam3smith:
before we get into the details -- could you explain what this is for specifically?

alisonkozol:
Hello,

I am a first time user of this repository. I am attempting to upload this .csl file so my colleagues can access it for a grant proposal. We are using a combination of Paperpile and Google docs. While I am able to upload the .csl file in Paperpile, I am unable to access this format through Google docs [specifically, the Paperpile plug-in in Google docs]. It seems like I need to first have the .csl file add to the repository here in order to access it through Google docs. 

This .csl format is based on the NIH bibliography style required for this grant proposal. 

Thank you!

rmzelle:
Does https://github.com/citation-style-language/styles/blob/master/national-library-of-medicine-grant-proposals.csl not suffice?

adam3smith:
you should also contact Paperpile about this -- from what I understand, custom styles should work in google docs (via Paperpile --> Citation Styles in that document).

alisonkozol:
There are some differences that differ from the protocol.

Our version:
Singh S, Bray M-A, Jones TR, Carpenter AE (2014). Pipeline for illumination correction of images for high-throughput microscopy. Journal of Microscopy 256(3):231-236 / doi:10.1111/jmi.12178 PMID: 25228240 PMCID: 4359755 (Research Article)

N'tnl library of medicine:
Chickinsky A. The development of a 3D DNA database. IEEE 35th Annual Northeast Bioengineering Conference [Internet]. 2009. p. 1–2. Available from: http://dx.doi.org/10.1109/NEBC.2009.4967734

-we do not specify the medium through which we found the article
-we like to include PMID and PMCID. 
-we specify that the citation is for a research article
-we do not like to include "Available from:" before the doi. 

I have corresponded with a rep. from Paperpile. Unfortunately, you can only access citation styles from the repository for a shared document like this.

adam3smith:
Unfortunately, we can't accept custom styles on the repository--styles posted here are available to millions of users in more than 30 different products, so we need to impose some limits.

FWIW, the NLM style Rintze links to (the one that says grants and proposals) does include the PMCID (though not both PMID and PMCID -- it uses PMCID, then falls back to PMID). 
The DOI isn't listed at all, what you see is the URL. There should be an option in Paperpile to not include the URL.

If at least some of these requirements are more broadly applicable to NIH/NLM grant applications, we can see what we can do about including them in the existing or a new, more generic style.

alisonkozol:
I understand completely. I suppose the main issue with this style are the brackets around the in-text citations. Grant proposals often have page limits, and brackets take up a lot of space! Subscripts without brackets may be a better option for a majority of grant-writers.

PR #1538, opened 2015-05-06
American medical association changes

serpodrick:
Changes for american-medical-association and friends, as indicated in the commit messages.

http://validator.citationstyles.org/ hasn't seemed to be working for at least the last few days: "Validation is taking longer than expected! (more than 6 seconds)". I'm hoping Travis has more patience.

### DIFF
--- a/american-medical-association-alphabetical.csl
+++ b/american-medical-association-alphabetical.csl
@@ -123,7 +123,8 @@
       <choose>
         <if type="bill book graphic legislation motion_picture report song" match="any">
           <group suffix="." prefix=" " delimiter=" ">
-            <text variable="volume" prefix="Vol " suffix="."/>
+            <text term="volume" form="short"  text-case="capitalize-first" strip-periods="true"/>
+            <text variable="volume" prefix=" " suffix="."/>
             <text macro="edition"/>
             <text macro="editor" prefix="(" suffix=")"/>
           </group>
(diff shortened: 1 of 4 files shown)


rmzelle:
The validator works for me (test e.g. http://validator.citationstyles.org/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fcitation-style-language%2Fstyles%2Fmaster%2Fapa.csl&version=1.0 ).

rmzelle:
For easier localization, it would also be better to use 

```
<text term="volume" form="short" strip-periods="true"/>
<text variable="volume" prefix=" " suffix="."/>
```

instead of `<text variable="volume" prefix="Vol " suffix="."/>`

https://github.com/citation-style-language/locales/blob/master/locales-en-US.xml#L191

rmzelle:
And instead of `prefix=" Vol "`, it would be better to place the delimiter on a `<group/>` element. E.g. you can rewrite

```
        <else-if type="chapter paper-conference entry-dictionary entry-encyclopedia" match="any">
          <group prefix=" ">
            <text macro="editor"/>
            <text variable="container-title" font-style="italic" prefix=" " suffix="." text-case="title"/>
            <text variable="volume" prefix=" Vol " suffix="."/>
            <text macro="edition" prefix=" "/>
            <text variable="collection-title" prefix=" " suffix="."/>
            <group suffix=".">
              <group suffix="." prefix="; ">
                <date variable="issued">
                  <date-part name="year"/>
                </date>
                <text variable="page" prefix=":"/>
              </group>
            </group>
          </group>
        </else-if>
```

to

```
        <else-if type="chapter paper-conference entry-dictionary entry-encyclopedia" match="any">
            <text term="in" text-case="capitalize-first" suffix=":"/>
            <text macro="editor"/>
            <text variable="container-title" font-style="italic" suffix="." text-case="title"/>
            <text variable="volume" prefix="Vol " suffix="."/>
            <text macro="edition"/>
            <group suffix=".">
              <text macro="publisher"/>
              <group suffix="." prefix="; ">
                <date variable="issued">
                  <date-part name="year"/>
                </date>
                <text variable="page" prefix=":"/>
              </group>
            </group>
          </group>
        </else-if>
```

serpodrick:
There you go.

Your link to the validator worked the first time for me (though it found 9 errors not of my making!), but after that it stopped working. I'm not sure what the problem is, so maybe it's on my end.

rmzelle:
@serpodrick, sorry, can't reproduce any issues with the validator. The link I sent intentionally produced some errors (it's validating the current version of apa.csl against the old 1.0 schema). Producing validation errors is a good way to make sure validation works correctly. :)

@adam3smith, happy with this?

adam3smith:
I'll check tonight & will merge if there are no concerns from me.

american-medical-association-alphabetical.csl
```diff
@@ -211,9 +213,19 @@
           <group prefix=" " suffix=".">
             <text variable="container-title" font-style="italic" form="short"/>
             <group delimiter=";" prefix=" ">
-              <date variable="issued">
-                <date-part name="year"/>
-              </date>
+             <choose>
```
adam3smith: I don't agree with your reading of the manual here. I think, if you look at the example, they're referring to journal articles with neither volume nor issue. So in line 217 you should have ```variable="issue volume"``` instead.

adam3smith:
Apart from that one comment, everything looks great, thanks!

serpodrick:
Committed.

rmzelle:
Great, thanks Pod!

PR #1542, opened 2015-05-08
[ASM] Re-generate dependent styles with short title

aurimasv:
citation-style-language/utilities#22

### DIFF
--- a/dependent/antimicrobial-agents-and-chemotherapy.csl
+++ b/dependent/antimicrobial-agents-and-chemotherapy.csl
@@ -3,15 +3,16 @@
   <!-- Generated with https://github.com/citation-style-language/utilities/tree/master/generate_dependent_styles/data/asm -->
   <info>
     <title>Antimicrobial Agents and Chemotherapy</title>
+    <title-short>AAC</title-short>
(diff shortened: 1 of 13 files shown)


rmzelle:
Thanks! Are you submitting? :)

aurimasv:
I am indeed. :-)

rmzelle:
Good luck!

rmzelle:
@aurimasv, shouldn't the ASM CSL style be double-spaced? (as it happens, I'm prepping a manuscript for another ASM journal)

aurimasv:
It's possible. The text is and since they don't explicitly mention anything about references, I guess they are too. Looking at recently accepted manuscripts to MCB, about half of them have double-spaced references and half have single-spaced, but given CSL's popularity that might not mean much (it does mean that the editors probably don't care).

aurimasv:
Good luck, btw!

rmzelle:
http://aem.asm.org/site/misc/journal-ita_org.xhtml#link00 says "Type every portion of the manuscript double-spaced (a minimum of 6 mm between lines), including figure legends, table footnotes, and references"

aurimasv:
Good catch. Ooops.

PR #1547, opened 2015-05-12
correction of drawing "locator" in citation

libora:
Code of "locator" drawing in citation was unified wit author-date style

### DIFF
--- a/iso690-numeric-brackets-cs.csl
+++ b/iso690-numeric-brackets-cs.csl
@@ -420,8 +420,15 @@
       <key variable="citation-number"/>
     </sort>
     <layout prefix="[" suffix="]" delimiter=", ">
-      <text variable="citation-number"/>
-      <text variable="locator" prefix=" s. "/>
+      <group delimiter=", ">
+        <group delimiter=" ">
+          <text variable="citation-number"/>
+        </group>
+        <group delimiter=" ">
+          <label variable="locator" form="short"/>
+          <text variable="locator"/>
+        </group>
+      </group>
     </layout>
   </citation>
   <bibliography second-field-align="flush">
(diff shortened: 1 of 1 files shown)


iso690-numeric-brackets-cs.csl
```diff
@@ -420,8 +420,15 @@
...
+      <group delimiter=", ">
+        <group delimiter=" ">
+          <text variable="citation-number"/>
+        </group>
+        <group delimiter=" ">
+          <label variable="locator" form="short"/>
+          <text variable="locator"/>
+        </group>
+      </group>
     </layout>
```
zuphilip: Okay, I see what you are doing, but this can be simplified. Try this (you may want to indent it correctly):
```
     <layout prefix="[" suffix="]" delimiter=", ">


          <text variable="citation-number"/>

        <group delimiter=" ">
          <text variable="locator"/>
        </group>

     </layout>
```
(A group with only one element is not useful and the group inside the layout is not necessary.)

libora: Hi zuphilip, your simpification is not working! 
correct citiation: [1 kap. 5, 2, 3 obr. 8]
citation from your simpification: [1kap. 5, 2, 3obr. 8]

correct simplification is:

    <layout prefix="[" suffix="]" delimiter=", ">
          <text variable="citation-number"/>
          <text variable="locator"/>
        </group>
    </layout>

But I don't like mixing two types of separators (delimiter and prefix) in one tag. I thing that my code is clearer and more understandable.

zuphilip: You are right about the outer group, but the one-child-element-group seems unnecessary. If you want to have a space between `citation-number` and the other two variables, then how about:
```
<layout prefix="[" suffix="]" delimiter=", ">
  <group delimiter=" ">
    <text variable="citation-number"/>
    <group delimiter=" ">
      <text variable="locator"/>
    </group>
  </group>
</layout>
```

libora: Yes, this is a better code as my. Thanks

2015-05-13 9:07 GMT+02:00 Philipp Zumstein <notifications@github.com>:

rmzelle:
Thanks both!

PR #1548, opened 2015-05-12
corection of code for "locator" drawing

libora:
Update part of drawing locator in citation to correct working


rmzelle:
Thanks!

PR #1551, opened 2015-05-13
Update methods-in-ecology-and-evolution.csl

zuphilip:
See https://forums.zotero.org/discussion/49371/edit-citation-style-for-methods-in-ecology-and-evolution/#Comment_224433


rmzelle:
Thanks!

PR #1552, opened 2015-05-14
Adds APA where the citation authors are separated by 'and' instead of…

cpina:
… '&'.

Some places like in Spain or France use this (localized in their
languages)

### DIFF
--- a/apa-and-as-text.csl
+++ b/apa-and-as-text.csl
@@ -1,0 +1,607 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
+  <info>
+    <title>American Psychological Association 6th edition (and as text)</title>
+    <title-short>APA</title-short>
+    <id>http://www.zotero.org/styles/apa-and-as-text</id>
+    <link href="http://www.zotero.org/styles/apa-and-as-text" rel="self"/>
+    <link href="http://owl.english.purdue.edu/owl/resource/560/01/" rel="documentation"/>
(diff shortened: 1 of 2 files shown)


rmzelle:
Oh, wait, did we decide to add this to the repo anyway?

cpina:
Oops, I understood that the style could be added to the repo after I mentioned in the email thread that there are 8 variants of apa such as apa-tr.csl or two apa-fr* and we discussed a correct name (I thought that the name proposals was for the repo, not for internal usage).

If can't be added using any name I'll give instructions to the affected users about how to use it without being in the repository; and apologies for getting over-excited creating it...

rmzelle:
I'll let Sebastian decide, since he does most of the maintenance on the APA styles. I don't think the title/ID is very clear, though. "and-as-text" probably only makes sense to people who know CSL. How about "American Psychological Association 6th edition (no ampersand)" and "apa-no-ampersand.csl"?

cpina:
Ok, I'll wait for @adam3smith 's input. I'm happy with either name. Thanks!

adam3smith:
I like no ampersand -- was still kind of sleepy when I suggested the
"text".
I'm Ok with adding this, since I always feel we're underserving the
non-English populations, though with the understanding that we'll remove it
once we can deal with it in locales as we should.

On Thu, May 14, 2015 at 5:06 PM, Carles Pina Estany <
notifications@github.com> wrote:

> Ok, I'll wait for @adam3smith <https://github.com/adam3smith> 's input.
> I'm happy with either name. Thanks!
>
> —
> <https://github.com/citation-style-language/styles/pull/1552#issuecomment-102182798>
> .
>



-- 
Sebastian Karcher, PhD
Northwestern University

cpina:
I've renamed it.

rmzelle:
Okay, thanks!

PR #1559, opened 2015-05-19
Fix report titles: quotes rather than italics

nickbart1980:
See https://forums.zotero.org/discussion/49362/report-type-in-chicago-style/


adam3smith:
thanks! You'll also do all the note/full note versions?

PR #1569, opened 2015-05-25
Harvard Edge Hill: Align with latest reported version of the style

fbennett:
Per this report to Zotero forums:

    https://forums.zotero.org/discussion/49581/harvard-edge-hill-university-style/#Item_0


adam3smith:
(please don't merge yet. Can't find this in the style guide)

fbennett:
Hmm. Good catch. Looking at the post again, the OP doesn't say actually say they are at Edge Hill U.

PR #1572, opened 2015-05-26
Create international-journal-of-osteoarchaeology.csl

pjaskulski:


international-journal-of-osteoarchaeology.csl
```diff
@@ -0,0 +1,248 @@
...
+      <name form="short" and="symbol" delimiter=", " initialize-with=". "/>
+      <substitute>
+        <names variable="editor"/>
+        <names variable="translator"/>
+        <text macro="anon"/>
+      </substitute>
+    </names>
+  </macro>
+  <macro name="access">
+    <choose>
```
adam3smith: would we really ever want both a DOI and a URL? You'd get this the way this is coded together with the DOI macro in line 240.

international-journal-of-osteoarchaeology.csl
```diff
@@ -0,0 +1,248 @@
...
+            <text macro="editor"/>
+          </group>
+          <group prefix=". ">
+            <text variable="container-title" font-style="italic"/>
+            <group prefix=" " delimiter=" ">
+              <group>
+                <text variable="volume" font-weight="bold"/>
+                <text macro="published-date"/>
+              </group>
+              <text macro="pages" prefix=" : "/>
```
adam3smith: are you sure about the space before the colon? I've never seen that in English. You also have an additional space via the delimiter of the group in row 234

pjaskulski:
This extra space occurs on "for Authors" page (http://onlinelibrary.wiley.com/journal/10.1002/%28ISSN%291099-1212/homepage/ForAuthors.html) in the References section.

But this is probably the mistake. Unnecessary space before the colon does not occur in the printed journal . I'll change it, also DOI/URL issue.

adam3smith:
Great thanks. Agree that the space in the author instructions is just a typo. They're often quite bad for Wiley journals.

PR #1575, opened 2015-06-04
journal-of-reconstructive-microsurgery.csl

go2roland:
The journal of reconstructive microsurgery style is AMA style besides some minor changes:

- References must be listed double-spaced, numbered, in AMA style, using Index Medicus
journal title abbreviations, and in the same sequence as they are cited in the text.

- By way of exception to AMA style, do not italicize book titles or journal title abbreviations and do not put a period at the end of a reference.


rmzelle:
If you look at https://github.com/citation-style-language/styles/pull/1575/files, you can see that you edited the dependent jama.csl style, which is incorrect. To add a new style, use the "+" icon at https://github.com/citation-style-language/styles. (you can make a new pull request after doing that)

adam3smith:
superseded by https://github.com/citation-style-language/styles/pull/1576

PR #1577, opened 2015-06-05
council-of-science-edtiors - The semicolon is a delimiter, not a prefix.

serpodrick:
Two of the CSE styles specify the semicolon as a prefix for the group containing the issued and pages macro. Really, it's a delimiter between that group and the preceding publisher macro. This change makes an extraneous semi-colon go away.


adam3smith:
never going to complain about replacing affixes with delimiters. Thanks!

PR #1579, opened 2015-06-09
Update peerj.csl

burnmeister:
inserted web pages and blog post entries such as 'available at' and 'accessed'

### DIFF
--- a/peerj.csl
+++ b/peerj.csl
@@ -14,7 +14,7 @@
     <category field="biology"/>
     <eissn>2167-8359</eissn>
     <summary>style for the journal PeerJ</summary>
-    <updated>2015-06-09T12:40:12+00:00</updated>
+    <updated>2015-06-26T10:51:31+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="author">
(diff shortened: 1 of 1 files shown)


rmzelle:
`et-al-min="10" et-al-use-first="10"` doesn't make sense. If you have 10 authors, you wouldn't show all 10 and "et al.". ("use-first" should always be less than "min")

Does PeerJ have an "et al." requirement?

zuphilip:
The last reference in the latest article https://peerj.com/articles/cs-8/ is actually showing all 11 authors. However, in the HTML the last author is behind the " +1 more" link. My guess would be that we shouldn't use `et-al` in the references, i.e. this was correct before.

burnmeister:
I also don't see any 'et al' requirements in the reference section guidelines ( https://peerj.com/about/author-instructions/#reference-section )

zuphilip:
<blockquote>Each journal reference should be listed using this format: the full list of Authors with initials...</blockquote>
This sounds like they really want to have all authors and never an et-al.

adam3smith:
OK, since there's now a consensus, could you remove the new et-al from the style? You can update your pull request by using the pencil (edit) button here:
https://github.com/burnmeister/styles/blob/patch-7/peerj.csl

Let us know when you're done, since we don't get notified on updated pull requests.

burnmeister:
I have removed et-al-min="10" et-al-use-first="10"  and updated the pull request

rmzelle:
Thanks!

PR #1580, opened 2015-06-09
PMC-22661 - add support for ahead-of-print

Klortho:
This pull request modifies both american-medical-association.csl and apa.csl, to add support for ahead-of-print, as described in the xbiblio-devel email thread, [How to encode and process "ahead of print"](http://sourceforge.net/p/xbiblio/mailman/message/34179904/).

Note that american-medical-association.csl uses a new date field, `epub-date`; so this PR should not be merged before support for that is added.

### DIFF
--- a/american-medical-association.csl
+++ b/american-medical-association.csl
@@ -77,6 +77,17 @@
         <text variable="title"/>
       </else>
     </choose>
+    <choose>
+      <if variable='volume'>
+      </if>
+      <else>
+        <group prefix=' [' suffix=']'>
+          <!-- FIXME: this should be localized -->
+          <text value='published online ahead of print '/>
+          <date variable="epub-date" form="text"/>
+        </group>
+      </else>
+    </choose>
   </macro>
   <macro name="publisher">
     <group delimiter=": ">
@@ -97,6 +108,7 @@
       </else>
     </choose>
   </macro>
+
   <citation collapse="citation-number">
     <sort>
       <key variable="citation-number"/>
--- a/apa.csl
+++ b/apa.csl
@@ -150,6 +150,14 @@
       </if>
       <else>
         <choose>
+          <if variable='volume'>
+          </if>
+          <else>
+            <!-- FIXME: this should be localized -->
+            <text value='Advance online publication. '/>
+          </else>
+        </choose>
+        <choose>
           <if variable="DOI">
             <text variable="DOI" prefix="http://doi.org/"/>
           </if>


adam3smith:
you may want to split those up then. We're not going to add a new field super quickly - that'd be part of a new CSL release and that's not going to happen within, say, the next month.

Klortho:
Well, is it okay if this just stays open until then? We won't be merging this repo back into our deployed service until after this is dealt with.

adam3smith:
Since the next version will likely happen in a different branch, I don't
think that makes a lot of sense, to be honest, but if it makes things much
easier for you, I also don't mind terribly.

rmzelle:
Ideally I'd like to have just one open ticket for a new date variable (probably in the "schema" repo). I'm pretty sure that we won't be calling it "epub-date", though, so a styles repo PR for that is a bit premature.

rmzelle:
Alternative names:

"issued-online"
"issued-electronic"
"available"

Klortho:
I like "issued-online". Close this if it makes things easier for you.

rmzelle:
And I don't want to rain on your parade, but it will take a while to get this into CSL (even though I agree we should add this as a date variable). You probably noticed we don't have CSL releases very often, and prepping a new release takes several months at least, and we haven't really started on the next one.

ZomboM:
No rain at all -- that's the beauty of open source, and git in particular. We can work in our own branches until it's added. Thanks for the consideration.

bdarcus:
I wonder if we might want to start to think about easy-to-release minor updates (variable additions) vs major updates that involve more fundamental changes?

adam3smith:
Agree. I think it'll take us some time to put out something that's truly a new version, but it'd be nice to catch up on terms and a couple of additional variables. 
You want to bring it up on xbiblio to think through the logistics?

Also closing this thread as per Chris's comment above.

rmzelle:
> I think it'll take us some time to put out something that's truly a new version, but it'd be nice to catch up on terms and a couple of additional variables. 
You want to bring it up on xbiblio to think through the logistics?

We can just repeat what we did for CSL 1.0.1, which was a backwards-compatible release. I'm not sure we want very frequent releases, since that complicates workflows that rely on validation.

adam3smith:
Right, I was thinking maybe once a year if there's a need. We're approaching 3 years since 1.0.1

rmzelle:
Yes, it's definitely time for a new release. (and we might want to switch to correct semantic versioning, so the next backwards-compatible releases would be 1.1, and the next incompatible release 2.0).

fbennett:
I'm just waking up to @Klortho's issue, led back here from a citeproc-js pull request that has been languishing. No problem with the change here, and `issued-online` for the variable name is fine. I'll wait for a schema change before moving it into the citeproc-js repo.

Issue #1578, opened 2015-06-08
Pflügers Archiv - European Journal of Physiology


zuphilip:
Currently assigned to Springer Basic (brackets), but should be reassigned to Springer Basic (brackets, alphabetical). In which list should this exactly be changed?

reference: https://forums.zotero.org/discussion/49691/style-error-pfluegers-archiv-european-journal-of-physiology/

Jmuccigr:
Took a shot at updating this myself. See pull request [1581](https://github.com/citation-style-language/styles/pull/1581).

PR #1581, opened 2015-06-09
Switch to alphabetical

Jmuccigr:
Following up on issue [#1578](https://github.com/citation-style-language/styles/issues/1578).


adam3smith:
thanks! (note to self, @rmzelle and @zuphilip -- this still needs to be changed in utilities).

PR #1586, opened 2015-06-12
remove name-as-sort-order="all"

reagle:
First Monday requires "First Last, First Last, and First Last", remove the attribute that was preventing that.

### DIFF
--- a/first-monday.csl
+++ b/first-monday.csl
@@ -25,7 +25,7 @@
   </macro>
   <macro name="author">
     <names variable="author">
-      <name name-as-sort-order="all" and="text" sort-separator=" " delimiter=", " delimiter-precedes-last="never"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks! Given what you write, you should also delete the delimiter-precedes-last="never" attribute (since there is a comma before the and).

reagle:
I've not used the edit-and-merge on Github feature before. Are you asking me to do this? I'm not sure how.

adam3smith:
no, you can simply update your pull request by using the pencil button here to edit the file:
https://github.com/reagle/styles/blob/patch-1/first-monday.csl 
let us know when you're done

reagle:
Okay, done.

adam3smith:
great, thanks!

PR #1588, opened 2015-06-16
american-chemical-society Misc. changes

serpodrick:
Hi

Three changes for the ACS styles as specified in the commit messages.

### DIFF
--- a/american-chemical-society-author-date.csl
+++ b/american-chemical-society-author-date.csl
@@ -248,7 +248,7 @@
             </date>
           </group>
         </else-if>
-        <else-if type="article-journal">
+        <else-if type="article-journal review" match="any">
           <group delimiter=" ">
             <text macro="title" suffix="."/>
             <text variable="container-title" font-style="italic" form="short"/>
--- a/american-chemical-society-author-date.csl
+++ b/american-chemical-society-author-date.csl
@@ -169,8 +169,10 @@
             <text variable="container-title" font-style="italic" suffix="."/>
             <text macro="edition"/>
             <text macro="publisher"/>
-            <text macro="full-issued" suffix=","/>
-            <text macro="pages"/>
+            <group delimiter=", ">
+              <text macro="full-issued"/>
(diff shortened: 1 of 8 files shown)


rmzelle:
Thanks!

adam3smith:
Hey @serpodrick -- was there a particular issue you were addressing with this? It seems like it may be causing problems for Papers. I think that's mainly their implementation issue, but I'm also not entirely clear what scenario you had in mind with this change. E.g. in Mendeley, article is a fallback category, so just using the style's fallback (i.e. else) formatting makes sense to me?

rmzelle:
Specifically, the problem is that apparently, all "article-journal" items match against "article" in Papers, which creates incorrect output.

serpodrick:
I am translating contribution types from an in-house database and I was trying to account for articles that aren't in a journal, a magazine or a newspaper. This particular change was intended to account for "Newsletter Articles", but it's not the only one.

I don't think citeproc-js uses article when the type is article-journal.

rmzelle:
> I don't think citeproc-js uses article when the type is article-journal.

No, it doesn't, and CSL processors shouldn't. It would be nice if we could fix this in the style, though, at least as an interim option.

One solution would probably be to move up `<else-if type="article-journal review" match="any">` as the first entry of `<choose/>`. That way any "article-journal" item would first match that.

serpodrick:
I pushed a change to do that: e297ab14f6813deefbe314e263824e38abfc68a8. Other than validating the files, I didn't test them. (Since this request is merged, it looks like I would need to create a new request, right?)

rmzelle:
@adam3smith, should we go with the solution from https://github.com/citation-style-language/styles/commit/e297ab14f6813deefbe314e263824e38abfc68a8 ?

adam3smith:
Away from computer, but going by descriptions, yes, sound good

rmzelle:
@serpodrick, yes, we'd appreciate a fresh PR.

PR #1589, opened 2015-06-17
Create journal-of-physical-therapy-science.csl

libora:
Discussion on forum is here 
https://forums.zotero.org/discussion/49784/style-request-journal-of-physical-therapy-science/

### DIFF
--- a/journal-of-physical-therapy-science.csl
+++ b/journal-of-physical-therapy-science.csl
@@ -20,6 +20,10 @@
     <terms>
       <term name="page" form="short">
         <single>p</single>
+        <multiple>p</multiple>
+    </term>
+    <term name="number-of-pages" form="short">
+        <single>p</single>
         <multiple>pp</multiple>
     </term>
   </terms>
@@ -27,7 +31,7 @@
   <macro name="accessed">
     <choose>
       <if variable="accessed">
-        <text term="accessed" suffix=" "/>
+        <text term="accessed" suffix=" " text-case="capitalize-first"/>
         <date variable="accessed">
           <date-part name="month" form="short" suffix=". "/>
           <date-part name="day" suffix=", "/>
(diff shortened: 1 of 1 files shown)


journal-of-physical-therapy-science.csl
```diff
@@ -0,0 +1,151 @@
...
+              </group>
+            </else-if>
+            <else>
+              <text macro="container-title" suffix=", "/>
+            </else>
+          </choose>
+          <date variable="issued">
+            <date-part name="year" suffix=", "/>
+          </date>
+          <text variable="volume" suffix=": "/>
```
adam3smith: Books and Chapter can have volumes, too, and the way you have this coded, you'd end their citation with something like ```3:``` where they do. I'd code the three item types entirely separately, i.e. extend the ```<else>``` from line 138 down to line 146. And then maybe include the publication date in the publisher macro?

Also, since we do see journal articles without page numbers (e.g. online first publications) this would look weird. I'd prefer if both the comma after year and the colon after volume were set as delimiters in a group.

libora: "Books and Chapter can have volumes, too" in instruction for authors is nothing about volumes for books:
"Book citations should give authors, book title, edition, city, publisher, year, and page numbers, if necessary.
example:
.2) Gibson MJ: Falls in later life. In: Improving the Health of Older People; A World View. New York: Oxford University Press, 1990, pp 296–315."

"I'd code the three item types entirely separately, i.e. extend the <else> from line 138 down to line 146." - OK I will prepare separate code for 1. web page, 2. book, 3. articles, 4. chapters, 5. paper conference and 6. others

"I'd prefer if both the comma after year and the colon after volume were set as delimiters in a group." - I agree

adam3smith: right, what I meant with "Books and Chapter can have volumes too" is that you might have those in the Zotero data for a book item, and then they'd show up in citations, which, as you note, doesn't correspond to the guide. 
Thanks for fixing this, let us know when you're done.

adam3smith:
@libora could you fix the remaining issues we agreed on as per above? I would like to merge this.

libora:
Yes I can do it in this week. Last months I have not free time, sorry.

2015-08-02 20:10 GMT+02:00 Sebastian Karcher <notifications@github.com>:

libora:
I hope, that now is code of style O.K.

adam3smith:
looks good, thanks!

PR #1590, opened 2015-06-17
Update psychological-medicine.csl

zuphilip:
reported https://forums.zotero.org/discussion/29448/style-error-psychological-medicine/#Comment_226722
* add point after title
* adjust some uppercase afterwards


adam3smith:
thanks!

PR #1594, opened 2015-06-23
Update pnas.csl

and3k:
PNAS does not have links after all articles. I modified pnas.csl accordingly, so URLs are only shown if there is no page number (like e.g. in The ISME Journal CSL).

### DIFF
--- a/pnas.csl
+++ b/pnas.csl
@@ -14,6 +14,10 @@
       <name>Rintze Zelle</name>
       <uri>http://twitter.com/rintzezelle</uri>
     </contributor>
+    <contributor>
+      <name>Bela Hausmann</name>
+      <uri>https://github.com/and3k</uri>
+    </contributor>
     <category citation-format="numeric"/>
     <category field="science"/>
     <issn>0027-8424</issn>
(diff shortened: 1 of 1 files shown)


adam3smith:
what reference manager/citeproc are you working with? Pretty much all styles are written with that particular logic implied (because that's what Zotero does by default and Mendeley offers as an option).

and3k:
I’m using Mendeley and let it export my citations to BibTeX (where the URL field is filled out by Mendeley). With the current version of pnas.csl I then end up with these URLs being in all my references, which should not be for articles, books, etc, only for some special types.

adam3smith:
Right, so the Mendeley part doesn't matter. The BibTex it exports doesn't care about citation styles. 
There is no solution/flag for this in pandoc? I can accept this PR since there's not much of a downside, but this is really a larger issue that I think should be addressed on the pandoc-citeproc end, as you'll otherwise get incorrect URLs in 100s of styles. Would you mind bringing up with the developers there? 

I'd leave this open for a couple of days so they can take a look.

and3k:
I’m not sure I get it…?

Isn’t the behaviour of pandoc-citeproc correct as per the CSL syntax, i.e. the current style says if there is a a URL field always use it? And all the other software are interpreting the CSL incorrectly?

The "new" URL behaviour is also already implemented in some of the CSL, e.g. https://github.com/citation-style-language/styles/blob/master/the-isme-journal.csl

adam3smith:
CSL doesn't specify which data should be passed on to the citeproc, so there's not right or wrong here. What Mendeley and Zotero do is to simply ommit the URL from the data sent to the citeproc, which then interprets it according to the spec.

and3k:
OK. I see.

Even if Mendeley/Zotero/etc. makes decisions based on the entry type or user configuration to include or not include the URL (or any other field), IMHO it still should be the responsibility of the CSL to ensure the output is correct. 

In this example with PNAS, there should never be an URL appended to journal article citation, even if its in the data.

I also don’t think pandoc-citeproc should make the decicion if to include URLs or not, but maybe @jgm thinks differently about that…

On a side note, the Mendeley option to include or not include URLs only effects the "Copy as formatted citation" option, but neither automatic export nor copy as BibTeX honour it.

Anyway, it’s not terrible important to me that this is merged since I can easily modify the CSL myself, so feel free to close it if you disagree :)

Thanks!
Bela

rmzelle:
@dstillman, @simonster, does either one of you remember how the "Include URLs of paper articles in references" ("Cite/Styles" tab of preferences) ended up in Zotero? It always struck me as weird that Zotero offered this UI option since we can just test for the presence of a page-range in CSL. Was it just a very common citation format customization request? We might be better off without it, instead of pushing of the adoption of this setting in other products (like pandoc-citeproc).

adam3smith:
I guess I'll be accepting this then. 
We should still discuss the larger issue, but probably somehere else @rmzelle

rmzelle:
Yeah, we should raise this on xbiblio-devel.

PR #1596, opened 2015-06-24
Add style for "Environnement, Risques & Santé"

fredomatic:


environnement-risques-and-sante.csl
```diff
@@ -0,0 +1,130 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" default-locale="en-US">
+  <info>
+    <title>Environnement Risques et SantÃ©</title>
```
adam3smith: you'll want to fix the title here.

environnement-risques-and-sante.csl
```diff
@@ -0,0 +1,130 @@
...
+    <author>
+      <name>Frederic Y. Bois</name>
+      <email>frederic.bois@ineris.fr</email>
+    </author>
+    <category citation-format="numeric"/>
+    <category field="medicine"/>
+    <category field="science"/>
+    <issn>1635-0421</issn>
+    <eissn>1952-3398</eissn>
+    <summary>Style for the French journal Environnement Risques et SantÃ©</summary>
```
adam3smith: same here.

environnement-risques-and-sante.csl
adam3smith: since the journal uses & in its title, I'd use that, too. You'll have to write it as ```&amp;```

environnement-risques-and-sante.csl
```diff
@@ -0,0 +1,130 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" default-locale="en-US">
+  <info>
+    <title>Environnement Risques et SantÃ©</title>
+    <id>http://www.zotero.org/styles/environnement-risques-and-sante</id>
```
adam3smith: but then given this is French, use "et" in the id/self link/filename, i.e. ```environnement-risques-et-sante```

environnement-risques-and-sante.csl
```diff
@@ -0,0 +1,130 @@
...
+    <link href="http://www.zotero.org/styles/environnement-risques-and-sante" rel="self"/>
```
adam3smith: same here and in the filename.

adam3smith:
Rest looks good.
You can update the style using the pencil icon here: https://github.com/fredomatic/styles/blob/patch-2/environnement-risques-and-sante.csl
Let us know when you're done.

rmzelle:
@fredomatic, let us know if you need help with this.

fredomatic:
Thanks for the proposed help. I am very busy at the moment, that's why I
did not move after Sebastian Karcher's messages. I am used to version
control management, if not on gitub, so I should be OK. I'll probably do
it next week.

Frederic

rmzelle:
Okay, thanks! Don't mean to rush you.

fredomatic:
OK, done I think.

Frederic

rmzelle:
Great, thanks!

PR #1597, opened 2015-06-25
Update mohr-siebeck-recht.csl

zuphilip:
See https://forums.zotero.org/discussion/50104/keine-seitenzahl-bei-mohrsiebeck-in-fussnote/#Comment_227251

I didn't touch the remaining code, which could maybe also improved.

### DIFF
--- a/mohr-siebeck-recht.csl
+++ b/mohr-siebeck-recht.csl
@@ -11,7 +11,7 @@
     </author>
     <category citation-format="note"/>
     <category field="law"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
what happens with form="symbol" if there is no symbol defined? Does it revert to short?

zuphilip:
Eh... yes
<blockquote>“symbol” first falls back to “short”, and “verb” and “short” both fall back to “long”</blockquote>
from http://docs.citationstyles.org/en/stable/specification.html#terms .

Anyway, would you prefer short? This was almost a 50-50 decision from my part...

adam3smith:
thanks, I could have looked myself. I don't have a preference between the two, either. Do you have any sense what'd be more likely in a German law text?

zuphilip:
I guess one would like to have the "§" symbol which is defined as <a href="https://github.com/citation-style-language/locales/blob/master/locales-de-DE.xml#L197">a symbol</a>.

adam3smith:
cool, let's go with this, then.

PR #1605, opened 2015-07-01
Create meteorological-applications.csl

ziyang:
Style for the Meteorological Applications. This style is almost entirely copied from peerj.csl, whose author is Burnmeister (sybrandvs@yahoo.com).


rmzelle:
@adam3smith, metadata looks good to me.

rmzelle:
Thanks!

PR #1607, opened 2015-07-03
Delete journal-of-oral-biosciences.csl

Sankar-GEO-ELS:
As per email from the Publishing Editor and clarification with Carles I propose deleting this dependent style as the output has some errors in it.


rmzelle:
Superseded by #1608.

PR #1612, opened 2015-07-06
journal-of-cardiology-cases.csl

Sankar-GEO-ELS:
deleting dependent and creating independent style for journal-of-cardiology-cases


adam3smith:
perfect, thanks!

PR #1613, opened 2015-07-06
Improve interpunctation in mohr-siebeck-recht.csl

zuphilip:
See https://forums.zotero.org/discussion/50104?page=1#Item_11


rmzelle:
Thanks!

PR #1618, opened 2015-07-11
Create mlz-russian-gost-r-7-0-11-2011-full-note.csl

Vladimir-VS:
This offers a multilingual Russian GOST R 7.0.11 (dissertations standard) compliant style with full note citation and alphabetic sorting.


rmzelle:
We don't accept MLZ styles in the CSL repository. Maybe @fbennett will be able to find a home for this on his MLZ website.

PR #1620, opened 2015-07-14

vl2:
Minor fixes.

### DIFF
--- a/international-labour-organization.csl
+++ b/international-labour-organization.csl
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+<style class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB" xmlns="http://purl.org/net/xbiblio/csl">
+<!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
     <title>International Labour Organization</title>
     <id>http://www.zotero.org/styles/international-labour-organization</id>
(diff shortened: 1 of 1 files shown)


international-labour-organization.csl
```diff
@@ -158,12 +160,12 @@
             </group>
           </group>
         </else-if>
-        <else-if type="chapter paper-conference" match="any">
-          <text macro="title" prefix=" " suffix="."/>
+        <else-if type="chapter" match="any">
+          <text macro="title" prefix=" &quot;" suffix="&quot;,"/>
```
rmzelle: It's better to use the "quotes" attribute here (`quotes="true"`). See http://docs.citationstyles.org/en/stable/specification.html#quotes

international-labour-organization.csl
```diff
@@ -190,17 +192,24 @@
...
               <text macro="publisher"/>
-              <text variable="volume"/>
+              <text variable="volume" text-case="capitalize-first"/>
               <text macro="pages"/>
             </group>
           </group>
         </else-if>
+        <else-if type="article-newspaper" match="any">
+          <group>
+            <text macro="title" prefix=" &quot;" suffix=".&quot;"/>
```
rmzelle: It's better to use the "quotes" attribute here (`quotes="true"`). See http://docs.citationstyles.org/en/stable/specification.html#quotes

adam3smith:
Please fix the two quotes issues mentioned by @rmzelle above, otherwise this is good to go. 
You can update your pull request by using the pencil icon here:
https://github.com/vl2/styles/blob/patch-3/international-labour-organization.csl
Let us know when you're done of if you need any help.

rmzelle:
@vl2, if you could update this pull request, that would be great.

international-labour-organization.csl
vl2: Thanks for the validation. Just back from the holidays and heard there are
also some other changes needed.
I'll update the pull request with "quotes" + other modifications asap.

rmzelle:
Closing for inactivity.

@vl2, we can still accept this pull request if you make the changes we requested above and post a comment here that you have done so.

PR #1622, opened 2015-07-20
reviews-of-modern-physics-with-titles Changes

serpodrick:
Hi

Some changes for reviews-of-modern-physics-with-titles for conference proceedings and magazine and newspaper articles.


rmzelle:
Thanks!

PR #1623, opened 2015-07-20
CSE and ACS - in term

serpodrick:
Hello

One similar change for american-chemical-society, council-of-science-editors and friends, to prevent extraneous appearance of "in".


rmzelle:
Looks like two styles don't have the "title" macro.

serpodrick:
Fixed.

rmzelle:
Thanks!

PR #1631, opened 2015-07-24
Update mammal-review.csl

icalandra:
According to the guidelines,
1) "et al." should be in italics in the text. New lines 43 and 54 added.
2) When there are more than one citation for a given year by the same first author, it should be differentiated with letters but not by adding more authors. Guidelines: "Where more than one reference by the same author(s) appeared in the same year, use '2009a, b' in both text and References". Line 98 edited (new line 100).
3) The URL should not appear in the reference list (no mention of it and not in the examples in the guidelines). Lines 165-167 deleted.
4) For a book chapter, the page numbers should be preceded by ",". This works, except if there is a collection, which ends with a period. Line 144 edited (new line 146).
5) For a book, title and publisher should both end with a period. Modifications to line 129 (new line 131) and to line 132 (new line 134), respectively.

### DIFF
--- a/mammal-review.csl
+++ b/mammal-review.csl
@@ -18,7 +18,7 @@
     <category field="zoology"/>
     <issn>0305-1838</issn>
     <eissn>1365-2907</eissn>
-    <updated>2012-09-09T22:24:43+00:00</updated>
+    <updated>2015-03-24T13:44:48+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
@@ -40,6 +40,7 @@
     <names variable="author">
       <name name-as-sort-order="all" sort-separator=" " initialize-with="" delimiter=", " delimiter-precedes-et-al="never"/>
       <label form="short" prefix=" (" suffix=")" strip-periods="true"/>
+      <et-al font-style="italic"/>
       <substitute>
         <names variable="editor"/>
         <names variable="translator"/>
@@ -50,6 +51,7 @@
   <macro name="author-short">
     <names variable="author">
       <name form="short" and="symbol" delimiter=", " initialize-with=". "/>
+      <et-al font-style="italic"/>
       <substitute>
         <names variable="editor"/>
         <names variable="translator"/>
@@ -95,7 +97,7 @@
       </if>
     </choose>
   </macro>
-  <citation collapse="year-suffix" et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true">
+  <citation collapse="year-suffix" et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true">
     <sort>
       <key variable="issued"/>
       <key variable="author"/>
@@ -126,10 +128,10 @@
       <choose>
         <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
           <group>
-            <text macro="title" prefix=" "/>
+            <text macro="title" prefix=" ", suffix="."/>
             <text macro="edition"/>
           </group>
-          <text prefix=" " suffix=" " macro="publisher"/>
+          <text prefix=" " suffix=". " macro="publisher"/>
         </if>
         <else-if type="chapter paper-conference" match="any">
           <text macro="title" prefix=" "/>
@@ -141,7 +143,7 @@
           </group>
           <group>
             <text variable="container-title" font-style="italic" prefix=" " suffix=","/>
-            <text variable="collection-title" prefix=" " suffix="."/>
+            <text variable="collection-title" prefix=" " suffix=","/>
             <group suffix=".">
               <text variable="page" prefix=" " suffix="."/>
               <text macro="publisher" prefix=" "/>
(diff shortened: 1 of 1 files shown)


icalandra:
I have no idea what's wrong the the checks; I don't understand the errors... Can anyone help please?

icalandra:
I had a typo in the first version. I corrected it in the second and the checks are now fine.

rmzelle:
Thanks. The most recent free article I could find from Jan 2015 doesn't use italics for "et al.", but maybe this changed recently.

And this isn't your fault, but the current style is rather sloppy in how it places punctuation. It's generally much better to place delimiting punctuation (many of the spaces, commas, and periods) on `<group/>` element or on "delimiter" attributes. See https://zoteromusings.wordpress.com/2013/10/28/writing-csl-features-and-best-practices/ for some more background. E.g. you could rewrite

```
          <group>
            <text variable="container-title" font-style="italic" prefix=" " suffix=","/>
            <text variable="collection-title" prefix=" " suffix=","/>
            <group suffix=".">
              <text variable="page" prefix=" " suffix="."/>
            </group>
          </group>
```

to

```
          <group prefix=" " delimiter=", " suffix=".">
            <group delimiter=". ">
              <text variable="page"/>
              <text macro="publisher"/>
            </group>
          </group>
```

rmzelle:
(let us know if you would like to try your hand at improving the use of `<group/>` and delimiters. Otherwise I'll accept this as is.

icalandra:
Sorry to answer that late. I'm currently in holidays so I'll check that next week (and yes, I'd like to try out with <group>)
Thank you for your answer in any case!

icalandra:
Ok, I'm back from holidays (...).

You're right about the italics. It looks like the guidelines do not correspond to the published articles...
I have tried to play with <group>. Can you please check it?

rmzelle:
Better, thanks!

icalandra:
I have just downloaded the latest version of this style on the Zotero repository and the mistakes for URL and citation disambiguation are still there.
<text variable="URL" text-decoration="none" prefix=" "/> 
and 
disambiguate-add-names="true" 
should be deleted

rmzelle:
I just checked and https://www.zotero.org/styles/mammal-review?source=1 currently offers the new version (check e.g. the timestamp in the style).

icalandra:
Ok, there's something weird here that I do not understand. When I check the source, it is indeed correct. But when I download the file (either through updating in Firefox or the csl file itself), I get an old version...
What am I doing wrong?

rmzelle:
Are you using Zotero Standalone? You might have to install the style via the Preferences panel (see https://www.zotero.org/support/preferences/cite#styles).

Otherwise you can reset your styles (see https://www.zotero.org/support/preferences/advanced), but note that that would delete any custom-installed styles from your Zotero installation.

icalandra:
Still the same problem. 
I am using Zotero for Firefox so I started Firefox again but no change.

The problem is really that the csl file I download from the repository is not the latest version, although the source (when I click on the "source" button next to the name) is indeed the latest.

Just to make sure we're talking about the same thing, I go there https://www.zotero.org/styles/ and then type mammal review and click on the result. Then, I can download the csl file and/or update the style. As I said, these are not the latest versions.

It seems to me that the problem is on the repository itself, as if the style was not linked to the correct file.

adam3smith:
the repository is fine. just tested. How are you making sure the version you download is not the latest version?

rmzelle:
The entry for Mammal Review shows the new timestamp ("Mammal Review (2015-08-03 08:55:01)", see https://www.zotero.org/styles/?q=mammal%20review ), and if I download the file, either via https://www.zotero.org/styles/mammal-review?source=1 or by right-clicking the style title and downloading the file, I get the latest version as well.

As long as you see the new timestamp on the website, there really isn't a problem with the Zotero Style Repository.

icalandra:
@adam3smith : I was sure of it because of the time stamp and because of the errors
@rmzelle : Ok, I feel stupid now, but I can assure you that 1h ago, I didn't get the latest version (either as a file, or as an update in Zotero). Now, I do, doing exactly the same procedure.

So everything is fine now :)

icalandra:
Sorry for the bother

rmzelle:
@dstillman, can you check the time for the server https://github.com/citation-style-language/distribution-updater runs on? Seems a bit off. https://github.com/citation-style-language/styles-distribution/commit/7740658f5814c7272243556a2a855ce599f672ac used "2015-08-03T08:55:01+00:00", even though it was processed between "Aug 3, 2015, 10:40 PM EDT" (https://github.com/citation-style-language/styles/commit/df3d4356455b3dff18f71e2459e307f1b790cc7d) and "Aug 3, 2015, 10:43 PM EDT" (https://github.com/citation-style-language/styles-distribution/commit/7740658f5814c7272243556a2a855ce599f672ac). With EDT being UTC -4:00, that doesn't seem to make sense.

rmzelle:
@icalandra, no problem. But I don't think https://www.zotero.org/styles/ caches, so reloading the page should have given you the new style, and it looks like the repository was updated shortly after I accepted your pull request.

icalandra:
I even closed Firefox.

dstillman:
@rmzelle: https://github.com/citation-style-language/styles/commit/e6279d0ea2f9cdc4e08e7831c857c05950c32696 (Mon Aug 3 08:55:01 2015 +0000)

rmzelle:
Oh, it uses the file modification date, then?

dstillman:
Well, it uses the commit timestamp from the file, not the merge commit's timestamp.

rmzelle:
My bad.

PR #1632, opened 2015-07-25
Update neurosurgical-review.csl

zuphilip:
https://forums.zotero.org/discussion/50782/neurosurgical-review/


zuphilip:
Should we add this to the skip list
https://github.com/citation-style-language/utilities/blob/master/generate_dependent_styles/data/springer/_skip.txt
similar to "acta-neuropathologica"?

adam3smith:
thanks! Yes, I'll handle the skip list

PR #1633, opened 2015-07-25
Update austrian-legal.csl

zuphilip:
Improve the interpunctation and simplify some code, cf https://forums.zotero.org/discussion/50736/fehler-bei-zitierstil-moor-siebeck/#Item_6


rmzelle:
Should we try to keep the XML comments? @adam3smith?

(@zuphilip, unfortunately the CSL editor strips XML comments from styles)

zuphilip:
I manually added some XML comments before submitting. However, the other comments are (at least for me) mostly unnecessary.

adam3smith:
I'm OK taking these out. Agree with @zuphilip we should be judging this case-by-case.

rmzelle:
Thanks!

PR #1635, opened 2015-07-28
Update archiv-fur-geschichte-der-philosophie.csl

zuphilip:
The other link to the documentation seems not to work currently. However, it seems there might be an English and German version of the instructions for authors. 

FYI
<blockquote>Das AGPh nimmt Beiträge in den Sprachen Deutsch, Englisch und Französisch zur Begutachtung an.</blockquote>


adam3smith:
thanks. We could consider removing the default locale, there doesn't appear to be anything hard-coded locale wise (though at first glance the overall quality of the style doesn't look terribly high).

PR #1636, opened 2015-07-30
Update journal-of-separation-science.csl

zuphilip:
* This covers the open points from https://forums.zotero.org/discussion/27948?page=1#Item_8 
 * Add some other improvements/clean-up


rmzelle:
Thanks!

PR #1638, opened 2015-08-01

FrancoisC37:
The whole corrections, I've done a mistake with my post earlier this day (and i don't know how to edit it...).
In this one, the macro "collection" is correct!
Sorry

### DIFF
--- a/documents-d-archeologie-francaise.csl
+++ b/documents-d-archeologie-francaise.csl
@@ -356,10 +356,8 @@
     </choose>
   </macro>
   <macro name="collection">
-    <group delimiter=", ">
-<text variable="collection-title" prefix=", coll. « " suffix=" »"/>
+<text variable="collection-title" prefix=", coll. " quotes="true"/>
 <text variable="collection-number" prefix=", "/>
-    </group>
   </macro>
   <macro name="page">
     <choose>
@@ -449,7 +447,7 @@
               <text macro="publisher" prefix="&#160;: "/>
               <text macro="url" prefix=". "/>
               <text macro="accessed" prefix=", "/>
-              <text macro="collection" prefix=", coll. « " suffix=" »"/>
+              <text macro="collection"/>
               <text variable="volume" prefix=", "/>
               <text variable="number-of-volumes" prefix=", "/>
               <text macro="isbn" prefix=". "/>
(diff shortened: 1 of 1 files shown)


documents-d-archeologie-francaise.csl
```diff
@@ -357,8 +357,8 @@
   </macro>
   <macro name="collection">
     <group delimiter=", ">
-      <text variable="collection-title"/>
-      <text variable="collection-number"/>
+<text variable="collection-title" prefix=", coll. « " suffix=" »"/>
```
adam3smith: it would be much better to set the guillemets using ```quotes="true"```, i.e. in this case
```<text variable="collection-title" prefix=", coll. " quotes="true"/>``` among other advantages, it makes the space between guillemet and title non-breaking, so that you won't get a line break right there. 

Also, if you could indent this properly, that'd be great. You can use: http://rintze.zelle.me/style-formatter/ to do this for all lines with one click.

File: documents-d-archeologie-francaise.csl [hunk 2]
```diff
@@ -449,7 +449,7 @@
               <text macro="publisher" prefix="&#160;: "/>
               <text macro="url" prefix=". "/>
               <text macro="accessed" prefix=", "/>
-              <text macro="collection" prefix=", coll. « " suffix=" »."/>
+              <text macro="collection" prefix=", coll. « " suffix=" »"/>
```
adam3smith: again, use quotes="true"

documents-d-archeologie-francaise.csl
```diff
@@ -493,7 +493,8 @@
               <text macro="publisher" suffix=", "/>
               <text macro="archive" suffix=", "/>
               <text macro="page"/>
-              <text macro="collection" prefix=", coll. « " suffix=" »."/>
+<text variable="collection-title" prefix=", coll. « " suffix=" »"/>
```
adam3smith: ditto

adam3smith:
You can update your pull request by using the pencil icon here: https://github.com/FrancoisC37/styles/blob/patch-7/documents-d-archeologie-francaise.csl
Let us know when you're done.

rmzelle:
Thanks!

PR #1639, opened 2015-08-02
Update the-auk.csl

zuphilip:
Asked https://forums.zotero.org/discussion/26002/style-request-the-auk/#Item_12 . There were more differences in the implementation and the updated style guide, e.g. doi, editors label in long, et al.


adam3smith:
great, thanks

PR #1641, opened 2015-08-03
Update onderstepoort-journal-of-veterinary-research.csl

zuphilip:
This make interpunctation more stable. Created a large group with `delimiter=, " and delete the now unnecessary prefix/suffix.

Ref https://forums.zotero.org/discussion/50936/replace-comma-with-fullstop-period-if-no-url/#Item_2


adam3smith:
that looks great, thanks!

PR #1167, opened 2014-10-08
add new T&F dependent styles and style updates

adam3smith:

PR #1644, opened 2015-08-06
Added a new style for Elsevier's 'SoftwareX' journal.

StuntsPT:
It didn't exist and I had need of it, so I created it.
Might as well help others too.


adam3smith:
thanks. This looks like it might be the same as
Elsevier (numeric, without titles): https://www.zotero.org/styles?q=Elsevier%20%28numeric%2C%20without%20titles%29

StuntsPT:
Indeed, it seems exactly the same. Shame I couldn't find it before.
What's your policy on duplicates? (with different names, of course).

adam3smith:
we have dependent styles for that -- essentially links to the main style with the name of the journal:
http://docs.citationstyles.org/en/latest/specification.html#dependent-styles
I can take care of that, thanks again for submitting! (leaving this open until I have the dependent style up)

StuntsPT:
Please take a look at https://github.com/citation-style-language/utilities/pull/23
Is this what you meant?

adam3smith:
the original, independent style is still in this PR (that's why the test fails).

StuntsPT:
Sorry about that, forgot to `git rm` it.

adam3smith:
great, thanks!
If you're interested in doing more -- we have stuff to do at (almost) every level of technical difficulty -- ranging from pretty demanding tasks like working on the visual editor or helping to create a style repository (like the one at zotero.org/styles for citationstyles.org) to wrangling with other dependent style lists like https://github.com/citation-style-language/styles/pull/1167 to just helping with style requests by other users (for which there's a long queue right now). Just let us know if you're interested and if so in what and we can set you up.

StuntsPT:
Cool!
Let me take a look at what's on the TODO list and I'll let you know how I can help with.

rmzelle:
Do you prefer Ruby or JavaScript? :)

StuntsPT:
Python. xD
I've done some work on JS and didn't really like it.
I've played a bit with ruby too (just the tutorials), and I did like it better than JS.
So I guess Ruby it is. =-)

StuntsPT:
I've been browsing the issues in both repositories and I didn't find anything I can really help with.
I'll try to do some follow up on #1167 as suggested.
I also find the creation of a new front-end quite important but I'm no expert in that area. I can probably contribute with some suggestions, though.
I'll see what I can do in the next few days to get my feet wet. =-)
I'll be in touch in the relevant issues, though,

rmzelle:
Sorry I didn't get back to you sooner. Been too busy.

A relatively easy target for development would be https://github.com/citation-style-language/utilities/blob/master/generate_dependent_styles/generate_styles_from_data.rb. There are a bunch of changes I'd like to make, but haven't found the time yet. From the top of my head, it would be good if the script:

* would be refactored a bit so it's not just a linear script
* would have an option to place all generated styles into a single output directory instead of publisher-specific ones
* would be smarter about processing the various data files. E.g. ideally, a warning should be given for journals in the skip list that aren't generated anyway. It would also be good to detect duplicates across data sets.
* all input files should properly support blank lines and comments ("# ..."). Some do, but it's a bit of a hack right now.

PR #1645, opened 2015-08-06
Update british-ecological-society.csl

icalandra:
There were lots of brackets and commas for the editors in an edited book. I tried to fix it (lines 136-142); hopefully I did not screw up with groups!
There was also a problem with the prefix and suffix of collection title (line 133). It was not separated from book title (so I added a comma as prefix) and it ended with a period although the editor list comes right after in brackets.

### DIFF
--- a/british-ecological-society.csl
+++ b/british-ecological-society.csl
@@ -20,7 +20,7 @@
     </contributor>
     <category citation-format="author-date"/>
     <category field="biology"/>
-    <updated>2013-04-05T01:56:49+00:00</updated>
+    <updated>2014-09-06T22:02:33+00:00</updated>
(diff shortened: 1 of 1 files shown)


adam3smith:
looks good, thanks!. (You can, btw., delete empty affixes such as ```suffix=""```` entirely for easier readability.

icalandra:
You're right, but I thought about it after pulling the request...
By the way, just to make sure: the style for Functional Ecology is linked to this one, right?

rmzelle:
> By the way, just to make sure: the style for Functional Ecology is linked to this one, right?

Yes, see the "independent-parent" link in https://github.com/citation-style-language/styles/blob/master/dependent/functional-ecology.csl.

PR #1648, opened 2015-08-07
New style for Svensk Exegetisk Arsbok

carljohanberglund:
CSL style for Svensk Exegetisk Arsbok, ISSN 1100-2298

### DIFF
--- a/svensk-exegetisk-arsbok.csl
+++ b/svensk-exegetisk-arsbok.csl
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" page-range-format="chicago" demote-non-dropping-particle="sort-only" default-locale="sv-SE">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" page-range-format="chicago" demote-non-dropping-particle="sort-only">
   <info>
-    <title>Svensk Exegetisk Årsbok</title>
+    <title>Svensk exegetisk årsbok (full note)</title>
     <title-short>SEÅ</title-short>
     <id>http://www.zotero.org/styles/svensk-exegetisk-arsbok</id>
     <link href="http://www.zotero.org/styles/svensk-exegetisk-arsbok" rel="self"/>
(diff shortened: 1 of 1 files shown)


svensk-exegetisk-arsbok.csl
```diff
@@ -0,0 +1,782 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" page-range-format="chicago" demote-non-dropping-particle="sort-only" default-locale="sv-SE">
```
adam3smith: given that they accept manuscripts in "Swedish, Danish, Norwegian, English, German and French." wouldn't it make more sense to remove the default locale?

adam3smith:
It looks like this has an ISSN, right? 	1100-2298? Could we add this?

adam3smith:
Otherwise this looks great.

svensk-exegetisk-arsbok.csl
```diff
@@ -0,0 +1,782 @@
...
+    <link href="http://www.zotero.org/styles/svensk-exegetisk-arsbok" rel="self"/>
+    <link href="http://www.zotero.org/styles/society-of-biblical-literature-fullnote-bibliography" rel="template"/>
+    <link href="http://exegetiskasallskapet.se/SEA/anvisningar.html" rel="documentation"/>
+    <author>
+      <name>Carl Johan Berglund</name>
+      <email>cajo@cajo.nu</email>
+    </author>
+    <category citation-format="note"/>
+    <category field="theology"/>
+    <issn>1100-2298</issn>
```
rmzelle: @adam3smith, the ISSN is already here.

svensk-exegetisk-arsbok.csl
```diff
@@ -0,0 +1,782 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" page-range-format="chicago" demote-non-dropping-particle="sort-only" default-locale="sv-SE">
+  <info>
+    <title>Svensk Exegetisk Årsbok</title>
```
rmzelle: Should the journal title be in title case (which is mostly an English habit)? Maybe "Svensk exegetisk årsbok" is better?

carljohanberglund: It frequently happens with Swedish names of periodicals (as opposed to book titles, which are always in sentence case). In this particular case, however, you are absolutely right!

svensk-exegetisk-arsbok.csl
carljohanberglund: I'm unsure about the function of the default locale, but I am currently unable to get Zotero 4 to produce Swedish footnotes in Word 2011 without this default locale in place. Swedish is the most common language used in this periodical, so it would make sense for a default setting (as opposed to a mandatory one).

adam3smith: at least in Zotero, you can't override the style default. In Zotero 4.27+ you should be able to adjust the language used by any style without a default locale in the document preferences, though.

carljohanberglund:
Default locale removed. Upgrading to Zotero 4.0.28 solved the issue necessitating it.

rmzelle:
Does the journal use multiple styles? If not, it's better to remove the " (full note)" from the title.

carljohanberglund:
Yes, it also uses an author-date format.

rmzelle:
Okay, thanks!

PR #1650, opened 2015-08-08
Create harvard-university-of-worcester.csl

University-of-Worcester:
This is the first commit of our style - version 1.0 - hope all goes well!
- Chris


harvard-university-of-worcester.csl
```diff
@@ -0,0 +1,376 @@
...
+      </if>
+    </choose>
+  </macro>
+  <macro name="italicised-title">
+    <text variable="title" font-style="italic"/>
+  </macro>
+  <macro name="standard-title">
+    <text variable="title"/>
+  </macro>
+  <macro name="publisher-place">
```
adam3smith: Don't solve this with text values -- that's awfully fragile. 
Something like

```
<group delimter=", " suffix=".">
  <text variable="publisher/>
  <text variable="page"/>
</group>
```

Will do the same more elegantly and more robustly.

harvard-university-of-worcester.csl
```diff
@@ -0,0 +1,376 @@
...
+  </macro>
+  <macro name="medium-online">
+    <choose>
+      <if match="any" variable="medium">
+        <text variable="medium" prefix="[" suffix="]"/>
+      </if>
+      <else>
+        <choose>
+          <if match="any" variable="DOI URL">
+            <text value="[Online]"/>
```
adam3smith: lets do ```<text term="online" text-case="capitalize-first" prefix="[" suffix="]"/>``` Same output, but style is going to be better suited to future modifications.

harvard-university-of-worcester.csl
```diff
@@ -0,0 +1,376 @@
...
+            </date>
+          </if>
+          <else>
+            <date date-parts="year" form="text" variable="issued">
+              <date-part name="year" range-delimiter=""/>
+            </date>
+          </else>
+        </choose>
+      </if>
+      <else-if type="bill" match="any">
```
adam3smith: just do ```<else-if type="bill"/>``` for empty output.

harvard-university-of-worcester.csl
```diff
@@ -0,0 +1,376 @@
...
+        <choose>
+          <if match="all" variable="number">
+            <text variable="number" prefix=", episode " suffix="."/>
+          </if>
+        </choose>
+      </if>
+    </choose>
+  </macro>
+  <macro name="single-quote-title">
+    <text variable="title" prefix=" '" suffix="'."/>
```
adam3smith: don't hardcode single quotes -- use ```quotes="true"``` instead.

harvard-university-of-worcester.csl
```diff
@@ -0,0 +1,376 @@
...
+          <date-part name="day"/>
+          <date-part name="month"/>
+        </date>
+      </if>
+    </choose>
+  </macro>
+  <macro name="interviewer">
+    <text value="Interviewed by "/>
+    <names variable="interviewer"/>
+    <text value=" for "/>
```
adam3smith: haven't followed the logic here entirely, but wouldn't it make more sense to pack the "for" into a group delimiter, in case you have an interview without a magazine/TV station?

harvard-university-of-worcester.csl
```diff
@@ -0,0 +1,376 @@
...
+  </macro>
+  <macro name="receipient">
+    <names variable="recipient"/>
+  </macro>
+  <citation collapse="year" disambiguate-add-year-suffix="true" et-al-use-first="3" et-al-min="4">
```
adam3smith: you're sure you don't want any sort order for in-text citations?

harvard-university-of-worcester.csl
```diff
@@ -0,0 +1,376 @@
...
+        <group>
+          <label form="short" variable="locator"/>
+          <text variable="locator"/>
+        </group>
+      </group>
+    </layout>
+  </citation>
+  <bibliography et-al-use-first="4" et-al-min="5" hanging-indent="true">
+    <sort>
+      <key macro="author"/>
```
adam3smith: not sorting by date? That'd be more common for Harvard styles.

harvard-university-of-worcester.csl
```diff
@@ -0,0 +1,376 @@
...
+        <choose>
+          <if match="none" variable="publisher-place">
+            <text macro="event-place"/>
+          </if>
+        </choose>
+        <text macro="event-date"/>
+      </group>
+      <text macro="publisher-place" prefix=" "/>
+      <text macro="pages" prefix=" " suffix="."/>
+      <choose>
```
adam3smith: again, try to find a way -- likely a group -- to solve this without putting punctuation into value fields.

harvard-university-of-worcester.csl
```diff
@@ -0,0 +1,376 @@
...
+  <info>
+    <title>Harvard - University of Worcester</title>
+    <id>http://www.zotero.org/styles/harvard-university-of-worcester</id>
+    <link rel="self" href="http://www.zotero.org/styles/harvard-university-of-worcester"/>
+    <link rel="documentation" href="http://www.worc.ac.uk/ils/documents/Harvard_referencing"/>
+    <author>
+      <name>Christopher Hill</name>
+    </author>
+    <contributor>
+      <name>Christopher Hill</name>
```
adam3smith: remove yourself as contributor -- author is enough.

harvard-university-of-worcester.csl
```diff
@@ -0,0 +1,376 @@
...
+      <if type="webpage interview personal_communication post-weblog" match="any">
+        <date form="text" date-parts="year-month-day" variable="issued">
+          <date-part name="day"/>
+          <date-part name="month"/>
+        </date>
+      </if>
+    </choose>
+  </macro>
+  <macro name="interviewer">
+    <text value="Interviewed by "/>
```
adam3smith: using something like ```<label form="verb" text-case="capitalize-first"/>``` within the ```<names>``` node would be a lot more elegant here.

University-of-Worcester:
Thank you @adam3smith for your comments. Was not expecting such decisive action so quickly! I'll follow up all of them and do a new commit. FYI, I've been using Mendeley's Visual Editor for the most part, perhaps explains some of the vestiges?
Chris

adam3smith:
yeah, the visual editor allows you to do all the things I suggest, but it makes some of them -- especially groups -- a bit unintuitive to use.

harvard-university-of-worcester.csl
University-of-Worcester: Addressed

harvard-university-of-worcester.csl
University-of-Worcester: Addressed

harvard-university-of-worcester.csl
University-of-Worcester: Addressed

harvard-university-of-worcester.csl
University-of-Worcester: Addressed...I think

harvard-university-of-worcester.csl
University-of-Worcester: Addressed

harvard-university-of-worcester.csl
University-of-Worcester: Addressed. Only limitation is you can't (as far as I can tell) declare past tense for the verb.

adam3smith: you can redefine any term at the beginning of the style:

```
<locale>
  <terms> 
    <term name="interviewer" form="verb">interviewed by</term>
  </terms>
</locale>
```

typically placed right below ```</info>``` though it'll work anywhere.

harvard-university-of-worcester.csl
University-of-Worcester: It is a remit for this university to do things differently simply because it can. The guide I am following explicitly states it's ordered by author/editor A-Z so...there we go.

harvard-university-of-worcester.csl
University-of-Worcester: Correct. Addressed

harvard-university-of-worcester.csl
University-of-Worcester: nothing explicit in the guide but one example suggests A-Z by author/editor. Seems sensible to do so.

harvard-university-of-worcester.csl
University-of-Worcester: Addressed

harvard-university-of-worcester.csl
```diff
@@ -0,0 +1,366 @@
...
+        <text macro="anon"/>
+      </substitute>
+    </names>
+  </macro>
+  <macro name="author-short">
+    <names variable="author">
+      <name form="short" and="text" delimiter-precedes-last="never"/>
+      <et-al font-style="normal"/>
+      <substitute>
+        <names variable="translator"/>
```
adam3smith: I didn't see that before, but wouldn't you want the same logic here (for in text citations) that you generate with if/else-if for the first element in the bibliography below? I.e. titles for broadcast, otherwise editors etc.?

Normally (and I'll leave it up to you if you want to do that or not) you'd pack that into ```<substitute>``` for both the author and author-short macro: that way you're sure that they correspond (and in-text citations are easy to match to their bibliography counterparts).

adam3smith:
I have one more comment in line above -- otherwise all good from my side. Let us know when you're ready for this to be merged.

harvard-university-of-worcester.csl
University-of-Worcester: I've followed your advice and wrapped all that conditional logic into a substitute-author macro to use in both bibliography and in-text citations. Should maintain consistency better that way.

University-of-Worcester:
Ok, if everything looks good, this is ready to merge. Thank you for all your assistance @adam3smith.

harvard-university-of-worcester.csl
adam3smith: you left ```<names variable="translator"/>``` here in the author-short macro but don't have it in author (and I'd guess you don't want it). Is that an oversight? Otherwise this looks ready to merge.

University-of-Worcester: Indeed, it was initially an oversight, however, upon re-consideration, we'd decided to surface the translator for the in-text citation (if there is one) but keep it an exclusive field in the bibliography.

University-of-Worcester:
Just to follow up that last comment; some of what we have regarding in-text citations *may* change in future but for now the guidelines do not specify how all the different source types are to be presented in the same meticulous detail as they do for bibliography references (as you may have gathered from their lack of implementation initially). I agree that there needs to be some more discussion and guidance on this as more students (hopefully) take up usage of CSL and relevant software. Although there's also issue of what makes a reputable source; a discussion we like like to have with students as well.

Apologies if all this is inappropriate for a pull request(!) but as someone who clearly takes keen interest in the quality of styles, @adam3smith, I thought you'd benefit from a bit more honest background information on where we're at with this internally :) Your input has been invaluable.

Chris

adam3smith:
thanks!

PR #1651, opened 2015-08-10
New style for Svensk exegetisk årsbok

carljohanberglund:
Now with correct capitalization.


rmzelle:
Do you mind fixing this in your original pull request (#1648)? That way we keep everything together. Just use the pencil/edit icon at https://github.com/carljohanberglund/styles/blob/patch-3/svensk-exegetisk-arsbok.csl.

And you would like to fix the style to Swedish, even though the journal accepts manuscript in other languages?

PR #1653, opened 2015-08-14
Create international-journal-of-food-science-and-technology.csl

libora:
Discussion on Zotero forum: https://forums.zotero.org/discussion/51089/style-requestinternational-journal-of-food-science-and-technology

### DIFF
--- a/international-journal-of-food-science-and-technology.csl
+++ b/international-journal-of-food-science-and-technology.csl
@@ -4,6 +4,7 @@
     <title>International Journal of Food Science &amp; Technology</title>
     <id>http://www.zotero.org/styles/international-journal-of-food-science-and-technology</id>
     <link href="http://www.zotero.org/styles/international-journal-of-food-science-and-technology" rel="self"/>
+    <link href="http://www.zotero.org/styles/ecological-entomology" rel="template"/>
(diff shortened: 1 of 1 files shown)


international-journal-of-food-science-and-technology.csl
```diff
@@ -0,0 +1,226 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" default-locale="en-US" version="1.0">
+  <info>
+    <title>International Journal of Food Science &amp; Technology</title>
+    <id>http://www.zotero.org/styles/international-journal-of-food-science-and-technology</id>
+    <link href="http://www.zotero.org/styles/international-journal-of-food-science-and-technology" rel="self"/>
```
adam3smith: we really want the template links in the styles where they apply and the "BY" part of the style does make them, technically, a legal necessity. So in this case, please add
```<link href="http://www.zotero.org/styles/ecological-entomology" rel="template"/>``` and do remember keeping a note or the original style for future contribution. FWIW, I simply copy the original ```<link  rel="self>``` line and change self to template as I start editing a style. That way I never forget.

adam3smith:
Looks good otherwise. Thanks!

rmzelle:
Great, thanks!

PR #1656, opened 2015-08-20
Web sites & Sorting thirdly by title

DanisKiziridis:
The web site bibliographic entry format wasn't in agreement with what the guidelines suggest:
Format from journal's guidelines: Authorship or Source. (Year). Title of web document or web page (italics). Available at: [URL]. Last accessed DD MONTH YYYY. 
Example from journal's guidelines: National Science Foundation (2010). Ocean acidification (italics). Available at: http://www.nsf.gov/pubs/2010/nsf10530/nsf10530.htm. Last accessed 7 September 2010. 

In the case of same authors in same year, the style wasn't sorting by title.

### DIFF
--- a/ecology-letters.csl
+++ b/ecology-letters.csl
@@ -17,7 +17,7 @@
     <category field="biology"/>
     <issn>1461-023X</issn>
     <eissn>1461-0248</eissn>
-    <updated>2013-02-20T13:28:44+00:00</updated>
+    <updated>2015-05-20T08:52:23+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="container">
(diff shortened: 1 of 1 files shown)


ecology-letters.csl
```diff
@@ -73,9 +73,15 @@
     <choose>
       <if type="webpage">
         <group>
-          <text value="URL" suffix=" "/>
-          <text variable="URL"/>
+          <text value="Available at:" suffix=" "/>
```
adam3smith: Could you change this to 
```<text term="available at" text-case="capitalize-first" suffix=": "/>```
I'm trying to avoid using ```text value``` as much as possible.

File: ecology-letters.csl [hunk 2]
```diff
@@ -207,9 +212,10 @@
     <sort>
       <key macro="author"/>
       <key macro="issued" sort="ascending"/>
+      <key macro="title"/>
     </sort>
     <layout>
-      <group display="block">
+      <group display="block">
```
adam3smith: remove the stray space you added here at the end of the line).

adam3smith:
Looking great otherwise, thanks. You can update by using the pencil icon here: https://github.com/DanisKiziridis/styles/blob/patch-1/ecology-letters.csl
Let us know when you're done.

DanisKiziridis:
The 2 edits done, thanks for the comments!

rmzelle:
Great, thanks!

PR #1660, opened 2015-08-25
Update apa.csl to italicize book titles

drackham:
Book titles should be italicized in APA 6th.  See: https://owl.english.purdue.edu/owl/resource/560/08/

### DIFF
--- a/apa.csl
+++ b/apa.csl
@@ -184,7 +184,7 @@
         <choose>
           <if variable="version" type="book" match="all">
             <!---This is a hack until we have a computer program type -->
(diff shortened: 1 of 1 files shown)


adam3smith:
You're misunderstanding the code: that code only evaluates to true when an item type book has  a "version" variable--which would only be the case for software, for which we don't have yet have a proper item type. 
For all other books (i.e. all actual books), line 190 applies and book titles are italicized. If that doesn't work for you, contact support for your reference manager of choice. The CSL style is correct.

PR #1662, opened 2015-08-26
Added type for website, blogpost and podcast

rm2342:
(as part of "article-newspaper")

### DIFF
--- a/neue-juristische-wochenschrift.csl
+++ b/neue-juristische-wochenschrift.csl
@@ -109,7 +109,7 @@
             <text macro="firstpage-locator"/>
           </group>
         </if>
-        <else-if type="article-newspaper post-weblog webpage song">
+        <else-if type="article-newspaper post-weblog webpage song" match="any">
           <text macro="author-note" suffix=", "/>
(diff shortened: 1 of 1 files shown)


neue-juristische-wochenschrift.csl
```diff
@@ -109,7 +109,7 @@
             <text macro="firstpage-locator"/>
           </group>
         </if>
-        <else-if type="article-newspaper">
+        <else-if type="article-newspaper post-weblog webpage song">
```
adam3smith: this needs match="any" (if/else-if default to match="all" which would never evaluate to true in this case.)

rm2342: Thank you. Funnily, it worked fine without the match="any". I added it
anyway, of course.

adam3smith:
Thanks. Yes, Zotero may still be too forgiving here, but other CSL implementations (e.g. pandoc's) will stumble over this.

rm2342:
Thank you. I will update my other styles in the same manner in the near
future (wanted to run one style through the process first).

PR #1664, opened 2015-08-26
Create the_biological_bulletin.csl

nicolewebster:
handmade style to fit the Biological Bulletin journal literature cited requirements (http://www.biolbull.org/site/misc/ifora.xhtml#references)

### DIFF
--- a/the_biological_bulletin.csl
+++ b/the_biological_bulletin.csl
@@ -6,15 +6,12 @@
     <title-short>Biol.Bull.</title-short>
     <id>http://www.zotero.org/styles/the-biological-bulletin</id>
     <link href="http://www.zotero.org/styles/the-biological-bulletin" rel="self"/>
-    <link href="http://www.zotero.org/styles/journal-of-evolutionary-biology" rel="template"/>
-    <link href="http://www.evolutionary-ecology.com/advice/AdviceFrames.html" rel="documentation"/>
     <author>
-      <name>Alex Chubaty</name>
-      <email>achubaty@sfu.ca</email>
+      <name>Nicole Webster</name>
     </author>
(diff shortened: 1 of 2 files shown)


adam3smith:
The only thing wrong is the filename -- it should be the-biological-bulletin.csl (hyphens, not underscores).

We'd also ask you to put the template and documentation link back in, please. Looks great otherwise.

the_biological_bulletin.csl
```diff
@@ -0,0 +1,161 @@
...
+    <title>The Biological Bulletin</title>
+    <title-short>Biol.Bull.</title-short>
+    <id>http://www.zotero.org/styles/the-biological-bulletin</id>
+    <link href="http://www.zotero.org/styles/the-biological-bulletin" rel="self"/>
+    <author>
+      <name>Nicole Webster</name>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="biology"/>
+    <issnl>1939-8697</issnl>
```
adam3smith: oh, also we don't really use issnl -- just go with issn (or eissn, whichever applies).

rmzelle: Yes, please use:

```
<issn>0006-3185</issn>
<eissn>1939-8697</eissn>
```

ISSN-L wasn't a bad idea, but it hasn't really seen much uptake, and most publishers only show ISSN and eISSN, so we focus on those two as well.

rmzelle:
Sorry, but can you remove the extra quote mark that you just introduced?

`<link href=""http://www.biolbull.org/site/misc/ifora.xhtml#references" rel="documentation"/>` should be `<link href="http://www.biolbull.org/site/misc/ifora.xhtml#references" rel="documentation"/>`.

PR #1665, opened 2015-08-27
Changed Bibliography "hanging-indent" to "false"

University-of-Worcester:
Changing this bibliography property as it is undesirable to have a hanging indent in our style. (This was left over from a starter-style and not picked up using the Visual Editor)


adam3smith:
thanks!

PR #1666, opened 2015-08-28
reviews-of-modern-physics-with-titles Changes

serpodrick:
Hi

These changes make reviews-of-modern-physics-with-titles do something sensible for article, review and review-book, which were previously ignored.

Ser

### DIFF
--- a/reviews-of-modern-physics-with-titles.csl
+++ b/reviews-of-modern-physics-with-titles.csl
@@ -201,7 +201,7 @@
   </macro>
   <macro name="locators">
     <choose>
-      <if type="article-journal">
+      <if type="article-journal review review-book" match="any">
         <text variable="volume" prefix=" " font-weight="bold"/>
       </if>
       <else-if type="legal_case">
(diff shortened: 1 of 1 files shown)


reviews-of-modern-physics-with-titles.csl
```diff
@@ -243,7 +243,7 @@
           </group>
         </group>
       </if>
-      <else-if type="article-journal article-magazine article-newspaper" match="any">
+      <else-if type="article-journal article-magazine article review review-book" match="any">
```
adam3smith: why are we loosing newspaper articles here? I'd expect page numbers to be printed where they exist, no?

serpodrick: The if node for article-newspaper starts at line 234 and ends at line 245. If the test for the if node is true, it doesn't look at the else-if nodes, right?

adam3smith: absolutely; so this was there by mistake before. Thanks.

PR #1668, opened 2015-08-31
Update journal-of-separation-science.csl

zuphilip:
cf. https://forums.zotero.org/discussion/27948/style-request-journal-of-separation-science-jss/#Item_10


adam3smith:
thanks!

PR #1669, opened 2015-08-31
Create these-de-medecine-paris-xi.csl

Jishin42:
New style added for French Medecine Thesis, based on http://www.zotero.org/styles/medecine-sciences.

### DIFF
--- a/these-de-medecine-paris-xi.csl
+++ b/these-de-medecine-paris-xi.csl
@@ -14,8 +14,6 @@
     <category citation-format="numeric"/>
     <category field="medicine"/>
     <category field="science"/>
-    <issn>0767-0974</issn>
-    <eissn>1958-5381</eissn>
     <summary>Style pour une these de medecine Paris XI</summary>
     <updated>2015-08-31T21:20:00+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
--- a/these-de-medecine-paris-xi.csl
+++ b/these-de-medecine-paris-xi.csl
@@ -14,7 +14,7 @@
     <category citation-format="numeric"/>
     <category field="medicine"/>
     <category field="science"/>
-    <summary>Style pour une these de medecine Paris XI</summary>
+    <summary>Style pour une these de medecine Paris XI (Universite Paris Sud)</summary>
     <updated>2015-08-31T21:20:00+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
@@ -28,7 +28,6 @@
     <names variable="author" suffix=".">
       <name name-as-sort-order="all" sort-separator=" " initialize-with="" delimiter=", " delimiter-precedes-last="always"/>
       <label form="short" prefix=" "/>
-      <et-al />
       <substitute>
         <names variable="editor"/>
         <text variable="title"/>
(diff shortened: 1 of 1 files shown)


these-de-medecine-paris-xi.csl
```diff
@@ -0,0 +1,141 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" default-locale="fr-FR">
+<!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>These de medecine Paris XI</title>
+    <id>http://www.zotero.org/styles/these-de-medecine-paris-xi</id>
+    <link href="http://www.zotero.org/styles/these-de-medecine-paris-xi" rel="self"/>
+    <link href="http://www.zotero.org/styles/medecine-sciences" rel="template"/>
+    <link href="http://www.zotero.org/styles/plos" rel="template"/>
+    <link href="http://www.medecinesciences.org" rel="documentation"/>
```
adam3smith: is the style for Paris XI documented anywhere? We really would like online documentation, especially for styles that don't correspond to any easily accessible publications like journal styles.

these-de-medecine-paris-xi.csl
```diff
@@ -0,0 +1,141 @@
...
+    <names variable="editor" delimiter=", ">
+      <name initialize-with="" name-as-sort-order="all" sort-separator=" " delimiter=", " delimiter-precedes-last="always"/>
+      <label form="long" prefix=", "/>
+    </names>
+  </macro>
+  <macro name="author">
+    <names variable="author" suffix=".">
+      <name name-as-sort-order="all" sort-separator=" " initialize-with="" delimiter=", " delimiter-precedes-last="always"/>
+      <label form="short" prefix=" "/>
+      <et-al />
```
adam3smith: delete this line -- without attribute, et-al doesn't do anything.

these-de-medecine-paris-xi.csl
```diff
@@ -0,0 +1,141 @@
...
+		<else>
+          <text macro="title" prefix=" " suffix="."/>
+          <text variable="container-title" form="short"  strip-periods="true" prefix=" " suffix="."/>
+          <text macro="year-date" prefix="" suffix=";"/>
+          <choose>
+		    <if match="none" variable="volume">
+              <text variable="DOI" prefix="doi:"/>
+            </if>
+            <else>
+			  <text variable="volume" prefix="" suffix=""/>
```
adam3smith: remove empty prefix/suffix

adam3smith:
Otherwise this looks fine. My my concern is actually the documentation.

these-de-medecine-paris-xi.csl
rmzelle: It also looks like Paris XI is better known as "Université Paris-Sud".

Jishin42: The only documentation I have is a document given by the university. It is in French.
I uploaded it on a personal cloud (http://ovh.to/hCBLzv6) but is there a common repository for this kind of document ? Should I commit it in this repository ?

adam3smith: as long as you can keep the link to the document on the cloud permanent, that's great. We'd rather not host style manuals here, no.

rmzelle:
We need to come up with a better name for this style, as well. Do I understand correctly that this citation style's scope is limited to the Diplôme d'Etudes Spécialisées (DES) de Psychiatrie of the Université Paris-Sud?

rmzelle:
We're still willing to accept this once we figure out a good name for the style.

Closing now because of inactivity (if we hear back from you we can reopen).

PR #1674, opened 2015-09-01
This style is the citation style of the german journal "Im Gespräch - Hefte der Martin Buber-Gesellschaft". 

mgorki:


adam3smith:
Here's why the check fails:
1. There are a [range of validation](http://validator.citationstyles.org/?url=https%3A%2F%2Fgithub.com%2Fmgorki%2Fstyles%2Fraw%2Fpatch-1%2Fim-gespraech-buber-gesellschaft.csl&version=1.0.1#source-code) issues, I'll go through them inline.
2. You use a macro "contributors-short" that you don't define
3. You define macros "date", "number-of-volumes" that you don't use

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,651 @@
...
+      <name>Michael Gorki</name>
+      <email>michael.gorki@gmx.de</email>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="humanities"/>
+    <issn>1616-6094</issn>
+    <updated>2015-09-01T13:28:08+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="de">
```
adam3smith: why are you defining all of these terms in the style? They're part of CSL, you only need to redefine the ones that differ from https://github.com/citation-style-language/locales/blob/master/locales-de-DE.xml

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,651 @@
...
+      <name delimiter="/" delimiter-precedes-last="always"/>
+    </names>
+  </macro>
+  <macro name="volume">
+    <group suffix=".">
+      <text term="volume" form="short" suffix=" "/>
+      <text variable="volume"/>
+    </group>
+  </macro>
+  <macro name="Band aus Reihe">
```
adam3smith: you can't have spaces in macro names. I'd suggest ```band-aus-reihe```

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,651 @@
...
+                  <text value=". "/>
+                </if>
+              </choose>
+            </group>
+            <text macro="collection-editor"/>
+            <text variable="volume" prefix=", Bd. "/>
+          </group>
+        </group>
+      </if>
+      <if match="none" variable="title">
```
adam3smith: you can't have more than one ```if```` condition per macro. In this case, since the two are exhaustive, I'd just go with ```<else>``` here.

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,651 @@
...
+    <choose>
+      <if match="any" variable="title">
+        <group>
+          <text macro="editor" prefix=" "/>
+          <group prefix=" (" suffix=") ">
+            <group>
+              <text variable="collection-title"/>
+              <choose>
+                <if match="any" variable="collection-editor">
+                  <text value=". "/>
```
adam3smith: don't put punctuation in text values, that's ugly and fragile. Instead, use a group delimiter:

```
<group delimiter=". ">
  <text macro="collection-editor"/>
</group>
```

im-gespraech-buber-gesellschaft.csl
mgorki: Thank your very much for your kind help! I'm going to fix this error immediately.

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,651 @@
...
+  </macro>
+  <macro name="container-title">
+    <choose>
+      <if match="any" variable="container-title">
+        <text macro="editor" prefix=" "/>
+        <group prefix=" (" suffix=") ">
+          <group>
+            <text variable="collection-title"/>
+            <choose>
+              <if match="any" variable="collection-editor">
```
adam3smith: same here.

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,651 @@
...
+      <if match="any" variable="edition">
+        <text variable="edition" prefix=" " suffix=". "/>
+        <label plural="never" variable="edition"/>
+      </if>
+    </choose>
+  </macro>
+  <macro name="container-author">
+    <choose>
+      <if match="any" variable="container-author">
+        <text value="Ders.: "/>
```
adam3smith: I don't think this'll work, CSL can't handle ders.
Container author would typically be the author of a book where you're citing e.g. the introduction by someone else. E.g. in Nabokov, Vladimir, "Einleitung," in Kafka, Franz, *Die Metamorphose,* ...
Kafka is the container-author.

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,651 @@
...
+      </if>
+    </choose>
+  </macro>
+  <macro name="number-of-volumes">
+    <text variable="number-of-volumes" prefix=" von "/>
```
adam3smith: I don't think you want the "von" here: You'll frequently have items with number-of-volumes but no specific volume number (e.g. when citing a multivolume work in its entirety).

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,651 @@
...
+          <text macro="citation-locator"/>
+        </if>
+        <else-if match="any" position="ibid">
+          <text term="ibid"/>
+        </else-if>
+        <else>
+          <group>
+            <choose>
+              <if match="any" variable="title-short">
+                <text variable="title-short"/>
```
adam3smith: this seems odd -- what are you trying to test here? You're aware that you can test for position="subsequent"?

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,651 @@
...
+                  <choose>
+                    <if type="webpage" match="all">
+                      <choose>
+                        <if match="any" variable="issued">
+                          <date form="text" date-parts="year-month-day" variable="issued">
+                            <date-part name="year"/>
+                          </date>
+                        </if>
+                        <else>
+                          <choose>
```
adam3smith: needlessly complicated. Just continue with: ```<else-if variable="accessed">``` instead of starting a new choose element.

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,651 @@
...
+  </citation>
+  <bibliography et-al-min="9" et-al-use-first="8" subsequent-author-substitute="-" subsequent-author-substitute-rule="complete-each" hanging-indent="true">
+    <sort>
+      <key macro="author"/>
+      <key macro="year-date"/>
+      <key variable="title"/>
+    </sort>
+    <layout suffix=".">
+      <choose>
+        <if position="ibid-with-locator">
```
adam3smith: I'm surprised validation doesn't fail here, but you can't test for position in the bibliography.

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,651 @@
...
+              <if match="any" variable="collection-editor">
+                <text value=". "/>
+              </if>
+            </choose>
+          </group>
+          <text macro="collection-editor"/>
+          <text variable="volume" prefix=", Bd. "/>
+        </group>
+      </if>
+      <if match="none" variable="container-title">
```
adam3smith: again--can't have two consecutive ifs. Not sure what you're trying to do here, though.

mgorki:
I fixed all the essential errors. Is there any option in csl to implement the "Ders." (the same author) more acurately? It would be very nice to have something like "Gershom Scholem: Martin Bubers Auffassung des Judentums. In: Ders.: Judaica Bd. 2. Frankfurt a.M.: Suhrkamp 5. Auflage 1995, S. 133–192." and not: "Gershom Scholem: Martin Bubers Auffassung des Judentums. In: Gershom Scholem: Judaica Bd. 2. Frankfurt a.M.: Suhrkamp 5. Auflage 1995, S. 133–192.

zuphilip:
No, that is not possible, see https://www.zotero.org/support/kb/idem . In German this may be even more complicated because it may also be "Dies." instead of "Ders." if it is a women (depending on the citation rules).

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,426 @@
...
+        <text term="no date" form="short"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="pages">
+    <choose>
+      <if type="article-journal article-magazine article-newspaper broadcast thesis webpage chapter paper-conference" match="any">
+        <label strip-periods="true" prefix=", " suffix=". " variable="page" form="short"/>
+        <text variable="page"/>
+        <text variable="note" prefix=" [erstmals " suffix="]"/>
```
adam3smith: what is this for? I don't like using the note variable for workarounds. CSL is used by many different programs, "note" is used somewhat differently by most of them.

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,426 @@
...
+        <label variable="locator" form="short" suffix=". " strip-periods="true"/>
+        <text variable="page"/>
+        <text variable="note" prefix=" [erstmals " suffix="]"/>
+      </else-if>
+    </choose>
+  </macro>
+  <macro name="citation-locator">
+    <choose>
+      <if type="manuscript" match="none">
+        <label variable="locator" form="short" suffix=". " strip-periods="true"/>
```
adam3smith: remove strip-periods="true" and the . in the suffix -- they're redundant.

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,426 @@
...
+  <macro name="date">
+    <choose>
+      <if match="any" variable="issued">
+        <date form="text" variable="issued"/>
+      </if>
+    </choose>
+  </macro>
+  <macro name="collection-editor">
+    <names variable="collection-editor" prefix="Hrsg. von ">
+      <name delimiter="/" delimiter-precedes-last="always"/>
```
adam3smith: I'd much rather have Hrsg. von treated as ```<label form="verb-short"/>```

im-gespraech-buber-gesellschaft.csl
adam3smith: still would like this fixed. Let me know if you have questions.

im-gespraech-buber-gesellschaft.csl
adam3smith: this still seems wrong to me. Could you explain? (Deutsch ist OK, wenns einfacher ist)

im-gespraech-buber-gesellschaft.csl
adam3smith: still needs to be cleaned up.

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,426 @@
...
+            </group>
+          </group>
+        </else>
+      </choose>
+    </layout>
+  </citation>
+  <bibliography et-al-min="9" et-al-use-first="8" subsequent-author-substitute="-" subsequent-author-substitute-rule="complete-each" hanging-indent="true">
+    <layout suffix=".">
+      <choose>
+        <if position="ibid-with-locator">
```
adam3smith: you still have the position tests here. They don't do anything in the bibliography. Remove.

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,426 @@
...
+            <text variable="original-title" prefix=" (" suffix=". "/>
+            <text value="LAND" suffix=" "/>
+            <date date-parts="year" form="text" variable="issued" suffix=", "/>
+            <names variable="director" prefix="R: " suffix=")"/>
+          </group>
+        </else-if>
+        <else-if type="broadcast" match="any">
+          <group>
+            <text variable="title" font-style="italic"/>
+            <text value="ORIGINAL-TITEL" prefix=" (" suffix=". "/>
```
adam3smith: why not use variable here? I don't think we want placeholders in styles.

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,426 @@
...
+            <text variable="title" font-style="normal" suffix=". "/>
+            <text variable="container-title" font-style="italic" prefix="in: "/>
+            <text macro="year-date"/>
+            <text macro="pages" prefix=", "/>
+          </group>
+        </else-if>
+        <else-if type="motion_picture" match="any">
+          <group>
+            <text variable="title" font-style="italic"/>
+            <text variable="original-title" prefix=" (" suffix=". "/>
```
adam3smith: If you want to place parentheses around a group of items, set them as affixes on a group. The way you have this now, if you have a film w/o director or original titles, you'll have unbalanced parentheses.

im-gespraech-buber-gesellschaft.csl
mgorki: I'm using the note for the original publication date. As far as I know there is no other option to include it, is there?

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,426 @@
...
+        </else-if>
+        <else-if type="manuscript" match="any">
+          <group>
+            <text macro="author"/>
+            <group>
+              <choose>
+                <if match="any" variable="title">
+                  <text variable="title" suffix="."/>
+                  <date form="text" variable="issued" prefix=" "/>
+                  <text value="."/>
```
adam3smith: again, no punctuation in text value. Normally set as group affix or delimiter.

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,426 @@
...
+                </if>
+              </choose>
+              <text macro="archive" prefix=" "/>
+            </group>
+            <choose>
+              <if match="none" variable="archive_location">
+                <text macro="citation-locator" prefix=", "/>
```
adam3smith: citation-locator makes no sense in the bibliography. Remove.

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,426 @@
...
+      <label form="short" prefix=" (" suffix=")" text-case="capitalize-first"/>
+      <substitute>
+        <names variable="editor" suffix=": "/>
+        <text macro="anon"/>
+      </substitute>
+    </names>
+  </macro>
+  <macro name="author-short">
+    <names variable="author">
+      <name form="short" delimiter="/" delimiter-precedes-last="always" et-al-min="4" et-al-use-first="2" initialize-with=". "/>
```
adam3smith: couldn't you just set the et al.  in citation and be done with it rather than doing it three times?
Also, sure about the 2? (Smith/Meyer et al. 1999) is certainly possible but seems odd.

adam3smith:
Thanks!
A bunch more comments inline. Note that the automatic checks are just minimum formal requirements, we do require that you fix any issues we raise during manual review before we can accept the style.

im-gespraech-buber-gesellschaft.csl
adam3smith: sure. ```<text variable="original-date"/>```
in Zotero and Mendeley you can enter this by putting ``` {:original-date: 1876}``` into the Extra (Mendeley: Note) field.

im-gespraech-buber-gesellschaft.csl
mgorki: You are completely right. I had not thought about this.

im-gespraech-buber-gesellschaft.csl
mgorki: I think this would be a good idea in most cases. In those cases however were there is no author but an editor it would produce strange results. By now in such a case the result would be "Wolfgang Röllig (Hrsg.): Das Gilgamesch-Epos. (Reclam-Bibliothek, Bd. 10702) Stuttgart: Reclam 2009.". If I treat "Hrsg. von" as <label form="verb-short"/> the result would be "Wolfgang Röllig (Hrsg. von): Das Gilgamesch-Epos. (Reclam-Bibliothek, Bd. 10702) Stuttgart: Reclam 2009.", wouldn't it?

adam3smith: no, not if you just set it here. I.e. instead of putting Hrsg. von in the prefix, do:

```
<names variable="collection-editor">
  <label form="verb-short" suffix=" "/>
  <name delimiter="/" delimiter-precedes-last="always"/>
</names>
```

That won't affect the label for editors any differently than what you have here. It just makes the style more systematic and hardcodes less German into the style so it's easier to internationalize.

im-gespraech-buber-gesellschaft.csl
mgorki: The idea is, that if there is a short title it should be used instead of (Author Year). So for example if the book "Ich und Du" by Martin Buber has its title also as its short title it should be cited inline as "(Ich und Du, S. 27)" instead of "(Buber 1996)".

im-gespraech-buber-gesellschaft.csl
mgorki: You're right. I'm going to change it.

im-gespraech-buber-gesellschaft.csl
mgorki: I know this is not that usual, but the "(Smith/Meyer et al. 1999)" version is intended in this citation style.

mgorki:
Thank you very much for all your comments and Help! I think the code should be a lot better by now.

im-gespraech-buber-gesellschaft.csl
adam3smith: OK, that would do it, but lots of items have short titles. E.g. Zotero will automatically create short titles for any item with a colon, exclamation mark, or question mark in its title. This seems like it'd be very confusing. Is this intended to be just for books by Buber himself? I don't think this would be a suitable solution, then.

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,428 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
+  <info>
+    <title>Im Gespräch - Hefte der Martin Buber-Gesellschaft (German)</title>
+    <title-short>Im Gespräch</title-short>
+    <id>http://www.zotero.org/styles/im-gespraech-buber-gesellschaft</id>
+    <link href="http://www.zotero.org/styles/im-gespraech-buber-gesellschaft" rel="self"/>
+    <author>
```
adam3smith: Since this is pretty unusual in parts -- is this style documented somewhere? Do you have direct contact to Ulrich Gorki or is the last name just a coincidence? We feel pretty strongly about having documentation for our styles available should issues come up. Presumably the journal sends something to authors? Could they put that online?

im-gespraech-buber-gesellschaft.csl
adam3smith: OK, good. First part still holds, though, you can simplify this a lot by moving all the et-al instructions into ```<citation>```.

adam3smith:
Great, almost there. You can see where there are still outstanding comments by scrolling through the "Files Changed" view. Every comment by me should either be incorporated or have a brief note by you explaining why what I'm suggesting won't work for your purposes.

im-gespraech-buber-gesellschaft.csl
mgorki: Not necessarily. There are a lot of "classics" which are normally cited only by their short title. Which there are depends a lot on the individual case. For example in the context of an article where Kant or Heidegger are cited a lot an inline citation should be something like (Sein und Zeit S. 79) oder (KrV S. 516) bzw. (KrV B 516).

adam3smith: OK, but would those appear regularly in the bibliography? It seems odd to have "Sein und Zeit" in the text refer to "Heidegger, Martin (1981): Sein und Zeit" in the bibliography. But if that's indeed what's intended, we can go with it.

im-gespraech-buber-gesellschaft.csl
mgorki: I'm actually his son but also a member of the redaction. There is a word-document we send to the authors. I could convert it to PNG, GIF or JPG an upload it here.

mgorki: aa nachweis-richtlinien ig allgemein 01 2013_s1
aa nachweis-richtlinien ig allgemein 01 2013_s2

This "manual" is from 2013, at the moment we are working on a better manual but there are in the rules themselves only very small changes.

im-gespraech-buber-gesellschaft.csl
mgorki: Thanks for your comment. I fixed it.

im-gespraech-buber-gesellschaft.csl
mgorki: I fixed it.

im-gespraech-buber-gesellschaft.csl
mgorki: You are right, it's a mistake. I fixed it.

im-gespraech-buber-gesellschaft.csl
mgorki: Yes, it's meant to be like this. It's kind of scribal abbreviations.

im-gespraech-buber-gesellschaft.csl
mgorki: Sorry, I do not understand what exactly you are proposing to change. On which coose element are you referring?

im-gespraech-buber-gesellschaft.csl
mgorki: If you like we could actually also send you an exemplar of the last issue.

im-gespraech-buber-gesellschaft.csl
adam3smith: instead of

```
<else>
  <choose>
<if variable="accessed">
```

do 

```
<else-if variable="accessed">
```

(and obviously remove the ```</choose>``` and correct the ```</else>``` to ```</else-if>``` below). It does exactly the same thing&is shorter&easier to read.

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,428 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
+  <info>
+    <title>Im Gespräch - Hefte der Martin Buber-Gesellschaft (German)</title>
+    <title-short>Im Gespräch</title-short>
+    <id>http://www.zotero.org/styles/im-gespraech-buber-gesellschaft</id>
+    <link href="http://www.zotero.org/styles/im-gespraech-buber-gesellschaft" rel="self"/>
```
adam3smith: below is great. You can simply link to your comment as documentation, i.e.
add ```<link href="https://github.com/citation-style-language/styles/pull/1674/files#r38487900" rel="documentation"/>``` right here. If you'd rather, you can also upload this in any other format anywhere else and link to it as long as it's going to be reasonably permanent. The point of this is that we want to have a reference if someone reports an error.

im-gespraech-buber-gesellschaft.csl
adam3smith: this is great. See above.

im-gespraech-buber-gesellschaft.csl
mgorki: Thanks, I think it should be better now.

im-gespraech-buber-gesellschaft.csl
mgorki: It is sure that the style is not that usual. In parts this is due to the purpose of this style to give precise reference in cases where the exact edition of a text matters.

adam3smith: no problem; there are a lot of specialty styles out there. That's why I wanted the documentation; see https://github.com/citation-style-language/styles/pull/1674/files#r38489949 on how to include it in the style.

Apart from that, there are only two remaining small bits ([here](https://github.com/citation-style-language/styles/pull/1674/files#r38465777) and [here](https://github.com/citation-style-language/styles/pull/1674/files#r38431529)) and we're ready to go.

mgorki: Thank you for your comment. I fixed those two redundancies.

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,425 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
```
rmzelle: Assuming this journal only publishes in German, the style should be fixed to this language, which can be done by adding `default-locale="de-DE"` to the `<style>` tag.

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,425 @@
...
+  <info>
+    <title>Im Gespräch - Hefte der Martin Buber-Gesellschaft (German)</title>
```
rmzelle: @adam3smith, how do you feel about having the subtitle in the title here? I usually try to shy away from that.

im-gespraech-buber-gesellschaft.csl
```diff
@@ -0,0 +1,425 @@
...
+    <title-short>Im Gespräch</title-short>
+    <id>http://www.zotero.org/styles/im-gespraech-buber-gesellschaft</id>
```
rmzelle: We simply drop diacritics for the style file name and ID, and rarely use (parts of) subtitles. I much rather see `im-gesprach.csl` than `im-gespraech-buber-gesellschaft.csl`.

im-gespraech-buber-gesellschaft.csl
adam3smith: I think in this context keeping the subtitle makes sense -- it seems pretty important for letting people know what the journal is about.

rmzelle: Okay, fine with me, then. Still would like to see "-buber-gesellschaft" removed from the file name, though.

im-gespraech-buber-gesellschaft.csl
mgorki: I think this is a good idea as the journal - as you correctly asume - publishes almost only in german.

im-gespraech-buber-gesellschaft.csl
mgorki: Or how do would you feel with "im-gespraech-buber.csl"?

rmzelle: No, please make it "im-gesprach.csl".

im-gespraech-buber-gesellschaft.csl
mgorki: So it should be;

  <info>
    <title>Im Gespräch - Hefte der Martin Buber-Gesellschaft (German)</title>
    <title-short>Im Gespräch</title-short>
    <id>http://www.zotero.org/styles/im-gesprach</id>
    <link href="http://www.zotero.org/styles/im-gesprach" rel="self"/> 

correct?

rmzelle: Yes, please. See https://help.github.com/articles/renaming-a-file/ for instructions on how to rename the style. You also will have to adjust the style ID and "self" link.

rmzelle:
Looks good to me. I'll let @adam3smith merge.

adam3smith:
yup -- thanks for your patience going through al of these comments.

mgorki:
Thank you for your comments and the support.

rmzelle:
Thanks for contributing!

PR #1675, opened 2015-09-02

nchachereau:
This is a major update to a custom style for history students in Switzerland. See https://github.com/citation-style-language/styles/pull/234 for the original pull request.

We now have a proper documentation:
- [in French](https://www.infoclio.ch/fr/node/138218)
- [in German](https://www.infoclio.ch/de/node/133932)

Changes include:
- Support for additional item types such as manuscript, personal_communication, and improvements for audio and film types
- URL or DOI are included in the references
- Archival information is included, allowing better citation of primary sources
- Support for original-date


rmzelle:
Great, thanks. Note that the empty lines will eventually get removed when we standardize the indenting of the styles.

PR #1677, opened 2015-09-06
Set "demote-non-dropping-particle" to "display-and-sort"

nickbart1980:
CMS 16e calls for "van den Keere" (in-text) and "Keere, Pieter van den" (index & references) (16.71), and "al-Hakim" (in-text) and "Hakim, Tawfiq al-" (index & references) (16.76). This requires "display-and-sort".

Actually, _all_ Chicago styles, and, most likely, most other styles, too, need to be updated. Could someone batch-edit at least all of the Chicago styles?

### DIFF
--- a/chicago-author-date.csl
+++ b/chicago-author-date.csl
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" page-range-format="chicago">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="display-and-sort" page-range-format="chicago">
   <info>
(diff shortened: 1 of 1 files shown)


nickbart1980:
Just pinging. CMoS quite clearly demands "display-and-sort", and now that particle parsing has been adjusted, there shouldn’t be any problems with this fix.

adam3smith:
I'm trying to track down what prompted this: https://github.com/citation-style-language/styles/commit/77a87152668f06f7570ece9a339e9a0f8bf1544f
but can't find it. But yeah, the manual seems pretty clear on this. It's not, though, a very common rule. Most journals tend to go with "never" here.

rmzelle:
@adam3smith, so, should we just revert https://github.com/citation-style-language/styles/commit/77a87152668f06f7570ece9a339e9a0f8bf1544f ?

adam3smith:
yeah, I suppose. I wasn't able to find the cause for that edit.

PR #1680, opened 2015-09-08
Update the-isme-journal.csl

and3k:
* fixed editor macro: sort-separator is just a blank space, and no "and" symbol
* books never have doi/URLs
* doi’s have a pre-label of "e-pub ahead of print", and there is a space after doi
* disambiguate-add-names and disambiguate-add-givenname should both be false
* fixed master layout, so there is always a period before the access macro
* fixed book chapter layout: editor first, then container/etc, separated by periods

Changes are based on http://www.nature.com/ismej/ismej_new_gta.pdf and the articles from the current ISMEJ.

### DIFF
--- a/the-isme-journal.csl
+++ b/the-isme-journal.csl
@@ -5,6 +5,7 @@
     <id>http://www.zotero.org/styles/the-isme-journal</id>
     <link href="http://www.zotero.org/styles/the-isme-journal" rel="self"/>
     <link href="http://mts-isme.nature.com/cgi-bin/main.plex?form_type=display_auth_instructions" rel="documentation"/>
+    <link href="http://www.nature.com/ismej/ismej_new_gta.pdf" rel="documentation"/>
     <author>
       <name>Jessica Leigh</name>
       <email>blackwednesday@gmail.com</email>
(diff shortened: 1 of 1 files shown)


the-isme-journal.csl
```diff
@@ -5,6 +5,7 @@
     <id>http://www.zotero.org/styles/the-isme-journal</id>
     <link href="http://www.zotero.org/styles/the-isme-journal" rel="self"/>
     <link href="http://mts-isme.nature.com/cgi-bin/main.plex?form_type=display_auth_instructions" rel="documentation"/>
+    <link href="http://www.nature.com/ismej/ismej_new_gta.pdf" rel="documentation"/>
```
rmzelle: If this documentation link supersedes the old one, it's better to just replace the original instead of adding a second one.

and3k: The PDF link is where the actual documentation lives, but based on the filename I’m not confident that this may be not be renamed. The first link looks more stable/permalinky.

rmzelle:
Thanks (sorry for letting this sit for so long!).

PR #1681, opened 2015-09-09

ibadger:
Changes to the style which bring it in line with the Harvard used at UCA. Includes updates to way accessed dates are written, initials when more than one author/editor is cited in the bibliography and mentioned as a director or translator, changes to the format of map references.

Style has been tested through your 'CSL Style and Locale Validator'. Hopefully it is ok


rmzelle:
Thanks! Sorry we never responded to this.

ibadger:
Not a problem, thanks

PR #1682, opened 2015-09-10
minor corrections

jmiba:
added full stop after speech title; changed presented-at term to avoid German grammar (gender) problems with "masculine" conferences; changed case of accessed term to comply with orthography


rmzelle:
Accepted via manual commit, thanks. I just placed the `<locale/>` section directly after `<info/>`, and moved the colon from the term definition to the suffix.

PR #1684, opened 2015-09-13
Create chicago-fullnote-bibliography-linked-doi.csl

jgerth:
Same style but intended for use with web pages where hyperlinks of DOIs are ideal.

### DIFF
--- a/chicago-fullnote-bibliography-linked-doi.csl
+++ b/chicago-fullnote-bibliography-linked-doi.csl
@@ -1,0 +1,1259 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="never" page-range-format="chicago">
+  <info>
+    <title>Chicago Manual of Style 16th edition (full note with linked DOI)</title>
+    <id>http://www.zotero.org/styles/chicago-fullnote-bibliography-linked-doi</id>
+    <link href="http://www.zotero.org/styles/chicago-fullnote-bibliography-linked-doi" rel="self"/>
+    <link href="http://www.chicagomanualofstyle.org/tools_citationguide.html" rel="documentation"/>
+    <link href="http://www.zotero.org/styles/chicago-fullnote-bibliography" rel="template"/>
+    <author>
+      <name>Jordan Gerth</name>
+      <email>jgerth@uwalumni.com</email>
+    </author>
+    <category citation-format="note"/>
+    <category field="generic-base"/>
+    <summary>Chicago format with full notes and bibliography</summary>
+    <updated>2015-09-13T06:45:00+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="en">
+    <terms>
+      <term name="editor" form="verb-short">ed.</term>
+      <term name="translator" form="verb-short">trans.</term>
+      <term name="editortranslator" form="verb-short">
+        <single>ed. and trans.</single>
+        <multiple>ed. and trans.</multiple>
+      </term>
+      <term name="editortranslator" form="verb">
+        <single>Edited and translated by</single>
+        <multiple>Edited and translated by</multiple>
+      </term>
+      <term name="translator" form="short">trans.</term>
+    </terms>
+  </locale>
+  <macro name="editor-translator">
+    <group delimiter=", ">
+      <group delimiter=" ">
+        <choose>
+          <if variable="container-author reviewed-author" match="any">
+            <group>
+              <names variable="container-author reviewed-author">
+                <label form="verb-short" text-case="lowercase" suffix=" "/>
+                <name and="text" delimiter=", "/>
+              </names>
+            </group>
+          </if>
+        </choose>
+      </group>
+      <names variable="editor translator" delimiter=", ">
+        <label form="verb-short" text-case="lowercase" suffix=" "/>
+        <name and="text" delimiter=", "/>
+      </names>
+    </group>
+  </macro>
+  <macro name="secondary-contributors-note">
+    <choose>
+      <if type="chapter paper-conference" match="none">
+        <text macro="editor-translator"/>
+      </if>
+    </choose>
+  </macro>
+  <macro name="container-contributors-note">
+    <choose>
+      <if type="chapter paper-conference" match="any">
+        <text macro="editor-translator"/>
+      </if>
+    </choose>
+  </macro>
+  <macro name="secondary-contributors">
+    <choose>
+      <if type="chapter paper-conference" match="none">
+        <names variable="editor translator" delimiter=". ">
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks, but I don't think we'll take this. We already have a large number of CMoS variants to maintain, and using prefixes and suffixes to create hyperlinks is not a robust solution because CSL processors won't know to properly escape the DOI itself (see e.g. also http://www.doi.org/syntax.html ).

jgerth:
Please provide an example. This solution helps online bibliographies
immensely and the Chicago style is very popular. The convention that I
followed is employed by other styles and conforms to your community
standards. By not accepting this, the needs of the community are not met.

adam3smith:
Our community standards for what kinds of styles we are accepting are [here](https://github.com/citation-style-language/styles/wiki/Criteria-for-Accepting-Styles) and I think this one falls squarely into the category of styles we explicitly say we may not accept. 

We do use this format for APA because it's explicitly recommended by APA. That's not the case for the Chicago manual.

Not sure what you'd want an example for, but no, we're not going to accept this, sorry. Custom-hosted styles are perfectly usable for most purposes, though, now including the Zotero API.

jgerth:
If there is another way to make certain URLs and DOIs as hyperlinks for web use, I would be happy to conform to a different format. I agree that the method currently proposed is not elegant (though we can agree that it is used elsewhere).

Chicago style FAQs give flexibility for online content, so while not explicitly prescribed, it is not in violation of the style either. I reviewed your community standards again and do not see the conflict. This is not a personal use type CSL. It has broad application to anyone who wants their bibliography in a web format, and, as previously mentioned, Chicago style is not a narrow style. As an academic, this is valuable.

Thank you for reconsidering.

adam3smith:
I mean -- it depends on how you're generating your online bibliographies, but it wouldn't be hard to run a routine that simply converts `doi:10.xyzlsdfkj`
into `doi:<a href="http://doi.org/10.xyzlsdfkj">10.xyzlsdfkj</a>` when generating your website. If you generate html with pandoc, I believe it does that automatically, for example. 
In general, all of us at CSL believe that our citation styles shouldn't be in the business of generating bibliographies optimized for the web. 
That's really the job of whichever tool you're using to do the actual publishing/generating of the website with (e.g. I would see this as a very reasonable feature request for ZotPress, if that doesn't already do this). Again, the reason we're doing this for APA is because it's a requirement for their *print* bibliographies.

It's also not something that anyone else has ever requested and as we've said, there's no reason you'll need this accepted to the repository for your personal use. 
Additional styles put a significant burden on the very few long-time volunteers running this project and performing maintenance, so I'm sorry to say, the answer remains no.

jgerth:
Certainly I could manually add the link, but that runs counter to the concept of having a CSL in the first place. I am trying to use this CSL with ZotPress, which requires that the style be part of the community.

I am a volunteer too, willing to contribute. I am not the only user stuck in this predicament of wanting to use a Chicago style for a web page bibliography where DOIs are clickable. But it is true that even for my personal use, I cannot use a personal CSL with ZotPress (that would solve the problem, but that is not the issue before us).

I am extremely disappointed. This type of style is far from something that is “optimized for the web”. Instead, it is a necessity for sharing citations and quickly accessing references in the digital era. If you wish to reject it, that is fine, but you should also update your community standards to clarify your purpose so that new bibliography formats for online use can be handled by a different CSL community. Thanks.

rmzelle:
> Please provide an example.

The page I linked to gives this DOI as example: `http://dx.doi.org/10.1000/456#789` (which should be escaped as `http://dx.doi.org/10.1000/456%23789`, which wouldn't happen with your style)

> I am trying to use this CSL with ZotPress

According to https://wordpress.org/plugins/zotpress/changelog/, ZotPress already automatically hyperlinks DOIs since version 5.2.5.

> I cannot use a personal CSL with ZotPress

The Zotero Web API only gained the ability to use non-repository CSL styles in July (see https://twitter.com/rintzezelle/status/618493401600786432), and it doesn't seem like ZotPress has seen an update since. You might want to put this forward as a feature request.

> that would solve the problem, but that is not the issue before us

Well, it could be the issue before us, if you were simply looking for a way to achieve the desired result for yourself. We're happy to help you with that, but I respectfully disagree on the separate topic of whether adding your style variant to the repository would carry a net benefit.

rmzelle:
I also just found https://wordpress.org/support/topic/links-for-all-dois-chicago-style. I'll ask for clarification on how DOI hyperlinking is supposed to work in ZotPress.

And by the way, I'm not disputing that CSL could benefit from better hyperlinking options. But I don't think we should (widely) adopt the approach of hard-coding escaped HTML tags in CSL styles, since this can result in incorrect URLs as explained above.

adam3smith:
(For future reference, Katie responded to Rintze's post in the ZotPress support thread referenced above. DOI hyperlinking should work in ZotPress and should be doing exactly what I describe above automatically already. If that's not working for you, make sure you're running the latest version of ZotPress and if so, follow up there.)

rmzelle:
@adam3smith, well, the creator of Zotpress did write "It should be completely implemented in the next release (in a week or two).", but I'm not sure whether that refers to DOI hyperlinking (which apparently has been a feature in the last few versions of Zotpress) or to escaping of tricky DOIs.

jgerth:
Thanks for the communication. I looked into this, and do not believe that DOI hyperlinking is in fact implemented in ZotPress as of this writing, at least for the Chicago style. I did try it, just to be certain.

While I am appreciative of the volunteers in this community in maintaining thousands of different styles, I still think we need to do better with providing community CSL styles that allow for links (whether that be for DOIs or otherwise). You are exactly right that the current way (and I believe only way) of doing this, though, is not ideal.

Thanks again.

adam3smith:
yeah, sorry, I misread that. Next version.
But as that thread shows, this can be done with relative ease on the implementation side and I'm absolutely convinced that that's where it belongs. 

(Btw., I'm almost certain the approach in this pull request wouldn't actually have worked to generate hyperlinks with ZotPress anyway: ZotPress uses citeproc-node generated HTML bibliographies, which escape html tags. You can see this at work here: 
https://api.zotero.org/groups/4211/items/4DARFPWW?format=bib&style=https://raw.githubusercontent.com/jgerth/styles/638e56e4f5d98bcbfcacab5b174f73d327bee503/chicago-fullnote-bibliography-linked-doi.csl )

rmzelle:
@adam3smith, think we should open a ticket for this? We could e.g. add an attribute to `<text/>` to give a hint to the CSL processor that the displayed variable should be hyperlinked. I don't think CSL styles should be in the business of generating URL targets, though, and it might be better to keep this entirely outside the CSL schema. E.g. we could write up expectations for CSL processors, and say that they should allow for hyperlinking of variables like DOIs, URLs, and PMIDs, using whatever URL scheme makes the most sense. It probably depends on whether this should be a per-style toggle.

adam3smith:
I think we should try this on xbiblio first. Personally I feel rather strongly that we should do something along these lines:
> it might be better to keep this entirely outside the CSL schema. E.g. we could write up expectations for CSL processors, and say that they should allow for hyperlinking of variables like DOIs, URLs, and PMIDs, using whatever URL scheme makes the most sense. I

and IIRC, Bruce sees it the same way, but others may disagree.

PR #1685, opened 2015-09-14
journal-of-business-logistics.csl

yja100:
Citation for Journal of Business Logistics (modified Chicago Style)

### DIFF
--- a/journal-of-business-logistics.csl
+++ b/journal-of-business-logistics.csl
@@ -1,0 +1,543 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" page-range-format="chicago">
+
+  <info>
+    <title>Journal of Business Logistics</title>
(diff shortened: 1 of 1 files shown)


journal-of-business-logistics.csl
```diff
@@ -0,0 +1,530 @@
...
+    <id>http://www.zotero.org/styles/journal-of-business-logistics</id>
+	<link href="http://www.zotero.org/styles/journal-of-business-logistics" rel="self"/>
+	<link href="http://onlinelibrary.wiley.com/journal/10.1002/(ISSN)2158-1592/homepage/ForAuthors.html"  rel="documentation"/>
+	<link href="http://www.zotero.org/styles/chicago-author-date" rel="template"/>
+	<author>
+      <name>Yousef Abdulsalam</name>
+      <email>y.abdulsalam@asu.edu</email>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="generic-base"/>
```
adam3smith: please change to field="social_science" (yes, we should have business, but we don't; generic-base is reserved for cross-disciplinary styles like Chicago or APA)

PR #1687, opened 2015-09-14
Include p. & pp. in citation locators

reagle:
FM wants p. & pp. in citation locators http://firstmonday.org/ojs/index.php/fm/about/submissions#cite

### DIFF
--- a/first-monday.csl
+++ b/first-monday.csl
@@ -116,7 +116,7 @@
           <text macro="author-short"/>
           <text macro="year-date"/>
         </group>
-        <group delimiter=" " prefix=", ">
+        <group delimiter=" ">
           <label variable="locator" form="short"/>
           <text variable="locator"/>
         </group>
--- a/first-monday.csl
+++ b/first-monday.csl
@@ -116,7 +116,10 @@
           <text macro="author-short"/>
           <text macro="year-date"/>
         </group>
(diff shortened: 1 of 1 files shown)


first-monday.csl
```diff
@@ -116,7 +116,10 @@
           <text macro="author-short"/>
           <text macro="year-date"/>
         </group>
-        <text variable="locator"/>
+        <group delimiter=" " prefix=", ">
```
adam3smith: no need for the prefix here - that's in the group delimiter on line 114

adam3smith:
@reagle would you mind updating this?

reagle:
Sorry, could you be more specific? I think I suggested this via github but I'm not sure how to check it out, modify, and recommit. I'll close and trust you know what to do :)

rmzelle:
@reagle, I think @adam3smith is asking you to reopen this pull request, visit https://github.com/reagle/styles/blob/patch-2/first-monday.csl and click the pencil/edit icon there, and remove the prefix. That should update this pull request.

adam3smith:
yup, that's exactly what I meant.

reagle:
Okay, I think I've done so.

adam3smith:
you did indeed. Thanks!

PR #1689, opened 2015-09-16
Elsevier 2015 09 16

cpina:
Updates styles based on Elsevier's report.


cpina:
I need to update my branch first.

PR #1690, opened 2015-09-17

roaldfrosig:
Style requirements updated according to new documentation. Bills and legislation updated to synchronize citations and bibliography.


rmzelle:
Thanks!

PR #1693, opened 2015-09-18
Update to fix webpages and et. al

University-of-Worcester:


rmzelle:
Is there a reason why you reordered so many XML attributes? E.g.

```
-        <label text-case="lowercase" variable="page" form="short"/>
+        <label form="short" variable="page" text-case="lowercase"/>
```

doesn't have any effect, except making it difficult to see what the real changes are in this pull request.

Also, and this was already wrong before, "et-al-min" and "et-al-use-first" should never have the same value (see e.g. `et-al-min="19" et-al-use-first="19"`). "et-al-min" is the minimum number of authors that triggers et-al abbreviation, and "et-al-use-first" is the number of authors to show before "et al." in that case. "et-al-use-first" should therefore always have a lower number.

OctanMan:
Hi, I'm the former student who authored the style. The et-al-min/first issue is my bad. This commit was simultaneously a handover/walkthrough for those taking responsibility for it in future. They're still new to the process so I'll try and assist where possible to rectify my error asap.

The re-ordering seems to be a by-product of the Visual Editor, perhaps updated to order attributes alphabetically since I last used it? But I sympathise with the annoyance.

rmzelle:
@adam3smith, have you ever seen the Visual Editor do this?

University-of-Worcester:
Hi,
I'm new to editing CSL files (taking over from OctanMan) and have made a couple of changes to the file. I'm unclear on whether this pull request has merged (as per the instructions). Can you please let me know if I need to make those changes via a new pull request instead?

rmzelle:
@University-of-Worcester, we haven't accepted the pull request yet because of the attribute reordering issue. It's best to create a new pull request, and see if you can keep the attributes in order (make sure you're starting from the latest version at https://github.com/citation-style-language/styles/blob/master/harvard-university-of-worcester.csl, and maybe make the required edits by hand).

rmzelle:
Continued in #1742.

PR #1694, opened 2015-09-18
Create universite-de-sherbrooke-departement-de-geomatique.csl

ArMoraer:
New style for Université de Sherbrooke - Département de géomatique appliquée

### DIFF
--- a/universite-de-sherbrooke-departement-de-geomatique.csl
+++ b/universite-de-sherbrooke-departement-de-geomatique.csl
@@ -95,7 +95,8 @@
       </substitute>
     </names>
   </macro>
-  <macro name="access">
+  <!-- Unused macro; commented to pass the Travis CI test -->
+  <!--<macro name="access">
     <choose>
       <if type="thesis">
         <choose>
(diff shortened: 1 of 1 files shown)


ArMoraer:
Last commit passed the Travis CI checks

rmzelle:
After looking at http://www.usherbrooke.ca/geomatique/fileadmin/sites/geomatique/espace-etudiant/guide-redaction2004.pdf, I was wondering whether the Département de géomatique is a subdepartment of Département de géographie et télédétection? Or do they just share the same citation format?

Secondly, it would be better to replace

```xml
```

by

```xml
    <category field="geography"/>
```

(assuming that geomatics is a subtopic of geography; see also http://docs.citationstyles.org/en/stable/specification.html#appendix-i-categories)

ArMoraer:
Hi rmzelle,
"Département de géographie et télédétection" is actually the old name of "Département de géomatique". However, your remarks about the category elements are relevant because I forgot to update them. Since Geomatics is a multidisciplinary field that belongs to Geography, Physics and Computer Science, I have set the following categories : 

```xml
<category field="geography"/>
<category field="physics"/>
<category field="engineering"/>
```

Don't hesitate to correct me if you feel these categories are wrong.

rmzelle:
Thanks for clarifying, and thanks for contributing!

PR #1696, opened 2015-09-19
Harvard - York St John University

stvn-s:
I hope I have followed the instructions to get this style reviewed and hopefully posted on the Zotero Style repoitory


zuphilip:
The automatic check fails because your file is name `harvard-york-st-john-university` instead `harvard-york-st-john-university.csl`, i.e. you are missing the file ending. You can change its name to pass the automatic Travis check (also there might be another review at the name later).

stvn-s:
Thanks for your advice - will make that change

stvn-s:
The original file seems to have disappeared from the repository. Does that mean I should resubmit the file? Or am I missing something?

zuphilip:
No, the file is in your fork. You can go to https://github.com/stvn-s/styles/blob/master/harvard-york-st-john-university and click on the pen (edit) to rename the file and add the file ending. In this way also the the pull request here will be updated.

stvn-s:
I have updated the filename.
Please let me know if there is anything else I need to do

rmzelle:
Which CSL style did you base this one on?

Also, just noting for posterity that http://www.yorksj.ac.uk/information-learning-services/information-learning-services/faqs-for-staff--students/online-guides.aspx works as an alternative documentation.

stvn-s:
It depends how strictly you view "dependency". The style is not really based on another style as such. It is a re-work/adaptation of the Anglia Ruskin University style which I prepared a few years. Some elements have remained the same; other elements have been removed completely; and significant parts of some others have been re-written. 

Yes, there is more than one place to find documentation - if you think that is the best link to use, why not include it in the CSL file.

rmzelle:
If there is any significant dependency left, it would still be good to acknowledge the Anglia Ruskin style via a template link. It is good for attribution and helps us track the lineage of styles. See also https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link (in this case, the link would be `<link href="http://www.zotero.org/styles/harvard-anglia-ruskin-university" rel="template"/>`.

> Yes, there is more than one place to find documentation - if you think that is the best link to use, why not include it in the CSL file.

I think the current link suffices. By noting the other one in the pull request, we already have a record of it.

stvn-s:
There is probably enough dependency to note the link. I have inserted the link as suggested.

rmzelle:
Thanks!

stvn-s:
Does that mean everything is now in place to get the style added to the Style Repository? If there is anything else I need to do, please let me know.
Thanks.

rmzelle:
It just takes a few minutes, but the style made it into the Zotero repository now as well: https://www.zotero.org/styles?q=york%20st%20john

stvn-s:
Brilliant - thanks for your help

rmzelle:
You're welcome!

PR #1697, opened 2015-09-20
Update harvard7de.csl

mmoole:
fix superfluous space before comma at Volume / Issue in Bibliography


rmzelle:
Thanks.

We just had another call to get rid of this style, though :).

http://xbiblio-devel.2463403.n2.nabble.com/Proposal-Retire-duplicate-quot-Harvard-quot-styles-harvard1-csl-amp-harvard7de-csl-tt7579422.html

PR #1702, opened 2015-09-21
2015 09 21 elsevier

cpina:
Elsevier did major changes and consolidation of depending styles.

This pull request reflects it and it's in sync with the last utilities repo information (for Elsevier styles).


cpina:
The Elsevier Guide for Authors hopefully are in sync with the internal state - I'm checking a few more. If not the GfA should be fixed on Elsevier's side.

cpina:
I'll check this and why rake passed locally.

rmzelle:

The only thing we could still do is redirect any deleted styles to their original parent. That way, even though a journal might have been discontinued, people can (theoretically) still edit their manuscripts while keeping the same citation format. (see e.g. https://github.com/cpina/styles/commit/acd98285ea6279966fc58d2e202392aa1af5dfc7 )

adam3smith:
yeah, I'm happy, too. @cpina -- can you easily generate a list of styles that you have deleted (i.e. discontinued journals?)

rmzelle:
Git should make that relatively easy. See e.g. http://stackoverflow.com/questions/6017987/is-there-a-way-in-git-to-list-all-deleted-files-in-the-repository

adam3smith:
Right, but that would also catch the ones that were just turned into dependent styles, since dependents are in a different folder.

rmzelle:
It looks like the only deleted ID is "materials-discovery-today", which @cpina already redirected to "materials-discovery", so this should be good to go.

Carles, did you check for journals that have transferred away from Elsevier, or which have been discontinued?

(I used:

```
$ git log -3 --diff-filter=A --summary --author="Carles Pina" > new.txt
$ git log -2 --diff-filter=D --summary --author="Carles Pina" > old.txt
```

)

adam3smith:
@cpina could you check on two journals, both reported as errors to Mendeley:
Injury -- uses Elsevier Vancouver, their style looks very different (it's Vancouver, but superscript, with bold volumes and italic journal titles). This may just be their guide: http://www.elsevier.com/journals/injury/0020-1383/guide-for-authors a recent open access article did follow Elsevier Vancouver: http://www.injuryjournal.com/article/S0020-1383%2815%2900301-0/pdf
The Veterinary Journal -- Uses Elsevier Harvard but journal wants  full journal titles, not abbreviations. This is still the case in very recent (forthcoming) articles.

PR #1704, opened 2015-09-23
chimia.csl

mpbraendle:
A style for CHIMIA, the International Journal for Chemistry and Official Membership Journal of the Swiss Chemical Society (SCS) and its Divisions.
    This style has been tested with Mendeley and Zotero and has been validated using the CSL Schema 1.0.1. Papers for Mac does not correctly format webpages and patents due to unknown reasons.
    The style has many limits due to CSL constraints, most notably the inability to include pages/page ranges for books, the wrong square brackets for two items cited together, and the lack of support for citing 1a)... b)...c)....

### DIFF
--- a/chimia.csl
+++ b/chimia.csl
@@ -10,15 +10,12 @@
       <email>mpbraendle@gmail.com</email>
       <uri>http://www.id.uzh.ch/cl/iframe/org/visitenk/index.php?id=mb</uri>
     </author>
-    <contributor><name /></contributor>
     <category citation-format="numeric"/>
     <category field="chemistry"/>
     <issn>0009-4293</issn>
-    <eissn>0009-4293</eissn>
     <summary>A style for CHIMIA, the International Journal for Chemistry and Official Membership Journal of the Swiss Chemical Society (SCS) and its Divisions.
-    This style has been tested with Mendeley and Zotero and has been validated using the CSL Schema 1.0.1. Papers for Mac does not correctly format webpages and patents due to unknown reasons.
-    The style has many limits due to CSL constraints, most notably the inability to include pages/page ranges for books, the wrong square brackets for two items cited together, and the lack of support for citing 1a)... b)...c)....</summary>
-    <updated>2015-08-31T22:30:00+00:00</updated>
+    This style has been tested with Mendeley and Zotero and has been validated using the CSL Schema 1.0.1. Papers for Mac does not correctly format webpages and patents due to unknown reasons.</summary>
+    <updated>2015-10-10T216:30:00+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <!-- MACROS, sorted alphabetically -->  
@@ -32,6 +29,20 @@
         <names variable="translator"/>
       </substitute>
     </names>
+  </macro>
+  
+  <!-- computer program -->
+  <macro name="computerprogram">
+    <group delimiter=", ">
+      <group delimiter=" ">
+        <text variable="title" text-case="title"/>
+        <text macro="program-version"/>
+      </group>
+      <text macro="author"/>
+      <text macro="publisher"/>
+      <text macro="year-date"/>
+    </group>
+    <text macro="doi" prefix=", "/>
   </macro>
   
   <!-- DOI -->
@@ -54,9 +65,8 @@
   <!-- editor -->
   <macro name="editor">
     <group delimiter=" ">
-      <!-- <text term="editor" plural="true" text-case="capitalize-first"/> -->
       <names variable="editor">
-        <label form="short" text-case="capitalize-first" suffix=". " strip-periods="true"/>
+        <label form="short" text-case="capitalize-first" suffix=" "/>
         <name initialize-with=". " delimiter=", " delimiter-precedes-last="always"/>
       </names>
     </group>
@@ -199,13 +209,21 @@
         </if>
         <!-- book and report -->
         <else-if type="book report bill graphic legal_case legislation motion_picture song" match="any">
-          <text macro="author" suffix=", "/>
-          <group delimiter=", ">
-            <text variable="title" prefix="‘" suffix="’" />
-            <text macro="publisher"/>
-            <text macro="year-date"/>
-          </group>
-          <text macro="doi" prefix=", "/>
+          <choose>
+            <!-- Zotero maps computer program to book, so check on version variable (may fail) -->
+            <if variable="version">
+              <text macro="computerprogram"/>
+            </if>
+            <else>
+              <text macro="author" suffix=", "/>
+              <group delimiter=", ">
+                <text variable="title" prefix="‘" suffix="’" />
+                <text macro="publisher"/>
+                <text macro="year-date"/>
+              </group>
+              <text macro="doi" prefix=", "/>
+            </else>
+          </choose>
         </else-if>
         <!-- book chapter or conference proceedings -->
         <else-if type="chapter paper-conference entry-encyclopedia" match="any">
@@ -240,16 +258,7 @@
         </else-if>
         <!-- computer program (is mapped to type "article" in Mendeley) -->
         <else-if type="article">
-          <group delimiter=", ">
-            <group delimiter=" ">
-              <text variable="title" text-case="title"/>
-              <text macro="program-version"/>
-            </group>
-            <text macro="author"/>
-            <text macro="publisher"/>
-            <text macro="year-date"/>
-          </group>
-          <text macro="doi" prefix=", "/>
+          <text macro="computerprogram"/>
         </else-if>
         <!-- patent -->
         <else-if type="patent">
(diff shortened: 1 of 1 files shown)


chimia.csl
```diff
@@ -0,0 +1,320 @@
...
+              <text variable="publisher"/>
+              <text macro="thesis-number"/>
+            </group>
+            <text variable="publisher-place"/>
+            <text macro="year-date"/>
+          </group>
+          <text macro="doi" prefix=", "/>
+        </else-if>
+        <!-- computer program (is mapped to type "article" in Mendeley) -->
+        <else-if type="article">
```
adam3smith: not very happy with this. Couldn't we maybe do this by testing for the version variable? Zotero maps software to book. Mendeley also maps a whole lot of other item types to article.

chimia.csl
```diff
@@ -0,0 +1,320 @@
...
+      <text term="edition" form="short"/>
+    </group>
+  </macro>
+  
+  <!-- editor -->
+  <macro name="editor">
+    <group delimiter=" ">
+      <!-- <text term="editor" plural="true" text-case="capitalize-first"/> -->
+      <names variable="editor">
+        <label form="short" text-case="capitalize-first" suffix=". " strip-periods="true"/>
```
adam3smith: use `suffix=" "/>` rather than stripping&adding the period.

chimia.csl
```diff
@@ -0,0 +1,320 @@
...
+    <title>CHIMIA</title>
+    <id>http://www.zotero.org/styles/chimia</id>
+    <link href="http://www.zotero.org/styles/chimia" rel="self"/>
+    <link href="https://chimia.ch/index.php?option=com_content&amp;view=category&amp;layout=blog&amp;id=109&amp;Itemid=489&amp;lang=en" rel="documentation"/>
+    <author>
+      <name>Martin Brändle</name>
+      <email>mpbraendle@gmail.com</email>
+      <uri>http://www.id.uzh.ch/cl/iframe/org/visitenk/index.php?id=mb</uri>
+    </author>
+    <contributor><name /></contributor>
```
adam3smith: delete this line.

adam3smith:
Thanks. A couple of comments inline. Apart from those:

>  the wrong square brackets for two items cited together
what is this referring to? I'm not aware that there's anything about bracketing citations that CSL wouldn't be able to handle, but I could be wrong.

The automated check fails because
1. you use a macro publisher-place that isn't defined
2. you use the same ISSN twice -- just include it once, likely as eissn.

rmzelle:
(you can update the current pull request via the pencil icon at https://github.com/mpbraendle/styles/blob/master/chimia.csl )

rmzelle:
And regard the lack of support for "composite references" in CSL, are you, or do you know, an editor at  CHIMIA? We've always had trouble finding somebody who is knowledgeable about this type of citations, and it's still not clear what the requirements typically are.

For instance, for this journal, are composite sorted in any way (e.g. by date), or are they just printed in the order in which they are cited?

rmzelle:
@mpbraendle, let us know if you'd like to get back to this pull request. Otherwise we'll close it in a few days.

chimia.csl
mpbraendle: Testing on the version variable is not a sufficient condition, since it may be empty. Not every software has a version attached to it, and authors may be lax in citing correctly.
I would rather prefer that in the next version, a type "computer program" (or similar), as Endnote does have.

chimia.csl
mpbraendle: Ok.

mpbraendle:
Hi rmzelle and Adam Smith,

sorry for answering so late, I was very busy the last two weeks and did not find the time to invest one hour or so into fixing chimia.csl according to your suggestions.

Yes, I'm an Editor (among other things responsible for Web) at Chimia. 

CHIMIA's Instructions to Authors on citations are very sparse, and defined by examples you have to deduce the citation style. The Technical Editor told me that it follows to some extent the citation style of Angewandte Chemie, so I took the angewandte-chemie.csl as as start point. This one had this comment on composite references. But I can delete it.

CHIMIA's instructions tell: "Whenever possible, composite references should be used, instead of a series of invidual ones." Usually, these are used if you have several references support or prove a statement. The format then is: [citation number] a) citation1; b) citation2; c) citation3 ...  . Endnote has an option for this in the Number format: "Use one number for references always cited together", specification of separator, "Use letters to separate grouped references", letter format and so on. Sorting is upon the will of the author.

I will correct what I can and update the pull request today.

mpbraendle:
Regarding ISSN: CHIMIA has  the same ISSN for <issn> and <issnl>. You may look up portal.issn.org

rmzelle:
> Regarding ISSN: CHIMIA has the same ISSN for `<issn>` and `<issnl>`

Yes, that's usually the case. However, your style uses `<issn>` and `<eissn>`, and the e-ISSN (online) and print ISSN always differ. I can't find the e-ISSN for this journal, so it would be better to remove `<eissn>0009-4293</eissn>`.

rmzelle:
> CHIMIA's instructions tell: "Whenever possible, composite references should be used, instead of a series of invidual ones." Usually, these are used if you have several references support or prove a statement. The format then is: [citation number] a) citation1; b) citation2; c) citation3 ... . Endnote has an option for this in the Number format: "Use one number for references always cited together", specification of separator, "Use letters to separate grouped references", letter format and so on.

(if you could find links to online EndNote documentation on this topic, or just post a screenshot here of this window, that would be helpful)

So, what happens if the author first cites an article by Doe et al. in the first sentence, and then cites the same article plus several more (e.g. from Brown et al. and Johnson & Johnson) in the second sentence? Does the earlier cite to Doe et al. make the composite reference impossible?

chimia.csl
mpbraendle: The basic problem is:
1) CSL  has no type "software" or "computer program" as Endnote has
2) Mendeley and Zotero map this document type to different types in CSL
3) Mendeley, Papers, Zotero store the program version in different fields (version, edition or number), depending on what the author does.

In the worst case 2) and 3) together would require checking on 6 different combinations, and you still have no clue if the document type is still a computer program. 
So implementation of 1) in the next version of CSL is  absolutely essential.

mpbraendle: I leave the "article" part as it is. For Zotero, although rarely used by chemists - they prefer Endnote or Mendeley - I will add a check on version in the "book" section. Although I'm not happy with this. The easiest solution would be to add correct document types, so that reference management software not has to map arbitrarily.

mpbraendle:
Yes, in such a case, references must be split. Citation would be:

phrase1[1] . phrase2[1-2]

References would be:

[1] Doe et al
[2] a) Miller et al, b) Stevenson et al, 

A composite reference also can be cited individually: phrase3[2b]

mpbraendle:
Regarding ISSN, I use issnl now which is correct. 
Stay tuned, I will update the chimia.csl in a few hours.

rmzelle:
> Regarding ISSN, I use issnl now which is correct. 

No, please just use the ISSN. We have a `<issnl>` element in CSL, but we prefer to use `<issn>` instead.

mpbraendle:
Ok. I just use issn

chimia.csl
adam3smith: yes, no disagreement with the need for a software item type.

adam3smith:
still wondering about this:

>the wrong square brackets for two items cited together

is this referring to the fact that two references should be [1]
[1] a)ref 1; b) ref 2
instead of [1,2] or something else?

chimia.csl
rmzelle: > yes, no disagreement with the need for a software item type.

Do we have a ticket? Can't find one in the zotero-bits repo or schema/documentation CSL repos.

chimia.csl
mpbraendle: Ok.

mpbraendle:
OK, I have implemented your suggestions, as far this was possible.

chimia.csl
```diff
@@ -0,0 +1,329 @@
...
+      <name>Martin Brändle</name>
+      <email>mpbraendle@gmail.com</email>
+      <uri>http://www.id.uzh.ch/cl/iframe/org/visitenk/index.php?id=mb</uri>
+    </author>
+    <category citation-format="numeric"/>
+    <category field="chemistry"/>
+    <issn>0009-4293</issn>
+    <summary>A style for CHIMIA, the International Journal for Chemistry and Official Membership Journal of the Swiss Chemical Society (SCS) and its Divisions.
+    This style has been tested with Mendeley and Zotero and has been validated using the CSL Schema 1.0.1. Papers for Mac does not correctly format webpages and patents due to unknown reasons.</summary>
+    <updated>2015-10-10T216:30:00+00:00</updated>
```
adam3smith: typo in the date here (216)

adam3smith:
that's the only reason the check fails as far as I can see. Everything else looks good.

mpbraendle:
Is corrected now.

chimia.csl
rmzelle: @adam3smith? We should probably create one in the zotero-bits repo if one doesn't yet exist.

adam3smith:
thanks!

chimia.csl
adam3smith: yes, agree

rmzelle: Done: https://github.com/avram/zotero-bits/issues/69

PR #1708, opened 2015-10-04
Add director to contributors and contributors-short macros

nickbart1980:


adam3smith:
thanks. If you're adding this to other styles, make sure to test those with director data in pandoc. It's possible a couple of styles may have some workaround based on type="motion_picture" implemented. As I said, adding director won't break anything where that variable doesn't exist, but may conflict with workarounds where it does.

nickbart1980:
Ok, sure. This one has indeed been tested in pandoc.

PR #1712, opened 2015-10-08
[onderstepoort] Order inline citations alphabetically

zuphilip:
<blockquote>In the style guide a the bottom of p.3 ( http://openjournals.net/files/Ref/HARVARD2009%20Reference%20guide.pdf ), where several authors are listed at once, the authors are listed alphabetically.</blockquote> see https://forums.zotero.org/discussion/52840/style-error-onderstepoortjournalofveterinaryresearchcsl/#Item_2

### DIFF
--- a/onderstepoort-journal-of-veterinary-research.csl
+++ b/onderstepoort-journal-of-veterinary-research.csl
@@ -115,6 +115,9 @@
     </group>
   </macro>
   <citation et-al-min="3" et-al-use-first="1" disambiguate-add-names="true" disambiguate-add-givenname="true" disambiguate-add-year-suffix="true">
+    <sort>
+      <key macro="author-short"/>
+    </sort>
     <layout delimiter="; " prefix="(" suffix=")">
       <group delimiter=", ">
         <group delimiter=" ">


zuphilip:
BTW the version at http://editor.citationstyles.org/styleInfo/?styleId=http%3A%2F%2Fwww.zotero.org%2Fstyles%2Fonderstepoort-journal-of-veterinary-research is not reflecting the last changes, which are done a month ago. Maybe some update is needed?

adam3smith:
yes, unfortunately the styles in the online editor aren't update automatically (nor is the citeproc)

adam3smith:
thanks!

rmzelle:
> Maybe some update is needed?

As it happens, I got a similar comment last week per email from Papers, in response to which I wrote:

> Yes, the situation with the CSL editor is a bit unfortunate. It currently needs to be manually updated to use a newer copy of the styles repository (this eventually should be automated so that the
editor doesn't lag months behind). Also, the creator, Steve Ridout, worked on it based on a one-time grant, and has since moved on. The project is currently without a maintainer, and nobody else on the CSL team has attempted to get familiar with the code base (we're a bit short on manpower and experienced JavaScript developers). There are some instructions on how to update it, though, which seem relatively straightforward: https://github.com/citation-style-editor/csl-editor-demo-site/blob/master/README.md.

The instructions don't seem overly complicated (citeproc.js was e.g. updated in https://github.com/citation-style-editor/csl-editor/commit/be7bb931a3e9cac91a93da47fa032e8e6db11ee9, and the CSL styles in https://github.com/citation-style-editor/csl-editor/commit/bf1644b2434e1670a22e8767440253d2b09a2f59 ), and the site seems to be hosted by GitHub. We might be able to figure out between ourselves how to update the site.

PR #1713, opened 2015-10-09
Enable shelbot

inukshuk:
We should be able to enable the bot now, that the distribution updater should ignore failed builds.

See https://github.com/citation-style-language/distribution-updater/pull/1#issuecomment-144756239

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ notifications:
     on_failure: always
   webhooks:
     urls:
+      - https://shelbot.herokuapp.com/build
       - https://styles-update.zotero.org:8827/
     on_success: always
-    on_failure: never
+    on_failure: always
+    on_start: never


rmzelle:
Thanks. Is there any documentation on what "on_start" does? I couldn't find any description.

inukshuk:
Honestly, I have not tried it. It is mentioned in the [docs](http://docs.travis-ci.com/user/notifications/#Webhook-notification), but they don't elaborate on any specifics. Judging from the name, I'd assume it's a hook that is triggered when the build starts -- which we don't require since the bot should be triggered by GitHub directly. I just turned it off to make sure it's off.

Speaking of which, though, we still need to enable the GitHub hook. In the project settings there is a page for webhooks. There, you need to enable a custom hook to notify 'https://shelbot.herokuapp.com/pull_request' with content type json. The only event we need to trigger is 'Pull Request' (you need to select 'let me choose individual events'.

adam3smith:
Done. Hope I did that right:
screenshot from 2015-10-10 11 42 29

inukshuk:
Looks perfect. This should trigger the bot to comment on every PR from now on.

inukshuk:
Actually, hang on, I'll check if we're using the secret token.

inukshuk:
No, it should work already, but we may want to set a secret to make sure the PR message we receive are legit.

PR #1715, opened 2015-10-09
New style for Biology Education Centre, Uppsala University, English.

perenstrom:


adam3smith:
how similar are the English and the Swedish style apart from the terms/numbers? Any chance we could wrap this into one style and people can just toggle it by setting the locale (which now both Zotero and Mendeley support).

perenstrom:
They are similar, but not similar enough to merge them we feel. We have decided to keep them separate for clarity for our students as well.

adam3smith:
hmm -- I just ran a diff on these. All differences, and there aren't many, are in terms and could easily be handled by using terms and defining them in a locale section.
We have a strong preference for not letting the number of styles grow more than necessary, so everything that _can_ be handled in a single style should be. That's why we have robust localization features (You can obviously always host styles locally if you don't agree with that).

So, I'd ask you to turn this into a single style. If you run into problems with that, let us know and we'll be happy to help.

perenstrom:
Okay. I'll have to read up on this and return then. 

Should these two be replaced by one new pull?

adam3smith:
You can just close one PR and modify the other one. (of if you find it much easier you can also close both and start a new one. We have a small preference for the former, though)

The basic idea would be to set no default locale, to define any term that is currently hardcoded in affixes as a term and then define it/them as needed in the locale section of the style. At a quick glance, that just affects the two prefixes in these lines

```
        <else-if type="webpage post-weblog" match="any">
          <group suffix=".">
            <group prefix="WWW document">
              <date form="text" variable="issued" prefix=" "/>
              <text variable="URL" prefix=": " suffix=". "/>
            </group>
            <date form="text" variable="accessed" prefix="Accessed "/>
          </group>
        </else-if>
```

Accessed already has a term, so that's easy. WWW document you could just use the term "online" for, i.e. `<term name="online">WWW document</term>` and the equivalent for Swedish and then just use `<text term="online"/>`

perenstrom:
Thanks for the help. Working on it now. A question though. Non language-specific terms, should I keep them in a separate <locale> tag?

perenstrom:
And sorry, I can't find any info about this in the documentation. How do I use terms as prefixes?

And can I still keep the difference with date form as well? Text in english and numeric in swedish?

perenstrom:
Would this be correct, if the terms are defined correctly?
```
          <group suffix=".">
            <group>
              <text term="online" />
              <date form="text" variable="issued" prefix=" "/>
              <text variable="URL" prefix=": " suffix=". "/>
            </group>
            <group>
              <text term="accessed">
              <date form="text" variable="accessed" prefix=" "/>
            </group>
          </group>
```

rmzelle:
> And can I still keep the difference with date form as well? Text in english and numeric in swedish?

Yes. The trick here would be to redefine the "text" form of the English date format as a numeric format, or redefine the Swedish "numeric" date format as a text format.

> Would this be correct, if the terms are defined correctly?

It's hard to tell without knowing what the output should look like, but in general, yes, that looks okay. Is the style guide for this format available online anywhere? (the current documentation link just points to the website of your department)

perenstrom:
How would I redefine that?

The style guide can be found in our folder Presenting Science (http://www.ibg.uu.se/digitalAssets/331/331418_1presenting-science-2011-webb-eng.pdf, page 11).

rmzelle:
As an example, you could add the following date format definitions to your style (you already have the necessary `<locale/>` elements, so just reuse those):

```
<locale xml:lang="en-US">  
  <date form="text">
    <date-part name="month" suffix=" "/>
    <date-part name="day" suffix=", "/>
    <date-part name="year"/>
  </date>
</locale
<locale xml:lang="sv-SE">
  <date form="text">
    <date-part name="month" form="numeric-leading-zeros" suffix="/"/>
    <date-part name="day" form="numeric-leading-zeros" suffix="/"/>
    <date-part name="year"/>
  </date>
</locale>
```

Then just use `<date variable="..." form="text"/>` throughout your style.

(what I did here was take the "text" and "numeric" formats from https://raw.githubusercontent.com/citation-style-language/locales/master/locales-en-US.xml, and redefine them as the "text" format for "en-US" and "sv-SE" within the style. Modify as necessary. See also http://docs.citationstyles.org/en/stable/specification.html#localized-date-formats )

Can you update the "documentation" link to point to that PDF?

perenstrom:
Thank you for the explanation. So the ```/```-suffix in the Swedish version is to make it ```month/day/year```?

I'll update the link in next commit as well!

rmzelle:
> So the /-suffix in the Swedish version is to make it month/day/year?

Yes, that's the default US English "numeric" format. For Swedish perhaps you'd want something like:

```
  <date form="text">
    <date-part name="year"/>
    <date-part name="month" form="numeric-leading-zeros" prefix="-" range-delimiter="/"/>
    <date-part name="day" form="numeric-leading-zeros" prefix="-" range-delimiter="/"/>
  </date>
```

(from https://raw.githubusercontent.com/citation-style-language/locales/master/locales-sv-SE.xml )

perenstrom:
Thank you for your help!

perenstrom:
Fixed the date, which seems to work correctly now (https://github.com/citation-style-language/styles/commit/a78b69f3d8b8a303090aff178fbc0c833facd8f5).

Another question though. Can I restrict the language pulldown in Zotero to only Swedish and English UK when using my style?

And another thing. How can I make the term ```accessed``` capitalise correctly?

perenstrom:
After a bit of googling I managed to fix it. I added ```text-case="capitalize-first"``` where I was using the term.

perenstrom:
Just to perhaps lift my previous question that might be missed:

Is there any way of restricting which languages there are to chose from for the style in Zotero? I've noticed that some styles only allow certain languages.

rmzelle:
> Is there any way of restricting which languages there are to chose from for the style in Zotero?

You can use the `default-locale` attribute on the root `<style/>` element to restrict the style to one particular language (see http://docs.citationstyles.org/en/stable/specification.html#the-root-element-cs-style for an example). To limit a style to Swedish and British English, you can create a dependent style (see http://docs.citationstyles.org/en/stable/specification.html#dependent-styles ) for the second locale. For example, you could:

* restrict "uppsala-universitet-institutionen-for-biologisk-grundutbildning.csl" to "sv-SE"
* create a dependent style "uppsala-universitet-institutionen-for-biologisk-grundutbildning-en.csl" restricted to "en-GB"

perenstrom:
> create a dependent style "uppsala-universitet-institutionen-for-biologisk-grundutbildning-en.csl" restricted to "en-GB"

This is what I tried doing in the beginning, which was suggested against. So I'll keep it as one file now.

Will different English language settings fall back to the one I have available in the file, or will en-US fall back to default locale files?

Other than this I have no more questions and the style should be ready to be merged!

rmzelle:
> This is what I tried doing in the beginning, which was suggested against.

Where did we do this, exactly?

adam3smith:
I asked not to create two independent styles if at all possible -- if having a separate dependent style is highly desirable, I'm OK with it, though I'd still have a preference for having just one style (@rmzelle may disagree with that, though).

> Will different English language settings fall back to the one I have available in the file, or will en-US fall back to default locale files?

The one in the style. See here: http://docs.citationstyles.org/en/stable/specification.html#locale-fallback

rmzelle:
Right, so @adam3smith primary concern above was the creation of two *independent* styles. We're much more open to accepting a single *independent* style accompanied by a *dependent* style in a secondary language.

We have two options. We can either leave the locale of the independent style undefined, in which case the user needs to use the style in the desired locale (e.g. in Zotero and Mendeley this can be easily done via a language drop-down menu). Alternatively, we can have an independent style for one language, and a dependent one for the second one. In that case the user can just install whichever version they need.

I think the latter solution is friendlier to the user, and acceptable provided that the number of locales is limited (to 2-3, say). So I'm totally willing to accept a Swedish independent and English dependent style in this case.

perenstrom:
Thanks for both of your input. I think I'll stick with one file for the moment and see if we get some respons from our students and teachers. If they seem to unlike the language dropdown we can consider splitting it up into two different files again later.

Perhaps I should look into pushing for the posibility to restrict the language dropdown to multiple languages. But that is for another forum.

adam3smith:
great, thanks!

PR #1727, opened 2015-10-13
Add style for journal CSSE

andersjohansson:
This style is based on APA 6th ed (no ampersand) (https://www.zotero.org/styles/apa-no-ampersand) with just a few small changes to conform to the style guide for Cultural Studies of Science Education.
- No comma between author and year in citations
- Use ampersand in bibliography


adam3smith:
Looks good, thanks. I want to take a look at the APA no ampersand style to see where that's used and make sure it shouldn't, in fact, look like this one. I believe we created that for Springer journal(s), too.

rmzelle:
(the "no ampersand" variant originates here: https://github.com/citation-style-language/styles/pull/1552 )

adam3smith:
sorry for the delay here & thanks for contributing

PR #1733, opened 2015-10-15
Update centre-de-recherche-sur-les-civilisations-de-l-asie-orientale.csl

PhilPons:


rmzelle:
@PhilPons, could you correct the timestamp? `<updated>2015-10-14</updated>` is not a valid format (see http://books.xmlschemata.org/relaxng/ch19-77049.html for background). E.g. use `2015-10-14T12:00:00+00:00`. You can update your style via the edit (pencil) icon at https://github.com/PhilPons/styles/blob/patch-1/centre-de-recherche-sur-les-civilisations-de-l-asie-orientale.csl

PhilPons:
It's done !
Thank you for the information.
Best regards

rmzelle:
Great. Validation now showed another error in:

```
              <choose>
                <if variable="volume">
                  <text term="volume" form="short" suffix=". "/>
                  <text variable="volume"/>
                </if>
                <if variable="number-of-volumes">
                  <text variable="number-of-volumes" suffix=" "/>
                  <text term="volume" form="short" suffix="."/>
                </if>
              </choose>
```

You can only use a single `<if/>` section per `<choose/>`. See http://docs.citationstyles.org/en/stable/specification.html#choose. If these tests are independent, use two `<choose/>` sections.

rmzelle:
How should references look like if they have both a volume and number-of-volumes? Just curious.

PhilPons:
Hi,

Not well I guess. But we assume that this case doesn't appear. It's probably not relevant to have both informations ?

Thank you for your help
De : Rintze M. Zelle [notifications@github.com]
Envoyé : jeudi 15 octobre 2015 18:33
À : citation-style-language/styles
Cc : PONS Philippe
Objet : Re: [styles] Update centre-de-recherche-sur-les-civilisations-de-l-asie-orientale.csl (#1733)


How should references look like if they have both a volume and number-of-volumes? Just curious.

rmzelle:
It might be better to change

```
            <group>
              <choose>
                <if variable="volume">
                  <text term="volume" form="short" suffix=". "/>
                  <text variable="volume"/>
                </if>
              </choose>
              <choose>
                <if variable="number-of-volumes">
                  <text variable="number-of-volumes" suffix=" "/>
                  <text term="volume" form="short" suffix="."/>
                </if>
              </choose>
            </group>
```

into

```
              <group delimiter="/">
                <group delimiter=" ">
                  <text term="volume" form="short" suffix="."/>
                  <text variable="volume"/>
                </group>
                <group delimiter=" ">
                  <text variable="number-of-volumes"/>
                  <choose>
                    <if variable="volume" match="none">
                      <text term="volume" form="short" suffix="."/>
                    </if>
                  </choose>
                </group>
              </group>
```

The latter should give:

* only "volume": "vol. 5"
* only "number-of-volumes": "12 vol."
* both "volume" and "number-of-volumes": "vol. 5/12" (this is what the style did before)

Maybe you could test whether this works as expected?

PhilPons:
Such a solution seems perfect !
Thank you so much for all your help !

I change the CSL, test this work and create, just now, a new pull request. 

Thank you.

rmzelle:
Can you make the changes in this pull request? That way it's easier for us to keep track of the style's history. Just use the edit icon at https://github.com/PhilPons/styles/blob/patch-1/centre-de-recherche-sur-les-civilisations-de-l-asie-orientale.csl again.

rmzelle:
Great, thanks!

PR #1739, opened 2015-10-21
update universita-pontificia-salesiana styles

xavi-mat:
Make dependent some styles (en, pt-br, es, fr)
Update (it) style.
(they weren't because there was a bug; see https://github.com/citation-style-language/styles/pull/1194)


adam3smith:
thanks for doing that.

PR #1741, opened 2015-10-22
Update stem-cells-and-development.csl

### LINKED ISSUE #1591: Stem cells and development, issue with listing the references
When I use the Stem cells and development style the references should be listed numerically in the text but instead they are listed alphabetically in the reference list, thus not starting the reference count at [1] in the text. 

Has anyone else experienced this? And do you have any suggestions on how to change it?

Bw Rós

rmzelle:
Fix #1591.

### DIFF
--- a/stem-cells-and-development.csl
+++ b/stem-cells-and-development.csl
@@ -50,11 +50,6 @@
     </layout>
   </citation>
   <bibliography et-al-min="51" et-al-use-first="50" line-spacing="2">
-    <sort>
-      <key macro="author"/>
-      <key macro="issued"/>
-      <key variable="title"/>
-    </sort>
     <layout suffix=".">
       <text variable="citation-number" suffix=". "/>
       <group delimiter=" ">


adam3smith:
look at that! happy csl-bot (FYI @inukshuk )

rmzelle:
:sparkles::sparkles::sparkles:

inukshuk:
At last : ) Thanks guys!

I would suggest that we re-enable the PR welcome message again and start to collect a little bit of experience with the bot on real PRs. Then we can start thinking about the next improvements.

inukshuk:
Ooops, just noticed that you already re-enabled the PR hook.

rmzelle:
> I would suggest that we re-enable the PR welcome message again

It's on already, right?

inukshuk:
Yes, it's on, my bad!

rmzelle:
Could you look into https://github.com/inukshuk/Sheldon/issues/16, though? For some of the PRs I get "csl-bot" as the OP (original post) author in GMail, which is a bit confusing. My best guess is that the email notification for the initial welcome comment sometimes arrives a few seconds before GitHub's own email, which a delay of 10-30 s would probably fix.

PR #1743, opened 2015-10-23
[apa.cls] Add possibility to cite data sets

zuphilip:
This is described here http://blog.apastyle.org/apastyle/2013/12/how-to-cite-a-data-set-in-apa-style.html . However, we will use the newest recommondation about how to format doi's here as well. I added also the case, that it is possible to have data with a title contained in a data set with a title as outlined here https://www.zotero.org/support/dev/translators/datasets

Let me know what you think.

### DIFF
--- a/apa.csl
+++ b/apa.csl
@@ -191,6 +191,19 @@
           </else>
         </choose>
       </if>
+      <else-if type="dataset" match="any">
+        <choose>
+          <if match="none" variable="container-title">
+            <group delimiter=" ">
+              <text variable="title" font-style="italic"/>
+              <text value="Data set" prefix="[" suffix="]"/>
+            </group>
+          </if>
+          <else>
+            <text variable="title"/>
+          </else>
+        </choose>
+      </else-if>
       <else-if variable="reviewed-author">
         <choose>
           <if variable="reviewed-title">
@@ -336,7 +349,7 @@
                     <date-part prefix=", " name="month"/>
                   </date>
                 </if>
-                <else-if type="article-journal bill book chapter graphic legal_case legislation motion_picture paper-conference report song" match="none">
+                <else-if type="article-journal bill book chapter graphic legal_case legislation motion_picture paper-conference report song dataset" match="none">
                   <date variable="issued">
                     <date-part prefix=", " name="month"/>
                     <date-part prefix=" " name="day"/>
@@ -363,7 +376,7 @@
   </macro>
   <macro name="issued-sort">
     <choose>
-      <if type="article-journal bill book chapter graphic legal_case legislation motion_picture paper-conference report song" match="none">
+      <if type="article-journal bill book chapter graphic legal_case legislation motion_picture paper-conference report song dataset" match="none">
         <date variable="issued">
           <date-part name="year"/>
           <date-part name="month"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
awesome in general. Only question is how we should handle the [Data set]. I'm not very happy with hardcoding it in the style because it breaks localization. As you know, the solution we've been using is to put genre in square brackets -- which also is, obviously everything but ideal. 
Obviously the real solution is to have a term (I'd like a term for every CSL item type), but what do you think makes more sense in the meantime. cc @rmzelle

zuphilip:
The variable `genre` is for `journalArticle` in Zotero not present. Thus, we had to add it also as part of `extra`. Would you suggest to replace the hardcoded "Data set" with the extra-variable OR just to prefer the extra-variable if present?

adam3smith:
I'm honestly not sure. Let's see what Rintze says. I think it could be argued that your current approach is better because we'd essentially just swap out text value for text term to properly implement it once we have a term.

bwiernik:
Given that apa.csl has the <format> macro for placing the contents of <medium> in square brackets for any item (to allow APA's arbitrary format descriptions), the dataset solution should be sure to accommodate that. If we hard code [Data set] into the style, then the simplest solution would probably be to place it as a conditional when the <format> macro is called in the <title-plus-extra> macro.

That being said, using <genre> or <medium> might not necessarily be bad. Consider the second example from this style blog post (http://blog.apastyle.org/apastyle/2013/12/how-to-cite-a-data-set-in-apa-style.html). If the style uses a term, flexible descriptions like "data set with code book" would not be possible (if they were desired). It might be best to supply a term/hard coded text only on the condition that <medium> and <genre> do not exist. 

For datasets with <container-title>, do you have an example reference or indication that the [Data set] should be printed after the container title? Generally, material in parentheses and square brackets always follow the item title.

zuphilip:
The `format` macro with the `medium` variable sounds fine for me. However, we should make sure that the corresponding variable will be part of the coming zotero itemType. Actually, the part with `container-title`, I have coded without an example or any guidelines. Thus, we may just simplify delete the additional hack. I didn't know about the general rule about material in parentheses.

I will make the changes later, maybe tomorrow. @adam3smith Do you agree with `medium` instead of `genre`?

bwiernik:
I have never seen an example with the [Format] information anywhere but after the item title, so I think we can simplify and just have it in the `format` macro.

adam3smith:
yes, everything that bwiernik says (include it in the medium macro and add the "or" condition to allow a default [Data set] and "overriding" the default with the medium field) makes perfect sense--go ahead with that.

bwiernik:
While you are at it, adding a similar conditional for computer programs would be nice.

adam3smith:
That's not that easy, since csl doesn't have an item type

bwiernik:
Using the conditional:

<if type="book" variable="version" match="all">

That is used elsewhere in the style?

zuphilip:
I updated the code for `dataset`. (The suggestion for computer file would be fine for me, but I would like to wait for the opinion from Sebastian or Rintze.)

adam3smith:
I don't have objections against doing this for software, but I think makes sense to keep that separate since the logic is more complicated

zuphilip:
Thank you. I updated the <a href="https://www.zotero.org/support/dev/translators/datasets">documentation here</a> accordingly to include `medium`, CC @aurimasv

rmzelle:
Sorry for the late response, but I'm fine with this. Hard-coding obviously doesn't localize, but apa.csl already has some features that don't localize well (like some non-localizing dates), and it's more robust than relying on field content (which can contain typos and formatting inconsistencies).

(also, @adam3smith, regarding "(I'd like a term for every CSL item type)", feel free to create a ticket for that; I don't think we have one)

PR #1744, opened 2015-10-24
Create national-science-foundation-chemistry-grant-proposal

roditg:


rmzelle:
Does https://github.com/citation-style-language/styles/blob/master/national-library-of-medicine-grant-proposals.csl not work for you?

(your pull request is failing because your style is missing a ".csl" extension, by the way)

roditg:
I need a citation layout format that has the title first, then the authors, etc.

adam3smith:
do you have documentation for the style? This looks very basic/fragile: let's maybe figure out what NSF wants exactly before you put too much more work into the submission here.

Though FWIW, The current style fails because you have 
1. several missing elements of info (updated and all links)
2. several smart quotes instead of regular quotes, the first of those in line 26 -- never copy & paste from a word processor--that break XML parsing entirely

adam3smith:
(and see the link for validation -- at a minimum, your style should pass validation at http://validator.citationstyles.org )

adam3smith:
Hmm -- not sure what happened here. Any particular reason you deleted&closed this?

roditg:
Figure out 2 things: 1. How to actually modify styles in papers without creating a pull request.  2. that Papers doesn't have away of grouping references under the same number but with subheadings, making this relatively useless for what I need.

PR #1747, opened 2015-10-28
APA-no-doi-no-issue: Remove DOI from output

Seanmcn:
So unless I'm missing something, why does apa-no-doi-no-issue actually display the DOI?

### DIFF
--- a/apa-no-doi-no-issue.csl
+++ b/apa-no-doi-no-issue.csl
@@ -130,32 +130,25 @@
       <else>
         <choose>
           <if variable="page" match="none">
-            <choose>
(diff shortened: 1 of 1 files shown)


adam3smith:
it displays DOIs for items without page ranges, so as to capture online only articles. That is on purpose (it's a style we're using for several journals). What are you using this for?

rmzelle:
As @adam3smith says, this variant doesn't print DOIs for articles that already can be tracked down via page numbers. There might be a good reason to accept a true no-DOI variant, so feel free to make the case for such a style, but the current style should stay as is in this respect.

PR #1750, opened 2015-11-04

roaldfrosig:
Changed the date format for legislation, and removed several instances of wrongly applied italic font-style.


rmzelle:
Thanks!

PR #1751, opened 2015-11-06
Update zookeys.csl

stuckyb:
Minor bug fix so that "et al." inline citations display correctly when a reference is used two or more times.

### DIFF
--- a/zookeys.csl
+++ b/zookeys.csl
@@ -97,8 +97,8 @@
   <macro name="publisher">
     <choose>
       <if type="thesis" match="none">
-        <group prefix="" suffix="" delimiter=", ">
-          <text variable="publisher" suffix=""/>
+        <group delimiter=", ">
+          <text variable="publisher"/>
           <text variable="publisher-place"/>
         </group>
         <text variable="genre" prefix=". "/>
@@ -211,7 +211,7 @@
             <text macro="edition"/>
             <text macro="editor" suffix="."/>
           </group>
-          <group prefix=" " suffix="" delimiter=", ">
+          <group prefix=" " delimiter=", ">
             <text macro="publisher"/>
             <text variable="page" prefix=" " suffix="pp."/>
           </group>
--- a/zookeys.csl
+++ b/zookeys.csl
@@ -47,7 +47,7 @@
   </macro>
   <macro name="author-short">
     <names variable="author">
-      <name form="short" delimiter=" " and="text" delimiter-precedes-last="never" et-al-min="3" et-al-use-first="1" initialize-with=". "/>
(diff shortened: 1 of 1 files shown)


zookeys.csl
```diff
@@ -47,7 +47,7 @@
   </macro>
   <macro name="author-short">
     <names variable="author">
-      <name form="short" delimiter=" " and="text" delimiter-precedes-last="never" et-al-min="3" et-al-use-first="1" initialize-with=". "/>
+      <name form="short" delimiter=" " and="text" delimiter-precedes-last="never" et-al-min="3" et-al-use-first="1" et-al-subsequent-min="3" et-al-subsequent-use-first="1" initialize-with=". "/>
```
adam3smith: I'd much rather you remove the et al settings here entirely and instead include them in the citation line (line 154) below. There's no reason to have the useless settings in there and, as you've seen, it can cause all types of problems when working with styles.

adam3smith:
You can modify your pull request by clicking the pencil icon here:
https://github.com/citation-style-language/styles/pull/1751/files

stuckyb:
I didn't even realize the other "et al" properties were down there.  
it's fixed now.

adam3smith:
the actual change you made is great, but do you have any idea where all the additional stuff is coming from? All the `suffix=""` and the so on? None of that is actually doing anything, but it is a lot of clutter that we'd rather not have in the style.

stuckyb:
Yep -- those came from the Javascript-based CSL editor that the official Zotero site recommends using.  Since it added those to the style definitions, I figured it might be preferred to be explicit about those properties, so I left them in.

adam3smith:
hmm, that's both unfortunate and surprising. I know the editor will leave these in as you edit the respective line, but it definitely shouldn't be adding them. (So no, we prefer them to be explicit. On the contrary).

stuckyb:
Okay, the empty suffix and prefix property definitions are all gone.

adam3smith:
Great, thanks -- and you were using the version of the editor at http://editor.citationstyles.org/ ?

stuckyb:
Yes, I was.  Who knows, perhaps I accidentally told it to add those empty properties to the XML somehow, or introduced them at some other point without realizing it.

adam3smith:
Any chance you could have been working with an older version of the style? Those were removed since you last modified the style on September 12, 2013.

adam3smith:
Anyway, if you're going to work with the editor again, just make sure to do a hard reload (hold shift while reloading) to make sure to grab the most recent version, which also includes up-to-date styles.

stuckyb:
Ah, yes -- that probably was it!

PR #1752, opened 2015-11-07
Update zookeys.csl

stuckyb:
Small bug fix to ensure that references that are disambiguated by appending an alphabetic character to the year (e.g., Smith 1996a, Smith 1996b) display correctly in the bibliography.


adam3smith:
thanks!

PR #1754, opened 2015-11-11

roaldfrosig:
Three small changes:

1) Added "date unknown" when no date is given.
2) Changed "pp." to "p." for multiple pages.
3) Added a space after citation number, to make the bibliography block align correctly.

### DIFF
--- a/university-college-lillebaelt-vancouver.csl
+++ b/university-college-lillebaelt-vancouver.csl
@@ -18,6 +18,11 @@
       <term name="online">internet</term>
       <term name="accessed">cited</term>
       <term name="available at">available from</term>
+      <term name="no date">date unknown</term>
+      <term name="page" form="short">
+        <single>p.</single>
+        <multiple>p.</multiple>
+      </term>
     </terms>
   </locale>
   <!--                                                 CITATION MACRO -->
@@ -159,6 +164,9 @@
   <macro name="year">
     <choose>
       <if type="legislation bill"/>
+      <else-if variable="issued accessed" match="none">
+        <text term="no date" prefix="[" suffix="]"/>
+      </else-if>
       <else-if type="article-newspaper broadcast" match="any">
         <date variable="issued" delimiter=" ">
           <date-part name="year"/>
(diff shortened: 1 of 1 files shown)


university-college-lillebaelt-vancouver.csl
```diff
@@ -270,7 +278,7 @@
   </citation>
   <bibliography et-al-min="7" et-al-use-first="6" second-field-align="flush">
     <layout>
-      <text variable="citation-number" suffix="."/>
+      <text variable="citation-number" suffix=". "/>
```
adam3smith: on which reference manager are you testing this? The space here really shouldn't be necessary.

rmzelle: Because of the `second-field-align="flush"`?

adam3smith: yes -- or am I missing something? I though if anything the space could mess with things once you get into higher numbers (i.e. 3 digits)

rmzelle: No, makes sense. Just making sure I follow your logic.

roaldfrosig: As the style is now (without the space), if you insert a bibliography with
only a few references, and later add more, the bibliography will look like
this after the tenth citation:


[image: Inline image 2]
I was rather puzzled about this, especially because the standard Vancouver
style (the one that Zotero is 'born' with) doesn't have this problem. So I
checked how that csl was written, and noticed that the citation number had
". " as suffix rather than just "."

I guess the real problem is, that the size of the hanging indent is decided
when the bibliography is made, and not adjusted later. Adding a space is an
unelegant fix, and it probably won't work if there is more than 99
citations, but the style is meant for bachelor students, and it is very
rare that they reach that kind of citation count in their papers.

I noticed that if you change citation style in your document, and then
change it back to Vancouver (or any other style with
second-field-align=flush I guess) the bibliography will realign. It's what
I would do, if I just used the style for myself. But I'm trying to 'sell'
Zotero to our students, and want it to work as seamlessly as possible for
them.

adam3smith: That makes sense, though it's not ideal on the Zotero/citeproc-js side of things. @fbennett -- is this something citeproc does or do we need to look at the Word add-ons?

fbennett: The makeBibliography() function delivers a full list of citations, and a note of the offset. IIRC, partial updates based on the list are handled in Zotero in integration.js.

PR #1755, opened 2015-11-11
Update et-al settings for journal-of-infectious-diseases.csl

rmzelle:
Requested via email, confirmed at http://www.oxfordjournals.org/our_journals/jid/for_authors/ms_prep.html.

### DIFF
--- a/journal-of-infectious-diseases.csl
+++ b/journal-of-infectious-diseases.csl
@@ -29,23 +29,6 @@
       <term name="section" form="short">sect.</term>
     </terms>
   </locale>
-  <locale xml:lang="fr">
-    <date form="text" delimiter=" ">
-      <date-part name="day"/>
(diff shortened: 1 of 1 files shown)


rmzelle:
Uhm, it looks like this PR resulted in two Travis builds (one for the new branch and one for a (virtual) merge?), which triggered two csl-bot "Good news" posts?

https://travis-ci.org/citation-style-language/styles/builds/90541506
https://travis-ci.org/citation-style-language/styles/builds/90541638

@inukshuk?

inukshuk:
Yes, this is normal... but Sheldon responds only to requests which are part of a PR, which the first one isn't. Also if you look at the first build there is no mention of the issue number (1755) -- without this Sheldon couldn't have posted here, so I think this was caused by something else.

PR #1757, opened 2015-11-12
Update harvard-university-of-leeds.csl

jesusbagpuss:
Most changes suggested by David Kane:
http://csl.mendeley.com/styles/25843611/harvard-university-of-leeds-2015-16 

Also updated link to 2015/16 style changes.

### DIFF
--- a/harvard-university-of-leeds.csl
+++ b/harvard-university-of-leeds.csl
@@ -36,7 +36,7 @@
   </macro>
   <macro name="author">
     <names variable="author">
-      <name and="text" delimiter-precedes-last="never" et-al-min="0" initialize-with="." name-as-sort-order="all"/>
+      <name and="text" delimiter-precedes-last="never" initialize-with="." name-as-sort-order="all"/>
       <label form="short" prefix=" (" suffix=")."/>
       <substitute>
         <names variable="editor"/>
@@ -127,7 +127,7 @@
       </group>
     </layout>
   </citation>
-  <bibliography et-al-min="3" et-al-use-first="1" hanging-indent="true">
+  <bibliography hanging-indent="true">
     <sort>
       <key macro="author"/>
       <key variable="title"/>
@@ -156,9 +156,7 @@
           <group prefix=" ">
             <text term="in" text-case="capitalize-first" font-style="italic" suffix=":"/>
             <names variable="editor" delimiter=", " prefix=" " suffix=".">
-              <name and="text" delimiter-precedes-last="never" initialize-with=". ">
-                <name-part name="family" text-case="capitalize-first"/>
-              </name>
+              <name and="text" delimiter-precedes-last="never" initialize-with=". "/>
               <label form="short" strip-periods="false" prefix=", "/>
             </names>
             <group prefix=" " suffix="." delimiter=" ">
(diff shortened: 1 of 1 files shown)


harvard-university-of-leeds.csl
```diff
@@ -32,7 +36,7 @@
   </macro>
   <macro name="author">
     <names variable="author">
-      <name and="text" delimiter-precedes-last="never" initialize-with="." name-as-sort-order="all"/>
+      <name and="text" delimiter-precedes-last="never" et-al-min="0" initialize-with="." name-as-sort-order="all"/>
```
adam3smith: what's the `et-al-min="0"` doing? That's not a number that makes sense.

adam3smith: if you mean to list all authors in the bibliography, set the et-al-min in the bibliography line to some arbitrary high number (99 or so)

jesusbagpuss: Hi, I was sent the updates by someone else - I'll check what is intended in the bibliography and let you know.
What's the behaviour when ```et-al-min``` isn't defined?

adam3smith: it first falls back to what's in the bibliography, then to what's in style, then it doesn't do et al ever. But it appears to be more robust to just set it to a high number, which is why I'm recommending that.

harvard-university-of-leeds.csl
```diff
@@ -153,7 +157,7 @@
             <text term="in" text-case="capitalize-first" font-style="italic" suffix=":"/>
             <names variable="editor" delimiter=", " prefix=" " suffix=".">
               <name and="text" delimiter-precedes-last="never" initialize-with=". ">
-                <name-part name="family" text-case="uppercase"/>
+                <name-part name="family" text-case="capitalize-first"/>
```
rmzelle: @adam3smith, isn't it also rather uncommon to change the capitalization of names to "capitalize-first"?

harvard-university-of-leeds.csl
rmzelle: > But it appears to be more robust to just set it to a high number

That's just because it prevents all authors to be printed for silly papers with thousands of authors, right? (see http://www.nature.com/news/physics-paper-sets-record-with-more-than-5-000-authors-1.17567 )

adam3smith: Yes, that, and it also means you don't inherit a different setting set at a higher level.

harvard-university-of-leeds.csl
adam3smith: yeah -- most of the time this does nothing; in some very rare cases it probably does things you don't want (like bell Hooks or e.e. Cummings)

rmzelle: @jesusbagpuss, can you remove the "text-case" attribute here?

harvard-university-of-leeds.csl
jesusbagpuss: @adam3smith - I've just been clarifying this.
The in-text citation should have 'et al' if there are 3 or more authors, and in which case only the first author should be listed.

The change should actually be to the bibliography section.
The guidance is
> **All author names should be included in the reference.**
> Where a source has a very long list of authors, eg in the case of some scientific articles, 
> you might wish to consult your tutor on whether to use "et al" in place of some author names.

Trying to code 'ask your tutor' into a CSL is a bit like trying to knit fog.
Our Endnote style has a limitless author list for the bibliography, as much as I think it's silly, I think the same should be applied here!

harvard-university-of-leeds.csl
jesusbagpuss: I've just been talking to our referencing gurus about this. The CSL file I was supplied with was edited in Mendeley - and I suspect the 'none' (or however it's described) wasn't appealing to the person updating the style.
So yes, I think this should be removed.

rmzelle:
Great, thanks.

PR #1758, opened 2015-11-12
Create technische-universitat-dresden-finanzwirtschaft-und-finanzdien…

jmiba:
…stleistungen-note.csl

### DIFF
--- a/technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-note.csl
+++ b/technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-note.csl
@@ -16,8 +16,9 @@
     </contributor>
     <category citation-format="note"/>
     <category field="social_science"/>
+    <link href="http://www.zotero.org/styles/harvard1" rel="template"/>
     <summary>Zitierstil entsprechend den Zitierrichtlinien des Lehrstuhls Finanzwirtschaft und
-Finanzdienstleistungen der Fakultät Wirtschaftswissenschaften, Technische Universität Dresden. Stil basiert auf dem Zitierstil Harvard Reference format 1 (author-date)</summary>
+Finanzdienstleistungen der Fakultät Wirtschaftswissenschaften, Technische Universität Dresden.</summary>
     <updated>2015-11-12T18:05:38+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
--- a/technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-note.csl
+++ b/technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-note.csl
@@ -15,7 +15,7 @@
       <email>julian.onions@gmail.com</email>
     </contributor>
     <category citation-format="note"/>
-    <category field="generic-base"/>
+    <category field="social_science"/>
     <summary>Zitierstil entsprechend den Zitierrichtlinien des Lehrstuhls Finanzwirtschaft und
 Finanzdienstleistungen der Fakultät Wirtschaftswissenschaften, Technische Universität Dresden. Stil basiert auf dem Zitierstil Harvard Reference format 1 (author-date)</summary>
     <updated>2015-11-12T18:05:38+00:00</updated>
--- a/technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-note.csl
+++ b/technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-note.csl
@@ -6,6 +6,7 @@
     <id>http://www.zotero.org/styles/technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-note</id>
     <link href="http://www.zotero.org/styles/technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-note" rel="self"/>
     <link href="http://www.zotero.org/styles/harvard1" rel="template"/>
+    <link href="https://bildungsportal.sachsen.de/opal/url/RepositoryEntry/683835403" rel="documentation"/>
     <author>
       <name>Jens Mittelbach</name>
       <email>jens.mittelbach@slub-dresden.de</email>
@@ -18,9 +19,15 @@
     <category citation-format="note"/>
     <category field="social_science"/>
     <summary>Zitierstil entsprechend den Zitierrichtlinien des Lehrstuhls Finanzwirtschaft und Finanzdienstleistungen der Fakultät Wirtschaftswissenschaften, Technische Universität Dresden.</summary>
-    <updated>2015-11-13T11:46:45+00:00</updated>
+    <updated>2015-11-14T10:58:54+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
+  <locale xml:lang="de">
+    <terms>
+      <term name="available at">Internet</term>
+      <term name="accessed">Stand</term>
+    </terms>
+  </locale>
   <macro name="editor">
     <names variable="editor" font-variant="normal" delimiter=", ">
       <name font-variant="small-caps" delimiter="; " et-al-min="0" initialize="false" name-as-sort-order="all"/>
@@ -53,7 +60,6 @@
     </names>
   </macro>
   <macro name="access">
-    <text macro="delimiter"/>
     <choose>
       <if variable="URL">
         <text term="available at" suffix=": "/>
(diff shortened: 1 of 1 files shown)


technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-note.csl
```diff
@@ -0,0 +1,209 @@
...
+      <uri>http://www.slub-dresden.de</uri>
+    </author>
+    <contributor>
+      <name>Julian Onions</name>
+      <email>julian.onions@gmail.com</email>
+    </contributor>
+    <category citation-format="note"/>
+    <category field="generic-base"/>
+    <summary>Zitierstil entsprechend den Zitierrichtlinien des Lehrstuhls Finanzwirtschaft und
+Finanzdienstleistungen der Fakultät Wirtschaftswissenschaften, Technische Universität Dresden. Stil basiert auf dem Zitierstil Harvard Reference format 1 (author-date)</summary>
```
zuphilip: Add
```
<link href="http://www.zotero.org/styles/harvard1" rel="template"/>
```
above and you can delete the last sentence here. Moreover, you can then delete the contributor as well, because he is credited in the linked style.

technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-note.csl
```diff
@@ -0,0 +1,209 @@
...
+      <name>Jens Mittelbach</name>
+      <email>jens.mittelbach@slub-dresden.de</email>
+      <uri>http://www.slub-dresden.de</uri>
+    </author>
+    <contributor>
+      <name>Julian Onions</name>
+      <email>julian.onions@gmail.com</email>
+    </contributor>
+    <category citation-format="note"/>
+    <category field="generic-base"/>
```
zuphilip: I think we normally use `social_science` also for economics here, at least that would be consistent with with your other style https://www.zotero.org/styles/technische-universitat-dresden-wirtschaftswissenschaften?source=1 . I suppose that these two chairs use different citaiton styles also they are at the same departement, correct?

jmiba: I changed to social_science. You are right: the chairs in this department use their own styles. It's insane.

zuphilip: Yeah, it is the same here and I agree with your comment about that.

Ping @adam3smith or @rmzelle for continuation from here

technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-note.csl
```diff
@@ -0,0 +1,243 @@
...
+          <text macro="year-date"/>
+        </group>
+        <group>
+          <label suffix=" " variable="locator" form="short"/>
+          <text variable="locator"/>
+        </group>
+      </group>
+    </layout>
+  </citation>
+  <bibliography et-al-use-first="0" hanging-indent="true">
```
adam3smith: remove the `et-al-use-first` here. I don't think it will do anything in most reference managers, but it's a bit unclear what it _should_ do, so I'd rather not have it in the style.

technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-note.csl
```diff
@@ -0,0 +1,243 @@
...
+              <label suffix=" " variable="page" form="short"/>
+              <text variable="page"/>
+            </group>
+          </group>
+        </else>
+      </choose>
+      <text macro="access" suffix="."/>
+    </layout>
+  </bibliography>
+  <locale xml:lang="de">
```
adam3smith: while this is perfectly valid, could you move this section (i.e. the locale) right behind the info section, i.e. after line 23? That's where it's in all other styles.

technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-note.csl
```diff
@@ -0,0 +1,243 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="note" version="1.0" demote-non-dropping-particle="sort-only" xmlns="http://purl.org/net/xbiblio/csl">
+  <info>
+    <title>Technische Universität Dresden - Finanzwirtschaft und Finanzdienstleistungen (note)</title>
+    <title-short>TUD Betriebswirtschaftslehre (note)</title-short>
+    <id>http://www.zotero.org/styles/technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-note</id>
+    <link href="http://www.zotero.org/styles/technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-note" rel="self"/>
```
adam3smith: Does the Lehrstuhl document that style online somewhere? We really like to have links to documentation where at all possible. If so, include as `rel="documentation"` link.

adam3smith:
See my three comments inline. After addressing those, we're good to merge.

technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-note.csl
jmiba: It does, but the problem is it is not openly accessible. Should I include the link anyway?

technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-note.csl
jmiba: Will do that.

technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-note.csl
jmiba: I'll remove that.

rmzelle:
@inukshuk, it might be nice to do something about repeated comments. Would it be possible for the csl-bot to delete its one comments, so that it e.g. only keeps the latest "Good news" or "Bad news" comment.

technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-note.csl
rmzelle: @jmiba, please delete

```
    <contributor>
      <name>Julian Onions</name>
      <email>julian.onions@gmail.com</email>
    </contributor>
```

rmzelle:
@jmiba, will this style only be used in German? If so, the style should be set to German (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#3---style-locale for documentation).

jmiba:
@rmzelle, this style could also be used in other languages. Some students and researchers at this chair do write their papers in English.

inukshuk:
@rmzelle we could change the notification schedule to `on_success: change` and `on_failure: always` (or perhaps `change` too?) -- this would at least omit the a string of successive success messages.

We can look into deleting old comments, but I think this might be rather more confusing than helpful.

technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-note.csl
```diff
@@ -0,0 +1,229 @@
...
+      <name font-variant="small-caps" delimiter="; " et-al-min="0" initialize="false" name-as-sort-order="all"/>
+      <label form="short" font-variant="normal" prefix=" (" suffix="):"/>
+    </names>
+  </macro>
+  <macro name="anon">
+    <text term="anonymous" form="short" text-case="capitalize-first" strip-periods="true"/>
+  </macro>
+  <macro name="author">
+    <names variable="author" font-variant="normal">
+      <name font-variant="small-caps" delimiter="; " delimiter-precedes-last="never" et-al-min="0" initialize="false" initialize-with="." name-as-sort-order="all"/>
```
adam3smith: if you could remove the `et-al-min="0"` here, too? (and above in line 29). This applies to all your current style submissions. From my side that's the last issue, everything else looks good.

adam3smith:
@rmzelle these are all good to merge from my side. Let me know if I should wait or if we're good to go here.

rmzelle:
> we could change the notification schedule to on_success: change and on_failure: always (or perhaps change too?) -- this would at least omit the a string of successive success messages.

@inukshuk, yeah, that would already help a lot.

rmzelle:
@jmiba, okay, the styles don't need a `default-locale`, then. This one should be good to merge.

inukshuk:
@rmzelle I was just going to do that, but we can't as this would seriously impact the distribution updater (since we can't set the schedule individually for each web hook).

rmzelle:
@inukshuk, yeah, I was afraid of that. Created a ticket at Travis-CI (https://github.com/travis-ci/travis-ci/issues/5121).

rmzelle:
@inukshuk, based on http://docs.travis-ci.com/user/notifications/#Webhooks-Delivery-Format, I guess we could only suppress comments for `"status_message": "Still Failing"` right now.

rmzelle:
Thanks, @jmiba!

PR #1762, opened 2015-11-14
technische-universitat-dresden-finanzwirtschaft-und-finanzdienstleistungen-author-date

jmiba:
Yet another style of the chair Finanzwirtschaft at TU Dresden


rmzelle:
Do you have some example references that show how this style differs from #1761? They both print the short title in the in-text citation, but this one doesn't print it again in the bibliography?

jmiba:
@rmzelle, this is a mistake. The short title should not be printed in the inline citation. I will change that.

rmzelle:
Okay, looks good now.

adam3smith:
thanks!

PR #1764, opened 2015-11-16
Omit "medium" field in APA blog citations.

bgradin:
According to the Publication Manual of the American Psychological Association, 6th Ed., p. 215, the general citation form for internet message boards, electronic mailing lists, and other online communities (including blogs) is:

```
Author, A. A. (Year, Month Day). Title of Post [Description of form]. Retrieved from http://www.xxxx
```

Currently, if a citation with a `type` of `post-weblog` or `webpage` has a `medium` specified, the medium will be outputted in brackets after the genre, which is also in brackets.

This is awkward aesthetically, but aside from that, a medium is probably not needed in blog or webpage citations, since in general, they will already have a description of form (genre) specified.

### DIFF
--- a/apa.csl
+++ b/apa.csl
@@ -244,6 +244,7 @@
           </group>
           <text macro="edition"/>
         </group>
+        <text macro="format" prefix=" [" suffix="]"/>
       </if>
       <else-if type="post-weblog webpage" match="any">
         <text variable="genre" prefix=" [" suffix="]"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
I guess my question is why you would ever have medium and genre for a website or weblog in the data? I think we're doing this as is to accommodate various mappings from different reference managers. What are you working with?

PR #1765, opened 2015-11-21
Update earthquake-engineering-and-structural-dynamics.csl

bbbear2002:
Revise "Newspaper" citation.
Before:
Pear R. Crisis puts tax moves into play. The New York Times 2008; .
After:
Pear R. Crisis puts tax moves into play. The New York Times 2008.

### DIFF
--- a/earthquake-engineering-and-structural-dynamics.csl
+++ b/earthquake-engineering-and-structural-dynamics.csl
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" page-range-format="expanded" initialize-with-hyphen="false" demote-non-dropping-particle="sort-only" default-locale="en-US">
-  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+<!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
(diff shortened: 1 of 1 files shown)


earthquake-engineering-and-structural-dynamics.csl
```diff
@@ -91,7 +91,7 @@
     </layout>
   </citation>
   <bibliography et-al-min="7" et-al-use-first="6" second-field-align="flush" entry-spacing="0">
-    <layout suffix=".">
+    <layout suffix=". ">
```
adam3smith: could you remove the space here? It's probably not going to make a difference in many circumstances, but it doesn't make any sense to put a space after the final period.

earthquake-engineering-and-structural-dynamics.csl
```diff
@@ -148,8 +148,8 @@
                 <text variable="container-title" text-case="title" strip-periods="true" font-style="italic"/>
               </group>
               <group>
-                <text macro="year-date" suffix="; "/>
-                <text variable="volume" font-weight="bold"/>
+                <text macro="year-date"/>
+                <text variable="volume" font-weight="bold" prefix="; "/>
                 <text variable="issue" prefix="(" suffix=")"/>
```
adam3smith: I think this will actually print the semicolon in bold, which, while hard to see, isn't right. Could you use

```
<group delimiter="; ">
   <text macro="year-date"/>
   <text variable="volume" font-weight="bold"/>
</group>
```

instead?

PR #1774, opened 2015-11-23
Update soziale-welt.csl

frederik-elwert:
Made small changes to better reflect the author guidelines of the journal.
(In agreement with the original author of the style.)


adam3smith:
great, thanks!

PR #1777, opened 2015-11-25
Create Cladistics.csl

vidovics:
Cladistics journal reference style. The reference style is similar to Elsevier Harvard, with italicized et al. for in-text reference. The bibliography has no italics, there is a "." at the end of the author list and the date is out of brackets. There is a "." after the journal title.

### DIFF
--- a/Cladistics.csl
+++ b/Cladistics.csl
@@ -1,0 +1,254 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" 
+
+version="1.0" demote-non-dropping-particle="never" default-locale="en-US">
+  <info>
+    <title>Cladistics</title>
+    <id>http://www.zotero.org/styles/Cladistics</id>
+    <link href="http://www.zotero.org/styles/cladistics" rel="self"/>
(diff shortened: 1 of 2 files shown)


rmzelle:
Do you have a version of this style without all these extra line breaks? E.g.

```
    <link href="http://onlinelibrary.wiley.com/journal/10.1111/%28ISSN
%291096-0031/homepage/ForAuthors.html" rel="documentation"/>
```

should be

```
    <link href="http://onlinelibrary.wiley.com/journal/10.1111/%28ISSN%291096-0031/homepage/ForAuthors.html" rel="documentation"/>
```

rmzelle:
Great, thanks.

Please take a look at https://github.com/citation-style-language/styles/wiki/Style-Requirements. In particular, the style file name should become "cladistics.csl" (lowercase; see also https://help.github.com/articles/renaming-a-file/ ). In addition, `<category field="generic-base"/>` should be deleted, and the "template" link should point to the CSL style you used as a starting point (if there is one, otherwise remove it). We also would like to see the eISSN included (`<eissn>1096-0031</eissn>`).

rmzelle:
Perfect, thanks!

vidovics:
Thanks for your help.

PR #1780, opened 2015-11-27
Minor fixes to the style

Bibref:
Updates the access macro to include DOI. Also modified the number of pages variable for thesis so it appears. Updated the date variable for newspaper articles so it appears between ( and ).

### DIFF
--- a/ets-ecole-de-technologie-superieure.csl
+++ b/ets-ecole-de-technologie-superieure.csl
@@ -122,21 +122,28 @@
     <text variable="issue" prefix=" "/>
   </macro>
   <macro name="access">
-    <group delimiter=". ">
-      <text variable="URL" prefix="&lt;" suffix="&gt;"/>
-      <choose>
-        <if variable="URL" match="any">
-          <group delimiter=" ">
-            <text value="Consulté le"/>
-            <date variable="accessed">
-              <date-part name="day" suffix=" "/>
-              <date-part name="month" suffix=" "/>
-              <date-part name="year"/>
-            </date>
-          </group>
-        </if>
-      </choose>
-    </group>
+		<choose>
+			<if variable="DOI" match="any">
+				<group delimiter=". ">
+					<text variable="DOI" prefix="&lt;http://dx.doi.org/" suffix="&gt;"/>
+					<date variable="accessed" delimiter=" " prefix="Consulté le ">
+						<date-part name="day"/>
+						<date-part name="month"/>
+						<date-part name="year"/>
+					</date>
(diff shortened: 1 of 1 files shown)


ets-ecole-de-technologie-superieure.csl
```diff
@@ -122,21 +122,28 @@
...
-            <date variable="accessed">
-              <date-part name="day" suffix=" "/>
-              <date-part name="month" suffix=" "/>
-              <date-part name="year"/>
-            </date>
-          </group>
-        </if>
-      </choose>
-    </group>
+		<choose>
```
adam3smith: we'd much rather you have the style with two spaces indenting rather than tabs. (All styles on the repo follow this convention). If you're preferred editor doesn't do that for you, we actually have a formatter tool that will do it for you:

ets-ecole-de-technologie-superieure.csl
```diff
@@ -122,21 +122,28 @@
...
+			<if variable="DOI" match="any">
+				<group delimiter=". ">
+					<text variable="DOI" prefix="&lt;http://dx.doi.org/" suffix="&gt;"/>
```
adam3smith: this is fine, but if you have any input on the style guidelines, doi.org actually prefers `http://doi.org`, i.e. without the dx. part, now, so I'd encourage you to use that (though we'll accept either)

ets-ecole-de-technologie-superieure.csl
```diff
@@ -162,7 +169,14 @@
             <text variable="genre"/>
             <text variable="publisher-place"/>
             <text variable="publisher"/>
-            <text variable="page" form="short" suffix=" p"/>
+            <choose>
```
adam3smith: indenting here is fine, so not sure what happened above.

adam3smith:
Please fix the indenting as per above. You can update your pull request by using the pencil icon at the top right here: https://github.com/citation-style-language/styles/pull/1780/files

ets-ecole-de-technologie-superieure.csl
Bibref: Sorry about the indenting. I always seem to have a problem with that when I copy my modifs. into GitHub. It might be my text editor giving me troubles... I will check into it for next time.

ets-ecole-de-technologie-superieure.csl
Bibref: Great, I didn't know about that. I modified that part. Thanks!

rmzelle:
Thanks!

PR #1782, opened 2015-11-30
sunway-college-johor-bahru-2015.csl

SunwayCollege:
Guide to Harvard Referencing by Sunway College Johor Bahru

### DIFF
--- a/sunway-college-johor-bahru-2015.csl
+++ b/sunway-college-johor-bahru-2015.csl
@@ -1,0 +1,333 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" default-locale="en-GB" version="1.0" demote-non-dropping-particle="sort-only">
+  <!--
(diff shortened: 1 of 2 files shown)


SunwayCollege:
Thank you

adam3smith:
@rmzelle -- all good from my side, but I figured you may have thoughts about naming.

rmzelle:
@SunwayCollege, thanks for contributing!

Are multiple versions of your style guide in use right now? We prefer to host only the most recent version, and leave out the "2015" from the style title and ID.

Does SCJB also have a style guide for a numeric style, or do you only use the Harvard-based author-date system?

SunwayCollege:
Hi admin

"2015" is removed from style title and ID
SCJB uses Harvard- based author-date system.

Thank you

rmzelle:
Thanks. Finally, do you care a lot about using "Harvard" to describe the style? At CSL we greatly prefer "author-date" (since Harvard styles don't have a real link to the real Harvard).

SunwayCollege:
Hi

The college do emphasis on "author-date" system. However, students are not able to grasp the concept. Thus, the lecturers will use "harvard" referencing  term to guide the students.

rmzelle:
Okay, thanks!

PR #1790, opened 2015-12-01
Proposal: lluelles-noibid.csl

f-mb:
I propose to rename lluelles-lexisnexis.csl as lluelles-noibid.csl as, basically, this is the Lluelles Style but with no Ibid. It has been used by LexisNexis Québec for years, but more and more people rely on it. I think it would be better to rename it as the "Guide Lluelles - No Ibid".

With this pull request:
- lluelles-lexisnexis.csl would become lluelles-noibid.csl 
- and a dependent style referring to lluelles-noibid.csl would be create for LexisNexis Québec, with the former ID lluelles-lexisnexis.csl.

I am not sure this is how I/we should do that... so open to any suggestion!

Best,

Florian

### DIFF
--- a/dependent/lluelles-lexisnexis.csl
+++ b/dependent/lluelles-lexisnexis.csl
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" default-locale="fr-CA">
+  <info>
+    <title>LexisNexis Québec (Guide Lluelles, no Ibid., French - Canada)</title>
+    <title-short>LexisNexis Québec</title-short>
+    <id>http://www.zotero.org/styles/lluelles-lexisnexis</id>
+    <link href="http://www.zotero.org/styles/lluelles-lexisnexis" rel="self"/>
+    <link href="http://www.zotero.org/styles/lluelles-no-ibid" rel="independent-parent"/>
+    <link href="http://f-mb.github.io/cslegal/" rel="documentation"/>
+    <category citation-format="note"/>
+    <category field="law"/>
+    <summary>Ce style est adapté aux règles de rédaction des publications québécoises en langue française de LexisNexis Canada : notes complètes, sans Ibid., basé sur la 7e édition du Guide des références pour la rédaction juridique (Guide Lluelles), et sur lluelles.csl et lluelles-noibid.csl.</summary>
+    <updated>2014-05-15T12:00:00+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+</style>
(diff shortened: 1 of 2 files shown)


f-mb:
I guess this was not the right way to do it... Travis failed... I guess because of the IDs mixing...

lluelles-lexisnexis.csl
```diff
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" default-locale="fr-CA">
   <info>
-    <title>LexisNexis Québec (Guide Lluelles modifié, French - Canada)</title>
-    <id>http://www.zotero.org/styles/lluelles-lexisnexis</id>
-    <link href="http://www.zotero.org/styles/lluelles-lexisnexis" rel="self"/>
+    <title>Guide des références pour la rédaction juridique 7e édition (Notes complètes) (Guide Lluelles, No Ibid, French - Canada)</title>
+    <id>http://www.zotero.org/styles/lluelles-noibid</id>
```
adam3smith: you'd also have to change the filename of this one to correspond to the ID. Instructions are the same as for any other edit.

rmzelle: Per our existing "no ibid" styles, please use "no Ibid." in the title (no capitalized "No", and a period behind "Ibid", since it's an abbreviation). For the style name, ID, and "self" link, please use "-no-ibid" instead of "-noibid".

dependent/lluelles-lexisnexis.csl
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" default-locale="fr-CA">
+  <info>
+    <title>LexisNexis Québec (Guide Lluelles modifié, French - Canada)</title>
+    <title-short>LexisNexis Québec</title-short>
+    <id>http://www.zotero.org/styles/lluelles-lexisnexis</id>
+    <link href="http://www.zotero.org/styles/lluelles-lexisnexis" rel="self"/>
+    <link href="http://www.zotero.org/styles/lluelles-noibid" rel="independent-parent"/>
+    <link href="http://f-mb.github.io/cslegal/" rel="documentation"/>
```
rmzelle: Do you want to update this to http://f-mb.org/cslegal/?

f-mb:
Sorry! 
I fixed it and merged all commits into one.
(CSLégal webpage is also up to date.)

rmzelle:
(what I meant was, you use "http://f-mb.github.io/cslegal/" in the styles, but this automatically redirects to "http://f-mb.org/cslegal/", so I was wondering if you preferred the latter; not that it matters much)

f-mb:
I prefer the github,io link, as this is the permanent and actual location of the document.

rmzelle:
Okay, sure. Thanks!

PR #1791, opened 2015-12-03
Purh

NMonteix:
Before merging, there is a problem to be solved, that could be linked with Zotero: 
If a document is only available online (shows an URL) and has pages indicated (e.g. an electronic book which chapters has been published as PDF), it seems impossible to obtain both the URL and the pages in bibliography. Without pages, the URL is shown; with pages, only those are shown, the URL disappears.
Is it a behaviour linked with the CSL file as it is?


changer-d-epoque.csl
```diff
@@ -0,0 +1,356 @@
...
+            <text term="ibid" text-case="capitalize-first" font-style="italic"/>
+          </else-if>
+          <else-if position="subsequent">
+            <group delimiter=", ">
+              <text macro="author"/>
+              <text variable="title" form="short" font-style="italic"/>
+              <group delimiter=" ">
+                <text term="cited" font-style="italic"/>
+                <text value="supra" font-style="italic"/>
+                <text term="note" form="short" suffix=" $$"/>
```
adam3smith: this seems wrong -- that would give you the extra field in Zotero and the note field in Mendeley. 
I'm guessing you want `first-reference-note-number`? (and please confirm that the $$ is on purpose and not a forgotten placeholder).

adam3smith:
The URL is Zotero behavior, nothing in the CSL style. By default, Zotero treats URL="false" when articles have a page range. That behavior can be toggled using the "include URL" checkbox in the Cite/Styles tab of the preferences.

Please fix the error I note in line, then this is good to merge from my side.

changer-d-epoque.csl
NMonteix: In fact, I wanted not to hard code "n." as a recall for the first reference note, hence the term="note"
Testing it with the editor tool, it seems to work, it gives a text such as "op. cit. supra n. $$," the $$ being an on purpose reminder to give the first note where the work is cited.

If first-reference-note-number works with Zotero and Word it would be perfect. How is it supposed to be coded? It only appears in the specs under the variables, not anywhere else

NMonteix:
Ok
Thanks for the tip

changer-d-epoque.csl
adam3smith: oh sorry, I read that as variable, not term, so that makes more sense.
But yes, `first-reference-note-number`  works well with Zotero or Mendeley and word (doesn't work in Pandoc). 
There's nothing special to it, you just use it like any other variable, i.e. something like

```
<group delimiter=" ">
  <text term="note" form="short"/>
</group>

will give you "n. 23" (assuming this was first cited in the 23rd footnote, obviously).

NMonteix: Brilliant!
I will make the correction and might change a few things once the editor will have answered me.
I will keep you posted once everything will have been done
Thanks!

PR #1793, opened 2015-12-03
Two new styles for Zeitschrift für Religionswissenschaft.

frederik-elwert:
These styles follow the guidelines for the [Zeitschrift für Religionswissenschaft](http://www.degruyter.com/view/j/zfr).


frederik-elwert:
The automated testing is incredibly useful! Just two remarks:

* The styles first had a file name containing umlauts, which were rejected by travis. These file names were in fact suggested and inserted by the [visual CSL editor](http://editor.citationstyles.org/). Maybe that tool should be updated to make file name suggestions in line with the automated tests.
* Travis also rejected having two styles with the same ISSN. However, the respective journal explicitly allows for two different citation systems (author-year and notes). Thus, I think I was rightfully assigning the same ISSN to the two styles.

rmzelle:
> Thus, I think I was rightfully assigning the same ISSN to the two styles.

Yeah, it's fine to use the same ISSN twice in this case. We can set up an exception for the journal instead: https://github.com/citation-style-language/styles/blob/master/spec/spec_helper.rb#L8 (we mainly check to make sure people don't just copy a style from a different journal and forget to update the ISSN).

adam3smith:
1) we'l fix, see ticket above.
2) we can add to a skip list. I'll do that and add the ISSN back in for you, it's easier with test running locally to be sure to get that right.

adam3smith:
What should we do with the language here: It looks like this is almost entirely German but they accept non-German articles? What's the actual ratio here? Are we talking 1 in 100, 1 in 10 or 1 in 2 in languages other than German? 

If most folks are going to be using this in German, we should just put the default locale in (and, in that case, add German to the style title without changing the filename, id as in `(note, German)`.

rmzelle:
@frederik-elwert, also a few requests/questions:

* will this style be used in more than one language, or just German?
* please rename/adjust the styles to use "zeitschrift-fur-religionswissenschaft-author-date" instead of "zeitschrift-fur-religionswissenschaft-author-year", and "zeitschrift-fur-religionswissenschaft-note" instead of "zeitschrift-fur-religionswissenschaft-notes" (that's more consistent with the rest of our repository styles)
* please remove prior style authors/contributors, and add a "template" link instead (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link )
* `<issn>2194-508X</issn>` should be `<eissn>2194-508X</eissn>`. You can also add the print ISSN (`<issn>0943-8610</issn>`; see https://www.wzb.eu/en/library/collections-search/e-journals?page=detail.phtml&bibid=WZB&colors=7&lang=en&jour_id=186193 )

(PS. looks like @adam3smith and me are jumping on this pull request at the same time :) )

frederik-elwert:
Thanks for the comments!

> What should we do with the language here: It looks like this is almost entirely German but they accept non-German articles? What's the actual ratio here? Are we talking 1 in 100, 1 in 10 or 1 in 2 in languages other than German?

Most are German, but a significant minority is in English (~ 1 in 10). Does that still qualify for default-locale and the title addition? I’ll just follow your advice.

I fixed the rest, thanks for the hints. I added the eissn and fixed the issn only in the -note style, in order to avoid a test failure. Should I leave it to you to copy both to the -author-date one as well?

rmzelle:
> Most are German, but a significant minority is in English (~ 1 in 10). Does that still qualify for default-locale and the title addition? I’ll just follow your advice.

Our options are either to force the current two styles to German, and add "dependent" styles for English (these would just be aliases to the German ones, but with a different locale), or to just leave things as there are, in which case the author needs to select the right locale when using the style. Since we already have two variants for this journal we might want to do the latter (otherwise we end up with four styles, which is a bit much).

> Should I leave it to you to copy both to the -author-date one as well?

Just go ahead and add them to the other style as well. Then also add both ISSNs to https://github.com/frederik-elwert/styles/blob/patch-1/spec/spec_helper.rb (change the line `  1546-1726 2108-6419 0035-2969 1958-5691` to `  1546-1726 2108-6419 0035-2969 1958-5691 0943-8610 2194-508X`). That should allow the test report to pass.

frederik-elwert:
> Our options are either to force the current two styles to German, and add "dependent" styles for English (these would just be aliases to the German ones, but with a different locale), or to just leave things as there are, in which case the author needs to select the right locale when using the style. Since we already have two variants for this journal we might want to do the latter (otherwise we end up with four styles, which is a bit much).

Yes, I agree. I’d leave it as it is, and authors should set their locale themselves.

> Just go ahead and add them to the other style as well.

Done.

zeitschrift-fur-religionswissenschaft-note.csl
```diff
@@ -0,0 +1,1235 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="never" page-range-format="chicago">
+<!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>Zeitschrift für Religionswissenschaft (notes)</title>
```
rmzelle: @frederik-elwert, can you change this to `<title>Zeitschrift für Religionswissenschaft (note)</title>`?

zeitschrift-fur-religionswissenschaft-author-date.csl
```diff
@@ -0,0 +1,262 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="chicago">
+<!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>Zeitschrift für Religionswissenschaft (author-year)</title>
```
rmzelle: @frederik-elwert, "author-date", please.

rmzelle:

adam3smith:
yup. Thanks!

rmzelle:
@frederik-elwert, thanks for contributing!

frederik-elwert:
Thanks for the helpful feedback!

rmzelle:
We're also happy you like the automatic test reports! We've only been using the @csl-bot posts for a little over a month now. Let us know if you can think of any improvements (we're still hoping to be able to make the Travis error reports more readable).

PR #1802, opened 2015-12-07
Update article Encyclopedia citation in SYCJB

SunwayCollege:
Update article encyclopedia citation in SYCJB

### DIFF
--- a/sunway-college-johor-bahru.csl
+++ b/sunway-college-johor-bahru.csl
@@ -4,12 +4,15 @@
 Sunway College Johor Bahru (Harvard-based author-date system) citation style using CSL1.0.1 (http://citationstyles.org)
 
 NOTES
-
-- This citation style is based on the January 2013 version of the SCJB style guide and Roadmap to Referencing (http://www.sunway.edu.my/jb/images/library/Sunway%20JB%20Guide%20to%20Harvard%20Referencing_20150529.pdf).
+20 Nov 2015
+- This citation style is based on the May 2015 version of the SCJB style guide and Roadmap to Referencing (http://www.sunway.edu.my/jb/images/library/Sunway%20JB%20Guide%20to%20Harvard%20Referencing_20150529.pdf).
 - Resource type is assumed to be electronic if both URL and Accessed date are filled out for an item.
 - This style was tested with Zotero 4.0.8 (http://www.zotero.org/).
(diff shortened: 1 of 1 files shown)


sunway-college-johor-bahru.csl
```diff
@@ -149,10 +152,15 @@ TO DO
   </macro>
   <macro name="title">
     <choose>
-      <if type="chapter paper-conference article-journal article-magazine article-newspaper broadcast interview manuscript map personal_communication speech thesis map post post-weblog" match="none">
+    <!-- encyclopedia type -->
+	  <if type="entry-encyclopedia article-newspaper">
+		<text variable="title" prefix="&apos;" suffix="&apos;"/>
```
rmzelle: It is better to do this with CSL quotes:

`<text variable="title" quotes="true"/>`

(see http://docs.citationstyles.org/en/stable/specification.html#quotes)

SunwayCollege:
Hi

If used 

    <text variable="title" quotes=true/>

The result will be (Global Warming 2015)

It is not the same as SYCJB guideline.

That why we used

    <text variable="title" prefix"&apos;" suffix="&apos;"/>

The result will be ('Global Warming' 2015).
It is single-quote not double quote mark as according to guideline

Thank you

sunway-college-johor-bahru.csl
```diff
@@ -149,10 +152,16 @@ TO DO
   </macro>
   <macro name="title">
     <choose>
-      <if type="chapter paper-conference article-journal article-magazine article-newspaper broadcast interview manuscript map personal_communication speech thesis map post post-weblog" match="none">
+    <!-- encyclopedia type  newspaer-->
+	  <if type="entry-encyclopedia article-newspaper">
```
adam3smith: even if this may currently work correctly in Zotero or Mendeley, this isn't right. you want
`<if type="entry-encyclopedia article-newspaper" match="any">`
otherwise, the match condition should default to match="all", which in this case is impossible (and item can't be two item types at the same time.

adam3smith:
I take it the single-quote issue is set? The code you ave should give yo proper single quotes.

adam3smith:
Thanks!
Future submission, please run through http://formatter.citationstyles.org/ to get all the indenting standardized prior to submission. Makes reviewing easier.

PR #1805, opened 2015-12-07
Add styles for "Collection de l'Ecole française de Rome"

NMonteix:
One thing at a time... First the ISSN problem
Sorry for the mess


rmzelle:
Great. So now you can add the style to https://github.com/NMonteix/styles/tree/patch-1 (either by checking out this branch and adding the files, or by using the "New file" button at this link).

NMonteix:
I don't understand the Travis report: the specified macro is used in the lay-out. Which would be the unused macro?

rmzelle:
Looks good, except "collection-de-l-ecole-francaise-de-rome-note.csl" has an unused macro "locator-cit" which is not allowed.

P.S. You're looking at https://github.com/NMonteix/styles/blob/patch-1/collection-de-l-ecole-francaise-de-rome-note.csl, right? I don't see a call to "locator-cit" there.

adam3smith:
only in the full-note version, not in the note version.

NMonteix:
Sorry, my mistake, I read too quickly

rmzelle:
You're getting another error since the unused "locator-cit" macro was the only one to call the "locator" macro, so you'll have to delete the latter one too. (the testing tool isn't smart enough to catch this in a single go)

rmzelle:

rmzelle:
(oh, and it would be better to change the title of this PR first, which I'll do)

collection-de-l-ecole-francaise-de-rome-full-note.csl
```diff
@@ -0,0 +1,319 @@
...
+        <text macro="pages"/>
+      </group>
+    </macro>
+    <citation disambiguate-add-year-suffix="true" et-al-subsequent-min="2" et-al-subsequent-use-first="1">
+      <layout delimiter=" ; ">
+        <choose>
+          <if position="subsequent">
+            <group delimiter=", ">
+              <text macro="author"/>
+              <text variable="title" form="short" font-style="italic" suffix="..."/>
```
adam3smith: you actually mean to put this here? If so, I'd use the ellipsis character rather than three single periods, i.e. … rather than ... (looks the same, but try selecting. Has the advantage of being non-breakable).

collection-de-l-ecole-francaise-de-rome-full-note.csl
```diff
@@ -0,0 +1,319 @@
...
+        <text variable="container-title" font-style="italic" suffix=","/>
+        <text macro="publisher" suffix=","/>
+        <group delimiter=" " suffix=",">
+          <text macro="year-date"/>
+          <text macro="collection"/>
+        </group>
+        <text macro="pages"/>
+      </group>
+    </macro>
+    <citation disambiguate-add-year-suffix="true" et-al-subsequent-min="2" et-al-subsequent-use-first="1">
```
adam3smith: you really mean to set et-al-min at 2? That's extremely odd, no?

adam3smith:
two minor issue inline. Please fix or respond, otherwise this looks good.

collection-de-l-ecole-francaise-de-rome-full-note.csl
NMonteix: I tried with the ASCII code &#133; for the ellipsis but obtain a blank 
square in Word, so I got back to three dots
What would be the correct code?

collection-de-l-ecole-francaise-de-rome-full-note.csl
NMonteix: Indeed... I would say the whole idea of a full note with a call back to 
a previous note is odd, but modern and contemporary historians seem to 
prefer this to a sharp author-note, maybe because they mainly cite 
archives and not so much litterature.
Given the first citation gives all of the names and the reader can find 
this first citation with the indication of the note, I thought 
interesting to gain space in notes, it is more easier for reading. The 
editor doesn't seem to have specification regarding this.

collection-de-l-ecole-francaise-de-rome-full-note.csl
rmzelle: I always just copy/paste my funky characters from wikipedia: "…" (from https://en.wikipedia.org/wiki/Ellipsis)

rmzelle: And `&#133;` seems to be "Unicode Character 'NEXT LINE (NEL)' (U+0085)" (http://www.fileformat.info/info/unicode/char/0085/index.htm), so that's not right.

rmzelle: `&#8230;` probably? http://www.fileformat.info/info/unicode/char/2026/index.htm

NMonteix: Pasted from adam3smith 's message
Thanks for the tip though

NMonteix: So http://www.nationalfinder.com/html/char-asc.htm being wrong...

collection-de-l-ecole-francaise-de-rome-full-note.csl
adam3smith: OK, just to be clear , you're getting something like 
Smith et al. instead of Smith et Marx, yes? It's specifically the 2 that I'm wondering about, because that's also gramatically wrong for et al. (which stands for and other*s*)

NMonteix: Agreed...
Shifting to 3
But, yes, I had Smith et al. instead of Smith et Marx

rmzelle: > It's specifically the 2 that I'm wondering about, because that's also gramatically wrong for et al. (which stands for and others)\

Is it? Singular for "alia" is "alius", so that would still abbreviate the same. :P (https://en.wiktionary.org/wiki/alius)

NMonteix: Both in Latin and for the second author, it would indeed be better 
considering /et al/. as the abreviation for /et alii/. For the second 
and last author's sake (and some respect for Latin grammar), I have 
changed...
Hence I guess the struggle for being first named in hard science...

rmzelle:
Thanks!

collection-de-l-ecole-francaise-de-rome-full-note.csl
adam3smith: For completeness sake:

et al. abbreviates et alii, not et alia (which is "and other things"). My
Latin is a bit rusty, but IIRC, et alius makes just as little sense in
Latin as "and other" would make in English.

On Mon, Dec 7, 2015 at 1:58 PM, Rintze M. Zelle <notifications@github.com>
wrote:

> In collection-de-l-ecole-francaise-de-rome-full-note.csl
> <https://github.com/citation-style-language/styles/pull/1805#discussion_r46868791>
> :
>
> > +    </macro>
> > +    <macro name="chapter">
> > +      <group delimiter=" ">
> > +        <text macro="title" quotes="true" suffix=","/>
> > +        <text term="in"/>
> > +        <text macro="editor" suffix=","/>
> > +        <text variable="container-title" font-style="italic" suffix=","/>
> > +        <text macro="publisher" suffix=","/>
> > +        <group delimiter=" " suffix=",">
> > +          <text macro="year-date"/>
> > +          <text macro="collection"/>
> > +        </group>
> > +        <text macro="pages"/>
> > +      </group>
> > +    </macro>
> > +    <citation disambiguate-add-year-suffix="true" et-al-subsequent-min="2" et-al-subsequent-use-first="1">
>
> It's specifically the 2 that I'm wondering about, because that's also
> gramatically wrong for et al. (which stands for and others)\
>
> Is it? Singular for "alia" is "alius", so that would still abbreviate the
> same. :P (https://en.wiktionary.org/wiki/alius)
>
> —
> <https://github.com/citation-style-language/styles/pull/1805/files#r46868791>
> .
>



-- 
Sebastian Karcher, PhD
www.sebastiankarcher.com

NMonteix:
To you!

rmzelle:
@NMonteix, every once in a while I standardize the indentation of all the styles in the repository, and I noticed some of your recent styles used four spaces for the first indentation level instead of two (see e.g. https://github.com/citation-style-language/styles/commit/83a675ff9f08509ae897d1312168c955815dcebb#diff-8e6816a90312c664c7b2cced8a96e787 ). Could you use 2-spaces-per-level for any future styles? Thanks!

(you should be able to use our formatter to quickly get correct indenting: http://formatter.citationstyles.org/ )

NMonteix:
@rmzelle, I will (I guess I used styles coded some time ago as 
templates, starting from my files and not the reindented version; I will 
reindent them)

rmzelle:
Great, thanks.

PR #1806, opened 2015-12-07
Update to us-geological-survey style

kknierim:
This update changes the report citation type to include a URL and access date,
following recent requirements by the USGS.
Note: the addition may not correctly ignore urls when one is not provided with the report

### DIFF
--- a/us-geological-survey.csl
+++ b/us-geological-survey.csl
@@ -132,24 +132,26 @@
           </group>
         </else-if>
         <else-if type="webpage">
-          <text variable="title" prefix=" " suffix=","/>
-          <group prefix=" ">
-            <text term="accessed"/>
-          </group>
-          <group prefix=" ">
-            <date variable="accessed">
-              <date-part name="month" form="long" suffix=" "/>
-              <date-part name="day" form="numeric" suffix=", "/>
-              <date-part name="year" form="long"/>
-            </date>
-          </group>
-          <group prefix=" ">
-            <text term="at"/>
-            <text variable="container-title" prefix=" "/>
-          </group>
-          <group prefix=" ">
-            <text term="at"/>
-            <text variable="URL" prefix=" "/>
+          <group prefix=" " delimiter=", ">
+            <text variable="title"/>
+            <group delimiter=" ">
+              <text term="accessed"/>
+              <date variable="accessed">
+                <date-part name="month" form="long" suffix=" "/>
+                <date-part name="day" form="numeric" suffix=", "/>
+                <date-part name="year" form="long"/>
+              </date>
+            </group>
+            <group delimiter=" ">
+              <group delimiter=" ">
+                <text term="at"/>
+                <text variable="container-title"/>
+              </group>
+              <group delimiter=" ">
+                <text term="at"/>
+                <text variable="URL"/>
+              </group>
+            </group>
           </group>
         </else-if>
         <else-if type="chapter paper-conference" match="any">
--- a/us-geological-survey.csl
+++ b/us-geological-survey.csl
@@ -96,7 +96,7 @@
       </group>
       <choose>
         <if type="report">
-          <group prefix=" " delimiter=", " suffix=".">
+          <group prefix=" " delimiter=", " suffix=".,">
             <group delimiter=" ">
               <text variable="title" suffix=":"/>
               <text variable="publisher"/>
@@ -107,6 +107,24 @@
               <text variable="page" suffix=" "/>
               <label variable="page" form="short" plural="never"/>
             </group>
+          </group>
+          <group prefix=" ">
+            <text term="accessed"/>
+          <group prefix=" " suffix=",">
+            <date variable="accessed">
+              <date-part name="month" form="long" suffix=" "/>
+              <date-part name="day" form="numeric" suffix=", "/>
+              <date-part name="year" form="long"/>
+            </date>
+          </group>
+          <group prefix=" ">
+            <text term="at"/>
+            <text variable="container-title" prefix=" " suffix=" "/>
+          </group>
+          <group prefix=" " suffix=".">
+            <text term="at"/>
+            <text variable="URL" prefix=" "/>
+          </group>
           </group>
         </if>
         <else-if type="bill book graphic legal_case legislation motion_picture report song" match="any">
--- a/us-geological-survey.csl
+++ b/us-geological-survey.csl
@@ -107,27 +107,27 @@
               <text variable="page" suffix=" "/>
               <label variable="page" form="short" plural="never"/>
             </group>
-          <group prefix=" ">
-            <text term="accessed"/>
-          <group prefix=" " suffix=",">
-            <date variable="accessed">
-              <date-part name="month" form="long" suffix=" "/>
-              <date-part name="day" form="numeric" suffix=", "/>
-              <date-part name="year" form="long"/>
-            </date>
-          </group>
-          <group prefix=" ">
-            <text term="at"/>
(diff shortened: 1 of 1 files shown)


us-geological-survey.csl
```diff
@@ -96,7 +96,7 @@
       </group>
       <choose>
         <if type="report">
-          <group prefix=" " delimiter=", " suffix=".">
+          <group prefix=" " delimiter=", " suffix=".,">
```
rmzelle: `suffix=".,"` seems wrong.

kknierim:
Yea, I was trying to use the <group> info to correctly display the punctuation. The citations needs to look like this (as an example):

Adamski, J.C., Celik, I., Camas, H., Arslan, O., Yegin, E., Kufrevioglu, O.I., Colombo, J.C., Bilos, C., Campanaro, M., and Rodriguez Presa, M.J., 1997, Nutrients and pesticides in ground water of the Ozark Plateaus in Arkansas, Kansas, Missouri, and Oklahoma: U.S. Geological Survey 96-4313, 34 p., accessed May 8, 2014, at http://pubs.usgs.gov/wri/wri96-4313/.

So that the citation reads "blah blah XX p., accessed Month Day, Year, at url."  Note the comma after the page (p) and the second comma after the date.

I bullied the code so it works, but probably not the correct way to do it.

us-geological-survey.csl
adam3smith: We'd want this handled with a delimiter, so you don't have a report without URL ending on `.,.` which currently could happen. Remove the suffix entirely (including the period). 
Then remove `</group>` from line 110 and place it after line 128 instead. Please test to make sure I haven't gotten something mixed up.

adam3smith: otherwise this looks good. Instructions for updating the pull request above.

adam3smith:
OK, last one, then we're set (thanks!): Could you make sure the indenting is correct? Most easily by using our dedicated tool:

kknierim:
Yes, I will check the indenting. Before that, I am wondering if I have an extra period... seems I get this with the code:

damski, J.C., Celik, I., Camas, H., Arslan, O., Yegin, E., Kufrevioglu, O.I., Colombo, J.C., Bilos, C., Campanaro, M., and Rodriguez Presa, M.J., 1997, Nutrients and pesticides in ground water of the Ozark Plateaus in Arkansas, Kansas, Missouri, and Oklahoma: U.S. Geological Survey Water-Resources Investigations Report 96-4313, 34 p., accessed May 8, 2014, at http://pubs.usgs.gov/wri/wri96-4313/. .

Note the extra space and period at the end. Not sure where to fix that?

zuphilip:
Delete the period here in the suffix should fix your problem:
```
 <else-if type="webpage">
   ...
          <group prefix=" " suffix=".">
            <text term="at"/>
            <text variable="URL" prefix=" "/>
          </group>
```

kknierim:
Thanks @zuphilip . Looks like your suggestion is for a webpage, but my example is from a report (apologies that I wasn't more clear). I think the webpage type works fine (doesn't include an extra period).

zuphilip:
The same group occurs also in the `report` case, but the strange indentation makes it really hard to read. You should also correct the webpage thing because that works AFAIK because the processor is forgiveful.

us-geological-survey.csl
```diff
@@ -107,6 +107,24 @@
...
+            <text variable="container-title" prefix=" " suffix=" "/>
+          </group>
+          <group prefix=" " suffix=".">
+            <text term="at"/>
+            <text variable="URL" prefix=" "/>
+          </group>
+          </group>
           </group>
         </if>
         <else-if type="bill book graphic legal_case legislation motion_picture report song" match="any">
```
zuphilip: BTW can you delete `report` here? This case is handeld already above. (I know that this wasn't your change.)

kknierim:
Goodness, yes. Apologies on the weird indenting. I ran it through the CSL Style Formatter for this commit and made changes as @zuphilip suggested. Hopefully this will make it easier to read. Appreciate the help.

kknierim:
And wondering if the suffix term should just be removed completely... so not even including it as ```suffix = " "``` , but just removing the suffix term from the report and web page?

Another question, should "report" be in line 50: 
           ```<if type="bill book graphic legal_case legislation motion_picture report song" match="any">```

adam3smith:
Good calls. Yes, remove those suffixes for both. 
For line 50, all that code makes little sense in the context of the style, pretty sure it's just a leftover. Replace lines 49 to 56 by `<text variable="title" form="short"/>`.

kknierim:
Ok, I removed the extra lines that @adam3smith suggested and tested the code; it works well for reports and produces the following examples:
Adamski, J.C., Celik, I., Camas, H., Arslan, O., Yegin, E., Kufrevioglu, O.I., Colombo, J.C., Bilos, C., Campanaro, M., and Rodriguez Presa, M.J., 1997, Nutrients and pesticides in ground water of the Ozark Plateaus in Arkansas, Kansas, Missouri, and Oklahoma: U.S. Geological Survey Water-Resources Investigations Report 96-4313, 34 p., accessed May 8, 2014, at http://pubs.usgs.gov/wri/wri96-4313/.
Macfarlane, P.A., and Hathaway, L.R., 1987, The hydrogeology and chemical quality of ground waters from the lower Paleozoic aquifers in the Tri-State region of Kansas, Missouri, and Oklahoma: Kansas Geological Survey 9, 44 p.

Note that when a URL is missing, the punctuation ends correctly (thanks for that help!).

I worked on the webpage type to try and get it to match the report type; I fixed the formatting of the date so it matches USGS requirements. I can't seem to get a "," after the date though. See the following example:
National Atmospheric Deposition Program, 2014, NADP/NTN Monitoring Location AR27, accessed May 9, 2014 at http://nadp.sws.uiuc.edu/sites/siteinfo.asp?id=AR27&net=NTN.

Note there is not a comma after the date and before that "at...". When I try to get the comma in there, I end up breaking the code, and am not sure I'm adding a comma in the right spot. 
Suggestions?

rmzelle:
You can try changing

```xml
        <else-if type="webpage">
          <text variable="title" prefix=" " suffix=","/>
          <group prefix=" ">
            <text term="accessed"/>
          </group>
          <group prefix=" ">
            <date variable="accessed">
              <date-part name="month" form="long" suffix=" "/>
              <date-part name="day" form="numeric" suffix=", "/>
            </date>
          </group>
          <group prefix=" ">
            <text term="at"/>
            <text variable="container-title" prefix=" "/>
          </group>
          <group prefix=" ">
            <text term="at"/>
            <text variable="URL" prefix=" "/>
          </group>
        </else-if>
```

to

```xml
        <else-if type="webpage">
          <group prefix=" " delimiter=", ">
            <text variable="title"/>
            <group delimiter=" ">
              <text term="accessed"/>
              <date variable="accessed">
                <date-part name="month" form="long" suffix=" "/>
                <date-part name="day" form="numeric" suffix=", "/>
              </date>
            </group>
            <group delimiter=" ">
              <group delimiter=" ">
                <text term="at"/>
              </group>
              <group delimiter=" ">
                <text term="at"/>
                <text variable="URL"/>
              </group>
            </group>
          </group>
        </else-if>
```

rmzelle:
(basically, every time you use a term like "at" or "accessed", you'll want to use a group so that the term doesn't get printed if the variable in the group doesn't hold a value. Then there is just one more group for the ", " delimiters)

kknierim:
Thanks @rmzelle. That worked! I tested the code in Zotero, and it works great for reports and webpages.

rmzelle:
Thanks!

PR #1808, opened 2015-12-08
Update apa.csl

agruber:
The "name" and "label" commands were not applied to the substitute variables "editor" and "translator".


adam3smith:
No, that's a bug in the reference manager you're using. Those should be inherited in substitute. Which software is this?

agruber:
This happens with citeproc.js version 1.1.7.

rmzelle:
That's a somewhat outdated version from April, see https://bitbucket.org/fbennett/citeproc-js/downloads.

rmzelle:
And while this might be a real issue for you, it's not a style problem, so closing this. If you can reproduce this with the latest version of citeproc-js, the bug should be reported to that project.

PR #1810, opened 2015-12-08
I've changed volume/number of volume

smarchand:
I have fixed a troubleshooting on volume/number of volumes

### DIFF
--- a/ens-de-lyon-centre-d-ingenierie-documentaire.csl
+++ b/ens-de-lyon-centre-d-ingenierie-documentaire.csl
@@ -19,7 +19,7 @@
     <category field="social_science"/>
     <category field="philosophy"/>
     <category field="literature"/>
-    <updated>2015-12-01T08:18:35+00:00</updated>
+    <updated>2015-12-08T14:56:57+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="fr">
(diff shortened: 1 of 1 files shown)


adam3smith:
Would you mind running this through the formatter at http://formatter.citationstyles.org/ and updating your pull request as described above? I'd like the indenting to be correct.

smarchand:
I've made the reindentation as you requested.
thanks

adam3smith:
thanks!

ens-de-lyon-centre-d-ingenierie-documentaire.csl
```diff
@@ -173,9 +173,16 @@
             <date-part name="year"/>
           </date>
           <group>
-            <text term="volume" form="short" suffix=".&#160;"/>
-            <text variable="volume"/>
-            <text variable="number-of-volumes" prefix="/"/>
+            <choose>
+              <if variable="volume" match="none">
+                <text variable="number-of-volumes" suffix="&#160;vol."/>
```
rmzelle: @smarchand, in general it's better to use CSL terms (and groups) whenever possible, by the way. Here you could have used

```xml
<group delimiter="&#160;">
  <text variable="number-of-volumes"/>
</group>
```

instead of

`<text variable="number-of-volumes" suffix="&#160;vol."/>`

We can leave this as is, but something to keep in mind for future edits. (using terms make it possible to automatically translate styles into other languages)

smarchand: thanks . I will try that right away !

2015-12-10 6:07 GMT+01:00 Rintze M. Zelle <notifications@github.com>:

> In ens-de-lyon-centre-d-ingenierie-documentaire.csl
> <https://github.com/citation-style-language/styles/pull/1810#discussion_r47189463>
> :
>
> > @@ -173,9 +173,16 @@
> >              <date-part name="year"/>
> >            </date>
> >            <group>
> > -            <text term="volume" form="short" suffix=".&#160;"/>
> > -            <text variable="volume"/>
> > -            <text variable="number-of-volumes" prefix="/"/>
> > +            <choose>
> > +              <if variable="volume" match="none">
> > +                <text variable="number-of-volumes" suffix="&#160;vol."/>
>
> @smarchand <https://github.com/smarchand>, in general it's better to use
> CSL terms (and groups) whenever possible, by the way. Here you could have
> used
>
> <group delimiter="&#160;">
>   <text variable="number-of-volumes"/>
>   <text term="volume" form="short"/>
> </group>
>
> instead of
>
> <text variable="number-of-volumes" suffix="&#160;vol."/>
>
> We can leave this as is, but something to keep in mind for future edits.
> (using terms make it possible to automatically translate styles into other
> languages)
>
> —
> <https://github.com/citation-style-language/styles/pull/1810/files#r47189463>
> .
>



-- 
Stéphane

PR #1814, opened 2015-12-10
minor fixes vancouver-fr-ca.csl

f-mb:
- Add full support for graphic and motion_picture
- Fix bug with editions
- Fix bus with authors, book authors, book editors


rmzelle:
Could you reindent the style with 2 spaces per level (e.g. with http://formatter.citationstyles.org/)? You're also introducing some empty lines which will be stripped later anyway, so it would be better to remove those as well.

f-mb:
Done. Sorry.

rmzelle:
Great, thanks!

PR #1815, opened 2015-12-10
Create the-chemical-society-of-japan.csl

shojitakahashi:
This style is for the two journals published by The Chemical Society of Japan.
Bulletin of the Chemical Society of Japan
Chemistry Letters The dependent styles for these journals will be submitted separately.

Shoji Takahashi
Solution Consultant (e-Platform and Content), A&G
Elsevier Japan KK
E-mail: s.takahashi@elsevier.com

### DIFF
--- a/the-chemical-society-of-japan.csl
+++ b/the-chemical-society-of-japan.csl
@@ -2,7 +2,7 @@
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
   <info>
     <title>The Chemical Society of Japan</title>
-    <id>http://www.zotero.org/styles/styles/the-chemical-society-of-japan</id>
+    <id>http://www.zotero.org/styles/the-chemical-society-of-japan</id>
     <link href="http://www.zotero.org/styles/the-chemical-society-of-japan" rel="self"/>
     <link href="http://www.csj.jp/journals/styles/ref.html" rel="documentation"/>
     <author>
--- a/the-chemical-society-of-japan.csl
+++ b/the-chemical-society-of-japan.csl
@@ -1,0 +1,189 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>The Chemical Society of Japan</title>
+    <id>http://www.zotero.org/styles/styles/the-chemical-society-of-japan</id>
+    <link href="http://www.zotero.org/styles/the-chemical-society-of-japan" rel="self"/>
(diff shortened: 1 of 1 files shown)


the-chemical-society-of-japan.csl
```diff
@@ -0,0 +1,189 @@
...
+      <else>
+        <text variable="page"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="publisher">
+    <text variable="publisher" suffix=", "/>
+    <text variable="publisher-place"/>
+  </macro>
+  <citation collapse="citation-number">
```
adam3smith: You'll want to add

```
<sort>
</sort>
```
here, too. Otherwise you could be getting something like `5,3` for subsequent citations.

the-chemical-society-of-japan.csl
```diff
@@ -0,0 +1,189 @@
...
+        <label suffix=" " variable="page" form="short"/>
+        <text variable="page"/>
+      </else-if>
+      <else>
+        <text variable="page"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="publisher">
+    <text variable="publisher" suffix=", "/>
```
adam3smith: could we do this in a group instead, please. That'll make it more robust, e.g. for missing places. I.e.

```
<group delimiter=", ">
  <text variable="publisher"/>
</group>
```

the-chemical-society-of-japan.csl
shojitakahashi: Thanks for your advice.  I made the change.

the-chemical-society-of-japan.csl
shojitakahashi: Thanks for your advice.  I made the change.

rmzelle:

the-chemical-society-of-japan.csl
```diff
@@ -0,0 +1,194 @@
...
+      <label form="verb-short" plural="never" suffix=" "/>
+      <name initialize="false" initialize-with=". "/>
+    </names>
+  </macro>
+  <macro name="locators">
+    <choose>
+      <if type="article-journal" match="any">
+        <number font-style="italic" variable="volume"/>
+      </if>
+      <else-if type="chapter" match="any">
```
adam3smith: You'll actually want to place a group around lines 79-82, too. That way, the terms are only printed if you have the variable. Also, I'd prefer to use text rather than number at least for volume so that we're a bit more flexible with how volumes look (they're normally numbers, but I'm sure we can find number/letter combination in some contexts).

adam3smith:
@shojitakahashi see my last comment above, sorry I missed that before. Then we're good to merge.

the-chemical-society-of-japan.csl
shojitakahashi: I made the necessary change.  Hope it will be OK this time.

adam3smith:
great, thanks. Now go ahead and fix that filename on the other pull request.

PR #1816, opened 2015-12-10
Create bulletin-of-the-chemical-society-of-japan.csl

shojitakahashi:
This is a dependent style for the-chemical-society-of-japan.csl

### DIFF
--- a/dependent/bulletin-of-the-chemical-society-of-japan.csl
+++ b/dependent/bulletin-of-the-chemical-society-of-japan.csl
@@ -2,9 +2,9 @@
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
   <info>
     <title>Bulletin of the Chemical Society of Japan</title>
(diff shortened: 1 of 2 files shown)


adam3smith:
The filename for this one should be bulletin-of-the-chemical-society-of-japan.csl -- but actually hold off on making that change: this will fail Travis tests until we've accepted your other pull request for the parent style. Once that's done, you renaming the file here will trigger another run of the tests which should then be fine.

shojitakahashi:
Thanks.  I will wait.  I think the filename is already correct (bulletin-of-the-chemical-society-of-japan.csl).  Am I looking at a different thing?

adam3smith:
You probably mean the id. the filename is this here and it's currently still set to the-chemical-society-of-japan.csl . But as I said, hold off on the fix, please

screenshot from 2015-12-10 10 28 19

adam3smith:
thanks!

rmzelle:
(in general, it's preferable to update the "documentation" link for dependent styles; for this style, it could be http://www.journal.csj.jp/bcsj/submission/manuscript-preparation#mp-raa )

PR #1817, opened 2015-12-10
Create chemistry-letters.csl

shojitakahashi:
This is a dependent style for the-chemical-society-of-japan.csl.


adam3smith:
and thanks again!

PR #1818, opened 2015-12-11
Create analytical-sciences.csl

shojitakahashi:


shojitakahashi:
I looked at my codes again and again but I cannot figure out what is wrong.  Could you please identify what is wrong with my code?  Hope this message (before the validation succeeds) will be brought to your attention...

rmzelle:
The problem is that you defined the "editor" macro twice, which the validate doesn't catch, while it apparently trips up the automatic Travis test setup.

adam3smith:
thanks!

rmzelle:
@inukshuk, would it be easy to provide a more information error for styles that use the same name for multiple macros? As you can see in https://travis-ci.org/citation-style-language/styles/builds/96259840, Travis currently only reports a failed parse but doesn't give a reason.

inukshuk:
@rmzelle I think this should be quite easy; will take a look.

rmzelle:
Great, thanks. We can catch it in validation with an additional Schematron rule, too, by the way.

PR #1819, opened 2015-12-11
Create analytical-sciences.csl

rmzelle:


adam3smith:
@rmzelle -- I'm good with this as is. I'm not quite sure what's going on with the two PRs, so I'll let you handle the merging.

rmzelle:
I created this to troubleshoot #1818 (I figured initially that it might be an encoding issue, since I didn't see anything wrong at first). I'll let Shoji fix up his original PR and then I'll close this.

PR #1822, opened 2015-12-12
Include entry-dictionary/entry-encyclopedia with book sections

adunning:
Generally speaking, entry-dictionary and entry-encyclopedia are conceived as a type of book section. These were latecomers to CSL, and do not seem to have been properly included in the appropriate macros for most styles. This is most notably manifested by printing the secondary contributors before the container title.

I have found that this can be fixed in all styles I have tested by replacing `"chapter paper-conference"` with `"chapter entry-dictionary entry-encyclopedia paper-conference"`, and this pull request applies this across the entire repository. I have not found any ill effects, but would strongly recommend further scrutiny.

For instance, if one runs this example through chicago-author-date.csl:

``` json
{"id":"test","type":"entry-encyclopedia","title":"Test Entry","container-title":"The Test Encyclopedia","publisher":"Citeproc","publisher-place":"Oxford","editor":[{"family":"Smith","given":"Adam"}],"author":[{"family":"Smith","given":"Adam"}],"issued":{"date-parts":[[2004]]}}
```

This is the result before these changes:

> Smith, Adam. 2004. “Test Entry.” Edited by Adam Smith. <em>The Test Encyclopedia</em>. Oxford: Citeproc.

Whereas this is expected:

> Smith, Adam. 2004. “Test Entry.” <em>The Test Encyclopedia</em>. Edited by Adam Smith. Oxford: Citeproc.


archeologie-medievale.csl
```diff
@@ -273,7 +273,7 @@
                   <text macro="publisher"/>
                 </group>
               </else-if>
-              <else-if type="chapter paper-conference" match="any">
+              <else-if type="chapter entry-dictionary entry-encyclopedia paper-conference" match="any">
```
zuphilip: That looks wrong because one of the cases before is
```
<else-if type="article-journal article-magazine article-newspaper broadcast personal_communication thesis entry-dictionary entry-encyclopedia" match="any">
```

zuphilip:
My first concern here is that if `entry-dictionary` or `entry-encyclopedia` are already used "somewhere" in the style before, then it might that we cannot add these two cases also to all occurences of the `"chapter paper-conference"`. It might be that they are missing on porpuse for some conditional formatting. The example I found above shows that problem. Thus, I would suggest to (at least) restrict only to the "clear" cases.

Second, it might (in theory) also be the possible that the cases `entry-dictionary` or `entry-encyclopedia` are missing on purpose, i.e. some formatting are only valid for `article-journal` and `chapter` but not for the other two cases. I don't know if this happening in practice at all or often.

adam3smith:
Thanks Andrew for doing this. 
I generally do think this is worthwhile, but I share @zuphilip's concerns, in particular no. 1. (I think no. 2, while certainly valid in theory, is indeed mostly theoretical. I don't think we have any styles where these would be treated as fallback-type items).

The simplest solution might be to only apply this fix where the dictionary and encyclopedia item types aren't otherwise explicitly mentioned in the style. That may be overly conservative, but we want to be overly conservative when applying batch-fixes without testing individual styles.

rmzelle:
(I'd like to note as well that citing encyclopedias and dictionaries is exceedingly rare for styles for the sciences, where this change would just add clutter)

rmzelle:
@adunning, based on the reactions above, I don't think there is a good way to batch correct the entire repository without checking individual styles by hand, so I'm closing this (just before the first anniversary of this ticket). We're still open to accept a more targeted fix (with would probably have to exclude all STEM journals, and address the concerns @zuphilip and @adam3smith expressed above).

adunning:
Sorry, never had the time to make the adjustments. Should probably still be fixed in Chicago, at least.

rmzelle:
> Generally speaking, entry-dictionary and entry-encyclopedia are conceived as a type of book section. These were latecomers to CSL

Just curious, but did you find when these item types were introduced? As far as I can tell, they already existed in the first numbered release, CSL 0.8, in 2009. So they're far from recent additions.

> Sorry, never had the time to make the adjustments. Should probably still be fixed in Chicago, at least.

New ticket created (see above).

PR #1827, opened 2015-12-13
Create yakugaku-zasshi.csl

shojitakahashi:
This is a dependent style of Chemical and Pharmaceutical Bulletin.

<title>Yakugaku Zasshi (Journal of the Pharmaceutical Society of Japan)</title>
The reason for the second name in parentheses is that the official name of this journal is "Yakugaku Zasshi" in Japanese but it also has an English name "Journal of the Pharmaceutical Society of Japan" too.
http://yakushi.pharm.or.jp/


rmzelle:
Thanks!

PR #1828, opened 2015-12-13
Update sist02.csl

shojitakahashi:
I created this style about a year ago.  It was my first time to create a new style with only one month of experience of using CSL Editor.
After a year of experience, I am more comfortable with CSL and have updated this style.
As a result, I have rewritten almost all the lines.
I believe this code is much cleaner than the old one, with more sophisticated use of groups, delimiters, etc.

.

### DIFF
--- a/sist02.csl
+++ b/sist02.csl
@@ -8,199 +8,133 @@
     <author>
       <name>Shoji Takahashi</name>
       <email>s.takahashi@elsevier.com</email>
-      <uri>http://www.mendeley.com/profiles/shoji-takahashi/</uri>
+      <uri>http://www.mendeley.com/profiles/shoji-takahashi3/</uri>
     </author>
     <contributor>
       <name>Eriko Amano</name>
       <uri>http://www.mendeley.com/profiles/eriko-amano/</uri>
     </contributor>
     <category citation-format="numeric"/>
-    <updated>2014-11-26T05:09:41+00:00</updated>
+    <updated>2015-12-13T06:04:50+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="ja">
     <terms>
-      <term name="chapter" form="short">ch.</term>
-      <term name="presented at">presented at the</term>
-      <term name="available at">available</term>
+      <term name="open-quote">“</term>
+      <term name="close-quote">”</term>
     </terms>
   </locale>
   <macro name="author">
     <names variable="author" suffix=". ">
       <name initialize="false" initialize-with=". " name-as-sort-order="all"/>
-      <label form="short" prefix=", " text-case="capitalize-first"/>
-      <substitute>
-        <names variable="translator"/>
-      </substitute>
     </names>
   </macro>
-  <macro name="edition">
-    <number prefix="第" variable="edition"/>
-    <text term="edition" form="short" suffix=", "/>
-  </macro>
   <macro name="event">
-    <text variable="event-place" suffix=", "/>
-    <date delimiter="" variable="event-date" suffix=". ">
-      <date-part name="year" range-delimiter="" suffix="-"/>
-      <date-part name="month" form="numeric-leading-zeros" range-delimiter="" suffix="-"/>
-      <date-part name="day" form="numeric-leading-zeros" range-delimiter=""/>
-    </date>
+    <group delimiter=", ">
+      <text variable="event-place"/>
+      <date variable="event-date">
+        <date-part name="year" suffix="-"/>
+        <date-part name="month" form="numeric-leading-zeros" suffix="-"/>
+        <date-part name="day" form="numeric-leading-zeros"/>
+      </date>
+    </group>
   </macro>
   <macro name="title">
     <choose>
-      <if type="article-journal" match="any">
-        <text variable="collection-title" suffix=": "/>
-        <text variable="title" suffix=". "/>
-        <text variable="container-title"/>
-      </if>
-      <else-if type="chapter paper-conference report" match="any">
+      <if type="chapter report paper-conference article webpage" match="any">
         <choose>
-          <if match="any" variable="container-title">
-            <text variable="title" prefix="&quot;" suffix="&quot;. "/>
-            <text variable="container-title" suffix=". "/>
+          <if match="any" variable="container-title" type="webpage">
+            <text term="open-quote"/>
+            <text variable="title" quotes="false"/>
+            <text term="close-quote"/>
           </if>
           <else>
-            <text variable="title" suffix=". "/>
+            <text variable="title"/>
           </else>
         </choose>
-      </else-if>
-      <else-if type="webpage" match="any">
-        <text variable="title" prefix="&quot;" suffix="&quot;. "/>
-        <text variable="container-title" suffix=". "/>
-      </else-if>
+      </if>
       <else>
-        <text variable="title" suffix=". "/>
+        <text variable="title"/>
       </else>
     </choose>
   </macro>
   <macro name="issued">
-    <choose>
-      <if type="book chapter" match="any">
-        <choose>
-          <if match="any" variable="page">
-            <date variable="issued" suffix=", ">
-              <date-part name="year" range-delimiter=""/>
-            </date>
-          </if>
-          <else>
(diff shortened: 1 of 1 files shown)


sist02.csl
```diff
@@ -8,199 +8,133 @@
...
-      </if>
-      <else-if type="chapter paper-conference report" match="any">
+      <if type="chapter report paper-conference article webpage" match="any">
         <choose>
-          <if match="any" variable="container-title">
-            <text variable="title" prefix="&quot;" suffix="&quot;. "/>
-            <text variable="container-title" suffix=". "/>
+          <if match="any" variable="container-title" type="webpage">
+            <text term="open-quote"/>
+            <text variable="title" quotes="false"/>
```
adam3smith: why aren't you just doing `quotes="true"` here instead?

sist02.csl
```diff
@@ -8,199 +8,133 @@
...
-            <label plural="never" suffix=" " variable="page" form="short"/>
-            <choose>
-              <if match="any" variable="collection-title">
-                <text variable="page" suffix=", "/>
-              </if>
-              <else>
-                <text variable="page" suffix="."/>
-              </else>
-            </choose>
+            <label plural="never" suffix=". " variable="page" form="short"/>
```
adam3smith: for consistency, remove the periods in the suffix here. It's not doing anything (since it's already in the term) and  you don't have it elsewhere.

adam3smith:
indeed a big improvement, thanks!

sist02.csl
shojitakahashi: Hi,

Thanks.  I thought you would catch this.  Let me explain.

What I am trying to get here is a title enclosed with quotes followed by a comma like below.
“TITLE”,

FYI, locales-ja-JP.xml has <style-options punctuation-in-quote="false"/>.

1.
With the current coding, I can get the desired result in CSL Editor, Mendeley Desktop, and Word (see the attached sist02-1.png).
I added the <locale xml:lang="ja"> section because locales-ja-JP.xml has Japanese parentheses 『 』.
<locale xml:lang="ja">
  <terms>
    <term name="open-quote">“</term>
    <term name="close-quote">”</term>
  </terms>
</locale>
…
<choose>
   <if match="any" variable="container-title" type="webpage">
      <text term="open-quote"/>
      <text variable="title" quotes="false"/>
      <text term="close-quote"/>
    </if>
    <else>
      <text variable="title"/>
      </else>
</choose>

2.
When you use quotes="true", you get the desired result in CSL Editor but not in Mendeley Desktop or Word (see attached sist02-2.png).
<locale xml:lang="ja">
  <terms>
    <term name="open-quote">“</term>
    <term name="close-quote">”</term>
  </terms>
</locale>
…
<choose>
  <if match="any" variable="container-title" type="webpage">
     <text variable="title" quotes="true"/>
  </if>
  <else>
     <text variable="title"/>
  </else>
</choose>

3.
I also tried the following, without the <locale xml:lang="ja"> section this time, but again I get the desired result in CSL Editor but not in Mendeley Desktop or Word (see attached sist02-3.png).
<choose>
  <if match="any" variable="container-title" type="webpage">
    <text variable="title" quotes="false" prefix="&quot;" suffix="&quot;"/>
  </if>
  <else>
     <text variable="title"/>
  </else>
</choose>

Hope this explains the background of my coding.

Best regards,

Shoji

Shoji Takahashi
Solution Consultant (e-Platform and Content), A&G
Elsevier Japan KK
Tel: +81-3-5561-5994
Fax: +81-3-5561-0451
Address:
Higashi-Azabu 1-Chome Bldg. 4F
1-9-15 Higashi-Azabu, Minato-ku, Tokyo
106-0044 JAPAN

sist02.csl
shojitakahashi: Hi,

Thanks.  I looked at this issue and found that the period included in the term is lost when the default-locale is set to ja-JP.  See the attached screenshot.

Best regards,

Shoji

sist02.csl
adam3smith: I'll take a look best as I can in Mendeley, but unless I'm missing something, that's really something they need to fix. Your 2.) is the way to go and should work reliably.

sist02.csl
adam3smith: That's very odd: 
https://github.com/citation-style-language/locales/blob/master/locales-ja-JP.xml#L172 has the period in the locale. You yourself fixed that more than a year ago. 

(btw. -- you can't e-mail screenshots. You'd need to drag them to the message here)

sist02.csl
shojitakahashi: Thanks.  Then, do you suggest that I would change the coding to 2) and then raise the problem with the Mendeley team?

Shoji

sist02.csl
shojitakahashi: Again, this might be an issue with Mendeley.  If having the period in the suffix does not do any harm as you said in your earlier message, will it be OK just leave my coding as it is for now?  I will raise the problem with the Mendeley team.

Shoji

adam3smith: fine to leave it in, but please do also report to Mendeley.

sist02.csl
adam3smith: I'd much prefer that, yes. I've double-checked 2.) in Zotero and it works correctly there, too. One possibility is that Mendeley is simply using an older version of citerproc-js, in which case this will fix itself, but would be good to check with them, yes. I'd much rather not have the workaround in the style, though.

sist02.csl
shojitakahashi: Thanks.  I will report this to Mendeley.

Shoji

adam3smith:
thanks!

sist02.csl
shojitakahashi: Thanks.  I just submitted the change.

I looked at what I was doing in my old style and below is what I did there.  I believe you (and I too) hate this!
<text variable="title" prefix="&quot;" suffix="&quot;. "/>

Shoji

adam3smith: that's right, I'm no fan of the hardcoded quotes at all, so this is much
better (it will also give you proper typographical curly quotes instead of
ugly straight ones).

On Sun, Dec 13, 2015 at 4:44 PM, Shoji Takahashi <notifications@github.com>
wrote:

> In sist02.csl
> <https://github.com/citation-style-language/styles/pull/1828#discussion_r47451753>
> :
>
> >          <choose>
> > -          <if match="any" variable="container-title">
> > -            <text variable="title" prefix="&quot;" suffix="&quot;. "/>
> > -            <text variable="container-title" suffix=". "/>
> > +          <if match="any" variable="container-title" type="webpage">
> > +            <text term="open-quote"/>
> > +            <text variable="title" quotes="false"/>
>
> Thanks. I just submitted the change. I looked at what I was doing in my
> old style and below is what I did there. I believe you (and I too) hate
> this! <text variable="title" prefix="&quot;" suffix="&quot;. "/> Shoji
> From: Sebastian Karcher [mailto:notifications@github.com] Sent: Monday,
> December 14, 2015 7:34 AM To: citation-style-language/styles Cc: Takahashi,
> Shoji (ELS-TOK) Subject: Re: [styles] Update sist02.csl (#1828
> <https://github.com/citation-style-language/styles/pull/1828>) In
> sist02.csl<#1828 (comment)
> <https://github.com/citation-style-language/styles/pull/1828#discussion_r47451593>
> >:
> <choose> - <if match="any" variable="container-title"> - <text
> variable="title" prefix="&quot;" suffix="&quot;. "/> - <text
> variable="container-title" suffix=". "/> + <if match="any"
> variable="container-title" type="webpage"> + <text term="open-quote"/> +
> <text variable="title" quotes="false"/>
> I'd much prefer that, yes. I've double-checked 2.) in Zotero and it works
> correctly there, too. One possibility is that Mendeley is simply using an
> older version of citerproc-js, in which case this will fix itself, but
> would be good to check with them, yes. I'd much rather not have the
> workaround in the style, though. — Reply to this email directly or view it
> on GitHub<
> https://github.com/citation-style-language/styles/pull/1828/files#r47451593
> >.
>
> —
> <https://github.com/citation-style-language/styles/pull/1828/files#r47451753>
> .
>



-- 
Sebastian Karcher, PhD
www.sebastiankarcher.com

sist02.csl
rmzelle: (so it might just be that Mendeley is using outdated CSL locale files)

adam3smith: I considered that, but the punctuation-in-quotes parameter has been in the locale since we moved them all to 1.0.1

PR #1830, opened 2015-12-14

zuphilip:
See https://forums.zotero.org/discussion/55443/style-error-medicine-science-in-sports-exercise/#Item_0


rmzelle:
Thanks!

PR #1833, opened 2015-12-15
Create angelaki.csl

### LINKED ISSUE #1171: Angelaki: Journal of the Theoretical Humanities
This is a Taylor & Francis journal, which is not in the list from the publisher but uses a standard style. = For the future add-list:

https://forums.zotero.org/discussion/12652/style-request-modern-language-association-note-with-bibliography/#Item_0

zuphilip:
Fixes #1171

### DIFF
--- a/dependent/angelaki.csl
+++ b/dependent/angelaki.csl
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" default-locale="en-US">
+  <info>
+    <title>Angelaki</title>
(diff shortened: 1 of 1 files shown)


zuphilip:
I haven't added the subtitle `Journal of the Theoretical Humanities` because the title `Angelika` doesn't look ambiguous for me. Or do you prefer to add the subtitle as well?

rmzelle:
Thanks. And yes, I try to avoid including subtitles at all cost :).

zuphilip:
Okay. I couldn't find anything about subtitle in the https://github.com/citation-style-language/styles/wiki/Style-Requirements

rmzelle:
I rather keep Style Requirements as compact as possible. This doesn't come up very often.

PR #1839, opened 2015-12-16
Create Management science.csl 

jimapps:


adam3smith:
Thanks! The CSL looks good. I'm wondering, though, what you based the style on? I saw the author-bib style they provide, but that looks like it's not what the journal actually follows. E.g. this is a recent article:
http://pubsonline.informs.org/doi/pdf/10.1287/mnsc.2014.1935
We also do have another style -- *Organization Science* -- that should follow the same outline. There are some small differences, but they're quite similar. 
Just trying to get a sense of what the best way forward is that will get us a correct style and the least amount of redundancy possible.

jimapps:
Thanks for your feedback. I think the required author-bib styles for both Management Science and Organization Science are the same. I didn't find that out. 

I guess sometimes authors don't strictly follow the guide (https://www.informs.org/content/download/17000/187215/file/Management-Science-template.zip).

rmzelle:
Was there any particular reason why you created a second pull request? #1837 and this one are identical.

jimapps:
I don't know how the pull request works. And the there were some failures in my first pull request.

jimapps:
Is there a way to include some customized abbreviations for journals? For example, INFORMS uses Management Sci. instead of Manag. Sci.

rmzelle:
> And the there were some failures in my first pull request.

It's possible to update a pull request once it has been created, so there is no need to create a new one if the automatic tests fail.

rmzelle:
@adam3smith, https://www.informs.org/Find-Research-Publications/INFORMS-Journals/Author-Portal/LaTeX-Style-Files says "One of the enclosed PDF files is a guideline to the INFORMS reference style and includes a comprehensive sample.", so it looks like all INFORMS journals share the same citation style.

> Is there a way to include some customized abbreviations for journals?

That depends on the reference manager (like Zotero, Mendeley, etc.) that you use the style with. In CSL, you can only tell the style to use full or abbreviated journal titles. It's not possible to define specific abbreviations.

jimapps:
@adam3smith, I check the current issues from management science, organization science, and marketing science. I find that INFORMS is using new style, as the link you provide.

jimapps:
@rmzelle, thanks for the suggestion and information on abbreviations.

adam3smith:
Actually, hold off on this for a moment. I'd first like to decide what the best way forward is

rmzelle:
Are you using our validator to make sure your style validates? It currently has some problems, see e.g. http://validator.citationstyles.org/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fjimapps%2Fstyles%2F5944a90f98af2848fd2c227fabe53ff8b5372e05%2Fmanagement-science.csl&version=1.0.1

jimapps:
@rmzelle Thanks. I change the style based on a different one.

rmzelle:
@jimapps, I see you based your latest version on organization-science.csl. Did you have to make any changes to the style format, or was it a perfect match?

adam3smith:
no, this is a significant re-write of organization-science -- if anything we should map OS onto this one (as their actual citation format appears to have changed).

adam3smith:
(the reason this is waiting is that I want to turn this into a general Informs style and create dependents, content-wise this is great)

rmzelle:
> (the reason this is waiting is that I want to turn this into a general Informs style and create dependents, content-wise this is great)

@adam3smith, I assume you'd still like to do this? If you want, I can compile journal metadata for https://github.com/citation-style-language/journals. (by hand, if necessary, since there only seem to be 14 journals; http://pubsonline.informs.org/)

Would that help speed things along?

adam3smith:
yes, it definitely would (and I should have a very good amount of time to work on various back-logs this weekend).

rmzelle:
@adam3smith, shall I just use the style from this PR to create an `INFORMS` independent? I can generate the dependents as well.

adam3smith:
yes, please. As per above I did review the style proper.

rmzelle:
@jimapps, a very belated thanks for this submission. We added this as a general INFORMS style, and will add additional CSL styles for all INFORMS journals that use this format.

PR #1841, opened 2015-12-18
Create wu-master-management.csl

georgd:


wu-master-management.csl
```diff
@@ -0,0 +1,652 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" name-delimiter="/" et-al-min="4" et-al-use-first="3" et-al-subsequent-min="4" et-al-subsequent-use-first="3" initialize="false" initialize-with="." demote-non-dropping-particle="never" xmlns="http://purl.org/net/xbiblio/csl">
+<!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>WU Master Management</title>
+    <id>http://www.zotero.org/styles/wu-master-management.csl</id>
```
rmzelle: Style IDs shouldn't have ".csl" extensions (e.g. `<id>http://www.zotero.org/styles/wu-master-management</id>` instead of `<id>http://www.zotero.org/styles/wu-master-management.csl</id>`)

rmzelle: The same goes for the "self" link, by the way.

rmzelle:
Is there a more direct link to the WU Master Management style guide you can provide?

georgd:
@rmzelle what do you mean with more direct link? There’s actually no real style guide.

rmzelle:
> There’s actually no real style guide.

How do you know what the style should look like, then? Is there a webpage with examples, maybe?

georgd:
Ah, ok. The style was developped in a reference implementation in EndNote.

georgd:
Currently, there’s actually a document but it’s accessible to the students only.

rmzelle:
And is this style specific to the Management master, or is it used across the Wirtschaftsuniversität Wien?

georgd:
It’s specific to the Management master but used also for at least one other program (for which I created a dependent style).

georgd:
So it’s used by more than 150 new students each year.

rmzelle:
I'm just asking to make sure we name the style correctly. Could you change the title to "Wirtschaftsuniversität Wien - Management", and the style ID and "self" link to "wirtschaftsuniversitat-wien-management" (with matching style file name)?

Also please add a "documentation" link (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#10---documentation-link ) to "http://wu.ac.at/library".

And will this style be primarily used in a single language? It's possible to "fix" styles to a certain locale (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#3---style-locale).

georgd:
I’ll spell out the university name, but I’d prefer to keep the “master” in the title (and the id and so on).

The style will be used both in English and German, however most students write in German. It looks like one can’t change the language in Zotero when a default language is given, so I’ll not set it.

rmzelle:
You can consolidate the `<locale xml:lang="de-DE">` and `<locale xml:lang="de-AT">` sections by just using the less specific `<locale xml:lang="de">` (which would be used for any German locale).

@adam3smith, any other requests from your side? I guess the et-al attributes might need some cleanup?

georgd:
Ok, consolidated them. This was also my first attempt but whyever, in my first tests I had troubles when I set the language to de-AT in the Zotero plugin in Word. Now it seems to work.

rmzelle:
Thanks.

The only thing remaining I'd like you to change is the et-al settings. The style currently has:

`<style ... et-al-min="4" et-al-use-first="3" et-al-subsequent-min="4" et-al-subsequent-use-first="3" ...>`

`<name ... et-al-min="4" et-al-use-first="3" et-al-subsequent-min="4" et-al-subsequent-use-first="3" .../>`

`<citation et-al-min="4" et-al-use-first="3" et-al-subsequent-min="4" et-al-subsequent-use-first="3" ...>`

`<bibliography ... et-al-min="8" et-al-use-first="6" et-al-use-last="true" ...>`

The way CSL works, the et-al settings on `<style/>` are overridden by those on `<citation/>` and `<bibliography/>`, which in turn are overridden by those on `<name/>`. For our repository, we prefer the settings to be only placed on `<citation/>` and `<bibliography/>`. Can you remove them from `<style/>` and `<name/>`? (I don't think this should change how the style behaves)

georgd:
Fine, I think, I got it now.

rmzelle:
Awesome, thanks!

georgd:
Great, thank you for your help!

PR #1845, opened 2015-12-28
Create gms-medizin-bibliothek-information.csl

zuphilip:


dependent/gms-medizin-bibliothek-information.csl
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" default-locale="en-US">
+  <info>
+    <title>GMS Medizin, Bibliothek, Information</title>
```
rmzelle: Shouldn't this be "GMS Medizin - Bibliothek - Information"? (I try to avoid any type of dashes other than regular hyphens)

zuphilip: Yes, it looks that they are currently writing the title with hyphens. I will correct it. Thanks

rmzelle:
Thanks!

PR #1847, opened 2015-12-29
Fixed journal abbrevs to not be followed by periods, fixed "Accessed on" to include space before

patrickmdnet:


rmzelle:
Thanks!

PR #1852, opened 2016-01-04

stvn-s:
Changes at lines 222 and 235 to rectify formatting issues.
YSJU style requires that page details for edited chapter and journal needs a space between p. and actual page number.


rmzelle:
Thanks.

(just note that in general, although it doesn't matter as much in this case, we prefer delimiters for separating punctuation, e.g. `<group delimiter=" ">`)

PR #1853, opened 2016-01-05

stvn-s:
Edits to lines 221 and 222; 234 and 235 - replacing 'suffix' with 'delimiter' for punctuation for these groups
(This was suggested as part of a previous pull request.)

This change is not urgent, but brings the previous changes into line with preferred option.


rmzelle:
Thanks!

PR #1855, opened 2016-01-09
Update AoM and apa.csl

bwiernik:
- Add support for status variable
- Prefer DOI over other URLs

### DIFF
--- a/academy-of-management-review.csl
+++ b/academy-of-management-review.csl
@@ -104,11 +104,7 @@
             </choose>
           </else-if>
           <else-if variable="issued" match="any">
-            <choose>
-              <if variable="status" match="all">
-                <text variable="status" text-case="capitalize-first"/>
-              </if>
-            </choose>
+            <text variable="status" text-case="capitalize-first"/>
           </else-if>
         </choose>
       </if>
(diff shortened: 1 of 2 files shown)


academy-of-management-review.csl
```diff
@@ -99,7 +102,14 @@
                 </date>
               </if>
             </choose>
-          </if>
+          </else-if>
+          <else-if variable="issued" match="any">
+            <choose>
```
adam3smith: unless I'm missing something, you don't need to test for status here. just use the current line 109.

adam3smith:
thanks -- see the above nit (which is cosmetic, i.e. shouldn't affect output, but I'd like to avoid clutter where possible).

academy-of-management-review.csl
bwiernik: I think it's necessary. If there is no `issued`, `status` gets printed in its place. `status` shouldn't get printed twice in that case.

bwiernik:
Oh, forgot to change branches when pushing a commit. Also fixed a minor bug in apa.csl

academy-of-management-review.csl
adam3smith: i understand why you need to test for issued. But why do you need to test for status? You should never need to test for a variable in a `choose` element with just one `if`. That's implicit -- does that make sense? Or what am I missing?

bwiernik: Ah, right. I wasn't thinking there.

adam3smith:
thanks!

PR #1856, opened 2016-01-10
Add Journal of Universal Computer Science (J.UCS)

GeorgLink:
Added csl for Journal of Universal Computer Science (J.UCS)

### DIFF
--- a/journal-of-universal-computer-science.csl
+++ b/journal-of-universal-computer-science.csl
@@ -3,7 +3,7 @@
 <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
     <title>Journal of Universal Computer Science</title>
-    <title-short>J.USC</title-short>
+    <title-short>J.UCS</title-short>
     <id>http://www.zotero.org/styles/journal-of-universal-computer-science</id>
     <link href="http://www.zotero.org/styles/journal-of-universal-computer-science" rel="self"/>
     <link rel="template" href="http://www.zotero.org/styles/apa-single-spaced"/>
--- a/journal-of-universal-computer-science.csl
+++ b/journal-of-universal-computer-science.csl
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style class="in-text" version="1.0" and="text" delimiter-precedes-last="never" demote-non-dropping-particle="never" default-locale="en-GB" xmlns="http://purl.org/net/xbiblio/csl">
+<style class="in-text" version="1.0" delimiter-precedes-last="never" demote-non-dropping-particle="never" default-locale="en-GB" xmlns="http://purl.org/net/xbiblio/csl">
 <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
     <title>Journal of Universal Computer Science</title>
--- a/journal-of-universal-computer-science.csl
+++ b/journal-of-universal-computer-science.csl
@@ -150,15 +150,15 @@
       <if type="report thesis book graphic motion_picture report song manuscript speech" match="any">
         <choose>
           <if variable="version" type="book" match="all">
-            <text variable="title"/>
+            <text variable="title" text-case="title"/>
           </if>
           <else>
-            <text variable="title" font-style="normal"/>
+            <text variable="title" text-case="title" font-style="normal"/>
           </else>
         </choose>
       </if>
       <else>
-        <text variable="title"/>
+        <text variable="title" text-case="title"/>
       </else>
     </choose>
   </macro>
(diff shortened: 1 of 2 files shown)


GeorgLink:
I fixed a typo in the short title.

journal-of-universal-computer-science.csl
```diff
@@ -0,0 +1,529 @@
...
+    <title-short>J.UCS</title-short>
+    <id>http://www.zotero.org/styles/journal-of-universal-computer-science</id>
+    <link href="http://www.zotero.org/styles/journal-of-universal-computer-science" rel="self"/>
+    <link rel="template" href="http://www.zotero.org/styles/apa-single-spaced"/>
+    <link href="http://www.jucs.org/ujs/jucs/info/submissions/style_guide.html" rel="documentation"/>
+    <author>
+      <name>Georg Link</name>
+      <email>glink@unomaha.edu</email>
+    </author>
+    <category citation-format="author-date"/>
```
adam3smith: could you add `<category field="science"/>` (or any other field from the [list](http://docs.citationstyles.org/en/latest/specification.html#appendix-i-categories) that you think is suitable; I know it's kind of restrictive, but still better to tag at all).

journal-of-universal-computer-science.csl
```diff
@@ -0,0 +1,529 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" and="text" delimiter-precedes-last="never" demote-non-dropping-particle="never" xmlns="http://purl.org/net/xbiblio/csl">
```
adam3smith: not quite sure I understand why you're setting `delimiter-precedes-last="never"` here but then set it to `"always"` [below](https://github.com/citation-style-language/styles/pull/1856/files#diff-553bedbd5da131219a6a3f1fbab7280eR55)
It seems odd this would differ within the style? And if it does, better to set it explicitly each time for clarity.

adam3smith: Presumably the journal is English only, right? In that case, please add `default-locale="en-US"` to this line.

journal-of-universal-computer-science.csl
```diff
@@ -0,0 +1,529 @@
...
+    </sort>
+    <layout delimiter=", " prefix="[" suffix="]">
+      <group delimiter=" ">
+        <text macro="author-short"/>
+        <text macro="issued-year"/>
+        <text macro="citation-locator"/>
+      </group>
+    </layout>
+  </citation>
+  <bibliography et-al-min="8" et-al-use-first="6" et-al-use-last="true" entry-spacing="0">
```
adam3smith: I'm not sure what the journal wants to do about et-al, but I'd definitely removfe the `et-al-use-last="true"` here, which is part of APAs somewhat odd rule to put ellipses between the 6th and the last author instead of using et-al.

adam3smith:
See comments inline. Looks good otherwise. Thanks!

GeorgLink:
adam3smith, I have incorporated your suggestions.

The style guide uses the "and" optionally for the in-line tag but never in the bibliography
http://www.jucs.org/ujs/jucs/info/submissions/style_guide.html
How can I remove the "and" in front of the last author?

I tried text-case="title" but it caused the words in the title to be merged without space in Docear4Word. Thus, it is reverted.

journal-of-universal-computer-science.csl
GeorgLink: Thank you. I had not found the list when I went looking for it and `computer_science` does not work.

journal-of-universal-computer-science.csl
```diff
@@ -0,0 +1,530 @@
...
+          <else>
+            <text macro="title"/>
+          </else>
+        </choose>
+      </substitute>
+    </names>
+  </macro>
+  <macro name="author-short">
+    <names variable="author">
+      <name form="short" and="text" initialize-with=". "/>
```
adam3smith: you want to remove the `and="text"` here to remove the and between authors. That was what you were asking about, right?

adam3smith:
Looks good to me otherwise. Let me know when you think it's ready to merge @GeorgLink .

@rmzelle -- could you take a quick look and see if you have a problem with a) the merge in the commit history and b) fixing the contributing.md type in a regular commit?

rmzelle:
> could you take a quick look and see if you have a problem with a) the merge in the commit history and b) fixing the contributing.md type in a regular commit?

It's fine.

GeorgLink:
@rmzelle I reverted the typo fix in contributing.md

@adam3smith thanks for the hint how to remove the `and`. It is good to go.

rmzelle:
Thanks!

Because of the large number of commits I just committed the latest version myself.

rmzelle:
Also thanks for finding the typo. Fixed here: https://github.com/citation-style-language/styles/commit/3b09114650a13d1a1d43c276d4e0ea007213eb85

GeorgLink:
You are welcome. Happy I could contribute. And it's cool how quickly this got accepted. :-D

PR #1857, opened 2016-01-11
Add Utah Geological Survey Style

inkenbrandt:
Style of the the Utah Geological Survey, Utah Geological Association, and the Journal of the Intermountain West.

Modified from the U.S. Geological Survey style template


rmzelle:
Great, thanks.

We'd also be happy to accept a *dependent* style for the journal (see https://github.com/citation-style-language/styles/blob/master/dependent/nature-biotechnology.csl for an example of what those look like). Would that be the "Geology of the Intermountain West" journal? (http://www.utahgeology.org/openjournal/index.php/GIW/index)

inkenbrandt:
Thanks! Yes, that is the right journal (ISSN 2380-7601).  The journal formatting is dependent on the style of the Survey.  I will submit a pull request for a dependent style for the Geology of the Intermountain West.

PR #1859, opened 2016-01-11
Fix case of text variable `Title`

inkenbrandt:
Titles should be sentence case.


adam3smith:
no, don't use that attribute. It's impossible to automate sentence case (think proper names, e.g.), so CSL styles should assume that titles are entered in sentence case and use `text-case="title"` where title case is required. Here's the relevant KB entry for Zotero: https://www.zotero.org/support/kb/sentence_casing, but this applies to all CSL implementations.

PR #1863, opened 2016-01-13
Update invisu.csl

amendesd:
First update of the new year! This update uses the new and very useful language selector of Zotero developed by Rintze Zelle and Aurimas Vinckevicius. Thanks to them!

### DIFF
--- a/invisu.csl
+++ b/invisu.csl
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" page-range-format="expanded" default-locale="fr-FR">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" page-range-format="expanded">
   <info>
-    <title>InVisu (French)</title>
+    <title>InVisu (International)</title>
     <id>http://www.zotero.org/styles/invisu</id>
     <link href="http://www.zotero.org/styles/invisu" rel="self"/>
     <link href="http://invisu.inha.fr/-ZOTERO-" rel="documentation"/>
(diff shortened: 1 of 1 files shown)


invisu.csl
```diff
@@ -574,11 +673,13 @@
...
+          <text macro="author-sub"/>
           <group delimiter=", ">
             <text macro="disambiguate"/>
-            <group delimiter="&#160;">
-              <text macro="locators"/>
+            <group delimiter=" ">
+              <text term="note" prefix="(" />
+              <text variable="first-reference-note-number" suffix=")"/>
+              <text macro="locators"/>       
             </group>
```
rmzelle: It would be much better to rewrite this as:

```xml
            <group delimiter=" ">
              <group prefix="(" suffix=")" delimiter=" ">
                <text term="note"/>
              </group>
              <text macro="locators"/>       
            </group>
```

invisu.csl
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" page-range-format="expanded" default-locale="fr-FR">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" page-range-format="expanded">
   <info>
-    <title>InVisu (French)</title>
+    <title>InVisu (International)</title>
```
rmzelle: Could you delete the " (International)" from the style title? We have quite a few more styles without a `default-locale`, and none are currently tagged as such, and I'd like to keep things consistent.

rmzelle:
Glad you like the Zotero feature!!!

I have two minor comments, but otherwise this looks good.

rmzelle:
Thanks!

PR #1864, opened 2016-01-13
Create geobiology.csl

zuphilip:
See https://forums.zotero.org/discussion/52148?page=1#Item_6

### DIFF
--- a/geobiology.csl
+++ b/geobiology.csl
@@ -1,0 +1,225 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" default-locale="en-US" version="1.0">
+  <info>
+    <title>Geobiology</title>
+    <id>http://www.zotero.org/styles/geobiology</id>
+    <link rel="self" href="http://www.zotero.org/styles/geobiology"/>
+    <link href="http://www.zotero.org/styles/ecological-entomology" rel="template"/>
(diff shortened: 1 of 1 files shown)


geobiology.csl
```diff
@@ -0,0 +1,225 @@
...
+        <text term="no date" form="short"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="edition">
+    <choose>
+      <if is-numeric="edition">
+        <group delimiter=" ">
+          <number variable="edition" form="ordinal"/>
+          <text value="edn"/>
```
adam3smith: could you fix this (I assume it's from the template) to use the short form of the term (and redefine that to edn.)

geobiology.csl
```diff
@@ -0,0 +1,225 @@
...
+          <text variable="genre"/>
+          <text variable="number" prefix=" No. "/>
+        </group>
+      </if>
+      <else-if type="bill book graphic legal_case legislation motion_picture report song speech" match="any">
+        <text variable="title" font-style="italic"/>
+      </else-if>
+      <else-if type="webpage">
+        <text variable="title"/>
+        <text value="WWW Document" prefix=" [" suffix="]"/>
```
adam3smith: do you know if they use that? It's not in the author guidelines and I find it really ugly, so would rather not have it anywhere it's not required.

adam3smith:
Looks good otherwise. Thanks!

adam3smith:
Thanks!

PR #1865, opened 2016-01-16
[Zf Kunstgeschichte] Add comma in bibliography

zuphilip:
see https://forums.zotero.org/discussion/55918?page=1#Item_31


adam3smith:
thanks!

PR #1866, opened 2016-01-17
Changes in documents-d-archeologie-francaise

NMonteix:


documents-d-archeologie-francaise.csl
```diff
@@ -105,7 +106,12 @@
   </macro>
   <macro name="responsability">
     <choose>
-      <if variable="author editor translator collection-editor" match="any">
+      <if variable="version">
+        <group suffix=" (éd.)">
```
adam3smith: I don't quite understand -- why are all the labels in the style set as suffix? That seems rather awkward, but I may be missing a reason this is done?

NMonteix: The difficulty here is that the style defined by the editor wants different labels that not necessarily correspond with the real role.
For instance the line quoted above has been added because computer programs authors need to be labelled as editors.

NMonteix: ... but I can indeed use labels ...
Huge corrections done but this one to be done...

NMonteix: Done

documents-d-archeologie-francaise.csl
```diff
@@ -424,214 +406,164 @@
...
   <bibliography>
     <sort>
-      <key macro="responsability"/>
+      <key macro="author-short"/>
       <key macro="year-date"/>
       <key macro="title"/>
     </sort>
     <layout>
-      <group prefix="" suffix="" delimiter=" ">
+      <group prefix="" suffix="&#9;" delimiter=" ">
```
rmzelle: What is the `&#9;` for? (seems to be a tab? http://www.fileformat.info/info/unicode/char/0009/index.htm )

documents-d-archeologie-francaise.csl
```diff
@@ -45,23 +48,20 @@
     <choose>
       <if variable="author">
         <names variable="author">
-          <name form="short" delimiter=", " initialize-with=". " delimiter-precedes-last="never" name-as-sort-order="all" sort-separator=" " et-al-min="3" et-al-use-first="1" font-weight="bold">
-      </name>
+          <name form="short" delimiter=", " initialize-with=". " delimiter-precedes-last="never" name-as-sort-order="all" sort-separator=" " et-al-min="3" et-al-use-first="1" font-weight="bold"/>
```
rmzelle: It's generally easier to put some of these options on `<citation>` and `<bibliography>` (such as `et-al-min="3"` and `et-al-use-first="1"`) and remove them here. For a full list, see http://docs.citationstyles.org/en/stable/specification.html#inheritable-name-options.

documents-d-archeologie-francaise.csl
NMonteix: Indeed. It is there to avoid the author and date to be extended all the line long in Word (behavior linked with the block layout?).

documents-d-archeologie-francaise.csl
NMonteix: Sure, will do

documents-d-archeologie-francaise.csl
rmzelle: Can you post a screenshot without the workaround and point out what's wrong exactly?

NMonteix: Here it is

rmzelle: That didn't work. Instead of using an email attachment, can you drag and drop (or paste) the screenshot into the GitHub webpage?

NMonteix: csl-daf no tab

rmzelle: And this is with Zotero and MS Word, right?

rmzelle: Could you try to use `display="block"` on the first `<group>` as well, and check whether that fixes things?

NMonteix: If the last encoding should be right in the end, do not merge it yet, some last corrections are almost ready

NMonteix: csl-daf no tab-block
Here is the result with `
<group display="block" delimiter=" ">
<text macro="author-short"/>
<text macro="year-date" font-weight="bold"/>
</group>
<group display="block" suffix=".&#9;">`

rmzelle: And this is with Zotero and MS Word, right?

NMonteix: Indeed, last Zotero version and MS Word 15

rmzelle: @adam3smith, @fbennett, does either one of you know whether `display="block"` does/should result in justified text? I don't see it happening in the Zotero CSL edit pane.

NMonteix: In fact, it is a "normal" behavior in Word: a tabulation is needed before a line feed if you don't want the whole line to be justified:
word
On the other hand, OpenOffice doesn't show this behavior:
tab openoffice
In the peculiar case of this style, `display="block` should not be applied to the first (author-date) group.

rmzelle: @adam3smith, have you ever seen anything like https://github.com/citation-style-language/styles/pull/1866#discussion_r50144450 (and maybe know of a better workaround?). Or shall we just accept the PR as-is?

adam3smith: I'd like to take a closer look at this first. Seems odd to me. So hold off on merging, please.

documents-d-archeologie-francaise.csl
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="UTF-8"?>
```
adam3smith: sorry, this is actually the fault of our formatter tool, but could you lowercase utf-8 again, please?

adam3smith:
otherwise looks good.

NMonteix:
Done.

adam3smith:
Merci Bien!

NMonteix:
À vous!

PR #1868, opened 2016-01-20
Create wegcenter-homepage.csl

mir06:
Hello, I've created that style we use on our webpage. Maybe it's useful for others as well.

### DIFF
--- a/wegcenter-homepage.csl
+++ b/wegcenter-homepage.csl
@@ -3,8 +3,8 @@
   <info>
     <title>Wegener Center Homepage</title>
     <title-short>WEGC</title-short>
-    <id>http://wegcwww.uni-graz.at/publ/wegcpubl/wegcenter-homepage</id>
-    <link href="http://wegcwww.uni-graz.at/publ/wegcpubl/wegcenter-homepage" rel="self"/>
+    <id>http://www.zotero.org/styles/wegener-center-homepage</id>
+    <link href="http://www.zotero.org/styles/wegener-center-homepage" rel="self"/>
     <link href="http://wegcwww.uni-graz.at/publ/wegcpubl/style_manual.pdf" rel="documentation"/>
     <author>
       <name>Armin Leuprecht</name>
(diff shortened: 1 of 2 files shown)


rmzelle:
I don't seem to be able to access http://wegcwww.uni-graz.at/publ/wegcpubl/style_manual.pdf. Is that an intranet link?

And is this style also used for any publications, or just your website?

mir06:
It's no intranet link - I just put an empty pdf-file there as I still have to document the style.

The style will be used for our Wegener Center Verlag (see http://wegcenter.uni-graz.at/en/wegener-center-verlag)

rmzelle:
While we're happy to see that you're using CSL, the CSL repository might not be the best place for your style. See https://github.com/citation-style-language/styles/wiki/Criteria-for-Accepting-Styles for our rationale. In short, repository styles are automatically pushed to many users, and we think it's better to reject styles for which there is likely to be little outside demand.

zuphilip:
Well, I guess the [Wegener Center](http://wegcenter.uni-graz.at/en/wegener-center/) is something similar to a "university departement":
<blockquote>The Wegener Center for Climate and Global Change is an interdisciplinary, internationally oriented research institute of the University of Graz</blockquote>
And the Wegener Center Verlag ["publishes scientific project reports, Ph.D. theses and M.Sc. theses"](http://wegcenter.uni-graz.at/en/wegener-center-verlag/). Would the style be applicable to all these publications from different authors?

adam3smith:
yes, that'd be my question, too: how many people would be using the style and who are they (in the sense that I'd be more favorably inclined if we're talking about a changing group of outside contributors versus permanent staff)?

rmzelle:
(yes, if the style is for publications, instead of for a single website, then I misunderstood, and we'd be much more open to accepting the style)

mir06:
Yes, the Wegener Center is an University department. And yes, the Wegener Center Verlag publishes theses as well as scientific reports. It won't be very many people using the style.
@rmzelle I obviously did not read the criteria-document very carefully. I agree that in our case it's better to host the style-file by ourselfs - sorry for the work.

adam3smith:
OK, thanks, closing this then. FWIW, it looks like a well-done style that should serve you well.

PR #1870, opened 2016-01-21
Create ecosistemas.csl

Pakillo:
Create new CSL style for the Spanish journal 'Ecosistemas' (www.revistaecosistemas.net)


rmzelle:
Thanks!

PR #1871, opened 2016-01-25
Create ucla-ubc-urban-theory-research-seminar-leitner-peck-sheppard.csl

teo5:
This represents the style required for the University of California, Los Angeles and University of British Columbia geography departments' joint urban theory research seminar. It may also be used more widely within the geography departments.

Style guide may be found on the last page of the PDF at http://blogs.ubc.ca/peck/files/2016/01/UBCUCLA-syllabus-2016-FINAL.pdf


rmzelle:
> It may also be used more widely within the geography departments.

Do any of these departments prescribe this exact citation style? Is the seminar an annually recurring one?

I'm trying to get a feel of whether accepting this style would actually be a good thing. The repository styles are made available to hundreds of thousands of users, and some products even ship with a copy of the entire style repository, so I try to keep the repository free of styles that have either very limited or short-lived appeal. If the seminar is a one-off event, it might be just as handy to just give participants a download link to the CSL style. Thoughts?

(see also https://github.com/citation-style-language/styles/wiki/Criteria-for-Accepting-Styles)

rmzelle:

PR #1872, opened 2016-01-25
IUCr: fix et-al font-style and list all authors in bibliography

touwwouter:
* IUCr: et-al font-style should be italic (55fa3a6)
* IUCr: all authors must be given (16447c0)


rmzelle:
Thanks. Could you update the documentation link as well (to "http://journals.iucr.org/d/services/notesforauthors.html")?

rmzelle:
Bedankt!

PR #1873, opened 2016-01-29
Update for zeitschrift-fur-kunstgeschichte.csl

blitzgneisserin:
translator(s) were missing (fixed)
names were saparated by commas only (instead of commas + and; fixed)

### DIFF
--- a/zeitschrift-fur-kunstgeschichte.csl
+++ b/zeitschrift-fur-kunstgeschichte.csl
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="UTF-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only">
   <info>
     <title>Zeitschrift für Kunstgeschichte</title>
@@ -17,7 +17,7 @@
     <category field="history"/>
     <issn>0044-2992</issn>
     <summary>From the editors: "Die Herausgeber werden im Falle von Editionen, Lexika und Ausstellungskatalogen dem Titel nachgestellt [use encyclopedia articles with or without container-title for that]. Bei gewöhnlichen Sammelbänden werden die Herausgeber dem Titel vorangestellt [use book for that]." Multilingual style; the information for exhibition catalogues should be entered in the field collection-title; locators may use the word "here" or "hier" in front of the page refering to which must be entered individually (the style outputs the locator as it is entered w/o any label or additional text).</summary>
-    <updated>2016-01-16T10:57:41+00:00</updated>
+    <updated>2016-01-29T04:44:25+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="de">
@@ -25,11 +25,15 @@
       <term name="note">wie Anm.</term>
       <term name="et-al">et al.</term>
       <term name="editor" form="short">Hg.</term>
+      <term name="translator" form="verb-short">übers. von</term>
+      <term name="editortranslator" form="verb-short">hg. und übers. von</term>
     </terms>
   </locale>
   <locale xml:lang="en">
     <terms>
       <term name="note">as note</term>
+      <term name="translator" form="verb-short">transl. by</term>
+      <term name="editortranslator" form="verb-short">ed. and transl. by</term>
     </terms>
   </locale>
   <macro name="author-short">
@@ -47,8 +51,8 @@
     </names>
   </macro>
   <macro name="editor">
-    <names variable="editor translator" delimiter=", ">
-      <name delimiter="" prefix=" " and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never"/>
+    <names variable="editor translator" delimiter=". ">
+      <name and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never"/>
       <label form="short" prefix=" (" suffix=")"/>
     </names>
   </macro>
@@ -81,6 +85,12 @@
       <text variable="collection-title"/>
       <text variable="collection-number"/>
     </group>
+  </macro>
+  <macro name="translator">
+    <names variable="editor translator" delimiter=". ">
+      <label form="verb-short"/>
+      <name prefix=" " and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never"/>
+    </names>
   </macro>
   <citation et-al-min="3" et-al-use-first="1" disambiguate-add-names="true">
     <layout delimiter="; " suffix=".">
@@ -126,8 +136,8 @@
                           <text macro="volume"/>
                         </group>
                         <names variable="editor">
+                          <name and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never"/>
                           <label form="verb-short" suffix=" "/>
-                          <name/>
                         </names>
                       </group>
                     </if>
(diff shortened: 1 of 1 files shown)


zeitschrift-fur-kunstgeschichte.csl
```diff
@@ -25,11 +25,15 @@
       <term name="note">wie Anm.</term>
       <term name="et-al">et al.</term>
       <term name="editor" form="short">Hg.</term>
+      <term name="translator" form="verb-short">übers. von</term>
+      <term name="editortranslator" form="verb-short">hg. und übers. von</term>
```
zuphilip: These terms should already be translated in `de-DE` standard locale exactly as you want them: https://github.com/citation-style-language/locales/blob/master/locales-de-DE.xml#L279

Thus, there should be no reason to define them here again, i.e. delete those two lines here.

zeitschrift-fur-kunstgeschichte.csl
```diff
@@ -25,11 +25,15 @@
...
     </terms>
   </locale>
   <locale xml:lang="en">
     <terms>
       <term name="note">as note</term>
+      <term name="translator" form="verb-short">transl. by</term>
+      <term name="editortranslator" form="verb-short">ed. and transl. by</term>
```
zuphilip: The [standard locale](https://github.com/citation-style-language/locales/blob/master/locales-en-US.xml#L311) would give:
```
<term name="translator" form="verb-short">trans. by</term>
<term name="editortranslator" form="verb-short">ed. &amp; trans. by</term>
```
Are you sure that we should choose different abbreviations here for "translated"?

zeitschrift-fur-kunstgeschichte.csl
```diff
@@ -47,8 +51,8 @@
     </names>
   </macro>
   <macro name="editor">
-    <names variable="editor translator" delimiter=", ">
-      <name delimiter="" prefix=" " and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never"/>
+    <names variable="editor translator" delimiter=". ">
```
zuphilip: The `delimiter=". "` looks wrong for me, but I don't have a counter example. It would result in something like "Editor1. Editor2. Editor3 and Editor4". However, commas would look better.

blitzgneisserin:
I think we should leave "hg. und übers." resp. "ed. and trans." webcause if there is nothing Zotero makes "ed. & trans.". "and/und" looks better, doesn't it? Btw. I copied those lines from "Kritische Ausgabe".

zeitschrift-fur-kunstgeschichte.csl
blitzgneisserin: The way I see it the delimiter does not matter here because Zotero either writes 2 names separated by "and" or , if there are more than 2 names "name et al."

zeitschrift-fur-kunstgeschichte.csl
zuphilip: Okay, I read your comment below. But the first line seems then not necessary. Moreover, we don't use "editortranslator" at the moment anywhere in the style. Note this would be one person who edited and translated the text.

blitzgneisserin: "editortranslaltor" was exactly what I needed:

"Laurentius Valla, Über den freien Willen - De libero arbitrio, hrsg. und übers. von Eckhard Keßler (lat.-dt. Ausg.), München 1987."

or:

"Leon Battista Alberti, Über die Malkunst, hg. und übers. von Oskar Bätschmann und Sandra Gianfreda, Darmstadt 2002."

zeitschrift-fur-kunstgeschichte.csl
blitzgneisserin: No, that is my mistake, "trans." ist fine

zeitschrift-fur-kunstgeschichte.csl
```diff
@@ -149,14 +159,29 @@
...
+                          <label form="short" prefix=" (" suffix=")"/>
+                        </names>
+                      </if>
+                    </choose>
+                    <group delimiter=" ">
+                      <text variable="title" font-style="italic" suffix=","/>
+                      <text macro="translator"/>
+                      <text macro="collection-with-number"/>
+                    </group>
+                  </else-if>
```
zuphilip: IMO you don't need here another `else-if`-case for the translator(s). Try to use another group, i.e. something like:
```
                   <else>
                     <choose>
                       <if match="none" variable="author">
                         <names variable="editor">
                           <name/>
                           <name and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never"/>
                         </names>
                       </if>
                     </choose>
                     <group delimiter=" ">
                       <group delimiter=", ">
                         <text macro="translator"/>
                       </group>
                       <text macro="collection-with-number"/>
                     </group>
                   </else>
```

zeitschrift-fur-kunstgeschichte.csl
```diff
@@ -228,14 +253,29 @@
...
+                      <label form="short" prefix=" (" suffix=")"/>
+                    </names>
+                  </if>
+                </choose>
+                <group delimiter=" ">
+                  <text variable="title" font-style="italic" suffix=","/>
+                  <text macro="translator"/>
+                  <text macro="collection-with-number"/>
+                </group>
+              </else-if>
```
zuphilip: Dito

zeitschrift-fur-kunstgeschichte.csl
zuphilip: You are right this works already here because of an exception I didn't know:

> If multiple variables are selected (separated by single spaces, see example below), each variable is independently rendered in the order specified, with one exception: when the selection consists of “editor” and “translator”, and when the contents of these two name variables is identical, then the contents of only one name variable is rendered. In addition, the “editortranslator” term is used if the cs:names element contains a cs:label element

http://docs.citationstyles.org/en/stable/specification.html#names

zeitschrift-fur-kunstgeschichte.csl
zuphilip: Okay, then you can delete the first line completely and adjust the second line accordingly.

zeitschrift-fur-kunstgeschichte.csl
zuphilip: Yeah, you are right, but I would still prefer to have here a comma. When at some point the et-al rule will change, then the style would already be "prepared" for that cases as well.

zuphilip:
Thank you @blitzgneisserin . The corresponding forum thread is https://forums.zotero.org/discussion/55918?page=1#Item_34 . This corrections look good in general, just some few improvements I mentioned above. Please fix these small issues and [update the style here](https://github.com/citation-style-language/styles/pull/1873#issuecomment-176958486).

zeitschrift-fur-kunstgeschichte.csl
blitzgneisserin: I know that something is unnecessary here but I do not understand the details, you know that. I am not that smart. I am not sure if I can do what you ask.
Anyway my solution works correctly, doesn't it? And zotero doesn't care.
I was just glad that I could fix this somehow. It was not easy for me.

zeitschrift-fur-kunstgeschichte.csl
blitzgneisserin: This is definitely a mistake that I copied form "Kritische Ausgabe"

zeitschrift-fur-kunstgeschichte.csl
blitzgneisserin: I am not sure if I understand your suggestion correctly but it would not work properly because of the structure of the macro "translator" - with edited books it would mention the editors twice. The macro "translator" is a renamed copy of "Kritische Ausgabe"'s macro "editor".

zuphilip: Don't debase yourself! Your updated style looks very good and I am only asking for some small nits here. I can change this issue here afterwards, if you prefer, but can you look over the terms and correct them?

zuphilip: (I had an error in my code above which I corrected now.)

blitzgneisserin:
Anyway thanks @zuphilip for the help and the corrections

zuphilip:
Superseeded by #1874

PR #1875, opened 2016-02-01
Updates to Neurology journal format

bbbruce:
These changes remove issue numbers, remove month and day from the journal article type, correct online article support, and add e-pub support according to the journal style.

### DIFF
--- a/neurology.csl
+++ b/neurology.csl
@@ -11,23 +11,9 @@
     <category field="medicine"/>
     <issn>0028-3878</issn>
     <eissn>1526-632X</eissn>
-    <updated>2015-01-31T19:00:00+00:00</updated>
+    <updated>2016-02-01T08:00:00+00:00</updated>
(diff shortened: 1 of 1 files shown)


neurology.csl
```diff
@@ -11,13 +11,15 @@
...
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
     <terms>
-      <term name="retrieved">available</term>
+      <term name="retrieved">accessed</term>
+      <term name="from">at</term>
       <term name="section" form="short">sect.</term>
+      <term name="internet">online</term>
     </terms>
```
adam3smith: at, accessed, and online all have terms of their own. Use those rather than redifining other ones, please. This is too confusing.

adam3smith:
Otherwise this looks good, thanks! See instructions above on how to update your pull request.

bbbruce:
Thanks for your suggestions.  It helped get rid of the cruft.

adam3smith:
perfect, thanks!

PR #1877, opened 2016-02-02
Create European-journal-of-international-law.csl

cathbrend:
I think this would be a interesting new style for international lawyers. I am a reference librarian based in Geneva, Switzerland, and some professors of my institution are part of this journal committee and recommend this style.


cathbrend:
I am sorry but I do not know how to fix this...

zuphilip:
You are missing the file ending `.csl`, i.e. rename the file to `european-journal-of-international-law.csl`

cathbrend:
Thanks, but I am still lost, sorry! I think I must replace "create  European-journal-of-international-law style" by "create European-journal-of-international-law.csl", but it is not in the code, it is the title of the pull request, how can I do it without creating a new pull request? I have tried, but apparently with no success...

adam3smith:
Does this help? https://help.github.com/articles/renaming-a-file/
the name of the pull request won't change and that's OK. It's the actual filename, at the top of the box with the file, that you need to add .csl to.

adam3smith:
almost there: you have two macros: `publisher` and `title-short` which aren't actually used. Delete them and the test will pass.

adam3smith:
Thanks!

PR #1879, opened 2016-02-09
IUCr: update author, editor, chapter, indentation

touwwouter:
* never precede last author with delimiter (5796760)
* fix book chapter style (51d874c)
* indent bibliographic entries(828eb46)
* citations should not be sorted (91e58cb)
* add names to disambiguate citations (cdb23ae)
* never precede et al. with delimiter (522a12e)

### DIFF
--- a/international-union-of-crystallography.csl
+++ b/international-union-of-crystallography.csl
@@ -51,10 +51,6 @@
   </macro>
   <!-- CITATION -->
   <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" collapse="year-suffix">
-    <sort>
-      <key macro="issued"/>
-      <key macro="author"/>
-    </sort>
     <layout prefix="(" suffix=")" delimiter="; ">
       <group delimiter=", ">
         <text macro="author-short"/>
--- a/international-union-of-crystallography.csl
+++ b/international-union-of-crystallography.csl
@@ -45,8 +45,8 @@
   </macro>
   <macro name="editor">
     <names variable="editor">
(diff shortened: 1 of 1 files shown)


touwwouter:
I would also like to change a few other things for which I could not find any hints in the specification, current issues or elsewhere. I can create a couple of new issues, but first I would like to make sure I haven't missed obvious solutions:

* Use `suffix=","` rather than `suffix="."` if `container-title-short`==`container-title` (or solve the problem otherwise.
* _Acta Cryst. D._ **66** should become _Acta Cryst._ D**66** (Acta letter not italic, no `.<space>` after letter); the same holds for other Acta Crystallographica journals. Other journal titles don't have special abbreviation-volume combinations.
* I use a custom journal abbreviation list to convert e.g. `Acta Crystallographica Section D: Biological Crystallography` to the non-standard abbreviation Acta Cryst. D. Could non-standard abbreviations be made part of the style specification?

adam3smith:
Only one of the above that's possible is the year suffix in italics. You can set `<text variable="year-suffix" font-style="italic"/>` explicitly.
The rest isn't. Not sure to what degree they're worthwhile, given that they're pretty specialized (and some of them rather unreasonable, as with the special format for Acta journals). Customized abbreviation lists for styles might happen at some point, but not anytime soon. The logistics of doing that right are quite complicated.

touwwouter:
Thanks! Indeed the the special format for Acta is quite unreasonable, but I hope the others will be implemented some time. I am not planning to make more changes soon, so please consider this pull-request final from my side.

PR #1880, opened 2016-02-10
remove space at "u.a." (et-al term)

DominikVogel:
Journal's style guide requests to use "u.a." instead of "et al.". the previous version used "u. a." (with a space between u. and a.). Fixed that.

### DIFF
--- a/der-moderne-staat.csl
+++ b/der-moderne-staat.csl
@@ -19,9 +19,9 @@
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="de">
-  	<terms>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks. Could you fix the indenting from tabs to two spaces? You can either do it manually or send the entire file through http://formatter.citationstyles.org/ (if you do the latter, please lowercase the utf-8 at the top of the style. The formatter currently uppercases that).

adam3smith:
perfect, thanks!

rmzelle:
> please lowercase the utf-8 at the top of the style. The formatter currently uppercases that

@adam3smith, this should now be fixed (see https://github.com/citation-style-language/formatter/issues/2, and hat-tip to @zuphilip for proving me wrong)

adam3smith:
yes, I followed that discussion. thanks!

PR #1882, opened 2016-02-11
Update conservation-biology.csl

zuphilip:
See https://forums.zotero.org/discussion/56734/style-error-conservation-biology/#Comment_246073


zuphilip:
This should fix all the points about the authors identified in the forum. Maybe, you can check in the forum that I am right about last point.

adam3smith:
yes, you're right on the URL. Thanks!

PR #1883, opened 2016-02-12
Update biological-journal-of-the-linnean-society.csl

ms609:
Updated to reflect edits made by journal copyeditor to recently typeset MS.

- list of authors in bibliography should not be truncated and replaced with et al; "Journal style is to include all author names for each reference in the reference list"
- Comma should not precede ampersand in in-text citations to references: correct style is "(Smith, Jones & Weston 2014)".  I believe that setting and="symbol" does not express the delimiter before the ampersand – do I need to do anything else?

### DIFF
--- a/biological-journal-of-the-linnean-society.csl
+++ b/biological-journal-of-the-linnean-society.csl
@@ -26,7 +26,7 @@
   </macro>
   <macro name="author">
     <names variable="author">
-      <name name-as-sort-order="all" sort-separator=" " initialize-with="" delimiter=", " and="symbol" delimiter-precedes-last="always" font-weight="bold"/>
+      <name name-as-sort-order="all" sort-separator=" " initialize-with="" delimiter=", " and="symbol" delimiter-precedes-last="never" font-weight="bold"/>
       <et-al font-style="italic"/>
       <label form="short" prefix=" (" suffix=")" text-case="capitalize-first"/>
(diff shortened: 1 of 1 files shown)


biological-journal-of-the-linnean-society.csl
```diff
@@ -26,7 +26,7 @@
   </macro>
   <macro name="author">
     <names variable="author">
-      <name name-as-sort-order="all" sort-separator=" " initialize-with="" delimiter=", " delimiter-precedes-last="always" font-weight="bold"/>
+      <name name-as-sort-order="all" sort-separator=" " initialize-with="" delimiter=", " and="symbol" delimiter-precedes-last="always" font-weight="bold"/>
```
adam3smith: you want to change `delimiter-precedes-last="always"` to `delimiter-precedes-last="never"` to prevent commas before the ampersand

adam3smith:
see note inline above. Looks great otherwise. 
Instructions on updating the pull request are in the automated message above. Thanks!

adam3smith:
@ms609 -- see my comment above. This should be a very quick update, would you mind quickly doing that so we can accept your fixes?

ms609:
I've made the change to delimiter-precedes-last="never".
Cheers

adam3smith:
thanks!

PR #1884, opened 2016-02-13
Update the-journal-of-indian-prosthodontic-society.csl

zuphilip:
From [Springer page](https://www.springer.com/medicine/dentistry/journal/13191):
<blockquote>This journal has been discontinued with effect from 13th January 2015. The Journal has moved to Medknow Publications http://www.journalonweb.com/jips/.</blockquote>


adam3smith:
thanks!

PR #1885, opened 2016-02-13
Springer2016

zuphilip:
The corresponding commits for the creation of these styles can be found at https://github.com/citation-style-language/utilities/pull/32

I split the changes in several commits to make it easier to review them.

I haven't thought about the `fachzeitschriften` further, here you just see the changes by strictly following the list and therefore changing some fachzeitschriften-styles. This might not be ideal. My guess is that there is a more general Springer style missing there with alphabetical ordering. But I fear that I cannot look deeper at the moment into this. Any practical solution would be appreciated and maybe we can postpone to look closer.

Let us now see, what Travis/Sheldon will complain...


zuphilip:
Trying to trigger Travis/Sheldo again by closing and reoping this PR.

zuphilip:
Okay, now I am happy with this PR and looking forward for your comments @adam3smith @rmzelle .

For the moment I would suggest to drop the `fachzeitschriften` commit. We can look afterwards closer at this style and the subtle differences. On the other hand the `bmc` commit should be pretty save, because also we know that all of these journals switched to "springer-vancouver-brackets". We just loose the `short-title` and the field. I feel that it is not worth to spend more time for these information. What do you think?

adam3smith:
Looks good to me in general. Agree BMC is fine as is. I'll trust you on Fachzeitschriften, so why don't you take that out. Anything else I should be looking at in particular?

zuphilip:
I deleted the Fachzeitschriften commit here. It might be worth to double check especially the additions and deletions (maybe also the "Rest" commit). Note that the updated versions are in the follow-up commits done.

rmzelle:
Should we get rid of "biomed-central.csl", or is it still used by any styles after this?

I looked over all the commits here and at those of https://github.com/citation-style-language/utilities/pull/32, and this all looks great. Thanks so much for working on this, @zuphilip! Since all the BMC CSL styles use the wrong format right now, I'd just go ahead and merge this. @adam3smith?

adam3smith:
yes, good to merge. If we don't have any other styles depending on BMC we should delete that or, perhaps better, turn it into a dependent style?

rmzelle:
> If we don't have any other styles depending on BMC we should delete that or, perhaps better, turn it into a dependent style?

I would just delete it and redirect users to a similar style in renamed-styles.json. No need to keep extinct style formats around.

rmzelle:
:sparkles: :+1: :fireworks:

adam3smith:
> :sparkles: :+1: :fireworks:

indeed. Thanks @zuphilip

zuphilip:
Thank you!

PR #1888, opened 2016-02-15
Simultaneous pages and locators

carljohanberglund:
In this update to svensk-exegetisk-arsbok.csl, I have added the full page ranges to the first reference of articles, even if a more specific page range is to be given.

Peter A. Brunt, “On Historical Fragments and Epitomes,” CQ 30/2 (1980): 477–94, hic 478–81.

Is there a way to replace the Latin "hic" with "here" if the document is in English, and with "här" if it is in Swedish?

### DIFF
--- a/svensk-exegetisk-arsbok.csl
+++ b/svensk-exegetisk-arsbok.csl
@@ -15,7 +15,7 @@
     <category field="theology"/>
     <issn>1100-2298</issn>
     <summary>Svensk exegetisk årsbok (full note)</summary>
-    <updated>2015-08-10T16:00:00+00:00</updated>
+    <updated>2016-02-15T16:00:00+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="sv-SE">
@@ -556,8 +556,11 @@
       <if variable="locator" match="none">
         <text macro="pages"/>
       </if>
-      <else-if type="article-journal">
-        <text variable="locator" prefix=": "/>
+      <else-if type="article-journal chapter" match="any">
+        <group delimiter=", hic ">
+          <text macro="pages"/>
+          <text variable="locator"/>
+        </group>
       </else-if>
       <else>
         <text macro="point-locators-subsequent" prefix=", "/>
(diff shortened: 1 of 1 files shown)


adam3smith:
thanks. 
Answer to your question is basically no, but so I undesrtand correctly: do you want this to depend on the language of what you're citing of what you're writing?

carljohanberglund:
OK. The language of what I'm writing.

rmzelle:
@adam3smith, wouldn't the following work? There isn't a "here" term, but you could use "at".

Instead of 

```xml
        <group delimiter=", hic ">
          <text macro="pages"/>
          <text variable="locator"/>
        </group>
```

use

```xml
  <locale xml:lang="sv-SE">
    <terms>
      <term name="at">här</term>
    </terms>
  </locale>
  <locale xml:lang="en">
    <terms>
      <term name="at">here</term>
    </terms>
  </locale>

  <!-- ... -->

        <text macro="pages"/>
        <group prefix=", " delimiter=" ">
          <text term="at"/>
          <text variable="locator"/>
        </group>
```

svensk-exegetisk-arsbok.csl
```diff
@@ -761,7 +764,7 @@
       </choose>
     </layout>
   </citation>
-  <bibliography hanging-indent="true" et-al-min="10" et-al-use-first="9" subsequent-author-substitute="&#8212;&#8212;&#8212;" entry-spacing="0">
+  <bibliography hanging-indent="true" et-al-min="10" et-al-use-first="9" subsequent-author-substitute="———" entry-spacing="0">
```
rmzelle: @carljohanberglund, just curious, do you know why `subsequent-author-substitute="&#8212;&#8212;&#8212;"` changed to `subsequent-author-substitute="———"` here?

carljohanberglund: Probably a cut-and-paste error. I'll revert it.

carljohanberglund:
@rmzelle Your suggestion works perfectly. Is there a way to specify "hic" as the default "at" term, to avoid  getting translations of "at" in languages other than the ones I specify explicitly?

carljohanberglund:
@rmzelle Got it to work with a default `<locale>:`

    <locale>
  	  <terms>
    		<term name="at">hic</term>
    	</terms>
    </locale>

svensk-exegetisk-arsbok.csl
rmzelle: Both are correct, but we escape certain characters for repository styles that are hard to identify/distinguish by eye, like the various types of dashes and special (e.g. non-breaking) spaces.

rmzelle:
Right, that's exactly how you'd do it.

PR #1889, opened 2016-02-16
Create national-institute-of-health-research-journals-library

rhiannon-ppl:
Style for the five journals in the National Institute of Health Research (NIHR) Journals Library. These open-access journals publish research funded by NIHR.

### DIFF
--- a/national-institute-of-health-research.csl
+++ b/national-institute-of-health-research.csl
@@ -15,14 +15,6 @@
     <category field="medicine"/>
     <issn>1366-5278</issn>
     <eissn>2046-4924</eissn>
-    <issn>2050-4365</issn>
-    <eissn>2050-4373</eissn>
-    <issn>2050-4349</issn>
-    <eissn>2050-4357</eissn>
-    <issn>2050-4322</issn>
-    <eissn>2050-4330</eissn>
-    <issn>2050-4381</issn>
-    <eissn>2050-439X</eissn>
     <summary>A style for journals in the NIHR Journals Library, resembles Vancouver style</summary>
     <updated>2016-02-05T11:44:17+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
--- a/national-institute-of-health-research.csl
+++ b/national-institute-of-health-research.csl
@@ -5,7 +5,7 @@
     <title-short>NIHR</title-short>
     <id>http://www.zotero.org/styles/national-institute-of-health-research</id>
     <link href="http://www.zotero.org/styles/national-institute-of-health-research" rel="self"/>
-    <link href="http://www.zotero.org/styles/elsevier-vancouver" rel="template"/>
+    <link href="http://www.zotero.org/styles/international-journal-of-infectious-diseases" rel="template"/>
     <link href="http://www.journalslibrary.nihr.ac.uk/information-for-authors/references" rel="documentation"/>
     <author>
       <name>Rhiannon Miller</name>
--- a/national-institute-of-health-research.csl
+++ b/national-institute-of-health-research.csl
@@ -13,8 +13,16 @@
     </author>
     <category citation-format="numeric"/>
     <category field="medicine"/>
-    <issn>1201-9712</issn>
-    <eissn>1878-3511</eissn>
+    <issn>1366-5278</issn>
+    <eissn>2046-4924</eissn>
+    <issn>2050-4365</issn>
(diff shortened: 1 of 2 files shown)


rmzelle:
Could you rename the style like so?:

```xml
        <title>National Institute of Health Research</title>
        <title-short>NIHR</title-short>
        <id>http://www.zotero.org/styles/national-institute-of-health-research</id>
        <link href="http://www.zotero.org/styles/national-institute-of-health-research" rel="self"/>
```

Also please change the file name to "national-institute-of-health-research.csl" (note the ".csl" extension, and see https://help.github.com/articles/renaming-a-file/ for instructions).

rmzelle:
And while you're at it, could you also reformat the style with http://formatter.citationstyles.org/ so that the indentation is 2 spaces per level?

rhiannon-ppl:
All these changes have now been made. Please let me know if you need me to do anything more.

Thanks and best wishes

Rhiannon


Prepress Projects
Rhiannon Miller • Programmes Manager
+44 (0)1738 450489

rmzelle:
Thanks. Could you delete:

```xml
    <issn>1201-9712</issn>
    <eissn>1878-3511</eissn>
```

Also, does that mean that you based the style on https://github.com/citation-style-language/styles/blob/master/international-journal-of-infectious-diseases.csl? If so, it would be better to change the "template" link from `<link href="http://www.zotero.org/styles/elsevier-vancouver" rel="template"/>` to `<link href="http://www.zotero.org/styles/international-journal-of-infectious-diseases" rel="template"/>`.

rhiannon-ppl:
Dear Rintze

You’re right. Fixed and done.


Prepress Projects
Rhiannon Miller • Programmes Manager
+44 (0)1738 450489

rmzelle:
Could you delete the ISSN and eISSN? Only journals have those, and this is a style for an organization.

rmzelle:
Also, once we accept this style, we can create "dependent" styles for the 5 individual journals that point to this one. See https://github.com/citation-style-language/styles/blob/master/dependent/nature-biotechnology.csl for an example of what such styles look like. That way each journal gets its own entry in the style repository.

rhiannon-ppl:
Thanks Rintze. Would you do that yourselves or should I?

Thanks and best wishes

Rhiannon

Prepress Projects
Rhiannon Miller • Programmes Manager
+44 (0)1738 450489

national-institute-of-health-research.csl
```diff
@@ -0,0 +1,204 @@
...
+    <choose>
+      <if variable="DOI">
+        <text variable="DOI" prefix="http://dx.doi.org/"/>
+      </if>
+      <else-if type="webpage post-weblog" match="any">
+        <choose>
+          <if variable="URL">
+            <group delimiter=" ">
+              <text variable="URL" prefix="URL: "/>
+              <date variable="accessed" prefix="(Accessed " form="text"/>
```
rmzelle: Where is the closing ")"?

(also, in general, it would be better to replace

```xml
<date variable="accessed" prefix="(Accessed " form="text"/>
```

with

```xml
<group prefix="(" suffix=")" delimiter=" ">
  <date variable="accessed" form="text"/>
</group>
```

That way you use a translatable term, which makes it easier to reuse this style in other languages.
)

rmzelle:
We can create the dependent styles.

rmzelle:
Also, it might make more sense to set the locale of this style to British English (`default-locale="en-GB"`) instead of US English (`default-locale="en-US"`).

national-institute-of-health-research.csl
rhiannon-ppl: Dear Rintze

Thanks for getting back to me.

The closing ) was on line 185:

        <else-if type="webpage">
          <group delimiter=" ">
            <text variable="title" form="long" font-style="italic" suffix="."/>
            <text variable="container-title" form="long" suffix="." text-case="title"/>
            <text macro="publisher"/>
            <text macro="access" suffix=")"/>
          </group>
        </else-if>	

However, I’ve taken your advice and replaced the code as you suggested. I’ve also made a few other minor changes including changing the default-locale.

Please let me know if there is anything more I need to do at this stage.

Thanks and best wishes

Rhiannon

Prepress Projects
Rhiannon Miller • Programmes Manager
+44 (0)1738 450489

rmzelle:
Thanks!

rmzelle:
I just added the dependent styles as well: https://github.com/citation-style-language/styles/pull/1893/files

rhiannon-ppl:
Thank you!

Best wishes

Rhiannon

Prepress Projects
Rhiannon Miller • Programmes Manager
+44 (0)1738 450489

PR #1891, opened 2016-02-18
Locator follows locale

carljohanberglund:
The "hic" separating the full page range from the locator will now be "here" if the document is written in English, "här" if it is in Swedish, and "hic" otherwise.


rmzelle:
Thanks!

(for the future, please use two spaces per indentation level in the XML. Some text editors allow you to automatically use that, or you can copy/paste your style into http://formatter.citationstyles.org/)

PR #1893, opened 2016-02-18
Add NIHR dependents

rmzelle:


rmzelle:
See also https://github.com/citation-style-language/utilities/commit/b23c749fe449ceb5d00b42af5ad9e65dee206ec1

PR #1894, opened 2016-02-18
New style for TU Dresden

jmiba:
Create
technische-universitat-dresden-betriebswirtschaftslehre-logistik-author-
date.csl


rmzelle:
Will this style be used in German only?

jmiba:
How it will be used I hardly can tell, yet I deliberately created it to be localizable to any language. However, I was not quite easy with some of the german terms, so I changed them. I expect some use of the style in English, at least. So it is not over the top to provide localization.

rmzelle:
I'm asking because if the style is primarily used in German, it might make sense to fix it to that language (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#3---style-locale ).

jmiba:
Yes, I know. But it is really meant to be used in different languages.

rmzelle:
Okay, in that case the style is good to go.

jmiba:
Thanks, very kind! And so fast!

rmzelle:
We aim to please. :)

Issue #1801, opened 2015-12-06
Travis Report unclear when filename is not matching


zuphilip:
On <a href="https://github.com/zuphilip/styles/commit/40cd80d2b57cc2bd7ffef704216ae6523a8b46d2">one of my commit</a> today the <a href="https://travis-ci.org/citation-style-language/styles/builds/95214564">travis report</a> was quite unclear:

```
  1) dependent style journal-of-hip-preserving-surgery the self-link (if present) is a valid style repository link
     Failure/Error: expect(style.self_link).to match(%r{http[s]?://www.zotero.org/styles/#{id}})
       expected "http://www.zotero.org/styles/journal-of-hip-preservation-surgery" to match /http[s]?:\/\/www.zotero.org\/styles\/journal-of-hip-preserving-surgery/


  2) dependent style journal-of-hip-preserving-surgery the id is a valid style repository link
     Failure/Error: expect(style.id).to eq("http://www.zotero.org/styles/#{id}")
       expected: "http://www.zotero.org/styles/journal-of-hip-preserving-surgery" got: "http://www.zotero.org/styles/journal-of-hip-preservation-surgery"
```

And actually the fix was to rename the file accordingly. One part of the confusion might be the use of id here. Isn't the variable `#{id}` corresponding to the file name? I suggest for example for 1) something like "the self-link (if present) matches the file name". What do you think?

(I opened a new issue, because I don't know the status of #1088.)

rmzelle:
Since we already test whether the style ID matches the "self" link (`expect(style.id).to eq(style.self_link)`, at https://github.com/citation-style-language/styles/blob/master/spec/dependent_styles_spec.rb#L38), the first test is rather redundant and can be removed. 

We could rewrite the second as:

```
  2) journal-of-hip-preserving-surgery (dependent): style ID must be a valid style repository link ("http://www.zotero.org/styles/" + style file name (without ".csl" extension))
     Failure/Error: expect(style.id).to eq("http://www.zotero.org/styles/#{id}")
       expected: "http://www.zotero.org/styles/journal-of-hip-preserving-surgery" got: "http://www.zotero.org/styles/journal-of-hip-preservation-surgery"
```

Also, it might be clearer if we rename "id" to "basename", since it's based on the filename instead of the style ID (see https://github.com/citation-style-language/styles/blob/master/spec/spec_helper.rb#L58 and  http://stackoverflow.com/questions/1094460/before-the-dot-in-a-file-name-what-is-it-called).

[...]

PR #1897, opened 2016-02-21
Update Travis tests

rmzelle:


rmzelle:
@inukshuk, do you know how `spec/independent_styles_spec.rb` and `spec/dependent_styles_spec.rb` are called? I couldn't find any references to them in `spec/spec_helper.rb`. I was also wondering whether it would be easy to define shared tests for independent and dependent styles, since many tests are the same for both style types.

rmzelle:
@zuphilip, @adam3smith, could you take a quick look to see if these changes make sense?

This PR also addresses @zuphilip's #1801.

spec/dependent_styles_spec.rb
```diff
@@ -1,82 +1,79 @@
-Dependents.each_pair do |id, (filename, path, style, reason)|
+Dependents.each_pair do |basename, (filename, path, style, reason)|
 
-  describe "dependent style #{id}" do
+  describe "#{basename} (dependent):" do
```
zuphilip: The colon `:` is most of the times not needed. E.g. 
```4or (dependent) must validate against the CSL 1.0.1 schema```
I think I would delete it again, at least in the default string.

Comment (zuphilip) [inline spec/dependent_styles_spec.rb]: Maybe it helps to clarify here
```#{basename} (dependent subdirectory)```
This will make the mentions of subdirectory below unneccessary.

File: spec/dependent_styles_spec.rb [hunk 2]
```diff
@@ -1,82 +1,79 @@
...
     end
 
-    it "was successfully parsed" do
+    it "must be parsable as a CSL style" do
       expect(style).to be_a(CSL::Style), reason
     end
 
     unless style.nil?
-      it "is dependent" do
+      it 'only dependent styles are allowed in the "dependent" subdirectory' do
```
zuphilip: This could then simply be
```it "must be a dependent style" do```

File: spec/dependent_styles_spec.rb [hunk 3]
```diff
@@ -1,82 +1,79 @@
...
         expect(style).to be_dependent
       end
 
-      it "has an info element" do
+      it "must contain an <info/> element" do
        expect(style).to have_info
       end
 
-      it "does not have any rendering elements" do
+      it '"dependent" styles may not contain <macro/>, <citation/>, or <bibliography/> elements' do
```
zuphilip: Should be enough to write
```it "may not contain <macro/>, <citation/>, or <bibliography/> elements" do```

File: spec/independent_styles_spec.rb [hunk 4]
```diff
@@ -1,79 +1,76 @@
...
         if style.has_template_link?
-          link = style.template_link
-
-          expect(link).to match(%r{http[s]?://www.zotero.org/styles/([a-z-]+)})
-          expect(Independents).to have_key(link[/[^\/]+$/])
+          parent_ID_link = style.template_link
+          link_prefix = "http://www.zotero.org/styles/"
+          
+          expect(parent_ID_link).to match(%r{^#{link_prefix}})
+          parent_ID = parent_ID_link[link_prefix.length..-1]
```
zuphilip: It would be better to replace the "parent"-part in the variable names here with "template".

zuphilip: BTW I haven't found a test that an independent style can't have a parent link. Is that test missing?

inukshuk:
@rmzelle all `*_spec.rb` files in the `spec` folder are loaded per rspec's default settings. You can put additional `*_spec.rb` there, for example with shared tests. If I remember correctly, I considered this, but decided the tests would be easier to read if I kept them together.

spec/independent_styles_spec.rb
inukshuk: The `independent?` predicate is [defined by the presence/absence of the link](https://github.com/inukshuk/csl-ruby/blob/master/lib/csl/style.rb#L108) so this is being tested implicitly. It may be helpful to add an explicit test with a more meaningful error message, or to change the 'must be independent' to 'may not contain a link to an independent parent'.

computer-und-recht.csl
```diff
@@ -5,7 +5,7 @@
     <title-short>CR</title-short>
     <id>http://www.zotero.org/styles/computer-und-recht</id>
     <link href="http://www.zotero.org/styles/computer-und-recht" rel="self"/>
-    <link href="https://www.zotero.org/styles/neue-juristische-wochenschrift" rel="template"/>
+    <link href="http://www.zotero.org/styles/neue-juristische-wochenschrift" rel="template"/>
```
inukshuk: Why don't we normalize all URLs to use https? Wouldn't that make for a better default?

inukshuk: Nevermind, I just saw the issue.

spec/dependent_styles_spec.rb
rmzelle: Agree that the colon isn't necessary. To stay as brief as possible, we could do:

```
apa must
dependent/nature-biotechnology must
```

spec/dependent_styles_spec.rb
rmzelle: Yes, but sometimes contributors are trying to contribute an independent style, but accidentally put it into the "dependent" subdirectory. In that case, their style shouldn't be a dependent style. They just have to move it to root.

spec/dependent_styles_spec.rb
rmzelle: Again, I'd like to be a little bit more wordy here in case the contributor misplaced an independent style.

spec/independent_styles_spec.rb
rmzelle: > It would be better to replace the "parent"-part in the variable names here with "template".

Yes.

rmzelle: > It may be helpful to add an explicit test with a more meaningful error message, or to change the 'must be independent' to 'may not contain a link to an independent parent'.

Yes, it would be better if the exact criteria for the test would be explained in the error message.

rmzelle:
> You can put additional *_spec.rb there, for example with shared tests.

I might consolidate things a bit, then, since there is so much overlap.

computer-und-recht.csl
rmzelle: Yes, @dstillman correctly mentioned that this isn't merely an implementation issue, since the "independent-parent" link is used both as a dereferenceable URI and as an identifier, so it's probably better to stick to HTTP for now.

I'd like to enforce HTTP for the "template" link as well for consistency for the time being.

spec/dependent_styles_spec.rb
zuphilip: Well, we agree above which will makes it then `dependent/nature-biotechnology only dependent styles are allowed in the "dependent" subdirectory` or `dependent/nature-biotechnology must be a dependent style` where you can still add `(check the directory you are adding your style)`

rmzelle: How about:

`dependent/nature-biotechnology must be a dependent style (independent styles must be placed in the root directory)`

zuphilip: :+1:

PR #1898, opened 2016-02-22
Minor update

jmiba:
Delimiters for authors etc. updated. Unnecessary font variant
attributes deleted.

### DIFF
--- a/technische-universitat-dresden-betriebswirtschaftslehre-logistik-author-date.csl
+++ b/technische-universitat-dresden-betriebswirtschaftslehre-logistik-author-date.csl
@@ -46,7 +46,7 @@
   </macro>
   <macro name="author-short">
     <names variable="author">
-      <name form="short" delimiter="/ " delimiter-precedes-last="never" et-al-min="3" et-al-use-first="1" initialize-with=". "/>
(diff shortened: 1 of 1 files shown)


technische-universitat-dresden-betriebswirtschaftslehre-logistik-author-date.csl
```diff
@@ -47,8 +46,7 @@
   </macro>
   <macro name="author-short">
     <names variable="author">
-      <name form="short" font-variant="normal" delimiter="/" delimiter-precedes-last="never" et-al-min="3" et-al-use-first="1" initialize-with=". "/>
-      <et-al font-variant="normal"/>
+      <name form="short" delimiter="/ " delimiter-precedes-last="never" et-al-min="3" et-al-use-first="1" initialize-with=". "/>
```
zuphilip: You should try to avoid to define the `et-al-min` and `et-al-use-first` in other nodes than `<bibliography>` or `<citation>`. Here you already defined the 3/1 rule in the [citation](https://github.com/jmiba/styles/blob/patch-8/technische-universitat-dresden-betriebswirtschaftslehre-logistik-author-date.csl#L123) and you can simply delete these two attributes here.

rmzelle:
Thanks!

PR #1901, opened 2016-02-23

roaldfrosig:
Looks for the law number in both the 'number' and the 'volume' variable. To comply with the recent Retsinformation.js - translator.


adam3smith:
thanks!

PR #1902, opened 2016-02-24
Update the-geological-society-of-america.csl

rdsteed:
Adds a group to the end of bibliography layout to support references to website content as well as document transferred from websites by adding "URL" and "accessed" date variables.  Follows instructions and examples  from http://www.geosociety.org/pubs/documents/GSA_RefGuide_Examples_000.pdf

### DIFF
--- a/the-geological-society-of-america.csl
+++ b/the-geological-society-of-america.csl
@@ -159,6 +159,13 @@
           </group>
         </else>
       </choose>
+      <group> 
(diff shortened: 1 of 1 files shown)


adam3smith:
(will wrap this with another fix. please don't merge yet @rmzelle )

adam3smith:
Thank you. I used your code with some extra tweaks in the fix referenced above.

PR #1903, opened 2016-02-24
Update nature.csl

icalandra:
This is the example from the Nature guidelines for citing a website:
"Buckley, M. and Reid, A., Global food safety - keeping food safe from farm to table. Technical report. (2010) Available at: insert website here. (Accessed: 4th November 2012)".

I have therefore edited lines 45-52 to format the URL and add the access date, and added lines 119-123 so that URL and access date are shown only for webpages. It works as intended (as far as I could check), but I'm not sure whether it is the best way to do so!

### DIFF
--- a/nature.csl
+++ b/nature.csl
@@ -41,15 +41,17 @@
       <else-if variable="DOI">
         <text variable="DOI" prefix="doi:"/>
       </else-if>
-      <else-if variable="URL">
-        <text variable="URL" prefix="Available at: " suffix=". "/>
-	<group prefix="(Accessed: " suffix=")">
+      <else-if type="webpage" variable="URL" match="all">        
+        <text term="available at" text-case="capitalize-first" suffix=": "/>
+        <text variable="URL" suffix=". "/>
+        <group prefix="(" suffix=")" delimiter=": ">
+          <text term="accessed" text-case="capitalize-first"/>
           <date variable="accessed">
             <date-part name="day" suffix=" " form="ordinal"/>
             <date-part name="month" suffix=" "/>
             <date-part name="year"/>
           </date>
-      	</group>      
+        </group>
       </else-if>
     </choose>
   </macro>
@@ -116,11 +118,7 @@
         </group>
         <text variable="page"/>
         <text macro="issuance"/>
-	<choose>
-	  <if type="webpage" match="all">
-            <text macro="access"/>
-	  </if>
-        </choose>
+        <text macro="access"/>
       </group>
     </layout>
   </bibliography>
--- a/nature.csl
+++ b/nature.csl
@@ -14,7 +14,7 @@
     <category field="generic-base"/>
     <issn>0028-0836</issn>
     <eissn>1476-4687</eissn>
-    <updated>2014-06-22T03:46:28+00:00</updated>
+    <updated>2014-09-10T06:11:43+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="title">
@@ -42,8 +42,14 @@
         <text variable="DOI" prefix="doi:"/>
       </else-if>
       <else-if variable="URL">
-        <text term="at"/>
-        <text variable="URL" prefix=" &lt;" suffix="&gt;"/>
+        <text variable="URL" prefix="Available at: " suffix=". "/>
+	<group prefix="(Accessed: " suffix=")">
+          <date variable="accessed">
+            <date-part name="day" suffix=" " form="ordinal"/>
+            <date-part name="month" suffix=" "/>
+            <date-part name="year"/>
+          </date>
+      	</group>      
       </else-if>
     </choose>
   </macro>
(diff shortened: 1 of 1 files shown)


rmzelle:
Where did you find that example? I don't see any concrete examples at http://www.nature.com/nature/authors/gta/index.html#a5.4.

icalandra:
Actually, I got that from Scientific Reports (http://www.nature.com/srep/publish/guidelines#references).

There, the guidelines say "Scientific Reports uses standard Nature referencing style". The Zotero style for Sci Rep is completely based on the one for Nature. So I assumed that Nature should cite websites as Sci Rep does. Am I wrong?

rmzelle:
Who knows. I'm not very impressed by Nature's inability to just give some clear examples in its style guide. I did find this example in Nature itself via Google:

> 2. Salmon, R. & Logan, A. Flaring Up: North Dakota Natural Gas Flaring More than Doubles in Two Years (Ceres, 2013); available at go.nature.com/jdks3y.

(http://www.nature.com/nature/journal/v518/n7539/abs/nature14199.html)

icalandra:
Ok, not very consistent indeed... So you might want to reject the pull request then. 
But could you please first tell me whether my correction is working as intended, and whether it could be made simpler? I'm still learning :)

rmzelle:
Well, currently you limit the "access" macro to webpage items, which probably isn't right since the "access" macro was originally also used to print the DOI for items without a volume number (e.g. early-access journal articles that haven't been assigned an issue yet). So you probably would want to move the item-type test into the macro.

@adam3smith, did you ever find a good authoritative source on the Nature style?

adam3smith:
I´ve had exactly the same issue with Nature. No idea what they want. I actually think going with @icalandra 's version, which is at least listed in a style guide by Nature *somewhere*, makes sense. @rmzelle -- are you OK with that?

@icalandra -- in addition to what @rmzelle says, we prefer if instead of hard-coding terms into the affixes, you set them via terms for `accessed` and `available at`, in this case both with text-case="capitalize-first". See http://docs.citationstyles.org/en/stable/specification.html#miscellaneous for a list of terms. Various advantages, including easier localization etc. Then also indent properly (2 spaces per level. You can use http://formatter.citationstyles.org/ ) and you're all set.

rmzelle:
> @rmzelle -- are you OK with that?

Yes.

> we prefer if instead of hard-coding terms into the affixes

Yeah, I forgot to mention that. So instead of

```xml
        <group prefix="(Accessed: " suffix=")">
          <date variable="accessed">
            <date-part name="day" suffix=" " form="ordinal"/>
            <date-part name="month" suffix=" "/>
            <date-part name="year"/>
          </date>
      	</group>
```

it would be better to have

```xml
        <group prefix="(" suffix=")" delimiter=": ">
          <date variable="accessed">
            <date-part name="day" suffix=" " form="ordinal"/>
            <date-part name="month" suffix=" "/>
            <date-part name="year"/>
          </date>
      	</group>
```

icalandra:
So, would it be correct?

<macro name="access">
    <choose>
      <if variable="volume"/>
      <else-if variable="DOI">
        <text variable="DOI" prefix="doi:"/>
      </else-if>
      <else-if type="webpage" match="all">        
        <text term="available at" text-case="capitalize-first" suffix=": "/>
        <text variable="URL" suffix=". "/>
        <group prefix="(" suffix=")" delimiter=": ">
          <date variable="accessed">
            <date-part name="day" suffix=" " form="ordinal"/>
            <date-part name="month" suffix=" "/>
            <date-part name="year"/>
          </date>
        </group>
      </else-if>
    </choose>
  </macro>

icalandra:
Why is my comment empty??

adam3smith:
To be safe, I'd do
`<else-if type="webpage" variable="URL" match="all">`       
unlikely someone has a webpage without URL, but who knows.
Looks good otherwise.

rmzelle:
Fixed that for you.

As for the code, ` match="all"` is unnecessary. You'll also want to use another group so that the "available at" term only prints if the URL variable isn't empty (since groups work as implicit conditionals, see http://docs.citationstyles.org/en/stable/specification.html#group).

So e.g.

```
        <group delimiter=": " suffix=". "/>
          <text term="available at" text-case="capitalize-first"/>
          <text variable="URL"/>
        </group>
```

Edit: @adam3smith's solution is better since it suppresses the accessed date if there is no URL, which is probably desirable.

icalandra:
So that would be

`<macro name="access">
    <choose>
      <if variable="volume"/>
      <else-if variable="DOI">
        <text variable="DOI" prefix="doi:"/>
      </else-if>
      <else-if type="webpage" variable="URL" match="all">        
        <text term="available at" text-case="capitalize-first" suffix=": "/>
        <text variable="URL" suffix=". "/>
        <group prefix="(" suffix=")" delimiter=": ">
          <date variable="accessed">
            <date-part name="day" suffix=" " form="ordinal"/>
            <date-part name="month" suffix=" "/>
            <date-part name="year"/>
          </date>
        </group>
      </else-if>
    </choose>
  </macro>`

adam3smith:
yes. Go ahead and update the pull request. If there are remaining issues we'll flag them there. Instructions are above.

icalandra:
Ok, done

adam3smith:
perfect, thanks!

rmzelle:
Thanks!

PR #1904, opened 2016-02-24

roaldfrosig:
Now looks in both the 'number' and the 'volume' field for the law-number. In order to comply with the recent Retsinformatin.dk translator.


adam3smith:
thanks!

PR #1910, opened 2016-02-26
Update associacao-brasileira-de-normas-tecnicas.csl

leonardopsantos:
Fixed Conference Proceedings missing a dot at the and of the conference's name


rmzelle:
Thanks!

PR #1916, opened 2016-03-02
Create HSU_ENGR.csl

djb489:
A modification of the ASCE citation style for use in by engineering students at Humboldt State University (or anyone else). This citation style only differs from the existing ASCE citation style by a single comma. More changes are anticipated.


rmzelle:
Continued in #1917.

(@djb489, it's generally preferable to correct your first pull request instead of opening a second one, but let's continue in #1917)

PR #1920, opened 2016-03-02
Create herpetological-conservation-and-biology.csl

tersiphone:
This is a newly created CSL style for the journal Herpetological Conservation and Biology.  Please do not yet post to the repository as it will require final approve from our Board of Directors before being made publicly available.  I am posting here as a pull request so that it can be preliminarily reviewed for technical correctness.

The instructions to authors can be found here:

http://www.herpconbio.org/InstructionsForAuthors.pdf.

### DIFF
--- a/herpetological-conservation-and-biology.csl
+++ b/herpetological-conservation-and-biology.csl
@@ -182,7 +182,7 @@
         <text variable="number" suffix=". "/>
         <text variable="URL" prefix="Available from " suffix=". "/>
         <group prefix="[" suffix="]">
-          <text value="accessed" text-case="capitalize-first" suffix=" "/>
+          <text term="accessed" text-case="capitalize-first" suffix=" "/>
           <date variable="accessed">
             <date-part name="day" form="numeric-leading-zeros" suffix=" "/>
             <date-part name="month" form="long" suffix=" "/>
--- a/herpetological-conservation-and-biology.csl
+++ b/herpetological-conservation-and-biology.csl
@@ -13,7 +13,6 @@
     <category field="biology"/>
     <category field="zoology"/>
     <issn>1931-7603</issn>
-    <eissn>1931-7603</eissn>
     <updated>2016-03-02T15:22:52</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
--- a/herpetological-conservation-and-biology.csl
+++ b/herpetological-conservation-and-biology.csl
@@ -93,17 +93,21 @@
         <choose>
           <if variable="collection-title">
             <choose>
-              <if is-numeric="edition">
-                <group delimiter=" ">
-                  <number text-case="capitalize-first" variable="edition" form="ordinal"/>
-                  <text term="edition" text-case="capitalize-first" suffix=". "/>
-                </group>
+              <if variable="edition">
+                <choose>
+                  <if is-numeric="edition">
+                    <group delimiter=" ">
+                      <number text-case="capitalize-first" variable="edition" form="ordinal"/>
+                      <text term="edition" text-case="capitalize-first" suffix=". "/>
+                    </group>
+                  </if>
+                  <else>
+                    <text variable="edition"/>
+                    <number text-case="capitalize-first" variable="edition" form="ordinal"/>
+                    <text term="edition" text-case="capitalize-first" suffix=". "/>
+                  </else>
+                </choose>
               </if>
-              <else>
-               <text variable="edition"/>
-               <number text-case="capitalize-first" variable="edition" form="ordinal"/>
-               <text term="edition" text-case="capitalize-first"  suffix=". "/>
-              </else>
             </choose>
             <group delimiter=" ">
               <text variable="collection-title" text-case="capitalize-all"/>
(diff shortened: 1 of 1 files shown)


tersiphone:
If I understand this failure correctly, it has to do with the fact that the ISSN and eISSN numbers indicated in the style are the same?  If that is the case, then I am a bit stuck because the journal only has an ISSN and not an eISSN.  I guess I can just remove the latter from the style?

If the failure was caused by something else, please let me know.

adam3smith:
yes, the fields are not obligatory. If there's not a separate eISSN, just use the ISSN.

tersiphone:
Done.

herpetological-conservation-and-biology.csl
```diff
@@ -0,0 +1,223 @@
...
+    <category citation-format="author-date"/>
+    <category field="biology"/>
+    <category field="zoology"/>
+    <issn>1931-7603</issn>
+    <updated>2016-03-02T15:22:52</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <macro name="author">
+    <names variable="author" suffix=".">
+      <name and="text" delimiter-precedes-last="always" et-al-min="10" et-al-use-first="10" initialize-with="." name-as-sort-order="first"/>
```
adam3smith: set the `et-al` below in citation, so you only have to do this once. Also, you want the `et-al-min` to always be at least one bigger than the `et-al-use-first`. Having them at the same number is non-sensical (it'd imply using et al for an article with 10 authors and printing the first ten authors)

herpetological-conservation-and-biology.csl
```diff
@@ -0,0 +1,223 @@
...
+      <label prefix=", "/>
+      <substitute>
+        <text macro="editor"/>
+      </substitute>
+    </names>
+  </macro>
+  <macro name="editor">
+    <names variable="editor">
+      <name and="text" delimiter-precedes-last="always" et-al-min="10" et-al-use-first="10" initialize-with="." name-as-sort-order="first"/>
```
adam3smith: see above

herpetological-conservation-and-biology.csl
```diff
@@ -0,0 +1,223 @@
...
+      <label form="short" text-case="capitalize-first" prefix=" (" suffix=")"/>
+    </names>
+  </macro>
+  <macro name="author-short">
+    <names variable="author">
+      <name form="short" delimiter="" and="text" et-al-min="3" et-al-use-first="1" initialize="false" initialize-with="" sort-separator=""/>
```
adam3smith: move the `et al` here to `<citation>`

herpetological-conservation-and-biology.csl
```diff
@@ -0,0 +1,223 @@
...
+        <text variable="number-of-pages" suffix=" p."/>
+      </if>
+    </choose>
+  </macro>
+  <macro name="webpage">
+    <choose>
+      <if type="webpage">
+        <text variable="number" suffix=". "/>
+        <text variable="URL" prefix="Available from " suffix=". "/>
+        <text value="accessed" text-case="capitalize-first" prefix="[" suffix=" "/>
```
adam3smith: why not `<text term...`?

adam3smith: I'd wrap the accessed term and date in a group with `prefix="["` and `suffix="]"` It's easier to read than the two separated brackets, but it'll also not look weird in the (admittedly unlikely) case that the accessed date is missing.

tersiphone:
Thanks @adam3smith.  I made all the changes you suggested, except the one "why not <text term...?"  I wasn't sure which term this was referring to.  Is it the "Accessed from"?  If so, I think I tried making that a term at first but it never rendered, which is why I made it a prefix.

Otherwise...?

herpetological-conservation-and-biology.csl
```diff
@@ -0,0 +1,229 @@
...
+      </if>
+    </choose>
+  </macro>
+  <macro name="webpage">
+    <choose>
+      <if type="webpage">
+        <text variable="number" suffix=". "/>
+        <text variable="URL" prefix="Available from " suffix=". "/>
+        <group prefix="[" suffix="]">
+          <text value="accessed" text-case="capitalize-first" suffix=" "/>
```
adam3smith: I meant this one: `<text term="accessed" text-case="capitalize-first" suffix=" "/>` should do exactly the same (and is more robust for future changes and localization to other languages)

tersiphone:
Right.  Done.

adam3smith:
(you can only do that with predefined terms, which is why "available from" didn't work. It's possible to do that, too, but it's a bit involved and I don't think it's worth the effort here. What you have now will work exactly as you want it to.)

adam3smith:
Thanks! We really appreciate journals helping to provide their own styles, it's a huge help.

tersiphone:
No worries.  Thanks for all your help!

What is the next step after I get editorial board approval for this to be hosted on the CSL and/or Zotero repository?

adam3smith:
this will be on the Zotero repository within the hour and available through Mendeley within about 24hs. That happens automatically.

tersiphone:
Yikes!!!  Can we prevent that from happening?  We need to get Board of Directors approval before it can be posted in the repositories.

adam3smith:
I have taken the style down. Let me know when you want me to put it up again, I saved a copy.

tersiphone:
Thanks.  Hopefully it will be in the coming days.

PR #1921, opened 2016-03-03
Consolidate RSpec tests

rmzelle:
This mostly eliminates doubly defined tests.


rmzelle:
@adam3smith, @inukshuk, I reorganized the style tests into a single file, an deduplicated almost all the shared tests. Should make it easier to keep things synchronized.

If I don't hear any objections I'll merge this in a day or two.

rmzelle:
I also found that running the test suite on a subset of styles often gives errors since not all styles are loaded.

E.g. `CSL_TEST="acs-nano.csl" bundle exec rake` gives the erroneous error:
> acs-nano: "template" link must point to an existing independent style

For some tests it would be enough to separately store all the file names of the independent styles (`'"template" link must point to an existing independent style'` and `'"independent-parent" link must point to an existing independent style'`). However, for one of the tests this is not enough, and we'd have to load the parent (`"must have the same citation-format as its independent-parent"`). But if we preload all independents that would make the `CSL_TEST` option quite a bit slower.

I don't use `CSL_TEST` very often, so it might not be worth to fix this.

inukshuk:
I've only ever used this for independent styles, but we could just load independent parents automatically when loading dependent styles via `CSL_TEST` -- that would still be much quicker than loading everything.

rmzelle:
> we could just load independent parents automatically when loading dependent styles via CSL_TEST

This PR should take care of that: https://github.com/citation-style-language/styles/pull/1938

With this only the parents of tested dependents are loaded, so it's still rather snappy.

PR #1914, opened 2016-03-01
Clara-revue-style-fr.csl

clararevue:
Citation style for the scientific journal Clara: Recherche/Architecture

PR #1923, opened 2016-03-03
Update and rename clara-recherche-architecture.csl to clara-architect…

clararevue:
…ure-recherche.csl


rmzelle:
Oh, sorry, did I recommend the wrong name? I think I copied "CLARA Recherche / Architecture" from http://editionsmardaga.com/Clara-no3-2015 but it should indeed be the other way around.

rmzelle:
Thanks! (fix of #1914)

PR #1924, opened 2016-03-03
Biologia

matejovo:


rmzelle:
Did you write this style from scratch? If you used another CSL style as template, please add a "template" link (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link).

And it doesn't really look like a "geography" journal, so could you remove `<category field="geography"/>`?

In addition, could you edit the style to add this as the first line?: `<?xml version="1.0" encoding="utf-8"?>`

rmzelle:
Oh, and we prefer it if you remove any authors that didn't contribute to this style (the "template" link should take care of any attribution). My guess is that you can remove

```
    <author>
      <name>Jorge Pinzon</name>
    </author>
```

He didn't work on the Biologia style, did he?

Let us know if you need any help editing this pull request (see also the instructions above in https://github.com/citation-style-language/styles/pull/1924#issuecomment-191906273)

rmzelle:
Great, thanks!

PR #1925, opened 2016-03-05
journal VTEI

libora:
Journal VTEI changed guide for authors. I deleted old dependent style and create 2 new independent styles. One for Czech and one for English (see https://forums.zotero.org/discussion/56868/set-up-delimiterprecedeslast-for-different-locale)


rmzelle:
Continued in #1926.

PR #1926, opened 2016-03-05
VTEI

libora:
created new independent style a deleted old dependent style because the change of guide for authors


rmzelle:
Thanks, but could you please create a new pull request and just change the existing file, instead of deleting and re-adding it. It's much easier for us to check the history of the file that way.

rmzelle:
Oh, wait, never mind. I didn't notice that you're replacing a dependent with an independent style.

rmzelle:
Thanks!

PR #1927, opened 2016-03-05
Create water-alternatives.csl

libora:
style for journal Water Alternatives, output from style was agreed by redaction.


rmzelle:
Great, thanks!

PR #1928, opened 2016-03-06
Adding option for type of speech.

Jmuccigr:


fold-and-r.csl
```diff
@@ -140,6 +140,16 @@
           <text variable="URL" prefix=" "/>
         </else>
       </choose>
+      <choose>
+        <if type="speech">
+          <group prefix=", at "  delimiter=", ">
```
adam3smith: you don't need the groups here (you never need groups with one element). Just do
```
<group suffix=".">
  <text variable="event" prefix=", at "/>
  <text variable="publisher-pace" prefix=", in "/>
</group>
```
which is also more robust than what you have (in case of a missing place variable).

Jmuccigr: Thanks for the tip. I've changed it further, since I realized that it was probably better just to have a separate `else-if` for the speech type.

(Every time I return to editing these files, I have to re-learn whatever little part of the csl I'm playing with.)

fold-and-r.csl
```diff
@@ -129,6 +129,15 @@
             </group>
           </group>
         </else-if>
+        <else-if type="speech">
+          <group delimiter=" ">
+            <text macro="title" prefix="'" suffix="'"/>
```
adam3smith: did we ever talk about why you're using straight quotes in the affixes rather than quotes="true" (which would be more elegant and produce nice curly quotes?

adam3smith: (you'd need to change the default-locale to `en-GB` for that to work.

Jmuccigr: No, I don't believe we've had that chat. :-)

The journal's guidelines call for single quote marks. So would that then be:

`<text macro="title" quotes="true"/>`

in the appropriate groups, along with the locale change up top? That seems to do it for me.

I also noted the guidelines call for a leading "in" before the journal title, so I added that too.

adam3smith:
Great! Last question -- at least for chapter it looks like you'd also want quotes and the current style would just print the title. I think just setting `quotes="true"` in https://github.com/citation-style-language/styles/pull/1928/files#diff-a03802aca1da523688424503e1158fcfR50 would work, but maybe there are instances where you just want a plain title (didn't see any in the styleguide). 
(Sorry, this is probably more than you bargained for, but it's always a good opportunity to combine edits on a style with some light QA).

Jmuccigr:
Isn't it **always** more than you bargain for? ;-)

The guidelines aren't super clear, but yes, I think the chapter needs the quote marks too. I also fixed the editor macro while I was at it. I'm sure there's more.

fold-and-r.csl
```diff
@@ -119,8 +119,8 @@
         </else-if>
         <else-if type="article-journal">
           <group delimiter=" ">
-            <text macro="title" prefix="'" suffix="'"/>
-            <text term="in" suffix=" "/>
+            <text macro="title" quotes="true"/>
+            <text term="in" prefix=", " suffix=" "/>
```
adam3smith:
```<text macro="title" quotes="true" suffix=","/>
<text term="in"/>
<text macro="journal"/>
```

the way you have this set up right now, you're getting a space between quotation mark and comma because of the group delimiter.

fold-and-r.csl
```diff
@@ -129,6 +129,15 @@
             </group>
           </group>
         </else-if>
+        <else-if type="speech">
+          <group delimiter=" ">
+            <text macro="title" quotes="true"/>
+            <group suffix=".">
+              <text variable="event" prefix=", at "/>
```
adam3smith: same here, actually -- you can't combine a space in a delimiter with a comma in a prefix. Here I'd remove the delimiter, which doesn't acutally do anything anymore.

fold-and-r.csl
Jmuccigr: So the group delimiter occurs between all members of the group? (Told you I didn't know how this all works!)

So, either delete the delimiter, or I could make the delimiter a comma and remove the comma in the prefixes, right? Actually I think I could then just remove the prefixes altogether.

adam3smith: What I have above (if placed in the group) would be one solution, i.e. put the comma in the suffix. When you want the comma in the delimiter, you'd need

```
<group delimiter=", ">
  <text macro="title" quotes="true"/>
  <group delimiter=" ">
    <text term="in"/>
    <text macro="journal"/>
  </group>
</group>
```

Hope that makes sense. The latter is considered better style.

Jmuccigr: Nope, I got lost.

What's better style? (And note that my current version uses the quotes="true" bit.)

adam3smith: The problem with your current version is the space it puts _before_ the comma, so that's clearly wrong.
If you just place everything in one group with ", " as the delimiter, you'll also get a comma between "in" and the journal title --> also wrong.
So I give you two options: either you place everything in one group with delimiter=" " and put a comma in a suffix on title (that's what I have four posts above) or you nest two groups with different delimiters (two posts above). Both produce the same output in most cases, but the second option is more elegant and will work sensibly even in the absence of a journal title (a very rare case, presumably).

Jmuccigr: OK, that seems to work. Thanks for your patience!

PR #1929, opened 2016-03-06
Update water-alternatives.csl

libora:
1. add section for type Interview into the macro Locators
2. deleted empty lines and change space -> tab


rmzelle:
Thanks!

(you can use http://formatter.citationstyles.org/ to quickly reformat your style with the right kind of indentation)

PR #1915, opened 2016-03-02
Create victoria-university-australia-harvard.csl

SunwayCollege:
Victorial UNiversity Melbourne Australia Harvard style is created based on http://libraryguides.vu.edu.au/harvard

PR #1931, opened 2016-03-07
Victoria University Harvard style uploaded and renamed

SunwayCollege:
Applied changes as recommended 
- Change file name
- change self link


rmzelle:
Is this the same as #1915?

SunwayCollege:
Yes. It is similar to #1915 but with the recommended changes as applied to 2645c04

rmzelle:
I accepted your style from #1915. Thanks for contributing!

Sorry for letting that pull request sit for a few days. I hadn't seen you had updated it. (in general, it's almost always better to stay in your original pull request; next time, feel free to leave a comment if you think you've been forgotten)

SunwayCollege:
Noted. Thank you for the advice.

PR #1938, opened 2016-03-09
Improve CSL_TEST support

rmzelle:
* Always load the parents of dependents that are being checked (so we can make sure the “citation-format” is the same between parent and dependent)
* Always use a list of all independents for looking up “template” and “independent-parent” IDs

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -146,4 +146,46 @@ Independents = Hash[collect_styles.each_with_index.map { |path, i|
   load_style(path)
 }]
 
+# Make sure we always have the basenames of all independent styles stored
+if ENV['CSL_TEST'] != nil
+  INDEPENDENTS_BASENAMES = Dir[File.join(STYLE_ROOT, '*.csl')].map { |path|
+    filename = File.basename(path)
+    basename = filename[0..-5]
+  }
+else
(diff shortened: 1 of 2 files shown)


rmzelle:
@inukshuk, can you take a quick look to check whether I coded according to Ruby convention? I wasn't sure about the variable naming (especially when it comes to capitalization; all-caps are constants, but otherwise ...?)

Verified that this works with:

```bash
$ CSL_TEST="academy-of-ma.*" bundle exec rake
$ CSL_TEST="acs-nano.csl" bundle exec rake
$ bundle exec rake
```

There are still a few tests that can silently fail (e.g. the check for uniqueness of ISSNs), but I don't really see how we can do that without loading all styles, which would mostly defeat the purpose of "CSL_TEST" (i.e. quick tests). At least we shouldn't have false positives anymore now.

inukshuk:
Yes that's great.

CSL_TEST is an advanced flag, so I would expect users (basically that's only us anyway) to be aware that it won't catch duplicate ISSNs or similar cases.

Constants are capitalized (usually all caps), local variables and methods are usually snake case ('template_id' would be more common than 'template_ID') and classes are pascal case.

rmzelle:
Okay, thanks!

PR #1940, opened 2016-03-09
Patch 4

libora:
Update one style and add new style

### DIFF
--- a/vodohospodarske-technicko-ekonomicke-informace-en.csl
+++ b/vodohospodarske-technicko-ekonomicke-informace-en.csl
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" default-locale="en-GB">
   <info>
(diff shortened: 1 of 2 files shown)


adam3smith:
@rmzelle -- are you OK with the filename here? Then I'd add it to the ISSN exception list. (context [here](https://forums.zotero.org/discussion/56868/set-up-delimiterprecedeslast-for-different-locale/#Comment_248370), we need separate styles because of non-localizable differences.

rmzelle:
I'm fine with the filename, but the titles of the Czech and English version should be aligned. Could we use "Vodohospodářské technicko-ekonomické informace (Czech)" and "Vodohospodářské technicko-ekonomické informace (English)"?

(if there is only one style format per journal, we usually don't include descriptors like "numeric, brackets" in the title)

I already added the ISSN exception (and restarted the Travis build): https://github.com/citation-style-language/styles/commit/e39144c2561f8925ae02295ae50a8ca5b1ea4674

libora:
Could we use "Vodohospodářské technicko-ekonomické informace (English)"? - of course yes, I will change, take me a moment

rmzelle:
Thanks!

PR #1943, opened 2016-03-09
Update universitat-bremen-institut-fur-politikwissenschaft.csl

JulianKoepff:
- updating the guideline document's URL
- updating contributor's e-mail address
- including URL for encyclopedy entries and blog posts
- including volume numbers for books and chapters
- removing space after volume number for journal articles if issue number is not present
- including date for interviews
- including duration for broadcast, motion_picture song and interview via note (not very elegant solution)
- for journal articles: replacing access date and volume/issue/pages for DOI if present

### DIFF
--- a/universitat-bremen-institut-fur-politikwissenschaft.csl
+++ b/universitat-bremen-institut-fur-politikwissenschaft.csl
@@ -205,7 +205,7 @@
     </choose>
     <choose>
       <if type="chapter" match="any">
-	    <group delimiter=", " prefix=" " suffix=".">
+        <group delimiter=", " prefix=" " suffix=".">
           <text variable="container-title" text-case="title"/>
           <choose>
             <if variable="volume" match="all">
@@ -215,12 +215,12 @@
               <text variable="number-of-volumes" suffix=" Bände"/>
             </else-if>
           </choose>
-		</group>
+        </group>
       </if>
       <else-if type="webpage" match="any">
         <text variable="container-title" suffix="."/>
       </else-if>
-	  <else-if type="interview" match="any">
+      <else-if type="interview" match="any">
         <text value="Interview am" text-case="capitalize-first"/>
       </else-if>
       <else>
--- a/universitat-bremen-institut-fur-politikwissenschaft.csl
+++ b/universitat-bremen-institut-fur-politikwissenschaft.csl
@@ -69,7 +69,7 @@
     <choose>
       <if variable="URL">
         <choose>
-          <if type="article-journal" variable="URL DOI" match="all">
+          <if type="article-journal" variable="DOI" match="all">
             <choose>
               <if variable="volume" match="none">
                 <group>
@@ -79,6 +79,26 @@
               </if>
             </choose>
           </if>
+          <else-if>
+            <choose>
+              <if type="article-journal">
+                <choose>
+                  <if variable="DOI volume" match="none">
+                    <text value="Text abrufbar unter:" prefix=" " suffix=" "/>
+                    <text variable="URL"/>
+                    <group prefix=" (" suffix=")">
+                      <text value="Zugriff am" suffix=" "/>
+                      <date variable="accessed">
+                        <date-part name="day" form="numeric" suffix="."/>
+                        <date-part name="month" form="numeric" suffix="."/>
+                        <date-part name="year" form="long"/>
+                      </date>
+                    </group>
+                  </if>
+                </choose>
+              </if>
+            </choose>
+          </else-if>
           <else-if type="article-journal" variable="DOI" match="none">
             <text value="Text abrufbar unter:" prefix=" " suffix=" "/>
             <text variable="URL"/>
--- a/universitat-bremen-institut-fur-politikwissenschaft.csl
+++ b/universitat-bremen-institut-fur-politikwissenschaft.csl
@@ -79,23 +79,19 @@
               </if>
             </choose>
           </if>
-          <else-if>
+          <else-if type="article-journal">
             <choose>
-              <if type="article-journal">
-                <choose>
-                  <if variable="DOI volume" match="none">
-                    <text value="Text abrufbar unter:" prefix=" " suffix=" "/>
-                    <text variable="URL"/>
-                    <group prefix=" (" suffix=")">
-                      <text value="Zugriff am" suffix=" "/>
-                      <date variable="accessed">
-                        <date-part name="day" form="numeric" suffix="."/>
-                        <date-part name="month" form="numeric" suffix="."/>
-                        <date-part name="year" form="long"/>
-                      </date>
-                    </group>
-                  </if>
-                </choose>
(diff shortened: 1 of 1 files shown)


rmzelle:
Could you run your style through http://formatter.citationstyles.org/ as well? Your indentation is a bit off.

JulianKoepff:
Reformatted the style

rmzelle:
@zuphilip, would you mind giving this a quick look-over? You last edited this style.

zuphilip:
I am afraid that I will not have time to look at this closely in the next two weeks. BTW I don't have more insights on the style and last time just corrected some space errors which was reported through the forum.

universitat-bremen-institut-fur-politikwissenschaft.csl
```diff
@@ -116,10 +132,15 @@
   </macro>
   <macro name="locator">
     <choose>
-      <if type="article-journal">
-        <text variable="volume" suffix=" "/>
-        <text variable="issue" prefix="(" suffix=")"/>
+      <if type="article-journal" variable="DOI">
```
adam3smith: please make `match="all"` explicit here.

universitat-bremen-institut-fur-politikwissenschaft.csl
```diff
@@ -116,10 +132,15 @@
...
+        <text variable="DOI" prefix="DOI: "/>
       </if>
+      <else-if type="article" match="any">
```
adam3smith: are you sure this is right? Don't you want `type="article-journal"` here? For what (Mendeley I assume?) item type do you want this?

universitat-bremen-institut-fur-politikwissenschaft.csl
```diff
@@ -233,7 +267,11 @@
           <text macro="published-date"/>
           <choose>
             <if match="any" variable="page">
-              <text macro="pages"/>
+              <choose>
+                <if match="none" variable="DOI">
```
adam3smith: no, this isn't right. The instructions say to only do this for e-only publication (Reine E-Journals). You're getting rid of page ranges for all journal articles with DOIs, which is pretty much any journal article published in the last 5 years (probably longer).

universitat-bremen-institut-fur-politikwissenschaft.csl
adam3smith: and as below, this is a misunderstanding of the instructions. The Schimmelfennig article is an example for an E-Journal, not for any journal article with a DOI. You should reverse this: Only where articles don't have a volume number (as is the case for most E-Journals) print the DOI.

universitat-bremen-institut-fur-politikwissenschaft.csl
```diff
@@ -67,24 +67,40 @@
...
-          <group prefix=" (" suffix=")">
-            <text value="Zugriff am" suffix=" "/>
-            <date variable="accessed">
-              <date-part name="day" form="numeric" suffix="."/>
-              <date-part name="month" form="numeric" suffix="."/>
-              <date-part name="year" form="long"/>
-            </date>
-          </group>
-        </group>
+      <if type="webpage article-journal article-magazine article-newspaper article entry-encyclopedia speech post-weblog interview" variable="URL" match="all">
```
adam3smith: I'm not sure why we're testing for item type here at all? These would seem to be general rules for all item types, no?

JulianKoepff:
Updated the pull request

universitat-bremen-institut-fur-politikwissenschaft.csl
```diff
@@ -67,24 +67,44 @@
...
-            <date variable="accessed">
-              <date-part name="day" form="numeric" suffix="."/>
-              <date-part name="month" form="numeric" suffix="."/>
-              <date-part name="year" form="long"/>
-            </date>
-          </group>
-        </group>
+      <if variable="URL">
+        <choose>
+          <if type="article-journal" variable="URL DOI" match="all">
```
adam3smith: delete URL here, you're already testing for it two lines above (doesn't do any harm, but is redundant).

universitat-bremen-institut-fur-politikwissenschaft.csl
```diff
@@ -67,24 +67,44 @@
...
+            <choose>
+              <if variable="volume" match="none">
+                <group>
+                  <text value="Text abrufbar unter:" prefix=" " suffix=" "/>
+                  <text variable="URL"/>
+                </group>
+              </if>
+            </choose>
+          </if>
+          <else-if type="article-journal" variable="DOI" match="none">
```
adam3smith: How about you just do `<else>` here -- that way you'd still get the full online information for online journal articles without DOIs, which may very well exist (though they should be rare). Also, currently, if you had, say, a chapter with a DOI, it would print neither DOI nor URL, which seems wrong to me.

adam3smith:
much better. Two quick fixes, then we should be ready to merge.

universitat-bremen-institut-fur-politikwissenschaft.csl
JulianKoepff: The instructions say that if an article has a volume and issue number then the URL should be excluded even if we got it from an internet source (p. 30). In fact, we want to print DOI numbers only for journal articles that are exclusively published online.

adam3smith: OK, so else doesn't work,  but neither does what you have right now: 
If you have an article with URL, no volume/issue, no DOI, you won't get anything -- no URL, no DOI.

JulianKoepff: You are right. I now repeated the command for this very case only which is not very elegant. Do you know a better solution?

adam3smith: not really, no. If you want to you can put 82-93 in a macro, that way you don't have to duplicate the code, but in terms of the `<choose>` structure, I don't think that can be simplified more, at least not in a way I can think of.

JulianKoepff:
Sorry, I don't understand what the test report wants me to change. Can you help me?

universitat-bremen-institut-fur-politikwissenschaft.csl
```diff
@@ -67,24 +67,64 @@
...
+            <choose>
+              <if variable="volume" match="none">
+                <group>
+                  <text value="Text abrufbar unter:" prefix=" " suffix=" "/>
+                  <text variable="URL"/>
+                </group>
+              </if>
+            </choose>
+          </if>
+          <else-if>
```
adam3smith: you can't have an `<else-if>` without an actual condition. I think you just want `<else-if type="article-journal"/>` here and then get rid of the extra text you have in 83+84

adam3smith:
Sorry, the failure to validate error message is completely incomprehensible, not your fault.

JulianKoepff:
Yes!! :rocket:

adam3smith:
Thanks!

PR #1918, opened 2016-03-02
Create FAPPZ_v01

matejovo:
citation style for FAFNR of CULS according http://studium.agrobiologie.cz/Nastenka_studijniho_oddeleni/Bakalarske_a_diplomove_prace/Zavazna_pravidla_citace_seznamy_pouzite_literatury.pdf

PR #1944, opened 2016-03-10
Improve feedback non-validating/non-parsing styles

rmzelle:
One more. This rolls back https://github.com/inukshuk/styles/commit/94bcce850fdd06fdd4ed31bbece971d679f32748 (see #1818), since we recently came to the conclusion that Nokogiri’s validation error messages aren’t very good: https://github.com/citation-style-language/styles/pull/1918#issuecomment-193882842

It’s probably much less confusing to steer users to our Jing-based validator, which usually gives much clearer validation messages.

With this pull request we also skip the “does-this-style-parse” check for invalid styles.

cc @inukshuk, @adam3smith

### DIFF
--- a/spec/styles_spec.rb
+++ b/spec/styles_spec.rb
@@ -1,14 +1,24 @@
-shared_examples "style" do |basename, (filename, path, style, reason), in_dependent_subdir|
-  it "must validate against the CSL 1.0.1 schema" do
-    expect(CSL.validate(path)).to eq([])
+shared_examples "style" do |basename, (filename, path, style), in_dependent_subdir|
+  let(:style_validates) { 
+    if CSL.validate(path).length == 0
+      true
+    else
+      false
+    end
+  }
+  
+  it "must validate against the CSL 1.0.1 schema (please check your style at http://validator.citationstyles.org/)" do
+    expect(style_validates).to be true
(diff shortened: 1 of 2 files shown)


rmzelle:
With this the feedback for non-validating styles is:

```
  1) acs-nano: must validate against the CSL 1.0.1 schema (please check your style at http://validator.citationstyles.org/)
     Failure/Error: expect(style_validates).to be true
     
       expected true
            got false
```

spec/styles_spec.rb
```diff
@@ -1,14 +1,24 @@
-shared_examples "style" do |basename, (filename, path, style, reason), in_dependent_subdir|
-  it "must validate against the CSL 1.0.1 schema" do
-    expect(CSL.validate(path)).to eq([])
+shared_examples "style" do |basename, (filename, path, style), in_dependent_subdir|
+  let(:style_validates) {
```
inukshuk: You can just say `let :style_validates { CSL.validate(path).length.zero? }`

inukshuk:
We could try to determine whether or not the test is being run on Travis and in that case try to determine the URL to the failed style (which must be available on GitHub after all) and include it in the link so we can link straight to the validation results.

PR #1945, opened 2016-03-10
Show ISSN if not ISBN - in styles based on ČSN ISO690:2011

libora:
Only Mendely users will see the change in output from style; Zotero has not implemented ISSN variable for a lot of document types now.
Some commits have incorrect description "for books" (my mistake); this change is for all document types, which have ISBN.

### DIFF
--- a/iso690-author-date-cs.csl
+++ b/iso690-author-date-cs.csl
@@ -445,15 +445,13 @@
     <choose>
       <if variable="DOI">
         <group>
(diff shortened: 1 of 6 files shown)


adam3smith:
This looks good to me -- ready on your end @libora ?

libora:
Yes, thats all.

adam3smith:
thanks!

PR #1946, opened 2016-03-13
Some adjustments

jmiba:
Minor changes in name delimiters and an updated et-al term for German.

### DIFF
--- a/technische-universitat-dresden-betriebswirtschaftslehre-logistik-author-date.csl
+++ b/technische-universitat-dresden-betriebswirtschaftslehre-logistik-author-date.csl
@@ -200,12 +200,10 @@
                 <text variable="event-place" suffix=", "/>
               </else-if>
               <else-if type="webpage" match="any">
-              <choose>
-                  <if match="any" variable="container-title">
-                    <text term="in" text-case="capitalize-first" suffix=": "/>
-                    <text variable="container-title" suffix=", "/>
-                  </if>
-                </choose>
+                <group delimiter=": " suffix=", ">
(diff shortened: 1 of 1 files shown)


technische-universitat-dresden-betriebswirtschaftslehre-logistik-author-date.csl
```diff
@@ -189,15 +190,23 @@
...
                 <text variable="event-place" suffix=", "/>
               </else-if>
+              <else-if type="webpage" match="any">
+              <choose>
+                  <if match="any" variable="container-title">
+                    <text term="in" text-case="capitalize-first" suffix=": "/>
+                    <text variable="container-title" suffix=", "/>
+                  </if>
+                </choose>
+              </else-if>
```
rmzelle: It's more concise to do:

```xml
              <else-if type="webpage" match="any">
                <group delimiter=": " suffix=", ">
                </group>
              </else-if>
```

instead of 

```xml
              <else-if type="webpage" match="any">
              <choose>
                  <if match="any" variable="container-title">
                    <text variable="container-title" suffix=", "/>
                  </if>
                </choose>
              </else-if>
```

Groups are implicit conditionals. If a `<group/>` calls one or more variables, any terms in the group are automatically suppressed if the variables are empty. See http://docs.citationstyles.org/en/stable/specification.html#group for background.

jmiba: That makes a lot of sense, thank you. I will change that in a minute.

rmzelle:
Great, thanks!

PR #1947, opened 2016-03-13
Create o-bib.csl

### LINKED ISSUE #1835: o-bib
The Open Access library journal <a href="https://www.o-bib.de/">o-bib</a> will use a standard style beginning in the next year:

" Ab Heft 1/2016 wird kein eigener Zitierstil mehr zugrunde gelegt, sondern der Chicago-Stil (deutsch, mit Fußnoten) verwendet."

Source: Südwest-Info: Mitteilungsblatt des VDB-Regionalverbands Südwest Nr. 28 (2015), S. 21. http://www.vdb-online.org/landesverbaende/sw/sw-info/

zuphilip:
This will close #1835

### DIFF
--- a/dependent/o-bib.csl
+++ b/dependent/o-bib.csl
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" default-locale="de-DE">
+  <info>
+    <title>o-bib</title>
+    <id>http://www.zotero.org/styles/o-bib</id>
+    <link href="http://www.zotero.org/styles/o-bib" rel="self"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
cool dass die das machen. Und wie immer besten Dank.

rmzelle:
Thanks!

Added "German" to the title: https://github.com/citation-style-language/styles/commit/ab68e564063d4635ef7b87efa4dcb74e40ca6ce6

zuphilip:
Sorry @rmzelle , I forgot the "German" in the title.

rmzelle:
(it should be possible to test for this with Sheldon, by the way)

adam3smith:
was just thinking that, since I keep forgetting. My suggestion would be to see if we have a default locale other than en and then check for parentheses in the title. 
Alternatively we'd have to put a whole locale mapping table in the test file, which I suppose we could do, too -- it's only 40 or so value pairs.

PR #1948, opened 2016-03-14
Change "container-title-short" to italic

scottcove:
Fixed the omission of the italics in the journal title in footnotes.


rmzelle:
Thanks!

PR #1951, opened 2016-03-14
Optimization

jmiba:
On second thoughts about @rmzelle’s comment about groups


rmzelle:
Right, that should be redundant.

PR #1962, opened 2016-03-18
update resuscitation.csl

Phil54520:
The old style based on elsevier is slightly different. This one match better to the style of the Resuscitation journal


Phil54520:
I am not able to update this style because it has been created on a dependent style. How am I suppose to propose an update?

adam3smith:
You´ll have to delete the resuscitation.csl file in the dependent folder (instructions here: https://github.com/blog/1545-deleting-files-on-github). You also need to fix the template link to point to an actual styles.
If deleting doesn't work for you (might be tricky simply because there are so many files in that folder), please just fix the template link and then let us know, we can delete on our end.

rmzelle:
(you can delete the "dependent" version at https://github.com/Phil54520/styles/blob/patch-2/dependent/resuscitation.csl)

Can you elaborate a little on what you had to change in the style? Elsevier is trying to standardize the use of citation formats across their journals, so it would be good if we can verify that this journal is different.

https://www.elsevier.com/journals/resuscitation/0300-9572/guide-for-authors would be a better "documentation" link, by the way.

Phil54520:
I don’t want to delete unless being sure to know what I am doing exactly.

The actual proposed style for the Resuscitation journal rely on “elsevier-vancouver” template, which is incorrect. The actual style is like : Vancouver (superscript in text, only year in date, no issue numbers)
So maybe it is easier just to refer to this style instead of rewriting the whole .csl file?
I definitely agree for the documentation link maybe we can change it.

So basically I would just need to refer to the actual vancouver (superscript in text, only year in date, no issue numbers) style?

What do you think?

Phil

rmzelle:
Yes, if there is an existing CSL style that has the desired citation format, definitely keep resuscitation.csl as a dependent style and just update the "independent-parent" link to that style (and remove the XML comment so we don't overwrite the style next time we automatically generate the Elsevier styles).

adam3smith:
You'd have to delete the new file you created -- but if this is confusing, you can just leave it here&we'll take care of it, might be a lot of effort for two lines of code changes.

Phil54520:
This is just really hard to understand how it works!!
I am not a newbie in coding and everything, but my gosh the way things are organised… for me it is just a big mess with no easy way to change the style.
I’m sorry but I really don’t know how to simply change things. I just want to change the reference (style based). And the robot always give me an error that I cannot create a new file.
Have a look if you can

Sorry

Phil

adam3smith:
I think the only thing left to do is to click the trash can (delete) icon here:
https://github.com/Phil54520/styles/blob/patch-2/resuscitation.csl
and you're done.

Phil54520:
Ok done, If you say so.
Will see the next update once it has been taken into account

adam3smith:
Thanks. Added in a separate commit to clean up the commit history.

PR #1963, opened 2016-03-19
Create IEEE with DOI

csps-uam:
Add DOI to journal, conference and conference proceedings if available.


rmzelle:
Continued in #1964.

@csps-uam, we only accept styles that are licensed under the Creative Commons BY-SA 3.0 license into the repository.

csps-uam:
Then I would ask you to modify your current IEEE citation styles in order to include DOI. ACM Papers allow it nowadays and it is quite useful. Thanks.

rmzelle:
> ACM Papers allow it nowadays

Do you have a link to this style guide so we can take a look?

csps-uam:
Hi, sure!,

you can check the Word template for papers in http://www.acm.org/publications/article-templates/proceedings-template.html/

Or you can check this pdf: https://drive.google.com/file/d/0B54enefzSxTgdlFta2dGMGFDRHc/view?usp=sharing

All references have the "DOI=" entry.

rmzelle:
Thanks. We have a dedicated style for ACM Proceedings, though (which also currently doesn't show the DOI), so it might be better to update that style: https://github.com/citation-style-language/styles/blob/master/acm-sig-proceedings.csl

@adam3smith, thoughts?

csps-uam:
I currently use this custom CSL for my IEEE and ACM papers submission, I just wanted to share it with other people. You are the ones who really know what are doing, so I leave the decision to you guys. Thanks for your awesome work.

Yet, I think offering this style as an additional option (just as IEEE with URL) won't harm and may be helpful to someone.

Regards.

PR #1964, opened 2016-03-19
Create IEEE with DOI

csps-uam:
Add DOI information, if available, to IEEE citation style (in reference list, not inline).

### DIFF
--- a/ieee-with-doi.csl
+++ b/ieee-with-doi.csl
@@ -33,8 +33,9 @@
     <category field="engineering"/>
     <category field="generic-base"/>
     <updated>2016-09-03T18:34:20+00:00</updated>
-    <rights license="http://gplv3.fsf.org/">GPL v3</rights>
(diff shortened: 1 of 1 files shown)


adam3smith:
what is this for? I don't see this in any of the IEEE style guides.

csps-uam:
DOI allows quick reference finding an access.

adam3smith:
I know why DOIs are useful in general and we do have styles that always print them (the Elsevier styles, APA, Chicago Manual for journal articles, etc.). Our [criteria for inclucion](https://github.com/citation-style-language/styles/wiki/Criteria-for-Accepting-Styles) are basically that a style is either used in a specific journal or published in a manual for wide use somewhere. I don't think that's the case here?

rmzelle:
@adam3smith, see #1963 for more discussion.

adam3smith:
Closing based on the discussion in #1963

PR #1965, opened 2016-03-20
Update ceska-zemedelska-univerzita-v-praze-fakulta-agrobiologie-potra…

matejovo:
"et al." if 3 and more authors. It was set to 6


rmzelle:
Thanks!

PR #1970, opened 2016-03-21
Append space to citation number suffix [ieee/ieee-with-url styles]

tribayf:
Hi,

There seems to be a space missing after the citation number in the `ieee ` and `ieee-with-url` styles. Judging by this IEEE [style guide](http://www.ieee.org/documents/ieeecitationref.pdf) it looks to be about the width of an em space, so I've appended `&#8195;` to the end of the citation number.

(Incidentally, I'm working with [jekyll-scholar](https://github.com/inukshuk/jekyll-scholar) and `&emsp;` was being translated to `&amp;emsp;`. I'm not sure what the root of this is, but the decimal encoding works fine.)

### DIFF
--- a/ieee-with-url.csl
+++ b/ieee-with-url.csl
@@ -220,7 +220,7 @@
   <bibliography entry-spacing="0" second-field-align="flush">
     <layout>
       <!-- Citation Number -->
-      <text variable="citation-number" prefix="[" suffix="]"/>
+      <text variable="citation-number" prefix="[" suffix="]&#8195;"/>
       <!-- Author(s) -->
       <text macro="author" suffix=", "/>
       <!-- Rest of Citation -->
--- a/ieee.csl
+++ b/ieee.csl
@@ -225,7 +225,7 @@
   <bibliography entry-spacing="0" second-field-align="flush">
     <layout suffix=".">
       <!-- Citation Number -->
-      <text variable="citation-number" prefix="[" suffix="]"/>
+      <text variable="citation-number" prefix="[" suffix="]&#8195;"/>
       <!-- Author(s) -->
       <text macro="author" suffix=", "/>
       <!-- Rest of Citation -->


adam3smith:
I'm not very inclined to take this, but open for disucssion. The thing is that `second-field-align="flush"` is supposed to take care of this (and does on citeproc-js based products). I think pandoc based products (which I believe jekyll scholar is -- is that Martin Fenner's implementation?) don't handle that command. 

The problem with making your change to accomodate pandoc is that I'm worried it might mess with the tabs, especially for higher numbers, in Word et al. Other thing I'm woried about is using em-space. Are we comfortable that that exists in practically any font?

tribayf:
Jekyll-scholar is @inukshuk, and uses [CiteProc-Ruby](https://github.com/inukshuk/citeproc-ruby). citeproc-js/ruby and CSL are all new to me, so I didn't realize the styles were dynamically configurable. Not sure what I can access through Jekyll, so I'll have to look into it more.

As a sensible default, however, what about using a plain old space?

adam3smith:
Yeah, let's switch this to regular space and I'll take it. (But just to be clear: this really shouldn't be necessary -- second-field-align="flush" should insert a tab after the citation number and then left align the rest of the reference. But we do the suffix after the citation number in a bunch of styles that have this implemented and I don't think have ever gotten any complaints, so I'm fine with adding it.)

inukshuk:
@tribayf did not follow this discussion, but FWIW, in jekyll-scholar we don't generally apply the bibliography styles, because the bibliography template can be fully customized (i.e., whether it is an ol, ul, something else etc.); it should be easy to adjust the template by customizing the bibliography template or adding CSS rules -- if this is something that can be achieved this way and should be added by the processor, let me know.

adam3smith:
So, after checking with @rmzelle, we've agreed to not accept this. Just to clarify, the style as is, with `second-field-align="flush"` should look like this:
untitled
(which is what e.g. Zotero produces).
this is different from just a space after the citation number, which gives you:
untitled
 I don't know enough about jekyll-scholar to suggest a good way forward, but either by customizing your CSS or by talking to @inukshuk this should be handled there. Essentially, IEEE bibliographies are `<ol>`s, so this should be easy enough to do.

tribayf:
Yeah, no problem! It does sound like maybe a styling issue with jekyll-scholar. Thanks for looking into this.

lucasdavid:
In case someone have this same problem in the future (and is not very good at Jekyll-scholar, like me), this is how I solved this problem:

1. I downloaded the [ieee.csl](https://github.com/citation-style-language/styles/blob/master/ieee.csl) style file into the assets folder and remove the line that declared the citation numbers:
    ```csl
    <text variable="citation-number" prefix="[" suffix="]"/>
    ```
2. I updated `_config.yml` file to point to this style file:
    ```yml
    scholar:
      style: assets/ieee
      ...
    ```
3. I updated the style.scss file to add the removed citation numbering:
    ```scss
    ol.bibliography {
      list-style: none;
      padding-left: 40px;
    }
    
    ol.bibliography li {}
    ol.bibliography li + li {
      // margin-bottom: .5em;  // space between citations
    }
    ol.bibliography { counter-reset: item }
    ol.bibliography li:before { 
        content: "[" counter(item) "] "; 
        counter-increment: item; 
        width:3em;
        position:absolute;
        text-indent:-2em;
    }
    ```

PR #1971, opened 2016-03-21
Created morphomuseum.csl

renaud-lebrun:
Citation style proposition for MorphoMuseuM journal (http://morphomuseum.com)


morphomuseum.csl
```diff
@@ -0,0 +1,256 @@
...
+    </choose>
+  </macro>
+  <macro name="website-title">
+    <choose>
+      <if type="webpage" match="any">
+        <text variable="title"/>
+      </if>
+    </choose>
+  </macro>
+  <macro name="doi">
```
adam3smith: you're not using this macro. Delete it.

PR #1972, opened 2016-03-22
Create morphomuseum.csl

renaud-lebrun:
Updated csl file for journal "MorphoMuseuM" (http://morphomuseum.com)

PR #1973, opened 2016-03-22
Create palaeovertebrata.csl

renaud-lebrun:
Style for "Palaeovertebrata" journal. http://palaeovertebrata.com

### DIFF
--- a/palaeovertebrata.csl
+++ b/palaeovertebrata.csl
@@ -1,0 +1,254 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" default-locale="en-US">
+  <info>
+    <title>Palaeovertebrata</title>
+    <id>http://www.zotero.org/styles/palaeovertebrata</id>
+    <link href="http://www.zotero.org/styles/palaeovertebrata" rel="self"/>
+    <link href="http://www.zotero.org/styles/elsevier-harvard" rel="template"/>
+    <link href="http://palaeovertebrata.com/instructionsToAuthors" rel="documentation"/>
+    <category citation-format="author-date"/>
+	 <author>
+      <name>Renaud Lebrun</name>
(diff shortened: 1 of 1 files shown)


palaeovertebrata.csl
```diff
@@ -0,0 +1,254 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" default-locale="en-US">
+  <info>
+    <title>Palaeovertebrata</title>
+    <id>http://www.zotero.org/styles/palaeovertebrata</id>
+    <link href="http://www.zotero.org/styles/palaeovertebrata" rel="self"/>
+    <link href="http://www.zotero.org/styles/elsevier-harvard" rel="template"/>
+    <link href="http://palaeovertebrata.com/instructionsToAuthors" rel="documentation"/>
+    <category citation-format="author-date"/>
```
rmzelle: Idem.

palaeovertebrata.csl
```diff
@@ -0,0 +1,254 @@
...
+	 <author>
+      <name>Renaud Lebrun</name>
+      <email>renaud.lebrun@umontpellier.fr</email>
+      <uri>http://www.mendeley.com/profiles/renaud-lebrun/</uri>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="biology"/>
+	<issn>0031-0247</issn>
+     <eissn>2274-0333</eissn>
```
rmzelle: Running this style through http://formatter.citationstyles.org/ should again correct the indentation here.

rmzelle:
Do MorphoMuseuM (#1972) and this journal share the exact same citation style? If so, it's better to create a "dependent" style for one of them, with the other acting as its "parent". See http://docs.citationstyles.org/en/stable/specification.html#dependent-styles and https://github.com/citation-style-language/styles/blob/master/dependent/nature-biotechnology.csl for an example). That way we don't need to define the same format in two different styles.

rmzelle:
@renaud-lebrun, I added this style in #2267, with the changes I requested.

PR #1974, opened 2016-03-22
Update instituto-de-pesquisas-tecnologicas.csl

samuelbarsanelli:
Just fixing some bugs I've found on the way...


rmzelle:
Thanks! Could you try running your style through http://formatter.citationstyles.org/, please, and update the pull request? Also, we always recommend starting from the latest version from the repository. I reformatted your style in December, but you're reverting those changes here.

samuelbarsanelli:
Done! Sorry about that, it's the first time I make an update, I'm still not familiarized with GitHub tools.

Samuel

Date: Tue, 22 Mar 2016 06:52:38 -0700

rmzelle:
Great, thanks! (we try to keep all styles in the repository consistently formatted, which makes it easier to read them and compare differences between versions)

PR #1975, opened 2016-03-23
Updated et al numbers to match specs

rmhowie:
Specs: http://onlinelibrary.wiley.com/journal/10.1111/(ISSN)1945-5100/homepage/ForAuthors.html

### DIFF
--- a/meteoritics-and-planetary-science.csl
+++ b/meteoritics-and-planetary-science.csl
@@ -112,7 +112,7 @@
     <text term="in" text-case="capitalize-first" suffix=" "/>
     <text variable="container-title" font-style="italic"/>
   </macro>
-  <citation et-al-min="2" et-al-use-first="1" et-al-subsequent-min="2" et-al-subsequent-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year">
+  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year">
     <layout prefix="(" suffix=")" delimiter="; ">
       <group delimiter=", ">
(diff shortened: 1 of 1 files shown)


rmhowie:
This change also modifies the bibliography, but I only wanted to change the et al. settings for in text citations. 

This is probably a Zotero issue, maybe they don't support different styles for the in text and bibliography

meteoritics-and-planetary-science.csl
```diff
@@ -112,7 +112,7 @@
     <text term="in" text-case="capitalize-first" suffix=" "/>
     <text variable="container-title" font-style="italic"/>
   </macro>
-  <citation et-al-min="10" et-al-use-first="1" et-al-subsequent-min="6" et-al-subsequent-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year">
+  <citation et-al-min="2" et-al-use-first="1" et-al-subsequent-min="2" et-al-subsequent-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year">
```
adam3smith: Please do:
`<citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year">`
I realize they say "two or more" in their instructions, but that's not what they mean (check the journal -- in text citation of two authors are (firstauthor and secondauthor year). The subsequent bit doesn't do anything.

adam3smith:
Works fine for me in Zotero.

rmhowie:
> I realize they say "two or more" in their instructions, but that's not what they mean (check the journal -- in text citation of two authors are (firstauthor and secondauthor year).

Oh yeah, you're right!

adam3smith:
great, thanks!

PR #1976, opened 2016-03-23
Updates styles with latest Elsevier report.

cpina:


rmzelle:
Thanks! I see that you already found the new "journals" repo.

rmzelle:
@cpina, by the way, for future updates, I recommend updating the timestamp in _template.csl. Only styles that really change will be updated (timestamps and XML comments are ignored when comparing old and newly generated styles). At least, that's how it should work.

cpina:
Ok, will do the next time. And yes, I've found the new repo. BTW, I'll create a PR for the journals repo with the new data as well...

adam3smith:
We'd want the deleted styles listed in renamed.json right?

cpina:
@adam3smith : good point. But I'm not sure what to write. I've manually checked these cases:
Asian Nursing Research - non-standard
Asian Pacific Journal of Tropical Biomedicine - non-standard
Chirurgie de la main - disappeared
Ciencias Clínicas - non-standard

So, one journal is not in the list any more. What do we do with discontinued journals?

Other 3 journals appear now as "non-standard". I could ignore them (and do not delete them) but I guess that we prefer a non-existing style than an incorrect one?

rmzelle:
Yeah, I was just thinking if we should be more consistent in how we use renamed-styles.json.

The main reason we created it, was that Dan Stillman was concerned that users would end up with styles that no longer updated, or duplicate styles, when I hand-checked the entire repo a few years ago and renamed a lot of styles for consistency to remove naming errors.

For journals that move away from an existing independent-parent (e.g. journals that leave Elsevier), and discontinued journals, maybe we should just delete them and not add them to renamed-styles.json.

cpina:
Yes, the renamed.json was specially when changing style IDs...

rmzelle:
I was also considering doing some Travis CI checks to make sure none of the retargeted journals in renamed-styles.json exist in the repo, and that all target journals in renamed-styles.json do exist. We could even remove very old entries (e.g. > 3 years) with git blame: http://zsoltfabok.com/blog/2012/02/git-blame-line-history/

adam3smith:
I thought @dstillman was concerned also with the existence of style IDs that update to a non-existing style and potential breakage from that. But I could be misremembering this. Dan, if you have a minute to chime in -- we'd rather just delete these, but if that's a concern we can do something else.

rmzelle:
The client (e.g. Zotero) could just switch over to the parent style if a dependent is no longer available. No real reason for us to include such deleted dependents in renamed-styles.json, I think.

adam3smith:
good point, yes.

cpina:
Hi - anything else that should be fixed before merging? Thank you!

adam3smith:
good from my end -- these are all updates, right? I have a couple of outstanding issues that don't seem to be reflected in this list, where guidelines and journal practices diverge from the Elsevier list (most commonly, by not using abbreviations for journals). What should we do for those cases? Put them on the skip list and fix?

cpina:
Hi - yes, updates based on the latest report from Elsevier.

I've just had an email from the Elsevier department regarding http://www.zotero.org/styles/elsevier-with-titles , I'll follow up by email.

Using/not using styles: I'd prefer to have a list of the problematic Guide for Authors so I can ping the correct department and they can fix it instead of adding exceptions (hard to maintain in long term).

adam3smith:
OK, will do. I think there are a couple on the Mendeley style error reports form/spreadsheet.

dependent/journal-of-investigative-dermatology-symposium-proceedings.csl
```diff
@@ -1,18 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" default-locale="en-GB">
-  <!-- Generated with https://github.com/citation-style-language/utilities/tree/master/generate_dependent_styles/data/npg -->
+<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" default-locale="en-US">
+  <!-- Elsevier, generated from "elsevier" metadata at https://github.com/citation-style-language/journals -->
```
rmzelle: Looks like we need to update the NPG dataset for this one (or just put it on the NPG skip list)

rmzelle:
Looks good to merge. @adam3smith, will you do the honors?

rmzelle:
(Travis is complaining because line https://github.com/citation-style-language/styles/blob/6a3c2b595e6bd5ff5eb2df3793ab705d366397c0/renamed-styles.json#L282 should be deleted after deleting hpb.csl)

rmzelle:
(and the error didn't show up before because I've been changing the Travis tests since this PR was created)

PR #1978, opened 2016-03-25
Add RSpec checks for renamed-styles.json

rmzelle:


spec/repository_spec.rb
```diff
@@ -26,3 +26,17 @@
     expect(EXTRA_FILES).to eq([])
   end
 end
+
+describe "The file \"renamed-styles.json\"" do
+  it "must be valid JSON" do
+    expect(JSON_IS_VALID).to be true
+  end
```
rmzelle: @inukshuk, can I bother you for a minute? I was wondering if you could think of a better way to do this `expect(JSON_IS_VALID).to be true` test? I assume it would involve catching the `JSON::ParserError` exception and testing that, but I wasn't sure how to do that.

rmzelle: (and we also have "publishers.json" in the "journals" repo, and "locales.json" in the "locales" repo, so I like to test for valid JSON there too)

zuphilip: For the `journals` repo you could also try to do something with the `jsonlint` npm package, see for example https://github.com/zuphilip/journals/pull/1

inukshuk: @rmzelle normally you would do something like `expect { JSON.parse(x) }.to_not raise(JSON::ParserError)` but since you're actually parsed the file to populate the renamed style lists there is little point in parsing the file one more time.

I'd suggest to rename `JSON_IS_VALID` to something more descriptive though.

In general, we've been adding quite a number of global constants... this is not good style, but the reason why I started doing this was simply for performance to ensure every style is only parsed once. If you need the renamed style lists only in the repository test, however, I would just set it up using before hooks there.

rmzelle: > If you need the renamed style lists only in the repository test, however, I would just set it up using before hooks there.

Okay, I'll try to do that.

> the reason why I started doing this was simply for performance to ensure every style is only parsed once.

Currently all the styles are stored simultaneously in memory, right? Do you think performance would be any faster if `let(:style_validates) { CSL.validate(path).length.zero? }` would use the loaded style instead of a file path? (see https://github.com/citation-style-language/styles/blob/master/spec/styles_spec.rb)

rmzelle: > For the journals repo you could also try to do something with the jsonlint npm package

Thanks, but I think I'd like to standardize on doing all checks via RSpec across the various repos. And I want to do more checks than just validating the JSON.

inukshuk: Good questions. Validation is certainly a factor, but I'd suggest to disable it temporarily to see what kind of performance hit we're actually talking about. The reason why this is a little complicated is that, yes, the styles are in memory, but that is actually the Ruby representation of each style; we could theoretically gain some speed if we read each XML file just once validate and parse the style right away -- it's not straightforward so it would be good to know exactly how much this would be worth (though it is probably a considerable amount of time).

rmzelle: Well, I get:

* on Windows 7, Ruby 2.1.0, SSD, with https://github.com/adoxa/ansicon installed (for proper color output, still prints thousands of lines):
```
Finished in 2 minutes 44.6 seconds (files took 38.24 seconds to load)
131958 examples, 0 failures
```

* after `SET CONTINUOUS_INTEGRATION=true` (per https://github.com/travis-ci/travis-ci/issues/1337):
```
Finished in 1 minute 12.98 seconds (files took 33.74 seconds to load)
131958 examples, 0 failures
```
(wow! Looks like the standard progress bar slows things down quite a bit; I'll check whether I get any performance improvement locally with `CONTINUOUS_INTEGRATION=true` on OS X later)

* after disabling style validation:
```
Finished in 37.49 seconds (files took 33.07 seconds to load)
123791 examples, 0 failures
```

On my Mac I also usually get something around 1 minute 10 seconds, with validation. So it looks like validation does make a big difference.

inukshuk: Well, validation itself takes up some time, as well, but basically each style is read and parsed (by nokogiri) twice. We'd have to make some changes to CSL-Ruby to allow us to feed nokogiri's parse results directly into the validator and style parser, then we could essentially run both in one go.

rmzelle: > but basically each style is read and parsed (by nokogiri) twice

Could I check how much time is spent on this by timing `CSL::Style.load(path)` (read/parse) vs. `CSL.validate(path)` (read/parse/validate)?

By the way, the profiling output is:

```
Top 10 slowest examples (9.98 seconds, 13.4% of total time):
  zeitschrift-fur-soziologie: must validate against the CSL 1.0.1 schema
     Please check your style at http://validator.citationstyles.org/
    1.68 seconds ./spec/styles_spec.rb:4
  zeitschrift-fur-soziologie: must be parsable as a CSL style
    1.6 seconds ./spec/styles_spec.rb:9
  pontifical-gregorian-university: must be parsable as a CSL style
    1.03 seconds ./spec/styles_spec.rb:9
  pontifical-gregorian-university: must validate against the CSL 1.0.1 schema
     Please check your style at http://validator.citationstyles.org/
    1.02 seconds ./spec/styles_spec.rb:4
  dependent/nigerian-journal-of-pharmaceutical-sciences: may not have <macro/>,
<citation/>, or <bibliography/> elements
    0.79808 seconds ./spec/styles_spec.rb:94
  dependent/remote-sensing-applications-society-and-environment: must have an <i
nfo/> element
    0.79308 seconds ./spec/styles_spec.rb:116
  dependent/the-journal-of-heart-and-lung-transplantation: must have a style ID
    0.78908 seconds ./spec/styles_spec.rb:120
  dependent/journalism-practice: must validate against the CSL 1.0.1 schema
     Please check your style at http://validator.citationstyles.org/
    0.76408 seconds ./spec/styles_spec.rb:4
  dependent/journal-of-hydrometeorology: may not have <macro/>, <citation/>, or
<bibliography/> elements
    0.75808 seconds ./spec/styles_spec.rb:94
  dependent/international-journal-of-surgery-open: must validate against the CSL
 1.0.1 schema
     Please check your style at http://validator.citationstyles.org/
    0.75508 seconds ./spec/styles_spec.rb:4

Top 10 slowest example groups:
  zeitschrift-fur-soziologie:
    0.1939 seconds average (3.3 seconds / 17 examples) ./spec/styles_spec.rb:155

  pontifical-gregorian-university:
    0.12107 seconds average (2.06 seconds / 17 examples) ./spec/styles_spec.rb:1
55
  soziologie:
    0.05136 seconds average (0.87309 seconds / 17 examples) ./spec/styles_spec.r
b:155
  dependent/nigerian-journal-of-pharmaceutical-sciences:
    0.05038 seconds average (0.80608 seconds / 16 examples) ./spec/styles_spec.r
b:162
  dependent/remote-sensing-applications-society-and-environment:
    0.05001 seconds average (0.80008 seconds / 16 examples) ./spec/styles_spec.r
b:162
  dependent/the-journal-of-heart-and-lung-transplantation:
    0.04988 seconds average (0.79808 seconds / 16 examples) ./spec/styles_spec.r
b:162
  dependent/journalism-practice:
    0.04825 seconds average (0.77208 seconds / 16 examples) ./spec/styles_spec.r
b:162
  dependent/journal-of-hydrometeorology:
    0.04782 seconds average (0.76508 seconds / 16 examples) ./spec/styles_spec.r
b:162
  dependent/hormones-and-cancer:
    0.04769 seconds average (0.76308 seconds / 16 examples) ./spec/styles_spec.r
b:162
  dependent/international-journal-of-surgery-open:
    0.04763 seconds average (0.76208 seconds / 16 examples) ./spec/styles_spec.r
b:162
```

(I'm not sure what makes https://github.com/citation-style-language/styles/blob/master/zeitschrift-fur-soziologie.csl so slow; I've seen this before, and the style isn't particularly complex)

rmzelle: Replace `CSL.validate(path)` by `CSL::Style.load(path)` gives

```
Finished in 45.12 seconds (files took 39 seconds to load)
123791 examples, 0 failures
```

so maybe the savings would only be 5-10 seconds here, since I came to 38 seconds without any validation.

rmzelle: Some of the profile times make no sense to me, though, e.g.:

```
dependent/the-journal-of-heart-and-lung-transplantation: must have a style ID
    0.78908 seconds ./spec/styles_spec.rb:120
```

which should be superfast. I assume there some hiccup during some of these examples, like garbage collection?

inukshuk: `Style.load` also parses the CSL in Ruby (that's not necessary for validation) so we'd save more. You would have to replace validation with only reading the file and parsing it with Nokogiri.

rmzelle: > Style.load also parses the CSL in Ruby (that's not necessary for validation) so we'd save more

We'd save less, no? I just showed that `Style.load` is only ca. 7 seconds slower than no validation at all.

inukshuk: You could time the actual test to see if the test takes so long (i.e., time the expect line only)... otherwise it could also have something to do with the way we're building up the examples dynamically.

inukshuk: >  We'd save less, no? I just showed that Style.load is only ca. 7 seconds slower than no validation at all.

Oh, that's right... so the bulk of the time is actual validation anyway.

rmzelle: Test suite performance is actually pretty decent nowadays, though, so this discussion is not a strong call to action.

rmzelle:
It's still not very good RSpec idiom, but it works. The new tests found some problems in renamed-styles.json, which now have been corrected.

spec/repository_spec.rb
rmzelle: Turns out that `let(:style_validates) { CSL.validate(path).length.zero? }` is run for every example that calls `style_validates`, which resulted in style validation being done twice. With `before(:all)` it seems we just saved about 20 seconds: https://github.com/citation-style-language/styles/commit/1e93b01700c7a05155ec68b5cb6e474e11a8f499

We're under 1 minute of test time, now:

```
Finished in 57.81 seconds (files took 45.74 seconds to load)
132028 examples, 0 failures
```

PR #1987, opened 2016-03-29
Update hiob-ludolf-centre-for-ethiopian-studies.csl

eu-genia:
volume information was doubled; removed

thanks again!


eu-genia:
Thanks again in advance for merging!

rmzelle:
Thanks!

PR #1990, opened 2016-03-30
Delete discontinued BMC styles

zuphilip:
See https://github.com/citation-style-language/journals/issues/6


rmzelle:
So we're just missing BMC Proceedings?

Last article was published in 2015, and I can't find the reference style for it: https://bmcproc.biomedcentral.com/submission-guidelines/preparing-your-manuscript#preparing+main+manuscript+text

>How to format your references
For information on Reference style, please see the relevant journal's Instructions for Authors.

(can't find the "Instructions for Authors")

zuphilip:
Yes, that is the only one missing. You are right about the missing instructions. Thus, we can argue (independent whether the journal still continues) that the style is unclear and therefore it is better not to link it to something which might be the wrong one. Thus, we can delete it as well. If in the future the situation will change, then we can still add it easily.

Do you agree with deletion?

rmzelle:
I would just delete it. If it pops back up on the Springer journal metadata we'll just add it again.

rmzelle:
Thanks!

PR #1991, opened 2016-03-30
Added a spacer in between contributers name + date and the actual source

nevrome:


adam3smith:
this should be on a new line in the style. Which processor/reference manager are you using?

nevrome:
I'm using Zotero to manage my literature and to export it as an BibTex-File. For writing and processing the text I use RMarkdown (RStudio>knitr>pandoc?).

adam3smith:
(@nevrome: unrelated, but you should look at using CSL JSON for Zotero/pandoc integration; much less loss during conversion than using bibtex).

@rmzelle -- can you think of any reason for not allowing spaces at the end of display elements or at the end of first fields where 2nd field align is used? While the processors should handle these correctly, it's hard and many don't and I wonder if there's any cost to accommodating these?

nevrome:
@adam3smith Thanks I'll look into this. At the moment my solution seams convenient - but you never know. 

Without my change the citation looked somehow like this:
`Huntley – Prentice 1993B. Huntley – C. Prentice, Holocene vegetation und climates of europe, in: Wright (ed.) 1993, 136–168`
I wanted it to like like this: 
`Huntley – Prentice 1993: B. Huntley – C. Prentice, Holocene vegetation und climates of europe, in: Wright (ed.) 1993, 136–168`

adam3smith:
right, the citation style, as is, should look like this
>Huntley – Prentice 1993: 
B. Huntley – C. Prentice, Holocene vegetation und climates of europe, in: Wright (ed.) 1993, 136–168

pandoc doesn't handle the relevant CSL (display).

nevrome:
Ok - I hope there's a solution beyond my pretty lazy approach?

adam3smith:
@rmzelle ping

rmzelle:
I pinged John MacFarlane at https://github.com/jgm/pandoc-citeproc/issues/85.

I really rather not add any hacks to repository styles to cover the shortcomings of particular CSL processors (especially since there doesn't seem to be any particular reason why `display="block"` hasn't been implemented yet). @nevrome, even if we don't accept this pull request, you can of course still use the customized style yourself.

nevrome:
Yeah - of course. Thanks for the dedication!

rmzelle:
> @rmzelle -- can you think of any reason for not allowing spaces at the end of display elements or at the end of first fields where 2nd field align is used? While the processors should handle these correctly, it's hard and many don't and I wonder if there's any cost to accommodating these?

I'm just not a fan since adding extra spaces adds clutter to the styles. And I'm not sure spaces can't lead to layout problems (e.g. related to justification).

nevrome:
@rmzelle Justification is not the problem in my case. Just for the record.

adam3smith:
Closing this as per the discussion above. FWIW, I think this should work better with pandoc since https://github.com/jgm/pandoc-citeproc/commit/6e00df61ce4ba62d2616196859922c9afe3d837c

PR #1992, opened 2016-03-31
Add dependent/international-archives-of-medicine.csl

zuphilip:
This journal is now published by iMed.pub and follows
Vancouver style. Until the end of 2014 it was published
by BioMed Central

https://github.com/citation-style-language/journals/issues/6


adam3smith:
thanks!

PR #1995, opened 2016-04-02
Update ceska-zemedelska-univerzita-v-praze-fakulta-agrobiologie-potra…

matejovo:

use "et" instead "and" because it use in Czech written works.


adam3smith:
somehow you added the + signs from the diff into the style. 
But also, this style is purposefully language neutral, I thought. If we want it in czech, we can just set a default locale, so I'm not sure I understand.

matejovo:
@adam3smith 

Faculty rules say: 
If you cite English written source in Czech written thesis write: (Berge **et** Wierup, 2012)
If you cite Czech written source in Czech written thesis write: (Novák **a** Dvořák, 2011) 
In case you use citation sw, you can use **et** in both

When I change locale to CS, It will write "a" everywhere and "et al" will be replaced "a kol."  I dont wan it.

What we need is change **and** by **et**

Clear?:)

adam3smith:
ok, sounds good, thanks!

PR #1997, opened 2016-04-05
Health Reform Observer: InText Defaut (Author-Date).csl

scottnb:
Based off of Harvard. Intended use for Health Reform Observer – Observatoire des Réformes de Santé.


rmzelle:
Continued in #2007.

PR #1999, opened 2016-04-05
Recent Patents in Drug Discovery and Delivery

hbien:
Added style for "Recent Patents in Drug Discovery and Delivery" from Elsevier press


hbien:
Re-named file to be consistent with actual journal title. Actual title is "Recent Patents on Drug Delivery & Formulation".

PR #2001, opened 2016-04-05
Canadian Journal of Economics: InText Author-(Date).csl

scottnb:
Based off of Chicago. Intended use to cite CJE.


rmzelle:
Continued in #2005.

PR #2002, opened 2016-04-05
Recent Patents on Drug Delivery & Formulation

hbien:
Fixed submission for "Recent Patents on Drug Delivery & Formulation" by Bentham Science publishers, not Elsevier.


rmzelle:

(@hbien, you can just edit styles in place in your pull request; no need to delete and re-add it)

rmzelle:
And it's almost always better to just fix up the original pull request. No need to create new ones if something had to be corrected.

(replaces #1999)

rmzelle:
Thanks!

PR #2005, opened 2016-04-06
Create canadian-journal-economics-default.csl

scottnb:
link: 
http://csl.mendeley.com/styles/371811861/canadian-journal-economics-default


scottnb:
Hi Rintze Zelle,

Thank you, I'll make the changes on same pull request.

I am new to adding to github. I looked at style requirements, reformatted
with CSL formatter and it passed the "verification" but I'm not sure of the
issues for #2005. Originally Chicago style but I made specific to Canadian
Journal of Economics. Should I leave a comment in the entry for assistance?
​

rmzelle:
The first issue to fix is that the style needs to be moved out of the "dependent" subdirectory (dependent styles are those that rely on another style for their citation format, like https://github.com/citation-style-language/styles/blob/master/dependent/nature-biotechnology.csl ).

See https://help.github.com/articles/moving-a-file-to-a-new-location/ (and start by clicking the edit button at https://github.com/scottnb/styles/blob/patch-6/dependent/canadian-journal-economics-default.csl ).

rmzelle:
Thanks. While it's good that you validated your style, for the CSL repository we have additional requirements as discussed in https://github.com/citation-style-language/styles/wiki/Style-Requirements. In particular, @csl-bot is still unhappy because (as highlighted in the test report):

* repository styles may not contain any unused macros (in your case, "editor" and "translator")
* repository styles must have a style ID and "self" link of the form "http://www.zotero.org/styles/canadian-journal-economics-default" instead of "http://csl.mendeley.com/styles/canadian-journal-economics-default"

rmzelle:
Does this journal have more than one citation format? Usually we just name journal styles after the journal title (e.g. "Canadian Journal of Economics" and "canadian-journal-of-economics.csl").

Will this style be used for both English and French submissions?

scottnb:
The journal has French submissions, but for our purposes we have been using
in English. Maybe I should specify in the code?

There is only one reference format, but since people have been using
in-text citations differently I started saving two separate links to make
it easier.

zuphilip:
I read in the [style guide](http://economics.ca/cje/CJE-Style-Guide-2014.pdf): "Style manual:	 Chicago Manual of	Style; Canadian spelling (Canadian Oxford )."

Thus, I already expect that one of the existing styles https://www.zotero.org/styles/?q=chicago with the language selection should match. No?

rmzelle:
> The journal has French submissions, but for our purposes we have been using
in English. Maybe I should specify in the code?

If you want, you can lock the style to Canadian French (locale code "fr-CA"; see https://github.com/citation-style-language/styles/wiki/Style-Requirements#3---style-locale ). Or you can keep the style as is, in which case users can use it in both English and Canadian French (but they would have to select the correct locale by hand when they use the style).

> There is only one reference format, but since people have been using
in-text citations differently I started saving two separate links to make
it easier.

We prefer to host just one CSL style per journal though, unless the journal specifically prescribes different citation formats.

And as @zuphilip remarked, it would indeed be preferable to turn this into a dependent style that depends on one of the existing Chicago styles. Whether that's feasible depends on whether you get the desired output by using on of those styles in Canadian French.

scottnb:
​Hi both, thanks for the input.

I think best to leave as is in English, then maybe later a dependent style
can be made from Chicago in Canadian French.

Okay, I have a few other journal styles I wanted to put out if this one
goes through successfully but I can just submit the "default" style and use
the "alternates" privately.

rmzelle:
Can you give an example of how the default differs from the alternates?

scottnb:
In my test report, is it picking up that "editor" and "translator" macro are missing something? (i.e. unused). Can I delete "translator" code if not likely to be relevant? I may have made changes under "editor" macro as needed & maybe I created an error from the CSL editor.

scottnb:
Default & Alternates differ by in-text citation only so more convenient while writing.
Default: This idea is supported by X (Smith and Scott 2008).
Alternate: As Smith and Scott (2008) report... 

It is a minor difference and I understand why only one version per journal is preferred.

rmzelle:
Neither macro is currently used in the style. If you don't need them, they should be deleted.

rmzelle:
And regarding your example, programs like Zotero and Mendeley allow you to "suppress" the authors, so you end up with "(2008)". You then have to write out the "Smith and Scott" part out by hand as part of the regular manuscript text. Since it's common to mix both formats in a single document, we wouldn't want to have separate styles for them.

scottnb:
Yes, if mixing styles then better to have default submission (author-date) and users can suppress authors as they need.

canadian-journal-economics-default.csl
```diff
@@ -0,0 +1,611 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" demote-non-dropping-particle="never" version="1.0" class="in-text" page-range-format="chicago">
+  <info>
+    <title>Canadian Journal of Economics: InText Default (Author-Date)</title>
```
rmzelle: Can you change this to `<title>Canadian Journal of Economics</title>`?

canadian-journal-economics-default.csl
```diff
@@ -0,0 +1,611 @@
...
+    <id>http://www.zotero.org/styles/canadian-journal-economics-default</id>
```
rmzelle: Can you change this to `<id>http://www.zotero.org/styles/canadian-journal-of-economics</id>` (the style file name and "self" link need to be updated to match as well)

canadian-journal-economics-default.csl
```diff
@@ -0,0 +1,611 @@
...
+    <link rel="self" href="http://www.zotero.org/styles/canadian-journal-economics-default"/>
+    <link rel="documentation" href="http://www.chicagomanualofstyle.org/tools_citationguide.html"/>
```
rmzelle: Can you update this "documentation" link? http://economics.ca/cje/en/style.php is probably the best link.

And can you add a "template" link so we know which CSL style you based yours on? See https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link

canadian-journal-economics-default.csl
```diff
@@ -0,0 +1,611 @@
...
+      <uri>http://arc.nucapt.northwestern.edu/Richard_Karnesky</uri>
+    </contributor>
+    <contributor>
+      <name>Naomi Scott</name>
+      <uri>http://www.mendeley.com/profiles/naomi-scott/</uri>
+    </contributor>
+    <contributor>
+      <name>Andrew Dunning</name>
+      <email>andrew.dunning@utoronto.ca</email>
+    </contributor>
```
rmzelle: Please remove all authors and contributors from the original style.

canadian-journal-economics-default.csl
```diff
@@ -0,0 +1,611 @@
...
+    <category citation-format="author-date"/>
+    <category field="generic-base"/>
```
rmzelle: Please change the field to something more appropriate. The options (http://docs.citationstyles.org/en/stable/specification.html#appendix-i-categories) are rather limited, but "social_science" is better than "generic-base")

scottnb:
I will make all the additional changes.

Just one more thought. Can I make the default as "author-(date)" on this submission? Might be easier to move parentheses when needed rather than typing authors- at least for users primarily writing in this format.

rmzelle:
> Can I make the default as "author-(date)" on this submission?

No. We use the "(author-date)" format for all CSL repository styles, and I'd like to keep things consistent.

We might make it easier it easier in the future to properly define both types.

scottnb:
Fair enough. I'll keep the other style for personal purposes. I just
submitted 2 other styles but kept in line with (author-date) formatting.

rmzelle:
The file name/ID should be "canadian-journal-of-economics.csl" (with "-of-").

(also lowercase "Chicago" in `<link href="http://www.zotero.org/styles/Chicago" rel="template"/>`, please)

rmzelle:
And you're more than welcome to include your own name as author/contributor to this style. I just meant removing the original authors from chicago.csl.

rmzelle:
Looks good! We should be almost there.

So, just to confirm: did you have to make changes to the body of the style to make it match the journal requirements, or was the "chicago-author-date" format already correct?

scottnb:
Excellent, thanks for all your help with this!

I made a few adjustments to the style using the CSL visual and code editor
from Mendeley.

rmzelle:
@adam3smith, can you take a brief look? Looks good to me.

rmzelle:
Thanks!

rmzelle:
@scottnb, could you help us figure out the correct et-al abbreviation settings for this style? I couldn't find this information in style documentation. Beyond what number of authors should et-al abbreviation happen, and how many authors should be shown before "et al."? (for in-text citations, and for the references in the bibliography, which may have different requirements)

The style currently has a lone "et-al-use-first" attribute at https://github.com/citation-style-language/styles/blob/08fbfbf97b751b1d1e55bd846a7fafabb74769bb/canadian-journal-of-economics.csl#L137, while https://github.com/citation-style-language/styles/blob/08fbfbf97b751b1d1e55bd846a7fafabb74769bb/canadian-journal-of-economics.csl#L549 has `et-al-use-first="19" et-al-min="4"`, and https://github.com/citation-style-language/styles/blob/08fbfbf97b751b1d1e55bd846a7fafabb74769bb/canadian-journal-of-economics.csl#L570 has `et-al-use-first="19" et-al-min="19"`.

Normally, "et-al-min" and "et-al-use-first" are defined together on an element, while the value of "et-al-min" must always be higher than the value of "et-al-use-first".

scottnb:
Hi Rintze,

For in-text there should *not *be use of et al. so really there is no
"et-al-min" and the "et-al-display-first" should display all authors. The
guide says if 3 or more authors are repeated, then use initials. E.g.
(Smith, Scott and Doe 2012; SSD), then future citations by these authors
are listed as "SSD". However, I think this feature has to be done manually.

Reference list should have all authors.

I'll check around tomorrow to see if more recent articles still follow this
style.

rmzelle:
Okay, thanks. So it looks like we can just delete all "et-al-..." attributes, then. I'll wait with making that change until you report back.

(We started running some more automated checks on our repository styles, which highlighted the strange settings in this CSL style)

rmzelle:
And you're right that CSL doesn't support an automatic condensing of "Smith, Scott and Doe" to "SSD".

scottnb:
No problem! It looks like many recent articles do use et al. but the guide
currently posted on the CJE website is consistent with my last comment. So
I think safe to go ahead with that.

Thank you,
Naomi

rmzelle:
Great, thanks for checking. I removed the et-al abbreviation from the style: https://github.com/citation-style-language/styles/commit/5e38d2aabba638edabfb9603388ea3d425d63fd1

PR #2009, opened 2016-04-07

roaldfrosig:
As the danish locale file has been updated, I have added an et-al term in the locale element, to keep our style unchanged.


adam3smith:
thanks!

PR #2010, opened 2016-04-07

roaldfrosig:
The danish locale file has recently been updated, so I'm adding the et-al term to the locale element, in order to maintain our style as it is.


adam3smith:
thanks!

PR #2011, opened 2016-04-07
Update coral-reefs.csl

RonaldP3:
Journal abbreviations should be without periods. Same is true for conference proceedings.


adam3smith:
thanks!

PR #2020, opened 2016-04-19
Update and rename vancouver.csl to materials-research-letters.csl

zeroicma:
Just change the style title to facilitate users finding the exact citation style of the specific journal

### DIFF
--- a/materials-research-letters.csl
+++ b/materials-research-letters.csl
@@ -1,10 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="minimal">
   <info>
-    <title>Vancouver</title>
-    <id>http://www.zotero.org/styles/vancouver</id>
-    <link href="http://www.zotero.org/styles/vancouver" rel="self"/>
-    <link href="http://www.nlm.nih.gov/bsd/uniform_requirements.html" rel="documentation"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
(this should be dependent of taylor and francis NLM -- let's check if we have this in the data, otherwise add manually).
@zeroicma -- thank. this isn't quite the right way to go about doing this (you'd actually overwrite the Vancouver style with what you're proposing) but thanks for pointing this out.

zeroicma:
Karcher

Thanks for pointing out. This is the first time I work on citation editing.
I may submit another indenpent one soon

Thanks again for your help!

On Tuesday, April 19, 2016, Sebastian Karcher <notifications@github.com>
wrote:

> (this should be dependent of taylor and francis NLM -- let's check if we
> have this in the data, otherwise add manually).
> @zeroicma <https://github.com/zeroicma> -- thank. this isn't quite the
> right way to go about doing this (you'd actually overwrite the Vancouver
> style with what you're proposing) but thanks for pointing this out.
>
> —
> <https://github.com/citation-style-language/styles/pull/2020#issuecomment-212143730>
>


-- 
Xiaolong

Ph.D Candidate
Department of Materials Science and Engineering

M.S. Candidate
Department of Statistics

North Carolina State University
Raleigh, NC 27695-7919

  Tel: 919-809-4579
Web: www4.ncsu.edu/~xma4

*"In case I won't see you, good afternoon, good evening and good night."*

adam3smith:
Supserseded by #2022

PR #2022, opened 2016-04-19
Create materials-research-letters.csl

zeroicma:
Help users directly find the citation style for the specific journal


rmzelle:
@adam3smith, as you mentioned in #2020, this should just be an automatically generated dependent of taylor-and-francis-national-library-of-medicine.csl.

It's already in the T&F metadata file in the "journals" repo, but apparently that's not in sync with the "styles" repo. @adam3smith, do you know where we stand with T&F metadata? Would it make sense to ask for a fresh set to update the "journals" and "styles" repos?

rmzelle:
Added this journal by hand: https://github.com/citation-style-language/styles/commit/22a8adc436472ed8df255e6723764f31dc38f3a3

Sorry for the long wait.

PR #2023, opened 2016-04-20
fixed spacing, simplified user input for theses

scottcove:
Simplified the user input for the genre value in thesis type by allowing the user to add only the level, and not requiring them to add the value "dissertation" in the (primarily Zotero) "type" field.  Also fixed spacing between publisher and publisher-place macros.


rmzelle:
Thanks!

PR #2024, opened 2016-04-21
Change et-al settings for elsevier-with-titles

rmzelle:
@cpina, @adam3smith, can you double-check?


adam3smith:
Sorry for not doing this when I promised I would and thanks @rmzelle for taking care of it. Should also include same change for http://www.zotero.org/styles/elsevier-with-titles-alphabetical (the w/o titles style is already without et al)

cpina:
Looks good to me!

PR #2028, opened 2016-04-23
Create universite-laval-departement-des-sciences-historiques.csl

CharlesQc:
The file is not based on a particular style, as it was coded from scratch. Name is chosen to fit other Universite Laval's files in the repository., but differs significantly in some ways. It is french.

P.S. First post - feel free to suggest upgrades


rmzelle:
Great, thanks!

PR #2033, opened 2016-04-23
Update polish-legal.csl

Jack1982:
Changes to how websites are cited, consistent with current trends.


adam3smith:
The changes look fine, but could you fix the indenting, please? You can either manually indent two spaces per level or, more conveniently, using 

See instructions above for updating your pull request.

rmzelle:
@Jack1982, if you accept this pull request we should be all set: https://github.com/Jack1982/styles/pull/1

rmzelle:
Thanks!

PR #2035, opened 2016-04-26
Fix name formatting in zoological-journal-of-the-linnean-society

rmzelle:


rmzelle:
A Mendeley user reported the following things that were wrong:

> According the journal: (Carlo, Barbeitos, & Lasker, 2011); (Hilbish, 1996).
> According Mendeley: (J.M. Carlo, M.S. Barbeitos, & H.R. Lasker, 2011); (T.J. Hilbish, 1996).

> In addition, according to Zoological Journal of the Linnean Society:  When papers are by three authors, use all names on the first mention and there after abbreviate to the first name et al. For papers by four or more authors, use et al. throughout.

I could confirm that the style never seems to use initials in in-text citations, but the et-al settings already look correct for the specified behavior.

adam3smith:
ah yes, I missed the second part of this, i.e. the initials.

rmzelle:
Missed? You never worked on this style, no?

adam3smith:
missed when I saw the report through the Mendeley error report form

rmzelle:
Ah, also got an email through the contact form. Know if Mendeley is unreliable regarding et-al-subsequent-* settings?

rmzelle:
(I recon not since it should be entirely within the scope of citeproc-js)

adam3smith:
not that I know of. Are you sure the sure the user understood the journal requirements? If it's the same person, didn't seem to be the case as reported on the form.

rmzelle:
I'll double-check with her (maria del mar soler hurtado).

adam3smith:
yeah, same person. Complained that (A. Hugall, J. Stanton, & C. Moritz, 1999) wasn't et al.

rmzelle:
Yeah, if that's the first time it's cited it's right.

rmzelle:
(haven't heard back from her regarding the et-al. issue)

PR #2036, opened 2016-04-26

CharlesQc:
This is a big update, which includes many changes. 

It goes like this:
1. Much cleaner code. I revised the syntax, mostly because there was still some mistakes with random double commas and periods. Use of groups helped a lot in that way. 
2. Both "citation" and "bibliography" layouts were reconstructed completely, following a MUCH simpler pattern: macro-macro-condition(for exceptions)-condition(for articles vs books, mostly)-macro. I isolated smaller logical conditions instead of ONE BIG condition. 
3. The "end-part" macro was also removed, since macros for publisher, year and pages were made adequate (and also because citation and bibliography are so different, "end-part" was basically the source of some problems).

As for smaller changes:
4. Removed the "---" substitution, as stated by the documentation. Authors will be repeated for each of their works in bibliography.
5. Removed "paper-conference" as condition for URL display.
6. Added " ; " between each cite in citation.
7. Added "Ibid." and "op. cit." conditions for citation, as stated by the documentation.
8. Added macros "pages-bibliography" and "pages-citation", as they are not used in the same way.

This version has been tested with 120 different entries now, and seems to be fully functional. Although, two technical problems remain:
1. Documentation asks for an explicit differentiation of "Doctorate thesis" and "Master thesis" ("Thèse de doctorat" and "Mémoire de maîtrise"). For now, I have not yet found a way to implement it, so only "Thèse" appears.
2. Even though "note" is stated as "class", Word does not show citation as footnote (ex: like '1' appears next to quote, and refers to the citation at the bottom of the page, preceded by the same '1'). We still have to manually "Add a footnote", and then select the correct entry (any clue would be helpful for that).

### DIFF
--- a/universite-laval-departement-des-sciences-historiques.csl
+++ b/universite-laval-departement-des-sciences-historiques.csl
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style class="note" version="1.0" initialize="false" initialize-with-hyphen="false" default-locale="fr-CA" xmlns="http://purl.org/net/xbiblio/csl">
+<!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
     <title>Université Laval - Département des sciences historiques (French - Canada)</title>
     <id>http://www.zotero.org/styles/universite-laval-departement-des-sciences-historiques</id>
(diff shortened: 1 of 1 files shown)


libora:
Close this Pull Request, please. For next time, you can edit your pull request directly and do not need to create a new one.

PR #2037, opened 2016-04-26

CharlesQc:
This is a big update, which includes many changes. 

It goes like this:
1. Much cleaner code. I revised the syntax, mostly because there was still some mistakes with random double commas and periods. Use of groups helped a lot in that way. 
2. Both "citation" and "bibliography" layout were reconstructed completely, following a MUCH simpler pattern: macro-macro-condition(for exceptions)-condition(for articles vs books, mostly)-macro. I isolated smaller logical conditions instead of ONE BIG condition. 
3. The "end-part" macro was also removed, since macros for publisher, year and pages were made adequate (and also because citation and bibliography are so different, "end-part" was basically the source of some problems).

As for smaller changes:
4. Removed the "---" substitution, as stated by the documentation. Authors will be repeated for each of their works in bibliography.
5. Removed "paper-conference" as condition for URL display.
6. Added " ; " between each cite in citation.
7. Added "Ibid." and "op. cit." conditions for citation, as stated by the documentation.
8. Added macros "pages-bibliography" and "pages-citation", as they are not used in the same way.

This version has been tested with 120 different entries now, and seems to be fully functional. Although, two technical problems remain:
1. Documentation asks for an explicit differentiation of "Doctorate thesis" and "Master thesis" ("Thèse de doctorat" and "Mémoire de maîtrise"). For now, I have not yet found a way to implement it, so only "Thèse" appears.
2. Even though "note" is stated as "class", Word does not show citation as footnote (ex: like '1' appears next to quote, and refers to the citation at the bottom of the page, preceded by the same '1'). We still have to manually "Add a footnote", and then select the correct entry (any clue would be helpful for that).

### DIFF
--- a/universite-laval-departement-des-sciences-historiques.csl
+++ b/universite-laval-departement-des-sciences-historiques.csl
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style class="note" version="1.0" initialize="false" initialize-with-hyphen="false" default-locale="fr-CA" xmlns="http://purl.org/net/xbiblio/csl">
+<!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
     <title>Université Laval - Département des sciences historiques (French - Canada)</title>
     <id>http://www.zotero.org/styles/universite-laval-departement-des-sciences-historiques</id>
@@ -12,7 +13,7 @@
     <category citation-format="note"/>
     <category field="history"/>
     <summary>Le style bibliographique pour citation et références du Département des sciences historiques de l'Université Laval    </summary>
-    <updated>2016-04-22T00:27:37+00:00</updated>
+    <updated>2016-04-26T05:41:58+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="author-bibliography">
@@ -31,7 +32,7 @@
     </names>
   </macro>
   <macro name="editor-bibliography">
-    <names variable="editor" suffix=". ">
+    <names variable="editor">
       <name and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="3" et-al-use-first="1" initialize="false" initialize-with="." name-as-sort-order="first">
         <name-part name="family" text-case="uppercase"/>
         <name-part name="given" text-case="capitalize-first"/>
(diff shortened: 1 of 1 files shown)


universite-laval-departement-des-sciences-historiques.csl
```diff
@@ -155,87 +158,127 @@
...
-          <text macro="edition" suffix="."/>
-          <text macro="end-part"/>
-          <text macro="collection"/>
-        </else>
-      </choose>
-      <text macro="access" suffix=","/>
-    </layout>
   </bibliography>
 </style>
+e
```
libora: Deleta last line with "e"

libora:
I do not see problem with doctoral and MSc theses. On "http://www.hst.ulaval.ca/services-et-ressources/guides-pedagogiques/regles-de-presentation-bibliographique/" the use the same rules for bibliography:
**Mémoire ou thèse**
 NOM, Prénom. « Titre ». Mémoire de maîtrise [ou Thèse de doctorat, ou autre identifiant], ville, nom de l'université, année, nombre total de pages.
 ROBERT, Jean-Claude. « Montréal 1821-1871, aspects de l'urbanisation ». Thèse de doctorat, Paris, École des hautes études en sciences sociales, 1977. iv, 491 p.

CharlesQc:
Thanks for the comment!

I am really sorry about the consecutive pull requests: I did not know I could edit the first one. I will edit this one, as suggested.

About the doctoral & master theses, the example used explains the difference, and therefore my problem (it could be something very simple that I do not know... I am still not an expert at CSL!):
We need to have something that will differentiate whether the work is a "Thèse de doctorat" OR a "Mémoire de maîtrise" in the bibliography, but I cannot find how to code something like this, as the only "document type" is the generic "thesis". 

For example: 
ROBERT, Jean-Claude. « Montréal 1821-1871, aspects de l'urbanisation ». **Thèse de doctorat**, Paris, École des hautes études en sciences sociales, 1977. iv, 491 p.

DAVID, Éric. « Le clergé catholique canadien-français et la théorie de l’évolution (1860-1950) ». **Mémoire de maîtrise**, Montréal, Université du Québec à Montréal, 2003, 99 p.

I will keep working on it!

CharlesQc:
Problem 2 (about footnotes) is solved. I just found out that it was [Mendeley-related issue](https://feedback.mendeley.com/forums/4941-general/suggestions/3425633-support-footnote-citations), as it does not recognize footnotes automatically (we need to insert a footnote manually, and then add the citation at the bottom).

libora:
@CharlesQc Yes I understand you now. Type of thesis is mapped to CSL variable "Genre" (in Zotero this field is named Type, in Mendely is directly named Genre).

adam3smith:
(we've been telling Mendeley they need to change this. @cpina -- pinging this again: any chance to get Thesis Type mapped to genre?)

libora:
adam3smith is true. In this version of Mendeley you cannot display type of thesis. But I think, your style should be for users of Zotero too, than I sugesst to add genre to your style.

adam3smith:
oh, we definitely want to do this with genre. I'm just saying that Mendeley should fix this (and that we've asked them to) so this may work for Mendeley, too, in the future, without the workaround of forcing it to display genre for thesis as you describe above.

libora:
@CharlesQc can you repleca line 269 by this code:

`       <choose>
          <if variable="genre">
            <text variable="genre"/>
          </if>
          <else>
            <text value="Thèse"/>
          </else>
        </choose>`

CharlesQc:
Thanks for the suggestion @libora and for the info @adam3smith . As stated in the commit, I used a macro because type "thesis" is used in the same way in citation and bibliography.

cpina:
@adam3smith : I've added a Mendeley ticket and I hope to do it soon. No reason for not changing it other than I forgot, apologies.

rmzelle:
Looks good here, @adam3smith. Will you merge?

PR #2038, opened 2016-04-26
Stripped periods

jmiba:
in journal abbreviations according to the guidelines of the faculty


rmzelle:
Can you remove the double space between the two attributes? (`form="short"  strip-periods="true"`)

rmzelle:
Thanks!

PR #2043, opened 2016-04-28
style for the "Journal fuer Kunstgeschichte"

blitzgneisserin:
This is a style for the "Journal fuer Kunstgeschichte" / "Journal of Art History". I have written this style a few months ago and you requested me to share it. Here is the referring discussion in the Zotero forum: https://forums.zotero.org/discussion/55979/stil-infoclio-anpassen/ 
Unfortunately there is no online documentation for this style. I have used my own copies in order to create it. Note that the field "short title" must be edited according to the instructions in "summary" if you want it to work properly.
I have never created a new file. I hope this works.


blitzgneisserin:
what is wrong here?

rmzelle:
The file name ("jfk"). This should be "journal-fur-kunstgeschichte.csl". See https://help.github.com/articles/renaming-a-file/

rmzelle:
You still need the ".csl" extension.

adam3smith:
should be almost there. Just delete the macro `creator-for-subsequent`.
Also, could you briefly explain this from the summary:
> Please edit the field "short title" and enter "author/editor" + "year" or "city" + "year".

rmzelle:
@adam3smith, @blitzgneisserin answered you in the commit message:

> Editing the short title manually is necessary because Zotero cannot produce short titles like "London 2016" or "Los Angeles 2003" automatically, but that is exactly how short titles of exhibition catalogues should look like in art historical papers. The short titles are also needed in the bibliography, which may look like this e. g. : http://simona.bplaced.net/wp-content/uploads/2016/01/jfk.png
Do you remember the discussion now?

adam3smith:
Thanks!

PR #2044, opened 2016-04-29
iso690-XXX-cz style update

libora:
replace standard space with non-breaking space in some cases
change date format from long to numeric for access date


adam3smith:
Thanks!

PR #2045, opened 2016-04-29
Docu journal-of-geriatric-psychiatry-and-neurology

zuphilip:
Update the docu link according to the 301 redirection (this is the first on my list).

I checked also the archived version http://web.archive.org/web/20140824045750/http://www.sagepub.com/journals/Journal201662/manuscriptSubmission which seems to be unchanged. However, I am not sure, that was really necessary.


rmzelle:
Thanks!

PR #2046, opened 2016-05-03
Significant update of De Buck

rjstapel:
My previous version of De Buck stylesheet still contained numerous mistakes. I have created a complete new update, which solves all issues known to me at the moment:
- Extended the application of the stylesheet to all possible file types
- Solved the issue that non-dropping particles were not capitalised in short-name format or first name in bibliography (G. de Koning > De Koning or De Koning, G.)
- Now able to refer to folio nrs. in books etc.
- Solved incorrect order of volume-nr. and series-nr.
- Added a book-author in book chapters etc.
- Added option for personal communication
- Added "recipient" and "interviewer" to relevant file types
- Added URL's to manuscripts
- Removed label "p." for page, which was always added to manuscripts.
- Solved issues with incorrect delimiters in both citation and bibliography
- Solved issue with translators
- Solved some issues with websites, presentations, theses, and other less common file types.
- Added the option of using date ranges (etc.) for publication dates, which are very common when working with Medieval/Early Modern material for instance, by including this information in the Zotero field "Extra" (note)
- Further minor changes

### DIFF
--- a/de-buck.csl
+++ b/de-buck.csl
@@ -20,18 +20,21 @@
     <category citation-format="note"/>
     <category field="generic-base"/>
     <summary>Annotatie en literatuurlijst volgens P. de Buck e.a., Zoeken en schrijven. Handleiding bij het maken van een historisch werkstuk (Haarlem 1982). Gebaseerd op "MHRA format with full notes and bibliography"</summary>
-    <updated>2013-01-20T12:26:52+01:00</updated>
+    <updated>2016-05-04T17:23:52+01:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="nl">
     <terms>
       <term name="et-al">e.a.</term>
-      <term name="editor" form="short">
+      <term name="editor" form="verb-short">
         <single>ed.</single>
-        <multiple>eds.</multiple>
+        <multiple>ed.</multiple>
       </term>
       <term name="edition" form="short">druk</term>
-      <term name="translator" form="verb-short">(vert.)</term>
+      <term name="translator" form="verb-short">
+        <single>(vertaler)</single>
+        <multiple>(vertalers)</multiple>
+      </term>
       <term name="volume" form="short">
         <single>dl.</single>
         <multiple>dls.</multiple>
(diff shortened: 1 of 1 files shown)


adam3smith:
Great, thanks!

You changed the indenting here, which makes this harder to review. Could you indent to two spaces/level, either using your own editor or http://formatter.citationstyles.org/

The other general comment I have is that this includes a lot of spaces in both delimiters, prefixes, and suffixes. That's very fragile and I can already spot some issues, e.g. 
https://github.com/citation-style-language/styles/pull/2046/files#diff-fba22804781d954ed366c23fabb02cc9R323 (that'll produce space -- comma -- space between label and name). There are also a lot of duplicate spaces caused by spaces in prefix and suffix -- citeproc-js (Zotero/Mendeley) will filter those out, but it's not great to rely on that. My recommendation would be to 1) use delimiters wherever possible and 2) where that not possible/too complicated, pick either prefix or suffix and only ever add spaces in one of them throughout the style.

rjstapel:
I tried to use the formatter you referred to, but not sure whether this produced the wanted result, there is still a large indent. I'll see what I can do about the double spaces, although that may take me some time.

adam3smith:
the formatting is perfect now, thanks

de-buck.csl
```diff
@@ -20,18 +20,21 @@
...
+      <term name="editor" form="verb-short">
         <single>ed.</single>
-        <multiple>eds.</multiple>
+        <multiple>ed.</multiple>
       </term>
       <term name="edition" form="short">druk</term>
-      <term name="translator" form="verb-short">(vert.)</term>
+      <term name="translator" form="verb-short">
+        <single>(vertaler)</single>
+        <multiple>(vertalers)</multiple>
```
rmzelle: If you don't mind, it's better to move the parenthesis out of the term defitions, and into the label (`<label ... prefix="(" suffix=")"/>`).

de-buck.csl
```diff
@@ -128,16 +146,8 @@
...
       </names>
-      <choose>
-        <if variable="author editor" match="any">
-          <names variable="translator" delimiter=", ">
-            <name and="text" delimiter=", " delimiter-precedes-last="never"/>
-            <label form="verb-short" text-case="lowercase" prefix=" " strip-periods="false"/>
-          </names>
-        </if>
-      </choose>
+      <!--<choose><if variable="author editor container-author" match="any"><names variable="translator" delimiter=", "><name and="text" delimiter=", " delimiter-precedes-last="never"/><label form="verb-short" text-case="lowercase" prefix=" " strip-periods="false"/></names></if></choose>-->
```
rmzelle: We prefer it if you delete commented XML code like that on line 150.

rmzelle:
Thanks!

rmzelle:
@rjstapel, we just got a report from a user of this style. They store non-date information in the item's "Extra" field, which messed up the citation. See https://forums.zotero.org/discussion/59393/

It's probably better to remove the use of the "note" CSL variable for date ranges. While it's an decent workaround for personal styles, we try to avoid using it in repository styles since it's a bit of a hack and it's often unclear to users what's going on. If you don't mind we can make the change ourselves.

(obviously we'd like Zotero to properly support date ranges, but this is the best we can do for now)

rjstapel:
Dear Rintze,
You are right, it was a good workaround for myself, but probably not helpful for larger scale use. If you want to make the adjustment, please do so.

All the best,
Rombert





Op 24 mei 2016, om 15:23 heeft Rintze M. Zelle <notifications@github.com<mailto:notifications@github.com>> het volgende geschreven:


@rjstapel<https://github.com/rjstapel>, we just got a report from a user of this style. They store non-date information in the item's "Extra" field, which messed up the citation. See https://forums.zotero.org/discussion/59393/

It's probably better to remove the use of the "note" CSL variable for date ranges. While it's an decent workaround for personal styles, we try to avoid using it in repository styles since it's a bit of a hack and it's often unclear to users what's going on. If you don't mind we can make the change ourselves.

(obviously we'd like Zotero to properly support date ranges, but this is the best we can do for now)

rmzelle:
I reverted the change: https://github.com/citation-style-language/styles/commit/69c9a3eee87fea04d29b0906726bc21d26a1387d

Thanks for chiming in!

PR #2047, opened 2016-05-04
Removed comma from citations

luciomarinelli:


adam3smith:
@luciomarinelli  -- thanks, looks good. I assume you did this using the visual editor? (it strips all xml comments, not sure if we want those back in).
@rmzelle -- thoughts on the comments? I think we can do without them in this case.

luciomarinelli:
Yes, used the online editor which stripped out the comments. I also suppose we can do without!

rmzelle:
These particular comments aren't very informative, so it's fine to remove them.

@luciomarinelli, your current code should be simplified a bit further from

```xml
      <group delimiter=" ">
        <group delimiter=" ">
          <text macro="author-cit"/>
          <text macro="date"/>
        </group>
      </group>
```

to

```xml
      <group delimiter=" ">
        <text macro="author-cit"/>
        <text macro="date"/>
      </group>
```

luciomarinelli:
ok, done

rmzelle:
Thanks!

PR #2048, opened 2016-05-10
Create reproduction-in-domestic-animals

okingfisher:
Style for the journal Reproduction in Domestic Animals, Wiley-Blackwell.

### DIFF
--- a/reproduction-in-domestic-animals
+++ b/reproduction-in-domestic-animals
@@ -1,0 +1,176 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" et-al-min="19" et-al-use-first="19" demote-non-dropping-particle="sort-only" default-locale="en-GB" xmlns="http://purl.org/net/xbiblio/csl">
+<!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>Reproduction in Domestic Animals</title>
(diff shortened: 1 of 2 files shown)


rmzelle:
Thanks. Could you add a ".csl" extension to the file name? (see https://help.github.com/articles/renaming-a-file/)

Also please add the eISSN number (`<eissn>1439-0531</eissn>`)

reproduction-in-domestic-animals
```diff
@@ -0,0 +1,176 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" et-al-min="19" et-al-use-first="19" demote-non-dropping-particle="sort-only" default-locale="en-GB" xmlns="http://purl.org/net/xbiblio/csl">
```
rmzelle: `et-al-min="19" et-al-use-first="19"`

et-al-use-first always must be a lower number than et-al-min (because it doesn't make sense to use et-al abbreviation for an item with 19 names and still show all 19 names)

rmzelle:
Thanks. You haven't yet addressed my [comment](https://github.com/citation-style-language/styles/pull/2048#discussion_r63086022) about `et-al-min="19" et-al-use-first="19"`, though. Could you fix that?

rmzelle:
Thanks!

By the way, do you remember which CSL style you used as a starting point?

okingfisher:
It could be Environmental Conservation, using http://editor.citationstyles.org/searchByExample/ (not quite close, though)

PR #2049, opened 2016-05-10
Sort inline citations by date then author

inkenbrandt:
Page 17 of the style guide (http://files.geology.utah.gov/online/c/c-105.pdf) states that when a sentence requires multiple
references, they are listed in chronologic-alphabetic order -- chronological first, then alphabetical.


adam3smith:
good catch, thanks!

PR #2050, opened 2016-05-10
Spell out "editor(s)" and add comma between and

inkenbrandt:
Following example of pg. 20 in the section `Documents having compilers/editors` of http://files.geology.utah.gov/online/c/c-105.pdf. Fixes the abbreviation `eds.` to the correct `editors` and adds a comma between the initials of the first editor and the word "and".

### DIFF
--- a/utah-geological-survey.csl
+++ b/utah-geological-survey.csl
@@ -135,9 +135,9 @@
           <group prefix=", ">
             <text term="in" font-style="italic"/>
             <names variable="editor translator" prefix=" " suffix="," delimiter=", ">
-              <name and="text" sort-separator=", " initialize-with=".," name-as-sort-order="all"/>
+              <name and="text" sort-separator=", " initialize-with="." name-as-sort-order="all" delimiter-precedes-last="always"/>
               <et-al term="and others"/>
-              <label form="long" prefix=" "/>
+              <label prefix=","/>
             </names>
             <text variable="event" prefix=" " suffix=" &#8211;"/>
             <text variable="container-title" prefix=" " suffix=":"/>
--- a/utah-geological-survey.csl
+++ b/utah-geological-survey.csl
@@ -137,7 +137,7 @@
             <names variable="editor translator" prefix=" " suffix="," delimiter=", ">
               <name and="text" sort-separator=", " initialize-with="." name-as-sort-order="all" delimiter-precedes-last="always"/>
               <et-al term="and others"/>
-              <label prefix=","/>
+              <label prefix=", "/>
             </names>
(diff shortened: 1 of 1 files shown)


utah-geological-survey.csl
```diff
@@ -135,9 +135,9 @@
           <group prefix=", ">
             <text term="in" font-style="italic"/>
             <names variable="editor translator" prefix=" " suffix="," delimiter=", ">
-              <name and="text" sort-separator=", " initialize-with="." name-as-sort-order="all"/>
+              <name and="text" sort-separator=", " initialize-with=".," name-as-sort-order="all"/>
               <et-al term="and others"/>
-              <label form="short" prefix=" "/>
+              <label form="long" prefix=" "/>
```
rmzelle: (it's not wrong, but typically we just delete `form="long"` since it's the default)

utah-geological-survey.csl
```diff
@@ -135,9 +135,9 @@
           <group prefix=", ">
             <text term="in" font-style="italic"/>
             <names variable="editor translator" prefix=" " suffix="," delimiter=", ">
-              <name and="text" sort-separator=", " initialize-with="." name-as-sort-order="all"/>
+              <name and="text" sort-separator=", " initialize-with=".," name-as-sort-order="all"/>
```
rmzelle: `initialize-with=".,"` can't be right. I'm pretty sure this would give names like "Doe, J.,B., and T.,S., Smith".

What you want is to use the "delimiter-precedes-last" attribute (and e.g. set it to "always"). See http://docs.citationstyles.org/en/stable/specification.html#name

utah-geological-survey.csl
```diff
@@ -135,9 +135,9 @@
           <group prefix=", ">
             <text term="in" font-style="italic"/>
             <names variable="editor translator" prefix=" " suffix="," delimiter=", ">
-              <name and="text" sort-separator=", " initialize-with="." name-as-sort-order="all"/>
+              <name and="text" sort-separator=", " initialize-with="." name-as-sort-order="all" delimiter-precedes-last="always"/>
               <et-al term="and others"/>
-              <label form="short" prefix=" "/>
+              <label prefix=","/>
```
rmzelle: Did you test this?

To get "in Miller, D.M., Todd, V.R., and Howard, K.A., editors" (p. 20), you'd probably need `<label prefix=", "/>`, no? (with added space)

rmzelle:
@inkenbrandt, thanks!

(by the way, it's better to leave a comment in the pull request than to add a note to your commit when you respond to me; we always see the former, but not necessarily the latter)

PR #2051, opened 2016-05-12
Create CSL for Land Degradation & Development

gilrockmendes:


rmzelle:
Thanks. Could you go through https://github.com/citation-style-language/styles/wiki/Style-Requirements and update your style as necessary?

(the file name should be "land-degradation-and-development.csl"; see https://help.github.com/articles/renaming-a-file/)

land-degradation-and-development.csl
```diff
@@ -0,0 +1,178 @@
...
+      <substitute>
+        <names variable="editor"/>
+        <names variable="translator"/>
+        <text macro="title"/>
+      </substitute>
+    </names>
+  </macro>
+  <macro name="author-short">
+    <names variable="author">
+      <name form="short" and="symbol" et-al-min="3" et-al-use-first="1" initialize-with=". "/>
```
adam3smith: remove `et-al-min="3" et-al-use-first="1" ` that's already covered in by the same info under citation.

land-degradation-and-development.csl
```diff
@@ -0,0 +1,178 @@
...
+      <et-al font-style="italic"/>
+      <substitute>
+        <names variable="editor"/>
+        <names variable="translator"/>
+        <text variable="container-title" font-style="italic"/>
```
adam3smith: this seems odd -- if you don't have the same line in the author macro, matching citation to bibliography would be quite tricky.

adam3smith:
See two nits in comments inline, everything else looks good.

rmzelle:
Add in #2261 with some minor changes.

Thanks, @gilrockmendes, and sorry for the long wait.

Issue #2053, opened 2016-05-15
Czech ISO690 styles accessed date format changes.


wilx:
I find the last changes in Czech ISO690 styles dubious. The accessed date looks wrong when rendered by Pandoc. The format defined in https://github.com/citation-style-language/styles/blob/master/iso690-author-date-cs.csl#L390 produces "vid. 2016-květen-12". This looks unlike any typical Czech date format. It should probably be like this to make any sense:

```{xml}
          <date variable="accessed">
            <date-part name="year"/>
            <date-part name="month" form="numeric-leading-zeros" prefix="-"/>
            <date-part name="day" form="numeric-leading-zeros" prefix="-"/>
          </date>
```
This applies to all of the Czech ISO690 styles.

adam3smith:
@libora -- this is your change. I can't judge what's common in Czech or in the local version of the ISO norms, but would agree that 2016-05-12 looks more plausible (esp. for ISO) than what we have now.

libora:
@adam3smith wilx is true, my mistake, sorry. I will chance it immediately when I will be by PC. Take me 2 hours. Libor

libora:
It is done - pull request #2054

rmzelle:
(this PR is related to #2039 and #2044)

rmzelle:
Thanks for reporting this, @wilx!

PR #2054, opened 2016-05-16
issue #2053

libora:
Make date format compatible with ISO 8601


adam3smith:
thanks. closes https://github.com/citation-style-language/styles/issues/2053

PR #2055, opened 2016-05-16
Update uppsala-universitet-institutionen-for-biologisk-grundutbildnin…

perenstrom:
…g.csl

Changed all container-title fields to long form according to our own instructions.


uppsala-universitet-institutionen-for-biologisk-grundutbildning.csl
```diff
@@ -14,29 +14,32 @@
...
       <term name="et-al">et al.</term>
     </terms>
   </locale>
+
   <locale xml:lang="en-GB">
     <date form="text">
-      <date-part name="day" suffix=" "/>
-      <date-part name="month" suffix=" "/>
-      <date-part name="year"/>
+      <date-part name="day" suffix=" " />
```
adam3smith: not sure where these come from, but could you remove the trailing spaces here (i.e. the space _after_ the last ") and below in row 40-42?

adam3smith:
thanks!

rmzelle:
(it doesn't matter much, but it's slightly preferred to just delete `form="long"`, which is the default, to keep style compact)

PR #2056, opened 2016-05-16
American journal of climate change

arunbte:
Citation style for the journal American Journal of Climate Change

PR #2063, opened 2016-05-17
Citation style for the journal American Journal of Climate Change

arunbte:
Citation style for the journal American Journal of Climate Change

Updated according the 'Style Requirements'


rmzelle:
Closing since we already have #2056, which looks to be in better shape.

PR #2066, opened 2016-05-18
Water resources research

arunbte:
Citation style for water resources research


libora:
@rmzelle AGU has same styles rules for all journals in their portfolio. I think the better way is to rename Water Resource Research style to AGU and create WRR style as dependent style. Do you agree?

rmzelle:
Looks like we already have a CSL style for the American Geophysical Union:

https://www.zotero.org/styles?q=american%20geophysical%20union

@arunbte, could you check if this has the right format for you?

(p.s. for a list of AGU journals, see http://publications.agu.org/journals/)

libora:
Style from @arunbte does not sort bibliography correctly for 2 or more authors. Styles from repository does it perfectly due sorting attribute "names-min" (unfortunately I do not find the exact description of this attribute in CSL specification).There is also some other differences in arunbte's style (doubled space, commas, italics font etc.) from style in repository. Some differences are errors but some cannot by verified in Guide for Authors (e.g. italics in name of conference).

libora:
@rmzelle I create table for all AGU journals - https://github.com/citation-style-language/journals/pull/9

arunbte:
Agreed. I don't know about the existing style. So, I'm reverting back the submitted one.

rmzelle:
@arunbte, we'll use @libora's one to get the WRR style into the repository.

As an aside, you're over-complicating your use of GitHub. First, it's rarely necessary to revert commits. Secondly, while you created a "water-resources-research" branch, somehow you started with a bunch of commits for the American Journal of Climate Change (probably by branching off "american-journal-of-climate-change" instead of off "master"). I'm also not sure why you created #2063 when we already had #2056.

In general, it's much easier to follow our instructions at https://github.com/citation-style-language/styles/blob/master/CONTRIBUTING.md#3-create-a-pull-request. You don't have to create any new pull requests now, but those instructions might help you in the future.

arunbte:
@rmzelle Hi, I'm new toGitHub. Just learning the things. Will avoid all those errors in future.

rmzelle:
No problem :).

PR #2067, opened 2016-05-20
Create QUTAPA.csl

healyn:
Hi, 

This is a style created for Queensland University of Technology. It is the APA 6th style with just doi: instead of http://doi

We would like to create this so that TALIS Aspire Readings lists can pull in the recommended citation style for QUT.

Thanks
Nicole Healy

### DIFF
--- a/QUTAPA.csl
+++ b/QUTAPA.csl
@@ -1,0 +1,617 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
+<!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>QUT APA</title>
+    <title-short>QUTAPA</title-short>
+    <id>http://www.zotero.org/styles/qut-apa</id>
+    <link href="http://www.zotero.org/styles/qut-apa" rel="self"/>
+    <link href="http://owl.english.purdue.edu/owl/resource/560/01/" rel="documentation"/>
+    <author>
+      <name>Simon Kornblith</name>
+      <email>simon@simonster.com</email>
+    </author>
+    <contributor>
+      <name>Bruce D'Arcus</name>
+    </contributor>
+    <contributor>
+      <name>Curtis M. Humphrey</name>
+    </contributor>
+    <contributor>
+      <name>Richard Karnesky</name>
+      <email>karnesky+zotero@gmail.com</email>
+      <uri>http://arc.nucapt.northwestern.edu/Richard_Karnesky</uri>
+    </contributor>
+    <contributor>
+      <name>Sebastian Karcher</name>
+    </contributor>
+    <category citation-format="author-date"/>
+    <category field="psychology"/>
+    <category field="generic-base"/>
+    <summary>Modified from APA 6th to suit QUT</summary>
+    <updated>2016-05-20T04:26:35+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="en">
+    <terms>
+      <term name="editortranslator" form="short">
+        <single>ed. &amp; trans.</single>
+        <multiple>eds. &amp; trans.</multiple>
+      </term>
+      <term name="translator" form="short">
+        <single>trans.</single>
+        <multiple>trans.</multiple>
+      </term>
+    </terms>
+  </locale>
+  <macro name="container-contributors">
+    <choose>
+      <if type="chapter paper-conference entry-dictionary entry-encyclopedia" match="any">
+        <group delimiter=", ">
+          <names variable="container-author" delimiter=", ">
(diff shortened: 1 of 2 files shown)


rmzelle:
@healyn, are you submitting #2068 and #2069 as well for the sole purpose of using these styles with Talis Aspire Reading Lists?

I'm sorry, but in general, we don't accept styles like these (see https://github.com/citation-style-language/styles/wiki/Criteria-for-Accepting-Styles) which are only expected to be used by a single user (in this case, you). The reason is that our repository styles are automatically made available to hundreds of thousands of users, and these variants are unlikely to offer value to any of them.

The better way would be to ask Talis if they can add support for CSL styles hosted outside of our repository.

(we would likely accept these styles if they're also used in writing; let me know if it's the case that these style variants are part of your university-wide style guide for academic writing)

adam3smith:
If these are used in writing, I'd very much urge whoever is in charge to not require alterations to standards. E.g. it makes a *lot* of sense to print DOIs as links and not as just the number preceded by doi:, which is why the URL format is now recommended by APA, CrossRef, and doi.org

healyn:
Hi,

Thanks for your responses. To answer your questions. The three styles I submitted are all to make sure that the QUT accepted styles can be used with the Talis Reading List System. I'm the project manager for the system implementation, and the system will be used by all 45,000 QUT students. These styles are also used university wide for writing. The university style guide can be found here http://www.citewrite.qut.edu.au/. This guide is both online and in printed form and is provided for all QUT students.

In reference to all three styles. I'm happy to withdraw #2068 as I've done some further investigation and negotiation and we are happy with the current Australian Guide to legal style and will use this.

For #2067, QUT has gone with doi: based on the advice from APA in their blog http://blog.apastyle.org/apastyle/digital-object-identifier-doi/ . In this blog it notes that doi: is correct as is http://dx.doi.org/. QUT just happens to have chosed the doi: version rather than the URL. 

For #2069 The QUT official advice to students is to use square brackets for the reference list. In fact Elsevier and SAGE both use square brackets for their version of Vancouver as well. It would be great given that this is for a university wide system if we could have this style approved. 

Thanks for your help and consideration.

Nicole Healy.

healyn:
Hi, 

I've just managed to find a style called "Taylor & Francis - National Library of Medicine" which will be fine to use instead of #2069 so i'll close this one as well. 

I would still appreciate your consideration of this APA request however. 

Thanks
Nicole.

rmzelle:
@adam3smith, I take it we'd accept the APA variant as long as it's clearly labeled as a QUT style?

@healyn, we would still appreciate it very much if you could advocate for a future switch within QUT to APA with `http://doi.org/` prefixed DOIs (and not `http://dx.doi.org/`, which is outdated*). The URL format is clearly preferred by APA (from the blog post: "Recently, however, CrossRef changed the format of the DOI to a more user-friendly one in the form of a URL"), and the fewer style variants we have in our repository, the easier it is to maintain.

*The DOI organization writes: "DOIs may ... use ... http://doi.org (or http://dx.doi.org, an earlier syntax that is fully supported but no longer preferred), see http://www.doi.org/doi_handbook/2_Numbering.html#htmlencoding

adam3smith:
For Vancouver, there's also https://www.zotero.org/styles?id=vancouver-brackets which would be exactly what you had originally (T&F has some small variations such as different et-al settings) so you could also look at that.
APA -- like @rmzelle , advocating for displaying the DOI in URL form would be by far my preferred solution. If that's not possible, though, I think I'd prefer turning this into a generic DOI variant of APA makes more sense (since we get this request an fair amount). I don't think it makes sense to label this as QUT format when it's something that other people want to use (even if foolishly). We could consider making it clear that the DOI format is a 2nd best solution via the title, maybe something like (old DOI format) or so?

rmzelle:
> I think I'd prefer turning this into a generic DOI variant of APA makes more sense (since we get this request an fair amount). I don't think it makes sense to label this as QUT format when it's something that other people want to use (even if foolishly). We could consider making it clear that the DOI format is a 2nd best solution via the title, maybe something like (old DOI format) or so?

Maybe `<title>American Psychological Association 6th edition ("doi:" DOI prefix, obsolete)</title>` and "apa-old-doi-prefix.csl"?

healyn:
Hi, 

Thanks for you assistance on this. I would really appreciate a style with the "old" doi format still available. Changing our writing guide and communicating the change university wide is a fairly big undertaking and will take some time even if we can get agreement from the university to make the change. It is definitely on the radar as an open discussion now, but we only update the writing guide once per year. 

Any assistance you can provide would be greatly appreciated.

Thanks
Nicole

rmzelle:
> It is definitely on the radar as an open discussion now

Great.

Looks like we'll just add an APA variant, then. Just need to figure out the best name.

adam3smith:
> Maybe <title>American Psychological Association 6th edition ("doi:" DOI prefix, obsolete)</title> and "apa-old-doi-prefix.csl"?

I'm OK with that. I think "obsolete" might be a bit strong? How about (old "doi:" DOI prefix)?

rmzelle:
Actually, I think it's enough to add "old" to the file name. The fact that it has a description within parenthesis already implicitly suggests the style isn't the default, and APA still accepts both variants. PR created: https://github.com/citation-style-language/styles/pull/2077

@adam3smith, will you review and merge?

PR #2072, opened 2016-05-22
Create canadian-geotechnical-journal.csl

libora:
Style for Canadian Geotechnical Journal. Discussion https://forums.zotero.org/discussion/59271/style-request-canadian-geotechnical-journal/#Comment_254267


adam3smith:
thanks!

PR #2079, opened 2016-05-25
OU Harvard improvements

pmelchor:
- Change documentation URL for easier access to the PDF version, which does not require an OU login
- Create 'director' macro for use in film references
- 'access' macro: use the 'available at' term instead of hardcoded text; replace hardcoded spacing with group delimiter
- Prepare citation and bibliography for type specific formats
- Create type specific formats for films

Example of film citation:
(_Lord of the Rings: The Two Towers_, 2003)

Example of film bibliography reference:
_Lord of the Rings: The Two Towers_ (2003) Directed by Peter Jackson [Film]. New York, Newline Productions Inc.

### DIFF
--- a/the-open-university-harvard.csl
+++ b/the-open-university-harvard.csl
@@ -15,7 +15,7 @@
     <category citation-format="author-date"/>
     <category field="generic-base"/>
     <summary>The Open University Harvard author-date style</summary>
-    <updated>2016-05-24T06:45:00+00:00</updated>
+    <updated>2016-05-25T07:13:00+00:00</updated>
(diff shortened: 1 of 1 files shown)


the-open-university-harvard.csl
```diff
@@ -151,16 +159,28 @@
...
+        </if>
+        <!-- Generic/fallback format (citation) -->
+        <else>
+          <group delimiter=", ">
+            <group delimiter=", ">
+              <text macro="author-short"/>
+              <text macro="year-date"/>
+            </group>
+            <group>
+              <text term="page" form="short" suffix=" "/>
```
adam3smith: Use label variable= locator here, so out works for other locator like chapter (sorry for formatting, typing on phone)

pmelchor:
@adam3smith Done!

adam3smith:
thanks!

PR #2088, opened 2016-06-01
Create pontifical-gregorian-university-de.csl

PUG-IS:
This is the German version of the Pontifical Gregorian University style. Shortly I'll add other languages (French,Spanish,English)

### DIFF
--- a/pontifical-gregorian-university-de.csl
+++ b/pontifical-gregorian-university-de.csl
@@ -1,0 +1,257 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" initialize-with-hyphen="true" page-range-format="expanded" demote-non-dropping-particle="sort-only" name-delimiter=" &#8211; " names-delimiter=" &#8211; " delimiter-precedes-last="always" et-al-min="4" et-al-use-first="1" name-form="long" initialize-with=". " sort-separator=", " default-locale="de-DE">
+  <info>
+    <title>Pontifical Gregorian University (German)</title>
+    <id>http://www.zotero.org/styles/pontifical-gregorian-university-de</id>
+    <link href="http://www.zotero.org/styles/pontifical-gregorian-university-de" rel="self"/>
+    <link href="http://www.unigre.it" rel="documentation"/>
+    <author>
+      <name>Information Systems</name>
+      <email>ced(@) unigre{.} it  (remove spaces and parenthesis/brackets but leave periods to get email)</email>
+    </author>
+    <category citation-format="note"/>
+    <category field="theology"/>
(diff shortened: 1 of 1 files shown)


rmzelle:
Do you know if it would be possible to cover most of these languages with a single independent style?

(for a similar situation, see https://www.zotero.org/styles?q=pontificia%20salesiana, where the English, French, Spanish, and Portuguese versions share the same style)

(for a description of independent and dependent styles, see http://docs.citationstyles.org/en/stable/primer.html#independent-and-dependent-styles)

PUG-IS:
Thanks @rmzelle ,
I think that i can do a dependent style for French and Spanish, the only change is about the translation of some "terms" elements (e.g. "editor").
The German style uses different kind of suffix and prefix than the others one, and i don't know if it is possible to use a dependent style.
The English style and the Italian style (already published) are equal, so probably I’ll just have to rename the title.
Is it the correct way?

rmzelle:
> I think that i can do a dependent style for French and Spanish, the only change is about the translation of some "terms" elements (e.g. "editor").

Yes, in that case a dependent style is definitely possible.

> The German style uses different kind of suffix and prefix than the others one, and i don't know if it is possible to use a dependent style.

Can you give an example? (the only thing close that is localizable are quote marks)

> The English style and the Italian style (already published) are equal, so probably I’ll just have to rename the title.

For an example of a dependent and independent style, see https://github.com/citation-style-language/styles/blob/master/dependent/universita-pontificia-salesiana-fr.csl and https://github.com/citation-style-language/styles/blob/master/universita-pontificia-salesiana.csl. The former relies on the latter, but provides an (overriding) "default-locale" value.

Since the repository already has the Italian version as an independent style, you can make an English dependent style that points to the Italian style. Use `default-locale="en-US"` for US English or "en-GB" for British English.

PUG-IS:
> I think that i can do a dependent style for French and Spanish, the only change is about the translation of some "terms" elements (e.g. "editor").

> Yes, in that case a dependent style is definitely possible.

Perfect...I'll do it.

>Can you give an example? (the only thing close that is localizable are quote marks)

Sure.
Italian,Spanish and French style:
`<text variable="title" form="short" quotes="false" prefix="«" suffix="»"/>`
German style:
`<text variable="title" form="short" quotes="false" prefix="„" suffix="“"/>`

Thanks.

adam3smith:
Perfect. If you use `quotes="true"` that will work for all of those languages exactly as you want it. I.e.
`<text variable="title" form="short" quotes="true"/>`

PUG-IS:
Ok...I'm editing the main style ( Italian ) to fix some problems and then i will create dependent styles for the other languages.

PUG-IS:
Ok...I'm editing the main style ( Italian ) to fix some problems and then i will create dependent styles for the other languages.

PR #2091, opened 2016-06-03
Add issue number

anthracene:
Adds parenthetical issue number immediately following volume, e.g.
volume(issue)
per Radiology style guide: http://pubs.rsna.org/page/radiology/pia/preparation#references

### DIFF
--- a/radiology.csl
+++ b/radiology.csl
@@ -180,6 +180,11 @@
     <choose>
       <if type="article-journal article-magazine" match="any">
         <text variable="volume"/>
+        <choose>
+          <if match="any" variable="issue">
+            <text variable="issue" prefix="(" suffix=")"/>
+          </if>
+        </choose>
       </if>
     </choose>
   </macro>
--- a/radiology.csl
+++ b/radiology.csl
@@ -180,11 +180,7 @@
     <choose>
       <if type="article-journal article-magazine" match="any">
         <text variable="volume"/>
-        <choose>
-          <if match="any" variable="issue">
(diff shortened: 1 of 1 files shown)


radiology.csl
```diff
@@ -180,6 +180,11 @@
     <choose>
       <if type="article-journal article-magazine" match="any">
         <text variable="volume"/>
+        <choose>
+          <if match="any" variable="issue">
+            <text variable="issue" prefix="(" suffix=")"/>
+          </if>
+        </choose>
```
rmzelle: The whole `<choose/>` block is unnecessary here. You can just use `<text variable="issue" prefix="(" suffix=")"/>` instead. Can you update your pull request?

anthracene: Thanks; wasn't clear to me from the documentation that the text node would
be skipped if the variable was empty -- I was worried about getting () in
the citation if there was no issue #.

I believe I've updated the pull request as you asked. I'm new to github, so
if it didn't come through please let me know and I'll try again.

Thanks,

John

rmzelle: The CSL specification is a bit dense, but it's there:

http://docs.citationstyles.org/en/stable/specification.html#affixes
> The attribute value is either added before (prefix) or after (suffix) the output of the element carrying the attribute, but *affixes are only rendered if the element produces output*.

rmzelle:
Perfect, thanks.

anthracene:
Great, thanks for your help!

John

PR #2094, opened 2016-06-07
Create Skandinavistik.csl

arqen:
Citation style for German Scandinavian Studies

I took the Style of Heilmann (https://github.com/citation-style-language/styles/blob/master/geistes-und-kulturwissenschaften-heilmann.csl) and changed it according to the style sheet for most of the German Scandinavian Studies.

- using full stops instead of commas to section the parts

- title is in italics (if it's a monography) or the journal title is in italics


rmzelle:
Hi. Sorry for the late response.

Is there an official style guide for this citation format that we can link to?

rmzelle:
Closing because of inactivity. @arqen, if you get back to us we can just reopen this pull request.

PR #2096, opened 2016-06-07

ibadger:
Tidy up lines 218-220 on how directors are introduced


adam3smith:
Thanks!

PR #2099, opened 2016-06-10
Update harvard-stellenbosch-university.csl

Yusufras:
change:

```xml
<else-if type="chapter paper-conference entry-encyclopedia entry-dictionary" match="any">
         <text macro="title"/>
            <group delimiter=" ">
              <text macro="book-details"/>
```

to:

```xml
          <text macro="title" prefix=" " suffix="."/>
            <text macro="editor"/>
            <text variable="event" suffix="."/>
            <group suffix="." delimiter=": ">
              <text macro="pages"/>
```

### DIFF
--- a/harvard-stellenbosch-university.csl
+++ b/harvard-stellenbosch-university.csl
@@ -274,12 +274,17 @@
             <text variable="genre"/>
             <text variable="publisher"/>
           </else-if>
-          <else-if type="chapter paper-conference entry-encyclopedia entry-dictionary" match="any">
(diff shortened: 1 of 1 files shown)


rmzelle:
@Yusufras, validation is failing because you introduced an additional `<group>` tag that isn't closed with `</group>`. You also now call an "editor" macro that doesn't exist in the style.

rmzelle:
(it's preferable to update this pull request instead of creating a new one)

PR #2101, opened 2016-06-13
Create negative-results.csl

Negative-Results:
We are pleased to introduce you to Negative Results, a journal that publishes articles that challenge current knowledge, existing models, and tenets. The articles are based on rigorous, and well-documented results that do not support these models or even refute them. Scientists usually communicate their findings by publishing results in specialized peer-reviewed journals. However negative findings – those that do not agree with what the researchers hypothesized – are often disregarded, neglected and therefore not put forward for publication. The stigma surrounding negative findings means that they are a low priority for publication.

### DIFF
--- a/dependent/negative-results.csl
+++ b/dependent/negative-results.csl
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" default-locale="en-US">
+ 
+  <info>
+    <title>Negative Results</title>
+    <title-short>Neg Res</title-short>
+    <id>http://www.zotero.org/styles/negative-results</id>
+    <link href="http://www.zotero.org/styles/negative-results" rel="self"/>
+    <link href="http://www.zotero.org/styles/cell" rel="independent-parent"/>
+    <category citation-format="author-date"/>
+    <category field="science"/>
+    <issn></issn>
+    <eissn></eissn>
+    <updated>2016-06-10T01:40:32+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+</style>
--- a/dependent/negative-results.csl
+++ b/dependent/negative-results.csl
@@ -9,8 +9,6 @@
     <link href="http://www.zotero.org/styles/cell" rel="independent-parent"/>
     <category citation-format="author-date"/>
     <category field="science"/>
-    <issn></issn>
-    <eissn></eissn>
     <updated>2016-06-10T01:40:32+00:00</updated>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks! Is this journal actually published by cell press? Do you have a link to your homepage and/or the guidelines for contributors?

Negative-Results:
No, our journal will not published by cell press. And regarding the second question :
- homepage : http://www.negative-results.org
- the guidelines for contributors : http://www.negative-results.org/index.php/instructions-for-authors/
Thanks

adam3smith:
Thanks. I guess I don't understand why you're linking to the Cell citation style then? I'm all in favor of you using an existing style, but what you require on your page is

Jackson, B.T. Kelly, A.J. McCoy, T. Gaffry, L.C. James, B.M. Collins, S. Honing, P.R. Evans, D.J. Owen A large-scale conformational change couples membrane recruitment to cargo binding in the AP2 clathrin adaptor complexCell, 141 (2010), pp. 1220–1229

whereas Cell style is

Watson, J.D., and Crick, F.H.C. (1953). Molecular structure of nucleic acids: a structure for deoxyribose nucleic acid. Nature 171, 737–738.

i.e. quite distinct.
(in your example, Jackson is missing initials, btw.)

Negative-Results:
Indeed, our webmaster make a mistake and normally it's ok now. So, I know we use Cell citation style but if it's possible we would like to do that with our journal name, only for our identity.

dependent/negative-results.csl
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" default-locale="en-US">
+ 
+  <info>
+    <title>Negative Results</title>
+    <title-short>Neg Res</title-short>
+    <id>http://www.zotero.org/styles/negative-results</id>
+    <link href="http://www.zotero.org/styles/negative-results" rel="self"/>
+    <link href="http://www.zotero.org/styles/cell" rel="independent-parent"/>
```
adam3smith: please add
`<link href="http://www.negative-results.org/index.php/instructions-for-authors/#References" rel="documentation"/>` here

dependent/negative-results.csl
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" default-locale="en-US">
+
```
adam3smith: remove empty line

adam3smith:
OK, no problem. As long as Cell style is actually what you want. No ISSN yet, correct?
See my two comments in the code. Otherwise this is good to go.

Negative-Results:
No ISSN yet, we are a young journal and we need one publication for that. Thanks a lot

adam3smith:
thanks!

PR #2102, opened 2016-06-13
Update seminaire-saint-sulpice-ecole-theologie.csl

guschartier:
Minor changes about footnote and update style, website


seminaire-saint-sulpice-ecole-theologie.csl
```diff
@@ -7,26 +7,30 @@
...
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="fr">
     <terms>
-      <term name="cited"/>
-      <term name="page" form="short">
+        <term name="in">
+            <single>dans</single>
+            <multiple>dans</multiple>
+        </term>
```
rmzelle: Can you just use `<term name="in">dans</term>`? There is no need to define singular and plural versions.

rmzelle:
You current

```
        <term name="in">
            <single>dans</single>
        </term>
```

should be

```
        <term name="in">dans</term>
```

guschartier:
thanks

rmzelle:
Thanks!

PR #2103, opened 2016-06-14
Create pontifical-gregorian-university-fr

PUG-IS:


dependent/pontifical-gregorian-university-fr.csl
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" initialize-with-hyphen="true" page-range-format="expanded" demote-non-dropping-particle="sort-only" name-delimiter=" &#8211; " names-delimiter=" &#8211; " delimiter-precedes-last="always" et-al-min="4" et-al-use-first="1" name-form="long" initialize-with=". " sort-separator=", " default-locale="fr-FR">
```
adam3smith: @rmzelle -- do you see any problems with the formatting attributes (like name delimiters etc.) here for a dependent style? Do you know, btw., if they would override attributes in a parent style (I assume not)?

rmzelle: It's better to strip them, but they don't do anything. The only attribute in dependent styles that has any effect is "default-locale".

rmzelle: @PUG-IS, you can reduce the second line to:

```
<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" default-locale="fr-FR">
```

PUG-IS: Done. Thanks. 👍

adam3smith:
Thanks!

PUG-IS:
@rmzelle , There is a problem with dependent styles. It seems that dependent style use 
`...class="in-text"` 
settings even though their parent style is set to 
`...class="note"`.
So actually the dependent styles that I created yesterday don't insert footnotes anymore.

rmzelle:
> It seems that dependent style use 
> ...class="in-text" 
> settings even though their parent style is set to 
> ...class="note".

Is that with Zotero?

PUG-IS:
Yes, Zotero for Firefox (4.0.29.10).  I tried on Windows and Mac.

rmzelle:
Have you tried reinstalling the parent and dependent styles?

I've never seen Zotero reports that this doesn't work as expected. You can also check your "styles" folder in your Zotero data directory to check that both the parent and dependent are set to "note".

adam3smith:
doesn't work for me either (i.e. I can replicate the issue). The dependent style doesn't have `class="note"` set, but they shouldn't have to, right?

PUG-IS:
> Have you tried reinstalling the parent and dependent styles?
Yes, i tried..nothing changed.

adam3smith:
same thing -- MLA note works, Angelaki doesn't.
I vaguely seem to remember that this has come up before?

rmzelle:
Don't recall. Think either Zotero or citeproc-js is defaulting to "in-text" for the dependent when the dependent doesn't have "class" specified on `<style>`?

PUG-IS:
Setting class="note" in the dependent style seems fix...

rmzelle:
Probably needs a fix in Zotero at https://github.com/zotero/zotero/blob/045554dd8e58a8a02bbac2cf621790db67797944/chrome/content/zotero/xpcom/style.js#L711. (the code right below does look up the parent: https://github.com/zotero/zotero/blob/045554dd8e58a8a02bbac2cf621790db67797944/chrome/content/zotero/xpcom/style.js#L721)

PUG-IS:
@rmzelle,  Do you think that I can fix the problem temporaly adding class="note" in dependent style?

adam3smith:
I'd be OK with that since it's legal CSL and I'm not sure how quick we'll be able to get the fix into Zotero. @rmzelle -- agree?

rmzelle:
Sure. I'll file a ticket with Zotero, though, since this should be fixed. Surprised we haven't gotten more reports for this.

PUG-IS:
Thanks!

PR #2108, opened 2016-06-14
Create world-s-poultry-science-journal.csl

andywales:
'World's Poultry Science Journal' csl style file, adapted from 'Ecology' csl style


rmzelle:
Thanks.

Could you change the file name to "worlds-poultry-science-journal.csl"? (and update the style ID and "self" link to match?)

Also please delete `<link href="http://esapubs.org/esapubs/AuthorInstructions.htm" rel="documentation"/>`, and add the eISSN (`<eissn>1743-4777</eissn>`). And run the style through http://formatter.citationstyles.org/, since the indenting is a bit off in the `<info/>` section.

rmzelle:
Great, thanks!

PR #2110, opened 2016-06-15
Improved computer programs layout and bibliography item sorting 

joejoinerr:


harvard-leeds-beckett-university.csl
```diff
@@ -15,6 +15,65 @@
...
+      <else-if variable="author">
+        <names variable="author">
+          <name initialize-with=". " name-as-sort-order="all" delimiter=", " delimiter-precedes-last="never" and="symbol"/>
+        </names>
+      </else-if>
+      <else-if variable="editor">
+        <names variable="editor">
+          <name initialize-with=". " name-as-sort-order="all" delimiter=", " delimiter-precedes-last="never" and="symbol" suffix=" ed."/>
+        </names>
+      </else-if>
```
rmzelle: You can simplify

```xml
      <else-if variable="author">
        <names variable="author">
          <name initialize-with=". " name-as-sort-order="all" delimiter=", " delimiter-precedes-last="never" and="symbol"/>
        </names>
      </else-if>
      <else-if variable="editor">
        <names variable="editor">
          <name initialize-with=". " name-as-sort-order="all" delimiter=", " delimiter-precedes-last="never" and="symbol" suffix=" ed."/>
        </names>
      </else-if>
```

to

```xml
      <else-if variable="author editor" match="any">
        <names variable="author">
          <name initialize-with=". " name-as-sort-order="all" delimiter=", " delimiter-precedes-last="never" and="symbol"/>
          <substitute>
            <names variable="editor"/>
          </substitute>
        </names>
      </else-if>
```

(the "author" label term is empty)

rmzelle: If you do this, you might have to add a `<locale/>` section though to redefine the "short" form of the "editor" term, if you don't want "eds." for multiple editors. I.e.:

```xml
  <locale xml:lang="en">
    <terms>
        <single>ed.</single>
        <multiple>ed.</multiple>
      </term>
    </terms>
  </locale>
```

joejoinerr: Can you clarify the purpose of `<label form="short" prefix=" "/>`? If it's empty why should it exist?

Also, how is the "ed." suffix rendered in your simplified example?

rmzelle: > If it's empty why should it exist?

It's only empty for "author", not for the substituting "editor".



> how is the "ed." suffix rendered in your simplified example?

Through `<label/>`. It calls the short form of the "editor" term as defined in the CSL locale file (or as redefined in the style itself).





> Is there any way to get it to show after each name e.g. "Gobo, G. ed. & Mauceri, S. ed."?

That's rather non-standard. I guess that's only possible with your original code.

joejoinerr: No matter. I've just referred to the docs and it says this:

> ‘ed.’ and ‘eds.’ are suitable abbreviations for ‘editor’ and ‘editors’

That voids the need for that extra `<locale/>` section too. I'll update with the changes.

rmzelle:
Great, thanks!

PR #2111, opened 2016-06-17
Update pontifical-gregorian-university.csl

PUG-IS:
Update et-al-min value.

### DIFF
--- a/pontifical-gregorian-university.csl
+++ b/pontifical-gregorian-university.csl
@@ -77,7 +77,7 @@
   </locale>
   <macro name="autori-nota">
     <names variable="author" font-variant="normal" font-weight="normal">
-      <name font-variant="normal" font-weight="normal" et-al-min="4" et-al-use-first="1" initialize-with=".">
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks.

By the way, it might be possible to simplify the CSL style a little. `et-al-min` and `et-al-use-first` are both inheritable name options (see http://docs.citationstyles.org/en/stable/specification.html#inheritable-name-options), and can be defined on `<style>`, `<citation>`, `<bibliography>`, and `<name>`. Values on `<name>` override those on `<style>`, `<citation>`, and `<bibliography>`, and values on `<citation>` and `<bibliography>` override those on `<style>`.

Your style currently has et-al attributes on `<style>`, `<bibliography>`, and all `<name>` elements. It looks like you can remove all of them except those on `<style>`.

PUG-IS:
Thanks. I just updated the style code, except on `<bibliography>`.

rmzelle:
Thanks!

PR #2113, opened 2016-06-17
Create journal-systematic-palaeontology.csl

nsvitek:
Proposed csl file for the Journal of Systematic Palaeontology, based on the csl file for the Geological Society of London. Changes to template style are as follows:
(1) citation ordering and formatting edited to order by author then shortened year (copying code from Methods in Ecology & Evolution csl file).
(2) bibliography authors formatted to be upper and lowercase (not small caps) and emboldened.
(3) space between authors' initials in bibliography added.
(4) semi-colon (not comma) used to separate authors in citation.
(5) number of pages added to book citation.
(6) publisher/publisher place ordering reversed.

### DIFF
--- a/journal-systematic-palaeontology.csl
+++ b/journal-systematic-palaeontology.csl
@@ -11,7 +11,7 @@
     </author>
     <category citation-format="author-date"/>
     <category field="biology"/>
-	<issn>1477-2019</issn>
+    <issn>1477-2019</issn>
     <eissn>1478-0941</eissn>
     <summary>style for the Journal of Systematic Palaeontology</summary>
     <updated>2016-06-17T01:36:46+00:00</updated>
@@ -32,12 +32,12 @@
     </names>
   </macro>
   <macro name="author">
-    <names variable="author">
+    <names variable="author" font-weight="bold">
       <name name-as-sort-order="all" and="symbol" sort-separator=", " initialize-with=". " delimiter-precedes-last="never" delimiter=", ">
-        <name-part name="family" font-weight="bold"/>
-        <name-part name="given" font-weight="bold"/>
+        <name-part name="family"/>
+        <name-part name="given"/>
       </name>
-      <et-al font-weight="bold"/>
+      <et-al/>
       <label form="short" prefix=" (" suffix=")"/>
       <substitute>
         <names variable="editor"/>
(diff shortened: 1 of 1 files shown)


journal-systematic-palaeontology.csl
```diff
@@ -0,0 +1,238 @@
...
+          <text macro="year-date"/>
+        </group>
+        <group delimiter=" ">
+          <label variable="locator" form="short"/>
+          <text variable="locator"/>
+        </group>
+      </group>
+    </layout>
+  </citation>
+  <bibliography hanging-indent="true" et-al-min="8" et-al-use-first="2">
```
adam3smith: These are somewhat odd et al settings and I can't find anything on them in the documentation. Where do you take them from?

journal-systematic-palaeontology.csl
```diff
@@ -0,0 +1,238 @@
...
+    </names>
+  </macro>
+  <macro name="author">
+    <names variable="author">
+      <name name-as-sort-order="all" and="symbol" sort-separator=", " initialize-with=". " delimiter-precedes-last="never" delimiter=", ">
+        <name-part name="family" font-weight="bold"/>
+        <name-part name="given" font-weight="bold"/>
+      </name>
+      <et-al font-weight="bold"/>
+      <label form="short" prefix=" (" suffix=")"/>
```
adam3smith: so the label shouldn't be bold? Again, not in the documentation -- is there an example in an article that you were able to find? Otherwise you could save yourself the trouble of the separate font-weight statements and just bold the whole macro.

adam3smith:
See my two questions inline. Beyond that, could you run this through our formatter to fix indenting?

journal-systematic-palaeontology.csl
nsvitek: I think I just screwed up where to put the semi-colon in the code in order to get the separator correct. Line 156: "; " changed to " ".

journal-systematic-palaeontology.csl
nsvitek: Oh, thanks! Yes, simple bad coding on my part. Will bold the whole macro then run through formatter.

journal-systematic-palaeontology.csl
adam3smith: no, I'm referring to the 8/2 et al settings. It's not impossible, but it's quite rare to show the first _two_ authors when using et al. You sure that's correct?

nsvitek: I see. I hadn't noticed that. No, it's not correct. Removed.

adam3smith:
Thanks!

rmzelle:
(fixed the file name: https://github.com/citation-style-language/styles/commit/96b8a95ef0038fdfef4ffba8cf3941fe9da50395)

PR #2114, opened 2016-06-17
Add class="note" to pontifical-gregorian-university dependents

PUG-IS:
add class="note", see [https://github.com/citation-style-language/styles/pull/2103](2103)

PR #2115, opened 2016-06-17
Update pontifical-gregorian-university-en.csl

PUG-IS:
add class="note", see [https://github.com/citation-style-language/styles/pull/2103](2103)


rmzelle:
Will be covered by #2114.

PR #2121, opened 2016-06-21

roaldfrosig:
Cleaned up the code, and added descriptions to the macros.


adam3smith:
thanks!

PR #2127, opened 2016-06-26
Create scienceasia.csl

asaiz:
This is the citation style for ScienceAsia (http://www.scienceasia.org). I modified the PNAS style to have the journal volume number in bold, etc. [following the examples given in the detailed instructions for authors, available at http://scienceasia.org/scias_guide.pdf].

### DIFF
--- a/scienceasia.csl
+++ b/scienceasia.csl
@@ -1,0 +1,157 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" page-range-format="minimal" demote-non-dropping-particle="sort-only" default-locale="en-GB" xmlns="http://purl.org/net/xbiblio/csl">
+<!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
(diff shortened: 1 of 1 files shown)


adam3smith:
@rmzelle -- looks good to me; bit tired, so leaving this for a 2nd look by you.

scienceasia.csl
```diff
@@ -0,0 +1,157 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" page-range-format="minimal" demote-non-dropping-particle="sort-only" default-locale="en-GB" xmlns="http://purl.org/net/xbiblio/csl">
+<!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>ScienceAsia</title>
+    <title-short>ScienceAsia</title-short>
```
rmzelle: `<title-short>ScienceAsia</title-short>` is rather redundant, since it's the same as the title. Not a real problem, but it could be deleted.

rmzelle:
Yeah, looks fine, thanks!

PR #2129, opened 2016-06-27
Add "original-*" variables

xavi-mat:
- Add "original-*" variables.
- Delete "note" variable (to allow the use of the workaround {:original-*:TEXT} )
- Update documentation.


teologia-catalunya.csl
```diff
@@ -468,8 +469,15 @@
           </group>
         </else>
       </choose>
+      <group delimiter=", " prefix=" (=" suffix=")">
```
rmzelle: `prefix=" (="`

Is the equal sign here correct?

xavi-mat: Yes, it is. To avoid confussion with the series.

adam3smith:
looks great, thanks.

PR #2130, opened 2016-06-28
Create journal-of-clinical-sleep-medicine.csl

libora:
Zotero forum thread is here: https://forums.zotero.org/discussion/60275/style-request-journal-of-clinical-sleep-medicine/


adam3smith:
Thanks!

PR #2131, opened 2016-06-28

agruber:
Apply title case transformation to container-title in the else conditional that is called for article-journal. It is done so already for article-newspaper and article-magazine.

### DIFF
--- a/american-sociological-association.csl
+++ b/american-sociological-association.csl
@@ -210,7 +210,7 @@
             <text macro="editor" prefix=" "/>
           </group>
           <group prefix=" " suffix="." delimiter=" ">
-            <text variable="container-title" font-style="italic"/>
+            <text variable="container-title" font-style="italic" text-case="title"/>
             <group delimiter=":">
               <group>
                 <text variable="volume"/>


adam3smith:
I'm a bit worried about doing this for journals where we may get unintended results for non-English journal titles. At the same time, how often does this come up?

agruber:
It is already in place for article-newspaper and article-magazine, so also this will then give an incorrect result. There are two routes to go from here: either delete it from article-newspaper and article-magazine, or add it to article-journal. Leaving it the way it is, is just inconsistent.

agruber:
I have counted the number of CSL files that match the pattern 'variable="container-title".+text-case="title"'. Looks like there are 344 files with this expression. So that is a more general problem and needs to be address at a higher level. That could e.g. be done by checking the language field.

adam3smith:
CSL already checks the language field, which is why I don't consider it a problem for book titles and, for the most part, newspapers and magazines. The problem is that for journals, in particular, the language of the work doesn't necessarily match that of the journal title (historical German and Latin titles now publishing in English are probably the biggest group).

agruber:
On the other way it will we wrong for all English journal names, which is by far the largest group.

rmzelle:
> On the other way it will be wrong for all English journal names

Well, that's only an issue when journal titles are (improperly) stored in lowercase in the metadata, right? In my experience with Zotero, I rarely encounter lowercased titles. 

But I don't have a strong opinion either way (although I'm not a fan of title-casing, which is inherently lossy, and I don't think it should be encouraged; maybe the US/UK will get rid of it when they switch over to the metric system).

adam3smith:
> n journal titles are (improperly) stored in lowercase in the metadata, right?

That's actually the core of the problem here: I used to think this was incorrect, but have since been told that technically speaking (i.e. by international cataloging standards), journal titles should be in sentence case and that's what Pubmed does. I believe we do title case journals for APA for that reason and so I think we might have to go that way, though I'm not liking it very much (obviously)

rmzelle:
So we should title-case in the style, then?

adam3smith:
yeah -- let's. I'd still have been curious where the sentence cased journal titles came from, in particular for a sociology style, but the number of problematic journals in that field is equally small.

PR #2132, opened 2016-06-28
Update statistika-statistics-and-economy-journal.csl

libora:
Apply title case transformation to title and container-title.


adam3smith:
thanks!

PR #2134, opened 2016-06-29
SBL 2nd Ed. - Series title should always be short

jackweinbender:
Series titles, like journal titles, should be short forms in the note as well as the bibliography.

### DIFF
--- a/society-of-biblical-literature-fullnote-bibliography.csl
+++ b/society-of-biblical-literature-fullnote-bibliography.csl
@@ -370,7 +370,7 @@
     </choose>
   </macro>
   <macro name="collection-title">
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks. Note that, afaik, this will not do anything in most reference managers, since they don't have a series abbreviation function/variable, but it also doesn't do any harm there. I'd guess that pandoc is the only product that'd currently do this.

jackweinbender:
Yeah, I've been working to make my own abbreviations file so that I can use the full form in Zotero and have uniform short forms in my bibliography (and yes, this is with pandoc). I always used to have Series and Shortseries fields in my .bib files—which worked nicely with biber/biblatex.

PR #2135, opened 2016-06-29
SBL 2nd Ed. - Series title should be title case

jackweinbender:
Another (somewhat) theoretical suggestion. In those cases where the full series title is used (see #2134 ), it should be in title-case.


rmzelle:
@adam3smith, will you tend to this? (looks fine to me, but I didn't check the style guide)

adam3smith:
Thanks!

PR #2136, opened 2016-06-30
acta paediatrica style

mcshaz:
currently down as Vancouver, but this is not correct (no brackets,
superscript, year only and no issue number). 

The dependent style as described above has been rebased on the Vancouver style, with the changes described (no brackets, superscript, year only, no issue number) as the vancouver.css file provided seemed more extensive and encompassing (cutural date support etc.)


adam3smith:
Thanks, that makes sense. I'll need a bit of time to review to make sure that the modified style (which has multiple dependents) performs as expected, but I generally agree that using the general-purpose Vancouver style as a basis for that makes a lot of sense.

mcshaz:
Thank you for taking the time to have a look. 

In modern development, if I were to submit code to an open source coding project (as opposed to an XML library such as this), it would have to have passed unit tests. It would be a fantastic feature for a project like this.

Almost all journals include 3 or 4 examples on the instruction for authors. In an ideal world, contributors would also pass in the records for the 3 or 4 examples (format would have to be agreed - JSON, XML etc), and the output directly pasted from the instructions for authors. All these examples would then be run every time a pull request was received, to ensure no dependents were 'broken' by any change.

Please note also I have subsequently changed the acta paediatrica dependent to referring to vancouver-sage, as I had missed the italics required by this journal.

adam3smith:
we obviously know about unit tests, and @fbennett developed a framework for this years ago, but we have more than 1300 different CSL styles. It's just not feasible for us to create a test library of that size, especially since it's not as simple as just running a single set of test JSON corresponding to examples. There are settings that don't show up in examples (sort, et al.), complicated cases such as missing fields, etc. So realistically, you'd be looking at a minimum of 10, but for more complex styles like Vancouver more like 30+ sets of tests. I just don't see that as happening.

adam3smith:
Changing the Sage Vancouver style definitely is a no go, though. That style is purposefully and in several significant ways different from regular Vancouver (spacing, et al rules, abbreviations for things like editors, display of online sources, page-range display) and it follows a different style guide.

mcshaz:
I will follow up the unit test framework written by @fbennett - thanks for the link. I was actually thinking prospectively rather than retrospectively generating tests cases, and in this case it is only for about 4 journals, so definitely doable for me as a potential contributor, even if I only provide some proof that the test cases had passed on my machine using the framework.

To be clear, I have **not touched** vancouver-sage, I have changed the parent for acta-paediatrica from vancouver to vancouver-sage

adam3smith:
oh sorry, I get it -- you left the changes to the superscript vancouver in place. 
But Sage Vancouver doesn't appear right -- just going quickly, it doesn't shorten page ranges it does 4/3 et al. instead of 7/6 it does (eds) for editors instead of , eds. -- those are the things I can tell at one look.

rmzelle:
> In modern development, if I were to submit code to an open source coding project (as opposed to an XML library such as this), it would have to have passed unit tests. It would be a fantastic feature for a project like this.
> 
> Almost all journals include 3 or 4 examples on the instruction for authors. In an ideal world, contributors would also pass in the records for the 3 or 4 examples (format would have to be agreed - JSON, XML etc), and the output directly pasted from the instructions for authors. All these examples would then be run every time a pull request was received, to ensure no dependents were 'broken' by any change.

I think @mz2 (from the Manuscripts app) also did some work in this direction (storing the guidelines from each journal in a standardized format), but I'm not sure if he made enough progress that we could start considering automatic testing of CSL styles based on published journal guidelines.

mcshaz:
> But Sage Vancouver doesn't appear right -- just going quickly, it doesn't shorten page ranges it does 4/3 et al. instead of 7/6 it does (eds) for editors instead of , eds. -- those are the things I can tell at one look.

very astute - I guess that is what makes you an admin on this repo

the only thing I have to work on for Acta Paediatrica is the examples from the web page with instructions for authors (frustratingly, there is no description or reference to any standard they would like references to comply to):   
> 1. Kühl C, Andersen GE, Hertel J, Mölsted-Pedersen L. Metabolic events in infants of diabetic mothers during the first 24 hours after birth. _Acta Paediatr_ 1982; 71:19-25.
> 2. Feigin RD. Bacterial meningitis beyond the neonatal period. In: Feigin RD, Cherry JD, eds. _Textbook of pediatric infectious diseases._ 2nd ed. Philadelphia: WB Saunders, 1987.
> 3. Jones G. Textbook of paediatrics. Uppsala: _Almqvist & Wiksell_, 1974: 193-9.
> 4. D'Hondt E, Berge E, Colinet G. Production and quality control of the Oka strain live varicella vaccine. _Postgrad Med J_ 1985; 61 Suppl 4:53-6.
> For a journal article in electronic format use the following style:
> 5. Morse SS. Factors in the emergence of infectious diseases. Emerg Infect Dis [serial online] 1995 Jan-Mar [cited 1996 Jun 5]; 1(1): [24 screens]. Available from: URL: www.cdc.gov\ncidod\EID\eid.htm

thus you are spot on - I need short page ranges, and eds. without brackets. There is also inconsistency int he published examples with book (2) having italics on the title and book (3) italic on the publisher. I haven't even tried to get (5) working, but it looks like I should:
-create an individual .csl file for acta paediatrica rather than using a parent
-try and find a formal description of citation style - it is frustrating the instructions for authors are so vague

mz2:
@rmzelle the computer readable guidelines we have been building for Manuscripts (still hoping to open them up) are not really trying to add extra validity checks on top of what CSL provide, it's more that for the specific case of citation style description that is one of the many parts in publication guidelins, the Manuscripts way of describing the required citation style is a reference to a CSL identifier. So, no help from us for this task am afraid…

adam3smith:
> create an individual .csl file for acta paediatrica rather than using a parent
unfortunately, yes

> -try and find a formal description of citation style - it is frustrating the instructions for authors are so vague

you won't. The best you can do is take everything you can infer from these examples and base the rest of decisions on Citing Medicine (i.e. Vancouver), on which it is based: http://www.ncbi.nlm.nih.gov/books/NBK7256/. This is quite normal for journal styles guides, btw. Very few give you more detail than that. And virtually none published by Wiley.

rmzelle:
> the Manuscripts way of describing the required citation style is a reference to a CSL identifier.

Ah, okay. That might still be interesting to us to catch CSL styles that use the wrong format for a journal (or identify journals for which we can generate new dependent CSL styles), but yeah, that wouldn't help with validating the format itself.

PR #2137, opened 2016-07-01
Create apa-fr-hepfr.csl

wookie909:


adam3smith:
and none of the existing French APA versions will do? This would be the third one...

wookie909:
Well, the existing styles apply to French Canada, which is a rather
different context than Switzerland (I have made quite a few modifications).
But of course the final decision is yours.

adam3smith:
If it's a different style, we'll take it. I was just making sure that's the case 
@rmzelle -- thoughts on naming? The style guide says it's "inspired by" APA, but it does follow APA very closely. On the other hand, the barrage of APA styles we show users may be getting overwhelming.

rmzelle:
Yeah, it might be better to rename this to:

```xml
<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" default-locale="fr-FR">
  <info>
    <title>Haute Ecole pédagogique Fribourg (French)</title>
    <title-short>HEPFR</title-short>
    <id>http://www.zotero.org/styles/haute-ecole-pedagogique-fribourg</id>
    <link href="http://www.zotero.org/styles/haute-ecole-pedagogique-fribourg" rel="self"/>
    <link href="http://www.zotero.org/styles/apa-old-doi-prefix" rel="template"/>
    <link href="https://www.hepfr.ch/sites/default/files/guide_apa_hepfr.pdf" rel="documentation"/>
    <author>
      <name>Frederic Spycher</name>
      <email>spycher.frederic@gmail.com</email>
      <uri>http://lib.fyi</uri>
    </author>
    <category citation-format="author-date"/>
    <updated>2016-07-01T16:14:00+00:00</updated>
  </info>
```

(changed "default-locale", title, short title, "self" link, style ID, and category)

and change the file name to "haute-ecole-pedagogique-fribourg.csl".

rmzelle:
Although, does your institution have style guides for other citation formats as well? If so, we might want to clarify that this is the APA variant. (quite a few universities have guides for both numeric ["vancouver"] and author-date ["harvard"] citation formats)

wookie909:
We only use APA, but we also have a German variant for German-speaking
students and faculty, as we live in a bilingual region (we don't have a CSL
for that yet, though. Although maybe we could try to cover both locale in
one file; I'll have to investigate that).

I agree to the changes that you proposed.

rmzelle:
> We only use APA, but we also have a German variant for German-speaking students and faculty, as we live in a bilingual region (we don't have a CSL for that yet, though. Although maybe we could try to cover both locale in one file; I'll have to investigate that).

Okay, in that case we'll accept the style with my proposed changes above. If you can cover both French and German with a single CSL style, you could keep the current version as is and create a German "dependent" CSL style (which would just act as an alias to the French version, but with a German locale). Or you could delete the "default-locale" and have a single style without a fixed locale.

rmzelle:
Perfect, thanks.

PR #2139, opened 2016-07-05
Two small corrections to the AGLC style

tabaxter:
Users of the style at Melbourne Law School have requested that the issue number be available for journal articles. Per rule 4.4 of AGLC this should be in the format of Vol(Issue). 

As well, short titles were appearing with an extra comma due to the delimiter on the group the macro was in. This shifts short title outside the group to fix that.

E.g.: 
"Overseas Tankship (UK) Ltd v Morts Dock and Engineering Co Ltd [1961] AC 388, 390, (‘Wagon Mound [No. 1]’)."
was displaying instead of
"Overseas Tankship (UK) Ltd v Morts Dock and Engineering Co Ltd [1961] AC 388, 390 (‘Wagon Mound [No. 1]’)."

### DIFF
--- a/australian-guide-to-legal-citation.csl
+++ b/australian-guide-to-legal-citation.csl
@@ -32,7 +32,7 @@
     <category citation-format="note"/>
     <category field="law"/>
     <summary>A modification of the Bluebook legal citation style for Australian conditions.</summary>
-    <updated>2015-11-10T00:40:00+00:00</updated>
+    <updated>2016-07-05T05:38:10+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale>
@@ -361,7 +361,10 @@
     <choose>
       <if type="article-journal article-magazine article-newspaper legal_case" match="any">
         <group delimiter=" ">
-          <text variable="volume"/>
+          <group>
+            <text variable="volume"/>
+            <text variable="issue" prefix="(" suffix=")"/>
+          </group>
           <choose>
             <if type="legal_case">
               <choose>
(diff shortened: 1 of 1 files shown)


adam3smith:
(@rmzelle - don't merge yet; want to double check the changes don't affect other item types)

tabaxter:
@adam3smith Thanks for getting on to it so fast! Let me know if I've stuffed anything up.

adam3smith:
I'm good with the 2nd change. On the first change, that's always a trade-off, because the style requires leaving out the issue number where journals are continuously paginated. I thought I understood that was the overwhelming norm for relevant journals?

tabaxter:
The norm is definitely that volumes are consecutively paginated. The problem with the current version of the csl is that it doesn't allow for them at all. By my short assessment (and if there's a standard policy you guys follow in these instances, I'm happy to go with that instead), it is easier to delete the issue numbers where they aren't relevant than to insert them where they are. Either the author or the editor can delete without leaving the screen. To insert where they are requires more effort.

My personal approach to AGLC (because I write almost exclusively in that format and occasionally train people in using Zotero for it) is to omit issue numbers from the individual entry, but i get that's not going to be the preferred approach.

adam3smith:
> By my short assessment (and if there's a standard policy you guys follow in these instances, I'm happy to go with that instead), it is easier to delete the issue numbers where they aren't relevant than to insert them where they are.

That's our general approach, too. It's only when individually paginated journals are virtually absent in a field that we remove the issue numbers.

> My personal approach to AGLC (because I write almost exclusively in that format and occasionally train people in using Zotero for it) is to omit issue numbers from the individual entry, but i get that's not going to be the preferred approach.

yes, obviously we'd like to discourage that as a general approach, even though it makes sense in your case.
At least a theoretical solution for a fix on this is on our long-term agenda, but as you can imagine it's pretty hard to get this right because there's no easy way of telling which journals are continuously paginated and which aren't.

adam3smith:
Given discussion above, merging this. Thanks!

PR #2141, opened 2016-07-05
journal-of-materials-research.csl

xyz135:
No style for Journal of Materials Research exists at this point. I have created this new style following the instructions at:

[http://www.mrs.org/jmr-format/](url)

The style is largely based on the "american-physics-society.csl" file but with the necessary changes, e.g.,

1. The delimiter after the last author's name is ":".
2. The full title of the journal is added which ends with ".".
3. The journal name is abbreviated and appears in italics.
4. The volume number is in bold followed by the issue number in parenthesis.
5. Only the first page number appears.
6. The year appears, in parenthesis, at the end.

### DIFF
--- a/journal-of-materials-research.csl
+++ b/journal-of-materials-research.csl
@@ -2,19 +2,19 @@
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
   <info>
     <title>Journal of Materials Research</title>
-	<title-short>J. Mater. Res.</title-short>
+    <title-short>J. Mater. Res.</title-short>
     <id>http://www.zotero.org/styles/journal-of-materials-research</id>
-	<link href="http://www.zotero.org/styles/journal-of-materials-research" rel="self"/>
-	<link href="http://www.zotero.org/styles/american-physics-society" rel="template"/>
-	<link href="http://www.mrs.org/jmr-format" rel="documentation"/>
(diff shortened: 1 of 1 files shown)


xyz135:
Great!

xyz135:
Made minor changes in <info> section and the section for book etc. -- the book title etc. should not be italicized; so that was updated. It looks good to me now.

journal-of-materials-research.csl
```diff
@@ -0,0 +1,192 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>Journal of Materials Research</title>
+	<title-short>J. Mater. Res.</title-short>
+    <id>http://www.zotero.org/styles/journal-of-materials-research</id>
+	<link href="http://www.zotero.org/styles/journal-of-materials-research" rel="self"/>
```
adam3smith: add `<link href="http://www.zotero.org/styles/american-physics-society" rel="template"/>`
here please

adam3smith:
could you run the style through http://formatter.citationstyles.org/ to fix indenting. With that and the template link, this looks good to me. Thanks!

xyz135:
@adam3smith: added the template link; will do the check right away! thanks.

adam3smith:
Thanks!

xyz135:
Thanks!

PR #2142, opened 2016-07-07
Create south-african-journal-of-geology.csl

hierosarkhos:


adam3smith:
This looks good. Is this based off an existing style? If so, we'd like to have a link to that style with `link="template"`

south-african-journal-of-geology.csl
```diff
@@ -0,0 +1,165 @@
...
+    <eissn>1996-8590</eissn>
+    <summary>Style for the South African Journal of Geology.</summary>
+    <updated>2016-07-07T12:07:56+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="en-GB">
+    <terms>
+      <term name="editor" form="short">
+        <single>Editor</single>
+        <multiple>Editors</multiple>
```
rmzelle: @hierosarkhos, this is slightly odd. It would be more straightforward to just use the default non-abbreviated "long" form. It's also our custom to take define terms in their standard (lowercase) form, and take care of capitalization when the term is used.

E.g. you could delete

```xml
  <locale xml:lang="en-GB">
    <terms>
        <single>Editor</single>
        <multiple>Editors</multiple>
      </term>
    </terms>
</locale>
```

and replace

```xml
  <macro name="editor">
      <name and="text" sort-separator=", " initialize-with="." delimiter-precedes-last="never" delimiter=", "/>
```

with

```xml
  <macro name="editor">
      <name and="text" sort-separator=", " initialize-with="." delimiter-precedes-last="never" delimiter=", "/>
        <label form="short" text-case="capitalize-first" prefix=" (" suffix=")"/>
```

rmzelle: The only reason for doing what you're doing would be if you'd want "long" non-abbreviated terms for "editor", but "short" abbreviated terms for "translator" when editors and translators are substituted in the "author" and "author-short" macros.

south-african-journal-of-geology.csl
```diff
@@ -0,0 +1,165 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+<!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>South African Journal of Geology</title>
+    <title-short>Geol. South Africa</title-short>
+    <id>http://www.zotero.org/styles/south-african-journal-of-geology</id>
+    <link rel="self" href="http://www.zotero.org/styles/south-african-journal-of-geology"/>
```
rmzelle: Can you add a "template" link? See https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link

hierosarkhos: @rmzelle & @adam3smith, I had started with meteorological-applications.csl as a base and intended to only make minor edits but soon found that the "minor" where more pervasive that originally anticipated. I then borrowed from the-geological-society-of-america.csl, the-geological-society-of-london.csl, APA,csl and harvard-anglia-ruskin-university.csl. By the time I was finished I wasn't sure where some the code came from and how much of the base was left.

south-african-journal-of-geology.csl
hierosarkhos: @rmzelle, Thank you for the guidance. I'm going to implement the suggestion. I've asked around and wasn't able to find anyone who has seen the term "translator" used, in either the short or long form, in any of the references that they are aware of.

rmzelle:
Great, thanks!

south-african-journal-of-geology.csl
rmzelle: Okay, that's fine, then.

PR #2145, opened 2016-07-13
Create geochronometria.csl

dbobak:
Citation style for journal "Geochronometria"
[http://www.geochronometria.pl/](http://www.geochronometria.pl/)
[http://www.degruyter.com/view/j/geochr](http://www.degruyter.com/view/j/geochr)


rmzelle:
Great, thanks!

PR #2153, opened 2016-07-21
Update Elementa style with numerous improvements

akokai:
Fixed several details for a more complete match to the journal style guide.

- Journal titles and issue numbers (important!)
- Collection/series titles and numbers
- Missing dates
- Book Title Case
- other minor things


rmzelle:
Thanks!

PR #2155, opened 2016-07-23
Update Elementa style

akokai:
Improve bibliography layout for Elementa.
- Fix weird layout artifact for reports.
- Add event fields for conference papers.

Apologies for the frequent PRs. I keep finding more things to fix as I write.

### DIFF
--- a/elementa.csl
+++ b/elementa.csl
@@ -15,7 +15,7 @@
     <category citation-format="author-date"/>
     <category field="science"/>
     <summary>Based on The Council of Science Editors style, Name-Year system: author-date in text, sorted in alphabetical order by author.</summary>
-    <updated>2016-07-16T08:00:00+00:00</updated>
+    <updated>2016-07-23T08:00:00+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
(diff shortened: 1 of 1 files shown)


akokai:
Canceling this because I realize there's a lot of work left to do on the legal citation style, which also extends into the CSE-author-date style that this was based on. I will submit new PRs for each style when I'm done.

rmzelle:
Okay, thanks.

(does CSE need to support legal items? It's scope it more the natural sciences, no?)

akokai:
Yeah, I might just leave CSE the way it is because I have a feeling I will not get it right, and more people probably use it. The CSE style guide is non-free, so I don't know for sure, but other sources say CSE should use Bluebook style for legal citations (like APA). Elementa definitely needs something like that because it's an interdisciplinary journal.

rmzelle:
> Elementa definitely needs something like that because it's an interdisciplinary journal.

I don't see any legal types listed at https://home.elementascience.org/for-authors/style-guide/, though.

In general, CSL is somewhat limited in its support of legal materials, and we don't fully support the requirements of Bluebook. If you need extensive support for legal items, your best option is to use Juris-M (a fork of Zotero; https://juris-m.github.io/) in combination with its Juris-M styles (forked from CSL).

akokai:
Thanks for the tips. Far from needing extensive legal citation support, I just want it not to output something nonsensical -- if anything, for my own use, but why not share it with the community. What I've done is copied the parts of APA that deal with legal citations. Will set up a new PR.

rmzelle:
(replaced by #2160)

PR #2154, opened 2016-07-21
Date inconsistencies

jmiba:
Use of macro year-date instead of date parts

PR #2156, opened 2016-07-25
Date inconsistencies amended

jmiba:
Use of macro year-date instead of date parts


rmzelle:
Much better, thanks.

rmzelle:
(replaces #2154)

PR #2160, opened 2016-07-26
Rewritten Elementa style

akokai:
- Reorganized the macros to parallel how `apa.csl` is organized (but with Elementa style specifications). Improved readability of the CSL document, in my opinion.
- Added the legal citation handling that's in APA. 
- Checked against the journal guidelines with many citation types...


rmzelle:
Great, thanks for all your work on this!

PR #2163, opened 2016-07-28
Update cell-numeric.csl

ms609:
Having just had a manuscript back with edits from Current Biology, I have identified (and implemented) the following changes as necessary:
- Author initials should be separated by period but no space (A.B., not A. B. )
- Et al in bibliography author list should be italicised
- Citations in text should be delimited by comma, not comma-space
[and Update metadata]


adam3smith:
perfect, thanks

PR #2164, opened 2016-07-28
Create apa-6-doi-no-issue.csl

pntfx4:
Based upon American Psychological Association 6th edition ("doi:" DOI prefix)(2016-05-25 06:41:20) but revised to remove journal issue and correctly style conference proceedings.

### DIFF
--- a/apa-6-doi-no-issue.csl
+++ b/apa-6-doi-no-issue.csl
@@ -1,0 +1,624 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
+  <info>
+    <title>American Psychological Association 6th edition ("doi:" DOI prefix, no issue numbers)</title>
+    <title-short>APA</title-short>
+    <id>http://www.zotero.org/styles/apa-6-doi-no-issue</id>
+    <link href="http://www.zotero.org/styles/apa-6-doi-no-issue" rel="self"/>
+    <link href="http://owl.english.purdue.edu/owl/resource/560/01/" rel="documentation"/>
+    <author>
+      <name>Simon Kornblith</name>
+      <email>simon@simonster.com</email>
+    </author>
+    <contributor>
+      <name>Bruce D'Arcus</name>
+    </contributor>
+    <contributor>
+      <name>Curtis M. Humphrey</name>
+    </contributor>
+    <contributor>
+      <name>Richard Karnesky</name>
+      <email>karnesky+zotero@gmail.com</email>
+      <uri>http://arc.nucapt.northwestern.edu/Richard_Karnesky</uri>
+    </contributor>
+    <contributor>
+      <name>Sebastian Karcher</name>
+    </contributor>
+    <category citation-format="author-date"/>
+    <category field="psychology"/>
+    <category field="generic-base"/>
+    <summary>APA 6th edition, but with DOIs and without issue numbers</summary>
+    <updated>2016-05-25T06:41:20+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="en">
+    <terms>
+      <term name="editortranslator" form="short">
+        <single>ed. &amp; trans.</single>
+        <multiple>eds. &amp; trans.</multiple>
+      </term>
+      <term name="translator" form="short">
+        <single>trans.</single>
+        <multiple>trans.</multiple>
+      </term>
+    </terms>
+  </locale>
+  <macro name="container-contributors">
+    <choose>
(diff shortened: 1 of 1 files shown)


adam3smith:
what's this for specifically? We would very much like to avoid the proliferation of APA styles. It becomes impossible to maintain -- it kind of is already.

pntfx4:
This corrects the currently posted styles. Specifically, per the APA publication manual:

Journal issue numbers should not be included. - this is a problem with all currently posted APA styles

Titles of conference proceedings should not be proceeded with "In". - this is a problem with American Psychological Association 6th edition ("doi:" DOI prefix)(2016-05-25 06:41:20)

rmzelle:
There is a discussion on the rationale to have the CSL APA styles include issue numbers at https://forums.zotero.org/discussion/32375/apa-issue-number/, which I think is still up to date.

In short, issue numbers are necessary for journal articles that restart numbering for each issue, and there is no easy way for software to know which journals do or don't. It's arguably better to include information that may be redundant than omit information that may be crucial.

pntfx4:
Thank you for posting the discussion. While I recognize the arguments made within, the benefits of an open-source citation approach is that it is easy to fork the style and allow the user to make the decision between including the issue or not. In this view, the submitted style fixes the issue regarding conference proceedings while still correctly providing DOI information.

rmzelle:
> the benefits of an open-source citation approach is that it is easy to fork the style and allow the user to make the decision between including the issue or not.

Well, you're of course more than welcome to modify a CSL style from the repository for personal use. But @adam3smith and I, as curators of the CSL styles repository, think there is a real cost to adding even more APA variants to the repository, so we're always hesitant to accept submissions like this one.

> the submitted style fixes the issue regarding conference proceedings

We'd be happy to accept any corrections to the existing APA CSL styles in the repository.

pntfx4:
I completely understand the hesitancy. I agree that there are too many APA styles listed. If your approach as curators is to restrict styles though, for transparency sake then it would seem that the following styles should be removed:

American Psychological Association 6th edition (single-spaced bibliography)(2015-03-19 02:36:13)
since this if a formatting edit that can easily be done in a word processor without any changes to the reference tags.

American Psychological Association 6th edition (no DOIs, no issue numbers)(2015-03-19 02:36:13)
since this is an inaccurate variant of the APA style which says that if DOIs are available they should be included.

American Psychological Association 6th edition (no ampersand)(2015-08-09 02:13:24)
since this is an inaccurate variant of the APA style

American Psychological Association 6th edition (curriculum vitae, sorted by descending date)(2015-03-19 02:36:13)
since this is an inaccurate variant of the APA style

American Psychological Association 6th edition(2016-05-25 22:36:10)
since this is an inaccurate variant of the APA style as it leaves out DOI information.

rmzelle:
> If your approach as curators is to restrict styles though, for transparency sake then it would seem that the following styles should be removed:

Maybe. Although we would probably get complaints it we removed any. FYI, the rationale for adding these styles:

> American Psychological Association 6th edition (single-spaced bibliography)(2015-03-19 02:36:13)
> since this if a formatting edit that can easily be done in a word processor without any changes to the reference tags.

See https://github.com/citation-style-language/styles/pull/1017#issuecomment-48116370

> American Psychological Association 6th edition (no DOIs, no issue numbers)(2015-03-19 02:36:13)
> since this is an inaccurate variant of the APA style which says that if DOIs are available they should be included.

Added because journals like Organizational Study require this variant. See https://github.com/citation-style-language/styles/commit/d5a5a441ae50aaeb0e7456c96a99151c81aaa8eb

> American Psychological Association 6th edition (no ampersand)(2015-08-09 02:13:24)
> since this is an inaccurate variant of the APA style

Added to improve localization for Spanish/French. See https://github.com/citation-style-language/styles/pull/1552

> American Psychological Association 6th edition (curriculum vitae, sorted by descending date)(2015-03-19 02:36:13)
> since this is an inaccurate variant of the APA style

Added as our only CV style (with items sorted descending by date). See https://github.com/citation-style-language/styles/commit/141ff52d366c5071718f440117f668b20088ac63#diff-28177f2f506d4d3d6d0d6c0cfe10d828

> American Psychological Association 6th edition(2016-05-25 22:36:10)
> since this is an inaccurate variant of the APA style as it leaves out DOI information.

Are you referring to apa.csl? That does print DOIs for me.

rmzelle:
Anyway, we try our best to make the repository as useful as possible, and ultimately have to rely on our own judgement. The situation around APA is particularly tricky because it's such a popular style, and has very detailed and changing requirements.

adam3smith:
I checked the remaining issue here -- the "In" for conference proceedings. The relevant example in the APA manual is 39, conference published in book form, which does have the "In". Example 38, proceedings published regularly, is from PNAS, which should just be (and is in all ref managers I'm aware of) treated like any other journal.
The DOI issue is simply a misunderstanding. The current standard APA style does include it, but in the now preferred format as a URL (i.e. preceded by http://doi.org/)  See here for more: https://www.zotero.org/support/kb/doi_in_apa
Thanks again for contributing. Closing this

PR #2167, opened 2016-07-29
Update journal-of-systematic-palaeontology.csl

nsvitek:
align citation sorting with style guidelines (http://www.tandf.co.uk/journals/authors/style/reference/ref_tjsp.pdf). Specifically: list multiple publications by same first author chronologically. In code: if same first author and same year, additionally sort by month.

### DIFF
--- a/journal-of-systematic-palaeontology.csl
+++ b/journal-of-systematic-palaeontology.csl
@@ -113,6 +113,18 @@
       </else>
     </choose>
   </macro>
+  <macro name="month-date">
+    <choose>
+      <if variable="issued">
+        <date variable="issued">
+          <date-part name="month"/>
+        </date>
+      </if>
+      <else>
+        <text term="no date" form="short"/>
+      </else>
+    </choose>
+  </macro>
   <macro name="edition">
     <choose>
       <if is-numeric="edition">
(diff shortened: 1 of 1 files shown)


journal-of-systematic-palaeontology.csl
```diff
@@ -171,6 +183,7 @@
       <key macro="author-count" names-min="3" names-use-first="3"/>
       <key macro="author" names-min="3" names-use-first="1"/>
       <key macro="year-date"/>
+     <key macro="month-date"/>
```
rmzelle: It might be more straightforward to use a dedicated macro for year/month sorting, e.g.:

```
  <macro name="year-month">
    <date variable="issued">
      <date-part name="year"/>
      <date-part name="month"/>
    </date>
  </macro>
...
       <key macro="year-month"/>
```

(I don't think you need the "no date" bit for sorting)

adam3smith: I actually disagree with this as a reading of the styleguide: it says nothing about what level "chronological" refers to. I think it makes little sense (and looks completely random) to sort by parts of the date that aren't actually displayed anywhere, which is why I typically use year-date in sort macros. 
If there's any indication that the style actually means "chronological" quite literally, we should just use <key variable="issued"/>

I'd be inclined to close this

rmzelle: Ah, yes, good catch.

@nsvitek, I agree with @adam3smith that the style guide doesn't explicitly require sorting by the full date, and it's quite likely that the current sorting is correct.

rmzelle: (if you hear differently from the journal editors we'd of course be happy to change this)

PR #2169, opened 2016-08-01
MLA8 Vol prefix is added regardless of if volume variable is set

lightman76:
Found that the MLA 8 format was always printing the vol.  prefix even when no volume was specified.  The diff doesn't look clean but I just wrapped the handling of the volume code in an choose if block to test if the volume was set.

### DIFF
--- a/modern-language-association-8th-edition.csl
+++ b/modern-language-association-8th-edition.csl
@@ -122,24 +122,28 @@
   </macro>
   <macro name="number">
     <group delimiter=", ">
-      <group delimiter=" ">
+      <group>
         <choose>
-          <if variable="volume" match="any">
-            <choose>
-              <!--lowercase if we have a preceding element-->
-              <if variable="edition container-title" match="any">
-                <text term="volume" form="short"/>
-              </if>
-              <else-if variable="author editor" match="all">
-                <!--other contributors preceding the volume-->
-                <text term="volume" form="short"/>
-              </else-if>
-              <else>
-                <text term="volume" form="short" text-case="capitalize-first"/>
-              </else>
-            </choose>
-            <text variable="volume"/>
+          <!--lowercase if we have a preceding element-->
+          <if variable="edition container-title" match="any">
+            <group delimiter=" ">
+              <text term="volume" form="short"/>
+              <text variable="volume"/>
+            </group>
           </if>
+          <!--other contributors preceding the volume-->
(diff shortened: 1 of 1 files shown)


adam3smith:
@rmzelle -- not merging anything while travelling, but unless you have a better idea how to do this with a group (and the necessary conditionals for uppercase&lowercase) this lgtm.

rmzelle:
I guess the following would work as well:

```
  <macro name="number">
    <group delimiter=", ">
      <group delimiter=" ">
        <choose>
          <!--lowercase if we have a preceding element-->
          <if variable="edition container-title" match="any">
            <group>
              <text variable="volume"/>
            </group>
          </if>
          <!--other contributors preceding the volume-->
          <else-if variable="author editor" match="all">
            <group>
              <text variable="volume"/>
            </group>
          </else-if>
          <else>
            <group>
              <text term="volume" form="short" text-case="capitalize-first"/>
              <text variable="volume"/>
            </group>
          </else>
        </choose>
      </group>
```

Not sure if that's preferable, though.

lightman76:
@rmzelle - I'm pretty new to the CSL definitions.  I think your version looks cleaner.  
Happy to update the PR to that if that's the more canonical way to handle this type of situation.

rmzelle:
Okay, please update the PR, then. And please test that my version works as expected. I quickly wrote it out by hand and didn't test it. I just noticed my version probably needs delimiters on the inner groups:

```
  <macro name="number">
    <group delimiter=", ">
      <group>
        <choose>
          <!--lowercase if we have a preceding element-->
          <if variable="edition container-title" match="any">
            <group delimiter=" ">
              <text variable="volume"/>
            </group>
          </if>
          <!--other contributors preceding the volume-->
          <else-if variable="author editor" match="all">
            <group delimiter=" ">
              <text variable="volume"/>
            </group>
          </else-if>
          <else>
            <group delimiter=" ">
              <text term="volume" form="short" text-case="capitalize-first"/>
              <text variable="volume"/>
            </group>
          </else>
        </choose>
      </group>
```

(the `<group>` and `</group>` on the third and last lines might also be unnecessary, but I'm not a 100% sure)

lightman76:
Pull request has been updated.  Please let me know if any other changes are needed.

adam3smith:
great -- thanks for spotting & fixing.

PR #2170, opened 2016-08-02
Minor update and cleanup

jmiba:
Periods should not be stripped. Unnecessary attributes deleted.


rmzelle:
(superseded by #2171)

@jmiba, we recommend to always start from the latest version of the style (or entire repository, if you are working on the desktop with a cloned repository), so you don't get PRs like this one.

PR #2172, opened 2016-08-04
Replace "complete-all" with "partial-each"  in bibliography

meiradarocha:
To conform with Technical Standards Brazilian Association ABNT NBR-6023-2002, p. 21, item 9.1.2.


rmzelle:
Thanks!

Issue #2173, opened 2016-08-05
ACM Sigchi 2016 conference paper format error


wbt:
This bug is filed on [acm-sigchi-proceedings.csl](https://github.com/citation-style-language/styles/blob/master/acm-sigchi-proceedings.csl). 
Sample 7 in the [SIGHCHI Proceedings format](http://cscw.acm.org/2016/submit/sigchi-word-proceedings.zip) is:

> Scott R. Klemmer, Michael Thomsen, Ethan Phelps-Goodman, Robert Lee, and James A. Landay. 2002. Where do web sites come from?: capturing and interacting with design history. In _Proceedings of the SIGCHI Conference on Human Factors in Computing Systems_ (CHI '02), 1-8. http://doi.acm.org/10.1145/503376.503378

Visiting that URL, capturing the information with the Zotero translator, and citing it in that style produces:

> Scott R. Klemmer, Michael Thomsen, Ethan Phelps-Goodman, Robert Lee, and James A. Landay. 2002. Where Do Web Sites Come from?: Capturing and Interacting with Design History. _Proceedings of the SIGCHI Conference on Human Factors in Computing Systems_, ACM, 1–8. http://doi.org/10.1145/503376.503378

Differences: 
- Title casing on the article title appears more correct than the "official"
- The non-italicized "In" before the italicized "Proceedings" is absent
- The non-italicized short version of the conference name "(CHI '02)" is absent, even if copied into the Conference Name or Volume field.
- The publisher is listed

I prefer the "buggy" version produced by the first (and last) differences, but on the others would prefer that the citation style file match the official sample.

rmzelle:
@adam3smith, can you give my PR a quick look-over? There was a bug in the `<choose/>` block (it tested twice for "paper-conference").

@wbt, I have a fix ready that should address the last three issues (the conference name should be in the "Series" field). For the first issue, Zotero/CSL don't attempt to automatically convert 'Title Case' to 'Sentence case'. See https://www.zotero.org/support/kb/sentence_casing for more explanation. (however, you can right-click titles of items in your Zotero library to quickly convert them to sentence case)

wbt:
I do have "CHI '02" in the Series field as well as in "Conference Name" and "Volume" but that doesn't solve it.

rmzelle:
I haven't updated the style yet.

adam3smith:
@wbt -- the update style will be available via Zotero within ~1h.

wbt:
[...]

PR #2175, opened 2016-08-08
Update acm-sigchi-proceedings.csl for conference papers

rmzelle:
Addresses #2173


adam3smith:
yes, all good, thanks!

PR #2179, opened 2016-08-11
Added a style for Deutsche Lebensmittel-Rundschau (DLR)

Mailaender:
It was created with http://editor.citationstyles.org taking articles in the journal for reference.

### DIFF
--- a/deutsche-lebensmittel-rundschau.csl
+++ b/deutsche-lebensmittel-rundschau.csl
@@ -1,0 +1,62 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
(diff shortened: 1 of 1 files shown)


rmzelle:
Does DLR take open submissions? Is there a guide for authors online that describes the citation format?

(the CSL repository is primarily for journals with open submissions, see https://github.com/citation-style-language/styles/wiki/Criteria-for-Accepting-Styles)

Mailaender:
Improved the style a bit further. The self description is in German only: http://www.dlr-online.de/mediadaten.html I would classify it as a very local, but open submission and peer-reviewed journal.

adam3smith:
very local is fine and I see it's peer reviewed, but I don't see any way to submit (and thus also no instructions for contributors)? How would one go about submitting to this? (I'm asking both because we do want to make sure putting this in the general repository makes sense and because we really do like to have documentation we and others can refer to if there are any doubts in the future).

Mailaender:
I believe the journal has no guidelines and everyone just uses what he likes so indeed you may be right: this is probably not suitable for inclusion.

PR #2182, opened 2016-08-14
Update Ruby to 2.3

rmzelle:


rmzelle:
Even while Ruby 2.3.0 was released in 2015, it doesn't look like Travis recognizes "2.3" yet, so specifying specific point release "2.3.1" instead for now. See https://github.com/travis-ci/travis-ci/issues/5361#issuecomment-167531389

PR #2183, opened 2016-08-15
Removed excess comma that was affecting patents in the catch-all else section of the bibliography area

marcralew:


rmzelle:
Thanks!

PR #2185, opened 2016-08-19
Add mercatus.csl

elidourado:
House style for the Mercatus Center at George Mason University.

### DIFF
--- a/mercatus-center.csl
+++ b/mercatus-center.csl
@@ -1,10 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="never" page-range-format="chicago">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="never" page-range-format="chicago" default-locale="en-US">
   <info>
-    <title>Mercatus House Style</title>
-    <id>http://www.zotero.org/styles/mercatus</id>
-    <link href="http://www.zotero.org/styles/mercatus" rel="self"/>
+    <title>Mercatus Center</title>
+    <id>http://www.zotero.org/styles/mercatus-center</id>
+    <link href="http://www.zotero.org/styles/mercatus-center" rel="self"/>
     <link href="https://chimsit.sharepoint.com/Mercatus/Mercatus%20Document%20Library/Mercatus-StyleGuide_v2.1.pdf" rel="documentation"/>
+	<link href="http://www.zotero.org/styles/chicago-fullnote-bibliography" rel="template"/>
     <author>
       <name>Eli Dourado</name>
       <email>elidourado@gmail.com</email>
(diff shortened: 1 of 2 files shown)


rmzelle:
Thanks. Is the style guide publicly available?

Could you rename the style to "mercatus-center.csl", and update the style ID and "self" link to match? The title should be "Mercatus Center". Also please add a "template" link (https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link), and, assuming the style is only used in US English, a default-locale of "en-US" (https://github.com/citation-style-language/styles/wiki/Style-Requirements#3---style-locale).

rmzelle:
Oh, and our tests currently fail because the style has a whole bunch on unused macros, which must be removed:


>1) mercatus: may not have any unused macros
>
>Failure/Error: expect(available_macros - used_macros).to eq([])
>
>expected: []
>
>got: ["access", "container-contributors", "container-title", "contributors", "contributors-sort", "description", "issue-join-with-comma", "issue-join-with-period", "issue-join-with-space", "locators-chapter", "locators-join-with-comma", "locators-join-with-period", "locators-join-with-space", "locators-journal-join-with-colon", "locators-journal-join-with-comma", "secondary-contributors"]

elidourado:
The style guide is not publicly available.

I have renamed, updated urls, added default locale, etc., and removed unused macros.

adam3smith:
We're generally a reluctant to host house styles unless either
a) They're for a massive organization like a UN suborganization or
b) At least some of the publications using the style have open(ish) submission

Is b) the case here?

elidourado:
The Mercatus Center is a nonprofit research center at George Mason University that publishes hundreds of papers and policy briefs per year. It doesn't exactly have open submissions, but it does publish work from researchers around the United States and the world who do not necessarily have any other affiliation with Mercatus. It would be extremely useful to have the style in the repository for bundling with citation managers so that authors could more easily follow the appropriate citation style guidelines.

adam3smith:
If you'd be able to make the house style guidelines available online, that'd be very helpful. 
It's a borderline case, but I'd be OK with hosting the style, but it's not great for us to host a style when we have no ability to check its correctness in the case of error reports.

elidourado:
Thanks for your feedback and willingness to host the style. I understand that it's a borderline case, and I appreciate your understanding.

I will discuss the possibility of publishing the style guide with the Mercatus publications team, but fair warning: that internal discussion may take some time and is not fully under my control. In the meantime, if others agree, it would be helpful to include the style in the repository to demonstrate value to the decision-makers in this regard.

adam3smith:
@rmzelle what do you think?

rmzelle:
@adam3smith, looks like a big enough organization to accept the style. And it's GMU, so they get brownie points for creating Zotero.

@elidourado, we'd also be more than happy if you could just attach a copy of the style guide (or the relevant pages that concern citations) to this GitHub issue, assuming that doesn't get you in trouble (that would still make it publicly available, but the center wouldn't have to formally publish it). Like @adam3smith mentioned, we really like to be able to point to a style guide so that we can confirm and fix any reported issues. It also helps us years later in determining whether a style is likely to still be in use.

rmzelle:
(otherwise it looks good to merge)

elidourado:
Thanks, everyone. I got approval to informally post the citation section of our style guide. It's attached below.

[Mercatus-CitationStyle-v2.1.pdf](https://github.com/citation-style-language/styles/files/439352/Mercatus-CitationStyle-v2.1.pdf)

rmzelle:
Thanks!

PR #2189, opened 2016-08-24
Update publisher note (type = thesis)

rjstapel:
Type "thesis": removed prefix "unpublished" and changed the order in the publisher information, according to De Buck, Zoeken en Schrijven.
[Not sure if I actually proposed this file change already or not (could not find any evidence in my administration), so sorry if this is a duplicate]

### DIFF
--- a/de-buck.csl
+++ b/de-buck.csl
@@ -20,7 +20,7 @@
     <category citation-format="note"/>
     <category field="generic-base"/>
     <summary>Annotatie en literatuurlijst volgens P. de Buck e.a., Zoeken en schrijven. Handleiding bij het maken van een historisch werkstuk (Haarlem 1982). Gebaseerd op "MHRA format with full notes and bibliography"</summary>
-    <updated>2016-05-25T12:00:00+01:00</updated>
+    <updated>2016-08-24T19:15:50+00:00</updated>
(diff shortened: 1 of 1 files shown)


rmzelle:
> Type "thesis": removed prefix "unpublished" and changed the order in the publisher information, according to De Buck, Zoeken en Schrijven.

Did you test whether your proposed version gives correct results? I'm wondering whether the delimiter outside of the `<choose/>` group has any impact on the various `<text/>` elements, or whether you get something like "genrepublisher,publisherplace1990" instead of "genre publisher, publisherplace 1990".

```xml
          <group delimiter=" ">
            <choose>
              <if variable="title" match="none"/>
              <else-if type="thesis">
                <text variable="genre"/>
                <text variable="publisher" suffix=","/>
                <text macro="issued"/>
              </else-if>
```

> [Not sure if I actually proposed this file change already or not (could not find any evidence in my administration), so sorry if this is a duplicate]

You made the change a few days ago (https://github.com/rjstapel/styles/commit/68eafa628cfb7229ae5c57fb7f0a57cea702efe4), but you didn't create a pull request for that change. So no, it doesn't look like you ever proposed this before.

rjstapel:
In Zotero at least, this did not produce any problems. 

Yes, I forgot to make the pull request.

rmzelle:
Okay, thanks. (I don't think the CSL specification currently discusses whether this should work or not)

PR #2192, opened 2016-08-31
hanging indents

brevans:
There doesn't seem to be a quick and permanent way to do this locally, also this should be standard anyhow.


rmzelle:
Great, thanks. (and sorry for the long wait)

PR #2206, opened 2016-09-07

rmzelle:
change and "text" to "symbol"
change name as sort order from first to always

### DIFF
--- a/south-african-journal-of-animal-science.csl
+++ b/south-african-journal-of-animal-science.csl
@@ -20,7 +20,7 @@
   </info>
   <macro name="editor-translator">
     <names variable="editor translator" prefix="(" suffix=")" delimiter=", ">
-      <name and="text" initialize-with="" delimiter=", "/>
+      <name and="symbol" initialize-with="" delimiter=", "/>
       <label form="short" prefix=", " text-case="capitalize-first"/>
     </names>
   </macro>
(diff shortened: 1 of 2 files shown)


south-african-journal-of-animal-science.csl
```diff
@@ -39,7 +39,7 @@
   </macro>
   <macro name="author">
     <names variable="author">
-      <name delimiter=", " and="text" delimiter-precedes-last="always" initialize-with=". " name-as-sort-order="first"/>
+      <name delimiter=", " and="symbol" delimiter-precedes-last="always" initialize-with=". " name-as-sort-order="always"/>
```
rmzelle: @Yusufras, `name-as-sort-order="always"` is not valid CSL. It should be `name-as-sort-order="all"` (see the description of "name-as-sort-order" under http://docs.citationstyles.org/en/stable/specification.html#name). Could you correct this?

rmzelle: (you can update this pull request by using the edit icon at https://github.com/Yusufras/styles/blob/patch-16/south-african-journal-of-animal-science.csl)

Yusufras:
The changes made does not seem to work, although i followed the process.

rmzelle:
> The changes made does not seem to work, although i followed the process.

No, you were fine. I've just been too busy to take a look at this.

rmzelle:
All done!

PR #2212, opened 2016-09-08
Update aquatic-invasions.csl

mccullermi:
Fixed issues with punctuation and added variables where needed. General tidying.

### DIFF
--- a/aquatic-invasions.csl
+++ b/aquatic-invasions.csl
@@ -14,7 +14,7 @@
     <category citation-format="author-date"/>
     <category field="biology"/>
     <eissn>1818-5487</eissn>
-    <updated>2016-09-08T20:41:46+00:00</updated>
+    <updated>2016-09-08T22:00:30+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="edition">
@@ -41,9 +41,6 @@
         <text variable="title" form="short" quotes="true"/>
       </substitute>
     </names>
-  </macro>
-  <macro name="title">
-    <text variable="title"/>
   </macro>
   <macro name="publisher">
     <choose>
@@ -160,7 +157,7 @@
           <choose>
             <if type="bill graphic legal_case legislation motion_picture song" match="any">
               <group delimiter=" ">
-                <text macro="title"/>
+                <text variable="title"/>
                 <group delimiter=", " suffix=". ">
                   <text macro="edition"/>
                   <text variable="collection-title"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
(just as a note -- if you want to make improvements to this, please do so within this pull request. Don't start a new one).

mccullermi:
Thanks! I'm still new to this..

adam3smith:
feel free to ask for help if you're stuck

aquatic-invasions.csl
```diff
@@ -161,50 +160,101 @@
             <date-part name="year"/>
           </date>
           <choose>
-            <if type="bill graphic legal_case legislation motion_picture song book" match="any">
-              <group delimiter=", " suffix=". ">
-                <text macro="title"/>
+            <if type="bill graphic legal_case legislation motion_picture song" match="any">
+              <group delimiter=" ">
+                <group>
```
adam3smith: remove this group (doesn't do anything since it has just one element)

aquatic-invasions.csl
```diff
@@ -161,50 +160,101 @@
...
                   <text macro="publisher"/>
                 </group>
               </group>
               <text macro="page"/>
             </else-if>
-            <else-if type="chapter paper-conference" match="any">
+            <else-if type="chapter" match="any">
+              <group delimiter=", " suffix=". ">
+                <group delimiter=" ">
+                  <group suffix=". ">
```
adam3smith: Same here: remove the group, move the suffix down to the variable, and remove the trailing space in the suffix, since that'll be inserted by the delimiter anyway

aquatic-invasions.csl
```diff
@@ -161,50 +154,101 @@
...
+                </group>
+                <group delimiter=", " prefix=" ">
+                  <group delimiter=" ">
+                    <text variable="collection-title"/>
+                    <text variable="collection-number"/>
+                  </group>
+                  <text macro="edition"/>
+                </group>
+              </group>
+              <group delimiter=", ">
```
adam3smith: remove group here, too

adam3smith:
a couple of small nits -- see my comments inline. Looks good otherwise, thanks!

aquatic-invasions.csl
mccullermi: The issue here is that when I remove the group, it defaults to the delimiter of the preceding group and I need (for this ref type), a period instead of a comma between the title and container-title but a comma between other things. I played around with it for a while and this seemed to be the best way to get that, but again, still new to this!

adam3smith: A group with a single element never changes anything. Removing the group (i.e. lines 232 and 234) and moving the suffix down to the variable, this will get the space delimiter from the group starting in line 231 -- which it's already getting anyway.

mccullermi: No, I see what you mean. I had kept the title macro from the template style used and there is no reason to have it. Removing it because if I do your suggestion as-is, it changes the title everywhere. Now updating titles and removing unneeded groups. Thanks!

rmzelle:
Thanks!

Issue #2197, opened 2016-09-03
Update DOI display from latest Crossref guidelines?


adunning:
The [Crossref DOI display guidelines](http://www.crossref.org/02publishers/doi_display_guidelines.html) were recently updated: the format `doi:10.NNNN/doisuffix` is now discouraged, recommending instead that all DOIs be displayed with a resolver, e.g. `https://doi.org/10.NNNN/doisuffix`. There are many styles, of course, that use either the old format or the longer resolver address; it will probably be some time before they are updated, but in the meantime, I am finding that editors (at least the ones who want to include DOIs at all) prefer the newer guidelines.

As this is a purely technical matter, I think it would be sane to assume that the latest Crossref guidelines should overrule older style guides, but this is obviously open to discussion.

adam3smith:
as you know, we're doing that for APA and there, in spite of clear statements by APA itself, we're getting a ton of complaints -- so I'd be quite hesitant to do it with Chicago Manual, where we wouldn't be able to point to similar statements by the editors. Which other styles do you have in mind?

adunning:
Interesting that has happened with APA; not what I would have expected, and given that it's probably best not to change it for Chicago for now. What about the case of e.g. MHRA, which still uses the old `http://dx.doi.org/` resolver? I can't imagine anyone would complain about that being shortened.

adam3smith:
> I can't imagine anyone would complain about that being shortened.

you need to widen your imagination -- it happens all the time -- but that one I'm willing to fight. We should use doi.org/ as a resolver across the board. Do we use dx.doi.org anywhere but in MHRA?
I'd gladly take a PR for this -- I'm super busy right now.

rmzelle:
> Do we use dx.doi.org anywhere but in MHRA?

In quite a few styles (search for "dx.doi"):

```
/Users/rintzezelle/Documents/CSL/styles/advances-in-alzheimers-disease.csl
59: <text variable="DOI" prefix="http://dx.doi.org/"/>
/Users/rintzezelle/Documents/CSL/styles/american-journal-of-plant-sciences.csl
106: <text variable="DOI" prefix="http://dx.doi.org/"/>
/Users/rintzezelle/Documents/CSL/styles/american-society-of-civil-engineers.csl
8: <link href="http://dx.doi.org/10.1061/9780784478998.ch17" rel="documentation"/>
9: <link href="http://dx.doi.org/10.1061/9780784479018.ch05" rel="documentation"/>
[...]

PR #2213, opened 2016-09-09
Switch from dx.doi.org to doi.org per official Crossref guidelines

rmzelle:
See https://github.com/citation-style-language/styles/issues/2197 for
discussion

### DIFF
--- a/advances-in-alzheimers-disease.csl
+++ b/advances-in-alzheimers-disease.csl
@@ -56,7 +56,7 @@
     <choose>
       <if variable="DOI">
         <group>
-          <text variable="DOI" prefix="http://dx.doi.org/"/>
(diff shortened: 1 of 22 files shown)


rmzelle:
@adam3smith, if you have time, feel free to give it a quick lookover. It's all a find-replace from "dx.doi.org" to "doi.org", with the exception of water-alternatives.csl.

rmzelle:
Qui tacet consentire videtur!

adam3smith:
I only taceted for 13 hs ;), but this looks good, thanks!

rmzelle:
Yeah, I know, but it's a rather straightforward change and you said you were busy anyway.

PR #2216, opened 2016-09-09
Update oncotarget.csl

taraonco:
Added a period/full stop at the end of references with DOIs

### DIFF
--- a/oncotarget.csl
+++ b/oncotarget.csl
@@ -84,7 +84,7 @@
         <choose>
           <if variable="DOI">
             <text variable="DOI" prefix="doi: "/>
-            <text variable="DOI" suffx="."/>
+            <text variable="DOI" suffix="."/>
           </if>
           <else>
             <choose>
--- a/oncotarget.csl
+++ b/oncotarget.csl
@@ -84,6 +84,7 @@
         <choose>
           <if variable="DOI">
             <text variable="DOI" prefix="doi: "/>
+            <text variable="DOI" suffx="."/>
           </if>
           <else>
             <choose>
--- a/oncotarget.csl
+++ b/oncotarget.csl
@@ -83,8 +83,7 @@
       <else>
         <choose>
           <if variable="DOI">
-            <text variable="DOI" prefix="doi: "/>
-            <text variable="DOI" suffix="."/>
+            <text variable="DOI" prefix="doi: " suffix="."/>
           </if>
           <else>
             <choose>


rmzelle:
Thanks! I made a small edit to prevent the DOI from being printed twice.

(confirmed that Oncotarget uses a period after the DOI in http://www.impactjournals.com/oncotarget/index.php?journal=oncotarget&page=article&op=view&path%5B%5D=11160 )

taraonco:
Thank you, I can see now that it was indeed duplicating the DOIs. However, I just tried out the new file (now available here https://www.zotero.org/styles?q=oncotarget), and it does not seem to add a period after the DOI.

rmzelle:
Did you test that by installing the latest version, or just by looking at the preview from the style repository? The latter might not have updated yet.

taraonco:
I installed the latest version, replacing my previous Oncotarget style in Zotero.

rmzelle:
It works for me:

> Gomes RR, Glienke C, Videira SIR, Lombard L, Groenewald JZ, Crous PW. Diaporthe: a genus of endophytic, saprobic and plant pathogenic fungi. Persoonia [Internet]. 2013 [cited 2016 Aug 26]; 31: 1–41. doi: 10.3767/003158513X666844.

taraonco:
I have been testing it out on the attached MS 14545... file, which was formatted using Zotero. When I go to "Set Document Preferences" and set the style to Oncotarget, DOIs are added, but without periods.


I also tried generating a random bibliography from a few files in my library, and they exported without DOIs (Untitled Bibliography).


I am sure that I am using the new file. Almost all the references I am using Zotero for are journal references. Not sure if this might make a difference.

rmzelle:
(you can only attach files via the GitHub website, not via email)

You can try resetting your CSL styles in Zotero (see https://www.zotero.org/support/preferences/advanced#database_maintenance). Alternatively, you can use the Zotero CSL Editor and copy and paste the style from the CSL repository, to check if that gives the expected result. I'm quite sure the style now prints a period after the DOI, so this seems like an installation issue.

taraonco:
Here is the file I was trying to send before. I tried both resetting my styles, and deleting the style from my library completely then downloading it anew, but I am still getting the same result.
[MS 14545 PII 11228-Zotero.docx](https://github.com/citation-style-language/styles/files/465206/MS.14545.PII.11228-Zotero.docx)

adam3smith:
in an existing document, switch to a different style and back.

taraonco:
That seemed to do it, thanks!

PR #2217, opened 2016-09-09
Create oncoscience.csl

taraonco:
Identical to the Oncotarget.csl style that is already in the repository (dependent)


rmzelle:
Thanks! (and sorry for the wait)

PR #2218, opened 2016-09-09
Create genescancer.csl

taraonco:
Identical to the Oncotarget.csl style that is already in the repository (dependent)


adam3smith:
thanks here, too, looks great.

Issue #2196, opened 2016-09-02
Chicago Illustrations CSL is ostensibly incorrect


dsifford:
CC: @alexwatkins 

Hi there,

I think the CSL might be incorrect for `chicago-figures.csl`.

![image](https://cloud.githubusercontent.com/assets/5240018/18191498/03b8eb16-7099-11e6-851e-1dac5baffc32.png)

I'm aware that the style isn't _supposed_ to render a bibliography, however there are other styles that similarly do not render bibliographies, and do not create this issue. For example, the most recent style added to this repo: [mercatus-center.csl](https://github.com/citation-style-language/styles/blob/master/mercatus-center.csl).

I'm fairly certain just removing [these lines](https://github.com/citation-style-language/styles/blob/master/chicago-figures.csl#L32-L69) will fix this error (but I'm certainly no expert with CSL XML, so I'll leave that for the group to decide on)

**Edit:** Tried the above in the CSL code editor and it just duplicated the inline citations, so chock that one as a 'no'. I'm totally unskilled with CSL, so I'll leave it to you experts to school me on 😄 

It's not a huge deal as-is, but the issue in my case (a citeproc-js user) is: Having bibliography fields set causes citeproc to call somewhere in the ballpark of 20-40 individual function calls (including recursion loops) before eventually returning an error, which is much slower in comparison to what happens with `mercatus-center.csl`.

Thanks so much for everyone's work on this 😄

alexwatkins:
I'm certainly no CSL expert, it does render bibliography entries but only
for images and book sections (so that you can cite an image from a book
such as a frontispiece). Since the example bibliography doesn't include
those two as examples it just says there's no printed form. That's what I
was going for here,but welcome any fixes that make it work better!

-Alex

rmzelle:
The Mercatus Center CSL style is a "note" style, and for that type of citation format bibliographies are often optional, since the (foot)notes (rendered by `<citation/>`) are usually pretty elaborate, and enough to locate the cited item. 

This is typically not the case for "in-text" CSL styles. In-text citations are usually pretty sparse (e.g. "(fig. 1)" doesn't tell you anything about the item), so for these styles the bibliography is pretty much required.

It looks like the CSL spec just says the the bibliography is optional, but it's just rather nonsensical for in-text styles, so it's not too surprising citeproc-js throws an error.

rmzelle:
[...]

PR #2220, opened 2016-09-12
Reorganize Chicago Figures code

rmzelle:
Reorganize code a little in response to #2196.

Better use of cs:choose, groups, and the style will now print at least something for most item types.


rmzelle:
Not fully tested, since the current stable version of Zotero has a citation bug that affects this particular style. Should be fixed in the next release (https://forums.zotero.org/discussion/62288/citeproc-js-bug), at which point I'll revisit this and merge.

rmzelle:
Looks good in Zotero 4.0.29.15.

PR #2227, opened 2016-09-17
Show locators on ibid-with-locator

fbennett:
Should more or less address <a href="https://forums.zotero.org/discussion/60384/legal-case-references-not-showing-page-numbers-after-ibid">Legal case references not showing page numbers after ibid.</a>


adam3smith:
thanks!

PR #2233, opened 2016-09-21
Fix spacing and punctuation for Protein Engineering, Design, and Selection

BYachnin:
Hi.  I've made a few minor changes to the spacing and punctuation in the bibliography section for the current PEDS citation style.  The current requirements can be found here: http://www.oxfordjournals.org/our_journals/proeng/for_authors/manuscript.html#Sections of the Manuscript

### DIFF
--- a/protein-engineering-design-and-selection.csl
+++ b/protein-engineering-design-and-selection.csl
@@ -16,7 +16,7 @@
     <issn>1741-0126</issn>
     <eissn>1741-0134</eissn>
     <published>2014-09-06T21:00:00-06:00</published>
-    <updated>2016-09-20T22:05:00-04:00</updated>
+    <updated>2016-09-21T20:45:00-04:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="author-short">
@@ -35,15 +35,14 @@
     </date>
   </macro>
   <macro name="publisher">
-    <group delimiter=": ">
-      <text variable="publisher-place"/>
+    <group delimiter=", ">
       <text variable="publisher"/>
+      <text variable="publisher-place"/>
     </group>
   </macro>
   <macro name="editor">
     <names variable="editor">
-      <name initialize-with=". " and="text" delimiter-precedes-last="never"/>
-      <label form="short"/>
+      <name name-as-sort-order="all" sort-separator="," initialize-with="." and="text" delimiter-precedes-last="never"/>
     </names>
   </macro>
   <citation et-al-min="3" et-al-use-first="1" and="text" disambiguate-add-year-suffix="true" collapse="year-suffix">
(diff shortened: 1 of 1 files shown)


protein-engineering-design-and-selection.csl
```diff
@@ -16,7 +16,8 @@
     <issn>1741-0126</issn>
     <eissn>1741-0134</eissn>
     <published>2014-09-06T21:00:00-06:00</published>
-    <updated>2014-09-07T00:00:00-06:00</updated>
+    <updated>2014-09-11T02:57:16+00:00</updated>
+    <updated>2016-09-20T22:05:00-04:00</updated>
```
adam3smith: you have two update timestamps. I think that's the only issue that prevent validation from goting through.

protein-engineering-design-and-selection.csl
```diff
@@ -69,11 +70,11 @@
         <text macro="issued" prefix="(" suffix=")"/>
         <choose>
           <if type="article article-magazine article-newspaper article-journal review" match="any">
-            <group delimiter=",">
+            <group delimiter=", ">
               <text variable="container-title" form="short" font-style="italic" text-case="title"/>
               <text variable="volume" font-weight="bold"/>
+			  <text variable="page" suffix="."/>
```
adam3smith: although you should also fix indenting here.

rmzelle:
http://oxfordjournals.org/ seems to be down right now, but judging from the Google Cache (http://webcache.googleusercontent.com/search?q=cache:fhn407JIAYUJ:www.oxfordjournals.org/proeng/for_authors/manuscript.html+&cd=1&hl=en&ct=clnk&gl=us), the "editor" macro seems to need to use the same parameters as used in the "author" macro (`name-as-sort-order="all" sort-separator="," initialize-with="."`):

> Norris,K.E., Norris,F. and Brunfeldt,K. (1980) In Köster,H. (ed), Nucleic Acids Synthesis: Applications to Molecular Biology and Genetic Engineering. IRL Press, London, pp. 233-241.

BYachnin:
I have now updated the publisher and editor parts to match up with the style requirements.

rmzelle:
( @adam3smith, see https://github.com/citation-style-language/styles/pull/2233#commitcomment-19124736)

adam3smith:
@rmzelle OK, good by me, but clearly I'm not fresh enough to merge today anymore.

rmzelle:
@BYachnin, let us know if we introduced any new issues. @adam3smith and I tried to improve the style a little bit more, but it was rather late when we looked at it, and we might have overlooked something.

rmzelle:
And thanks!

BYachnin:
@rmzelle and @adam3smith , it looks to me like you guys just changed things around to make it less hard coded, plus a couple of other corrections that I overlooked.  I took a quick look and everything looked fine.  I'll be submitting a paper to PEDS soon, so I will keep an eye on the new specification and let you know if anything seems off.

rmzelle:
> it looks to me like you guys just changed things around to make it less hard coded, plus a couple of other corrections that I overlooked.

Yes, that was the plan. We try to avoid hardcoding text so that styles can be more easily used in other languages.

PR #2191, opened 2016-08-26
Create vancouver-drsm.csl

pedrom34:
CSL Style adaptation of Vancouver citation style for Service Médical de l'Assurance Maladie (French health insurance)

PR #2235, opened 2016-09-22
Update service-medical-de-l-assurance-maladie.csl

pedrom34:
Two small fixes related to book chapter citation that my users requested. In accordance with documentation (https://github.com/citation-style-language/styles/pull/2191), Abstract shall not be part of the citation style, and given name are always initialized.


rmzelle:
Thanks!

PR #2237, opened 2016-09-23
Create cahiers-d-ethnomusicologie.csl

svictor9:
Style for the journal Cahiers d’ethnomusicologie (French). Adapted from  http://www.zotero.org/styles/iso690-author-date-en

### DIFF
--- a/cahiers-d-ethnomusicologie.csl
+++ b/cahiers-d-ethnomusicologie.csl
@@ -1,0 +1,489 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" default-locale="fr-FR">
+  <info>
+    <title>Cahiers d'ethnomusicologie (French)</title>
+    <id>http://www.zotero.org/styles/cahiers-d-ethnomusicologie</id>
+    <issn>1662-372X</issn>
+    <eissn>2235-7688</eissn>
+    <link
+	href="http://www.zotero.org/styles/cahiers-d-ethnomusicologie"
+	rel="self"/>
+    <link href="http://www.zotero.org/styles/iso690-author-date-en"
+	  rel="template" />
+    <link
+	href="http://www.adem-geneve.com/fr/cahiers-d-ethnomusicologie-contribution"
+	rel="documentation" />
+    <author>
+      <name>Victor A. Stoichita</name>
+      <email>svictor@svictor.net</email>
+      <uri>http://www.svictor.net</uri>
+    </author>
+    <category citation-format="author-date"/>
+    <summary>Style based on ISO 690:2010(F), V2 adapted from a template by L.
+    Mellifluo, R. Grolimund, M. Hardegger and P.-A. Giraud.</summary>
+    <updated>2016-09-24T23:32:13+00:00</updated>
+    <rights
+    license="http://creativecommons.org/licenses/by-sa/3.0/">This work
+    is licensed under a Creative Commons Attribution-ShareAlike 3.0
+    License</rights>
+  </info>
+  <locale>
+    <terms>
+      <term name="anonymous">Anon.</term>
+      <term name="no date">[sans date]</term>
+      <term name="in">in</term>
+      <term name="online">en ligne</term>
+      <term name="accessed">consulté le</term>
+      <term name="retrieved">disponible</term>
+      <term name="from">à l'adresse</term>
+      <term name="editor" form="short">
+	<single>dir</single>
+	<multiple>dirs.</multiple>
+      </term>
+      <term name="page" form="short">
+        <single>p.</single>
+        <multiple>p.</multiple>
+      </term>
+    </terms>
+  </locale>
+  <macro name="author">
+    <names variable="author">
+      <name and="text" name-as-sort-order="first" initialize="false" sort-separator=" " delimiter=", " delimiter-precedes-last="never">
+        <name-part name="family" text-case="uppercase"/>
+        <name-part name="given"/>
+      </name>
+    </names>
+  </macro>
+  <macro name="editor">
+    <names variable="editor">
+      <name and="text" name-as-sort-order="all" form="short" sort-separator=", " delimiter=", " delimiter-precedes-last="never">
+        <name-part name="family"/>
+        <name-part name="given"/>
+      </name>
+      <label prefix=" (" form="short" suffix=".)"/>
+    </names>
+  </macro>
+  <macro name="translator">
+    <names variable="translator">
+      <name and="text" name-as-sort-order="all" sort-separator=", " delimiter=", " delimiter-precedes-last="never">
+        <name-part name="family" text-case="uppercase"/>
+        <name-part name="given"/>
+      </name>
+      <label prefix=" (" form="short" suffix=".)"/>
+    </names>
+  </macro>
+  <macro name="responsability">
+    <choose>
+      <if variable="author editor translator" match="any">
+        <choose>
+          <if variable="author">
+            <text macro="author"/>
+          </if>
+          <else-if variable="editor">
+            <text macro="editor"/>
+          </else-if>
+          <else>
+            <text macro="translator"/>
+          </else>
+        </choose>
+      </if>
+      <else>
+        <text term="anonymous" text-case="uppercase"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="author-citation">
+    <choose>
+      <if variable="author editor translator" match="any">
+        <names variable="author">
+          <name form="short"/>
+          <substitute>
+            <names variable="editor"/>
+            <names variable="translator"/>
+          </substitute>
+        </names>
+      </if>
+      <else>
+        <text term="anonymous" text-case="uppercase"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="container-author">
+    <names variable="container-author">
+      <name and="text" name-as-sort-order="all" sort-separator=", " delimiter=", " delimiter-precedes-last="never">
+        <name-part name="family" text-case="uppercase"/>
+        <name-part name="given"/>
+      </name>
+    </names>
+  </macro>
+  <macro name="container-responsability">
+    <choose>
+      <if variable="container-author editor translator" match="any">
+        <choose>
+          <if variable="container-author">
+            <text macro="container-author"/>
+          </if>
+          <else-if variable="editor">
+            <text macro="editor"/>
+          </else-if>
+          <else>
+            <text macro="translator"/>
+          </else>
+        </choose>
+      </if>
+      <else>
+        <text term="anonymous" text-case="uppercase"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="year-date">
+    <choose>
+      <if variable="issued">
+        <date variable="issued">
+          <date-part name="year" form="long" suffix=" "/>
+        </date>
+      </if>
+      <else>
+        <text term="no date"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="title">
+    <choose>
+      <if type="book thesis map motion_picture song manuscript" match="any">
+        <text variable="title" font-style="italic"/>
+      </if>
+      <else-if type="paper-conference speech chapter article-journal article-magazine article-newspaper post-weblog post webpage broadcast" match="any">
+        <text variable="title" prefix="« " suffix=" ». "/>
+        <choose>
+          <if variable="container-author editor translator" match="any">
+            <text macro="container-responsability"/>
+            <choose>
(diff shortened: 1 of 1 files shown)


svictor9:
The author guidelines for this journal are here : http://www.adem.ch/fr/cahiers-d-ethnomusicologie-contribution

cahiers-d-ethnomusicologie.csl
```diff
@@ -0,0 +1,484 @@
...
+      </if>
+    </choose>
+  </macro>
+  <macro name="edition">
+    <text variable="edition" form="long"/>
+  </macro>
+  <macro name="publisher-place">
+    <choose>
+      <if type="patent manuscript article-newspaper broadcast motion_picture song" match="any">
+        <choose>
```
adam3smith: no need for this `<choose>` element. Simply `<text variable="publisher-place"/>` will do the same.

cahiers-d-ethnomusicologie.csl
```diff
@@ -0,0 +1,484 @@
...
+          <if variable="publisher-place">
+            <text variable="publisher-place"/>
+          </if>
+        </choose>
+      </if>
+      <else>
+        <choose>
+          <if variable="publisher-place">
```
adam3smith: (here you do need the if because it's followed by an else-if)

cahiers-d-ethnomusicologie.csl
```diff
@@ -0,0 +1,484 @@
...
+    <group>
+      <text variable="volume" prefix=" "/>
+      <text variable="issue" prefix="(" suffix=")"/>
+      <text variable="page" prefix=" : "/>
+    </group>
+  </macro>
+  <macro name="publisher">
+    <choose>
+      <if type="broadcast motion_picture song report" match="any">
+        <choose>
```
adam3smith: same here: no need for the `<choose> <if>`

cahiers-d-ethnomusicologie.csl
```diff
@@ -0,0 +1,484 @@
...
+      <if variable="URL">
+        <group>
+          <text term="retrieved" suffix=" " text-case="capitalize-first"/>
+          <text term="from" suffix=" : "/>
+          <text variable="URL"/>
+        </group>
+      </if>
+    </choose>
+  </macro>
+  <macro name="archive">
```
adam3smith: I'm not happy about that awkward `" : "` in a text value here. How about collapsing the two macros:

```
<macro name="archive">
  <group delimiter=" : ">
    <text variable="archive"/>
    <text variable="archive_location"/>
  </group>
  <choose>
    <if variable="archive_location" match="none">
      <text variable="call-number"/>
    </if>
  </choose>
</macro>
```
Still worried though: what happens when you have an archive and a call number? Right now they'd just be smashed together. Also, I find the call-number field here a bit hard to believe: this would print the call-number for any book added from a library catalog. Surely that can't be right? What's the goal here?

cahiers-d-ethnomusicologie.csl
```diff
@@ -0,0 +1,484 @@
...
+        <text variable="call-number"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="note">
+    <text variable="note"/>
+  </macro>
+  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year" year-suffix-delimiter=", " after-collapse-delimiter=" ; ">
+    <layout prefix="(" suffix=")" delimiter=" ; ">
+      <group delimiter=", p. ">
```
adam3smith: let's not hardcode this, instead use:
`<group delimiter=", ">` here and then use

```
<group delimiter=" ">
  <text variable="locator"/>
</group>
```
below

cahiers-d-ethnomusicologie.csl
```diff
@@ -0,0 +1,484 @@
...
+  <macro name="translator">
+    <names variable="translator">
+      <name and="text" name-as-sort-order="all" sort-separator=", " delimiter=", " delimiter-precedes-last="never">
+        <name-part name="family" text-case="uppercase"/>
+        <name-part name="given"/>
+      </name>
+      <label prefix=" (" form="short" suffix=".)"/>
+    </names>
+  </macro>
+  <macro name="responsability">
```
adam3smith: I'm pretty flexible with macro names, but I don't think we should use a non-existent English word. Either fix this (and all calls to it) to `responsibility` or maybe use the `responsable` which works in French and English (I think?)

cahiers-d-ethnomusicologie.csl
```diff
@@ -0,0 +1,484 @@
...
+          <if variable="container-title">
+            <text variable="container-title" font-style="italic"/>
+          </if>
+          <else>
+            <text variable="event" font-style="italic"/>
+          </else>
+        </choose>
+      </else-if>
+      <else-if type="report">
+        <text variable="number" suffix=" : "/>
```
adam3smith: This seems odd to me: wouldn't we usually add the number not to the report title but to its description? Something like
Author 1999. Report title. Number : Report Type. ?

cahiers-d-ethnomusicologie.csl
```diff
@@ -0,0 +1,484 @@
...
+          </else>
+        </choose>
+      </else>
+    </choose>
+  </macro>
+  <macro name="issue">
+    <group>
+      <text variable="volume" prefix=" "/>
+      <text variable="issue" prefix="(" suffix=")"/>
+      <text variable="page" prefix=" : "/>
```
adam3smith: This applies to all instances of " : " throughout the style. My understanding is that French typography requires non-breaking spaces for these. Do this by using `"&#160;:&#160;"` instead (unless I'm wrong about this).

cahiers-d-ethnomusicologie.csl
```diff
@@ -0,0 +1,484 @@
...
+          <else>
+            <text value="s.l." text-case="capitalize-first"/>
+          </else>
+        </choose>
+      </else>
+    </choose>
+  </macro>
+  <macro name="issue">
+    <group>
+      <text variable="volume" prefix=" "/>
```
adam3smith: no need for the prefix here: you're setting spaces in all suffixes in the bibliography below.

cahiers-d-ethnomusicologie.csl
```diff
@@ -0,0 +1,484 @@
...
+	  rel="template" />
+    <link
+	href="http://www.adem-geneve.com/fr/cahiers-d-ethnomusicologie-contribution"
+	rel="documentation" />
+    <author>
+      <name>Victor A. Stoichita</name>
+      <email>svictor@svictor.net</email>
+      <uri>http://www.svictor.net</uri>
+    </author>
+    <category citation-format="author-date"/>
```
adam3smith: add a field from the lists [here](http://docs.citationstyles.org/en/stable/specification.html#appendix-i-categories). I'd think `<category field="humanities"/>` fits best.

svictor9: Sure. "Humanities" is good.

cahiers-d-ethnomusicologie.csl
svictor9: Mais bien sûr :blush:

cahiers-d-ethnomusicologie.csl
svictor9: Yes. You’re probably right. There’s a problem here, which might come back later on.
The journal’s explicit guidelines are quite minimal. I sent the link in a previous comment. When the editors have to deal with documents which are not stated in the guidelines they try to do "something sensible", usually something which will look like the other references. 
Now I started this style from a template which claims to follow the ISO 690 norm for French bibliographies (its link is in the `<info>`). The template covers far more document types than those addressed in the journal guidelines. I hesitated between keeping and removing those extra types, and finally opted for keeping them, in case they come in handy some day. But quite frankly, I sometimes don’t know how people use those types. "Report" is an example of this: I don’t use it and never encountered it in my colleagues’ or my students’ bibliographies. I have no idea what the report "number" would represent in our discipline. That being said, I agree that in case someone uses it, putting a number right after the title doesn’t really make sense in this style. Following your suggestion, I then migrated the number to the description, between the "edition" and the "publisherplace".
If you prefer to deal with this kind of things — I mean document types which are not explicitly addressed in the journal’s guidelines — in another way, please let me know.

cahiers-d-ethnomusicologie.csl
svictor9: Indeed. Removed the `choose`

cahiers-d-ethnomusicologie.csl
svictor9: Got it :ok_hand:

cahiers-d-ethnomusicologie.csl
svictor9: ok

cahiers-d-ethnomusicologie.csl
svictor9: That’s right. Actually those were already a kind of non-breaking spaces (and worked as such on my computer) but inserted straight with a keyboard combination and probably lost in the copy/paste to Git. I changed all occurences.

cahiers-d-ethnomusicologie.csl
svictor9: ok

cahiers-d-ethnomusicologie.csl
svictor9: This is another example of something I don’t fully understand myself. It was in the original template when I first downloaded it. 

I just checked and there is an updated version of that template. This archive macro has been rewritten. I think it now addresses both the delimiter and the call-number problems so I replaced it in my style too.

cahiers-d-ethnomusicologie.csl
svictor9: Ok.
`

svictor9:
@adam3smith I just updated the file following the changes you requested. The csl-bot now complains that the file is not parsable as csl. I don’t understand why since the csl validator at http://validator.citationstyles.org/ is happy with it. Please tell me if I’m doing something wrong.

cahiers-d-ethnomusicologie.csl
```diff
@@ -0,0 +1,473 @@
...
+  <macro name="translator">
+    <names variable="translator">
+      <name and="text" name-as-sort-order="all" sort-separator=", " delimiter=", " delimiter-precedes-last="never">
+        <name-part name="family" text-case="uppercase"/>
+        <name-part name="given"/>
+      </name>
+      <label prefix=" (" form="short" suffix=".)"/>
+    </names>
+  </macro>
+  <macro name="respomsibility">
```
adam3smith: this should be `responsibility` with an n (same below for the container).

adam3smith:
To be honest, I don't understand this either. @rmzelle -- any idea what's going on?
Validation passes: http://validator.citationstyles.org/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fsvictor9%2Fstyles%2Fe53820df6bb15f6a1628d581c68c33b4c09f2385%2Fcahiers-d-ethnomusicologie.csl&version=1.0.1
the style runs in Zotero, so I'm at a loss. Also have never seen that error before, I believe.

cahiers-d-ethnomusicologie.csl
svictor9: Oh dear, typo again! Sorry!

adam3smith:
This all looks good to me now and I have no idea why the checks fail. @rmzelle -- might be something obvious. Could you give it a look?

cahiers-d-ethnomusicologie.csl
rmzelle: (if you use http://formatter.citationstyles.org/, be also aware that unescaped special whitespace characters, such as non-breaking spaces, are typically replaced by regular spaces by the reformatter; it shouldn't do that, but I haven't found a way to fix it yet)

cahiers-d-ethnomusicologie.csl
```diff
@@ -0,0 +1,473 @@
...
+  <macro name="archive_location">
+    <choose>
+      <if variable="archive_location">
+        <text variable="archive_location"/>
+      </if>
+      <else>
+        <text variable="call-number"/>
+      </else>
+    </choose>
+  </macro>
```
rmzelle: @svictor9, @adam3smith, I'm pretty sure that the problem is that there are two "archive_location" macros here (lines 299-318). (the rest of the macros are unique)

Just deleting one of them (they are identical anyway) should do the trick.

rmzelle: (otherwise the style looks good to merge to me)

svictor9: The duplicate macro was the problem indeed. Thanks!

rmzelle:
Thanks!

PR #2239, opened 2016-09-26
Create wirtschaftsuniversitat-wien-handel-und-marketing.csl

georgd:


adam3smith:
Thanks, looks good. (Note I made a small cosmetic change in line with CSL repository conventions, so please start from the repository version of the style if you ever need to make further changes)

PR #2241, opened 2016-09-27
Switch from "http" to "https" for DOI links

rmzelle:
Per #2197

### DIFF
--- a/academy-of-management-review.csl
+++ b/academy-of-management-review.csl
@@ -89,7 +89,7 @@
       <if variable="page" match="none">
         <choose>
           <if variable="DOI">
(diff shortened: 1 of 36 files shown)


rmzelle:
@adam3smith, feel free to merge. Simple find-and-replace of "http://doi.org/" by "https://doi.org/".

(I noticed that https://github.com/citation-style-language/styles/pull/2241/files#diff-15aaee93c6e5331268bb6e7773a9f146R99 is a bit funky with `prefix=" DOI:https://doi.org/"`; haven't checked if they actually require that)

adam3smith:
cool, thanks

PR #2246, opened 2016-09-29
Fixing the citation format for JID.

cparnot:
Using square brackets instead of parentheses.


rmzelle:
Thanks Charles!

cparnot:
Awesome, that was fast. And for the record, this was reported by a Zotero user to me.

PR #2247, opened 2016-09-29
Report parser errors

inukshuk:
This should print the respective parse error when tests could not be parsed. Should make errors like duplicate macros much easier to spot!


rmzelle:
Thanks!

PR #2250, opened 2016-09-29
Update cahiers-d-ethnomusicologie.csl

svictor9:
1. Corrected a couple discrepancies in `<citation>` between the style and the journal’s guidelines. Should now look like this: (Rouget 1990 : 247)

2. Added missing whitespace between journal title and issue number for articles


adam3smith:
Thanks!

PR #2252, opened 2016-09-29
Update national-library-of-medicine-grant-proposals.csl

CELISEROGERS:
only change is to make citations superscript


adam3smith:
does NLM not care one way or the other?

CELISEROGERS:
not sure what you mean but zotero had instructions on how to change this as other users were asking for this as well, it can be named something else but those writing nih grant proposals need a style that brings in pmcids and superscript takes up less room than brackets

adam3smith:
OK, let's see if anyone protests and we can always revert this.

PR #2253, opened 2016-09-30
SIGCHI proceedings format shouldn't default to "et al"

mporcheron:
The SIGCHI reference format is based on the ACM reference format and this does not include "et al" for any references (that I can find). Given no indication otherwise, I don't think the SIGCHI reference format should truncate the list of authors either.

Example with current CSL file:
csl-current

Example from the ACM Digital Library of the ACM reference:
acm

Example with the modified CSL file:
csl-proposed

### DIFF
--- a/acm-sigchi-proceedings.csl
+++ b/acm-sigchi-proceedings.csl
@@ -115,7 +115,7 @@
       </group>
     </layout>
   </citation>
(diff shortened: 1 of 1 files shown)


rmzelle:
@adam3smith, looks right to me. (can't find any instructions on or examples of the use of et-al in bibliographic references in the style guide, and the guide says the style is close to what the ACM Digital Library produces)

adam3smith:
OK, sounds good, thanks!

PR #2263, opened 2016-10-05
Create vigiliae-christianae.csl

carljohanberglund:
New CSL style for Vigiliae Christianae (ISSN 0042-6032; http://brill.com/vc)

### DIFF
--- a/vigiliae-christianae.csl
+++ b/vigiliae-christianae.csl
@@ -16,41 +16,36 @@
     <issn>0042-6032</issn>
     <eissn>1570-0720</eissn>
     <summary>Vigiliae Christianae</summary>
-    <updated>2016-10-06T09:13:00+00:00</updated>
+    <updated>2016-10-07T13:08:50+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
     <terms>
-      <term name="editor" form="verb-short">ed.</term>
-      <term name="translator" form="verb-short">trans.</term>
+      <term name="editor" form="short">ed.</term>
+      <term name="translator" form="short">trans.</term>
       <term name="at">here</term>
     </terms>
   </locale>
-    <locale xml:lang="de">
+  <locale xml:lang="de">
     <terms>
-      <term name="editor" form="verb-short">Hrsg.</term>
-      <term name="translator" form="verb-short">Übers.</term>
+      <term name="editor" form="short">Hrsg.</term>
+      <term name="translator" form="short">Übers.</term>
       <term name="at">hier</term>
     </terms>
   </locale>
   <locale xml:lang="fr">
     <terms>
-      <term name="editor" form="verb-short">éd.</term>
-      <term name="translator" form="verb-short">trad.</term>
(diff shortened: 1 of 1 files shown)


vigiliae-christianae.csl
```diff
@@ -0,0 +1,776 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" page-range-format="chicago" demote-non-dropping-particle="sort-only">
```
rmzelle: @carljohanberglund, it looks like this journal publishes in English, German, and French, so we probably don't want to fix the style to a single language, right?

carljohanberglund: French and German terms added.

vigiliae-christianae.csl
```diff
@@ -0,0 +1,790 @@
...
+    <text variable="collection-title"/>
+    <text variable="collection-number" prefix=" "/>
+  </macro>
+  <macro name="container-title">
+    <choose>
+      <if type="chapter paper-conference" match="any">
+        <group delimiter=" " suffix=" ">
+          <label variable="page" form="long" text-case="capitalize-first"/>
+          <text variable="page"/>
+          <text term="in"/>
```
adam3smith: I'd rather see the in in a group with the `container-title` below. Otherwise in the (admittedly rare, but esp. for conference papers not unthinkable) case that the container title is missing, the in will just hang in the air...

vigiliae-christianae.csl
```diff
@@ -0,0 +1,790 @@
...
+        <text macro="point-locators-subsequent" prefix=", "/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="pages">
+    <choose>
+      <if type="article-journal">
+        <text variable="page" prefix=" "/>
+      </if>
+      <else-if type="chapter paper-conference" match="any">
```
adam3smith: this seems pointless -- looks like you could just do `<text variable="page" prefix=" "/>` for the entire macro, but if there are some item types (magazine and newspaper articles??) that you do want to exclude, you definitely only need one if and not an else-if with the same content.

vigiliae-christianae.csl
```diff
@@ -0,0 +1,790 @@
...
+      <term name="editor" form="verb-short">éd.</term>
+      <term name="translator" form="verb-short">trad.</term>
+      <term name="at">ici</term>
+    </terms>
+  </locale>
+<locale>
+	  <terms>
+  		<term name="at">hic</term>
+  	</terms>
+  </locale>
```
rmzelle: @carljohanberglund, did you leave in lines 43-47 intentionally? (that define `<term name="at">hic</term>`)

carljohanberglund: Sometimes it's nice to have fallback terms in Latin, but I don't think it's necessary in this case.

vigiliae-christianae.csl
carljohanberglund: I dit not think of that case. Fixed now.

vigiliae-christianae.csl
carljohanberglund: That's true. Fixed now.

adam3smith:
@rmzelle all good by me. I'll let you merge.

rmzelle:
Thanks both!

PR #2265, opened 2016-10-06

Yusufras:
```
- <et-al term="and others" font-style="italic"/>
+ <et-al font-style="italic"/>

- <text variable="page" form="long" prefix=":"/>
+ <text variable="page" form="long" prefix=", "/>
```


rmzelle:
@Yusufras, I made a few small changes to simplify the style, and changed the DOI link from "http" to "https" (which is preferred).

PR #2266, opened 2016-10-06
Corrected use of et al. in author lists

furiousd:
Latest IEEE style guide says "If there are more than six names listed, use et al. after the first author." (p. 34)

[IEEE_style_manual.pdf](https://github.com/citation-style-language/styles/files/514202/IEEE_style_manual.pdf)


rmzelle:
Great, thanks. Let us know if you find any more discrepancies against the latest style guide (or feel free to create additional pull request).

I just removed the "prefix" on the et-al element, which is implicit.

PR #2270, opened 2016-10-14
Create sexual-health.csl

CRMTH:
Created new style for Sexual Health journal.


rmzelle:
@CRMTH, it looks like you already figured this out, but we prefer it if you update an existing pull request instead of creating a new one. Continued in #2271.

PR #2271, opened 2016-10-14
Create sexual-health.csl

CRMTH:
Created CSL template for Sexual Health journal.


rmzelle:
Thanks. Is there an existing CSL style you based this one on? If so, we'd like to link to that style with a "template" link (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link).

CRMTH:
We've added the template link to the code.

rmzelle:
Great, thanks!

PR #2273, opened 2016-10-17
Update nccr-mediality.csl

cnotz:
Short citations without italics (as in stylesheet) and prefix for second citations.

### DIFF
--- a/nccr-mediality.csl
+++ b/nccr-mediality.csl
@@ -283,7 +283,7 @@
   <macro name="access">
     <choose>
       <if variable="URL">
-        <text prefix="," variable="URL"/>
+        <text prefix=", " variable="URL"/>
         <group prefix=", " delimiter=" ">
           <text term="accessed" text-case="lowercase"/>
           <date variable="accessed">
--- a/nccr-mediality.csl
+++ b/nccr-mediality.csl
@@ -14,7 +14,7 @@
     <category citation-format="note"/>
     <category field="humanities"/>
     <summary>Style for NCCR Mediality.</summary>
-    <updated>2016-10-17T09:03:48+00:00</updated>
+    <updated>2016-10-18T10:51:48+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="de">
@@ -283,7 +283,7 @@
   <macro name="access">
     <choose>
       <if variable="URL">
-        <text variable="URL"/>
+        <text prefix="," variable="URL"/>
         <group prefix=", " delimiter=" ">
           <text term="accessed" text-case="lowercase"/>
           <date variable="accessed">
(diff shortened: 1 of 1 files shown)


nccr-mediality.csl
```diff
@@ -311,6 +311,7 @@
         <else-if position="subsequent">
           <text macro="contributors-short" suffix=", "/>
           <text macro="title-short"/>
+          <text variable="first-reference-note-number" prefix=" (Anm. " suffix=")"/>
```
rmzelle: @cnotz, just curious, but what does "Anm" stand for here?

@adam3smith, do you know if we have a CSL term for this?

adam3smith:
"Anm." stands for "Anmerkung," i.e. "note." No CSL term, though I think there should be (we [hardcode the equivalent](https://github.com/citation-style-language/styles/blob/master/oscola.csl#L594) in English styles, too).

nccr-mediality.csl
```diff
@@ -174,7 +174,7 @@
     <group delimiter=", ">
       <choose>
         <if type="book">
-          <text variable="title" form="short" font-style="italic"/>
+          <text variable="title" form="short"/>
```
adam3smith: give this, let's simplify the macro. You can reduce it to a single line, i.e. `<text variable="title" form="short"/>`

adam3smith:
One small nit above. Everything else looks good. Thanks

rmzelle:
> "Anm." stands for "Anmerkung," i.e. "note." No CSL term, though I think there should be (we hardcode the equivalent in English styles, too).

There is a "note" term, though. (see e.g. https://github.com/citation-style-language/locales/blob/020712589f08bd74adf53c99f808fbbefebd463f/locales-en-US.xml#L191)

cnotz:
Thank you for your critical eyes and your help simplificating this style!

nccr-mediality.csl
```diff
@@ -310,8 +300,9 @@
         </else-if>
         <else-if position="subsequent">
           <text macro="contributors-short" suffix=", "/>
-          <text macro="title-short"/>
-          <text variable="first-reference-note-number" prefix=" (Anm. " suffix=")"/>
+          <text variable="title" form="short"/>
+          <text term="note" form="short" prefix=" (" suffix="."/>
+          <text variable="first-reference-note-number" prefix=" " suffix=")"/>
```
rmzelle: @cnotz, it's better to replace

```xml
          <text term="note" form="short" prefix=" (" suffix="."/>
          <text variable="first-reference-note-number" prefix=" " suffix=")"/>
```

by

```xml
        <group prefix=" (" suffix=")" delimiter=" ">
          <text term="note" form="short"/>
        </group>
```

and change

```
<term name="note" form="short">Anm</term>
```

by

```
<term name="note" form="short">Anm.</term>
```

That clarifies how the parentheses belong together (it's a good habit to use groups wherever possible in styles for placing punctuation). (this `<group/>` would also prevent the "note" term from being printed if the "first-reference-note-number" variable doesn't have a value, although I guess that shouldn't really be possible for `position="subsequent"`)

In our world of CSL we also always define abbreviated terms with periods.

adam3smith:
@rmzelle I put in the changes you suggested, go ahead and merge if you're good with this otherwise.

rmzelle:
Great, thanks.

cnotz:
Thank you so much for your work!
> Am 21.10.2016 um 16:52 schrieb csl-bot <notifications@github.com>:
> 
> 😟

PR #2274, opened 2016-10-17
Create studia-historica-upsaliensa.csl

torfeur:
Style for student essays and dissertation series of the department of History, Uppsala University. According to the department style guidelines. Contains some "extras": Formats item type "Manuscript" and any item type with information in BOTH of the fields "Archive" and "Location In Archive" separately in the bibliography, in order to facilitate the use of Zotero (and others) to cite archival sources.


rmzelle:
@torfeur, could you remove the two macros in the style that aren't used? They're "archive" and "archive-note".

studia-historica-upsaliensa.csl
```diff
@@ -0,0 +1,588 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="note" version="1.0" and="symbol" initialize="false" demote-non-dropping-particle="never" xmlns="http://purl.org/net/xbiblio/csl">
```
adam3smith: Is this style used in multiple languages  or just in English (or Swedish)? IN the latter case, we'd want a default-locale here.

rmzelle: (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#3---style-locale)

studia-historica-upsaliensa.csl
```diff
@@ -0,0 +1,588 @@
...
+<!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>Studia Historica Upsaliensa </title>
+    <title-short>Studia</title-short>
+    <id>http://www.zotero.org/styles/studia-historica-upsaliensa</id>
+    <link href="http://www.zotero.org/styles/studia-historica-upsaliensa" rel="self"/>
+    <link href="http://www.zotero.org/styles/chicago-fullnote-bibliography" rel="template"/>
+    <link href="http://www.hist.uu.se" rel="documentation"/>
```
rmzelle: Is there an actual (online or offline) document that describes this citation format? Giving the URL of the department helps, but a direct link to the style guide would be more useful.

studia-historica-upsaliensa.csl
```diff
@@ -0,0 +1,588 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="note" version="1.0" and="symbol" initialize="false" demote-non-dropping-particle="never" xmlns="http://purl.org/net/xbiblio/csl">
+<!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>Studia Historica Upsaliensa </title>
```
rmzelle: We name all our university department styles as `<title>name of university - name of department</title>`. Would `<title>Uppsala universitet - Historiska institutionen</title>` be acceptable? (if so, the style's file name, style ID, and "self" link would have to be updated to match)

studia-historica-upsaliensa.csl
torfeur: I thought about that. Yes there is [http://hist.uu.se/Student/Masterutbildning/Skrivauppsats/Formaliainstruktioner/tabid/5600/language/en-US/Default.aspx](url)
The problem is that the information on that pages changes URL every time we update our homepage, something we do every three to four years. A few years from now the link above will most certainly be dead, but the information will still be found somewhere else on hist.uu.se

studia-historica-upsaliensa.csl
torfeur: I changed it. I see the difference between a common university style and a department specific one. So please disregard the question above. Dropped the "department" part of your suggestion. Seems to be inline with the naming of similar department styles in the repos so I hope that is fine with you.

studia-historica-upsaliensa.csl
adam3smith: Yes, this is good as is, then.

adam3smith:
The test fails because of the filename -- but I actually thought your case for keeping the name&ID as is was compelling. If there is such a series, why not use that as the style name (as you originally had)? @rmzelle thoughts?

studia-historica-upsaliensa.csl
torfeur: I changed the documentation link to the full description in my comment above.

torfeur:
I thought so to, but once I wrote it out it looks better. I'm usually in favor of titles that reflects the contents. Pretty much the same with both titles but nobody reads latin these days. Do I need to make a new pull request for the style with a new name?

adam3smith:
no -- I changed it for you (but you could have done so, too). I'm OK with this as is, then. I'll let @rmzelle  do the final review&merge

torfeur:
Thanks for your help! Much appreciated.

rmzelle:
> since when do you name all university department styles that way? I got the name from a style that you once wrote for the Swedish Univerisity of Agriculture . "Acta something something..." Studia Historia Upsaliensia" is the name of the dissertation series at the History department. A "full" title would be "Acta Universitatis Upsaliensis. Studia Historica Upsaliensia".

Well, maybe not all, but I was thinking of styles like these: https://www.zotero.org/styles?q=%20-%20facul. There can be valid reasons to name a style differently, but my goal is just to try to standardize style names as much as possible to make them easier to find.

If you're okay with the current name, let's stick with that. It also matches the [style guide](http://www.hist.uu.se/LinkClick.aspx?fileticket=ZORjavT%2fp%2bI%3d&tabid=3825&mid=15459&language=sv-SE) which says "Uppsala universitet" and "Historiska institutionen" on the front page.

Just a quick question, is the "Historia" correct? I noticed it's different from "Historiska". Is that just a declension difference? I'm fine with it, as long as it is intentional.

(the "Acta something something" style is https://github.com/citation-style-language/styles/blob/master/acta-universitatis-agriculturae-sueciae.csl, by the way)

torfeur:
Thanks for you comments and help.

I changed the name to "Uppsala universitet - Historia". There were plenty of examples of that too in the repos. I understand your quest for better names though. 

I don't know the correct grammatical terms for the difference between "Historiska" and "Historia", but the former is defining the word that follows ("institutionen"), where as the latter is a noun. "Historia" is of course "History", "Historiska" would be "of History" or something similar.
So yes ( a long answer to the short question)

Again, thank's!
(By the way - the documentation for that Acta style of yours links to [slu.se](url), that's hardly very specific...)

rmzelle:
Okay, thanks. Merged. Thanks for contributing.

> (By the way - the documentation for that Acta style of yours links to slu.se, that's hardly very specific...)

Yes, I don't think there was an online style guide for that style.

PR #2276, opened 2016-10-20
Update deutsches-archaologisches-institut.csl

bellerophons-pegasus:
Added display of booktitles for chapters. Changed 'et al.' to 'u. a.'. Stripped periods from title fields. Display of '(Hrsg.)' instead of '(eds.)'. Only display edition if number and superscripted. No commas before page numbers, except for articles. Included series number (and series title) for publications in series.
One problem left: In chapters the editors of the book are not displayed correctly; e.g. it should be 'M. Alexianu' and not 'Alexianu, Marius'

### DIFF
--- a/deutsches-archaologisches-institut.csl
+++ b/deutsches-archaologisches-institut.csl
@@ -4,8 +4,8 @@
   <info>
     <title>Deutsches Archäologisches Institut (German)</title>
     <title-short>DAI</title-short>
-    <id>http://www.zotero.org/styles/deutsches-archaologisches-institut</id>
-    <link href="http://www.zotero.org/styles/deutsches-archaologisches-institut" rel="self"/>
+    <id>http://www.zotero.org/styles/deutsches-archäologisches-institut-german</id>
+    <link href="http://www.zotero.org/styles/deutsches-archäologisches-institut-german" rel="self"/>
     <link href="http://www.zotero.org/styles/american-journal-of-archaeology" rel="template"/>
     <link href="http://www.dainst.org/publikationen/publizieren-beim-dai/richtlinien" rel="documentation"/>
     <author>
@@ -19,7 +19,7 @@
     <category citation-format="author-date"/>
     <category field="anthropology"/>
     <summary>New author-date style meant to meet citation specifications provided by DAI</summary>
-    <updated>2016-10-20T21:26:52+00:00</updated>
+    <updated>2016-10-25T15:48:55+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="secondary-contributors">
@@ -53,7 +53,7 @@
           <choose>
             <if variable="editor container-title" match="any">
               <text term="in" prefix=", " suffix=": "/>
-              <names variable="editor" font-variant="small-caps">
+              <names variable="editor">
                 <name delimiter=" &#8211; " delimiter-precedes-et-al="never" delimiter-precedes-last="always" initialize-with="."/>
                 <label form="short" prefix=" (" suffix=")" strip-periods="false"/>
               </names>
@@ -186,8 +186,11 @@
       <if type="bill book graphic legal_case motion_picture report song chapter paper-conference" match="any">
         <choose>
           <if is-numeric="edition">
-            <text variable="edition" vertical-align="sup"/>
+            <text variable="edition" vertical-align="sup" prefix=" "/>
           </if>
+          <else>
+            <text value=" "/>
+          </else>
         </choose>
       </if>
     </choose>
@@ -252,11 +255,6 @@
   <macro name="place-date">
     <choose>
       <if type="book thesis chapter" match="any">
-        <choose>
-          <if match="none" is-numeric="edition">
-            <text value=" "/>
-          </if>
-        </choose>
         <text variable="publisher-place" prefix="(" suffix=" "/>
         <text macro="date" suffix=")"/>
       </if>
@@ -348,7 +346,7 @@
         <text macro="locators"/>
         <text macro="issue"/>
         <text macro="series"/>
-        <text macro="edition" prefix=", "/>
+        <text macro="edition"/>
         <text macro="place-date"/>
         <text macro="journal-date"/>
         <text macro="locators-chapter"/>
--- a/deutsches-archaologisches-institut.csl
+++ b/deutsches-archaologisches-institut.csl
@@ -7,7 +7,7 @@
     <id>http://www.zotero.org/styles/deutsches-archaologisches-institut</id>
     <link href="http://www.zotero.org/styles/deutsches-archaologisches-institut" rel="self"/>
     <link href="http://www.zotero.org/styles/american-journal-of-archaeology" rel="template"/>
-    <link href="http://www.dainst.org/it/node/28770?ft=3" rel="documentation"/>
+    <link href="http://www.dainst.org/publikationen/publizieren-beim-dai/richtlinien" rel="documentation"/>
     <author>
       <name>Adam Rabinowitz</name>
       <email>adam.rabinowitz@gmail.com</email>
--- a/deutsches-archaologisches-institut.csl
+++ b/deutsches-archaologisches-institut.csl
@@ -2,7 +2,7 @@
 <style class="in-text" version="1.0" delimiter-precedes-et-al="never" initialize-with="." demote-non-dropping-particle="sort-only" default-locale="de" xmlns="http://purl.org/net/xbiblio/csl">
 <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/#) -->
   <info>
-    <title>Deutsches Archäologisches Institut</title>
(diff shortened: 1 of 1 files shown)


deutsches-archaologisches-institut.csl
```diff
@@ -225,10 +220,10 @@
   <macro name="locators-article">
     <choose>
       <if type="article-newspaper">
-        <group prefix=", " delimiter=", ">
+        <group delimiter=", ">
           <group delimiter=" ">
-            <text variable="edition"/>
-            <text term="edition"/>
+            <text variable="edition" prefix=", "/>
```
adam3smith: This is almost certainly wrong: 1) why would you have commas between the edition variable and the term? "Ausgabe, Hamburg"? But even if that's what you want, with the space delimiter and the spaces in prefix, you'd get "Ausgabe , Hamburg" which is definitely wrong.

deutsches-archaologisches-institut.csl
```diff
@@ -292,7 +278,7 @@
     </date>
   </macro>
   <macro name="collection-title">
-    <text variable="collection-title"/>
+    <text variable="container-title" strip-periods="true"/>
```
adam3smith: this seems unfortunate: a macro that's called collection-title shouldn't just be calling `container-title`. I'd get rid of the macro entirely and just use the `<text variable="container-title" strip-periods="true"/>` wherever you need it.

deutsches-archaologisches-institut.csl
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+<style class="in-text" version="1.0" delimiter-precedes-et-al="never" initialize-with="." demote-non-dropping-particle="sort-only" default-locale="de" xmlns="http://purl.org/net/xbiblio/csl">
+<!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/#) -->
   <info>
     <title>Deutsches Archäologisches Institut</title>
```
adam3smith: Since you're switching the style to German, could you change the title to `Deutsches Archäologisches Institut (German)` ? No need to change id/filename etc.

deutsches-archaologisches-institut.csl
bellerophons-pegasus: I hope I fixed it. The DAI has no special rules for newspaper articles

deutsches-archaologisches-institut.csl
bellerophons-pegasus: Got rid of it, I was using it on one place anyway...

bellerophons-pegasus:
Also changed the documentation link. It points to the page were all available guidelines can be downloaded

deutsches-archaologisches-institut.csl
```diff
@@ -1,20 +1,25 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+<style class="in-text" version="1.0" delimiter-precedes-et-al="never" initialize-with="." demote-non-dropping-particle="sort-only" default-locale="de" xmlns="http://purl.org/net/xbiblio/csl">
```
rmzelle: @bellerophons-pegasus, can you change `default-locale="de"` to `default-locale="de-DE"`? (that way the style explicitly uses the "de-DE" CSL locale file from https://github.com/citation-style-language/locales/)

deutsches-archaologisches-institut.csl
```diff
@@ -23,17 +28,17 @@
         <group delimiter=". ">
           <choose>
             <if variable="author">
-              <names variable="editor">
+              <names variable="editor" font-variant="normal">
```
rmzelle: You should be able to delete the ` font-variant="normal"` here (which is the default).

deutsches-archaologisches-institut.csl
```diff
@@ -23,17 +28,17 @@
...
                 <name delimiter=" &#8211; " initialize-with=". "/>
-                <label form="short" prefix=" (" suffix=")" strip-periods="true"/>
+                <label form="short" strip-periods="true" prefix=" (" suffix=")"/>
               </names>
             </if>
           </choose>
           <choose>
             <if variable="author editor" match="any">
-              <names variable="translator">
+              <names variable="translator" font-variant="normal">
```
rmzelle: You should again be able to delete the ` font-variant="normal"` here (which is the default).

deutsches-archaologisches-institut.csl
```diff
@@ -216,7 +211,7 @@
   <macro name="locators-chapter">
     <choose>
       <if type="chapter paper-conference" match="any">
-        <group prefix=", ">
+        <group prefix=" ">
           <text variable="page"/>
         </group>
```
rmzelle: You can just remove the group here and use `<text variable="page" prefix=" "/>`

deutsches-archaologisches-institut.csl
```diff
@@ -251,37 +246,28 @@
...
   <macro name="place-date">
     <choose>
-      <if type="book thesis" match="any">
-        <text variable="publisher-place" prefix=" (" suffix=" "/>
+      <if type="book thesis chapter" match="any">
+        <choose>
+          <if match="none" is-numeric="edition">
+            <text value=" "/>
+          </if>
+        </choose>
```
rmzelle: This conditional prefix here is rather awkward.

In general, it's much better to have macros output a block of content without any separating prefixes or suffixes surrounding the block. E.g. "(Berlin, 2009)" instead of " (Berlin, 2009)". You can then control the delimiters when calling the macros:

```
        <choose>
          <if is-numeric="edition">
            <group delimiter=", ">
              <text macro="series"/>
              <text macro="edition"/>
            </group>
            <text macro="place-date"/>
          </if>
          <else>
            <group delimiter=" ">
              <text macro="series"/>
              <text macro="place-date"/>
             </group>
          </else>
        </choose>
```

That makes it much easier to understand what's going on.

deutsches-archaologisches-institut.csl
bellerophons-pegasus: Deleted. Might have come from the CSL-editor

deutsches-archaologisches-institut.csl
bellerophons-pegasus: Changed

deutsches-archaologisches-institut.csl
rmzelle: (and it would be good to check if the code above works as desired; I couldn't find any example references at the documentation link, and hand-wrote the blurb without testing)

bellerophons-pegasus: Hm.
The thing is, the edition number, if available, is placed directly in front of the 'place-date' without any spacing. Otherwise I need a space. I've added the space as a condition in the group for the edition instead.

rmzelle: > The thing is, the edition number, if available, is placed directly in front of the 'place-date' without any spacing.

That's what

```
            <group delimiter=", ">
              <text macro="series"/>
              <text macro="edition"/>
            </group>
            <text macro="place-date"/>
```

should give, as long as these macros don't add any more spaces themselves.

rmzelle:
@adam3smith, let me know if you're okay with the style like this.

deutsches-archaologisches-institut.csl
```diff
@@ -23,17 +28,17 @@
         <group delimiter=". ">
           <choose>
             <if variable="author">
-              <names variable="editor">
-                <name delimiter=" &#8211; " initialize-with=". "/>
-                <label form="short" prefix=" (" suffix=")" strip-periods="true"/>
+              <names variable="editor" font-variant="normal">
+                <name delimiter=" – " initialize-with=". "/>
```
bellerophons-pegasus: Yes, it came from the csl editor. It is changed now.

deutsches-archaologisches-institut.csl
```diff
@@ -46,24 +51,20 @@
       <if type="chapter paper-conference" match="any">
         <group>
           <choose>
-            <if variable="author">
-              <text value="in" prefix=" " suffix=": "/>
-              <names variable="editor">
-                <name form="short" delimiter=" &#8211; " delimiter-precedes-last="always"/>
+            <if variable="editor container-title" match="any">
+              <text value=", in" suffix=": "/>
```
bellerophons-pegasus: Changed

bellerophons-pegasus:
Is there anything left to do for me?

adam3smith:
I won't get to this until the weekend. @rmzelle if you feel good about this, go ahead and merge. Otherwise I'll do that on Sunday.

rmzelle:
@bellerophons-pegasus, thanks! If you ever feel adventurous in the future, you can try to reorganize the style a little to rely more on prefixes, suffixes, and delimiters on `<group/>` elements, as I suggested at https://github.com/citation-style-language/styles/pull/2276#discussion_r84923684, but if it works, it works.

bellerophons-pegasus:
Awesome! Thank you. And I will keep that in mind.

PR #2277, opened 2016-10-23
Separate abstract from the rest using full stop in APA.

wilx:
Abstract in APA annotated style is not separated from previous
item. This change adds a full stop and a space prefix before the
abstract.


adam3smith:
Thanks, but the `display="block"` elements puts the abstract on a new line, so there should not and should not need to be a space here. Here's e.g. how this looks in Zotero (created via the LibreOffice add-on).
screenshot from 2016-10-22 21-33-50

Where this doesn't work, this is an implementation/processor issue and you should report it to the software you're using.

PR #2278, opened 2016-10-23
Updates to harvard-coventry-university.csl

tipichris:
Updates to disambiguation rules following clarification from style guide authors on the intention.

### DIFF
--- a/harvard-coventry-university.csl
+++ b/harvard-coventry-university.csl
@@ -280,7 +280,7 @@
       </if>
     </choose>
   </macro>
-  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-givenname="true" givenname-disambiguation-rule="all-names" name-as-sort-order="all" collapse="year">
+  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-givenname="true" givenname-disambiguation-rule="by-cite" name-as-sort-order="all" collapse="year">
     <layout prefix="(" suffix=")" delimiter=", ">
       <choose>
         <if locator="page">
--- a/harvard-coventry-university.csl
+++ b/harvard-coventry-university.csl
@@ -94,7 +94,7 @@
       </else-if>
       <else>
         <names variable="author">
-          <name name-as-sort-order="all" and="text" sort-separator=", " initialize-with="." delimiter=", "/>
+          <name name-as-sort-order="all" and="text" sort-separator=", " initialize-with="." delimiter=", " form="long"/>
           <label form="short" prefix=" " strip-periods="false"/>
           <substitute>
             <text macro="editor"/>


adam3smith:
Thanks! I've left the `form="long"` in there, but that shouldn't actually change anything. Disambiguation will automatically go to full names where initals aren't enough.

tipichris:
When I read the CSL docs it seemed `form="long"` wasn't needed, but empirically it was. Without it Jane Smith and John Smith get disambiguated with full names in the citation, but not in the list of references. At least in Zotero's csledit.xul. Perhaps something else in the style sheet is getting the way?

rmzelle:
@adam3smith, I made a shorter version of the style for troubleshooting: https://gist.github.com/rmzelle/f1f6545f7dde6419d7dfd3933209d19e

Behavior does seem a bit wonky in Zotero for Firefox 4.0.29.15. When citing a single item, I get `(Smith)` and `J. Smith` with or without `form="long"` on the "author" macro. When citing two ambiguous items, I get `(Joe Smith, Jane Smith)` and either `J. Smith` and `J. Smith` (no `form` attribute on the "author" macro) or `Joe Smith` and `Jane Smith` (`form="long"` on the "author" macro).

@fbennett?

I also noticed that http://docs.citationstyles.org/en/stable/specification.html#disambiguation doesn't seem to mention how disambiguation in cites should affect name expansion in the bibliography.

fbennett:
I built a test fixture around the troubleshooting version and confirmed that behavior ... but most of that looks correct, and the last bit (which would be an error) I can't reproduce.
* With a single item, or without disambiguation, `initialize-with-"true"` shortens `form="long"` (whether by default or explicitly declared) to initials.
* With `disambiguate-add-givenname="true"`, the default disambiguation rule, and the names in separate citations, those names will fully expand to `John Smith` and `Jane Smith` in the citation.
* Under the same conditions, the names should fully expand in the bibliography, with or without `form="long"`. Initials would definitely be wrong, but I don't get that result when `form` is omitted, at least in the latest version of the processor.

fbennett:
It's possible that Zotero is running an earlier version of the processor. A similar test against that release tag would show whether it's a transient processor bug.

rmzelle:
@fbennett, according to https://www.zotero.org/support/4.0_changelog, Zotero for Firefox 4.0.29.15 is running citeproc-js 1.1.136.

fbennett:
Weird one.
* I can reproduce the fault [1] in Zotero and in Juris-M under Firefox 47.
* I can reproduce the fault in the processor demo page, when viewed through Firefox 47 *or* through Google Chrome.
* I *cannot* reproduce the fault in the test framework, at least using any of the (older) JS shells that I have.

Success (at failure) in the demo page is a step forward. I'll keep working on it, and post back when I've at least come up with a plaster bandage to solve the problem.

[1] "J. Smith" in the bibliography, with disambiguation-add-givenname="true", no form attribute on the cs:names in the bibliography names macro, and a pair of ambiguous names as input.

fbennett:
Looking through the code, I can't see why it *does* work correctly in the test framework. >sigh<

However, there is a bright spot---it's not necessary to trace that out. Setting a global name-form="long" on the cs:style node fixes things in the browser, following the expected logic. If I force that parameter in as a default value at processor instantiation, this should clear. Errands this afternoon, but I'll get that in place this evening.

More soon(ish) ...

fbennett:
The fix I ruminated about above works, but it causes one existing test to fail. It's an edge case, and I'd better ask for guidance before finalizing the fix. The test is [here](https://github.com/citation-style-language/test-suite/blob/master/processor-tests/humans/disambiguate_AllNamesWithInitialsBibliography.txt).

The test applies the `givenname-disambiguation-rule="all-names-with-initials"`. The effect of that rule is to attempt disambiguation on all names, but to stop at initials, without attempting to expansion to the full name.

With the fix, a default value of `name-form="long"` is set on `cs:style`, so that the `form` attribute on `cs:name` is never undefined. Previously, with an undefined value for `form`, this test apparently always added initials, as shown in the current result string:
``` html
<div class="csl-bib-body">
  <div class="csl-entry">Asthma, A., Bronchitis, B., and Cold, C. (1980)</div>
  <div class="csl-entry">Asthma, A., Bronchitis, B., and Cold, O. (1995)</div>
</div>
```

 After the fix, the processor is (paradoxically) *ignoring* the `form="long"` attribute and (not-so-paradoxically) applying the rule as intended, to produce this result:
``` html
<div class="csl-bib-body">
  <div class="csl-entry">Asthma, Bronchitis, and Cold, C. (1980)</div>
  <div class="csl-entry">Asthma, Bronchitis, and Cold, O. (1995)</div>
</div>
```
The parameters are in conflict in this case, so it could be thrown either way. I think that the latter, post-fix output is a better result, but it does change processor behavior, and it may change the output in some existing styles.

It would be possible to preserve the existing behavior with another small tweak to the processor ... which way would you-all like to go with it?

rmzelle:
I think @adam3smith will agree with me that

```xml
<div class="csl-bib-body">
  <div class="csl-entry">Asthma, Bronchitis, and Cold, C. (1980)</div>
  <div class="csl-entry">Asthma, Bronchitis, and Cold, O. (1995)</div>
</div>
```

is the preferred result. I don't completely understand your actual fix: http://docs.citationstyles.org/en/stable/specification.html#name already says that `form="long"` is the default value for `<name/>`, so it's not clear to me why you need to have the processor place `name-form` on `<style/>`. Is that the approach you take for all default values throughout the CSL spec?

fbennett:
The small code change I made has that effect, since "long" will then percolate down unless it is overridden on a subordinate node. I'm really just talking to myself there about the implementation---it would have been enough to say that the processor now follows the spec.

I'll push the change soon.

PR #2279, opened 2016-10-24
Create harvard-orient-institut-beirut.csl

shyam-k:


rmzelle:
I noticed that this style was submitted twice by two different GitHub accounts (here and in #2281). We can deal with this submission in either issue, but one of them should be closed.

rmzelle:
Hi. A few questions:

* Is there an online description of this citation format?
* Does the OIB use more than one citation format (e.g. a "numeric" and an "author-date" variant)? If it's just this format, we would typically remove the "Harvard" from the style title.
* Will this CSL style be used in more than one language?

shyam-k:
Hi 

- No there is no description for this citation format
- Yes there is one variant with  descending date (https://github.com/citation-style-language/styles/pull/2283)
- No

rmzelle:
@shyam-k, thanks.

So will these styles only be used in English, or another language?

Can you also elaborate a little on the purpose of these styles? Are both variants (this one and the descending date one) meant to be used for print/online publications? Or will they, for example, only be used to display Zotero libraries on the OIB website?

shyam-k:
Hi @rmzelle, Thanks.
Yes. It should be used in English, German and Arabic languages.

There is no description available for this style.The style is based on the Harvard citation styles (hence the name) and tailored to the needs of OIB's( Orient-Institut Beirut (OIB) is a non-profit German academic research institute) annual report and website.
Ideally, the styles will be used for multiple purposes: starting with the OIB's annual report and its website. I would aim for broader application, but this is currently not yet relevant.

rmzelle:
> tailored to the needs of OIB's( Orient-Institut Beirut (OIB) is a non-profit German academic research institute) annual report and website.
> Ideally, the styles will be used for multiple purposes: starting with the OIB's annual report and its website. I would aim for broader application, but this is currently not yet relevant.

So, at least for now, these two styles would only be used internally within OIB? In that case it might be better to just distribute the style yourself within the organization. We think that the primary role of the CSL repository is to distribute styles for cases where the style author doesn't know the users of the style. In contrast, if a style is e.g. used by a publication with open submissions, we almost always accept it.

I hope that sounds reasonable. See also https://github.com/citation-style-language/styles/wiki/Criteria-for-Accepting-Styles.

rmzelle:
I'll close this for now, then. Once you start using this CSL style for OIB publications with open submissions, feel free to post here again.

PR #2281, opened 2016-10-24
Create harvard-orient-institut-beirut.csl

rajeshvr:
OIBR csl style added


rmzelle:
Hi. A few questions:

* Is there an online description of this citation format?
* Does the OIB use more than one citation format (e.g. a "numeric" and an "author-date" variant)? If it's just this format, we would typically remove the "Harvard" from the style title.
* Will this CSL style be used in more than one language?

rmzelle:
Also, I noticed that this style was submitted twice by two different GitHub accounts (here and in #2279). We can deal with this submission in either issue, but one of them should be closed.

rajeshvr:
I am closing this request as one of my colleagues also made a pull request of the same

rmzelle:
Okay, thanks.

PR #2283, opened 2016-10-25
harvard-orient-institut-beirut-desc.csl

rajeshvr:
This style is for descending format of oibr zotero library


rmzelle:
Closing. See #2279 for the reason.

PR #2285, opened 2016-10-28
Create Journal of Radiation Research

char-llie:
Only validation of existing style


rmzelle:
@char-llie, can you explain what you were trying to do? The style you tried to submit already exists in our repository (at https://github.com/citation-style-language/styles/blob/master/dependent/journal-of-radiation-research.csl).

PR #2288, opened 2016-10-28
Update geochronometria.csl

dbobak:
Removed suffix from the "genre" text variable

### DIFF
--- a/geochronometria.csl
+++ b/geochronometria.csl
@@ -17,7 +17,7 @@
     <issn>1733-8387</issn>
     <eissn>1897-1695</eissn>
     <summary>Geochronometria citation style.</summary>
-    <updated>2016-07-13T16:59:28+00:00</updated>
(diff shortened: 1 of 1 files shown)


geochronometria.csl
```diff
@@ -81,7 +81,7 @@
         </group>
       </if>
       <else>
-        <text variable="genre" suffix=". "/>
+        <text variable="genre"/>
         <group delimiter=": ">
           <text variable="publisher"/>
           <text variable="publisher-place"/>
```
rmzelle: @dbobak, did you actually test this change? It looks like this changes "genre. publisher: publisher-place" to "genrepublisher: publisher-place", which seems incorrect.

dbobak: @rmzelle, sorry for the late reply. And you're right, in some cases this change could cause problems. I group entire block and upload new version in a moment, this should correct the error.

rmzelle:
Superseded by #2293.

PR #2290, opened 2016-10-29
Update stroke.csl

crubb:
Hey community,

I made some minor updates to stroke.csl:

- Update link to documentation
- Honor guideline "References with more than 6 authors should list the first 6 authors followed by et al."

Best regards,
crubb


rmzelle:
Thanks!

I just changed the documentation to point directly to the reference style guide, which is more convenient.

PR #2292, opened 2016-11-02
revista-chilena-de-derecho-y-tecnologia.csl

juanitowww:


rmzelle:
Thanks. Accepted with some minor edits (see https://github.com/citation-style-language/styles/pull/2292/commits/33a19f74c1af8f1a666e120f618d4937e755247d).

rmzelle:
(the only change that might make a difference is that I changed the style to use the Chilean CSL locale, with `default-locale="es-CL"`; let me know if that gives any problems)

PR #2293, opened 2016-11-02
Update geochronometria.csl

dbobak:
suffix from the "genre" text variable and moved it to a group


rmzelle:
Much better, thanks.

(and for the future, you can just update your original pull request instead of creating a new one; see the instructions in the comment above)

PR #2294, opened 2016-11-07
Update journal-of-mammalogy.csl

rmzelle:
@adam3smith, could you review this?

PR for https://forums.zotero.org/discussion/63068/journal-of-mammalogy-errors. See http://jmammal.oxfordjournals.org/for_authors/index.html for ITA.

### DIFF
--- a/journal-of-mammalogy.csl
+++ b/journal-of-mammalogy.csl
@@ -141,8 +141,8 @@
       <else-if type="bill book chapter graphic legal_case legislation motion_picture paper-conference  song" match="any">
         <group prefix=". " delimiter=" ">
           <group delimiter=" ">
-            <label variable="locator" form="short" text-case="capitalize-first"/>
-            <text variable="locator"/>
+            <label variable="page" form="short" text-case="capitalize-first"/>
+            <text variable="page"/>
             <text term="in"/>
           </group>
           <text variable="container-title"/>
--- a/journal-of-mammalogy.csl
+++ b/journal-of-mammalogy.csl
@@ -41,7 +41,7 @@
   </macro>
   <macro name="author">
     <names variable="author" font-variant="small-caps">
-      <name name-as-sort-order="first" and="text" sort-separator=", " initialize-with=". " delimiter=", " delimiter-precedes-last="never"/>
+      <name name-as-sort-order="first" and="text" sort-separator=", " initialize-with=". " delimiter=", " delimiter-precedes-last="always"/>
       <label form="short" prefix=" (" suffix=")"/>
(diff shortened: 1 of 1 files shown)


journal-of-mammalogy.csl
```diff
@@ -140,9 +140,11 @@
...
       <else-if type="bill book chapter graphic legal_case legislation motion_picture paper-conference  song" match="any">
         <group prefix=". " delimiter=" ">
-          <label variable="page" form="short" text-case="capitalize-first" suffix=" "/>
-          <text variable="page"/>
-          <text term="in"/>
+          <group delimiter=" ">
+            <label variable="locator" form="short" text-case="capitalize-first"/>
+            <text variable="locator"/>
+            <text term="in"/>
+          </group>
```
rmzelle: I made the "in" term conditional by placing it in a group, and switched from "page" to "locator". Didn't really dig into the style guide, but seemed like a reasonable correction.

journal-of-mammalogy.csl
```diff
@@ -140,9 +140,11 @@
       </if>
       <else-if type="bill book chapter graphic legal_case legislation motion_picture paper-conference  song" match="any">
         <group prefix=". " delimiter=" ">
-          <label variable="page" form="short" text-case="capitalize-first" suffix=" "/>
-          <text variable="page"/>
-          <text term="in"/>
+          <group delimiter=" ">
+            <label variable="locator" form="short" text-case="capitalize-first"/>
```
adam3smith: no -- in spite of the name of the macro, (feel free to change), this is called in the bibliography and thus was correct as page (though should be wrapped in a group as you do)

adam3smith:
@rmzelle see comment [above](https://github.com/citation-style-language/styles/pull/2294#discussion_r86711551) -- rest looks good. Thanks!

journal-of-mammalogy.csl
rmzelle: Ah, I now found an example from the style guide:

> Fahr, J. 2013.  Rhinolophus ziama.  Ziama horseshoe bat.  Pp. 355–356 in Mammals of Africa. 4. Hedgehogs, shrews, and bats (M. Happold and D. C. D. Happold, eds.).  Bloomsbury Publishing. London, United Kingdom.

Makes sense. Seems only applicable to chapters, though, not books.

Anyway, I'll revert the "locator" change.

rmzelle: Made the change. I'll let you merge.

adam3smith:
great, thanks

PR #2299, opened 2016-11-10
Remove unnecessary suffix after available-at

pmelchor:
No colon in the OU Harvard specification.


rmzelle:
Another good catch, thanks.

PR #2303, opened 2016-11-14
MHRA fallback for non-numeric volume

pmelchor:
Print-out the volume field "as is" when it is not numeric (e.g. if the user has typed in the roman numerals).
See https://forums.zotero.org/discussion/63174/problem-with-modern-humanities-research-association-3rd-edition#latest

Side note: the Zotero Style Editor seems to ignore the "small-caps" font-variant setting for cs:text, but it works fine when inserted in a Word document.

### DIFF
--- a/modern-humanities-research-association.csl
+++ b/modern-humanities-research-association.csl
@@ -253,6 +253,9 @@
           <if is-numeric="volume">
             <number variable="volume" form="roman" font-variant="small-caps"/>
           </if>
+          <else>
+            <text variable="volume" font-variant="small-caps"/>
+          </else>
         </choose>
       </if>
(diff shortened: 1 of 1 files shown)


pmelchor:
Sorry, I "lied" about the font-variant working in Word. The small caps format is _not_ being applied to the text element.

rmzelle:
Works for me (with the Zotero Style Editor of Zotero Standalone 4.0.29.10, with a chapter item). Are you sure you're testing with a suitable item type?

pmelchor:
Oooops, you're right. Ready to go on my side then.

rmzelle:
Thanks!

PR #2305, opened 2016-11-15
Fix issue in zootaxa.csl with parenthesis surrounding year. 

mjy:
Format was not closing correctly.  Detected when year letter (e.g. 2000a) was assigned.  Not rigorously tested beyond a couple quick spot checks with citeproc-ruby.

### DIFF
--- a/zootaxa.csl
+++ b/zootaxa.csl
@@ -173,8 +173,8 @@
     </sort>
     <layout suffix=" ">
       <text macro="author" suffix=" ("/>
-      <date variable="issued">
-        <date-part name="year" suffix=")"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks. I made this more robust -- FWIW, if your change made a different, that's probably a bug in citeproc-ruby. The two should be identical. It's probably too inconsequential to track down/report, though.

mjy:
Didn't have enough experience to tell where the issue was, but fiddling around I found the "fix". It was consequential enough to report here, so maybe I'll cross reference this there if I can get a spec to show what's going on the citeproc side of things.  Thanks for repo!

PR #2307, opened 2016-11-16
Update meteoritics-and-planetary-science.csl

poikiloblastic:
-Sort in-line citations by author and year, and remove disambiguation option when multiple authors are similar - only year+suffix (e.g. 2016a) is used to disambiguate multiple papers by same author.
-Sort biblio by first author only, then by year. Ignores 2nd+ authorship in alphabetization, as per MaPS policy.
-Removed space after colon separating volume and issue as per MaPS policy.

### DIFF
--- a/meteoritics-and-planetary-science.csl
+++ b/meteoritics-and-planetary-science.csl
@@ -112,8 +112,12 @@
     <text term="in" text-case="capitalize-first" suffix=" "/>
     <text variable="container-title" font-style="italic"/>
   </macro>
-  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year">
-    <layout prefix="(" suffix=")" delimiter="; ">
+<citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" collapse="year">
+    <sort>
+      <key macro="author"/>
+      <key macro="year-date"/>
+    </sort>
+  <layout prefix="(" suffix=")" delimiter="; ">
       <group delimiter=", ">
         <group delimiter=" ">
           <text macro="author-short"/>
@@ -125,8 +129,8 @@
   </citation>
   <bibliography hanging-indent="true" et-al-min="10" et-al-use-first="1">
     <sort>
-      <key macro="author"/>
-      <key variable="title"/>
+      <key macro="author-short" names-use-first="1" names-min="2"/>
+      <key macro="year-date"/>
     </sort>
     <layout suffix=".">
       <group delimiter=" ">
@@ -183,7 +187,7 @@
           </group>
           <group prefix=" " suffix="." delimiter=" ">
             <text variable="container-title" font-style="italic"/>
-            <group delimiter=": ">
+            <group delimiter=":">
               <text variable="volume"/>
               <text variable="page"/>
             </group>
--- a/meteoritics-and-planetary-science.csl
+++ b/meteoritics-and-planetary-science.csl
@@ -112,12 +112,12 @@
     <text term="in" text-case="capitalize-first" suffix=" "/>
     <text variable="container-title" font-style="italic"/>
   </macro>
-<citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" collapse="year">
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks -- fixed indenting and applied the same sort rules to citations as to bibliography. They're not explicit, but nothing else makes sense.

meteoritics-and-planetary-science.csl
```diff
@@ -125,8 +129,8 @@
   </citation>
   <bibliography hanging-indent="true" et-al-min="10" et-al-use-first="1">
     <sort>
-      <key macro="author"/>
-      <key variable="title"/>
+      <key macro="author-short" names-use-first="1" names-min="2"/>
+      <key macro="year-date"/>
```
rmzelle: @adam3smith, it shouldn't really matter now for the style output, but shouldn't the bibliography sort on the "author" macro instead of the "author-short" one?

adam3smith: The two behave differently, though and I'm not sure which one is correct: As is, this would sort

Smith, Bert 1776
Smith, Adam 1848
Smith, Bert, 1919

with `<key macro="author"/>` you'd get
Smith, Adam 1848
Smith, Bert 1776
Smith, Bert 1919

poikiloblastic: rmzelle is right, it isn't sorted quite correctly. I think this code fixes it:
`   <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-givenname="true" givenname-disambiguation-rule="by-cite" initialize-with="." collapse="year">
   <sort>
      <key macro="year-date"/>
      <key macro="author-short"/>
    </sort>
    <layout prefix="(" suffix=")" delimiter="; " >
      <group delimiter=", ">
        <group delimiter=" ">
	   <text macro="author-short"/>
           <text macro="year-date"/>
        </group>`

That should disambiguate if different authors have the same surname. Creates a citation like
> (Kita et al. 1998, 2000, 2005; Zhang et al. 2015; J. Zhang et al. 2016; L.-Y. Zhang et al. 2016)
instead of all the Zhang et al.'s being lumped together. This is consistent with the Chicago Style, which MaPS is branched off of. Chicago may use full first names to disambiguate, but I've never seen that in MaPS.

adam3smith: sorry, I'm confused now. Rintze and I were talking about the bibliography, but your code is for the in-text citation.  Also, your code now is always chronological, which I don't think you want. I.e. it would do (Zang 2000; Kita 2005)

poikiloblastic: MaPS in-line citations are chronological, so it _should_ sort as (Zang 2000; Kita 2005). 
I thought fixing the disambiguation in the in-line citation would help the biblio sorting. There are still issues. Ideally, it would sort by first author (disambiguating by first name is authors have same surnames), then sorting within the authorship by date only (i.e., not considering latter authors).
For the biblio, using `<key macro="author"/>` is the better way, but does not work correctly in the following example:

> Zhang et al. 2012, 2016; H. Zhang et al. 2015; J. Zhang et al. 2015

is cited as (removing all but the 1st and 2nd authors and date):

> Zhang H., Yang Y., ... 2015. ...
Zhang J., Dauphas N., ... 2012. ...
Zhang J., Wang T., ... 2016. ...
Zhang J., Yang W., ...2015. ...

Note the 2015 reference is cited after the 2016 reference because of the 2nd author. Similar behavior occurs if using the "author-short" macro to sort.

adam3smith:
OK, but that last bit can just be fixed by
`<key macro="author" names-use-first="1" names-min="2"/>` or what am I missing? Cite disambiguation wouldn't affect sort-order in the bibliography. Those are separate issues.

poikiloblastic:
Yeah that seems to work well enough. Sorry for the confusion, I'm new here, and usually edit my own styles offline by trial and error.

I had opened another pull request to sort out this sorting and another issue, but just closed and deleted that one. Can you make the changes as discussed to sort things properly? I don't think I can make any official changes without making another pull request. I believe the in-line citations should disambiguate by first name if different authors with the same surname are cited for same year.

The other issue was with the `<bibliography>` tag- it should also be modified to remove usage of et-al. Author lists are never abbreviated in MaPS (I've seen more than 40 authors listed for single pub). So the tag should be truncated to `<bibliography hanging-indent="true">`.

PR #2308, opened 2016-11-16
MaPS Update biblio and et al

poikiloblastic:
-Biblio should never be truncated. All authors should be listed, regardless of number. My previous modification erroneously truncated the authorship list.
Unresolved Issue: In-line citations should be ordered by date preferentially, and author groups should stay together. So, for example, correct: (Leitz et al. 1999, 2009; Arding 2005). Incorrect would be (Arding 2005; Leitz et al. 1999, 2009) or (Arding 2005; Leitz et al. 1999, Leitz et al. 2009). I think the sort code messes with this format, as the original code grouped appropriately but did not arrange by date correctly. Can the in-line citation be sorted after the multiple papers by the same author have been grouped?

### DIFF
--- a/meteoritics-and-planetary-science.csl
+++ b/meteoritics-and-planetary-science.csl
@@ -114,8 +114,8 @@
   </macro>
   <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" collapse="year">
     <sort>
-      <key macro="author-short" names-use-first="1" names-min="2"/>
       <key macro="year-date"/>
+      <key macro="author-short" names-use-first="1" names-min="2"/>
     </sort>
     <layout prefix="(" suffix=")" delimiter="; ">
       <group delimiter=", ">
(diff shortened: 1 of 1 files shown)


poikiloblastic:
I was having trouble testing it yesterday, but it I think that placing the "year-date" criteria before "author-short" in the sort dialog fixed the second issue.

rmzelle:
> In-line citations should be ordered by date preferentially, and author groups should stay together. So, for example, correct: (Leitz et al. 1999, 2009; Arding 2005). Incorrect would be (Arding 2005; Leitz et al. 1999, 2009) or (Arding 2005; Leitz et al. 1999, Leitz et al. 2009). I think the sort code messes with this format, as the original code grouped appropriately but did not arrange by date correctly. Can the in-line citation be sorted after the multiple papers by the same author have been grouped?

As discussed at http://docs.citationstyles.org/en/stable/specification.html#cite-grouping, if you use `<citation ... collapse="...">`:

> cites in in-text citations with identical rendered names are grouped together, e.g. the year-sorted “(Doe 1999; Smith 2002; Doe 2006; Doe et al. 2007)” becomes “(Doe 1999; Doe 2006; Smith 2002; Doe et al. 2007)”.

(and with `collapse="year"`, this would become “(Doe 1999, 2006; Smith 2002; Doe et al. 2007)”)

Issue #1720, opened 2015-10-11
Journal of Sport and Health Science]


adam3smith:
https://forums.zotero.org/discussion/37949/style-request-journal-of-sport-and-health-science/#Item_7

•Campbell JL, Pedersen OK. The varieties of capitalism and hybrid success. Comp Polit Stud, 2007;40:307-2.
•Mares I. Firms and the welfare state: when, why, and how does social policy matter to employers? In: Hall PA, Soskice D, editors. Varieties of capitalism. The institutional foundations of comparative advantage. New York: Oxford University Press, 2001.p.184–213.

The ISSN and EISSN of the journal are

ISSN:
2213-2961 (Electronic)
2095-2546 (Print)
2213-2961 (Linking)

A link to online style documentation of my journal: http://www.jshs.org.cn/EN/column/column290.shtml

A link to a freely available paper formatted with the style I'm requesting: http://www.sciencedirect.com/science/article/pii/S2095254614000271

Thanks in advance!

Kindly regards,

Carol

---

adamsmith Jun 29th 2014 

there should not actually be a comma between journal abbreviation and year, correct? Journal and book titles should be in italics and journal volume numbers in bold, right?

I know Elsevier likes to have DOIs included for all articles in many of the journals, even if they may not appear in the end. What's your preference there?

---

1 Correct. There is NO comma between journal abbreviation and year.

2 Yes, Journal and book titles are be in italics, and journal volume numbers are in bold.

3 Yes, please keep the DOIs
http://doi.org/...

pmelchor:
Question: the template I have started from includes specific bibliography renderings for types that are not mentioned in the JSHS style guide (e.g. article-newspaper or legal_case). Which of the following do you recommend?

a) "Keep them. If there was a good match in the basic types it is likely that those other types would also have to be treated as in the template style if they are ever used".

b) "Get rid of them. Let's keep new styles as simple as possible".

c) Other?

Thanks!

adam3smith:
definitely a)

pmelchor:
Here's another multiple choice question. The template I started from is not great: lots of formatting in bibliography instead of in macros, lots of affixes that could be avoided with a few delimiters, etc. Which option do we go for?

a) "Quick and dirty" (fastest): just make as few changes as are necessary to obtain the desired output.

b) "The perfectionist" (slowest): improve and submit the template style first, then work on the new one.

[...]

PR #2312, opened 2016-11-19
Create journal-of-sport-and-health-science.csl

pmelchor:
This responds to issue https://github.com/citation-style-language/styles/issues/1720

After wasting some time with inadequate templates, I realized that 'For further details you are referred to “Uniform Requirements for Manuscripts submitted to Biomedical Journals”' points to Vancouver as fallback (correct me if I am wrong), so I went for a modified Vancouver. The proposed style (1) matches the output for the examples provided in the style guide and (2) includes the tweaks mentioned in https://forums.zotero.org/discussion/37949/style-request-journal-of-sport-and-health-science/.

As always, I will be happy to implement any corrections or improvements.

### DIFF
--- a/journal-of-sport-and-health-science.csl
+++ b/journal-of-sport-and-health-science.csl
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="minimal">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="minimal" default-locale="en-US">
   <info>
     <title>Journal of Sport and Health Science</title>
     <title-short>JSHS</title-short>
@@ -33,13 +33,6 @@
       <term name="available at">available from</term>
       <term name="section" form="short">sect.</term>
     </terms>
-  </locale>
-  <locale xml:lang="fr">
-    <date form="text" delimiter=" ">
-      <date-part name="day"/>
-      <date-part name="month" form="short" strip-periods="true"/>
-      <date-part name="year"/>
-    </date>
   </locale>
   <macro name="author">
     <names variable="author">
--- a/journal-of-sport-and-health-science.csl
+++ b/journal-of-sport-and-health-science.csl
@@ -1,0 +1,353 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="minimal">
+  <info>
+    <title>Journal of Sport and Health Science</title>
+    <title-short>JSHS</title-short>
+    <id>http://www.zotero.org/styles/journal-of-sport-and-health-science</id>
+    <link href="http://www.zotero.org/styles/journal-of-sport-and-health-science" rel="self"/>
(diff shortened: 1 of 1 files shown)


journal-of-sport-and-health-science.csl
```diff
@@ -0,0 +1,353 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="minimal">
```
adam3smith: since the journal is English only, you'll want to add `default-locale="en-US"` here

journal-of-sport-and-health-science.csl
```diff
@@ -0,0 +1,353 @@
...
+      <term name="collection-editor" form="long">
+        <single>editor</single>
+        <multiple>editors</multiple>
+      </term>
+      <term name="presented at">presented at</term>
+      <term name="available at">available from</term>
+      <term name="section" form="short">sect.</term>
+    </terms>
+  </locale>
+  <locale xml:lang="fr">
```
adam3smith: and since we're setting a default locale, no need for the French terms. Delete the French locale.

journal-of-sport-and-health-science.csl
pmelchor: > I'd typically just fix these myself, but since you're intending to do more -- thanks again! -- it makes sense to send it back to you for future reference

Sure, I prefer it this way so that I can learn de subtleties as I come across different styles. `default-locale` added.

journal-of-sport-and-health-science.csl
pmelchor: Done.

adam3smith:
Thanks! Another note: when you see this reference in the examples: 
 Van der Geer J, Hanraads JAJ, Lupton RA. The art of writing a scientific article. J Sci Commun 2000;163:51–9.
you'll always want to check if the style corresponds to one of the Elsevier in-house styles. It's one of the standard examples they use. That's not the case here, so going with a new style is exactly right, but might come up in the future.

pmelchor:
Great, thanks for the tip :-)

rmzelle:
> I realized that 'For further details you are referred to “Uniform Requirements for Manuscripts submitted to Biomedical Journals”' points to Vancouver as fallback (correct me if I am wrong), so I went for a modified Vancouver.

And yes, that's correct. See also the summary and documentation link of https://github.com/citation-style-language/styles/blob/master/vancouver.csl, and the official account at e.g. http://www.icmje.org/recommendations/archives/2006_urm.pdf (the most recent version doesn't mention Vancouver anymore):

> A small group of editors of general medical journals met informally in Vancouver, British Columbia, in 1978 to establish guidelines for the format of manuscripts submitted to their journals. The group became known as the Vancouver Group. Its requirements for manuscripts, including formats for bibliographic references developed by the National Library of Medicine, were first published in 1979. The Vancouver Group expanded and evolved into the International Committee of 
Medical Journal Editors (ICMJE), which meets annually. ...
>
> The ICJME has produced multiple editions of the Uniform Requirements for Manuscripts Submitted to Biomedical Journals.

pmelchor:
@rmzelle Yes, thanks for confirming! I actually went for a modified Vancouver because an online search led me to that same info :-)

rmzelle:
@pmelchor, also, as a bit of background, we (or at least I) have been trying to limit the use of "Vancouver" and "Harvard" as descriptors for our citation styles. We typically prefer "numeric" and "author-date", respectively, but we often still relent if the organization uses these monikers to refer to their styles (which is e.g. why we have a lot of "Harvard - University Name" styles).

PR #2313, opened 2016-11-19
Coventry - bibliography order

tipichris:
Tweak order for bibliography

### DIFF
--- a/harvard-coventry-university.csl
+++ b/harvard-coventry-university.csl
@@ -310,7 +310,8 @@
   <bibliography hanging-indent="true">
     <sort>
       <key macro="author"/>
-      <key macro="year-date" sort="ascending"/>
+      <key macro="year-date" sort="descending"/>
+      <key variable="citation-number" sort="ascending"/>
       <key variable="title"/>
     </sort>
     <layout>
--- a/harvard-coventry-university.csl
+++ b/harvard-coventry-university.csl
@@ -280,7 +280,7 @@
       </if>
     </choose>
   </macro>
-  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-givenname="true" givenname-disambiguation-rule="all-names" name-as-sort-order="all" collapse="year">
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks -- the commit message had two other issues mentioned (disambiguation/full names). I didn't check, but I assume those were already added in an earlier PR.

tipichris:
Yes. Sorry, Git confuses me! Perhaps I should have started with rebase upstream/master instead of merge upstream/master?
It looks like  438b0e5 and f3c11be in my branch somehow became 9c1d9a4 in master, so they got pushed again

adam3smith:
If you do `git pull upstream master` (which is the same as `git fetch` followed by `git merge`) before adding any changes to the repository that really should work. I'm guessing there was an automerge that made it to the pull request. 
What I think may be going on is that you keep working from the same branch. The standard procedure would be to have a separate branch for every PR and then delete it once the PR gets merged.

rmzelle:
> The standard procedure would be to have a separate branch for every PR and then delete it once the PR gets merged.

@tipichris, especially if you plan to commit a pull request for just a single style, it's usually just as easy to just have GitHub take care of things. If you follow our instructions on updating existing styles at https://github.com/citation-style-language/styles/blob/master/CONTRIBUTING.md#3b-submitting-changes-to-an-existing-style, and edit the version in our official repository, GitHub will automatically recognize that you don't have the rights to commit directly, and create a new up-to-date branch in your fork. That way you don't ever have to bother to keep the "master" branch in your fork up-to-date.

rmzelle:
@adam3smith, it doesn't really matter here, but I wonder if maybe GitHub's new rebase option would have worked here: https://github.com/blog/2243-rebase-and-merge-pull-requests

adam3smith:
Great that that's possible -- but my understanding would be that makes not difference at all here (since the end result is a single commit either way), but could help us in situations where previously merging wasn't possible?

rmzelle:
@adam3smith, yeah, I'm not really sure when you'd want to rebase on GitHub.

PR #2315, opened 2016-11-20
Update theologie-und-philosophie.csl

knorb:
Optimized script, minor style corrections


theologie-und-philosophie.csl
```diff
@@ -238,61 +272,64 @@
...
+          <text variable="container-title" prefix=". Online unter: "/>
           <text variable="URL" prefix=", "/>
           <group delimiter="; " prefix=" (" suffix=")">
             <text macro="date" prefix="Stand: "/>
             <text macro="accessed"/>
           </group>
         </if>
         <else-if type="speech" match="any">
-          <text value=", "/>
+          <text value=". "/>
```
adam3smith: could we find a better solution for this? I really don't like punctuation in text values. How about putting a group on the three rows below and adding a `prefix=". "`?

adam3smith:
Looks good in general. A couple of request inline.

adam3smith:
Besten Dank!

PR #2316, opened 2016-11-22
Minor corrections

ahahl:
"et al." added and a macro dys solved.

### DIFF
--- a/ens-de-lyon-centre-d-ingenierie-documentaire.csl
+++ b/ens-de-lyon-centre-d-ingenierie-documentaire.csl
@@ -24,9 +24,6 @@
   </info>
   <locale xml:lang="fr">
     <terms>
-      <term name="ordinal">e</term>
-      <term name="ordinal-01" gender-form="feminine" match="whole-number">re</term>
-      <term name="ordinal-01" gender-form="masculine" match="whole-number">er</term>
       <term name="cited">op.&#160;cit.</term>
       <term name="page" form="short">
         <single>p.</single>
@@ -50,6 +47,7 @@
           <name form="long" and="text" delimiter-precedes-last="never" sort-separator=" " initialize-with=".">
             <name-part name="family" font-variant="small-caps"/>
           </name>
+          <et-al font-style="italic"/>
         </names>
       </if>
       <else>
@@ -59,6 +57,7 @@
               <name form="long" and="text" delimiter-precedes-last="never" sort-separator=" " initialize-with=".">
                 <name-part name="family" font-variant="small-caps"/>
               </name>
+              <et-al font-style="italic"/>
             </names>
           </if>
           <else-if variable="editor">
@@ -66,6 +65,7 @@
               <name form="long" and="text" delimiter-precedes-last="never" sort-separator=" " initialize-with=".">
                 <name-part name="family" font-variant="small-caps"/>
               </name>
+              <et-al font-style="italic"/>
               <label form="short" prefix="&#160;(" suffix=".)"/>
             </names>
           </else-if>
@@ -81,6 +81,7 @@
           <name form="long" and="text" delimiter-precedes-last="never" name-as-sort-order="first" sort-separator=" ">
             <name-part name="family" font-variant="small-caps"/>
           </name>
+          <et-al font-style="italic"/>
         </names>
       </if>
       <else>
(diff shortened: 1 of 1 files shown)


ens-de-lyon-centre-d-ingenierie-documentaire.csl
```diff
@@ -175,16 +175,25 @@
...
-          <if variable="container-author" match="none">
-            <text macro="editor" prefix=" dans "/>
-            <text variable="container-title" prefix=", " font-style="italic" text-case="capitalize-first"/>
-          </if>
-          <else-if variable="container-author" match="any">
-            <text macro="book_author" prefix=" dans "/>
-            <text variable="container-title" prefix=", " font-style="italic" text-case="capitalize-first"/>
-          </else-if>
-        </choose>
+        <group prefix="dans " delimiter=", ">
```
rmzelle: @ahahl, it's better to rely on CSL terms wherever possible. You could e.g. change:

```xml
<group prefix="dans " delimiter=", ">
  ...
</group>
```

to

```xml
<group delimiter=" ">
  <text term="in"/>
  <group delimiter=", ">
     ...
  </group>
```

(you would also have to redefine the "in" term, which is translated as "in" for the "fr-FR" CSL locale file, with:
`<term name="in">dans</term>` )

rmzelle: Also, you can simplify

```xml
        <group prefix="dans " delimiter=", ">
          <choose>
            <if variable="container-author" match="none">
              <choose>
                <if variable="editor" match="any">
                  <text macro="editor"/>
                  <text variable="container-title" font-style="italic" text-case="capitalize-first"/>
                </if>
                <else>
                  <text variable="container-title" font-style="italic" text-case="capitalize-first"/>
                </else>
              </choose>
            </if>
            <else>
              <text macro="book_author"/>
              <text variable="container-title" font-style="italic" text-case="capitalize-first"/>
            </else>
          </choose>
        </group>
```

to

```xml
        <group prefix="dans " delimiter=", ">
          <choose>
            <if variable="container-author" match="none">
              <choose>
                <if variable="editor">
                  <text macro="editor"/>
                </if>
              </choose>
            </if>
            <else>
              <text macro="book_author"/>
            </else>
          </choose>
          <text variable="container-title" font-style="italic" text-case="capitalize-first"/>
        </group>
```

since you use `<text variable="container-title" font-style="italic" text-case="capitalize-first"/>` in all cases.

rmzelle:
Thanks. (the benefit of using CSL terms instead of verbatim text is that terms automatically translate, so using terms makes it easier to reuse CSL styles in other languages)

> I've made the changes you suggested (thanks again!) and tried working on the ordinals. However, I didn't manage to add the font-style "italic" to "et-al" in the locale settings (line 40), nor did I manage to set my ordinals to vertical-align="sup". There must be something I didn't understand, I will keep looking!

(next time, it's handier if you post comments like these in the pull request thread (i.e., here), instead of in the commit message, which we don't always read carefully)

> I didn't manage to add the font-style "italic" to "et-al" in the locale settings (line 40)

To italicize "et al.", you need to add italics formatting like so (see also http://docs.citationstyles.org/en/stable/specification.html#et-al):

```xml
              <name form="long" and="text" delimiter-precedes-last="never" sort-separator=" " initialize-with=".">
              </name>
              <et-al font-style="italic"/>
<label form="short" prefix="&#160;(" suffix=".)"/>
```

(note that the `<et-al/>` element must always directly follow `<name/>`)

> I manage to set my ordinals to vertical-align="sup"

You're trying to achieve something like "1ʳᵉ", right? There is currently no way to apply `vertical-align="sup"` to ordinal suffixes. The only thing you can do is use special (Unicode) characters that are always uppercase, as is done in the French locale file:

```xml
    <term name="ordinal">ᵉ</term>
    <term name="ordinal-01" gender-form="feminine" match="whole-number">ʳᵉ</term>
    <term name="ordinal-01" gender-form="masculine" match="whole-number">ᵉʳ</term>
```

(see https://github.com/citation-style-language/locales/blob/e10c58517b87d0bff0f7d717fd2134b1cb94976d/locales-fr-FR.xml#L74)

Removing these terms from your style should be enough to have your style rely on the French locale style here instead.

ahahl:
Thanks again, reading the locale file proved really helpful. I also added a new macro for first editions: I just discovered this possibility, it is great! I suppose the file is ok to be pulled now.
Best,

rmzelle:
Great, thanks!

PR #2320, opened 2016-11-23
long version of journal volume and issue was changed to short version

libora:
This pull request was initiated by Mr Kratochvíl. Both versions of representation of volume and issue of the journal are available, but the short version is more universal.

### DIFF
--- a/iso690-author-date-cs.csl
+++ b/iso690-author-date-cs.csl
@@ -335,22 +335,14 @@
   </macro>
   <macro name="issue">
     <group delimiter=", ">
-      <group delimiter="&#160;">
-        <label variable="volume" form="short"/>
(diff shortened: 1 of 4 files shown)


iso690-author-date-cs.csl
```diff
@@ -335,22 +335,14 @@
...
       <text macro="collection"/>
-      <group delimiter="&#160;">
-        <label variable="page" form="short"/>
-        <text variable="page"/>
-      </group>
+       <text variable="page"/>
     </group>
-  </macro>
-  <macro name="publisher">
+  </macro>publisher">
```
libora: Oh, stupid mistake in paste operation :(

rmzelle:
Thanks!

PR #2324, opened 2016-11-30
Update technische-universitat-dresden-historische-musikwissenschaft-n…

jmiba:
…ote.csl

Minor updates

### DIFF
--- a/technische-universitat-dresden-historische-musikwissenschaft-note.csl
+++ b/technische-universitat-dresden-historische-musikwissenschaft-note.csl
@@ -375,11 +375,11 @@
     <group>
       <choose>
         <if type="article-journal article-magazine article-newspaper broadcast interview manuscript map patent personal_communication song speech thesis article webpage chapter paper-conference" match="any">
-          <label suffix=" " variable="page" form="short"/>
+          <label suffix="&#160;" variable="page" form="short"/>
           <text variable="page"/>
         </if>
         <else-if type="entry-encyclopedia" match="any">
-          <label suffix=" " variable="locator" form="short"/>
+          <label suffix="&#160;" variable="locator" form="short"/>
           <text variable="page"/>
         </else-if>
       </choose>
@@ -397,7 +397,7 @@
     </choose>
   </macro>
   <macro name="citation-locator">
-    <label suffix=" " variable="locator" form="short"/>
+    <label suffix="&#160;" variable="locator" form="short"/>
     <text variable="locator"/>
   </macro>
   <macro name="sort-key">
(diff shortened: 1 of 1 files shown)


technische-universitat-dresden-historische-musikwissenschaft-note.csl
```diff
@@ -28,8 +28,8 @@
       <term name="issue" form="short">H.</term>
       <term name="translator" form="verb-short">deutsch von</term>
       <term name="presented at">gehalten auf</term>
-      <term name="no date" form="short">o.&#8239;J.</term>
-      <term name="et-al">u.&#8239;a.</term>
+      <term name="no date" form="short">o. J.</term>
+      <term name="et-al">u. a.</term>
```
rmzelle: @jmiba, if you can help it, can you keep special characters like the one above escaped when making future edits? It might be enough to always start with the latest version from our repository, or, if you already did that, it might be a setting in your text editor.

rmzelle: (every once in a while I run a script to escape special white space characters and the like, because they're otherwise hard to identify as such)

rmzelle: And unless you can easily restore the escape codes, I'll just merge this as is and re-escape these characters later.

jmiba: @rmzelle, thanks, good to know. I use the visual csl editor a lot, it automatically converts decimal code to special characters and I always thought this was a good thing. I will avoid this in the future. For now, could you merge anyway and have your script do the dirty work?

rmzelle: > I use the visual csl editor a lot, it automatically converts decimal code to special characters and I always thought this was a good thing.

Ah, I didn't know it did that. It might be better if it we change it to not do that. Do you happen to know when the visual editor does that conversion? As soon as you type in the codes, or only when you download the edited style?

> For now, could you merge anyway and have your script do the dirty work?

Will do.

technische-universitat-dresden-historische-musikwissenschaft-note.csl
```diff
@@ -364,11 +375,11 @@
     <group>
       <choose>
         <if type="article-journal article-magazine article-newspaper broadcast interview manuscript map patent personal_communication song speech thesis article webpage chapter paper-conference" match="any">
-          <label suffix="&#160;" variable="page" form="short"/>
+          <label suffix=" " variable="page" form="short"/>
```
rmzelle: @jmiba, it looks like you replaced "&#160;" (non-breaking space) by regular spaces here and on lines 382 and 400 below. Is that intentional/desired?

jmiba: No, indeed, those are errors. I'll correct them.

rmzelle:
Thanks!

PR #2325, opened 2016-11-30
Create society-of-biblical-literature-author-date-bibliography.csl

johnhpercival:
SBL 2nd edition author-date style

### DIFF
--- a/society-of-biblical-literature-author-date-bibliography.csl
+++ b/society-of-biblical-literature-author-date-bibliography.csl
@@ -1,0 +1,464 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only">
+  <info>
+    <title>Society of Biblical Literature 2nd edition (author-date)</title>
+    <id>http://www.zotero.org/styles/society-of-biblical-literature-author-date-bibliography</id>
+    <link href="http://www.zotero.org/styles/society-of-biblical-literature-author-date-bibliography" rel="self"/>
+    <link href="http://www.zotero.org/styles/society-of-biblical-literature-fullnote-bibliography" rel="template"/>
+    <link href="http://www.sbl-site.org/publications/SBLHandbookofStyle.aspx" rel="documentation"/>
+    <contributor>
+      <name>John Percival</name>
+      <email>john.percival@cantab.net</email>
+    </contributor>
+    <category citation-format="author-date"/>
+    <category field="theology"/>
+    <summary>Society of Biblical Literature format with author-date and bibliography</summary>
+    <updated>2016-11-30T23:37:43+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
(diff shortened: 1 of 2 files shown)


johnhpercival:
I have now fixed the issues in the test report

adam3smith:
Is this commonly referred to as "author-date"? Typically, we'd refer to parenthetical in-text styles as author-date. I would call this short-note or something along those lines, but we are somewhat flexible about the naming if a name is used specifically in the style guide or the relevant community.

johnhpercival:
It's following the SBL author-date formatting, but you're right that it's using notes not in-text style. I don't know the community well enough to comment of the behaviour. I can change to short-note if preferred?

johnhpercival:
Changed naming to shortnote, and brought locale options in line with SBL fullnote style

adam3smith:
Thanks -- which section of SBL is this described in? I have the manual somewhere and would want to make sure the naming makes sense in how they refer to it.

johnhpercival:
Author-date is in section 6.5 (p104) of SBLHSv2 (2014):

"Authors writing in the social sciences may elect to use an author-date form of citation rather than the more traditional note-based system described above. The primary purpose of author-date style is to eliminate cumbersome notes while still facilitating source references that do not interrupt the flow of the main text to a major degree. Because the style eliminates duplicate listings in notes and bibliography, it can save considerable space. If one chooses, it can even be combined with strictly substantive notes to produce a clear and “streamlined” overall appearance."

I'm not currently trying to make an inline author-date system, but a footnote based system that's built to SBLHSv2 spec, for use in situations when word count is strictly limited (ie my PhD!)

adam3smith:
sorry, I haven't forgotten this, but it's a bit tricky. I want to make sure a short-note version is mentioned in the SBL Manual at all; otherwise I'm a little hesitant to take it.

johnhpercival:
Yes, it's not an official SBL recommended style, so would a name change help? A number of PhD students here at the University of Cambridge use a style like this, and I've taken one of their styles and brought it more fully in line with SBL style. So there is interest in this kind of style.

rmzelle:
@johnhpercival, we're generally hesitant to accept variants of a style that are not explicitly described by its style guide. Accepting styles like these has some problems and costs: e.g. it is hard for us to check the style against the desired output (since there is no description), it's harder for users to quickly find the official version since there will be more search results, and it's harder for us to maintain a larger repository.

We decided on a curated approach for this repository, instead of accepting any style submission, which I think was a good decision, but this means drawing the line somewhere, and I don't think this style variant quite reaches the bar right now.

You're of course always free to distribute the CSL style yourself. And if, for example, a department of the University of Cambridge ever starts to host a guide that describes this style variant, we'd probably accept the style.

adam3smith:
Yes, was going to write something along those lines -- the criteria for acceptance of styles (https://github.com/citation-style-language/styles/wiki/Criteria-for-Accepting-Styles ) are somewhat flexible at the margins, but generally we require a clearly defined (and not too limited) user group and published style guidelines. If a version of this style is used, e.g., officially by the/a theology department at Cambridge, that'd almost certainly work for us, but absent this, we'd probably not take it.

adam3smith:
closing this given the above. If there is, in fact, a style guide along the lines of what I mention above, @johnhpercival , we'll be happy to reopen this.

PR #2326, opened 2016-11-30

johnhpercival:
Changes:
1) Removed default-locale, as it prevented Zotero from changing locale
2) Added en-GB locale options
3) Removed "no." prefix for journal article issue numbers (see https://forums.zotero.org/discussion/62504/style-error-society-of-biblical-literature-2nd-edition-full-note)

### DIFF
--- a/society-of-biblical-literature-fullnote-bibliography.csl
+++ b/society-of-biblical-literature-fullnote-bibliography.csl
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="chicago">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="chicago" default-locale="en-US">
   <info>
     <title>Society of Biblical Literature 2nd edition (full note)</title>
     <id>http://www.zotero.org/styles/society-of-biblical-literature-fullnote-bibliography</id>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks for the fix. With what scenario in mind are you removing the default-locale? I'm asking specifically because you're then adding an en-GB locale that just has en-GB behave exactly like en-US and in this case it'd make much more sense to just force en-US as the default-locale does.

adam3smith:
(just to be clear -- if there is a good reason for removing the locale, I'm happy to do it, that's why I asked about the scenario.)

johnhpercival:
Thanks, I had somewhat misunderstood the locale options. Now I've dug a bit deeper, I think I understand and have updated accordingly.

adam3smith:
But you still want to get rid of the default-locale? If so, just briefly explain what the use case here is. If people are using SBL with UK punctuation or in other languages, this may make perfect sense. 
Also, the date format wasn't right?

johnhpercival:
Thanks for your patience & perseverance. I removed default-locale so that people could change the locale setting in Zotero if they wanted to use it in another language. When default-locale is specified, it seems that Zotero will not let you choose a different locale.

Also, the date format, as far as I could tell, did not appear to be used anywhere, as all references to dates were handled part-by-part, rather than relying on locale formatting. I therefore considered that code redundant and removed it.

adam3smith:
OK, but are you actually aware of people wanting to use the style in a different language? Removing the default locale has costs, albeit small ones: with it set, all users get citations automatically formatted as they are in the SBL manual. Without it, non-US users have to actually select the US locale for that. Not hard, but another thing to know. If SBL is used in other language, by all means let's change this, but I want to make sure that we actually know that that's the case.

johnhpercival:
You're quite right about that cost, which I hadn't considered. I don't know if SBL is used in other languages. Is it better to go along the lines of, "if it ain't broke, don't fix"? ie re-instate default locale?

adam3smith:
OK, i've added it back -- if you do encounter folks who want to use this in different languages/locales, I'm happy to reconsider.

adam3smith:
Thanks!

PR #2328, opened 2016-11-30
Create Budownictwo i Architektura (polski).csl

lukboro:
New citation style for Budownictwo i Architektura journal. Based on IEEE styles. 

Version 1.0.


rmzelle:
@lukboro, I made some changes to make your style valid XML and valid CSL. Let me know if it works properly now. You can download an up-to-date copy at https://raw.githubusercontent.com/lukboro/styles/patch-1/budownictwo-i-architektura.csl

lukboro:
I have updated file (little changes).

@rmzelle : 
I seems that everything is ok. Works properly. Thanks.
Soon I will add an english version - journal needs styles for polish and english language. 

Ł.B

rmzelle:
@lukboro, for bilingual journals we typically do one of the following:

* create a single CSL style that works with both languages, and remove the `default-locale` attribute. This way the user needs to select either English or Polish as the style locale whenever they use it.
* create a single _independent_ CSL style that works with both languages, and fix it to one language with the `default-locale` attribute (e.g. Polish), and create a _dependent_ CSL style fixed to the other language (e.g. English) that links to the independent style. You can see an example of this at https://github.com/citation-style-language/styles/blob/master/pontifical-gregorian-university.csl (independent, Italian) and https://github.com/citation-style-language/styles/blob/master/dependent/pontifical-gregorian-university-en.csl (dependent, English).

Only when it's impossible to cover both languages in a single independent style do we accept two separate independent styles. Could you try and see if you can support both languages in a single style?

rmzelle:
I quickly compared your two styles side-by-side*, and it should be possible to use a single independent style. Let us know if you need help. It looks like you would need to:

* Use two `<locale/>` section, one for Polish (`<locale xml:lang="pl">`) and one for English (`<locale xml:lang="en">`)
* Remove the `default-locale` attribute
* If one language needs `and="text"`, and the other `and="symbol"`, the trick is to use `and="text"` for the combined style, and redefine the "and" term to an ampersand (&) in the locale section of the language that originally used `and="symbol"` (i.e. `<term name="and">&amp;</term>`, since ampersands need to be escaped in XML; see http://stackoverflow.com/a/12524924/1712389).
* Change 

```xml
    <names variable="editor">
    <!-- dorzucilem (ed. by )-->
      <name name-as-sort-order="all" prefix="(ed. by: " suffix=") " and="symbol" sort-separator=" " initialize-with=". " delimiter=", " delimiter-precedes-last="always"/>
    <!-- na co to?? <label form="short" prefix=", " text-case="capitalize-first"/> -->
    </names>
```

to

```xml
    <group prefix="(" suffix=") ">
      <text term="editor" form="verb-short" suffix=": "/>
      <names variable="editor" >
      <!-- dorzucilem (ed. by )-->
        <name name-as-sort-order="all" and="symbol" sort-separator=" " initialize-with=". " delimiter=", " delimiter-precedes-last="always"/>
      <!-- na co to?? <label form="short" prefix=", " text-case="capitalize-first"/> -->
      </names>
    </group>
```

so that the "ed. by" properly translates.

\* (I compared https://raw.githubusercontent.com/lukboro/styles/patch-1/budownictwo-i-architektura.csl and https://raw.githubusercontent.com/lukboro/styles/patch-2/budownictwo-i-architektura-english.csl at https://www.diffnow.com/?report=3byoc)

rmzelle:
And it's probably easiest to start from the Polish version, since the English version isn't valid CSL right now.

rmzelle:
@lukboro, could you try and see if https://raw.githubusercontent.com/lukboro/styles/patch-1/budownictwo-i-architektura-pl.csl and https://raw.githubusercontent.com/lukboro/styles/patch-1/dependent/budownictwo-i-architektura-en.csl work for you? You have to have both installed for the English version to work.

I updated this pull request, and put the English translations in your Polish style, in an English locale section. I then added an English _dependent_ CSL style that relies on the Polish style, but uses English translations. If this works I'll merge this pull request and close the other one.

lukboro:
@rmzelle

Polish version works fine. English does not. When I us it the result is bibliography in polish. I can not tell why it is happening. I use MS Word with Mendeley plugin. Medeley uses citation styles installed on computer, but citation style editor wants styles from ZoteroSR. So I do not know is it a problem with csl file or it is about repository.

rmzelle:
@adam3smith, do you know if Mendeley correctly supports locale-overrides from dependent styles? In my testing with the Italian (independent) and French (dependent) versions of https://www.zotero.org/styles?q=gregorian, I seem to be stuck on Italian with the French version (test e.g. with a book chapter with an editor).

rmzelle:
@lukboro, the fact that Mendeley doesn't correctly handles the English version is a bug in Mendeley. I'm trying to get hold of somebody at Mendeley to get it fixed, but that might take a while.

In the meantime, I can accept the current pull request as it is if you'd like. (I'm strongly against going back to two separate full independent CSL styles. We generally don't accept workarounds here in the repository to deal with bugs in individual reference managers.)

adam3smith:
(sorry, I didn't see me tagged here, but I also don't know the answer -- though of course with Mendeley you can set a locale with any style, even where defaults are set)

lukboro:
@rmzelle
Ok, let's pull existing part. The base version is Polish, so it’s ok.

rmzelle:
Okay, thanks!

PR #2330, opened 2016-12-01
Update kolner-zeitschrift-fur-soziologie-und-sozialpsychologie.csl

shaunss:
Book, articles and chapters should not show URL information even if available in the Zotero entry. Also, for chapters in volumes with multiple editors, the "und" before the last editor should not be preceeded by a komma.

### DIFF
--- a/kolner-zeitschrift-fur-soziologie-und-sozialpsychologie.csl
+++ b/kolner-zeitschrift-fur-soziologie-und-sozialpsychologie.csl
@@ -17,12 +17,15 @@
     <contributor>
       <name>Sebastian Karcher</name>
     </contributor>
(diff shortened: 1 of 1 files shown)


rmzelle:
@adam3smith, can I leave this with you? The PR looks fine to me, but the instructions at the documentation link are in German, so that's easier for you to check.

adam3smith:
Fixed in https://github.com/citation-style-language/styles/commit/13078f5faa9a437ea004777e849a83e7a29cc77e
Thanks @shaunss -- I implemented exactly the same changes you're proposing here, but there was a much more economical way to not print the access macro for those three item types.

PR #2333, opened 2016-12-04
Create technische-universitat-dresden-kunstgeschichte-note-german.csl

jmiba:


rmzelle:
@jmiba, thanks.

(We generally don't include the language in the style file name/ID, unless we have multiple locale-specific versions of the same style)

PR #2334, opened 2016-12-05
Update amphibia-reptilia.csl

RAWes:
I made a few corrections to get the style closer to journal requirements.


amphibia-reptilia.csl
```diff
@@ -42,11 +42,16 @@
   </macro>
   <macro name="editor">
     <names variable="editor">
-      <name delimiter-precedes-last="always" initialize-with="." name-as-sort-order="all"/>
-      <label form="short" text-case="capitalize-first" prefix=", "/>
+      <name delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="19" et-al-use-first="19" et-al-subsequent-min="19" et-al-subsequent-use-first="19" initialize-with="." name-as-sort-order="all"/>
```
rmzelle: Same here.

amphibia-reptilia.csl
```diff
@@ -42,11 +42,16 @@
...
+      <label form="short" text-case="capitalize-first" prefix=", " suffix=", "/>
     </names>
   </macro>
-  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" collapse="year">
+  <macro name="book_author">
+    <names variable="container-author">
+      <name suffix=", " delimiter-precedes-et-al="never" delimiter-precedes-last="never"/>
+    </names>
+  </macro>
+  <citation name-form="short" and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="4" et-al-use-first="1" et-al-subsequent-min="4" et-al-subsequent-use-first="1" disambiguate-add-year-suffix="true" collapse="year">
```
rmzelle: And here.

amphibia-reptilia.csl
```diff
@@ -58,13 +63,13 @@
...
   </citation>
-  <bibliography et-al-min="7" et-al-use-first="6">
+  <bibliography delimiter-precedes-et-al="never" delimiter-precedes-last="never" initialize-with="." name-as-sort-order="all" entry-spacing="2" line-spacing="2" hanging-indent="true">
     <sort>
       <key macro="author"/>
       <key macro="issued"/>
       <key variable="title"/>
     </sort>
-    <layout suffix=".">
+    <layout prefix="•&#9;" suffix=".">
```
rmzelle: @RAWes, I'm pretty sure the journal doesn't actually want bullet points. See e.g. the bibliography in the PDF at http://dx.doi.org/10.1163/15685381-00002859.

amphibia-reptilia.csl
RAWes: I use that in every names definition to avoid unwanted behaviour before I discovered that there were overal definitions in <citation> and <bibliography> that I could set as well. I'll take it out.

amphibia-reptilia.csl
RAWes: Bedankt Rintze. I hadn't seen a PDF yet, good to know, because it was not clear from the Instructions to authors. I'll change it as well, and I'll use the PDF to find definitions for the types that are not defined in the instructions (e.g. edited book).

rmzelle: @RAWes, yes, it almost always helps to look at one or two recent published papers (the one I picked is from the "Free Sample Issue" linked to at http://www.brill.com/amphibia-reptilia).

(also, feel free to write in Dutch if you prefer :) )

rmzelle:
@RAWes, are you still around to update this pull request? We can help you if you have any questions.

RAWes:
I am still around, I just haven’t come to updating it yet, I will put it on my list of things to do pretty soon.

Renate

rmzelle:
Okay, great.

Since you started your submission, we have become a little bit more strict in checking correct values for the et-al attributes (in particular, "et-al-min" must always have a larger value than "et-al-use-first"). I rolled back my corrections for your original style (https://github.com/citation-style-language/styles/commit/e1e992968d45be5ccc92df5264beb41ce362c261#diff-327ae3a87055b2f6ef20f8efa25c97de) to solve a so-called 'merge conflict', so this pull request becomes mergeable again, but you'll have to take another look at the values of these attributes.

rmzelle:
@RAWes, I'm closing this for now because of inactivity.

It might be easiest to just start a new pull request when you're ready to submit some changes, but you also leave a comment here and we'll reopen the pull request.

PR #2336, opened 2016-12-05
Update technische-universitat-dresden-medienwissenschaft-und-neuere-d…

jmiba:
…eutsche-literatur-note.csl

Substantial recoding to comply with requirements


technische-universitat-dresden-medienwissenschaft-und-neuere-deutsche-literatur-note.csl
```diff
@@ -180,14 +185,9 @@
...
-            <if variable="volume">
-              <group delimiter=" ">
-                <text term="volume" form="short"/>
-                <number variable="volume"/>
-              </group>
-            </if>
-          </choose>
+          <group delimiter=" " suffix=".">
+            <text macro="volume"/>
+          </group>
```
rmzelle: @jmiba, groups with just a single child element can always be replaced by just the child element, e.g.

```
          <text macro="volume" suffix="."/>
```

instead of

```xml
            <text macro="volume"/>
          </group>
```

rmzelle:
Thanks!

PR #2338, opened 2016-12-07
Create budownictwo-i-architektura-english.csl

lukboro:
Citation style for "Budownictwo i Architektura" journal. English language version (Journal allows to publish in Polish or English).

ŁB


rmzelle:
(see also #2328)

PR #2339, opened 2016-12-07
Update technische-universitat-dresden-kunstgeschichte-note.csl

jmiba:
Substantial recoding according to requirements

### DIFF
--- a/technische-universitat-dresden-kunstgeschichte-note.csl
+++ b/technische-universitat-dresden-kunstgeschichte-note.csl
@@ -388,9 +388,7 @@
     <text macro="e-resource"/>
     <choose>
       <if match="any" variable="URL">
-        <group delimiter=" " prefix=" (" suffix=")">
-          <date form="numeric" variable="accessed"/>
-        </group>
+        <date form="numeric" variable="accessed" prefix=" (" suffix=")"/>
       </if>
     </choose>
   </macro>
--- a/technische-universitat-dresden-kunstgeschichte-note.csl
+++ b/technische-universitat-dresden-kunstgeschichte-note.csl
@@ -424,7 +424,7 @@
           </group>
         </if>
         <else>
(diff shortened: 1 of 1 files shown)


technische-universitat-dresden-kunstgeschichte-note.csl
```diff
@@ -15,16 +15,17 @@
...
     <category field="humanities"/>
     <summary>Zitierstil entsprechend den Vorgaben des Fachbereichs Kunstgeschichte, Technische Universität Dresden.</summary>
-    <updated>2016-12-04T11:18:53+00:00</updated>
+    <updated>2016-12-07T12:14:31+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="de">
     <terms>
-      <term name="accessed">Stand</term>
+      <term name="accessed"></term>
```
rmzelle: @jmiba, is this empty term intentional? If so, it's probably better to just delete this line, and just not use the "accessed" term in the style.

jmiba: Well, you are right, @rmzelle, even though the term's emptiness was intentional, it's not elegant and should be deleted.

rmzelle:
Great, thanks. My only remaining request is that you standardize the indenting to 2-spaces per level. You can quickly fix this by copy and pasting your style at http://formatter.citationstyles.org/ (the only problem with this tool is that it converts unescaped special whitespace characters, like non-breaking spaces, to regular spaces in many browsers, so it's always a good idea to check the diff when you're committing the reindented version to GitHub). Could you do that as well?

rmzelle:
Ah, so that was just a single line. Should be all good now. Thanks!

PR #2340, opened 2016-12-08
Create revista-latinoamericana-de-recursos-naturales.csl

rcervantesb:
Create revista-latinoamericana-de-recursos-naturales.csl


rmzelle:
Continued in #2341.

@rcervantesb, the reason this pull request is wrong is because you're merging into our "1.0" branch, which is very outdated (our "master" branch contains all the current CSL 1.0.1 styles).

PR #2344, opened 2016-12-08
Update weed-science-society-of-america.csl

rmzelle:
Per https://forums.zotero.org/discussion/63478/request-for-update-to-weed-science-society-of-americas-citation-style


wyoweeds:
For multiple in-text citations, the WSSA journals require a semicolon delimiter. As an example (Smith 2009; Thompson et al. 2013; Vivaldi and Jones 2020). The current version of this style separates these grouped citations with a comma instead. 

I changed the delimiter from ", " to "; " in line 179 in the .csl file on my machine, and it seems to now produce the correct group citations.

PR #2348, opened 2016-12-10
2016-journal-of-dairy-science.csl

wintermind:
New style created from scratch based on the 2016 Journal of Dairy Science "Instructions to Authors: Style and Form"


rmzelle:
Superseded by #2349. (@wintermind, in the future, you can just update your existing pull request instead of creating a new one)

PR #2349, opened 2016-12-10
Update journal-of-dairy-science.csl

wintermind:
This is a complete rewrite of the style to ensure compliance with the  2016 Journal of Dairy Science "Instructions to Authors: Style and Form". The new CSL will correctly produce all of the example references in that document.


rmzelle:
@wintermind, thanks! Did you base this on an existing CSL style? If so, we'd like to have a "template" link (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link)

Other than that it looks good.

wintermind:
@mzelle, the original style was based on a template, but I started from scratch since that first style was never compliant with Journal of Dairy Science style guidelines.

rmzelle:
Okay, thanks! (as far as I can tell, The Rockefeller University Press never published this journal, so it's a bit surprising that the old CSL style used that format)

PR #2352, opened 2016-12-12
new style

hylandj:
New CSL Harvard style created based on Dublin City University's Harvard.

### DIFF
--- a/harvard-dcu.csl
+++ b/harvard-dcu.csl
@@ -4,13 +4,13 @@
   <info>
     <title>Harvard (Dublin City University)</title>
     <title-short>Harvard (DCU)</title-short>
-    <id>http://www.zotero.org/styles/harvard</id>
+    <id>http://www.zotero.org/styles/harvard-dcu</id>
     <link rel="self" href="http://www.zotero.org/styles/harvard-dcu"/>
     <link href="http://www.zotero.org/styles/harvard" rel="template"/>
     <link href="https://www101.dcu.ie/library/Citing&amp;ReferencingGuide/player.html" rel="documentation"/>
     <author>
       <name>Jack Hyland</name>
-      <email>jack.hyland@dcu,ie</email>
+      <email>jack.hyland@dcu.ie</email>
     </author>
     <category citation-format="author-date"/>
     <category field="generic-base"/>
--- a/harvard-dcu.csl
+++ b/harvard-dcu.csl
@@ -1,0 +1,189 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
+  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>Harvard (Dublin City University)</title>
+    <title-short>Harvard (DCU)</title-short>
+    <id>http://www.zotero.org/styles/harvard</id>
+    <link rel="self" href="http://www.zotero.org/styles/harvard-dcu"/>
+    <link href="http://www.zotero.org/styles/harvard" rel="template"/>
+    <link href="https://www101.dcu.ie/library/Citing&amp;ReferencingGuide/player.html" rel="documentation"/>
+    <author>
+      <name>Jack Hyland</name>
+      <email>jack.hyland@dcu,ie</email>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="generic-base"/>
+    <published>2016-12-12T14:36:00+00:00</published>
+    <updated>2016-12-12T14:36:07+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <macro name="editor">
+    <names variable="editor" delimiter=", ">
+      <name and="text" initialize-with=". "/>
+      <label form="short" text-case="lowercase" prefix=" (" suffix=")"/>
+    </names>
+  </macro>
+  <macro name="anon">
+    <text term="anonymous" form="short" text-case="capitalize-first" strip-periods="true"/>
+  </macro>
+  <macro name="author">
+    <names variable="author">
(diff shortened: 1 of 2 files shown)


harvard-dcu.csl
```diff
@@ -0,0 +1,189 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
+  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>Harvard (Dublin City University)</title>
+    <title-short>Harvard (DCU)</title-short>
+    <id>http://www.zotero.org/styles/harvard-dcu</id>
+    <link rel="self" href="http://www.zotero.org/styles/harvard-dcu"/>
+    <link href="http://www.zotero.org/styles/harvard" rel="template"/>
```
rmzelle: @hylandj, we always like to see a "template" link that points to the CSL style your started out with (assuming you based your style on another). We don't have a "harvard.csl" style in our repository, so our tests currently give an error. Can you change this link to one of an existing style?

See also https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link.

harvard-dcu.csl
```diff
@@ -0,0 +1,189 @@
...
+  </info>
+  <macro name="editor">
+    <names variable="editor" delimiter=", ">
+      <name and="text" initialize-with=". "/>
+      <label form="short" text-case="lowercase" prefix=" (" suffix=")"/>
+    </names>
+  </macro>
+  <macro name="anon">
+    <text term="anonymous" form="short" text-case="capitalize-first" strip-periods="true"/>
+  </macro>
```
rmzelle: The "anon" macro isn't used in the style. If you don't need it, please delete it.

harvard-dcu.csl
```diff
@@ -0,0 +1,189 @@
...
+  <macro name="author">
+    <names variable="author">
+      <name and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="0" et-al-use-first="4" et-al-use-last="true" initialize-with="." name-as-sort-order="all"/>
```
rmzelle: Do you need `et-al-min="0" et-al-use-first="4"` here? `et-al-min` must always have a higher value than `et-al-use-first`, and you already define values for these attributes in the bibliography (`<bibliography et-al-min="4" et-al-use-first="1">`), where this macro is called, so you probably don't even need them here.

hylandj: Hi @rmzelle - I've made those three changes:
-template linking to a diff. style
-deleted the anon module
-sorted the et al issue (hopefully)

harvard-dcu.csl
```diff
@@ -0,0 +1,185 @@
...
+      </else>
+    </choose>
+  </macro>
+  <macro name="pages">
+    <group>
+      <label variable="page" form="short" suffix=" "/>
+      <text variable="page"/>
+    </group>
+  </macro>
+  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true">
```
rmzelle: I just checked https://www101.dcu.ie/library/Citing&ReferencingGuide/player.html, and it looks like you need et-al abbreviation for in-text citations for four or more authors, in which case only the first author should be printed. And all authors should be printed in the bibliography.

With that, it's probably clearest to change `<citation et-al-min="3" et-al-use-first="1"` above to `<citation et-al-min="4" et-al-use-first="1"`, and remove `et-al-min="4" et-al-use-first="1"` on line 38. You can delete the `delimiter-precedes-et-al="never"` and `et-al-use-last="true"` attributes on line 28 as well.

hylandj:
> "I just checked https://www101.dcu.ie/library/Citing&ReferencingGuide/player.html, and it looks like you need et-al abbreviation for in-text citations for four or more authors, in which case only the first author should be printed. And all authors should be printed in the bibliography.

With that, it's probably clearest to change <citation et-al-min="3" et-al-use-first="1" above to <citation et-al-min="4" et-al-use-first="1", and remove et-al-min="4" et-al-use-first="1" on line 38. You can delete the delimiter-precedes-et-al="never" and et-al-use-last="true" attributes on line 28 as well."


Thanks @rmzelle - I've made those changes. I had been unsure about where to make the in-line citation et-al changes: in <citation> or in <macro name="author-short">

rmzelle:
@hylandj, I made a minor name change and fixed the style locale to British English, which I assume is desired. Thanks!

hylandj:
Name change is fine with me. Thanks, @rmzelle

hylandj:
Hi @rmzelle - since we worked on this I've noticed that Dublin City University Harvard is appearing in the Zotero Style Repository but it isn't listed in http://editor.citationstyles.org/about/ or other apps I've tried like RefMe. 

Why is this?

adam3smith:
we only update the editor's styles every couple of months or so and don't have any impact on how quickly other apps update their styles. I know Zotero does so almost instantaneously and Mendeley should typically only take a day. No idea about RefMe, Paperpile etc.

hylandj:
thanks @adam3smith - but why isn't it showing on http://editor.citationstyles.org/searchByName/ - surely this should be the first place it appears?

adam3smith:
no. As I say 

>we only update the editor's styles every couple of months or so 

(this has to be done manually and takes a while because we have to recalculate all citation style samples after updating from the style repository)

rmzelle:
> And it's not showing up on Mendeley - http://csl.mendeley.com/searchByName/

And as far as we're aware the Mendeley copy of the Visual Editor is updated even less frequently than ours. It's just not trivial to have that be done automatically.

hylandj:
Got it @adam3smith - I misread your first response.

hylandj:
question - I don't intend this style to have a hanging indent in the bibliography. It isn't specified in the CSL:

<bibliography>

It defaults to no indent in Zotero and Mendeley but citethisforme is adding one. 

Would changing the CSL to:

<bibliography hanging-indent="false">

fix this?

adam3smith:
I doubt it. CSL clearly specifies false as the default and I don't think we'd want to add unnecessary elements for that. I'd report that to citethisforme. My guess would be that it's actually a formatting issue and not a CSL issue, but only they can say for sure. 


BTW, you'll have seen that we updated the editor last weekend, so the style now shows up there.

PR #2353, opened 2016-12-12
Update to veterinary-medicine-austria.csl

damnation333:
It's been a while since I've done this. Sorry if I did something wrong.

### DIFF
--- a/veterinary-medicine-austria.csl
+++ b/veterinary-medicine-austria.csl
@@ -1,174 +1,212 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" default-locale="en-GB">
-  <info>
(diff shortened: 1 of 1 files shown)


rmzelle:
(I just reindented your submission to 2-spaces per indentation level, which is what we use for all styles in the repository; this makes it easier to see what you changed at https://github.com/citation-style-language/styles/pull/2353/files)

It looks like you started from a very old version of this style (2013, if not earlier). Over time, we have made a number of corrections to styles in the repository, and you're now reverting many of these.

Would it be doable for you to make your desired changes to our latest version (available at https://raw.githubusercontent.com/citation-style-language/styles/master/veterinary-medicine-austria.csl)? That would make it much easier for us to accept them.

damnation333:
I'm so sorry about that.
I've now gone through it and compared it all and hopefully did the right changes. the CSL validates after a bit of error finding, too.

rmzelle:
Replaced by #2355.

PR #2354, opened 2016-12-14
Add INFORMS style

rmzelle:
Replacing #1839.


rmzelle:
@adam3smith, nitpick, but thinking I should add the article ("The Institute"), per https://www.informs.org/About-INFORMS. Otherwise looks good I think.

adam3smith:
Wikipedia and EDIRC: https://edirc.repec.org/data/inforea.html have them without the article and they don't use the article in their address:
http://pubsonline.informs.org/page/orsc/contact-journal-staff
I'd go without.

PR #2355, opened 2016-12-14
Major update of veterinary-medicine-austria.csl

damnation333:
General update of style to match new style guidelines and also added some localization bits.


rmzelle:
@damnation333, thanks, much better. I made a few more simplifications.

PR #2360, opened 2016-12-16
Update service-medical-de-l-assurance-maladie.csl

pedrom34:
Fixes following my users' requests :
- Added a citation style for reports.
- Switched the style from numeric to author-date.


rmzelle:
Great, thanks!

PR #2362, opened 2016-12-17
Created new style review-of-international-studies.csl

damnation333:


review-of-international-studies.csl
```diff
@@ -0,0 +1,162 @@
...
+          <date-part name="year"/>
+        </date>
+        <group>
+          <label variable="locator" form="short"/>
+          <text variable="locator" prefix=" "/>
+        </group>
+      </group>
+    </layout>
+  </citation>
+  <bibliography hanging-indent="true" et-al-min="6" et-al-use-first="6">
```
adam3smith: You want the et-al-min to always be at least one higher than the et-al-use-first. Zotero may be doing this right in spite of this, but what you're literally saying is "for 6 authors, print the first 6 followed by et al" which doesn't make sense.

review-of-international-studies.csl
```diff
@@ -0,0 +1,162 @@
...
+          <text prefix=" " suffix="." macro="publisher"/>
+        </if>
+        <else-if type="chapter paper-conference" match="any">
+          <text macro="title" prefix=" "/>
+          <group prefix=", ">
+            <text term="in" suffix=": "/>
+            <group delimiter=", " suffix=".">
+              <names variable="editor translator" delimiter=", ">
+                <name and="symbol" sort-separator=", " initialize-with=". "/>
+                <label form="short" prefix=" (" suffix=")" text-case="capitalize-first"/>
```
adam3smith: It seems odd that the editor label would appear as `(ed.)` for book chapters, but as `, Ed.` (as per the editor translator macro) otherwise. Could you check on that and unify as needed? Ideally, I'd like you to reuse the macro here (which may require adjusting it, in particular taking out the parenthesis around the whole thing there).

review-of-international-studies.csl
```diff
@@ -0,0 +1,162 @@
...
+              <text variable="volume"/>
+              <text variable="issue" prefix=":"/>
+            </group>
+          </group>
+        </else>
+      </choose>
+      <date variable="issued" prefix=" (" suffix=")">
+        <date-part name="year"/>
+      </date>
+      <group prefix=", ">
```
adam3smith: pretty sure that will give you the page ranges twice for chapters, here and on line 129 above. Could you check on that?

review-of-international-studies.csl
```diff
@@ -0,0 +1,162 @@
...
+    </names>
+  </macro>
+  <macro name="author">
+    <names variable="author">
+      <name form="long"/>
+      <label form="short" text-case="capitalize-first" prefix=" (" suffix=")"/>
+      <substitute>
+        <names variable="editor"/>
+        <names variable="translator"/>
+        <choose>
```
adam3smith: the substitute for author and author-short should almost certainly correspond. If you're putting in the container-title here but use the title or short-title in the text, how is a reader going to find this in the bibliography?

rmzelle:
@damnation333, I see you made some minor changes, but the main issue is that this should be a "note" CSL style, not an "author-date" one.

damnation333:
Closed as complete style needs reworking to be a note style and not author-date style. Bibliographic parts would still work.

PR #2364, opened 2016-12-17
Add INFORMS dependents

rmzelle:
Accompanies #2354, generated from https://github.com/citation-style-language/journals/pull/14.


rmzelle:
@adam3smith, note the two deleted (and replaced) styles. I put the math journal on the skip list, since it uses a different format.

PR #2366, opened 2016-12-17
Create ALMA. Archivum Latinitatis Medii Aevi

novacellus:
Style is a modified version of Franziska Heimburger's work on Annales. Histoire, Sciences Sociales


rmzelle:
Thanks!

PR #2368, opened 2016-12-19
Fixed german locale issue

damnation333:
Overseen with all the other changes we made.


damnation333:
@biosainik Thank you. Does this make the default language german? Most users of this style at my Uni use this writing in English, but there is a few who decide to write in german. Should I still change the default-locale in this case?

adam3smith:
If there will be German and English language use of the style, you should entirely remove the default-locale. That way, users can choose the right language.

rmzelle:
@damnation333, thanks!

PR #2371, opened 2016-12-19
IOS Press Style

michelole:
IOS Press reference style as defined in the MS Word template available at http://www.iospress.nl/service/authors/.


rmzelle:
Hmmm, I'm not sure IOS Press uses a single citation style. I checked the instructions for two journals, and they already differ. Compare http://www.iospress.nl/journal/kidney-cancer/ (Vancouver) and http://www.iospress.nl/journal/journal-of-integrative-neuroscience/.

rmzelle:
@michelole, looking at http://www.iospress.nl/service/authors/ again, I see now that the Word template is only for book authors. Is that what you used this CSL style for?

michelole:
Yes, that's the style for books as seen on http://www.iospress.nl/service/authors/.

The same style is used at some conferences (e.g. http://medinfo2017.medmeeting.org/Content/32516 and http://informaticsforhealth.org/instructions-for-authors/), whose proceedings are published by IOS Press.

For reference, see also the official EndNote style at http://endnote.com/downloads/style/ios-press.

rmzelle:
So maybe we can name the style `<title>IOS Press (books)</title>`?

michelole:
Sure! I went ahead and committed the change.

rmzelle:
Thanks!

michelole:
That's great, thanks for the support!

PR #2373, opened 2016-12-24
Delete two spaces in who-europe-harvard.csl

zuphilip:
These two spaces are unnecessary and make problems when analyzing the XML files.

### DIFF
--- a/who-europe-harvard.csl
+++ b/who-europe-harvard.csl
@@ -51,8 +51,8 @@
       <if variable="URL">
         <group delimiter=", ">
           <text variable="URL"/>
-          <group>
-            <text value="accessed"/>
+          <group delimiter=" ">
+            <text term="accessed"/>
             <date variable="accessed">
               <date-part name="day" suffix=" "/>
               <date-part name="month" suffix=" "/>
--- a/who-europe-harvard.csl
+++ b/who-europe-harvard.csl
@@ -52,8 +52,8 @@
         <group delimiter=", ">
           <text variable="URL"/>
           <group>
-            <text value="accessed "/>
-            <date variable="accessed ">
+            <text value="accessed"/>
+            <date variable="accessed">
               <date-part name="day" suffix=" "/>
               <date-part name="month" suffix=" "/>
               <date-part name="year"/>


who-europe-harvard.csl
```diff
@@ -52,8 +52,8 @@
         <group delimiter=", ">
           <text variable="URL"/>
           <group>
-            <text value="accessed "/>
-            <date variable="accessed ">
+            <text value="accessed"/>
+            <date variable="accessed">
```
rmzelle: @zuphilip, shouldn't the `<text/>` element here call the "accessed" term instead of using verbatim text?

rmzelle:
@zuphilip, thanks. Do you know if there are more occurrences of this? I just ran https://github.com/citation-style-language/utilities/blob/master/csl-reindenting-and-info-reordering.py in the hope that that would fix this, but (of course) it wouldn't (https://github.com/citation-style-language/styles/commit/fb66c34920e55ca061252ad00d2281728a20a498).

It would be nice if we could automatically test for this. http://relaxng.org/compact-tutorial.html#id2814955 suggests that we can do this in the schema:

> By default, a literal string pattern will consider the string in the pattern to match the string in the document if the two strings are the same after the whitespace in both strings is normalized. Whitespace normalization strips leading and trailing whitespace characters, and collapses sequences of one or more whitespace characters to a single space character. This corresponds to the behaviour of an XML parser for an attribute that is declared as other than CDATA.
...
There are two datatypes built-in to every RELAX NG implementation. These are named string and token: token corresponds to the default comparison behavior of a literal string pattern; string compares strings without any whitespace normalization (other than the end-of-line and attribute value normalization automatically performed by an XML processor).

So it might suffice to change patterns like `element preferredFormat { "html" | "text" }` to `element preferredFormat { string "html" | string "text" }`.

who-europe-harvard.csl
zuphilip: This would certainly be more general. Moreover, it looks like the space after the text "accessed" should stay, but I will move it to a group delimiter. I will change that.

zuphilip:
This is the only occurrence of starting or ending spaces in the `variable` attribute I found.

rmzelle:
Thanks @zuphilip! Happy 🎄!

(let us know if your analysis yields anything interesting)

Issue #1721, opened 2015-10-11
Biosocieties


adam3smith:
https://forums.zotero.org/discussion/38105/style-request-biosocieties/#Item_4
ISSN: 1745-8552

Author instructions: http://www.palgrave-journals.com/biosoc/author_instructions.html#References

Open access sample article: http://www.palgrave-journals.com/biosoc/journal/v8/n3/pdf/biosoc201313a.pdf

I've found several styles that are close, but none that completely matches the requirements. Thanks for helping out!

Campbell, J.L. and Pedersen, O.K. (2007) The varieties of capitalism and hybrid success. _Comparative Political Studies_ 40(3): 307–332.

Mares, I. (2001) Firms and the welfare state: When, why, and how does social policy matter to employers? In: P.A. Hall and D. Soskice (eds.) _Varieties of capitalism. The institutional foundations of comparative advantage_. New York: Oxford University Press, pp. 184–213.

PR #2374, opened 2016-12-24
Create biosocieties.csl

damnation333:
see under issues #1721  for style-request.

### DIFF
--- a/biosocieties.csl
+++ b/biosocieties.csl
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
   <info>
     <title>BioSocieties</title>
     <id>http://www.zotero.org/styles/biosocieties</id>
@@ -12,8 +12,8 @@
     </author>
     <category citation-format="author-date"/>
     <category field="political_science"/>
+    <issn>1745-8552</issn>
     <eissn>1745-8560</eissn>
-    <issnl>1745-8552</issnl>
     <summary>BioSocieties author-date style based on Harvard</summary>
     <updated>2016-12-24T11:35:58+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
--- a/biosocieties.csl
+++ b/biosocieties.csl
@@ -157,20 +157,24 @@
             </group>
           </group>
         </else-if>  
-        <else-if type="article-newspaper">
+        <else-if type="article-newspaper" match="any">
           <text macro="title" prefix=" " suffix="."/>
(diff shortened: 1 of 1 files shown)


damnation333:
Can somebody help with the first issue? Didn't get it to validate because of the </else-if> line 172. Not sure why.

rmzelle:
Did you use the validator? See http://validator.citationstyles.org/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fdamnation333%2Fstyles%2Fmaster%2Fbiosocieties.csl&version=1.0.1

biosocieties.csl
```diff
@@ -0,0 +1,200 @@
...
+            <text macro="editor"/>
+            <text variable="container-title" font-style="normal" suffix="."/>
+            <text variable="collection-title" suffix="."/>
+            <text variable="event" suffix="."/>
+            <group suffix="." delimiter=" ">
+              <text macro="publisher" prefix=" " suffix=","/>
+              <text variable="issued" suffix=":">
+              <label variable="page" form="short" suffix=" "/>
+              <text variable="page"/>
+            <else-if>
```
rmzelle: @damnation333, there are a bunch of errors here. Just by eye, it looks like the groups started on lines 162 and 168 are both missing a closing tag (`</group>`), the `<else-if>` on line 173 should be a closing tag (`</else-if>`), and "issued" is a date variable, so you can't print it with `<text/>` on line 170. There might be more, but by fixing these issues and using the validator you can probably figure it out.

biosocieties.csl
```diff
@@ -157,20 +157,24 @@
...
-            <group suffix="." delimiter=" ">
-              <text macro="publisher" prefix=" " suffix=","/>
-              <text variable="issued" suffix=":">
-              <label variable="page" form="short" suffix=" "/>
-              <text variable="page"/>
-            <else-if>      
+          <group prefix=" " delimiter=", ">
+            <text variable="container-title" form="short" suffix=","/>
+            <text variable="collection-title"/>
+          </group>
```
rmzelle: @damnation333, `suffix=","` on line 163 should be redundant if you already have `delimiter=", "` on the group.

biosocieties.csl
```diff
@@ -157,20 +157,24 @@
...
+          <group prefix=" " suffix=":">
+            <date variable="issued">
+              <date-part name="day" suffix=" " range-delimiter="-"/>
+              <date-part name="month"/>
+            </date>
+          </group>
+          <group suffix="." delimiter=" ">
+            <text macro="publisher" prefix=" " suffix=","/>
+            <label variable="page" form="short" suffix=" "/>
```
rmzelle: @damnation333, this `suffix=" "` should also be redundant.

biosocieties.csl
damnation333: I tried, but it isn't right if I delete it.

rmzelle: I guess it could make a difference in cases where you have a "container-title" but not a "collection-title".

rmzelle:
Thanks. I changed the `<issnl/>` to `<issn/>` (ISSN-L hasn't really caught on, so we stick to print and electronic ISSNs), and fixed the style to US English.

rmzelle:
Closes #1721.

@damnation333, will you post to the original Zotero thread to let the requester know their style arrived?

damnation333:
@rmzelle will do. Thanks for merging :)

PR #2375, opened 2016-12-24
fixed errors of Annals of Botany

damnation333:
as pointed out here: https://forums.zotero.org/discussion/63643/errors-with-aob-bibliography-in-mendeley#latest

### DIFF
--- a/annals-of-botany.csl
+++ b/annals-of-botany.csl
@@ -105,24 +105,23 @@
           <text prefix=" " suffix="." macro="publisher"/>
         </if>
         <else-if type="webpage" match="any">
-          <group suffix=". ">
-            <text macro="title" font-style="italic" prefix=" "/>
-          </group>
+          <text macro="title" font-style="italic" prefix=" " suffix=". "/>
           <text variable="URL" prefix=" " suffix="."/>
           <date variable="accessed">
             <date-part name="day" form="numeric" suffix=" "/>
-             <date-part name="month" form="short" suffix=" "/>
-             <date-part name="year" form="long"/>
-           </date>
+            <date-part name="month" form="short" suffix=" "/>
+            <date-part name="year" form="long"/>
+          </date>
         </else-if>    
         <else-if type="chapter paper-conference" match="any">
           <text macro="title" prefix=" "/>
-          <group prefix=" In: ">
+          <group prefix=" ">
+            <text term="in" text-case="capitalize-first" suffix=": "/>
             <group>
               <group suffix=".">
                 <names variable="editor translator" delimiter="," suffix=" ">
                   <name name-as-sort-order="all" sort-separator=" " initialize-with="" delimiter=", " delimiter-precedes-last="always"/>
-                  <label form="short" prefix=", " suffix=""/>
+                  <label form="short" prefix=", " />
                 </names>
                 <text variable="collection-title" prefix=" " suffix="."/>
                 <text variable="container-title" font-style="italic" prefix=" " suffix="."/>
--- a/annals-of-botany.csl
+++ b/annals-of-botany.csl
@@ -105,14 +105,14 @@
           <text prefix=" " suffix="." macro="publisher"/>
         </if>
         <else-if type="webpage" match="any">
-          <group delimiter=". " suffix=".">
-            <text macro="title" font-style="italic" prefix=" "/>
-            <text variable="URL" prefix=" "/>
+          <group delimiter=". " prefix=" " suffix=".">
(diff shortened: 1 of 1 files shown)


annals-of-botany.csl
```diff
@@ -104,6 +104,13 @@
           </group>
           <text prefix=" " suffix="." macro="publisher"/>
         </if>
+        <else-if type="webpage" match="any">
+          <group suffix=". ">
+            <text macro="title" font-weight="italics" prefix=" "/>
+          </group>
+          <text variable="url" suffix=". "/>
+          <date variabiable="accessed"/>
+        </else-if>
```
rmzelle: @damnation333, see the validation results: http://validator.citationstyles.org/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fdamnation333%2Fstyles%2Fpatch-8%2Fannals-of-botany.csl&version=1.0.1

To use italics, use `font-style="italic"`. To print the URL, use the "URL" (all-caps) variable. And change "variabiable" to "variable".

annals-of-botany.csl
```diff
@@ -104,14 +104,25 @@
           </group>
           <text prefix=" " suffix="." macro="publisher"/>
         </if>
+        <else-if type="webpage" match="any">
+          <group suffix=". ">
+            <text macro="title" font-style="italic" prefix=" "/>
+          </group>
```
rmzelle: @damnation333, while it's good to put as much separating punctuation in groups as possible, groups with only a single child element only complicate things, and should be avoided. (moving the suffix from the group to the `text` element would allow you to delete lines 108 and 110)

annals-of-botany.csl
```diff
@@ -104,14 +104,25 @@
...
-          <group prefix=".">
+          <group prefix=" In: ">
             <group>
               <group suffix=".">
-                <names variable="editor translator">
-                  <name name-as-sort-order="all" sort-separator=" " initialize-with="" delimiter=", " delimiter-precedes-last="always" prefix=" In: " suffix=","/>
-                  <label form="short" prefix=" "/>
+                <names variable="editor translator" delimiter="," suffix=" ">
+                  <name name-as-sort-order="all" sort-separator=" " initialize-with="" delimiter=", " delimiter-precedes-last="always"/>
+                  <label form="short" prefix=", " suffix=""/>
```
rmzelle: @damnation333, it's also cleaner to delete empty prefixes and suffixes, like `suffix=""` here.

annals-of-botany.csl
```diff
@@ -104,14 +104,25 @@
...
+          <date variable="accessed">
+            <date-part name="day" form="numeric" suffix=" "/>
+             <date-part name="month" form="short" suffix=" "/>
+             <date-part name="year" form="long"/>
+           </date>
+        </else-if>    
         <else-if type="chapter paper-conference" match="any">
           <text macro="title" prefix=" "/>
-          <group prefix=".">
+          <group prefix=" In: ">
```
rmzelle: @damnation333, if possible, we'd like to use CSL terms instead of verbatim text, because terms can be automatically translated, and using them make CSL styles easier to (re)use in other languages. Here, you can change `<group prefix=" In: ">` to e.g.

```xml
<group prefix=" " >
```

(Don't take my nitpicking the wrong way, by the way. We appreciate your help. I'm just pointing out some things that can be improved)

damnation333: this is really helpful. Thanks for the little tips here. Exactly what I need to get better at this. Love it. Like solving puzzles.

annals-of-botany.csl
```diff
@@ -104,14 +104,24 @@
...
           <text prefix=" " suffix="." macro="publisher"/>
         </if>
+        <else-if type="webpage" match="any">
+          <text macro="title" font-style="italic" prefix=" " suffix=". "/>
+          <text variable="URL" prefix=" " suffix="."/>
+          <date variable="accessed">
+            <date-part name="day" form="numeric" suffix=" "/>
+            <date-part name="month" form="short" suffix=" "/>
+            <date-part name="year" form="long"/>
+          </date>
```
rmzelle: @damnation333, is this correct now? I don't see any spaces between the URL and the date (just "."), which probably isn't what you want.

rmzelle: (if you need ". " as a delimiter between the title and URL, and between the URL and "accessed" date, it would be much better to delete these suffixes and add a parent group with `delimiter=". "`)

rmzelle:
Thanks! (accepted after reindenting the style with http://formatter.citationstyles.org/)

PR #2379, opened 2016-12-26
create scandinavian-journal-of-rheumatology

damnation333:
Per https://forums.zotero.org/discussion/62560/style-request-scandinavian-journal-of-rheumatology

### DIFF
--- a/scandinavian-journal-of-rheumatology.csl
+++ b/scandinavian-journal-of-rheumatology.csl
@@ -34,9 +34,6 @@
     <names variable="editor" prefix=" In: " suffix=". ">
       <name sort-separator=" " initialize-with="" name-as-sort-order="all" delimiter=", " delimiter-precedes-last="always" et-al-min="5" et-al-use-first="3" delimiter-precedes-et-al="never"/>
       <label form="short" prefix=" "/>
-      <substitute>
-        <names variable="editor"/>
-      </substitute>
     </names>
   </macro>
   <macro name="publisher">
@@ -45,16 +42,6 @@
   </macro>
   <macro name="container-title">
     <text variable="container-title" form="short" vertical-align="baseline" suffix="."/>
-  </macro>
-  <macro name="title">
-    <choose>
-      <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
-        <text variable="title" suffix=". "/>
-      </if>
-      <else>
-        <text variable="title" suffix=". "/>
-      </else>
-    </choose>
   </macro>
   <macro name="volume">
     <text variable="volume"/>
@@ -95,13 +82,13 @@
       <text variable="citation-number" suffix=". "/>
       <choose>
         <if type="webpage">
-          <text macro="title"/>
+          <text variable="title" suffix=". "/>
           <text variable="URL" prefix=" (" suffix=")."/>
           <text macro="accessed-date"/>
         </if>
         <else>
           <text macro="author"/>
-          <text macro="title"/>
+		    <text variable="title" suffix=". "/>
           <choose>
             <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
               <group delimiter=" ">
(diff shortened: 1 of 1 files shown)


scandinavian-journal-of-rheumatology.csl
```diff
@@ -0,0 +1,140 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" page-range-format="minimal" demote-non-dropping-particle="sort-only">
```
rmzelle: Again need a "default-locale".

scandinavian-journal-of-rheumatology.csl
```diff
@@ -0,0 +1,140 @@
...
+  <info>
+    <title>Scandinavian Journal of Rheumatology</title>
+    <id>http://www.zotero.org/styles/scandinavian-journal-of-rheumatology</id>
+    <link href="http://www.zotero.org/styles/scandinavian-journal-of-rheumatology" rel="self"/>
+    <link href="http://www.zotero.org/styles/annals-of-oncology" rel="template"/>
+    <link href="www.oxfordjournals.org/our_journals/annonc/for_authors/general.html" rel="documentation"/>
```
rmzelle: This "documentation" link needs to be updated.

scandinavian-journal-of-rheumatology.csl
```diff
@@ -0,0 +1,140 @@
...
+      <name sort-separator=" " initialize-with="" name-as-sort-order="all" delimiter=", " delimiter-precedes-last="always"/>
+    </names>
+  </macro>
+  <macro name="editor">
+    <names variable="editor" prefix=" In: " suffix=". ">
+      <name sort-separator=" " initialize-with="" name-as-sort-order="all" delimiter=", " delimiter-precedes-last="always" et-al-min="5" et-al-use-first="3" delimiter-precedes-et-al="never"/>
+      <label form="short" prefix=" "/>
+      <substitute>
+        <names variable="editor"/>
+      </substitute>
```
rmzelle: @damnation333, I see that this same code exists in https://github.com/citation-style-language/styles/blob/master/annals-of-oncology.csl, but can you still fix it here? Just take out the whole `<substitute/>` (lines 37-39). It makes no sense to substitute the "editor" variable in case of a missing "editor" variable.

scandinavian-journal-of-rheumatology.csl
```diff
@@ -0,0 +1,140 @@
...
+  <macro name="title">
+    <choose>
+      <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
+        <text variable="title" suffix=". "/>
+      </if>
+      <else>
+        <text variable="title" suffix=". "/>
+      </else>
+    </choose>
+  </macro>
```
rmzelle: @damnation333, the `<choose/>` doesn't do anything here anymore, since, unlike as in annals-of-oncology.csl, here the outcome will always be the same. Can you removing all the testing lines? You could even get rid of the entire "title" macro and just use `<text variable="title" suffix=". "/>` where you used to call it.

scandinavian-journal-of-rheumatology.csl
```diff
@@ -0,0 +1,140 @@
...
+  <macro name="volume">
+    <text variable="volume"/>
+  </macro>
```
rmzelle: Here it's again simpler to delete the "volume" macro and directly call the variable where needed.

rmzelle:
Thanks for helping people out with their style requests! I don't really have the time to do them all myself. I made a few suggestions to simplify the style, but otherwise this looks good.

damnation333:
changes made. 
I can't just edit the code in Zotero's test pane and copy it all back, right? then it looks lik the whole thing was changes when only a few lines were changed?

rmzelle:
> I can't just edit the code in Zotero's test pane and copy it all back, right? then it looks lik the whole thing was changes when only a few lines were changed?

In my experience, the Zotero test pane doesn't change anything like indentation, so you should be able to just copy and paste code in, make some changes, and copy and paste it back out. If you don't see that, please be more specific and we can see if there is anything to troubleshoot. (you could create before-and-after versions at https://gist.github.com/, and share the URLs of the Gists here, and I'll take a look)

rmzelle:
Looks good, thanks!

PR #2380, opened 2016-12-27
Update iso690-author-date-en.csl

grolimur:
Replaced comma with a point after the publication date in the reference + removed space between "In" and ":"


rmzelle:
@grolimur, I'm not sure if I asked this before, but is there any documentation we can point to for this style format? And are the changes you're proposing specified in the ISO690 guidelines?

grolimur:
@rmzelle Yes, but the ISO 690 standard is paywalled. Do you think it's still a good idea to point to it?

rmzelle:
@grolimur, yes. And styles may have multiple documentation links, so it can't really hurt to add a link to the standard, even if it's not accessible by all. Can you update this pull request with one?

grolimur:
@rmzelle Done! I hope I did correctly.

rmzelle:
Great, thanks.

PR #2381, opened 2016-12-27
create current-pharmaceutical-design.csl

damnation333:
as requested here: https://forums.zotero.org/discussion/62618/style-request-current-pharmaceutical-design/

### DIFF
--- a/current-pharmaceutical-design.csl
+++ b/current-pharmaceutical-design.csl
@@ -4,6 +4,7 @@
     <title>Current Pharmaceutical Design</title>
     <title-short>CPD</title-short>
     <id>http://www.zotero.org/styles/current-pharmaceutical-design</id>
+    <link href="http://www.zotero.org/styles/current-pharmaceutical-design" rel="template"/>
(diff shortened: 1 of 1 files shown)


rmzelle:
@damnation333, you might want to add a link to this pull request to your post at https://forums.zotero.org/discussion/comment/266887/#Comment_266887 :). (although in general, unless you know the requesting user is comfortable with downloading styles from GitHub, it might be handier to just commit a first version to the CSL repository, and then fix up any issues that show up in a follow-up pull request)

rmzelle:
Looks good, thanks!

rmzelle:
And off-topic, I must admit that I find it slightly awkward to write your username here and on the Zotero forums. I realize it's unlikely to be your intent, but I personally associate "damnation" mostly with profanity (https://en.wikipedia.org/wiki/Damnation#As_profanity), especially when it's used without much context (as is the case with a username).

It's not a very big deal for me, so feel free to keep using it, but it bothered me just enough that I wanted to share my feelings about it.

Issue #1794, opened 2015-12-04
Human Mutation


adam3smith:
Our style looks in pretty bad shape:
https://forums.zotero.org/discussion/55222/citation-style-human-mutation-mistakes/#Comment_241570

damnation333:
@adam3smith Can this be closed now?

adam3smith:
closed via #2382

PR #2382, opened 2016-12-27
added comma for in-text citation

damnation333:
as requested here and pointed out in #1794 https://forums.zotero.org/discussion/55222/citation-style-human-mutation-mistakes/#Comment_241570
Citation disambiguation was tested and works correctly for me.
Journal, book and chapter styles were double checked and seem to work fine.

### DIFF
--- a/human-mutation.csl
+++ b/human-mutation.csl
@@ -13,11 +13,11 @@
     <category field="medicine"/>
     <issn>1059-7794</issn>
     <eissn>1098-1004</eissn>
-    <updated>2012-09-27T22:06:38+00:00</updated>
+    <updated>2016-12-27T20:07:23+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="author-short">
-    <names variable="author">
+    <names variable="author" suffix=",">
       <name form="short" and="text"/>
       <substitute>
         <names variable="editor"/>
--- a/human-mutation.csl
+++ b/human-mutation.csl
@@ -48,12 +48,14 @@
   </macro>
   <citation disambiguate-add-year-suffix="true" et-al-min="3" et-al-use-first="1" collapse="year">
     <sort>
-      <key macro="issued"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
It's possible this is still incorrect wrt to the comma -- did you check? I just meant to point out that the style had been modified since the error report, so you shouldn't expect it to be.

damnation333:
@adam3smith Just checked again. The comma was missing and should defo be there.

human-mutation.csl
```diff
@@ -13,11 +13,11 @@
     <category field="medicine"/>
     <issn>1059-7794</issn>
     <eissn>1098-1004</eissn>
-    <updated>2012-09-27T22:06:38+00:00</updated>
+    <updated>2016-12-27T20:07:23+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="author-short">
-    <names variable="author">
+    <names variable="author" suffix=",">
```
rmzelle: @damnation333, while this works, styles are easier to understand and edit if you put separating punctuation at the highest possible level. In particular, try to write macros in such a way that they don't generate output with dangling separating punctuation (i.e. "Miller et al." is better than "Miller et al.,").

In this case, it's best to change:

```xml
      <text macro="author-short"/>
      <text macro="issued" prefix=" "/>
    </layout>
```

to

```xml
      <group delimiter=", ">
        <text macro="author-short"/>
        <text macro="issued"/>
      </group>
    </layout>
```

damnation333: great. Thank you for the explanation. I'll keep an eye on these sort of things from now on.

adam3smith:
@damnation333 -- check out the last two changes I made before merging: once you add the delimiter, you'd also remove the affixes (i.e. prefix and suffix) for that author-short and issued macro. Also, make sure indenting stays consistent (for some reason there was an extra level in the sort section of the citation).

Thanks!

damnation333:
@adam3smith Thanks and sorry. So much to learn still.

rmzelle:
Thanks, both.

PR #2384, opened 2016-12-28
create veterinary-pathology.csl

damnation333:
Hi,
Had a go at this style. Took me the longest so far as this style is all about macros with lots of conditional layers, so it took me a lot of brain to figure it out. Potentially also made a mess.
Based on Vancouver (Foot note).
Request on Zotero Forum: https://forums.zotero.org/discussion/59276/style-request-veterinary-pathology/


rmzelle:
> Potentially also made a mess.

@adam3smith, if you want to take a look, see https://www.diffchecker.com/4OaklLP1.

@damnation333, if you're not confident about the edits you made in a new style, one thing you could do is first committing the original template style (but with the new file name), and then committing your changes in a second commit. That way we can much more easily see what you changed. (in this case, I used an online diff tool, but that's more of a workaround.

veterinary-pathology.csl
```diff
@@ -0,0 +1,314 @@
...
+        <group prefix="[" suffix="]" delimiter=" ">
+          <text term="cited" text-case="lowercase"/>
+          <date variable="accessed" form="text"/>
+        </group>
+      </if>
+    </choose>
+  </macro>
+  <macro name="container-title">
+    <choose>
+      <if type="article-journal article-magazine paper-conference article-newspaper review review-book entry-dictionary entry-encyclopedia" match="any">
```
damnation333: chapters' container-titles do still get printed in italics (throught the <else>), but with its punctuation. Otherwise it's not correct.

veterinary-pathology.csl
```diff
@@ -0,0 +1,305 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="expanded" default-locale="en-US">
+  <info>
+    <title>Veterinary Pathology</title>
+    <id>http://www.zotero.org/styles/veterinary-pathology</id>
+    <link href="http://www.zotero.org/styles/veterinary-pathology" rel="self"/>
+    <link href="http://www.zotero.org/styles/vancouver-superscript" rel="template"/>
+    <link href="https://mc.manuscriptcentral.com/societyimages/vetpath/Veterinary%20Pathology-Instructions%20to%20Authors%202016-4-18.pdf" rel="documentation"/>
```
rmzelle: (@damnation333, in general, I try avoiding linking to documentation URLs that have a date in them, such as "... 2016-4-18.pdf" here, since they're likely to become dead links relatively quick. https://us.sagepub.com/en-us/nam/veterinary-pathology/journal201966#submission-guidelines might be more stable in this case. [I'll make the change myself])

rmzelle:
Thanks!

PR #2387, opened 2016-12-29
Create Plant and Cell Physiology

damnation333:
https://forums.zotero.org/discussion/57493/style-request-plant-and-cell-physiology/

### DIFF
--- a/plant-&amp;-cell-physiology.csl
+++ b/plant-&amp;-cell-physiology.csl
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style class="in-text" version="1.0" et-al-min="6" et-al-use-first="6" demote-non-dropping-particle="sort-only" default-locale="en-GB" xmlns="http://purl.org/net/xbiblio/csl">
+<style class="in-text" version="1.0" et-al-min="7" et-al-use-first="6" demote-non-dropping-particle="sort-only" default-locale="en-GB" xmlns="http://purl.org/net/xbiblio/csl">
   <info>
     <title>Plant and Cell Physiology</title>
-    <id>http://www.zotero.org/styles/plant-and-cell-physiology</id>
-    <link href="http://www.zotero.org/styles/plant-and-cell-physiology" rel="self"/>
+    <id>http://www.zotero.org/styles/plant-&amp;-cell-physiology</id>
+    <link href="http://www.zotero.org/styles/plant-&amp;-cell-physiology" rel="self"/>
     <link href="http://www.zotero.org/styles/elsevier-harvard-without-titles" rel="template"/>
     <link href="http://www.oxfordjournals.org/our_journals/pcp/for_authors/general.html#advice" rel="documentation"/>
     <author>
@@ -12,10 +12,9 @@
     </author>
     <category citation-format="author-date"/>
     <category field="biology"/>
-    <category field="generic-base"/>
     <issn>0032-0781</issn>
     <eissn>1471-9053</eissn>
-    <summary>Style for Plant and Cell Physiology based on Elsevier Harvard (no titles).</summary>
+    <summary>Style for Plant &amp; Cell Physiology based on Elsevier Harvard (no titles).</summary>
     <updated>2016-12-29T15:40:11+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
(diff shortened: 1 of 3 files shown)


plant-and-cell-physiology.csl
```diff
@@ -0,0 +1,227 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" et-al-min="6" et-al-use-first="6" demote-non-dropping-particle="sort-only" default-locale="en-GB" xmlns="http://purl.org/net/xbiblio/csl">
```
rmzelle: @damnation333, `et-al-min` must always have a higher value than `et-al-use-first`.

plant-and-cell-physiology.csl
```diff
@@ -0,0 +1,227 @@
...
+  <info>
+    <title>Plant and Cell Physiology</title>
```
rmzelle: @damnation333, it looks like this title uses an ampersand instead of "and". You can use XML-escaped ampersands with `&amp;`.

plant-and-cell-physiology.csl
```diff
@@ -0,0 +1,227 @@
...
+    <link href="http://www.zotero.org/styles/plant-and-cell-physiology" rel="self"/>
+    <link href="http://www.zotero.org/styles/elsevier-harvard-without-titles" rel="template"/>
+    <link href="http://www.oxfordjournals.org/our_journals/pcp/for_authors/general.html#advice" rel="documentation"/>
+    <author>
+      <name>Patrick O'Brien, PhD</name>
+      <email>obrienpat86@gmail.com</email>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="biology"/>
+    <category field="generic-base"/>
```
rmzelle: @damnation333, we typically only use `field="generic-base"` for styles that can't be assigned to a single or small subset of disciplines, such as styles for general style guides (like APA) or for university. It should be deleted for almost all journals.

plant-and-cell-physiology.csl
```diff
@@ -0,0 +1,227 @@
...
+    <group delimiter=" ">
+      <choose>
+        <if is-numeric="edition">
+          <number variable="edition" form="ordinal"/>
+        </if>
+        <else>
+          <text variable="edition" suffix="."/>
+        </else>
+      </choose>
+      <text value="ed"/>
```
rmzelle: @damnation333, it's better to use CSL terms as much as possible. Here you can use `<text term="edition" form="short" strip-periods="true"/>` (see https://github.com/citation-style-language/locales/blob/master/locales-en-GB.xml#L34). It's a bit longer, but this way the style can be reused more easily in other languages.

plant-and-cell-physiology.csl
```diff
@@ -0,0 +1,227 @@
...
+    </group>
+  </macro>
+  <macro name="locators">
+    <choose>
+      <if type="article-journal article-magazine article-newspaper" match="any">
+        <group delimiter=": " prefix=" ">
+          <group>
+            <text variable="volume"/>
+          </group>
```
rmzelle: Useless group.

plant-and-cell-physiology.csl
damnation333: ah. now I get it. read the spec sheet several times and had a mind blockage with this one. cheers.

plant-and-cell-physiology.csl
damnation333: Great. Thanks. Thats an error in the template file, I think. Should we fix it there, too?

plant-and-cell-physiology.csl
rmzelle: Yeah, the spec could probably be made a little clearer, but "use-first" must by definition be lower than "min". "min" defines the minimum number of authors at which et-al abbreviation kicks in, and because "et al." means "and others", you must always print fewer names than that.

plant-and-cell-physiology.csl
rmzelle: We're happy to take a pull request. It's not really wrong, though, although it's not optimal.

plant-&amp;-cell-physiology.csl
```diff
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style class="in-text" version="1.0" et-al-min="6" et-al-use-first="6" demote-non-dropping-particle="sort-only" default-locale="en-GB" xmlns="http://purl.org/net/xbiblio/csl">
+<style class="in-text" version="1.0" et-al-min="7" et-al-use-first="6" demote-non-dropping-particle="sort-only" default-locale="en-GB" xmlns="http://purl.org/net/xbiblio/csl">
   <info>
     <title>Plant and Cell Physiology</title>
-    <id>http://www.zotero.org/styles/plant-and-cell-physiology</id>
-    <link href="http://www.zotero.org/styles/plant-and-cell-physiology" rel="self"/>
+    <id>http://www.zotero.org/styles/plant-&amp;-cell-physiology</id>
+    <link href="http://www.zotero.org/styles/plant-&amp;-cell-physiology" rel="self"/>
```
rmzelle: @damnation333, no, that's not right. We want to keep "plant-and-cell-physiology" for the style ID and file name. It's just the title that had to be changed: `<title>Plant &amp; Cell Physiology</title>`

rmzelle:
Thanks!

PR #2389, opened 2016-12-29
Create european-journal-of-endocrinology.csl

sylverestoermann:
Style for European Journal of Endocrinology, based on Vancouver style
Documentation (http://www.eje-online.org/site/misc/For-Authors.xhtml) is sparse and includes only three example references (2x article-journal, 1x chapter). Other details were derived from the bibliography of actual and recent publications in the journal.

### DIFF
--- a/european-journal-of-endocrinology.csl
+++ b/european-journal-of-endocrinology.csl
@@ -18,6 +18,11 @@
     <updated>2016-12-29T23:48:11+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
+  <locale xml:lang="en">
+    <terms>
+      <term name="edition" form="short">edn</term>
+    </terms>
+  </locale>
   <macro name="author">
     <names variable="author">
       <name initialize-with="" name-as-sort-order="all" sort-separator=" "/>
@@ -27,13 +32,12 @@
     </names>
   </macro>
   <macro name="editor">
-    <group prefix="Eds ">
-      <names variable="editor" suffix=". ">
-        <name and="symbol" initialize-with="" name-as-sort-order="all" sort-separator=" ">
-          <name-part name="given" text-case="uppercase"/>
-        </name>
-      </names>
-    </group>
+    <names variable="editor" suffix=". ">
+      <label form="short" strip-periods="true" text-case="capitalize-first" suffix=" "/>
+      <name and="symbol" initialize-with="" name-as-sort-order="all" sort-separator=" ">
+        <name-part name="given" text-case="uppercase"/>
+      </name>
+    </names>
   </macro>
   <macro name="publisher">
     <group delimiter=": " suffix=",">
@@ -80,7 +84,10 @@
   <macro name="edition">
     <choose>
       <if is-numeric="edition">
-        <number prefix="edn " variable="edition" suffix=", "/>
+        <group delimiter=" " suffix=", ">
+          <text term="edition" form="short"/>
+          <number variable="edition"/>
+        </group>
       </if>
       <else>
         <text variable="edition" suffix="."/>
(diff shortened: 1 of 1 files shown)


sylverestoermann:
Resolves https://forums.zotero.org/discussion/35797/style-request-european-journal-of-endocrinology

european-journal-of-endocrinology.csl
```diff
@@ -0,0 +1,143 @@
...
+    </names>
+  </macro>
+  <macro name="editor">
+    <group prefix="Eds ">
+      <names variable="editor" suffix=". ">
+        <name and="symbol" initialize-with="" name-as-sort-order="all" sort-separator=" ">
+          <name-part name="given" text-case="uppercase"/>
+        </name>
+      </names>
+    </group>
```
rmzelle: @adam3smith, do you know what we usually do in these cases? http://www.eje-online.org/site/misc/For-Authors.xhtml#refs indeed shows "Eds J-C Job & M Pierson", but http://docs.citationstyles.org/en/stable/specification.html#label-in-cs-names says

> The optional cs:label element (see label) must be included after the cs:name and cs:et-al elements, but before the cs:substitute element. 

So we can't put the label before `<name/>`. That seems like an oversight, and I couldn't find a ticket to allow it.

european-journal-of-endocrinology.csl
```diff
@@ -0,0 +1,143 @@
...
+      </else>
+    </choose>
+  </macro>
+  <macro name="title">
+    <text variable="title"/>
+  </macro>
+  <macro name="edition">
+    <choose>
+      <if is-numeric="edition">
+        <number prefix="edn " variable="edition" suffix=", "/>
```
rmzelle: @sylverestoermann, thanks. I already simplified the style a little by removing some superfluous `<group/>` elements (they don't add anything if they only have a single child element).

Otherwise, if you're up for it, the main thing you could improve is to rely less on verbatim text labels like `prefix="edn "` above, and instead use CSL terms, which make styles easier to reuse in other languages, and which can automatically adapt to whether the variable contains a singular or plural value.

E.g. here, you could change `<number prefix="edn " variable="edition" suffix=", "/>` to

```xml
        <group delimiter=" " suffix=", ">
          <text term="edition" form="short"/> 
          <number variable="edition"/>
        </group>
```

(in this case, because the standard translation for the "short" form of the "edition" in British English is "ed" [see https://github.com/citation-style-language/locales/blob/master/locales-en-GB.xml#L34 ], you'd have to redefine this term as well by adding a locale section after the info section, e.g.

```xml
  <locale xml:lang="en">
    <terms>
      <term name="edition" form="short">edn</term>
    </terms>
  </locale>
```
)

If that's too complicated we can accept the style like this as well.

sylverestoermann: Thank you for the revisions, I highly appreciate it.

I understand how technically it would be cleaner to use CSL terms, but I have a full-time job as physician, a business on the side and a loving wife and children. So time is sparse and I have to carefully select what to use it for. The journal has not dramatically changed its citation style in the past few years (cf. for example author guidelines in the Wayback Machine) and there likely won't be major changes in the future (especially regarding the abbreviations used). It is also unlikely that localization will be an issue unless maybe the whole scientific community decides to change its lingua franca (which would not happen overnight anyways). So, in essence, the benefit of achieving perfection in technical terms (no change in output) does not warrant the cost (I'm not that adept at CSL that I could do this quickly).

european-journal-of-endocrinology.csl
adam3smith: that must be a typo in the specs and we don't validate this, so it's OK to allow imo. Otherwise we're in trouble: 
We have cs:label before cs:name all over the place, e.g. for pretty much any verb form of the lable, see e.g. https://github.com/citation-style-language/styles/blob/master/chicago-fullnote-bibliography.csl#L56

rmzelle: Ah, you're right: https://github.com/citation-style-language/schema/blob/v1.0.1/csl.rnc#L423

Mea culpa. I created a ticket to correct this in the spec: https://github.com/citation-style-language/documentation/issues/45

european-journal-of-endocrinology.csl
rmzelle: Sure, no problem!

(just to clarify: were not concerned with this journal changes its publishing language, but with being able to reuse (parts of) this CSL style for publications in other languages, which is easier if the style doesn't use hard-coded English text. But it's not that big of a deal here, so don't worry about it.)

rmzelle:
@sylverestoermann, thanks for contributing!

sylverestoermann:
You're welcome. Thanks for your help!

PR #2392, opened 2017-01-03
plastic-and-reconstructive-surgery.csl

damnation333:
https://forums.zotero.org/discussion/61335/style-request-plastic-and-reconstructive-surgery/

### DIFF
--- a/dependent/plastic-and-reconstructive-surgery.csl
+++ b/dependent/plastic-and-reconstructive-surgery.csl
@@ -9,7 +9,7 @@
     <link href="http://edmgr.ovid.com/prs/accounts/ifauth.htm" rel="documentation"/>
     <category citation-format="numeric"/>
     <category field="medicine"/>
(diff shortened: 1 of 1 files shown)


dependent/plastic-and-reconstructive-surgery.csl
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" default-locale="en-US">
+  <!-- Generated with https://github.com/citation-style-language/utilities/tree/master/generate_dependent_styles/data/npg -->
```
rmzelle: @damnation333, we use these XML comments in automatically generated dependent styles to keep track from which metadata folder at https://github.com/citation-style-language/journals they originated. It should be absent from all manually edited styles.

rmzelle:
Thanks!

Issue #2317, opened 2016-11-23
Journal of Vacuum Science & Technology has moved


adam3smith:
journal-of-vacuum-science-and-technology-b-microelectronics-and-nanometer-structures
journal-of-vacuum-science-and-technology-a-vacuum-surfaces-and-films
seem to have moved from IEEE to AIP, including their citation styles. Need to check which style exactly they're using now and switch.

damnation333:
@adam3smith 
This could be closed as it's been dealt with? https://github.com/citation-style-language/journals/pull/17

rmzelle:
@damnation333, yup. Thanks.

PR #2396, opened 2017-01-03
Create journal-of-vacuum-science-and-technology-b-microelectronics-an…

damnation333:
…d-nanometer-structures.csl

see #2317
Created new independent style for journal.
The documentation is rather bad, so had to dig into actual reference lists in publications. Hard to find book and chapter references.
Part A is the same, so we can create an independent style for it.

### DIFF
--- a/journal-of-vacuum-science-and-technology-b-microelectronics-and-nanometer-structures.csl
+++ b/journal-of-vacuum-science-and-technology-b-microelectronics-and-nanometer-structures.csl
@@ -1,0 +1,140 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" and="text" demote-non-dropping-particle="sort-only" default-locale="en-US" xmlns="http://purl.org/net/xbiblio/csl">
+  <info>
+    <title>Journal of Vacuum Science &amp; Technology B: Microelectronics and Nanometer Structures</title>
+    <id>http://www.zotero.org/styles/journal-of-vacuum-science-and-technology-b-microelectronics-and-nanometer-structures</id>
+    <link href="http://www.zotero.org/styles/journal-of-vacuum-science-and-technology-b-microelectronics-and-nanometer-structures" rel="self"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
did you check if there are existing AIP styles that apply?

rmzelle:
> didn't look into that and could've saved myself the work.

Since you accumulated some experience points and might be ready to level up, are you aware of https://github.com/citation-style-language/journals? The majority of our dependent CSL styles have been automatically generated from spreadsheets with journal metadata stored in that repo. The files for AIP journals can be found at https://github.com/citation-style-language/journals/tree/master/aip. We can explain in more detail how things work over there.

rmzelle:
(so, in this particular case, if you can confirm for us that american-institute-of-physics.csl covers this style format, we would close this pull request, update https://github.com/citation-style-language/journals/blob/master/aip/_journals.tab with the two journals, and run https://github.com/citation-style-language/journals/blob/master/generate_styles.rb to regenerate all the AIP dependent CSL styles from the metadata)

damnation333:
@rmzelle 
Amazing. This is really great. Definitely makes it all easier then. Thanks for pointing it out and saving me all the manual labour. Makes more sense like that! :)
I cannot confirm though that this journal uses the general AIP style. (I could assume though that at some stage they will be though with time)

rmzelle:
> I cannot confirm though that this journal uses the general AIP style.

The journal instructions (e.g. http://avs.scitation.org/pb-assets/files/publications/jva/JVST_Article_Template_January_2016.doc via http://avs.scitation.org/jva/authors/manuscript) seem close enough to the instructions at https://publishing.aip.org/authors/preparing-your-manuscript, so it's probably safe to just use the AIP style for them.

damnation333:
closed for independent styles in #17 in /journals

rmzelle:
> closed for independent styles in # 17 in /journals

@damnation333, you can just paste in the entire `https://github.com/citation-style-language/journals/pull/17` URL, which GitHub will automatically turn into https://github.com/citation-style-language/journals/pull/17. The `#17` you used automatically linked to issue 17 in the styles repo, which isn't correct.

PR #2397, opened 2017-01-05
Changes currently being added to the style guide

georgd:
- Add pages to legislation for EU laws
- Replace "an der" with "," between thesis type and university name
- Remove unnecessary translation


rmzelle:
Thanks!

PR #2398, opened 2017-01-05
Create tropical-animal-health-and-production.csl

damnation333:
Code was here, but had to be modified: https://forums.zotero.org/discussion/42877/style-request-tropical-animal-health-and-production/
I am struggling with the chapters. How can I get it to display the page number at the very end? When I add it into the "publisher" macro it duplicates it for normal article-journals, too. 
The rest all works.

### DIFF
--- a/tropical-animal-health-and-production.csl
+++ b/tropical-animal-health-and-production.csl
@@ -126,9 +126,11 @@
       <text variable="publisher"/>
       <text variable="publisher-place"/>
     </group>
-    <if type="chapter">
-      <text variable="page" prefix=", "/>
-    </if>  
+    <choose>
+      <if type="chapter">
+        <text variable="page" prefix=", "/>
+      </if>  
+    </choose>  
   </macro>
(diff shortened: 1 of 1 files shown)


tropical-animal-health-and-production.csl
```diff
@@ -0,0 +1,163 @@
...
+          </if>
+        </choose>
+      </if>
+    </choose>
+  </macro>
+  <macro name="chapter-info">
+    <choose>
+      <if type="chapter paper-conference" match="any">
+        <group delimiter=", ">
+          <group delimiter=" ">
```
rmzelle: The outer group doesn't do anything here.

rmzelle:
> I am struggling with the chapters. How can I get it to display the page number at the very end? When I add it into the "publisher" macro it duplicates it for normal article-journals, too.

Well, there are a bunch of ways. You could make the output in the macro conditional on item type (e.g. `<if type="chapter">`), or you could split up the `<layout/>` section in the bibliography by item type (again with a conditional).

damnation333:
That works now, but I guess it doens't validate now because there is no <else> following the <if>. How can I make it have that <else> without doing anything?

rmzelle:
The problem is that `<if/>` needs to be wrapped inside a `<choose/>`. You can't use it by itself.

PR #2399, opened 2017-01-06
Create discovery-medicine.csl

damnation333:
https://forums.zotero.org/discussion/56278/style-request-discovery-medicine/
Checked with actual publications for books/chapters as not specified in guidelines

### DIFF
--- a/discovery-medicine.csl
+++ b/discovery-medicine.csl
@@ -72,26 +72,6 @@
       <date-part name="year"/>
     </date>
   </macro>
-  <macro name="access">
-    <choose>
-      <if variable="URL">
-        <group delimiter=" ">
-          <group prefix="[" suffix="]" delimiter=" ">
-            <text term="cited"/>
-            <date variable="accessed">
-              <date-part name="year"/>
-              <date-part name="month" prefix=" " form="short" strip-periods="true"/>
-              <date-part name="day" prefix=" "/>
-            </date>
-          </group>
-          <group>
-            <text value="Available from: "/>
-            <text variable="URL"/>
-          </group>
-        </group>
-      </if>
-    </choose>
-  </macro>
   <macro name="title">
     <choose>
       <if type="book" match="any">
--- a/discovery-medicine.csl
+++ b/discovery-medicine.csl
@@ -1,0 +1,168 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US" xmlns="http://purl.org/net/xbiblio/csl">
+  <info>
+    <title>Discovery Medicine</title>
+    <title-short>DM</title-short>
+    <id>http://www.zotero.org/styles/discovery-medicine</id>
+    <link href="http://www.zotero.org/styles/discovery-medicine" rel="self"/>
+    <link href="http://www.zotero.org/styles/brain" rel="template"/>
+    <link href="http://www.oxfordjournals.org/our_journals/brainj/for_authors/general.html" rel="documentation"/>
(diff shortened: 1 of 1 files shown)


discovery-medicine.csl
```diff
@@ -0,0 +1,148 @@
...
+      <choose>
+        <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
+          <group prefix=" " delimiter=" ">
+            <text macro="edition"/>
+            <text macro="publisher"/>
+          </group>
+        </if>
+        <else-if type="chapter paper-conference" match="any">
+          <text term="in" text-case="capitalize-first" font-style="italic" suffix=": "/>
+          <group delimiter=". " prefix=" " suffix=". ">
```
adam3smith: no need for the prefix here. You have a space from the suffix above.

discovery-medicine.csl
```diff
@@ -0,0 +1,148 @@
...
+          <number variable="edition" form="ordinal"/>
+          <text term="edition" form="short"/>
+        </group>
+      </if>
+      <else>
+        <text variable="edition" suffix="."/>
+      </else>
+    </choose>
+  </macro>
+  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year">
```
adam3smith: delete: `disambiguate-add-names="true"`:
"If there are multiple references with the same first author and same publishing year, please add a letter after the year to differentiate the references"
(With the preference it would add additional names if the 2nd, 3rd etc. author are different)

discovery-medicine.csl
```diff
@@ -0,0 +1,148 @@
...
+  <macro name="citation-locator">
+    <group>
+      <label variable="locator" form="short"/>
+      <text variable="locator" prefix=" "/>
+    </group>
+  </macro>
+  <macro name="author">
+    <names variable="author" suffix=". ">
+      <name sort-separator=" " initialize-with="" name-as-sort-order="all" delimiter=", " delimiter-precedes-last="always"/>
+    </names>
```
adam3smith: I always make sure the `substitute` elements in author date styles conform between author and author-short macro. Otherwise, how would authors find the bib entry corresponding to a citation. In other words, add editor, translator, and title as substitute elements here. 
With the editor and translator now options, I would then also suggest adding `<label form="short" prefix=" (" suffix=")"/>`

discovery-medicine.csl
```diff
@@ -0,0 +1,148 @@
...
+<style class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US" xmlns="http://purl.org/net/xbiblio/csl">
+  <info>
+    <title>Discovery Medicine</title>
+    <title-short>DM</title-short>
+    <id>http://www.zotero.org/styles/discovery-medicine</id>
+    <link href="http://www.zotero.org/styles/discovery-medicine" rel="self"/>
+    <link href="http://www.zotero.org/styles/brain" rel="template"/>
+    <link href="http://www.discoverymedicine.com/about/information-for-authors-of-discovery-medicine/" rel="documentation"/>
+    <author>
+      <name>O'Brien Patrick, PhD</name>
```
rmzelle: @damnation333, do you care about the order here? I've been changing a few `<name>O'Brien Patrick, PhD</name>` to `<name>Patrick O'Brien, PhD</name>` in recent submissions.

rmzelle:
@adam3smith, I let this sit because the "author" and "author-short" macros still not fully align ("author" has no title substitution). Do you want to see that changed?

rmzelle:
Thanks!

Issue #5, opened 2011-03-10
Create Quality Markers


adam3smith:
Inspired by [this tweet](http://twitter.com/mrgunn/status/45627036361957376).
It would be very useful if users could gauge the level of vetting and thoroughness that has gone into a citation style - e.g. CMoS and APA are very good, APSA is only soso, etc. Could be done either with some type of crowd evaluate tool, or simply indicated by the author/committer.

PR #2400, opened 2017-01-06
Create journal-of-the-american-adacemy-of-audiology.csl

damnation333:
https://forums.zotero.org/discussion/62471/style-request-journal-of-american-academy-of-audiology/p1


journal-of-the-american-adacemy-of-audiology.csl
```diff
@@ -0,0 +1,204 @@
...
+    <category field="medicine"/>
+    <issn>1050-0545</issn>
+    <eissn>2157-3107</eissn>
+    <updated>2017-01-06T12:09:37+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <macro name="editor">
+    <names variable="editor">
+      <name delimiter=", " name-as-sort-order="all" sort-separator=", " initialize-with="">
+        <name-part name="family" text-case="capitalize-first"/>
```
adam3smith: I don't quite get what this is doing? names would typically be capitalize-first, no?

journal-of-the-american-adacemy-of-audiology.csl
```diff
@@ -0,0 +1,204 @@
...
+                <date-part name="day" suffix=", "/>
+                <date-part name="year"/>
+              </date>
+            </group>
+          </if>
+        </choose>
+      </else>
+    </choose>
+  </macro>
+  <macro name="title">
```
adam3smith: this entire macro doesn't do anything. Delete it and replace any `<text macro="title"/>` calls by `<text variable="title"/>`

journal-of-the-american-adacemy-of-audiology.csl
```diff
@@ -0,0 +1,204 @@
...
+        </group>
+      </if>
+      <else>
+        <text variable="edition" suffix="."/>
+      </else>
+    </choose>
+  </macro>
+  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year">
+    <layout prefix="(" suffix=")" delimiter="; ">
+      <group delimiter=", ">
```
adam3smith: no need for two nested groups with identical delimiter. Remove one.

journal-of-the-american-adacemy-of-audiology.csl
```diff
@@ -0,0 +1,204 @@
...
+      <key variable="title"/>
+    </sort>
+    <layout suffix=".">
+      <group delimiter=" ">
+        <text macro="author" suffix="."/>
+        <text macro="year-date" prefix="(" suffix=")"/>
+      </group>
+      <choose>
+        <if type="article-newspaper article-magazine" match="any">
+          <group delimiter=" ">
```
adam3smith: no need for a group with a single element. remove.

journal-of-the-american-adacemy-of-audiology.csl
```diff
@@ -0,0 +1,204 @@
...
+  <macro name="editor">
+    <names variable="editor">
+      <name delimiter=", " name-as-sort-order="all" sort-separator=", " initialize-with="">
+        <name-part name="family" text-case="capitalize-first"/>
+        <name-part name="given" text-case="capitalize-first"/>
+      </name>
+      <label form="short" prefix=", "/>
+    </names>
+  </macro>
+  <macro name="series-editor">
```
adam3smith: This isn't by you, right? It's very odd a) becuase the macro is called series-editor, but it uses the "original-author" variable that we honestly don't know what it was supposed to mean (it literally has a question mark in the specifications) and also is the only place where & is used between authors. I'd consider either replacing it by `collection-editor` or possibly removing the macro (and everywhere it's called) entirely.

journal-of-the-american-adacemy-of-audiology.csl
```diff
@@ -0,0 +1,204 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>Journal of the American Academy of Audiology</title>
+    <title-short>JAAA</title-short>
+    <id>http://www.zotero.org/styles/journal-of-the-american-adacemy-of-audiology</id>
```
rmzelle: @damnation333, typo: "adacemy" should be "academy" (idem for file name and "self" link).

damnation333:
Thanks @adam3smith and @rmzelle 
Oftentimes I use styles as templates that have a lot of bad quality code in them. I wonder if I should stick with a few styles that have been very much peer-reviewed and are of higher quality to use as templates. Might be more work sometimes, but would ensure the styles are written better. I'm yet too unexperienced to spot these things myself and I am very thankful for looking over them.

adam3smith:
What I do is to pick out a style with relatively high code quality among the best matches. There's not an obvious list -- some styles (APA, Vancouver, Chicago) are obviously high quality, but there are many others that are quite good, so I wouldn't limit myself to these. Often you can tell looking at the code -- e.g. good use of groups and delimiters over prefix/suffix is often a helpful indicator.

journal-of-the-american-academy-of-audiology.csl
```diff
@@ -0,0 +1,183 @@
...
+    <category field="medicine"/>
+    <issn>1050-0545</issn>
+    <eissn>2157-3107</eissn>
+    <updated>2017-01-06T12:09:37+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <macro name="editor">
+    <names variable="editor">
+      <name delimiter=", " name-as-sort-order="all" sort-separator=", " initialize-with="">
+        <name-part name="family"/>
```
adam3smith: I don't think you need the name-parts here at all? Any reason for them? Note that the order is determined by `name-as-sort-order` above. These are just needed for styling, .e.g. if you wanted last names in bold all small caps.

adam3smith:
Thanks!

rmzelle:
> Oftentimes I use styles as templates that have a lot of bad quality code in them. I wonder if I should stick with a few styles that have been very much peer-reviewed and are of higher quality to use as templates.

Yes, it's a bit unfortunate that we don't have a good way to improve overall style quality in the repository. Fixing up all individual styles is way too much work to do one-by-one, and is often hard to automate, so the best we can do right now is indeed to try to start from high-quality styles when writing new ones.

It would be good if we had some automatic quality assignments. There are some (limited) criteria that we could test for as warning signs: e.g. lots of affixes with punctuation in them (which often should be put on groups instead) and lots of affixes with text in them (should often be replaced by CSL terms). Highly used styles are generally also higher quality, because more people have looked at them. There has been some talk of crowd-sourced quality scores as well (https://github.com/citation-style-language/styles/issues/5), but that would probably focus more on how well the style performs, rather than how well it's written.

If we'd want to change the "Popular" styles in the Search by Example tool, by the way, it looks like we can do that at https://github.com/citation-style-language/csl-editor/blob/c07fc753a9405b2930728696ff950a87881fc957/src/cslStyles.js#L12.

Issue #2359, opened 2016-12-15
Journal of Polymer Science Part A citation style


pingitoreat52:
There is no CSL style for the Journal of Polymer Science Part A: Polymer Chemistry (http://onlinelibrary.wiley.com/journal/10.1002/(ISSN)1099-0518).

adam3smith:
https://github.com/citation-style-language/styles/wiki/Requesting-Styles

damnation333:
http://onlinelibrary.wiley.com/journal/10.1002/(ISSN)1099-0518/homepage/ForAuthors.html#anchor53

Citations

Literature citations should be formatted to the following guidelines, using the Chemical Abstracts Service Source Index (CASSI) recommendations for title abbreviations.
Books	1. Koros, W. J.; Chern, R. T. In Handbook of Separation Process Technology; Rousseau, E. D.; Russell, B., Eds.; Wiley: New York, 1987; Vol. 2, Chapter 20, pp 34-45.
Book Series	2. Koros, W. J.; Chern, R. T. In Polymers for Fun and Profit; Rousseau, E. D., Ed.; ACS Symposium Series 518; American Chemical Society: Washington, DC, 1997; p 45.
Journals	3. Guan, L., Jiménez, M. E. G., Walowski, C., Boushehri, A., Prausnitz, J. M. and Radke, C. J. J. Appl. Polym. Sci. 2011, 122, 1457–1471.
Conference Proceedings	4. Medium, K. P. In Polymers Revealed, Proceedings of the 5th Annual Conference on Hidden Polymers, Istanbul, Turkey, Sept 5–21, 1976; Bluto, M., Ed.; Elsevier: New York, 1977.
Theses	5. Big, G. H. Polymers in the Automotive Industry. Ph.D. Thesis, University of Scoggins, Fort Lee, MN, June 1954.
Patents	6. Norman, L. O. (Michelin Co.). U.S. Patent 4,199,533, April 22, 1980.
Online Sources	7. J. A. Woolam Co., Ellipsometry tutorial. http://www.jawoollam.com/tutorial_1.html (accessed May 1, 2010). 

ISSN:
    0887-624X (Print)
    0887-624X (Linking)

@pingitoreat52 can you adapt the two citations as outlined in the "Requesting Styles" manual to the style of the journal?

rmzelle:
[...]

PR #2403, opened 2017-01-06
Create journal-of-polymer-science-part-a-polymer-chemistry.csl

damnation333:
See #2359


damnation333:
@rmzelle 
That it puts the "template" in front is done by the CSL Visual Editor btw. Sometimes is use that for ease. I'll keep an eye on it that it doesn't happen again.

rmzelle:
@damnation333, yes, we should just fix the Visual Editor so that it keeps the attributes in our preferred order. You can fix it quickly by running your style through http://formatter.citationstyles.org/.

(I also like to keep `xmlns="http://purl.org/net/xbiblio/csl"` directly after `<style `, which is something the formatter currently doesn't fix. XML officially doesn't care about attribute order, but I find styles easier to read if things are organized in a somewhat consistent matter)

PR #2405, opened 2017-01-08
Create journal-of-water-sanitation-and-hygiene-for-development.csl

damnation333:
https://forums.zotero.org/discussion/55496/new-style-request/

### DIFF
--- a/journal-of-water-sanitation-and-hygiene-for-development.csl
+++ b/journal-of-water-sanitation-and-hygiene-for-development.csl
@@ -5,7 +5,7 @@
     <title-short>JWaSHDev</title-short>
     <id>http://www.zotero.org/styles/journal-of-water-sanitation-and-hygiene-for-development</id>
     <link href="http://www.zotero.org/styles/journal-of-water-sanitation-and-hygiene-for-development" rel="self" />
-    <link href="http://www.zotero.org/styles/apa-old-doi-prefix" rel="template" />
+    <link href="http://www.zotero.org/styles/apa" rel="template" />
     <link href="http://washdev.iwaponline.com/content/instructions-authors" rel="documentation" />
     <author>
       <name>Patrick O'Brien, PhD</name>
@@ -140,7 +140,7 @@
       <else>
         <choose>
           <if variable="DOI">
-            <text variable="DOI" prefix="doi:" />
+            <text variable="DOI" prefix="https://doi.org/" />
           </if>
           <else>
             <choose>
--- a/journal-of-water-sanitation-and-hygiene-for-development.csl
+++ b/journal-of-water-sanitation-and-hygiene-for-development.csl
@@ -1,0 +1,604 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" default-locale="en-GB" demote-non-dropping-particle="never">
+  <info>
+    <title>Journal of Water Sanitation and Hygiene for Development</title>
+    <title-short>JWaSHDev</title-short>
+    <id>http://www.zotero.org/styles/journal-of-water-sanitation-and-hygiene-for-development</id>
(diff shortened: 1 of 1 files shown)


journal-of-water-sanitation-and-hygiene-for-development.csl
```diff
@@ -0,0 +1,604 @@
...
+    <link href="http://www.zotero.org/styles/journal-of-water-sanitation-and-hygiene-for-development" rel="self" />
+    <link href="http://www.zotero.org/styles/apa-old-doi-prefix" rel="template" />
+    <link href="http://washdev.iwaponline.com/content/instructions-authors" rel="documentation" />
+    <author>
+      <name>Patrick O'Brien, PhD</name>
+      <email>obrienpat86@gmail.com</email>
+    </author>
+    <category citation-format="author-date" />
+    <category field="science" />
+    <eissn>2043-9083</eissn>
```
rmzelle: @damnation333, you're a bit sloppy with mixing up the print and online ISSNs :). See the bottom of the page at http://washdev.iwaponline.com/: "ISSN Print: 2043-9083 | ISSN Online: 2408-9362"

journal-of-water-sanitation-and-hygiene-for-development.csl
```diff
@@ -0,0 +1,604 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" default-locale="en-GB" demote-non-dropping-particle="never">
+  <info>
+    <title>Journal of Water Sanitation and Hygiene for Development</title>
+    <title-short>JWaSHDev</title-short>
+    <id>http://www.zotero.org/styles/journal-of-water-sanitation-and-hygiene-for-development</id>
+    <link href="http://www.zotero.org/styles/journal-of-water-sanitation-and-hygiene-for-development" rel="self" />
+    <link href="http://www.zotero.org/styles/apa-old-doi-prefix" rel="template" />
```
rmzelle: @damnation333, I don't see any indication in http://washdev.iwaponline.com/content/instructions-authors that this journal requires "doi:" instead of "https://doi.org/". Since the latter is the preferred form nowadays (because it creates an URL that can be directly accessed), I wouldn't hesitate to push it unless the style guide explicitly wants something else, even if published articles use an outdated format like "doi:".

damnation333: I swear I saw it somewhere and made that choice specifically, but of course I can't find it now. Let me do some digging and come back here or make the changes.

journal-of-water-sanitation-and-hygiene-for-development.csl
damnation333: It's due to not having found a proper source yet. If I don't find them on Wikipedia (which lists ISSN and eISSN) I look on Worldcat.org, but that only ever lists one number (which so far I have assumed was the ISSN). I'll make sure this doesn't happen anymore. Sorry.

adam3smith:
The NLM Catalog is remarkably good for journal metadata, even where the journal has very little to do with medicine such as here:
https://www.ncbi.nlm.nih.gov/nlmcatalog/?term=journal-of-water-sanitation-and-hygiene-for-development

journal-of-water-sanitation-and-hygiene-for-development.csl
rmzelle: My favorite place to check ISSNs is NCBI's NLM Catalog. See e.g. https://www.ncbi.nlm.nih.gov/nlmcatalog/?term=2043-9083. They are typically very accurate, but they don't cover all journals. Almost anything about biology or medicine should be in there, though.

rmzelle: If I can only find a single ISSN, I often also just google the ISSN to see what pops up.

adam3smith:
great, thanks!

PR #2406, opened 2017-01-08
Create journal-of-polymer-science-part-b-polymer-physics.csl

damnation333:


damnation333:
oops. needs to be in the dependent folder. Can we move files in here or do I need to close this and make a new one?

rmzelle:
You can move it. See https://help.github.com/articles/moving-a-file-to-a-new-location/

If I got this correct, you should be able to start editing the file name, and then type in "dependent/" in front of the original file type.

rmzelle:
Thanks!

PR #2407, opened 2017-01-09
Improve Legal citations in APA

adam3smith:
align APA variants

### DIFF
--- a/apa-annotated-bibliography.csl
+++ b/apa-annotated-bibliography.csl
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
-  <!-- This style was edited with the Visual CSL Editor (http://steveridout.com/csl/visualEditor/) -->
   <info>
     <title>American Psychological Association 6th edition (annotated bibliography with abstract)</title>
     <title-short>APA</title-short>
@@ -25,6 +24,10 @@
     <contributor>
       <name>Sebastian Karcher</name>
     </contributor>
+    <contributor>
+      <name> Brenton M. Wiernik</name>
+      <email>zotero@wiernik.org</email>
+    </contributor>
     <category citation-format="author-date"/>
     <category field="psychology"/>
     <category field="generic-base"/>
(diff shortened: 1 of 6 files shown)


apa-annotated-bibliography.csl
```diff
@@ -47,20 +50,32 @@
...
+        <group delimiter=", ">
+          <names variable="container-author" delimiter=", ">
+            <name and="symbol" initialize-with=". " delimiter=", "/>
+            <label form="short" prefix=" (" text-case="title" suffix=")"/>
+          </names>
+          <names variable="editor translator" delimiter=", ">
+            <name and="symbol" initialize-with=". " delimiter=", "/>
+            <label form="short" prefix=" (" text-case="title" suffix=")"/>
+          </names>
+        </group>
```
rmzelle: @adam3smith, just curious, but isn't the original code here compacter? (apart from the variable order, shouldn't the output be the same?)

apa-annotated-bibliography.csl
```diff
@@ -413,7 +496,15 @@
...
-          <date variable="issued" form="text"/>
+	  <choose>
+	    <if variable="container-title" match="any">
+	      <!--Only print year for cases published in reporters-->
+	      <date variable="issued" form="numeric" date-parts="year"/>
+	    </if>
+	    <else>
+              <date variable="issued" form="text"/>
+	    </else>
+	  </choose>
```
rmzelle: @adam3smith, there are a bunch of redundant `match="any"`  attributes on elements with a single test.

apa-annotated-bibliography.csl
adam3smith: no, the output isn't the same (and given the spec wording on editor translator probably shouldn't be). This is an old fix in APA, just updating the other styles. Background here:
https://github.com/citation-style-language/styles/pull/1465

apa-annotated-bibliography.csl
adam3smith: yes, that's right.

apa-annotated-bibliography.csl
rmzelle: Ah, it's specific to editortranslator collapsing. Never mind, then.

apa-annotated-bibliography.csl
adam3smith: @rmzelle though looking at this (partly because there are a ton of these, most of them not new) -- I'm wondering if it makes sense to discourage explicitly setting match attributes. I don't think they make the style harder to read (the way single-member groups do, say) and they prevent people from forgetting to set the attribute when adding additional variables.

adam3smith:
I've gone ahead and accepted that and moved the question of match clean-up to a separate ticket.

PR #2411, opened 2017-01-09
Create international-brazilian-journal-of-urology.csl

damnation333:
https://forums.zotero.org/discussion/58755/style-request-international-braz-journal-of-urology/


international-brazilian-journal-of-urology.csl
```diff
@@ -0,0 +1,172 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" page-range-format="minimal-two" demote-non-dropping-particle="never" default-locale="en-US" xmlns="http://purl.org/net/xbiblio/csl">
+  <info>
+    <title>International Brazilian Journal of Urology</title>
```
rmzelle: @adam3smith, @damnation333, any thoughts on the title? It looks like the full title is really just "International Braz J Urol". Shall we go with "International Brazilian Journal of Urology" regardless? 

It's rather uncommon in that the journal doesn't seem to use the unabbreviated title anywhere. See also
https://www.ncbi.nlm.nih.gov/nlmcatalog/?term=1677-5538 and http://www.intbrazjurol.com.br/.

adam3smith: I'd be inclined to use the official, i.e. abbreviated title

rmzelle: Okay, let's do that, then. @damnation333, can you make the change?

Personally, I'm not a big fan of journals who adopt acronyms or abbreviations as their official title (I'm looking at you, [BMJ](http://www.bmj.com/)!), but I do think we should follow journal's own naming as much as possible.

damnation333: Made the changes. Very odd though. 
Thinking from a user perspective, will people find this journal in the repo if they type in the full name?

rmzelle: I've asked them at https://www.facebook.com/intbrazjurol/ via a private message (that Facebook page might be the only place that uses the full title) whether they can give any guidance.

rmzelle:
(I decided to go with the full unabbreviated title after all; I found some mentions at https://www.facebook.com/intbrazjurol/ and the page title of http://www.intbrazjurol.com.br/)

rmzelle:
Thanks!

PR #2416, opened 2017-01-12
Update the-journal-of-pure-and-applied-chemistry-research.csl

Firda09:
Hello, I have updated the CSL. is it still look like ACS style? how to get the real repository in Mendeley CSL so that I can get the real name of my repository (The Journal of Pure and Applied Chemistry Research) as I proposed? Thank you for your help.

### DIFF
--- a/the-journal-of-pure-and-applied-chemistry-research.csl
+++ b/the-journal-of-pure-and-applied-chemistry-research.csl
@@ -3,7 +3,7 @@
   <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
     <title>The Journal of Pure and Applied Chemistry Research</title>
-    <title-short>JPACR</title-short>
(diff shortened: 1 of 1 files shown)


rmzelle:
> Hello, I have updated the CSL. is it still look like ACS style? how to get the real repository in Mendeley CSL so that I can get the real name of my repository (The Journal of Pure and Applied Chemistry Research) as I proposed? Thank you for your help.

Sorry, I don't really understand what your question is. Can you elaborate a little? (if you're not a native speaker of English, feel free to write in your native language)

adam3smith:
(the CSL changes look good, btw.; I think the JAPCR acronym is more useful than the ISO abbreviation, so I'd revert that)

damnation333:
@Firda09 If your questions means that your newly edited style doesn't show up for you in Mendeley, it's because the old style from the repository and your new style share the same ID. So on your local computer change the ID field and file name to something like "the-journal-of-pure-and-applied-chemistry-research-test and it should appear in Mendeley.

PR #2417, opened 2017-01-12
Create journal-of-mechanical-science-and-technology.csl

damnation333:
Per https://forums.zotero.org/discussion/54602/style-request-journal-of-mechanical-science-and-technology


rmzelle:
Thanks!

PR #2418, opened 2017-01-12
Update journal-of-wildlife-diseases.csl

damnation333:
adapted style to fit with current papers.
Per: https://forums.zotero.org/discussion/51108/updated-style-request-journal-of-wildlife-diseases


rmzelle:
Thanks!

PR #2419, opened 2017-01-12
Create taxon.csl

damnation333:
Per: https://forums.zotero.org/discussion/47631/style-request-taxon


taxon.csl
```diff
@@ -0,0 +1,600 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
```
rmzelle: @damnation333, did you leave out the "default-locale" on purpose?

The instructions say "Manuscripts must be written in correct English (either American 
or British spelling is acceptable, but this must be consistent within 
each manuscript)", but the examples use dates in the format "(accessed 19 Jan 2012)", which is more typical of British English, so I'd go with that.

damnation333: Sorry. Wasn't on purpose. Thanks for spotting that.

rmzelle:
Thanks!

PR #2420, opened 2017-01-13
Create canadian-journal-of-physics.csl

damnation333:
Per https://github.com/citation-style-language/journals/issues/10


rmzelle:
Thanks!

PR #2421, opened 2017-01-13
Minor updates and improvements to the style

ahahl:
Hi, 
I have improved some functions (viz. the original edition parameters), added no-date and anonymous fields, added a disambiguation parameter, etc.


rmzelle:
(@ahahl, next time, we prefer it if you just update your first pull request, instead of closing it and creating a new one; @csl-bot gives instructions on how to do above)

PR #2423, opened 2017-01-13
Update and rename dependent/canada-journal-of-public-health.csl to ca…

damnation333:
…n-j-public-health.csl

https://forums.zotero.org/discussion/45290/citation-style-request-canadian-journal-of-public-health

This style was wrong. Completely made a new and independent style based on Vancouver/NLM. The name of this journal was also wrong. There is no "Canada Journal of Public Health"  and from what I've learned from you by now, you'll probably actually say the real name is "Can J Public Health". Le't see :)

Default locale left out as one can write in english or french.


rmzelle:
@damnation333, no, we'd want to use the full name here.

E.g. http://journal.cpha.ca/index.php/cjph starts with "The Canadian Journal of Public Health is ...", the articles themselves use the full journal title (see e.g. http://journal.cpha.ca/index.php/cjph/article/view/5986/3495), and PubMed uses the full name too (https://www.ncbi.nlm.nih.gov/nlmcatalog/?term=1920-7476).

The Brazil case was different, because that journal didn't seem to use the full name anywhere on its website, which is very rare.

rmzelle:
Thanks!

PR #2424, opened 2017-01-14
Create the-journal-of-parasitology.csl

damnation333:
Per https://forums.zotero.org/discussion/62380/style-request-journal-of-parasitology

Do we need the two "subsequent..." arguments in line 189?
Re line 99: is there a label for "number-of-pages" I can use instead of verbatim text?

### DIFF
--- a/the-journal-of-parasitology.csl
+++ b/the-journal-of-parasitology.csl
@@ -73,24 +73,6 @@
       </substitute>
     </names>
   </macro>
-  <macro name="access">
-    <choose>
-      <if variable="URL">
-        <group>
-          <text term="retrieved" text-case="capitalize-first" suffix=" "/>
-          <date variable="accessed" suffix=", ">
-            <date-part name="month" suffix=" "/>
-            <date-part name="day" suffix=", "/>
-            <date-part name="year"/>
-          </date>
-          <group>
-            <text term="from" suffix=" "/>
-            <text variable="URL"/>
-          </group>
-        </group>
-      </if>
-    </choose>
-  </macro>
   <macro name="publisher">
     <text variable="genre" suffix=", "/>
     <group delimiter=", ">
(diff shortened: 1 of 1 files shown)


rmzelle:
> Do we need the two "subsequent..." arguments in line 189?
> Re line 99: is there a label for "number-of-pages" I can use instead of verbatim text?

It's easier if you comment on your own pull request, at the relevant line.

the-journal-of-parasitology.csl
```diff
@@ -0,0 +1,188 @@
...
+    </sort>
+    <layout prefix="(" suffix=")" delimiter="; ">
+      <group delimiter=", ">
+        <text macro="author-short"/>
+        <text macro="issued-year"/>
+        <text macro="citation-locator"/>
+      </group>
+    </layout>
+  </citation>
+  <bibliography et-al-min="3" et-al-use-first="1" subsequent-author-substitute="———" subsequent-author-substitute-rule="complete-each" entry-spacing="0" line-spacing="2">
```
rmzelle: I don't see anything on subsequent author substitution in the guidelines, so I would take out `subsequent-author-substitute="———" subsequent-author-substitute-rule="complete-each"`.

the-journal-of-parasitology.csl
```diff
@@ -0,0 +1,188 @@
...
+        </choose>
+      </substitute>
+    </names>
+  </macro>
+  <macro name="publisher">
+    <text variable="genre" suffix=", "/>
+    <group delimiter=", ">
+      <text variable="publisher"/>
+      <text variable="publisher-place"/>
+      <text variable="number-of-pages" suffix=" p"/>
```
rmzelle: @damnation333, well, you could use re-use the "page" term.

```xml
<group delimiter=" ">
  <text variable="number-of-pages"/>
  <text term="page" form="short"/>
</group>
```

rmzelle:
Thanks!

PR #2426, opened 2017-01-14
Create stem-cells-translational-medicine.csl

damnation333:
per https://forums.zotero.org/discussion/52852/style-request-stem-cells-translational-medicine


stem-cells-translational-medicine.csl
```diff
@@ -0,0 +1,161 @@
...
+    <category field="medicine"/>
+    <issn>2157-6564</issn>
+    <eissn>2157-6580</eissn>
+    <summary>A style for Stem Cells Translational Medicine based on Elsevier Vancouver.</summary>
+    <updated>2017-01-14T17:42:41+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <macro name="author">
+    <names variable="author">
+      <name delimiter-precedes-last="always" et-al-min="4" et-al-use-first="3" initialize-with="" name-as-sort-order="all" sort-separator=" "/>
```
rmzelle: @damnation333, does this et-al setting arrangement make the most sense? (there are also et-al setting on `<bibliography>`, but none in the "editor" macro)

It would make the style a little easier to read if there are only et-al settings on `<bibliography>`, or in the "author" and "editor" macros.

rmzelle:
By the way, it looks like the sister journal "Stem Cells" has the same format (http://www.stemcellsportal.com/STEM_CELLS_Info_for_Contributors#_Toc372810938) so we could add a dependent. (in which case, I would prefer to make "Stem Cells" the independent one)

stem-cells-translational-medicine.csl
damnation333: @rmzelle
re the et-al: that's how it's in the guidelines. four or more authors, then the first 3 and an et al.

damnation333:
@rmzelle 
Stem Cells is already on the repo and I see it has a different format, although it should look like this one here.
shall I swap this code here into STEM CELLs and then make this here a dependent one?

damnation333:
@rmzelle 
Stem Cells is already on the repo and I see it has a different format, although it should look like this one here.
shall I swap this code here into STEM CELLs and then make this here a dependent one?

rmzelle:
> Stem Cells is already on the repo and I see it has a different format, although it should look like this one here.
> shall I swap this code here into STEM CELLs and then make this here a dependent one?

Done. Thanks for checking!

PR #2427, opened 2017-01-15
Create american-journal-of-translational-research.csl

damnation333:
Per https://forums.zotero.org/discussion/49289/style-request-american-journal-of-translational-research


rmzelle:
Thanks!

PR #2428, opened 2017-01-15
Create journal-of-korean-neurosurgical-society.csl

damnation333:
Per https://forums.zotero.org/discussion/56813/style-request-j-korean-neurosurgical-soc

A rather ugly style, I got to say.

### DIFF
--- a/journal-of-korean-neurosurgical-society.csl
+++ b/journal-of-korean-neurosurgical-society.csl
@@ -15,7 +15,6 @@
     <category field="medicine"/>
     <issn>2005-3711</issn>
     <eissn>1598-7876</eissn>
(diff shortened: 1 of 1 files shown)


journal-of-korean-neurosurgical-society.csl
```diff
@@ -15,7 +15,6 @@
     <category field="medicine"/>
     <issn>2005-3711</issn>
     <eissn>1598-7876</eissn>
-    <summary>A style for the Journal of Korean Neurosurgical Society based on ElesvierVancouver/Free Radical Research.</summary>
```
rmzelle: This wasn't right, right? (the template is "american-journal-of-translational-research")

rmzelle:
Thanks!

PR #2431, opened 2017-01-15
Create world-applied-sciences-journal.csl

damnation333:
Per https://forums.zotero.org/discussion/50981/i-need-your-urgent-help-style-request-world-applied-sciences-journal


rmzelle:
Thanks!

PR #2433, opened 2017-01-16
Create copeia.csl

libora:
Discussion: https://forums.zotero.org/discussion/comment/267833


rmzelle:
Thanks! (I deleted the "issued" macro, which was identical in content to the "issued-year" one)

PR #2434, opened 2017-01-16
Update Comparative Population Studies

bwiernik:
Remove reference to nonexistent term.

(Perhaps the validator could check for these?)

### DIFF
--- a/comparative-population-studies.csl
+++ b/comparative-population-studies.csl
@@ -170,7 +170,6 @@
           <text variable="title" text-case="title"/>
           <group delimiter=" ">
             <text variable="genre" text-case="title"/>
-            <text term="number" form="short"/>
             <number variable="number"/>
           </group>
         </group>


rmzelle:
> Remove reference to nonexistent term.
> 
> (Perhaps the validator could check for these?)

Well, it's actually a valid term according to the schema (see https://github.com/citation-style-language/schema/blob/v1.0.1/csl-terms.rnc#L130), but it looks like we may have missed listing this type of terms in the specification.

rmzelle:
(I also think the CSL locale files contain a subset of all allowed terms)

bwiernik:
Hmm, when I try to insert a report with the style as first submitted in Word, I get a "trmtext not found" error. Perhaps it is a bug with citeproc-js then? In any event, the term line should be removed (there should be no label for number; compare the parallel code on line 196).

rmzelle:
> Perhaps it is a bug with citeproc-js then?

Well, it looks more like a helpful warning that you're trying to use a term for which no translations are available in either the locale files or the style itself.

bwiernik:
"Helpful warning" -- it stops the document from updating at all and prevents the insertion of the citation.

rmzelle:
If it's that bad, it might be worthwhile reporting the issue to @fbennett. Frank, citeproc-js apparently really doesn't like cases where it can't find a translation for a term. See above. Could it be changed to fail more gently?

fbennett:
If the style is missing a term, it does seem like it would be better for the error to be called to the user's attention somehow. The processor can throw a more meaningful error (I'll set that up), but making that easily discoverable would be a job for the calling application. About all the processor could do to surface it would be to throw an ugly citation, which could have bad consequences for authors.

Is it possible to catch these in validation?

rmzelle:
> Is it possible to catch these in validation?

It's tricky, since we currently don't require that each term (or form) is defined in the locale files. I would just default to "" for undefined terms.

bwiernik:
Could it default to NULL (or similar) so that affixes and group delimiters are not printed around the "" ?

rmzelle:
Yeah, that probably would be better.

(I'm not totally sure why this bug occurs, though, since we previously already deleted the "author" terms, and we never hear any problems about those: https://github.com/citation-style-language/locales/commit/3b6a1b5c8ce8c94172e67270be7a64c88aefa03d)

fbennett:
If the absent term returns empty, affixes and delimiters should go away. So
failing silently without error would be okay?

rmzelle:
I think that would be preferable, yes. @adam3smith?

adam3smith:
yes, I'd rather have this failing silently

fbennett:
I've fixed it for cs:text w/term attribute, and pushed new versions of the Propachi plugins. It think cs:label is already tolerant, but if there a more snags, give a shout. @rmzelle @adam3smith @bwiernik

rmzelle:
@bwiernik, would you mind testing this with the original style?

rmzelle:
Also, thanks Frank!

bwiernik:
cs:text with term="number" still fails with the same error. cs:label with term="number" returns no error.

fbennett:
Ouch, teething pains. I did test that, something must have gone awry. Will
go through it more carefully tomorrow.

fbennett:
I've gotten it to load without error in Zotero 5.0 beta standalone, and the processor replacement seems to be happening. I'll smooth out the install code and do some more testing tomorrow. Looks like we're close though.

fbennett:
Ookay, I think it may be better now. There was a problem with acquiring Zotero from context at startup. Pretty simple, once I figured it out. When the plugin is present at startup, we tie the processor replacement to a final-ui-startup event, which works nicely to assure Zotero is available. When the plugin is installed in the running app, that event has already flown, so we need to install immediately.

The problem was immediate acquisition crashes everything at startup (ouch); and when I staged back to using the event, the new processor was never installed (hence the error on the term refused to go away). I thinik I've hit some simple incantations to handle both cases -- should be able to enable and disable the plugin and observe the results (remembering that you need to selected another style and select back to observe the change).

[Fingers crossed.](https://juris-m.github.io/downloads/#csl-stuff)

bwiernik:
Seems to be working now. 

But small citeproc-js bug, I think: Words in "quotes" are not being capitalized when set using Title Case. Example JSON:
```[
	{
		"id": "http://zotero.org/users/1386342/items/C8XP258M",
		"type": "article-journal",
		"title": "From \"distance\" to \"friction\": substituting metaphors and redirecting intercultural research",
		"container-title": "Academy of Management Review",
		"page": "905-923",
		"volume": "33",
		"issue": "4",
		"DOI": "10.5465/AMR.2008.34421999",
		"shortTitle": "From \"distance\" to \"friction\"",
		"journalAbbreviation": "AMR",
		"language": "en",
		"author": [
			{
				"family": "Shenkar",
				"given": "Oded"
			},
			{
				"family": "Luo",
				"given": "Yadong"
			},
			{
				"family": "Yeheskel",
				"given": "Orly"
			}
		],
		"issued": {
			"date-parts": [
				[
					"2008"
				]
			]
		}
	}
]
```

fbennett:
That's been a long-standing issue. I'll take a look very soon.

PR #2435, opened 2017-01-16
Create scandinavian-journal-of-medicine-and-science-in-sports.csl

damnation333:
Per https://forums.zotero.org/discussion/63208/style-request-scandinavian-journal-of-medicine-science-in-sports


rmzelle:
Thanks!

PR #1683, opened 2015-09-12
Create Current Gene Therapy (Vancouver)

glycoSM:

PR #2436, opened 2017-01-17
Create current-gene-therapy.csl

damnation333:
To continue: https://github.com/citation-style-language/styles/pull/1683
Adapted correctly for book, chapter, journal, web and patent.


rmzelle:
Thanks!

PR #2440, opened 2017-01-18
Update fachhochschule-vorarlberg.csl

damnation333:
to continue #1933 (and #1953)

Slight cleanup (locales, use of variables than text and fix quotation mark issue and made a few simple changes to match guidelines.


fachhochschule-vorarlberg-author-date.csl
```diff
@@ -1,33 +1,63 @@
...
   <info>
-    <title>Fachhochschule Vorarlberg (German)</title>
-    <title-short>FHV</title-short>
-    <id>http://www.zotero.org/styles/fachhochschule-vorarlberg</id>
-    <link href="http://www.zotero.org/styles/fachhochschule-vorarlberg" rel="self"/>
-    <link href="http://www.fhv.at/bibliothek/schreibzentrum/unterlagen-zum-wissenschaftlichen-arbeiten" rel="documentation"/>
+    <title>Fachhochschule Vorarlberg (author-date, German)</title>
+    <title-short>FHV Autor-Jahr</title-short>
+    <id>http://www.zotero.org/styles/fachhochschule-vorarlberg-author-date</id>
+    <link href="http://www.zotero.org/styles/fachhochschule-vorarlberg-author-date" rel="self"/>
```
rmzelle: @adam3smith, so you're okay with the ID/title change? I saw some (German) discussion at https://github.com/citation-style-language/styles/pull/1933 on this.

@damnation333, if we make this change, the old and new style ID should be included in https://github.com/citation-style-language/styles/blob/master/renamed-styles.json, so that smart programs (like Zotero) can automatically update users from the old to the new style.

fachhochschule-vorarlberg-author-date.csl
```diff
@@ -1,33 +1,63 @@
...
+    <terms>
+      <term name="et-al">et al.</term>
+      <term name="anonymous" form="short">n. a.</term>
+      <term name="no date" form="short">n. y.</term>
+      <term name="collection-editor" form="short">Ed.</term>
+      <term name="retrieved">retrieved</term>
+      <term name="composer" form="short">Comp.</term>
+      <term name="composer" form="long">Composer</term>
+    </terms>
+  </locale>
```
rmzelle: Since the style has a German "default-locale", this locale section will never be used. (@adam3smith, do you recall whether Mendeley allows the "default-locale" value to be overridden? If so, maybe that's why they added it)

fachhochschule-vorarlberg-author-date.csl
damnation333: @rmzelle 
I'm in contact with the user who was struggling with the Github last time and he also provided me with their footnote style now, so we'll need the author-date form.
I've made a pull request for the renamed-styles.json: https://github.com/citation-style-language/styles/pull/2444

PR #2441, opened 2017-01-18
Minor correction

ahahl:
Sorry for this pull request so soon after the previous one. I hadn't spotted the issue on former line 350 before I realised it created problems with Zotpress. Thanks


rmzelle:
No problem. Thanks!

PR #2444, opened 2017-01-19
Update renamed-styles.json

damnation333:
Per https://github.com/citation-style-language/styles/pull/2440


rmzelle:
(I added this change to #2440, so that things are kept more together)

PR #2446, opened 2017-01-19
Create california-agriculture.csl

damnation333:
Per https://forums.zotero.org/discussion/63984/style-request-california-agriculture


rmzelle:
Thanks, Obama!

PR #2450, opened 2017-01-20
Create strategic-organization.csl

damnation333:
https://forums.zotero.org/discussion/46260/style-request-strategic-organization


rmzelle:
Thanks!

PR #2453, opened 2017-01-23

damnation333:
Fix two errors: https://forums.zotero.org/discussion/64042/citation-error-veterinary-pathology#latest
Alphabetical bibliography and not via occurence
initialized names.


veterinary-pathology.csl
```diff
@@ -45,7 +45,7 @@
   </macro>
   <macro name="editor">
     <names variable="editor" suffix=".">
-      <name sort-separator=" " initialize-with="" name-as-sort-order="all" delimiter=", " delimiter-precedes-last="always"/>
+      <name sort-separator=" " initialize-with=" " name-as-sort-order="all" delimiter=", " delimiter-precedes-last="always"/>
```
rmzelle: @damnation333, I think both the "editor" and "author" macros need `initialize-with=""`. With `initialize-with=" "` you would get something like "Holman R T , Wiese H F , Smith A N " instead of "Holman RT, Wiese HF, Smith AN", no?

PR #2454, opened 2017-01-23
Delete harvard-university-of-west-london.csl

damnation333:
https://forums.zotero.org/discussion/64033/style-error-university-of-west-london#latest

Delete. Substituted with proper Cite It Right, Oscola and APA.


rmzelle:
Thanks!

PR #2455, opened 2017-01-23
Create science-advances.csl

damnation333:
Per https://forums.zotero.org/discussion/54448/style-request-science-advances


rmzelle:
Thanks!

PR #2456, opened 2017-01-24
Create journal-of-zoo-biology.csl

damnation333:
Per https://forums.zotero.org/discussion/63695/style-request-journal-of-zoo-biology#latest

This journal doesn't seem to have any ISSN yet.

### DIFF
--- a/journal-of-zoo-biology.csl
+++ b/journal-of-zoo-biology.csl
@@ -8,7 +8,7 @@
     <link href="http://www.zotero.org/styles/evolution" rel="template"/>
     <link href="http://www.escijournals.net/JZB/about/submissions#authorGuidelines" rel="documentation"/>
     <author>
-      <name>O'Brien Patrick, PhD</name>
+      <name>Patrick O'Brien, PhD</name>
       <email>obrienpat86@gmail.com</email>
     </author>
     <category citation-format="author-date"/>
@@ -129,10 +129,10 @@
         <text macro="author-short"/>
         <text macro="year-date"/>
       </group>
-        <group prefix=", " delimiter=" ">
-          <label variable="locator" form="short"/>
-          <text variable="locator"/>
-        </group>
+      <group prefix=", " delimiter=" ">
+        <label variable="locator" form="short"/>
+        <text variable="locator"/>
+      </group>
     </layout>
   </citation>
   <bibliography hanging-indent="true" entry-spacing="0" line-spacing="2">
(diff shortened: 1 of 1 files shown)


journal-of-zoo-biology.csl
```diff
@@ -0,0 +1,154 @@
...
+      </substitute>
+    </names>
+  </macro>
+  <macro name="author-short">
+    <names variable="author">
+      <name form="short" and="text" initialize-with=". "/>
+      <substitute>
+        <names variable="editor"/>
+        <names variable="translator"/>
+        <text macro="title"/>
```
adam3smith: this doesn't seem like a good choice -- I'd add a separate title-short macro, using the short title and, more importantly, doesn't include the edition information. Then also add it to the title macro.

journal-of-zoo-biology.csl
```diff
@@ -0,0 +1,154 @@
...
+      </substitute>
+    </names>
+  </macro>
+  <macro name="title">
+    <choose>
+      <if type="book report" match="any">
+        <group delimiter=" ">
```
adam3smith: this group doesn't do anything

journal-of-zoo-biology.csl
```diff
@@ -0,0 +1,154 @@
...
+  <macro name="year-date">
+    <choose>
+      <if variable="issued">
+        <date variable="issued">
+          <date-part name="year"/>
+        </date>
+        <text variable="year-suffix"/>
+      </if>
+      <else>
+        <text term="no date" form="short"/>
```
adam3smith: you'll want to add the year-suffix explicitly for no date, too, otherwise you can get ambiguous citations where  two items have the same author and no date.

journal-of-zoo-biology.csl
```diff
@@ -0,0 +1,154 @@
...
+      <text variable="publisher"/>
+      <text variable="publisher-place"/>
+    </group>
+  </macro>
+  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" collapse="year-suffix" year-suffix-delimiter=",">
+    <sort>
+      <key macro="year-date"/>
+    </sort>
+    <layout prefix="(" suffix=")" delimiter="; ">
+      <group delimiter=" ">
```
adam3smith: I know the style guide is bad, but (Smith 1776 pp. 3-4) seems unusual. It seems like there'd be some type of delimiter, probably a comma, before the page label&number.

journal-of-zoo-biology.csl
damnation333: Can you explain what the year-suffix does? Can't find anything in the CSL specs and never came across it before. :)

adam3smith: the year-suffix explicitly sets the location of the year suffix set by `disambiguate-use-year-suffix`. I actually think in this case you can just delete it and this should work fine (by default it gets appended after the year)

rmzelle:
Thanks both!

PR #2458, opened 2017-01-25

rmzelle:
Per https://forums.zotero.org/discussion/63923/make-a-non-listed-csl-1-0-1-valid

### DIFF
--- a/universite-du-quebec-a-montreal.csl
+++ b/universite-du-quebec-a-montreal.csl
@@ -1,487 +1,538 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="fr-CA">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" default-locale="fr-CA">
   <info>
-    <title>Université du Québec à Montréal (Bouthat) (French - Canada)</title>
-    <title-short>UQAM</title-short>
-    <id>http://www.zotero.org/styles/universite-du-quebec-a-montreal</id>
-    <link href="http://www.zotero.org/styles/universite-du-quebec-a-montreal" rel="self"/>
-    <link href="http://virtuolien.uqam.ca/tout/UQAM_BIB000435621" rel="documentation"/>
+    <title>Style UQAM test</title>
+    <title-short>Style UQAM</title-short>
+    <id>http://www.zotero.org/styles/style-UQAM-test</id>
+    <link href="http://www.zotero.org/styles/style-UQAM-test" rel="self"/>
+    <link href="http://www.zotero.org/styles/apa-fr" rel="template"/>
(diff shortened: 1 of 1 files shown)


universite-du-quebec-a-montreal.csl
```diff
@@ -1,487 +1,542 @@
...
-                <name and="text" delimiter=", "/>
-              </names>
-            </if>
-          </choose>
-        </group>
+      <if type="chapter paper-conference entry-encyclopedia entry-dictionary" match="any">
+        <names variable="editor container-author" delimiter=", " suffix=", ">
+          <name and="text" initialize-with=". " delimiter=", "/>
+          <label form="short" prefix=" (" text-case="title" suffix=")"/>
+          <label form="short" prefix=" (" suffix=")" text-case="lowercase"/>
```
rmzelle: These double labels are weird. I'm rather surprised it even validates.

(csl.rnc uses:

```
    element cs:names {
      names.attributes,
      ((names.name?, names.et-al?) & names.label?),
      names.substitute?
    }
```

which I would think would only allow a single `<label/>`)

universite-du-quebec-a-montreal.csl
```diff
@@ -1,487 +1,542 @@
...
-  <macro name="anon">
+  <macro name="secondary-contributors">
     <choose>
-      <if variable="author editor translator" match="none">
-        <text term="anonymous" form="short" text-case="capitalize-first" suffix="." strip-periods="false"/>
+      <if type="article-journal book chapter paper-conference entry-encyclopedia" match="none">
+        <names variable="editor" delimiter=", " prefix=" (" suffix=")">
+          <name name-as-sort-order="all" and="text" sort-separator=", " initialize-with=". " delimiter=", " delimiter-precedes-last="never"/>
+          <label form="short" prefix=", " text-case="title"/>
+          <label form="short" prefix=" (" suffix=")" text-case="lowercase"/>
```
rmzelle: Idem.

universite-du-quebec-a-montreal.csl
rmzelle: Nevermind. I copied that schema fragment from "master", where we already fixed this post-1.0.1: https://github.com/citation-style-language/schema/commit/850a86fb56f10c198ef051159df07386d19dd56a

rmzelle: Should still fix the style, though.

rmzelle: @damnation333, do you have a time to check which of these two labels we should keep here and below?

universite-du-quebec-a-montreal.csl
adam3smith: sorry, I didn't see the ping. The lowercase labels in parentheses are correct both here and above

universite-du-quebec-a-montreal.csl
rmzelle: (@damnation333, never mind, @adam3smith already helped out: https://github.com/citation-style-language/styles/pull/2458#discussion_r99272512)

Issue #1829, opened 2015-12-13
Style for Glossa


zuphilip:
It would be nice to provide a CSL style for the transformed journal <a href="http://glossa.ubiquitypress.com/">Glossa</a>, which is now a fair Open Access journal. The guidelines can be found <a href="http://glossa.ubiquitypress.com/about/submissions/#style-sheet">here</a> and they refer to <a href="http://www.eva.mpg.de/linguistics/past-research-resources/resources/generic-style-rules.html">The Generic Style Rules for Linguistics (version of December 2014)</a> with some adaptions.

rmzelle:
Added in #2459.

zuphilip:
Thank you very much @damnation333 !

PR #2459, opened 2017-01-25
Create glossa.csl

damnation333:
per https://github.com/citation-style-language/styles/issues/1829
Journal, book, chapter and webpage all adapted to their regulations.


rmzelle:
Thanks!

PR #2460, opened 2017-01-25
Exclude book from archive-note macro.

adunning:
This prevents the place from occurring again, should `archive-place` be provided for an entry. I cannot think of any situation where this would be desirable. The Better CSL JSON plugin exports `archive-place` for all books alongside `publisher-place`, for some reason.

### DIFF
--- a/modern-humanities-research-association.csl
+++ b/modern-humanities-research-association.csl
@@ -406,7 +406,7 @@
   <macro name="access-note">
     <group delimiter=", ">
       <choose>
-        <if type="article-journal bill chapter legal_case legislation paper-conference" match="none">
+        <if type="article-journal bill book chapter legal_case legislation paper-conference" match="none">
           <text macro="archive-note" prefix=", "/>
         </if>
       </choose>


rmzelle:
> The Better CSL JSON plugin exports archive-place for all books alongside publisher-place, for some reason.

With the same value for "archive-place" and "publisher-place"? And with "Better CSL JSON plugin", you mean the "Better CSL JSON" export of "Better Bib(La)TeX for Zotero/Juris-M" (https://github.com/retorquere/zotero-better-bibtex)?

adam3smith:
I'm not too keen on this. While I admit that it's going to be rare, I don't think it's inconceivable that a (perhaps sufficiently rare) book-type publication would be found in an archive, in which case the archive location would be needed in the location. The fact that any reference manager or plugin puts data into fields that clearly don't belong there (archive-location and publisher-place are two completely different things) shouldn't lead us to change citation styles.

rmzelle:
@adunning, is there any documentation on what's different about "Better CSL JSON"?

adunning:
Fixed in <https://github.com/retorquere/zotero-better-bibtex/issues/637>, so closing this. (I am still not absolutely convinced that the CSL file is useful as arranged, since conceivably some of the other excluded items could also be found in an archive, but it's a moot point.)

PR #2461, opened 2017-01-25
Create psychosomatic-medicine.csl

damnation333:
Per https://forums.zotero.org/discussion/46346/style-request-psychosomatic-medicine


rmzelle:
Thanks!

PR #2462, opened 2017-01-26
Update journal-of-animal-science.csl

hannahgr8:
Added in an option to include the DOI (if present), which is required in the most recent JAS author guidelines.


rmzelle:
Thanks!

PR #2463, opened 2017-01-27
Create vancouver-brackets-noURL.csl

djibe89:
Vancouver citation styles with brackets for citations and URL references only for webpages

### DIFF
--- a/vancouver-brackets-noURL.csl
+++ b/vancouver-brackets-noURL.csl
@@ -1,0 +1,354 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="minimal">
+  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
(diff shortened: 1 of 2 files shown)


djibe89:
Yes indeed.
In University of Lille 2 (France), academic professors don't like when there are http:// in the References.
But if you store URL to the article for future re-access to the article, you are in a dilemma. Store it or hide it from Bibliography.
My style allows to store URL of article, guideline and no Bibliography mess with http://.

djibe89:
Yes for Zotero.
I understand, but many of the people in my University find it very helpful.

rmzelle:
> In University of Lille 2 (France), academic professors don't like when there are http:// in the References.

Is this preference described in an official style guide from the university?

djibe89:
No. Just Vancouver.
But Professors always tell us to stop having http and using brackets to cite.
You can delete if you want.
I dfon't wanna spoil your time

adam3smith:
Thanks -- given that, I don't think we'll take this for the repository. However, as long as you don't change it in your github repository, your file will still be available, and you can pass on the link https://raw.githubusercontent.com/djibe89/styles/95c93325e3f7a2ea0b1f22186d12559cd438bf91/vancouver-brackets-nourl.csl to your colleagues.

We really do appreciate you submitting this, though.

PR #2464, opened 2017-01-27
Use serial comma for contributors in MHRA.

adunning:
See § 5.1 (c) of the Style Guide.

### DIFF
--- a/modern-humanities-research-association.csl
+++ b/modern-humanities-research-association.csl
@@ -68,7 +68,7 @@
   </macro>
   <macro name="contributors-note">
     <names variable="author">
-      <name and="text" sort-separator=", " delimiter=", " delimiter-precedes-last="never"/>
+      <name and="text" sort-separator=", " delimiter=", "/>
       <label form="short" prefix=", "/>
       <substitute>
         <text macro="title-note"/>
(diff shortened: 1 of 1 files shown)


rmzelle:
I couldn't find an example with three authors (e.g. on page 59-60 of http://www.mhra.org.uk/pdf/MHRA-Style-Guide-3rd-Edn.pdf), and I'm not 100% sure the § 5.1 (c) rule  (on page 33) applies to author lists in notes.

I did find an example in http://www.spurgeons.ac.uk/_uploads/downloads/Spurgeon's_Referencing_Guidelines_2013-2014.pdf, which claims to follow MHRA:

>Brian Haymes, Ruth M.B. Gouldbourne and Anthony R. 
Cross

which agrees with the current setting. @adunning, are you aware of any explicit examples by MHRA itself?

adunning:
There is an example on p. 87 (chap. 13) of the guide.

rmzelle:
> There is an example on p. 87 (chap. 13) of the guide.

But like the examples on page 80 in § 11.6, those examples are for bibliographic entries, right, not the in-text notes?

@adam3smith, I don't oppose accepting this, so feel free to merge, but I find the MHRA guide rather confusing on this, and can't tell whether this is really required.

adunning:
Yes, it's definitely confusing. Still, I would think that if they meant to have an exception to a rule stated on the third page of their guide, they would say so …

adam3smith:
There's also an example for a serial comma for a parenthetical citations with tree authors in the 3rd edition, so I think overall it's pretty clear they want the comma, though from a style guide perspective, not including a relevant example is indeed quite unfortunate.

adam3smith:
Thanks!

PR #2465, opened 2017-01-28
Create mind-and-language.csl

damnation333:
Per https://forums.zotero.org/discussion/64130/style-request-mind-language

### DIFF
--- a/mind-and-language.csl
+++ b/mind-and-language.csl
@@ -231,14 +231,14 @@
     <sort>
       <key macro="author"/>
     </sort>
-    <layout delimiter="; ">
+    <layout delimiter="; " prefix="(" suffix=")">
       <group delimiter=": ">
         <choose>
           <if match="any" position="ibid">
             <text term="ibid"/>
           </if>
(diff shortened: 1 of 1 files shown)


rmzelle:
@adam3smith, looks good to me. See http://onlinelibrary.wiley.com/doi/10.1111/mila.2016.31.issue-1/issuetoc for example papers.

bwiernik:
It looks like citation layout should have parentheses affixes. All of the in-text citations and most of the footnote citations have parentheses around them. Across different papers I've looked at for this journal, it seems fairly common to have citations in footnotes that do not have parentheses if the footnote is one clause of the form "For additional information, see Strause 2003; Engel, 2006". But citations with parentheses are so much more common that it would be a huge pain for authors, I suspect, to have to type them themselves.

rmzelle:
@damnation333, @bwiernik was clearly more eagle-eyed than me. Can you make his proposed change?

rmzelle:
Thanks both!

PR #2468, opened 2017-01-29
Update juristische-zitierweise.csl

zuphilip:
* Fix citations for journal articles, i.e. add first page, no label, locator in parenthesis
* Besides the normal citations to journals, there are some special "archive" journals which are cited differently, i.e. they use volume and the year is in parenthesis --> I covered that in the type of `magazineArticle` which then just has to manually switched to
* Add `genre` to legal cases which will add "Urt."/"Urteil" or "Beschl."/"Beschluss" or similar types
* Use `genre` in thesis instead of fixed term
* Handle webpage type in bibliography
* Improve spacing and delimiters in bibliography + some other cleanup work

All this should be in accordance to the Stüber's Richtlinien.

As a second step I would like then to look more closely to the commentaries (which are always a little hard to understand and deal with).

### DIFF
--- a/juristische-zitierweise.csl
+++ b/juristische-zitierweise.csl
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" class="note" default-locale="de-DE">
+  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
     <title>Juristische Zitierweise (Stüber) (German)</title>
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks!

zuphilip:
Thank you! Wow, that was fast.

I created the `locale` part also with the visual editor, i.e. if the position is wrong, then we should correct that in the csl-editor code.

rmzelle:
> I created the locale part also with the visual editor, i.e. if the position is wrong, then we should correct that in the csl-editor code.

Right. We already have a ticket (by you!) at https://github.com/citation-style-language/csl-editor/issues/193, though.

PR #2470, opened 2017-01-29
Create arctic-antarctic-and-alpine-research.csl

damnation333:
Per https://forums.zotero.org/discussion/54880/style-request-arctic-antarctic-and-alpine-research

Adapted for journal, book, chapter.


rmzelle:
Thanks!

PR #2471, opened 2017-01-29
Create plant-cell-and-environment.csl

damnation333:
Per https://forums.zotero.org/discussion/48191/style-request-plant-cell-environment
Yet another Wiley Journal. They really need to get better at documenting their author guidelines. Everything in bold when they got a variable in bold...
Free paper here: http://onlinelibrary.wiley.com/doi/10.1111/pce.12141/pdf

### DIFF
--- a/plant-cell-and-environment.csl
+++ b/plant-cell-and-environment.csl
@@ -93,9 +93,7 @@
     </sort>
     <layout suffix=".">
       <text macro="author"/>
-      <date variable="issued" prefix=" (" suffix=")">
-        <date-part name="year"/>
-      </date>
+      <text macro="year-date" prefix=" (" suffix=")"/>
       <choose>
         <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
           <group delimiter=" " prefix=" ">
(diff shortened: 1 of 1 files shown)


rmzelle:
> Yet another Wiley Journal. They really need to get better at documenting their author guidelines.

Yes. Unlike most big publishers, they also don't seem to have made any attempts at standardizing their citation formats. I once reached out in 2013, but the person from "Author Customer Services" I was talking to didn't seem to understand my question on whether they (plan to) use a small set of standard citation formats for their journals. Instead, I got a rather nonsensical reply:

> We sincerely appreciate your kind offer of generating CSL styles for our journals. However, as mentioned previously, we currently handle our journals’ citation/reference styles ourselves. We will get in touch with you if we wish to learn more about your offer.

(@zuphilip, if you have any good contacts with Wiley, may we could reach out again)

adam3smith:
I tried twice via individual contacts to fairly high-up Wiley reps, but never got anywhere.

rmzelle:
@damnation333, can you review my changes to the et-al settings?

Your example paper from 2013 is rather old, by the way. This one is from the February 2017 issue: http://onlinelibrary.wiley.com/doi/10.1111/pce.12809/epdf. See e.g.:

> (Baldwin 2013; Wang et al. 2011; Zhang et al. 2013).

and

> Zhang C., Xie Q., Anderson R.G., Ng G., Seitz N.C., Peterson T., … Lu H. (2013)
Crosstalk between the circadian clock and innate immunity in Arabidopsis.
PLoS Pathogens 9, e1003370.

adam3smith:
(my changes are just clean-up; nothing substantive)

zuphilip:
@rmzelle I have no contact at Wiley, but will keep this in my mind.
( Maybe, they will make at some point some offers for buying subscription to journals etc. and then I can try to ask some things about their citation styles, because it is important for us to support our scientists as much as possible :-)

damnation333:
@rmzelle 
Thanks for spotting those issues. Looks fine from my side. :)

rmzelle:
Thanks all!

PR #2473, opened 2017-01-31
Create european-journal-of-ultrasound.csl

damnation333:
Per: https://forums.zotero.org/discussion/61439/style-request-ultraschall-in-der-medizin-european-journal-of-ultrasound

Adapted for journals, books, chapters.

Set the default locale to English, as it seems everything is in english anyway. Sometimes there is little bits of german in the references (Hrsg), but sometimes notw. Seems like they don't know what they want/don't really check.

Free Papers here (on top there is three free ones): http://ultraschall.thieme.de/home


rmzelle:
Thanks!

PR #2475, opened 2017-02-02
Make dependent style (CSE) cell-transplantation.csl

damnation333:
Per https://forums.zotero.org/discussion/64226/cell-transplantation-style


adam3smith:
(there seem to be some issues with this still -- see thread linked above. Let's not merge for the time being)

damnation333:
replaced by https://github.com/citation-style-language/styles/pull/2478
(was not just the full CSE style, but needed a few changes >> independent style created)

PR #2476, opened 2017-02-03
Create clinical-neurophysiology.csl

luciomarinelli:


damnation333:
@luciomarinelli 
Hi,
How come you've created this? This is already in the repository as an automatically generated dependent Elsevier style. Is the style incorrect?
(for the style you provided documentation and default-locale are missing if this was to be taken forward)

luciomarinelli:
Yes, the currently available style for Clinical Neurophysiology refers to the Vancouver style, however the journal requires some changes so I copied the Vancouver style and modified it accordingly...

rmzelle:
> the currently available style for Clinical Neurophysiology refers to the Vancouver style, however the journal requires some changes so I copied the Vancouver style and modified it accordingly...

Can you give an example of how the original style and your version differ?

Also, I noticed that http://www.clinph-journal.com/content/authorinfo#idp1519696 says "There are no strict requirements on reference formatting at submission. References can be in any style or format as long as the style is consistent.". Given that this journal (still) seems to be published by Elsevier, who have put in an effort to standardize their citation styles, I'm not sure it makes sense to switch to a customized style here.

luciomarinelli:
I just had a manuscript reviewed in Clinical Neurophysiology. The reference were formatted with Zotero using the provided style. This is what the editorial office wrote: "References: no issue numbers are needed; volume and page numbers are sufficient. Also, do not list the month of publication."
My modified style solved these issues and I believe it could be of use for other people...

rmzelle:
@luciomarinelli, okay, if the editors came back with remarks, we'll probably take this.

@adam3smith, can you take a quick look whether this is okay now? I'll add the journal to the Elsevier skip list when we merge this.

adam3smith:
Generally yes, I'm OK with that but I'd like to a) check which other styles depend on vancouver author-date: I'd guess that any other Elsevier styles would also be incorrect. If there are others, it'd make more sense to create an Elsevier Vancouver (author-date) style that's based on the current Elsevier Vancouver. If this style is unique, we can just handle it via skip list as you suggest. Will try asap

rmzelle:
@adam3smith, ah, yes. Good point.

adam3smith:
Yeah 27 of 59 dependents using the style are on the Elsevier list. I'll work on fixing this, but might be a couple of days to do it right & carefully.

adam3smith:
superseded by https://github.com/citation-style-language/styles/commit/5490afd991ab05d393d7166fabd67343b7003edb
@luciomarinelli thank you again -- we switched out all Elsevier journals following this template to one that should be exactly right, so while we're not taking your modified style, this did lead us to fix 30 others, so super helpful.

luciomarinelli:
Glad about it... next monday I will do another check about inline citations since a coauthor pointed me something weird; if needed I will do a pull

PR #2477, opened 2017-02-04
Create edward-elgar-business-and-social-science.csl

damnation333:
@bwiernik asked me to help with this style as he's writing a paper for this currently. Why not upload it to the repo.
The style is a little bit inconsistent in all the various publications, we tried get it work ok.
Adapted for journals, books, chapter and speeches (presentations).

### DIFF
--- a/edward-elgar-business-and-social-sciences.csl
+++ b/edward-elgar-business-and-social-sciences.csl
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" page-range-format="minimal" demote-non-dropping-particle="never" default-locale="en-GB">
   <info>
-    <title>Edward Elgar Publishing - Business &amp;AMP; Social Science</title>
-    <id>http://www.zotero.org/styles/edward-elgar-business-and-social-science</id>
-    <link href="http://www.zotero.org/styles/edward-elgar-business-and-social-science" rel="self"/>
+    <title>Edward Elgar Publishing - Business &amp; Social Sciences</title>
+    <id>http://www.zotero.org/styles/edward-elgar-business-and-social-sciences</id>
+    <link href="http://www.zotero.org/styles/edward-elgar-business-and-social-sciences" rel="self"/>
     <link href="http://www.zotero.org/styles/british-journal-of-industrial-relations" rel="template"/>
     <link href="https://www.e-elgar.com/sites/e-elgar/files/images/shared/authoredguidebussocsci20151.pdf" rel="documentation"/>
     <author>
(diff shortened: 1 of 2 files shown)


edward-elgar-business-and-social-sciences.csl
```diff
@@ -0,0 +1,319 @@
...
+    </author>
+    <category citation-format="author-date"/>
+    <category field="social_science"/>
+    <summary>Style for Edward Elgar Publishing Business and Social Sciences (non-textbook) publications.</summary>
+    <updated>2017-02-03T13:40:05+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="en">
+    <terms>
+      <term name="and others">""</term>
```
rmzelle: @damnation333, @bwiernik, is this intentional? It looks rather ... peculiar.

damnation333: Good spot. I needed to get rid of the et all and that's the way I found it worked. I know it looks weird. Haha. Any way to simplify this?

bwiernik: et-al should definitely not be dropped.

edward-elgar-business-and-social-sciences.csl
```diff
@@ -165,7 +165,7 @@
       </group>
     </layout>
   </citation>
-  <bibliography hanging-indent="true" subsequent-author-substitute="———" subsequent-author-substitute-rule="partial-each">
+  <bibliography hanging-indent="true" subsequent-author-substitute="&#8212;&#8212;&#8212;" subsequent-author-substitute-rule="partial-each">
```
bwiernik: subsequent-author-substitute should be dropped. The style doesn't use it.

bwiernik: subsequent-author-substitute should be dropped. The style doesn't use it.

edward-elgar-business-and-social-sciences.csl
rmzelle: @damnation333, see http://docs.citationstyles.org/en/stable/specification.html#et-al. The "and others" term is only used when set explicitly on `<et-al/>`.

(and, like @bwiernik says, some kind of "et al." term should always be used when abbreviating name lists, since otherwise there is no indication that names were left out)

edward-elgar-business-and-social-sciences.csl
```diff
@@ -0,0 +1,314 @@
...
+      <group delimiter=": ">
+        <group delimiter=" ">
+          <text macro="author-short"/>
+          <text macro="year-date"/>
+        </group>
+        <text variable="locator"/>
+      </group>
+    </layout>
+  </citation>
+  <bibliography hanging-indent="true" subsequent-author-substitute="———" subsequent-author-substitute-rule="partial-each">
```
bwiernik: The bibliography uses a lot of very fragile suffixes and prefixes. It results in lots of missing commas, periods in the middle of the reference, etc.

rmzelle:
Continued in #2489.

PR #2478, opened 2017-02-06
Update cell-transplantation.csl

damnation333:
Changed to CSE base with 2 minor changes and adaption for internet sources.
Per https://forums.zotero.org/discussion/64226/cell-transplantation-style


rmzelle:
Thanks!

PR #2479, opened 2017-02-06
Create high-altitude-medicine-and-biology.csl

damnation333:
Per https://forums.zotero.org/discussion/64286/style-request-high-altitude-medicine-biology

Adapted for journal, book and chapter. Waiting for requestor to provide example webpage citation in order to adapt properly.


rmzelle:
Thanks!

PR #2480, opened 2017-02-06
Update ASME style

karnesky:
Updated URL for style guidance, use title casing and italics where called for, add report number.


rmzelle:
Thanks!

PR #2482, opened 2017-02-07
Update and rename dependent/australian-mammalogy.csl to australian-ma…

damnation333:
…mmalogy.csl

Per https://forums.zotero.org/discussion/comment/269665/#Comment_269665

Made into an independent style based on Emu - austral ornith.


australian-mammalogy.csl
```diff
@@ -0,0 +1,227 @@
...
+      </else>
+    </choose>
+  </macro>
+  <macro name="pages">
+    <group>
+      <label variable="page" form="short" plural="always" suffix=". "/>
+      <text variable="page"/>
+    </group>
+  </macro>
+  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true">
```
damnation333: Made a few minor chances and also deleted disambiguate-add-names="true" from the bibliography line in order to get the 2009a, 2009b disambiguation that was asked for. Not sure if that is the correct fix though.

adam3smith:
sorry, this overlapped with my forum comment, but if this is really just disambiguation, I think we should just make the changes to the parent style and leave this dependent. It's highly unlikely we ever knew how disambiguation should go one way or the other, and it's very likely they do it the same way in all their journals.

damnation333:
Closed as replaced by https://github.com/citation-style-language/styles/pull/2483
Changes made to parent style instead.

PR #2483, opened 2017-02-07
Update emu-austral-ornithology.csl

damnation333:
Per https://forums.zotero.org/discussion/64299/style-correction-request-australian-mammalogy
To replace https://github.com/citation-style-language/styles/pull/2482


adam3smith:
thanks!

PR #2484, opened 2017-02-08
Update ruby from 2.3.1 to 2.4.0

rmzelle:
(also update Gemfile.lock)


rmzelle:
(just running this a few times to make sure we don't get large performance regressions; seems like things stay the same. Did notice something odd in the Travis build reports, though: https://github.com/thekompanee/fuubar/issues/97)

PR #2485, opened 2017-02-08
Create periodicum-biologorum.csl

damnation333:
Per https://forums.zotero.org/discussion/60853/style-request-periodicum-biologorum
Journal homepage: http://hrcak.srce.hr/ojs/index.php/periodicum_biologorum/index
Journal on NLM: https://www.ncbi.nlm.nih.gov/nlmcatalog/?term=0031-5362
current article with journals, chapters and webpage: http://hrcak.srce.hr/ojs/index.php/periodicum_biologorum/article/view/3928/2540


rmzelle:
Thanks!

PR #2487, opened 2017-02-08
Create waterbirds.csl

damnation333:
Per https://forums.zotero.org/discussion/62872/style-request-waterbirds
It says they follow CSE (I only saw after making the style). doesn't perfectly look like CSE anyway.
Free articles: http://www.bioone.org/toc/cowa/38/4

### DIFF
--- a/waterbirds.csl
+++ b/waterbirds.csl
@@ -1,0 +1,143 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB" xmlns="http://purl.org/net/xbiblio/csl">
+  <info>
+    <title>Waterbirds</title>
+    <id>http://www.zotero.org/styles/waterbirds</id>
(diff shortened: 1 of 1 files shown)


rmzelle:
(@damnation333, unrelated question: I assume you're currently using http://formatter.citationstyles.org/ before submitting your styles? If so, could you stop doing that for a while for styles you edit with the Visual Editor? I'm trying to see if there is anything else we can fix directly in the Visual Editor)

damnation333:
(@rmzelle
i use the formatter probably in 30% of the styles when I actually do lots of manual edits and copy paste kind of things and I can't be bothered to fix it manually. I won't use it for a while. :)

rmzelle:
(I'm mostly curious to what extent the Visual Editor mixes up the order of the "rel" and "href" attributes on links)

rmzelle:
(please keep using it for styles with lots of manual editing; I'm just want to keep an eye on styles coming straight out of the Visual Editor)

rmzelle:
Thanks!

rmzelle:
:duck:

PR #2488, opened 2017-02-08
Create oceanography.csl

damnation333:
per https://forums.zotero.org/discussion/48673/style-request-oceanography-magazine
Adapted fro journal, book and chapter.

Free articles here: http://tos.org/oceanography/issue


rmzelle:
Thanks!

PR #2489, opened 2017-02-09
Create edward-elgar-business-and-social-science.csl

bwiernik:
Same style as #2477

I started with the style that @damnation333 wrote and cleaned it up a bunch and fixed a lot of little bugs. It works well for all of the non-legal item types now.


rmzelle:
Great, thanks. (they are a bit inconsistent in using "Science" and "Sciences", but the style guide uses the latter, so let's settle on that)

PR #2490, opened 2017-02-09
Create apa-fr-ecole-de-technologie-superieure.csl

fharnois:
This is a new style for École de technologie supérieure which consists of a french adapted version of the APA style.


rmzelle:
@adam3smith, any opinion on the best title and ID for this? Slightly concerned we might get a lot of APA variants eventually. We could also adopt the scheme used for https://github.com/citation-style-language/styles/blob/master/university-college-lillebaelt-apa.csl ("University College Lillebælt - APA (Danish)"), which would place the style under the university name. It would be inconsistent for the other two French APA variants we currently have, though. See also https://www.zotero.org/styles?q=apa&dependent=0

(I also regret the "Harvard - [university name]" scheme we have, and would rather switch them all to 
"[university name] (Harvard)")

rmzelle:
(@fharnois, looks good; I'm just discussing something with my fellow maintainer here on how to best and consistently name this type of styles)

adam3smith:
@rmzelle -- I like university first and I think we should then also do that for the Montreal version.

@fharnois -- and having this, a 3rd French APA version, is necessary? Neither of the two that we already have works? I see the guide cites the Provost manual. That wouldn't work here?

rmzelle:
> @fharnois -- and having this, a 3rd French APA version, is necessary? Neither of the two that we already have works? I see the guide cites the Provost manual. That wouldn't work here?

Ah, good point. @fharnois, did you check if either one of the two existing French Canadian APA variants already has the correct format (see https://www.zotero.org/styles?q=apa-fr)

fharnois:
Hi,
Yes, I checked but both styles availables have some differences with the one we use. apa-fr-udem doesn't use ampersands and lacks some document types we use. Provost is closer to ours but isn't exactly the same either since we use [ ] to define some document types and doesn't have a reference style for patents.

adam3smith:
adding patents to Provost and using [] for document types (as APA style does) would seem to make a lot of sense as long as it doesn't go against the (Provost) style guide. I'd much rather see existing styles improved than new ones (that may or may not get maintained in the future) added.

fharnois:
That is a good idea. Could you tell me who is responsible for the updating of the provost style?

adam3smith:
It was created by @f-mb, but anyone can suggest additions or modifications, and we will accept them if they conform with the APA style as outlined in the Provost book.

f-mb:
Provost was created by someone else. We simply updated the title because it was not the official APA french style (there is none) but the adaptation by Provost. There is also APA French Université de Montréal.

fharnois:
Thank you for all this info. Unfortunately this is not an option for us since students won't know they can use the Provost style instead of the ÉTS style we recommend in the guide and for EndNote. I guess we will post the style directly on our website instead.

I will still try to update the Provost style with the improvements mentioned above.

Thank you and have a nice day!

adam3smith:
I'd consider making the ETS style a dependent style pointing to Provost -- what do you think @rmzelle ? I just really don't want to have duplicate code.

fharnois:
Would this make it possible for students to search with ETS? IF that is the case, it is a great solution for us.

BTW, I created a pull request for the updates made to APA-Provost a few minutes ago.

rmzelle:
> I'd consider making the ETS style a dependent style pointing to Provost

If that works, yes, that would be the preferred option.

> Would this make it possible for students to search with ETS?

Yes. Our "dependent" CSL styles are just skeleton styles that depend on another style for defining an actual citation format, but they show up under their own title.

fharnois:
That sounds great. If you can integrate the changes made in APA-Provost, it will match the style we use right now wihile also respecting the guidelines of Provost.

rmzelle:
We'll take a look soon. As for the dependent for your university, would `<title>École de technologie supérieure - APA (French - Canada)</title>` be acceptable? That would be consistent with what we did for https://www.zotero.org/styles?q=id%3Auniversity-college-lillebaelt-apa. (we'll probably rename the `American Psychological Association 6th edition (Université de Montréal) (French - Canada)` style as well to  `Université de Montréal - APA (French - Canada)` )

fharnois:
That name is perfect and would be easy for students to find.

rmzelle:
Thanks! I turned it into a dependent style and we accepted the proposed changes to the Provost style.

rmzelle:
I renamed the Université de Montréal APA style as well: https://github.com/citation-style-language/styles/commit/04d2e46bacd6c5ad80a3d18f1fec86e3bba9288f

PR #2351, opened 2016-12-12
Update south-african-journal-of-enology-and-viticulture.csl

Yusufras:
Page prefix=(  )
and
et al min in text.

PR #2492, opened 2017-02-10
Update south-african-journal-of-enology-and-viticulture.csl

damnation333:
To continue https://github.com/citation-style-language/styles/pull/2351

Style properly adapted for journals, books, chapters. Style guide and actual publications match! This style was off from it.
author-short et-al stuff is just standard, et-al-min=3, use-first=1.

Style guidelines: http://www.journals.ac.za/index.php/sajev/about/submissions#authorGuidelines

Free publications all here: http://www.sawislibrary.co.za/dbtw-wpd/textbase/sajev.htm

(Rintze: This was all edited in the CSL editor, mostly via the Visual Editor)

### DIFF
--- a/south-african-journal-of-enology-and-viticulture.csl
+++ b/south-african-journal-of-enology-and-viticulture.csl
@@ -2,6 +2,7 @@
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
   <info>
     <title>South African Journal of Enology and Viticulture</title>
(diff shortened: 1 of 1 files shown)


adam3smith:
@rmzelle good by me. Leaving this for you bc I'm not sure what you want to check for with the visual editor.

rmzelle:
> Leaving this for you bc I'm not sure what you want to check for with the visual editor.

I'm looking for any discrepancies in the editor output vs. what the formatter and/or the Python script produce. If there any some, I rather fix the editor. (in this case, there weren't any differences with the formatter)

rmzelle:
Thanks!

PR #2494, opened 2017-02-10
Update Journal of Sleep Research

mborg031:
The bibliography has been changed to truncate the reference list if there are more than 6 authors to the first 3 authors with "et al.". 

This change is required as per the Journal of Sleep Research author guidelines (http://onlinelibrary.wiley.com/journal/10.1111/(ISSN)1365-2869/homepage/ForAuthors.html).


adam3smith:
Thanks! (note that "more than 6 authors" translates into `et-al-min="6"` -- I fixed that in the style before accepting it)

PR #2497, opened 2017-02-14
Update collection-du-centre-jean-berard.csl

NMonteix:
Shift from in-text to note, added container author as possible substitute


rmzelle:
Thanks!

PR #2506, opened 2017-02-15
Update to APA provost

fharnois:
Update to correct some inconsistencies regarding the use of volume field, to add some document types (e.g. Patent) and to add genre between [ ] for some documents. All those changes were made in respect with the Provost Guidelines in « Normes de présentation d'un travail de recherche » 4th edition.

### DIFF
--- a/apa-fr-provost.csl
+++ b/apa-fr-provost.csl
@@ -174,25 +174,15 @@
       <else-if type="report">
         <text variable="title" font-style="italic"/>
         <group prefix=" (" suffix=")" delimiter=" ">
-          <choose>
-            <if variable="number" match="any">
-              <text value="Rapport" suffix=" "/>
-              <text variable="number" prefix="No. "/>
-            </if>
-          </choose>
+          <text value="Rapport"/>
+          <text variable="number" prefix="No. "/>
         </group>
       </else-if>
       <else-if type="patent" match="any">
-        <choose>
-          <if variable="number" match="any">
-            <group>
-              <text value="Brevet" font-style="italic" suffix=" "/>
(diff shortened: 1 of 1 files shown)


apa-fr-provost.csl
```diff
@@ -174,36 +174,60 @@
...
         <group prefix=" (" suffix=")" delimiter=" ">
-          <text variable="genre"/>
-          <text variable="number" prefix="No. "/>
+          <choose>
+            <if variable="number" match="any">
+              <text value="Rapport" suffix=" "/>
+              <text variable="number" prefix="No. "/>
+            </if>
+          </choose>
         </group>
```
rmzelle: @fharnois, you should be able to change

```xml
        <group prefix=" (" suffix=")" delimiter=" ">
          <choose>
            <if variable="number" match="any">
              <text value="Rapport" suffix=" "/>
              <text variable="number" prefix="No. "/>
            </if>
          </choose>
        </group>
```

to the simpler

```xml
        <group prefix=" (" suffix=")" delimiter=" ">
          <text value="Rapport"/>
          <text variable="number" prefix="No. "/>
        </group>
```

(groups automatically suppress text values if all called variables are empty; see http://docs.citationstyles.org/en/stable/specification.html#group)

rmzelle:
@adam3smith, do you have time to review this? I'll prep the dependent style.

adam3smith:
@rmzelle cleaned this up some, gtg from my side now

rmzelle:
Thanks!

rmzelle:
(This PR was created in response to #2490)

PR #2508, opened 2017-02-16
Create european-journal-of-sustainable-development.csl

libora:
Style based on APA; it can be used also for papers on International Conference on Sustainable Development


rmzelle:
Thanks!

Issue #2509, opened 2017-02-16
Author disambiguation in taxon.csl


damnation333:
I've been emailed regarding this style and can't figure out how to get it right:
**problem**: Following up on this: I've been using the style in Mendeley and notice that in the in-text citations 1) there appears to be some disambiguation of author surnames going on, which we don't do (some are appearing with initials or full names rather than just surname; I have a suspicion that it is somehow not dealing well with minor variants of the same name). 2) Some (but not all) citations include multiple author names (our usual format for 3+ authors is of course "1st author & al."). Perhaps this has something to do with the vagaries of my own database, but I haven't seen this behaviour in the few other styles that I have used so far. Do you think a fix will be on the cards?

Guidelines:
If several articles by the same author[s] in the same year are
cited, add letters to the year (e.g., 2006a). This applies also to articles
by the same first author but two or more various co-authors that
have to be cited with “& al.” in the text.

Relevant code in https://github.com/citation-style-language/styles/blob/master/taxon.csl 
citation et-al-min="6" et-al-use-first="1" et-al-subsequent-min="3" et-al-subsequent-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year" givenname-disambiguation-rule="primary-name"

Thanks.

adam3smith:
1) is indeed an issue of inconsistent name entry in Mendeley. I'd assume adding the initial for first authors is otherwise correct
2) we'd want to delete disambiguate-add-names="true"

damnation333:
thanks! Now I got it.
I also had to change the et-al-min settings. The example citation I was using had 5 authors only... Silly me.
PR made for changes: https://github.com/citation-style-language/styles/pull/2510

PR #2510, opened 2017-02-16
Update taxon.csl to fix author disambiguation

damnation333:
see https://github.com/citation-style-language/styles/issues/2509
Fix author disambiguation

### DIFF
--- a/taxon.csl
+++ b/taxon.csl
@@ -572,7 +572,7 @@
       </group>
     </layout>
   </citation>
-  <bibliography delimiter-precedes-et-al="never" et-al-min="8" et-al-use-first="6" entry-spacing="0" line-spacing="2" hanging-indent="true">
+  <bibliography delimiter-precedes-et-al="never" entry-spacing="0" line-spacing="2" hanging-indent="true">
     <sort>
       <key macro="author"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
did you check over this otherwise? -- I just had a quick glance at a recent article (there's a free one in the Dec. 2016 issue) and it looks nothing like this style.

damnation333:
Nothing like it is a bit is maybe a bit harsh, but I've found some mistakes just now. (open paper: http://www.ingentaconnect.com/contentone/iapt/tax/2016/00000065/00000005/art00007)
I made this style not too long ago and I was contacted by the editor people, so they must be generally happy with it apart from the disambiguation stuff. (I did spot some issues though with title italication).

adam3smith:
no, you're right. I actually overlooked some things in the code, it looks much better than I thought it did -- I just saw the code taken from APA and the style being quite un-APAish, but see a lot of the key things are actually right. The et al. for the bibliography isn't, though. I don't think they do et al. there at all. See the Wang et al. article with 11 authors in the bibliography of the article you link to.

damnation333:
Thanks for pointing that out. I've fixed that!

rmzelle:
Thanks!

PR #2511, opened 2017-02-16
Create advanced-optical-materials.csl

mfschumann:
Add style template for Wiley's Advanced Optical Materials


rmzelle:
Thanks! (happy to another fan of Commander Keen as well!)

Issue #2512, opened 2017-02-16
Improve style naming


rmzelle:
As followup of https://github.com/citation-style-language/styles/pull/2490#issuecomment-280091164 and https://github.com/citation-style-language/styles/commit/04d2e46bacd6c5ad80a3d18f1fec86e3bba9288f#commitcomment-20913664, I'd like to propose some changes:

* I'd like to change titles like "Elsevier Vancouver" to "Elsevier - Vancouver". This would extend to all (mostly publisher-wide) styles that have "Vancouver" and "Harvard" variants. Since this generally only involves adding a hyphen (and a space), this wouldn't affect IDs.
* I'd like to change over all the "Harvard - [university name]" styles to "University name - Harvard". I'd like to update the IDs to reflect this change as well. We might get a few complaints here during the switch, but I think it's still a better solution long-term. (maybe we should ask Mendeley if they support style redirection via renamed-styles.json, since I think they're rather popular in UK universities, which have most of these styles)

(we could switch to "University name (Harvard)" as well, which we already use for a few styles, but maybe we should make an distinction between labels like "Harvard", "Vancouver", and "APA", and more basic labels like "numeric" and "author-date" (I think we should keep placing the latter between parentheses)).

rmzelle:
@adam3smith, let me know if you agree with this. We could do the Harvard styles in two stages (titles first, then style IDs), and wait to see what the response is after each change.

adam3smith:
I'm fine with the renaming patterns and agree keeping Harvard/Vancouver etc. out of parentheses.
My concern apart from whether Mendeley et al handle the renaming gracefully would be universities who either link directly to styles or list them with their current name. I have no sense if that's common at all, so no idea if this is a non-issue or a major issue, though.

rmzelle:
Also, if you agree, I'd like to change the title of "Taylor & Francis - American Psychological Association" to "Taylor & Francis - APA". More consistent with the other styles:

https://www.zotero.org/styles?q=-%20american
https://www.zotero.org/styles?q=-%20apa

bwiernik:
(I have never heard any psychologist refer to the style by its full name.)

adam3smith:
definitely OK with the T&F rename

rmzelle:
> definitely OK with the T&F rename

[...]

PR #2514, opened 2017-02-19
Move "Harvard"/"Oxford"/"Vancouver"

rmzelle:
Follow-up of #2512.

### DIFF
--- a/harvard-anglia-ruskin-university.csl
+++ b/harvard-anglia-ruskin-university.csl
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
   <info>
-    <title>Harvard - Anglia Ruskin University</title>
+    <title>Anglia Ruskin University - Harvard</title>
     <id>http://www.zotero.org/styles/harvard-anglia-ruskin-university</id>
     <link href="http://www.zotero.org/styles/harvard-anglia-ruskin-university" rel="self"/>
     <link href="http://libweb.anglia.ac.uk/referencing/harvard.htm" rel="documentation"/>
--- a/harvard-bournemouth-university.csl
+++ b/harvard-bournemouth-university.csl
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
   <info>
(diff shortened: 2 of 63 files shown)


rmzelle:
@adam3smith, for you to merge. Just doing the titles for now should make it easy to roll things back, and minimize disruption for programs that don't use renamed-styles.json, or documentation that linked directly to styles.

rmzelle:
There is a drawback of this, by the way, in that the focus of people who are specifically looking for a "Harvard" style will be drawn more to the remaining styles that will start with "Harvard":

Harvard Educational Review
Harvard Reference format 1 (author-date)
Harvard Reference format 7 (author-date, German)

I still think we should make the change, but it increases the importance of getting rid of the "format 1" and "format 7" titles.

rmzelle:
> it increases the importance of getting rid of the "format 1" (...) and "format 7" (...) titles.

@adam3smith, in the xbiblio-devel discussion it seemed like the consensus was to turn the "format 1" style into a dependent, as a prelude to its eventual removal, but that hasn't happened yet. I think that would be a good change, but we might also want to change the title to indicate its deprecated status, e.g. to "(retired) Harvard Reference format 1 (replaced by "Cite Them Right 9th edition - Harvard")". Maybe we should then give Mendeley 6 months or so to prepare on their end, and remove it altogether after that point (and add it to renamed-styles.json).

Do you know if Mendeley nowadays comes with the Cite Them Right style in fresh installations? I have it in my copy of Mendeley, but I don't remember if I installed it myself at some point.

bwiernik:
> Do you know if Mendeley nowadays comes with the Cite Them Right style in fresh installations? I have it in my copy of Mendeley, but I don't remember if I installed it myself at some point.

Yes, the default styles are:
AMA
APSA
APA
ASA
Chicago 16 (note, full note, author-date)
Harvard - Cite them Right 9
Harvard Reference format 1
IEEE
MHRA
MLA
NLM
Nature
Vancouver

adam3smith:
Cool, thanks. As per one of the threads, I prepared a branch with Harvard1 turned into a dependent. There are currently 25 styles linking to it as a template. @rmzelle -- remind me what we do for those cases?

I also agree that we should be getting rid of it, but making it dependent on a decent style seems like a good idea anyway (and we'll need to resolve the template question either way).

rmzelle:
> There are currently 25 styles linking to it as a template. @rmzelle -- remind me what we do for those cases?

We don't really have to enforce the current requirement that all "template" links point to existing styles, but it helps catch typos, and usually doesn't give problems. In almost all cases, I just change the "template" link to the same style you redirect the old style ID to in renamed-styles.json. Just don't think of it as purely an pointer to the originating style. It can also be a close relative that is of higher or equally high quality as the style containing the "template" link.

PR #2515, opened 2017-02-19
Create nucleic-acids-research-web-server-issue.csl

sjdv1982:
Adapted NAR style (from nucleic-acids-research.csl") to fulfill the Web Server Issue requirement
from https://academic.oup.com/nar/pages/submission_webserver:
"The references section must include active electronic DOI and PubMed Central links for each cited paper (where available). Please include a PubMed abstract link if the PMC link is not available"


rmzelle:
@sjdv1982, before we deal with the ISSN issue, do you know of any example of what these Web Server issue references should look like? I checked the last issue (https://academic.oup.com/nar/issue/44/W1), and they didn't include the DOI/PMCID links there in the references, and I'm wondering whether it's really the best solution to add them on separate lines like you did where (using `display="block"`).

rmzelle:
Oh, actually, it looks like that's what they want. From the PDF in the zip-file from

> Programs to help insert the links into the references section are available at: http://lobstah.bu.edu/how_to_add_electronic_links_for_NAR.zip 

image

rmzelle:
@adam3smith, feel free to merge. Looks good to me now (I diffed the style to the original NAR, and the URLs were the only real changes).

adam3smith:
I want to double-check on the interaction between `display="block"` and `hanging-indent="true"`. I think this may either not work or create problems. In both of those cases we should remove the indent (if I'm wrong and this works, I want to know that, too).

rmzelle:
> I want to double-check on the interaction between display="block" and hanging-indent="true"

Yes, good point. I added a `display="block"` group at the top of the layout, but I didn't confirm in csledit.xul that the hanging indent actually still worked.

rmzelle:
@adam3smith, also, there has been some discussion over at pandoc about how to add support for the "display" and other positioning attributes, and whether and how they should work together. Any guidance we can give there would probably be most welcome. See https://github.com/jgm/pandoc-citeproc/issues/85

adam3smith:
@fbennett -- is there a reason you can't combine `second-field-align="flush"` (or `"margin"`) with `display="block"`? If you look at the example Rintze posts [above](https://github.com/citation-style-language/styles/pull/2515#issuecomment-280989599), that's exactly what we'd want here (rather than the hanging indent in the style), but the processor throws 
TypeError: blob.strings is undefined
With this style https://gist.github.com/adam3smith/2eb6403073d88d0e680bea890dfaaa19
(note that the display element isn't even set on the first group)

fbennett:
That end of things is kind of a mess, yeah. Without checking directly, it's probably caused by nesting of `display` elements inside the processor. IIRC the `second-field-align` attribute forces the "first" node into one `display` wrapper, and whatever remains of the cite into a second. So if you use `display` anywhere, it will be nested inside the implicit outer wrapper, and the behavior of nested `display` attributes is undefined.

`second-field-align` and `display` are very different approaches to the block-layout problem. In the next iteration of CSL, it might be a good idea to opt fully for one or the other.

adam3smith:
OK, so I'm going with the hanging indent -- it's not quite right, but better than without it and doesn't seem to cause any issues.

PR #2521, opened 2017-02-21
Create renewable-agriculture-and-food-systems.csl

damnation333:
via https://forums.zotero.org/discussion/62811/style-request-renewble-agriculture-and-food-systems
adapted for journal, book, chapter, webpage.


rmzelle:
Thanks!

PR #2522, opened 2017-02-21
Create journal-of-australian-strength-and-conditioning.csl

damnation333:
via https://forums.zotero.org/discussion/62601/style-request-journal-of-australian-strength-and-conditioning
Quite badly documented and also inconsistent. The requestor posted a few citations that are, I guess, from a paper, but in those they got the issue numer and a few different little things.


rmzelle:
Thanks!

PR #2525, opened 2017-02-23
FIX springer-basic for cases where is same lead author

sem-geologist:
Previous behavior was not right:
i.e. same author lets say Mikolaj, who published in 2014 and 2009, the similar thing.
citing it before resulted in:

(Mikolaj 2014; Mikolaj 2009) instead of collapsing it into (Mikolaj 2014, 2009).

Another noncompliant to springer format was the situation with same author and year:
lets say Mikolaj et al published in 2009  even 4 publications:

previous behaviour was like this:

(Mikolaj et al. 2009a; Mikolaj et al. 2009b; Mikolaj et al. 2009c; Mikolaj et al. 2009d)

correct behaviour would be (Mikolaj et al. 2009a, b, c, d)

I am 100% sure that this is how it should look in geoscience springer journals (i.e. International Journal of Earth Sciences); I looked also to some other scientific Springer journal requirements in other fields, it looks the same, thats why I want to fix that in the `springer-base`, as it is propagated to those styles (If I understand correctly).

### DIFF
--- a/springer-basic-author-date.csl
+++ b/springer-basic-author-date.csl
@@ -84,7 +84,7 @@
       <date-part name="year"/>
     </date>
   </macro>
-  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-givenname="true" disambiguate-add-year-suffix="true" collapse="year-suffix" cite-group-delimiter=", ">
+  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" collapse="year-suffix" cite-group-delimiter=", ">
     <sort>
       <key variable="issued"/>
       <key macro="author"/>
--- a/springer-basic-author-date.csl
+++ b/springer-basic-author-date.csl
@@ -84,7 +84,7 @@
       <date-part name="year"/>
     </date>
   </macro>
-  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true">
+  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-givenname="true" disambiguate-add-year-suffix="true" collapse="year-suffix" cite-group-delimiter=", ">
     <sort>
       <key variable="issued"/>
       <key macro="author"/>


rmzelle:
The linked "documentation" PDFs are silent on disambiguation, but I can confirm the collapsing format in two recent articles by journals using this style:  http://link.springer.com/article/10.1007/s10482-016-0759-5 (`(Golinska et al. 2013a, b)` and `(Kämpfer 2012a, b)`) and http://link.springer.com/article/10.1007/s11240-016-1111-5 (`(Pinto et al. 2002, 2008a, b, 2010)`).

@sem-geologist, do you have any examples to support the addition of `disambiguate-add-givenname="true"`?

sem-geologist:
my excuse, `disambiguate-add-givenname="true"` there got by error, it is not needed.

I think the springer documentation is silent there about the collapse as it is meant for books (and Latex is preferred, which would automatically take care). Now most of journals have its own Instructions for authors and in it they have more complicated example of citation ie.: [http://www.springer.com/earth+sciences+and+geography/geology/journal/531](http://www.springer.com/earth+sciences+and+geography/geology/journal/531) -> Instruction for Authors -> References

I guess it is due to changed focus of publishing, twenty years ago it was quite common to write monographs and books, automatically it meant that author would not be able to publish 2 things in the same subject in the year. Now it got quite a common situation as publishing articles is the primary activity.

adam3smith:
@rmzelle I haven't checked Springer journals; I know the guide is vague on this and the change makes sense to me, so feel free to merge.

sem-geologist:
I just wanted to be sure: Does it mean that dependant styles as International journal of earth sciences will get this correction?

rmzelle:
Great, thanks!

> I just wanted to be sure: Does it mean that dependent styles as International journal of earth sciences will get this correction?

Well, the dependent style won't change, but whatever software product you use should recognize that the parent style it relies on was updated and download the latest version.

PR #2526, opened 2017-02-23
Create the-canadian-geographer.csl

damnation333:
Guidelines here: http://onlinelibrary.wiley.com/journal/10.1111/(ISSN)1541-0064/homepage/TCG_Style_Guide_2015-10-19.pdf
Open Access paper: http://onlinelibrary.wiley.com/doi/10.1111/j.1541-0064.2012.00434.x/epdf

via: https://forums.zotero.org/discussion/46072/style-request-canadian-geographer


rmzelle:
Thanks!

(I added a second documentation link, because I don't like it when there is no easy way to go from the PDF to the journal homepage)

damnation333:
Thanks for the two fixes (docu and polyglot!) :)

PR #2528, opened 2017-02-24
Create limnetica.csl

damnation333:
via https://forums.zotero.org/discussion/57315/style-request-limnetica

Bit of variation going on regarding the issue. Space between volume and issue or not.

Papers here: http://www.limnetica.com/Limnetica/Limne35/


rmzelle:
Thanks!

PR #2532, opened 2017-02-26
Organic geochemistry

rmzelle:
Per https://forums.zotero.org/discussion/64525/style-error

@adam3smith, can you review? First commit is plain Elsevier Harvard with info block of former dependent style, second commit is the requested changes.


rmzelle:
(see http://www.sciencedirect.com/science/article/pii/S0146638016302686 for an example article)

rmzelle:
(and when we merge this we should put this style on the "elsevier" skip list)

adam3smith:
Looks good, thanks!

PR #2535, opened 2017-02-27

vl2:
- Minor changes in conference paper, journal article, book, book section, blog and report.

### DIFF
--- a/international-labour-organization.csl
+++ b/international-labour-organization.csl
@@ -16,16 +16,19 @@
     <author>
       <name>Vesa Sivunen</name>
     </author>
+    <author>
+      <name>Stephane Givkovic</name>
+    </author>
     <category citation-format="author-date"/>
     <category field="generic-base"/>
     <summary>The Harvard author-date style adapted for ILO house style</summary>
-    <updated>2013-01-21T15:37:19+00:00</updated>
+    <updated>2017-02-27T12:28:51+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="editor">
-    <names variable="editor" delimiter=", ">
+    <names variable="editor" delimiter=",">
       <name name-as-sort-order="all" and="text" sort-separator=", " initialize-with=". " delimiter=", "/>
-      <label form="short" prefix=" "/>
+      <label form="short"/>
     </names>
   </macro>
   <macro name="anon">
@@ -69,7 +72,7 @@
   </macro>
   <macro name="title">
     <choose>
-      <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
+      <if type="book" match="any">
         <text variable="title" font-style="italic"/>
       </if>
       <else>
@@ -78,9 +81,10 @@
     </choose>
   </macro>
   <macro name="publisher">
-    <group delimiter=", ">
+    <group>
       <text variable="event"/>
-      <text variable="publisher-place" prefix="(" suffix=")"/>
+      <text variable="publisher-place" prefix=" (" suffix=", "/>
+      <text variable="publisher" suffix=")"/>
     </group>
   </macro>
   <macro name="year-date">
@@ -109,15 +113,15 @@
     </choose>
   </macro>
   <macro name="pages">
-    <group>
-      <label variable="page" form="short" suffix=" "/>
+    <group prefix="p">
+      <label suffix=" " variable="page" form="short"/>
       <text variable="page"/>
     </group>
   </macro>
   <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true">
     <layout prefix="(" suffix=")" delimiter="; ">
       <group delimiter=" ">
-        <text macro="author-short"/>
+        <text variable="abstract"/>
         <text macro="year-date"/>
         <group>
           <label variable="locator" form="short"/>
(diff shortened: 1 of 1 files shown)


international-labour-organization.csl
```diff
@@ -109,15 +113,15 @@
...
-      <label variable="page" form="short" suffix=" "/>
+    <group prefix="p">
+      <label suffix=" " variable="page" form="short"/>
       <text variable="page"/>
     </group>
   </macro>
   <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true">
     <layout prefix="(" suffix=")" delimiter="; ">
       <group delimiter=" ">
-        <text macro="author-short"/>
```
damnation333: Your style doesn't validate anymore as you're not using the author-short macro anymore. Something tells me that this might not be what you want.
Could you also update the documentation links?

international-labour-organization.csl
```diff
@@ -78,9 +81,10 @@
     </choose>
   </macro>
   <macro name="publisher">
-    <group delimiter=", ">
+    <group>
       <text variable="event"/>
-      <text variable="publisher-place" prefix="(" suffix=")"/>
+      <text variable="publisher-place" prefix=" (" suffix=", "/>
+      <text variable="publisher" suffix=")"/>
```
rmzelle: @vl2, parentheses should not be split over multiple elements. Instead of

```xml
    <group>
      <text variable="event"/>
      <text variable="publisher-place" prefix=" (" suffix=", "/>
      <text variable="publisher" suffix=")"/>
    </group>
```

you can use

```xml
    <group delimiter=" ">
      <text variable="event"/>
      <group prefix="(" suffix=")" delimiter=", ">
        <text variable="publisher"/>
      </group>
    </group>
```

international-labour-organization.csl
```diff
@@ -109,15 +113,15 @@
     </choose>
   </macro>
   <macro name="pages">
-    <group>
-      <label variable="page" form="short" suffix=" "/>
+    <group prefix="p">
+      <label suffix=" " variable="page" form="short"/>
       <text variable="page"/>
     </group>
```
rmzelle: @vl2, this is also not correct. `<label suffix=" " variable="page" form="short"/>` should already give "p. ". If you want "pp.", the proper way is to use the plural form of the label term, i.e. `<label variable="page" form="short" plural="true" suffix=" "/>`, and remove the `prefix="p"` from the group.

international-labour-organization.csl
rmzelle: (@adam3smith, @damnation333, this PR needs some careful review)

international-labour-organization.csl
```diff
@@ -137,70 +141,84 @@
...
         </if>
         <else-if type="paper-conference" match="any">
-          <text macro="title" font-style="italic" prefix=" " suffix=","/>
+          <text variable="title" quotes="false" font-style="normal" font-weight="normal" prefix=" " suffix=","/>
           <group prefix=" " delimiter=" ">
             <text macro="editor"/>
             <text variable="container-title" font-style="italic" suffix="."/>
             <text variable="collection-title" suffix=","/>
-            <text macro="edition"/>
+            <text macro="edition" suffix=" "/>
```
adam3smith: no need for the suffix here -- it's covered by the delimiter in the group above.

international-labour-organization.csl
```diff
@@ -137,70 +141,84 @@
...
               <text macro="pages"/>
             </group>
           </group>
         </else-if>
-        <else-if type="chapter paper-conference" match="any">
-          <text macro="title" prefix=" " suffix="."/>
+        <else-if type="chapter" match="any">
           <group prefix=" " delimiter=" ">
-            <text term="in" text-case="capitalize-first" suffix=":"/>
+            <text variable="title" font-style="normal" suffix=", "/>
```
adam3smith: remove the space in the suffix.

international-labour-organization.csl
```diff
@@ -137,70 +141,84 @@
...
+            <text term="in" text-case="lowercase" font-variant="normal"/>
             <text macro="editor"/>
             <text variable="container-title" font-style="italic" suffix="."/>
             <text variable="collection-title" suffix="."/>
+            <text variable="scale"/>
             <text macro="edition"/>
             <group suffix="." delimiter=", ">
-              <text macro="publisher"/>
+              <text macro="publisher" prefix=" "/>
```
adam3smith: no need for the prefix here.

international-labour-organization.csl
```diff
@@ -137,70 +141,84 @@
...
               <text macro="pages"/>
             </group>
           </group>
         </else-if>
-        <else-if type="thesis">
```
adam3smith: why are we removing thesis here?

international-labour-organization.csl
```diff
@@ -137,70 +141,84 @@
...
           </group>
         </else-if>
+        <else-if type="report" match="any">
+          <text macro="title" font-style="italic" prefix=" "/>
+          <group>
+            <text macro="editor"/>
+            <text variable="container-title"/>
+            <text macro="edition"/>
+            <group>
+              <text variable="publisher-place" prefix=" (" suffix=", "/>
```
adam3smith: see rmzelle's comment about parentheses above. These should go on the group. and the `suffix=", "` should be a delimiter.

adam3smith:
Continued here: https://github.com/citation-style-language/styles/pull/2567
FYI @vl2 , it's much harder for us to review your update when you put it into a new pull request rather than updating the one we reviewed, so please in the feature update pull requests instead of issuing new ones. Instructions are in the automated message above.

PR #2536, opened 2017-02-27
Create evolutionary-bioinformatics.csl

damnation333:
via https://forums.zotero.org/discussion/56459/style-request-evolutionary-bioinformatics-libertas-academica-journal

### DIFF
--- a/dependent/evolutionary-bioinformatics.csl
+++ b/dependent/evolutionary-bioinformatics.csl
@@ -5,12 +5,12 @@
     <id>http://www.zotero.org/styles/evolutionary-bioinformatics</id>
     <link href="http://www.zotero.org/styles/evolutionary-bioinformatics" rel="self"/>
     <link href="http://www.zotero.org/styles/sage-vancouver" rel="independent-parent"/>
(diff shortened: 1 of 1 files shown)


rmzelle:
It looks like all the SAGE journals listed at http://insights.sagepub.com/journal_homepage.php use AMA (didn't check if that's any different from SAGE Vancouver). http://insights.sagepub.com/author_resources.php?folder_id=135&sub_menu_id=137:

> Articles published in these journals use the AMA reference formatting style.  

I couldn't find a spreadsheet that lists these journals (which would make it a lot easier to generate CSL styles for them), and it's not totally clear to me what collection they belong to ("Insights"?). They e.g. aren't listed in the "SAGE Premier 2017 Title List" at https://us.sagepub.com/en-us/nam/title-lists.

damnation333:
Sounds like we should write to them and ask them?

adam3smith:
I've tried to write to Sage many times without any real success. I got a somewhat promising answer one, but that, too, dried up in spite of multiple follow-ups.

rmzelle:
Thanks!

rmzelle:
(looks like Travis CI created a ghost build? It wasn't me who triggered this)

PR #2538, opened 2017-02-28
Create research-on-biomedical-engineering.csl

damnation333:
via https://forums.zotero.org/discussion/64098/style-for-rbe-research-on-biomedical-engineering#latest
Instructions: http://www.rbejournal.org/authors
Sample: http://www.rbejournal.org/doi/10.1590/2446-4740.03615

### DIFF
--- a/research-on-biomedical-engineering.csl
+++ b/research-on-biomedical-engineering.csl
@@ -1,0 +1,201 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" page-range-format="minimal" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>Research on Biomedical Engineering</title>
+    <title-short>RBE</title-short>
+    <id>http://www.zotero.org/styles/research-on-biomedical-engineering</id>
+    <link href="http://www.zotero.org/styles/research-on-biomedical-engineering" rel="self"/>
+    <link href="http://www.zotero.org/styles/elsevier-vancouver-author-date" rel="template"/>
+    <link href="http://www.rbejournal.org/authors" rel="documentation"/>
+    <author>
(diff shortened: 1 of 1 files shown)


adam3smith:
great, thanks!

research-on-biomedical-engineering.csl
```diff
@@ -0,0 +1,199 @@
...
+          <text term="edition" form="short"/>
+        </group>
+      </if>
+      <else>
+        <text variable="edition"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="access">
+    <text variable="PMID" prefix="PMid:" suffix=". "/>
```
rmzelle: @damnation333, did you use "PMid" on purpose? I've never seen this abbreviation in mixed case. Also, http://www.rbejournal.org/authors gives as example:

> Agarwal S, Rao A. Modulation of chromatin structure regulates cytokine gene expression during T-cell differentiation. Immunity. 1998; 9(6):765-75. PubMed: PMID 12000309.

so shouldn't the prefix be "PubMed: PMID "?

research-on-biomedical-engineering.csl
```diff
@@ -0,0 +1,199 @@
...
+    <choose>
+      <if variable="DOI">
+        <text variable="DOI" prefix="doi: http://dx.doi.org/"/>
```
rmzelle: @damnation333, even if the examples are different, just use `https://doi.org/` instead of `http://dx.doi.org/`.

research-on-biomedical-engineering.csl
damnation333: @rmzelle Their guidelines do show Pubmed: PMID 1200309, but in their papers it is only: PMid:1200309

research-on-biomedical-engineering.csl
damnation333: I will make that change in https://github.com/citation-style-language/styles/pull/2539

PR #2539, opened 2017-03-01
Update research-on-biomedical-engineering.csl

damnation333:
See for requested changes: https://forums.zotero.org/discussion/comment/271133/#Comment_271133

How can I make the conditional that when the journal article is not published yet, but got a URL and to make it print the URL?

### DIFF
--- a/research-on-biomedical-engineering.csl
+++ b/research-on-biomedical-engineering.csl
@@ -88,7 +88,7 @@
     </choose>
     <choose>
       <if variable="DOI">
-        <text variable="DOI" prefix="doi: http://dx.doi.org/"/>
+        <text variable="DOI" prefix="doi: https://dx.doi.org/"/>
       </if>
       <else-if type="webpage post-weblog" match="any">
         <group delimiter=" " prefix=" [" suffix="]. ">
--- a/research-on-biomedical-engineering.csl
+++ b/research-on-biomedical-engineering.csl
@@ -15,7 +15,7 @@
     <category field="biology"/>
     <issn>2446-4732</issn>
     <eissn>2446-4740</eissn>
-    <updated>2017-02-28T10:09:51+00:00</updated>
+    <updated>2017-03-01T13:21:32+00:00</updated>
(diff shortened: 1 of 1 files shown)


rmzelle:
> How can I make the conditional that when the journal article is not published yet, but got a URL and to make it print the URL?

Typically you would test for the "page" variable. If it's empty, the article probably hasn't yet been assigned an issue.

research-on-biomedical-engineering.csl
```diff
@@ -88,7 +88,7 @@
     </choose>
     <choose>
       <if variable="DOI">
-        <text variable="DOI" prefix="doi: http://dx.doi.org/"/>
+        <text variable="DOI" prefix="doi: https://dx.doi.org/"/>
```
rmzelle: @damnation333, please drop the "dx." too.

research-on-biomedical-engineering.csl
```diff
@@ -63,13 +63,9 @@
...
-    <group delimiter="; ">
-      <group delimiter=": ">
-        <text variable="publisher-place"/>
-        <text variable="publisher"/>
-      </group>
-      <text macro="year-date"/>
-    </group>
+    <text variable="publisher-place" suffix=": " text-case="title"/>
+    <text variable="publisher" suffix="; "/>
+    <text macro="year-date"/>
```
rmzelle: @damnation333, you reverted all those beautiful group delimiters that @adam3smith added in the first PR. 😭

research-on-biomedical-engineering.csl
```diff
@@ -85,10 +81,14 @@
     </choose>
   </macro>
   <macro name="access">
-    <text variable="PMID" prefix="PMid:" suffix=". "/>
+    <choose>
+      <if match="any" variable="PMID">
+        <text variable="PMID" prefix="PMid:" suffix=". "/>
+      </if>
+    </choose>
```
rmzelle: @damnation333, same here. Could you incorporate @adam3smith's changes from https://github.com/citation-style-language/styles/pull/2538/commits/7d1a4bfc638cdeef436d2b2da4c642af64ed2647 ?

rmzelle:
Thanks!

PR #2541, opened 2017-03-02
Add source types and URLs to Royal Society styles

mdoube:
Add URL data for newspaper articles, webpages, books, only if a DOI is missing.


rmzelle:
Thanks. I made a few tweaks to clean up the CSL code that shouldn't have affected the output.

Where did you find the guidelines for e.g. newspaper articles? Did you find examples in one of the journals?

mdoube:
Examples are from recent (Feb 2017) articles in RSOS. There's some variability as to inclusion of (accessed on <date>), and no strict guidelines published. These are really corner cases: the great majority of citations are to journal articles or other items with DOIs.

rmzelle:
Okay, thanks!

PR #2542, opened 2017-03-02
Update music-theory-spectrum.csl

damnation333:
Via request here: https://forums.zotero.org/discussion/comment/271245/#Comment_271245
@rmzelle Can you confirm I made the right changes? Doing this blind due to lack of testing abilities. Thanks

### DIFF
--- a/music-theory-spectrum.csl
+++ b/music-theory-spectrum.csl
@@ -391,13 +391,13 @@
     <layout suffix="." delimiter="; ">
       <choose>
         <if position="ibid-with-locator">
-          <group delimiter=", ">
+          <group delimiter=" " prefix="(" suffix=")">
             <text term="ibid"/>
             <text macro="point-locators-subsequent"/>
           </group>
         </if>
         <else-if position="ibid">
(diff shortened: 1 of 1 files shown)


adam3smith:
(not right; fixing in commit right now)

music-theory-spectrum.csl
```diff
@@ -391,13 +391,13 @@
     <layout suffix="." delimiter="; ">
       <choose>
         <if position="ibid-with-locator">
-          <group delimiter=", ">
+          <group delimiter=" " prefix="(" suffix=")">
```
rmzelle: @damnation333, I think the parentheses belong on the `<text macro="point-locators-subsequent"/>` to get "Ibid. (187-8)." Currently the "ibid" term would be wrapped as well.

I also don't think you need the parentheses on the "ibid" term below.

adam3smith: please merge once test passes?

rmzelle:
Squashed!

image
(https://flic.kr/p/db5MqE)

PR #2548, opened 2017-03-05
Create caduceus-lane-style.csl

samatawy:
This style applies the settings requested frequently in healthcare documents particularly presentations and marketing material created by the pharmaceutical industry in the MENA region.
Thank you for your efforts.


rmzelle:
(continued in #2550)

PR #2550, opened 2017-03-05
Create caduceus-lane-style.csl

samatawy:
This style applies the settings requested frequently in healthcare documents particularly presentations and marketing material created by the pharmaceutical industry in the MENA region.
Thank you for your efforts.

### DIFF
--- a/caduceus-lane-style.csl
+++ b/caduceus-lane-style.csl
@@ -1,0 +1,125 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>Caduceus Lane Style</title>
+    <title-short>CLH</title-short>
+    <id>http://www.zotero.org/styles/caduceus-lane-style</id>
+    <link href="http://www.zotero.org/styles/caduceus-lane-style" rel="self"></link>
+    
+    <contributor>
+      <name>Sameh S El-Atawy</name>
+      <uri>http://www.caduceuslane.com/staff/selatawy</uri>
+    </contributor>
+    <category citation-format="numeric"/>
+    <category field="medicine"/>
+        <summary>A mix of Harvard and Vancouver styles for the pharmaceutical marketing industry</summary>
+    <updated>2017-03-05T12:28:58+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <macro name="author">
+    <names variable="author" suffix=" ">
+      <name and="text" initialize-with="." name-as-sort-order="all"/>
+      <label form="short" text-case="capitalize-first" prefix=", "/>
+      <substitute>
+        <names variable="editor"/>
+        <names variable="translator"/>
+      </substitute>
+    </names>
+    <text macro="year-date" prefix="(" suffix=") "/>
+  </macro>
+  <macro name="editor">
+    <names variable="editor">
+      <name initialize-with="." delimiter=", " delimiter-precedes-last="always"/>
+      <label form="short" prefix=" (" text-case="capitalize-first" suffix=")"/>
+    </names>
+  </macro>
+  <macro name="year-date">
+    <choose>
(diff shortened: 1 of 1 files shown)


rmzelle:
@samatawy, is there a reason you closed all your pull requests? In general, if there are issues with the original pull request, it's preferred to update it instead of creating a new pull request.

More importantly, though, we typically only accept styles for popular style guides and for publications that accept open submissions (see https://github.com/citation-style-language/styles/wiki/Criteria-for-Accepting-Styles). If this is a CSL style for internal use by your company, it's probably not a good candidate for inclusion into our repository.

samatawy:
Thank you for your feedback
I'm sorry about closing pull requests - I thought that would help the automated process. My mistake.

As regards the style, I totally understand your position. However, the style is already used pretty widely in the pharmaceutical and healthcare industries (at least in North Africa and the Gulf where we operate) but isn't available in common software, so users often have to format their bibliographies by hand.

Our clients include Pfizer, Novartis, Gilead, AstraZeneca, and others - who all prefer to use this style in their publications and marketing material. Let me contact them and see how far they really need to have this style available globally. I might need to contact you again shortly. I hope that would be OK.

Thanks again for your time.
Dr Sameh S El-Atawy

rmzelle:
> the style is already used pretty widely in the pharmaceutical and healthcare industries (at least in North Africa and the Gulf where we operate)

Is there an authoritative source that describe this citation format? A style manual or something like that?

samatawy:
Sorry to have been so late to reply, but I'm still waiting for responses
from our clients. It might take a few more days.

The style in question is just a mix of two well-known families:

   1. Vancouver-style numbering for in-line markers (since it takes up less
   space and can therefore be used in brochures, flyers, and presentations),
   and
   2. Harvard-style bibliography text (a bland variant only used to refer
   to journal articles and text book sources) that does not include either doi
   or url information.

In a few days I might get something more solid but this is what almost all
pharmaceutical marketing material uses for as far as I can tell. Hope to be
contacting you soon.

Thanks for your time.
Yours,
Dr Sameh El-Atawy

rmzelle:
> Our clients include Pfizer, Novartis, Gilead, AstraZeneca, and others - who all prefer to use this style in their publications and marketing material.

> this is what almost all pharmaceutical marketing material uses for as far as I can tell.

Do you know who came up with this style format? How do you know what the citation format is supposed to look like? Is it dictated by your clients or did your company decide on a certain format?

samatawy:
I have not been involved with the company since its beginning, but I guess
the style arose from (1) using a common format (2) in a space-constrained
environment. My understanding thus far is this:

   - The most commonly used tool (*de facto* standard) for generating
   citations in our industry is *http://www.citethisforme.com/
   <http://www.citethisforme.com/>* (other tools also have similar default
   styles).

   - Companies often select the Harvard style when using this tool (and
   demand us to do so) and check our referencing against it. For example:

Baccarani, M., Pane, F. and Saglio, G. (2008). Monitoring treatment of
> chronic myeloid leukemia. *Haematologica*, 93(2), pp.161-169.


Bloomgarden, Z. (1994). A Review of Current Trends in Diabetes. *Diabetes
> Care*, 17(7), pp.786-790.


   - However, they do not use in-line referencing in the same style - which
   they reject for being too verbose. They use simple numbering instead in
   Vancouver style (numbering per page, brochure, or slide).

   - This mix is loosely referred to as "Harvard-style" despite the
   mismatch between in-line and bibliographic styling. If reference tools
   allowed us to mix and match style parts, we would have no need for a
   special style.

This mixing is why we cannot use common software tools to automate
referencing and we have to do it all manually. The bibliography is copied
one by one, ordered, numbered, and the numbers are used in-line as they
appear. As you can see, this is a tedious process and prone to error.

Please note that we do not use this style in academic publications (journal
articles, etc.). It is used mainly within the pharmaceutical industry for
presentations and documents where space and readability are overriding
concerns (although I have also seen it in local theses on occasion).

In some situations we strictly follow standard styles for certain
documents. For example regulatory documents follow the style required by
the government agency (FDA for example), but this poses no problem as these
styles are already published.

Thank you for being so patient - I highly appreciate it.

Yours,
Dr Sameh El-Atawy

adam3smith:
And the requirements for this style are pretty strict? E.g. taking a style such as https://www.zotero.org/styles?q=id%3Aspringer-basic-brackets which is very similar in nature: i.e. the ordering is more akin to a Harvard style, but it uses numerical citations, doesn't work?

In that case, our minimum standard would be that we have the style documented somewhere, so we can check if issues come up. Ideally that'd be on a webpage somewhere, but we'd also take a document attached here.

PR #2551, opened 2017-03-05
Create journal-of-the-association-of-environmental-and-resource-econo…

damnation333:
…mists.csl

via https://forums.zotero.org/discussion/51352/style-request-journal-of-the-association-of-environmental-and-resource-economists

article (free): http://www.journals.uchicago.edu/doi/pdfplus/10.1086/689865


rmzelle:
Thanks!

PR #2555, opened 2017-03-07
Create journal-of-historical-linguistics.csl

damnation333:
via https://forums.zotero.org/discussion/63433/style-request-journal-of-historical-linguistics
adapted for book, chapter and journal (examples in thread)


rmzelle:
Thanks!

PR #2558, opened 2017-03-08
FAOSTYLE Major Update

swifterslb:
Complete revamping. FAO is preparing a new version of FAOSTYLE. The official document has not been updated yet, but the output results of this CSL style has been thoroughly checked by FAO's senior English editor, listed as a contributor. Many new item types have been added and the codes have been optimized compared to the last version.

### DIFF
--- a/food-and-agriculture-organization-of-the-united-nations.csl
+++ b/food-and-agriculture-organization-of-the-united-nations.csl
@@ -1,9 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
   <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
-  <!-- This style was designed to work with both Zotero and Mendeley. -->
   <info>
-    <title>Food and Agriculture Organization of the United Nations - revision test</title>
+    <title>Food and Agriculture Organization of the United Nations</title>
     <title-short>FAO</title-short>
     <id>http://www.zotero.org/styles/food-and-agriculture-organization-of-the-united-nations</id>
     <link href="http://www.zotero.org/styles/food-and-agriculture-organization-of-the-united-nations" rel="self"/>
@@ -20,7 +19,7 @@
     <category citation-format="author-date"/>
     <category field="science"/>
     <category field="social_science"/>
-    <summary>This style is created to meet the citation and bibliographical requirements of FAOSTYLE. Last update: March 2017.</summary>
+    <summary>This style is created to meet the citation and bibliographical requirements of FAOSTYLE, and has been tested with Zotero and Mendeley. Last update: March 2017.</summary>
     <updated>2017-03-08T16:33:05+01:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
--- a/food-and-agriculture-organization-of-the-united-nations.csl
+++ b/food-and-agriculture-organization-of-the-united-nations.csl
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
   <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
-  <!-- This style was designed to work with both Zotero and Mendeley. -->
   <info>
     <title>Food and Agriculture Organization of the United Nations</title>
     <title-short>FAO</title-short>
--- a/food-and-agriculture-organization-of-the-united-nations.csl
+++ b/food-and-agriculture-organization-of-the-united-nations.csl
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
   <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <!-- This style was designed to work with both Zotero and Mendeley. -->
   <info>
     <title>Food and Agriculture Organization of the United Nations</title>
     <title-short>FAO</title-short>
@@ -20,19 +21,16 @@
     <category field="science"/>
     <category field="social_science"/>
     <summary>This style is created to meet the citation and bibliographical requirements of FAOSTYLE, and has been tested with Zotero and Mendeley. Last update: March 2017.</summary>
-    <updated>2017-03-08T16:33:05+01:00</updated>
+    <updated>2017-03-10T01:12:03+01:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <!-- Locale settings for English. Settings for other languages could be added later to this section. -->
   <locale xml:lang="en">
     <style-options punctuation-in-quote="false"/>
     <terms>
-      <term name="editor" form="verb-short">ed. by</term>
-      <term name="editor" form="short">
-        <single>ed</single>
-        <multiple>eds</multiple>
-      </term>
       <term name="no date">undated</term>
+      <term name="available at">also available at</term>
+      <term name="presented at">Paper presented at</term>
     </terms>
     <date delimiter=" " form="text">
       <date-part name="day"/>
@@ -45,42 +43,30 @@
     <!-- XX ed. / XX & XX eds. -->
     <names variable="editor">
       <name and="symbol" delimiter-precedes-last="never" initialize-with="."/>
-      <label form="short" strip-periods="false" prefix=", "/>
+      <label form="short" strip-periods="true" prefix=", "/>
     </names>
   </macro>
   <macro name="author">
-    <choose>
-      <if match="any" variable="author editor translator">
-        <names variable="author">
-          <name and="symbol" delimiter-precedes-last="never" initialize-with="." name-as-sort-order="all"/>
-          <label form="short" prefix=", "/>
-          <substitute>
-            <names variable="editor"/>
-            <names variable="translator"/>
-          </substitute>
-        </names>
-      </if>
-      <else>
+    <names variable="author">
+      <name and="symbol" delimiter-precedes-last="never" initialize-with="." name-as-sort-order="all"/>
+      <label form="short" prefix=", "/>
+      <substitute>
+        <names variable="editor"/>
+        <names variable="translator"/>
         <text term="anonymous" text-case="capitalize-first"/>
-      </else>
-    </choose>
+      </substitute>
+    </names>
   </macro>
   <macro name="author-short">
-    <choose>
-      <if match="any" variable="author editor translator">
-        <names variable="author">
-          <name form="short" and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="4" et-al-use-first="1" initialize-with="."/>
-          <et-al font-style="italic"/>
-          <substitute>
-            <names variable="editor"/>
-            <names variable="translator"/>
-          </substitute>
-        </names>
-      </if>
-      <else>
+    <names variable="author">
+      <name form="short" and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="4" et-al-use-first="1" initialize-with="."/>
+      <et-al font-style="italic"/>
+      <substitute>
+        <names variable="editor"/>
+        <names variable="translator"/>
         <text term="anonymous" text-case="capitalize-first"/>
-      </else>
-    </choose>
+      </substitute>
+    </names>
   </macro>
   <macro name="date">
     <!-- date format: year / undated / put "forthcoming" or else in "Date" field when needed.-->
@@ -94,14 +80,10 @@
     </choose>
   </macro>
   <macro name="publisher-and-place">
-    <!-- place / place, publisher -->
-    <group>
+    <!-- place, publisher -->
+    <group delimiter=", ">
       <text variable="publisher-place"/>
-      <choose>
-        <if variable="publisher">
-          <text variable="publisher" prefix=", "/>
-        </if>
-      </choose>
+      <text variable="publisher"/>
     </group>
   </macro>
   <macro name="edition">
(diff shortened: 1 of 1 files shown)


rmzelle:
@swifterslb, you removed the "default-locale" attribute intentionally? Do you envision that the style will be used in languages other than English?

swifterslb:
@rmzelle Yes, I intentionally removed default-locale. Important FAO publications are usually published in at least English, French, and Spanish. I'd like to leave room to accommodate them. Whether this could be achieved is, of course, decided upon future consultation with editors of the other languages.

food-and-agriculture-organization-of-the-united-nations.csl
```diff
@@ -10,196 +10,368 @@
...
-    <names variable="author" suffix=". ">
-      <name and="symbol" delimiter-precedes-last="never" initialize-with="." name-as-sort-order="all"/>
-      <label form="short" prefix=" "/>
-      <substitute>
-        <names variable="editor"/>
-        <names variable="translator"/>
-        <text variable="title"/>
-      </substitute>
-    </names>
+    <choose>
```
adam3smith: There's no need for this choose -- if here. You can just add `<text term="anonymous" .../>` as the final element of substitute for the same effect

food-and-agriculture-organization-of-the-united-nations.csl
```diff
@@ -10,196 +10,368 @@
...
-    <names variable="author">
-      <name form="short" and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="4" et-al-use-first="1" initialize-with="."/>
-      <et-al font-style="italic"/>
-      <substitute>
-        <names variable="editor"/>
-        <names variable="translator"/>
-        <text variable="title"/>
-      </substitute>
-    </names>
+    <choose>
```
adam3smith: ditto on choose -- if

food-and-agriculture-organization-of-the-united-nations.csl
```diff
@@ -10,196 +10,368 @@
...
+        <text term="no date"/>
+      </if>
+      <else>
+        <date date-parts="year" form="text" variable="issued"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="publisher-and-place">
+    <!-- place / place, publisher -->
+    <group>
```
adam3smith: much simpler and more robust as

```
<group delimiter=", ">
  <text variable="publisher"/>
</group>
```

food-and-agriculture-organization-of-the-united-nations.csl
```diff
@@ -10,196 +10,368 @@
...
-    <text term="edition" font-style="italic"/>
+    <!-- Second edition -->
+    <number text-case="capitalize-first" variable="edition" form="long-ordinal"/>
+    <text term="edition" prefix=" "/>
+  </macro>
+  <macro name="Cite-and-URL">
+    <!-- [Cited day month year]. URL -->
+    <choose>
+      <if variable="URL">
+        <text term="cited" text-case="capitalize-first" prefix="["/>
```
adam3smith: again, better with group as

```
<group delimiter=" " prefix="[" suffix="].">
    <text term="cited" text-case="capitalize-first"/>
  <date form="text" variable="accessed"/>
</group>
```

food-and-agriculture-organization-of-the-united-nations.csl
```diff
@@ -10,196 +10,368 @@
...
+        <date form="text" variable="accessed" prefix=" " suffix="]."/>
+        <text variable="URL" prefix=" "/>
+      </if>
+    </choose>
+  </macro>
+  <macro name="Available-at">
+    <!-- (also available at URL). -->
+    <choose>
+      <if variable="URL">
```
adam3smith: again use group here, which also saves you the need for the if.
i.e. `<group delimiter=" " prefix="(also " suffix=").">` etc.

food-and-agriculture-organization-of-the-united-nations.csl
```diff
@@ -10,196 +10,368 @@
...
   </macro>
-  <macro name="Available-at-and-accessed">
-    <text term="available at" text-case="capitalize-first" suffix=": "/>
-    <text variable="URL" suffix=". "/>
-    <text term="accessed" text-case="capitalize-first" suffix=" "/>
-    <date form="text" variable="accessed" suffix="."/>
+  <macro name="number_of_pages-label">
+    <!-- 1 p. / x pp. -->
+    <choose>
+      <if is-numeric="number-of-pages">
```
adam3smith: and you're sure you don't want to print the number of pages if it's not numeric, e.g. just as a literal string?

adam3smith: Also, again, this can be done more elegantly with a group, saving you the choose -- if

food-and-agriculture-organization-of-the-united-nations.csl
```diff
@@ -10,196 +10,368 @@
...
-            <date date-parts="year" form="text" variable="issued"/>
-          </if>
-          <else>
-            <text value="undated"/>
-          </else>
-        </choose>
-        <text variable="locator"/>
-      </group>
+      <text macro="author-short" suffix=", "/>
+      <text macro="date"/>
```
adam3smith: just to confirm -- there will never be page numbers or other locators given in the text?

food-and-agriculture-organization-of-the-united-nations.csl
```diff
@@ -10,196 +10,368 @@
...
-              <if match="any" variable="URL">
-                <text macro="Available-at-and-accessed"/>
+              <if type="book">
+                <!-- book and computer program in Zotero -->
+                <group>
+                  <text variable="title" font-style="italic"/>
+                  <text variable="medium" prefix=" [" suffix="]"/>
+                  <!-- For computer program in Zotero. Can put "CD&#8211;ROM / DVD&#8211;ROM" in "System" field. -->
+                  <text variable="genre" prefix=" [" suffix="]"/>
+                  <!-- For eBook  in Mendeley. Can put "Kindle edition" in "Genre" field. -->
```
adam3smith: btw. you can do that in Zotero, too, by putting `genre: Kindle edition` into extra.

food-and-agriculture-organization-of-the-united-nations.csl
```diff
@@ -10,196 +10,368 @@
...
+    <summary>This style is created to meet the citation and bibliographical requirements of FAOSTYLE, and has been tested with Zotero and Mendeley. Last update: March 2017.</summary>
+    <updated>2017-03-08T16:33:05+01:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
+  <!-- Locale settings for English. Settings for other languages could be added later to this section. -->
   <locale xml:lang="en">
+    <style-options punctuation-in-quote="false"/>
     <terms>
       <term name="editor" form="verb-short">ed. by</term>
       <term name="editor" form="short">
```
adam3smith: If both forms are going to be without period, just remove the terms and use strip-periods="true" in the style for simplicity.

food-and-agriculture-organization-of-the-united-nations.csl
```diff
@@ -10,196 +10,368 @@
...
+                  <!-- "Series Number" field for book section in Zotero. -->
+                  <text variable="number" prefix=" No. "/>
+                  <!-- "Report Number" field for report in Zotero. -->
+                  <!-- Mendeley doesn't offer a "Series Number / Report Number" field. "No. xxx" must be put in "Series" field. -->
+                </group>
+                <group suffix=".">
+                  <text macro="publisher-and-place" prefix=" "/>
+                  <group>
+                    <text variable="number-of-pages" suffix=" "/>
+                    <label suffix="." variable="page" form="short"/>
```
adam3smith: No need for the suffix period here. That's already in the label (also, you have defined a Macro for this -- why not use that?

food-and-agriculture-organization-of-the-united-nations.csl
```diff
@@ -10,196 +10,368 @@
...
+                  <text macro="editor" prefix=" "/>
+                </group>
+                <text variable="container-title" font-style="italic"/>
+                <group suffix=".">
+                  <label suffix=" " variable="page" form="short"/>
+                  <text variable="page"/>
+                </group>
+                <text variable="collection-title"/>
+                <group suffix=".">
+                  <text term="presented at" prefix="Paper "/>
```
adam3smith: to facilitate locatlization, you'll want to definte the presented-at term to include paper. That way you don't have to hardcode the English "Paper"

food-and-agriculture-organization-of-the-united-nations.csl
```diff
@@ -10,196 +10,368 @@
...
+        <date form="text" variable="accessed" prefix=" " suffix="]."/>
+        <text variable="URL" prefix=" "/>
+      </if>
+    </choose>
+  </macro>
+  <macro name="Available-at">
+    <!-- (also available at URL). -->
+    <choose>
+      <if variable="URL">
+        <text term="available at" text-case="lowercase" prefix="(also "/>
```
adam3smith: Here, too, for better localization, you'll want to define the also as part of the term (which I don't think you use anywhere without also, right?)

food-and-agriculture-organization-of-the-united-nations.csl
swifterslb: Need help here. I write the substitute part as:
```
<substitute>
  <names variable="editor"/>
  <text term="anonymous" text-case="capitalize-first"/>
</substitute>
```
But the result is if there's no author/editor/translator, "Anonymous" is not shown. The author part is just missing. How to fix?

food-and-agriculture-organization-of-the-united-nations.csl
swifterslb: I put it this way because sometimes there is only publisher-place but no publisher. In this case, the output will be "publisher-place." (ending with a period), not "publisher-place," (ending with a comma, which is undesired).

food-and-agriculture-organization-of-the-united-nations.csl
adam3smith: works for me, could you include it in the style with all the other revisions and we can then test it? If there is a problem with how Zotero/Mendeley handle this we'd need to know.

food-and-agriculture-organization-of-the-united-nations.csl
adam3smith: Same with the group, which has the additional advantage that it also works if there is a publisher but no place.

food-and-agriculture-organization-of-the-united-nations.csl
swifterslb: I've never considered this. Will check with the editor.

food-and-agriculture-organization-of-the-united-nations.csl
swifterslb: Thanks! Does this mean you can add "Genre" field to any item type by this way in Zotero?

food-and-agriculture-organization-of-the-united-nations.csl
swifterslb: There's a "Speech presented at" before so it's kind of difficult. But your point of localization is good. I'll define the term and persuade the editor to change the output of the presentation type to "Paper presented at" in the style guide document. Presentation is very rarely used, anyway.

food-and-agriculture-organization-of-the-united-nations.csl
adam3smith: yup.

food-and-agriculture-organization-of-the-united-nations.csl
swifterslb: The editor confirms that locators will not be used in in-line citation.

food-and-agriculture-organization-of-the-united-nations.csl
```diff
@@ -51,155 +63,287 @@
...
-            <text variable="issue" form="short" prefix="(" suffix=")"/>
+          <!-- case -->
+          <group delimiter=" ">
+            <group>
+              <text variable="title"/>
+              <text variable="volume" prefix=", "/>
+            </group>
+            <text variable="container-title"/>
+            <text variable="page"/>
+            <text variable="authority" prefix="("/>
```
adam3smith: sorry, I overlooked this, but you don't want to distribute affixes over elements. Instead use a group

```
<group delimiter=" " prefix="(" suffx=")">
  <text variable="authority"/>
  <text variable="publisher"/>
  <!-- "Publisher" field in Mendeley -->
  <text macro="date"/>
</group>
```

swifterslb: Neither Zotero nor Mendely works... See attached screenshot. Should I revert to choose--if scheme?

image

adam3smith: OK, I see. It's supressing that because it's just a term, so you need to wrap it in a macro, i.e. create
```
<macro name="anonymous">
  <text term="anonymous" text-case="capitalize-first"/>
</macro>
```

and then use `<text macro="anonymous"/>` in substitute.

Your example, though, indicates a problem in your style: you presumably don't want anonymous for in-text citations of legal cases (as you explicitly don't have it in the bibliography).

swifterslb: Thanks so much for the kind help! Changes made in the new commit.

rmzelle:
> It's supressing that because it's just a term

(is that official CSL behavior?)

food-and-agriculture-organization-of-the-united-nations.csl
rmzelle: > It's supressing that because it's just a term

(is that official CSL behavior?)

adam3smith: I think it's borderline. @fbennett will have the details clearer, but I think cs:names is treated like a group, so when it only has a term in it, that will get discarded.

swifterslb: I have another question. To achieve the output that legal case is (_casetitle_, date) and all others are (author, date), I firstly wrote in the citation section without changing the author-short macro:
```
  <choose>
    <if match="none" type="legal_case"> 
      <text macro="author-short" suffix=", "/>
      <text macro="date"/>
    </if>
    <else>
      <text macro="date"/>
    </else>
  </choose>
</layout>
```
This renders fine when legal case is cited alone. However, if an (author, date) item precedes a case, the punctuation separating them becomes comma instead of semicolon: (author, date, _casetitle_, date), not (author, date; _casetitle_, date). I checked with other types and changed the author-short macro, and it worked fine, thus the last commit. Why is this?

rmzelle:
> However, if an (author, date) item precedes a case, the punctuation separating them becomes comma instead of semicolon: (author, date, casetitle, date), not (author, date; casetitle, date).

Which version of the style gives that wrong output? As long the "layout" element has `delimiter="; "`, the "; " should always separate (non-collapsed) items in the same citation.

swifterslb:
I didn't push the problematic version. If the citation part of the penultimate commit (8a5ff30) were written as I provided in the comment, it would behave as I described.

swifterslb:
I believe there are no problems the last commit. Could you please merge the pull request first? So that I could inform the senior editor that the new style is ready for download from the official source. We can always discuss the potential bugs later. Thank you very much.

adam3smith:
yup, the style as is is fine, thanks.
Could you put a version of the (presumably) malfunctioning style on gist.github.com as a secret gist and link to it here so we can look at that more closely? The citeproc processor used in Zotero&Mendeley includes some special code for legal citations, and it's likely misbehaving here.

swifterslb:
Thanks a lot. The secret gist is at: https://gist.github.com/swifterslb/827a69f935215adca94dff24b0baffbc

adam3smith:
@fbennett -- could you take a look? Using the CSL at  https://gist.github.com/swifterslb/827a69f935215adca94dff24b0baffbc and legal case and any other item, we get a comma between in-text citations, although it should be, as per line 125, a semicolon.

fbennett:
The bug apparently isn't triggered by legal style code: I get the same result when `article-journal` is set as the title-only exception. It's a puzzler, I'll take a look.

fbennett:
It turned out to be a simple thing, and it's a little surprising it hadn't come up earlier. In an author-date style using `collapse`, if a cite with an author was followed by one that forced rendering of something instead of the author (as here)---even if via `cs:substitute`, the processor assumed that the cite had been collapsed.

I've tweaked things so that rendering of a text variable will also be treated as a no-collapse. The amended processor is available in the [Propachi](https://juris-m.github.io/downloads/#csl-stuff) plugins.

PR #2560, opened 2017-03-09
Update apa-fr-provost.csl

lauremelli:
Add the support of the original date of publication in the "Extra" field, like for the APA style, following what is explained in this forum discussion: https://forums.zotero.org/discussion/3673/original-date-of-publication

### DIFF
--- a/apa-fr-provost.csl
+++ b/apa-fr-provost.csl
@@ -380,10 +380,15 @@
   <macro name="issued-year">
     <choose>
       <if variable="issued">
-        <date variable="issued">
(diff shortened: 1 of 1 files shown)


adam3smith:
Looks good, thanks!

lauremelli:
Is it normal that the modifications haven't yet been reported in the Zotero style repository?
Thanks!

rmzelle:
It seems up to date now: https://www.zotero.org/styles?q=id%3Aapa-fr-provost

(there is always some delay, but normally it's less than an hour)

PR #2562, opened 2017-03-10
Update journal-of-historical-linguistics.csl

damnation333:
via error request: https://forums.zotero.org/discussion/comment/271743/#Comment_271743 
It should come out as In Hall PA & Soskice D, eds., (comma between names and label), but comes out as In Hall PA & Soskice D eds.,

Am I missing something?

### DIFF
--- a/journal-of-historical-linguistics.csl
+++ b/journal-of-historical-linguistics.csl
@@ -16,7 +16,7 @@
     <category field="linguistics"/>
     <issn>2210-2116</issn>
     <eissn>2210-2124</eissn>
-    <updated>2017-03-10T10:33:29+00:00</updated>
+    <updated>2017-03-10T10:39:18+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="editor-translator">
@@ -28,7 +28,7 @@
   </macro>
   <macro name="author">
     <names variable="author">
-      <name delimiter-precedes-last="always" initialize="false" initialize-with="." name-as-sort-order="all" sort-separator=" "/>
+      <name delimiter-precedes-last="always" initialize="false" initialize-with="." name-as-sort-order="first"/>
       <et-al font-style="italic"/>
       <label form="short" prefix=" (" suffix=")" text-case="capitalize-first"/>
       <substitute>
--- a/journal-of-historical-linguistics.csl
+++ b/journal-of-historical-linguistics.csl
@@ -16,7 +16,7 @@
     <category field="linguistics"/>
     <issn>2210-2116</issn>
     <eissn>2210-2124</eissn>
-    <updated>2017-03-07T18:40:10+00:00</updated>
+    <updated>2017-03-10T10:33:29+00:00</updated>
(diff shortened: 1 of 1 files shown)


journal-of-historical-linguistics.csl
```diff
@@ -28,7 +28,7 @@
   </macro>
   <macro name="author">
     <names variable="author">
-      <name delimiter-precedes-last="always" initialize="false" initialize-with="." name-as-sort-order="all" sort-separator=" "/>
+      <name delimiter-precedes-last="always" initialize="false" initialize-with="." name-as-sort-order="first"/>
```
adam3smith: this should have `and="symbol"`

journal-of-historical-linguistics.csl
```diff
@@ -111,15 +111,13 @@
           <text macro="title" prefix=" "/>
           <group prefix=".">
             <group>
-              <group suffix=".">
-                <names variable="editor translator">
-                  <name name-as-sort-order="all" sort-separator=" " initialize-with="" delimiter=", " delimiter-precedes-last="always" prefix=" In: " suffix=","/>
+              <group prefix=" " suffix=".">
+                <text term="in" text-case="capitalize-first" suffix=" "/>
+                <names variable="editor translator" delimiter="," suffix=",">
+                  <name delimiter=" " and="symbol" delimiter-precedes-last="always" initialize-with="" name-as-sort-order="all" sort-separator=" "/>
```
adam3smith: I'm not seeing initials in the bibliographies, so remove `initialize-with=""`. I also see names in regular order, so remove the `name-as-sort-order`. Finally, I'm not seeing a comma before the ampersand, so set `delimiter-precedes-last="never"`
I'm working off this example:
image
Might also be there are different ones? But this makes sense in the context of the style

adam3smith: That same example also suggests that there shouldn't be an "in" and that the label should be before the name and be `verb-short`.

journal-of-historical-linguistics.csl
```diff
@@ -111,15 +111,13 @@
...
                   <label form="short" prefix=" "/>
                 </names>
-                <text variable="collection-title" prefix=" " suffix="." font-style="italic"/>
-                <text variable="container-title" prefix=" " suffix="." font-style="italic"/>
-                <text macro="publisher" prefix=" " suffix=", "/>
-                <text variable="page" suffix="."/>
+                <text variable="page" prefix=" " suffix="."/>
```
adam3smith: it looks like we're missing the publisher macro here now?

rmzelle:
> It should come out as In Hall PA & Soskice D, eds., (comma between names and label), but comes out as In Hall PA & Soskice D eds.,
> 
> Am I missing something?

You'd have to change the "label" element in

```xml
          <text macro="title" prefix=" "/>
          <group prefix=".">
            <group>
              <group prefix=" " suffix=".">
                <names variable="editor translator" delimiter="," suffix=",">
                  <name delimiter=" " and="symbol" delimiter-precedes-last="always" initialize-with="" name-as-sort-order="all" sort-separator=" "/>
                </names>
                <text variable="page" prefix=" " suffix="."/>
              </group>
            </group>
          </group>
        </else-if>
```

to `<label form="short" prefix=", "/>`.

damnation333:
Thanks for reviewing this style. That chapter section was completely messed up. sorry.

rmzelle:
Thanks!

PR #2563, opened 2017-03-10
Update european-journal-of-human-genetics.csl

damnation333:
via https://forums.zotero.org/discussion/64774/template-european-journal-of-human-genetics

### DIFF
--- a/european-journal-of-human-genetics.csl
+++ b/european-journal-of-human-genetics.csl
@@ -14,7 +14,7 @@
     <category field="medicine"/>
     <issn>1018-4813</issn>
     <eissn>1476-5438</eissn>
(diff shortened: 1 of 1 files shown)


european-journal-of-human-genetics.csl
```diff
@@ -15,7 +15,7 @@
     <issn>1018-4813</issn>
     <eissn>1476-5438</eissn>
     <summary>Superscript Vancouver style for some NPG journals, with et-al-min="7" and et-al-use-first="3"</summary>
```
rmzelle: The summary was already correct, but the style itself wasn't. That's ... very peculiar.

rmzelle:
Thanks!

Issue #74, opened 2012-01-01
Changed Harvard 7 Style (de) for Political Science


zasq:
Hello!

I have changed the Harvard 7 Style (de) developed by Julian Onions and Sven Rothe to suit the needs for citing in Harvard style in German texts, especially in Political Science. Mainly, I changed the in-text citation style to [author year: page], avoiding the abbreviation for pages (S.) and changing the delimiters (no comma between author and year). For example: (Asbach 1999: 45) instead of (Asbach, 1999: S. 45) which is not the correct citing style in text for Political Science.

The style has been uploaded to:

git://gist.github.com/1546825.git

It might be interesting for others, as the in-text citation style meets the requirements of the "amerikanische Zitierweise" oder "Harvard Zitiersystem" (Harvard citation style in German) for theses in Political Science/Social Sciences. Is it possible to add it to the zotero style repository?

Thanks in advance!

Best,

Saskia Mestern

mmoole:
aha, thanks to [the hint](https://github.com/citation-style-language/styles/pull/1697#issuecomment-142086163) from rmzelle I took again a look at the harvard7de style .. and since I did a bit of housekeeping lately, I found the source of all this evil :-) 
The harvard7de style was made according to a document dating back to 2002 that can be found at various locations, for example here:
http://qse.ifs.tuwien.ac.at/courses/Usability2/Literaturangaben.pdf
Yes, my professor told us to use this kind of style at that time (that was about 2007, at quite another university)..... 
As that would be to easy, I could not find the original source for this style document... But researching found out:
FHV: Fachhochschule Voralberg (the university name)
BM: could be anything, maybe an abbreviation for a study program... I didn't find something that makes sense.

So I went to their library site, and they even have a more recent document online:
http://www.fhv.at/media/pdf/bibliothek/schreibzentrum/leitfaden-wissenschaftlichen-arbeiten
Here they even talk about Zotero and do no more make 'one of a kind' things like flipping first name and family names of the authors in bibliography. They talk more generally of styles and do no more make such detailed specifications as in the document dating back to 2002. And they even have a style with their name 'Fachhochschule Voralberg' in the styles library at Zotero! :+1: 

So to conclude, I think this all comes down to how much the harvard7de style is used in reality. If there are lots of people using it, I think we should keep it, but maybe rename it to something like 'Harvard Fachhochschule Voralberg (archived)' or similar to express that it is outdated.
If the style isn't used at all, then let's delete it.
[...]

PR #2564, opened 2017-03-13
Delete Harvard7de

adam3smith:
and redirect to FH Voralberg. As discussed here: https://github.com/citation-style-language/styles/issues/74#issuecomment-285992173


adam3smith:
@rmzelle is that the right way to go, though? I don't think FHV is actually the basis for that style. Shouldn't we rather not use a template and move the Harvard7 creator into contributor?

rmzelle:
Well, if either of the two styles need a rewrite in the future, it might make sense to base them on FHV, so the template link might still hold some value. But, yeah, what you propose is possible too. I'm fine with either.

adam3smith:
(fiddlestick; that commit wasn't supposed to go in this branch. I'll sort it out)

PR #2565, opened 2017-03-14
Update scientia-agriculturae-bohemica.csl

libora:
add "collection-editor" to the substitute of the author
Explanation: for document type "Report" is "collection-editor" available only in Zotero and not "editor".

### DIFF
--- a/scientia-agriculturae-bohemica.csl
+++ b/scientia-agriculturae-bohemica.csl
@@ -35,6 +35,7 @@
       <et-al font-style="normal" font-weight="normal"/>
       <substitute>
         <names variable="editor"/>
+        <names variable="collection-editor"/>
         <names variable="translator"/>
         <text variable="title"/>
       </substitute>
@@ -45,6 +46,7 @@
       <name name-as-sort-order="all" sort-separator=" " initialize-with="" delimiter-precedes-last="never" delimiter-precedes-et-al="never"/>
       <substitute>
         <names variable="editor"/>
+        <names variable="collection-editor"/>
         <names variable="translator"/>
         <text variable="title"/>
       </substitute>


rmzelle:
@adam3smith, shall we take this? I guess it can't really hurt, and it would be a workaround for Zotero until https://github.com/citation-style-language/zotero-bits/issues/31 is addressed.

adam3smith:
I'm not convinced this is a good idea. Actual collection editors do exist, of course, and probably _shouldn't_ appear at the beginning of citations. 
Is there no way to add authors to items via Zotero's extra field? It looks like those are just parsed literally, unfortunately?

libora:
Extra field is not the best way due to formatting rules, I think. How should be editor written to the extra field correctly? I see three possible ways:
1. Add "editor" creator role to report citation-style-language/zotero-bits#31 (the best way)
2. Temporary change "series editor" role to "author", update bibliography in the document before its submission (this way is fully possible in this style template, but if I have to change something in the document in the future, e.g. due to comments from reviewers etc, I have to do it again. Will I remember it? I am not sure.)
3. Add "collection-editor" to the style template

adam3smith:
If this were used by Zotero only, I'd have no objection, but of course there are dozens of other projects using CSL now. I'm pretty sure several of them (at a minimum, of course, pandoc), do allow for editors in report and could also see (unlikely but possible) downsides to turning the series editor into a regular editor.

(For the extra field: Zotero can pick up variable from the extra field -- the way it does with DOIs for book chapters, e.g., but I don't think that works to recognize firstname/lastname authors; but if there is a way, I'd suggest that as a workaround)

rmzelle:
Okay, I'll close this, then. @libora, you can of course use keep using this version until Zotero is updated to include the "editor" role for reports.

libora:
@rmzelle OK I create my local version of this style. Is any way to add correctly "editor" to extra field?

rmzelle:
> Is any way to add correctly "editor" to extra field?

According to http://citeproc-js.readthedocs.io/en/latest/csl-json/markup.html#cheater-syntax-for-odd-fields, the syntax for this is:

```
editor: Thompson || Hunter S.
```

(use additional lines for multiple editors)

PR #2566, opened 2017-03-15
Create fungal-biology.csl

damnation333:
via https://forums.zotero.org/discussion/61736/style-request-fungal-biology-elsevier
Added DOI as specified in the guidelines even though they don't appear in final papers (Elsevier journal, although it doesn't follow their styles).
Open papers here: http://www.sciencedirect.com/science/journal/18786146/121/1


rmzelle:
Thanks!

PR #2567, opened 2017-03-15

vl2:
I've made changements on following damnation333, rmzelle and adam3smith corrections.

### DIFF
--- a/international-labour-organization.csl
+++ b/international-labour-organization.csl
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
   <info>
-    <title>International Labour Organization</title>
-    <id>http://www.zotero.org/styles/international-labour-organization</id>
-    <link href="http://www.zotero.org/styles/international-labour-organization" rel="self"/>
+    <title>International Labour Organization test</title>
+    <id>http://www.zotero.org/styles/international-labour-organization-test</id>
+    <link href="http://www.zotero.org/styles/international-labour-organization-test" rel="self"/>
     <link href="http://www.zotero.org/styles/harvard-cardiff-university" rel="template"/>
     <link href="http://labordoc.ilo.org/record/442275?ln=en" rel="documentation"/>
     <link href="http://www.issa.int/Media/Files/Liaison_and_general_ISSA_design/ILO-house-style-manual" rel="documentation"/>
@@ -22,7 +22,7 @@
     <category citation-format="author-date"/>
     <category field="generic-base"/>
     <summary>The Harvard author-date style adapted for ILO house style</summary>
-    <updated>2017-03-14T08:22:32+00:00</updated>
+    <updated>2017-03-21T20:35:42+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="editor">
@@ -75,6 +75,9 @@
       <if type="book" match="any">
         <text variable="title" font-style="italic"/>
       </if>
+      <else-if type="report" match="any">
+        <text variable="title" font-style="italic" prefix=" "/>
+      </else-if>
       <else>
         <text variable="title"/>
       </else>
@@ -147,7 +150,7 @@
           <group prefix=" " delimiter=". " suffix=".">
             <text macro="title"/>
             <text variable="collection-title"/>
-            <text macro="edition"/>
+            <text macro="edition" prefix="(" suffix=")"/>
             <text macro="editor"/>
             <text term="volume" form="short" text-case="capitalize-first"/>
             <text macro="publisher"/>
@@ -194,12 +197,12 @@
             <group suffix="." delimiter=", ">
               <text macro="publisher"/>
             </group>
-            <group>
+            <group delimiter=" ">
               <text term="volume" form="short" text-case="capitalize-first"/>
               <text variable="volume" form="short" suffix=", "/>
               <text macro="pages" suffix="."/>
+              <text variable="URL"/>
             </group>
-            <text variable="URL" prefix=" "/>
           </group>
         </else-if>
         <else-if type="report" match="any">
@@ -210,7 +213,7 @@
             <text macro="edition"/>
             <group prefix="(" suffix=")" delimiter=", ">
               <text variable="publisher-place"/>
-              <text variable="publisher"/>
+              <text value="Institution"/>
             </group>
             <text macro="pages"/>
             <text variable="URL"/>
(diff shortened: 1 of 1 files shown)


international-labour-organization.csl
```diff
@@ -16,16 +16,19 @@
...
-    <updated>2013-01-21T15:37:19+00:00</updated>
+    <updated>2017-03-14T08:22:32+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="editor">
-    <names variable="editor" delimiter=", ">
+    <names variable="editor" delimiter=",">
       <name name-as-sort-order="all" and="text" sort-separator=", " initialize-with=". " delimiter=", "/>
-      <label form="short" prefix=" "/>
+      <label form="short"/>
```
adam3smith: this seems wrong to me? Do you really want <code>Smith, A.ed.</code> without any space?

international-labour-organization.csl
```diff
@@ -110,14 +116,15 @@
...
-      <label variable="page" form="short" suffix=" "/>
+      <label variable="page" form="short" plural="contextual" suffix=" "/>
       <text variable="page"/>
     </group>
   </macro>
   <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true">
     <layout prefix="(" suffix=")" delimiter="; ">
       <group delimiter=" ">
         <text macro="author-short"/>
+        <text variable="abstract"/>
```
adam3smith: the abstract for in-text citations in a Harvard style? That has to be a mistake, no?

international-labour-organization.csl
```diff
@@ -110,14 +116,15 @@
...
         <text macro="year-date"/>
         <group>
```
adam3smith: And no punctuation at all for in-text citations? (Smith 1776 p.5)?

international-labour-organization.csl
```diff
@@ -137,16 +144,20 @@
         <date-part name="year"/>
       </date>
       <choose>
-        <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
+        <if type="book" match="any">
           <group prefix=" " delimiter=". " suffix=".">
-            <text macro="title"/>
+            <text macro="title" font-style="italic"/>
```
adam3smith: no need for the `font-style="italic"` here -- you're already ding that in the macro

international-labour-organization.csl
```diff
@@ -137,16 +144,20 @@
...
+            <text variable="collection-title"/>
             <text macro="edition"/>
             <text macro="editor"/>
+            <text variable="volume" form="short" text-case="capitalize-first"/>
```
adam3smith: this seems odd - while it's valid CSL, I don't think any reference managers offers a short for of the volume variable. Do you mean to include a label here that should be short, i.e. Vol.? You'd have to set that via a label or a term.

international-labour-organization.csl
```diff
@@ -137,16 +144,20 @@
...
             <text macro="publisher"/>
+            <text macro="pages"/>
+            <text variable="number-of-pages"/>
           </group>
         </if>
         <else-if type="paper-conference" match="any">
-          <text macro="title" font-style="italic" prefix=" " suffix=","/>
+          <text variable="title" quotes="false" font-style="normal" font-weight="normal" prefix=" " suffix=","/>
```
adam3smith: much easier to read without all the default settings. You can just remove `quotes="false" font-style="normal" font-weight="normal"`

international-labour-organization.csl
```diff
@@ -158,49 +169,62 @@
...
-          </group>
-        </else-if>
-        <else-if type="webpage">
-          <group prefix=" " suffix=".">
-            <text macro="title" suffix=" [Online]."/>
-            <text macro="edition"/>
-            <text prefix=" " macro="access" suffix="."/>
-          </group>
-        </else-if>
+        <else-if type="thesis"/>
```
adam3smith: I'd recommend against this -- as they currently work, Zotero and Mendeley will actually print error messages when you try to cite a thesis.

international-labour-organization.csl
```diff
@@ -158,49 +169,62 @@
...
             <text macro="editor"/>
             <text variable="container-title" font-style="italic" suffix=","/>
             <text macro="edition"/>
+            <text macro="publisher" prefix=" "/>
             <group suffix="." delimiter=", ">
               <text macro="publisher"/>
-              <text variable="volume"/>
-              <text macro="pages"/>
             </group>
+            <text term="volume" form="short" text-case="capitalize-first"/>
```
adam3smith: I'm thinking this is what you want further up, too. But you'll want to wrap these two into a group. Otherwise the volume term is getting printed even if there is no actual volume, i.e. you'd get `Vol. pp. 123-345`

international-labour-organization.csl
```diff
@@ -158,49 +169,62 @@
...
+            <text variable="volume" form="short" suffix=", "/>
+            <text macro="pages" suffix="."/>
+            <text variable="URL" prefix=" "/>
```
adam3smith: no need for the prefix here. I't already in the delimiter above.

international-labour-organization.csl
```diff
@@ -158,49 +169,62 @@
...
+          </group>
+        </else-if>
+        <else-if type="report" match="any">
+          <group prefix=" ">
```
adam3smith: no need for the group with a single item in it.

international-labour-organization.csl
```diff
@@ -158,49 +169,62 @@
...
+            <text macro="title" font-style="italic"/>
```
adam3smith: set the italics in the macro, not here, for uniformity.

international-labour-organization.csl
```diff
@@ -158,49 +169,62 @@
...
+            <text macro="edition"/>
+            <group prefix="(" suffix=")" delimiter=", ">
+              <text variable="publisher-place"/>
+              <text variable="publisher"/>
+            </group>
+            <text macro="pages"/>
+            <text variable="URL"/>
           </group>
         </else-if>
+        <else-if type="post-weblog" match="any">
```
adam3smith: I'd also put everythign here into a group with `delimiter=" "` and the get rid of the spaces in the suffixes. That's going to be more robust.

vl2:
There are some issues with your submission.
rspec ./spec/styles_spec.rb[728:10] # international-labour-organization: may not have any unused macros

Could someone advice please?

rmzelle:
> There are some issues with your submission.
> rspec ./spec/styles_spec.rb[728:10] # international-labour-organization: may not have any unused macros
> 
> Could someone advice please?

CSL styles in the repository are not allowed to include any macros that aren't used. In this case, there is an "access" macro that isn't used, so it should be deleted.

vl2:
Dear adam3smith,

we are a bit lost and would like some help. I have done most of the modifications asked already, but apparently something is going wrong. Could you please give some more guidance? After all the style passes all the tests and looks OK in CSL editor. Note: we will not need the "thesis" in our style.
Thanks in advance.

adam3smith:
No, I think this just got left behind because of vacation time -- the automated tests all pass. I'll take a look tonight or Rintze might merge before that.

rmzelle:
(I was leaving this for @adam3smith since he had done a review of the style code)

adam3smith:
Sorry for the delay

PR #2568, opened 2017-03-15
Update deutsche-gesellschaft-fur-psychologie.csl

damnation333:
via https://forums.zotero.org/discussion/63141/zitierstil-deutsche-gesellschaft-fuer-psychologie

### DIFF
--- a/deutsche-gesellschaft-fur-psychologie.csl
+++ b/deutsche-gesellschaft-fur-psychologie.csl
@@ -4,9 +4,9 @@
     <title>Deutsche Gesellschaft für Psychologie (German)</title>
     <id>http://www.zotero.org/styles/deutsche-gesellschaft-fur-psychologie</id>
     <link href="http://www.zotero.org/styles/deutsche-gesellschaft-fur-psychologie" rel="self"/>
-    <link href="http://www.psychologie.uni-bonn.de/studium/richtlinien-zur-manuskriptgestaltung" rel="documentation"/>
+    <link href="https://www.psychologie.uni-bonn.de/de/studium/richtlinien-zur-manuskriptgestaltung/view" rel="documentation"/>
     <link href="https://github.com/citation-style-language/styles/pull/432" rel="documentation"/>
-    <link href="http://hub.culturegraph.org/about/BSZ-267173628/html" rel="documentation"/>
+    <link href="https://www.hogrefe.de/shop/richtlinien-zur-manuskriptgestaltung-75884.html" rel="documentation"/>
     <author>
       <name>Daniel Hirsbrunner</name>
       <email>dhirsbrunner@gmx.ch</email>
--- a/deutsche-gesellschaft-fur-psychologie.csl
+++ b/deutsche-gesellschaft-fur-psychologie.csl
@@ -14,9 +14,14 @@
     <contributor>
       <name>Michael Zaugg</name>
     </contributor>
+    <contributor>
+      <name>Patrick O'Brien</name>
+      <email>obrienpat86@gmail.com</email>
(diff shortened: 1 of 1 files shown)


rmzelle:
@damnation333, can you update the "documentation" links? The first one is giving a 404, and the third is to the 3rd edition of the style guide. Maybe replace the last one with https://www.hogrefe.de/shop/richtlinien-zur-manuskriptgestaltung-75884.html?

deutsche-gesellschaft-fur-psychologie.csl
```diff
@@ -83,18 +88,22 @@
...
+                  <date variable="accessed">
+                    <date-part name="day" suffix="."/>
+                    <date-part name="month" form="numeric" suffix="."/>
+                    <date-part name="year"/>
+                  </date>
+                </group>
+                <group delimiter=" ">
+                  <text term="available at" text-case="capitalize-first" suffix=": "/>
+                  <text variable="URL"/>
+                </group>
```
rmzelle: @damnation333, can you move the ": " suffix here onto the group as delimiter?

deutsche-gesellschaft-fur-psychologie.csl
```diff
@@ -270,7 +289,7 @@
             <text macro="container-contributors"/>
             <text macro="secondary-contributors"/>
             <group delimiter=", ">
-              <text variable="container-title" font-style="italic"/>
+              <text variable="container-title" font-style="normal"/>
```
rmzelle: And delete the `font-style="normal"`?

deutsche-gesellschaft-fur-psychologie.csl
```diff
@@ -270,7 +289,7 @@
             <text macro="container-contributors"/>
             <text macro="secondary-contributors"/>
             <group delimiter=", ">
-              <text variable="container-title" font-style="italic"/>
+              <text variable="container-title"/>
```
adam3smith: as per the post on the forums, this has to be wrong, no? They still put containers in italics.

adam3smith:
Looks good to me. Let's see how it fares with users ;)

deutsche-gesellschaft-fur-psychologie.csl
```diff
@@ -121,23 +130,27 @@
...
+              <if variable="event" match="none">
+                <text variable="genre" suffix=", "/>
+              </if>
+            </choose>
+            <group delimiter=": ">
+              <text variable="publisher-place"/>
+              <text variable="publisher"/>
+            </group>
+          </else>
         </choose>
```
rmzelle: @damnation333, this is a little odd, by the way. You always print the publisher-place/publisher information, so it should ideally be moved out of the `<choose/>`. The `type="report thesis"` conditional can then be used to directly control the printing of the "genre" variable, e.g.:

```xml
            <choose>
              <if type="report thesis" variable="event" match="none">
                <text variable="genre" suffix=", "/>
              </if>
            </choose>
```

damnation333: I see. Thanks. Not sure if that was me or from the template, but I'll look out for such odd code constructs from now on.

adam3smith: I think this may well have been the template. In quite a number of styles the publisher/place order is reversed for reports and theses (don't ask me why), and for those cases, the current structure would make sense, of course.

bwiernik: In this case, this construct is due to APA's differing rules for genre/medium for reports and theses compared to other types. This choose is to avoid printing genre twice for these two types.

PR #2569, opened 2017-03-15
Update biomed-central.csl

damnation333:
Fix error with chapter as mentioned here: https://forums.zotero.org/discussion/64762/style-error-biomed-central
Please hold for merge in case user requests some other changes.

### DIFF
--- a/biomed-central.csl
+++ b/biomed-central.csl
@@ -132,9 +132,11 @@
               <group delimiter=". ">
                 <group delimiter=": ">
                   <text term="in" text-case="capitalize-first"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
I need to test this, but I think 

```
<group delimiter=": ">
<group delimiter=". ">                 
<text macro="editor"/>
<text macro="container-title"/>
</group>
</group>
```

will work better, i.e. including for chapters that don't have editors, which we see occasionally in the biosciences.

rmzelle:
Looks good. Thanks both. (no need to wait, we can always create another PR)

PR #2571, opened 2017-03-16
Update politische-vierteljahresschrift.csl

damnation333:
via https://forums.zotero.org/discussion/64799/style-update-request-politische-vierteljahresschrift

How do I get it to show only "u." instead of "und" in the in-text citations?

### DIFF
--- a/politische-vierteljahresschrift.csl
+++ b/politische-vierteljahresschrift.csl
@@ -189,4 +189,4 @@
       <text prefix=" " macro="access"/>
     </layout>
   </bibliography>
-</style> 
+</style>
--- a/politische-vierteljahresschrift.csl
+++ b/politische-vierteljahresschrift.csl
@@ -166,7 +166,7 @@
         </else-if>
         <else-if type="thesis">
           <group prefix=" " suffix="." delimiter=". ">
-            <text macro="title"/>
+            <text variable="title"/>
             <text variable="genre"/>
             <text macro="publisher"/>
           </group>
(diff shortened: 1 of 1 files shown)


rmzelle:
> How do I get it to show only "u." instead of "und" in the in-text citations?

I'm not sure if CSL processors support the use of `and="symbol"` while defining a "symbol" variant of the "and" term as e.g. "u.". The spec suggests it should be hardcoded, although what I propose should validate and might work.

We should probably change the spec to just allow the "and" attribute to switch between the "long" and "symbol" forms of the "and" term (and perhaps the "short" form as well, for this case, either by adding a third attribute value, like "text-short", or by replacing the "and" attribute by an "and" child element of cs:names (with a "form" attribute to select the desired term variant), but that might be a bit awkward). (this has come up before in personal email between @adam3smith and me around 5/13/15, regarding item-locale specific settings)

politische-vierteljahresschrift.csl
```diff
@@ -47,25 +56,30 @@
...
+            </group>
+          </if>
+        </choose>
+      </else>
     </choose>
   </macro>
-  <macro name="title">
-    <text variable="title"/>
-  </macro>
+  <macro name="title"/>
```
rmzelle: @damnation333, empty macros make @csl-bot sad.

adam3smith:
no, sorry, citeproc-js behaves according to spec. You can't override the ampersand for and="symbol"

rmzelle:
Remember if we have a ticket for this? (I couldn't find one in "schema" or "documentation", and couldn't dig up anything in our email conversation)

adam3smith:
no, not that I can remember

rmzelle:
Thanks both!

> Remember if we have a ticket for this?

I created a ticket at https://github.com/citation-style-language/schema/issues/137.

PR #2572, opened 2017-03-17
Create alcoholism-clinical-and-experimental-research.csl

damnation333:
via https://forums.zotero.org/discussion/64872/style-request-harvard-style-for-referencing


adam3smith:
Thanks!

PR #2573, opened 2017-03-17

cookie365:
The date when the item has been accessed should always be displayed.

### DIFF
--- a/fachhochschule-vorarlberg-author-date.csl
+++ b/fachhochschule-vorarlberg-author-date.csl
@@ -38,7 +38,7 @@
       <term name="composer" form="short">Komp.</term>
       <term name="composer" form="long">Komponist</term>
       <term name="accessed">Zugriff am</term>
-	  <term name="available at">Online im Internet</term>
+      <term name="available at">Online im Internet</term>
     </terms>
   </locale>
   <locale xml:lang="en">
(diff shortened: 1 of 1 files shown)


fachhochschule-vorarlberg-author-date.csl
```diff
@@ -165,25 +165,19 @@
...
-              </date>
+		<if variable="URL">
+			<group prefix=" " suffix=": ">
+			  <text value="Online im Internet"/>
             </group>
-          </if>
-        </choose>
+			<text variable="URL"/>
+			<group prefix=" (" delimiter=" " suffix=").">
+			  <text value="Zugriff am:"/>
```
damnation333: you can use the variable "accessed" instead of the hardcoded text. See the locale section for reference (line ~40)

rmzelle:
@cookie365, can you point to the documentation that supports this change? E.g. https://ilias.fhv.at/goto_ilias_fhv_at_file_272243_download.html doesn't seem to required the accessed date for all item types.

rmzelle:
(I also reindented the style with 2-spaces per level instead of tabs, with http://formatter.citationstyles.org/)

cookie365:
@rmzelle because of the required accessed date, i talked with them and it is very important that the url or doi is displayed in every type, if it is set

fachhochschule-vorarlberg-author-date.csl
```diff
@@ -92,54 +96,39 @@
...
-        <choose>
-          <if type="chapter">
-            <names variable="editor" delimiter="; ">
-              <name name-as-sort-order="all" sort-separator=", " delimiter="; " delimiter-precedes-last="never" form="long" suffix=""/>
-              <label form="short" prefix=" (" suffix=".): "/>
-            </names>
-          </if>
-        </choose>
-      </if>
+      <if variable="author"></if>
```
rmzelle: @cookie365, I think you can remove the entire `<choose/>` here to test for the presence of the "author" variable. If the "editor" variable has already been used through `<substitute/>` in the "author" macro, it should be automatically suppressed in the rest of the style. (see http://docs.citationstyles.org/en/stable/specification.html#substitute for background)

fachhochschule-vorarlberg-author-date.csl
```diff
@@ -92,54 +96,39 @@
...
       <else>
         <names variable="editor" delimiter="; ">
           <name name-as-sort-order="all" sort-separator=", " delimiter="; " delimiter-precedes-last="never" form="long" suffix=""/>
-          <label form="short" prefix=" (" suffix=".) "/>
+          <label form="short" prefix=" (" suffix=".): "/>
```
rmzelle: @cookie365, it's better to move the ": " in the suffix up to the `<names/>` element.

fachhochschule-vorarlberg-author-date.csl
```diff
@@ -151,60 +140,58 @@
...
+          <text term="available at"/>
+        </group>
+        <text variable="DOI"/>
       </if>
+      <else-if variable="URL">
+        <group suffix=": URL: ">
+          <text term="available at"/>
+        </group>
+        <text variable="URL"/>
+      </else-if>
```
rmzelle: @cookie365, groups with just a single child element are redundant. It's better to e.g. change 

```xml
      <if variable="DOI">
        <group suffix=": DOI: ">
          <text term="available at"/>
        </group>
        <text variable="DOI"/>
      </if>
```

to

```xml
      <if variable="DOI">
        <group delimiter=": ">
          <text term="available at"/>
          <text value="DOI"/>
          <text variable="DOI"/>
        </group>
      </if>
```

rmzelle:
Thanks!

PR #2575, opened 2017-03-19
Update politische-vierteljahresschrift.csl

damnation333:
Minor changes requested here: https://forums.zotero.org/discussion/comment/272308/#Comment_272308

I was nder the assumption CSL removes the dot suffix if the title if it has a question mark. Somehow this is giving me a "?." now for the Mares chapter.

### DIFF
--- a/politische-vierteljahresschrift.csl
+++ b/politische-vierteljahresschrift.csl
@@ -26,8 +26,8 @@
     </terms>
   </locale>
   <macro name="editor">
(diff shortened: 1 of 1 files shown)


adam3smith:
The reason the suppressed period didn't work was the additional (and unnecessary) styling on the title element. citeproc-js applies these things in turn, and if there are two many, the punctuation suppression fails. The alternative would have been to use a group with delimiters.

PR #2570, opened 2017-03-16
Update and rename dependent/asian-pacific-journal-of-tropical-disease…

damnation333:
….csl to asian-pacific-journal-of-tropical-disease.csl

via https://forums.zotero.org/discussion/64839/asian-pacific-journal-of-tropical-disease-style-is-not-correct
Style guide: https://www.elsevier.com/journals/asian-pacific-journal-of-tropical-medicine/1995-7645/guide-for-authors
Open access examples: http://www.sciencedirect.com/science/journal/19957645/9/1

PR #2579, opened 2017-03-22
Update asian-pacific-journal-of-tropical-disease.csl

damnation333:
Dependent parent made and the first dependent adapted as discussed here: https://forums.zotero.org/discussion/comment/272491/#Comment_272491

Can y'all check if it's correct and then I'll make the other 7 dependent styles.


rmzelle:
(replaced by #2570)

PR #2580, opened 2017-03-22

Yusufras:
macro name author short
`<et-al font-style="italic"/>`


rmzelle:
Thanks!

PR #2581, opened 2017-03-23
Update university-for-the-creative-arts-figures.csl

ibadger:
Hi

I tested the merged file this morning and noticed that when a URL was present the reference would put:

(Accessed on on 25 February 2013)

I think that my mistake matched what I'd done with s.d. earlier, and I had put the 'on' in manually as text which was leading to duplication. I have now removed the suffix.

Sorry for not noticing this earlier.

Ian

### DIFF
--- a/university-for-the-creative-arts-figures.csl
+++ b/university-for-the-creative-arts-figures.csl
@@ -21,6 +21,26 @@
       <term name="no date" form="short">s.d.</term>
     </terms>
   </locale>
+  <macro name="accessed">
+    <choose>
+      <if match="all" variable="URL">
+        <group delimiter=" ">
+          <group delimiter=": ">
+            <text term="at" text-case="capitalize-first"/>
+            <text variable="URL"/>
+          </group>
+          <group prefix="(" suffix=")" delimiter=" ">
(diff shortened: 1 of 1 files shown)


university-for-the-creative-arts-figures.csl
```diff
@@ -83,7 +83,7 @@
               <choose>
                 <if match="any" variable="URL">
                   <text variable="URL" prefix="At: "/>
-                  <text term="accessed" text-case="capitalize-first" prefix="(" suffix=" on"/>
+                  <text term="accessed" plural="true" text-case="capitalize-first" prefix="(" suffix=" "/>
```
rmzelle: @ibadger, I take it you added the `plural="true"` by accident? (there is no plural form of the "accessed" term)

rmzelle:
I simplified the style a little by putting the duplicated URL/accessed code in a reusable macro.

rmzelle:
Thanks! (the double "on on" was my fault, by the way)

ibadger:
Fantastic, thanks for your help. I have a much better understanding of Macros now.

I have another style, the non figures one in the repository, I will start having a look at that soon as I think  the links to the referencing guide  might be out of date

PR #2583, opened 2017-03-24
Create vita-latina.csl

Oxil-2:
First submission of the stylesheet for the French journal "Vita Latina"

### DIFF
--- a/vita-latina.csl
+++ b/vita-latina.csl
@@ -195,24 +195,6 @@ Remarques :
 	</choose>
   </macro>
   
-  <macro name="edition">
-    <choose>
-      <if type="book chapter paper-conference report" match="any">
-        <choose>
-          <if is-numeric="edition">
-            <group delimiter=" ">
-              <number variable="edition" form="ordinal"/>
-              <text term="edition" form="short"/>
-            </group>
-          </if>
-          <else>
-            <text variable="edition" text-case="capitalize-first" suffix="."/>
-          </else>
-        </choose>
-      </if>
-    </choose>
-  </macro>
-  
   <macro name="lieu-edition">
     <choose>
       <if type="book chapter thesis report paper-conference" match="any">
--- a/vita-latina.csl
+++ b/vita-latina.csl
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" default-locale="fr-FR" >
-
-<!--
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" default-locale="fr-FR">
+  <!--
 
 Feuille de style CSL 
 à destination du logiciel Zotero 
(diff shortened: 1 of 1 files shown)


vita-latina.csl
```diff
@@ -0,0 +1,305 @@
...
+  </citation>
+  <!-- INSTRUCTIONS DE FORMATAGE DE LA BIBLIOGRAPHIE -->
+  <bibliography et-al-min="11" et-al-use-first="7" subsequent-author-substitute="&#8212;" entry-spacing="0">
+    <sort>
+      <key macro="contributeurs-tri"/>
+      <key variable="issued"/>
+    </sort>
+    <layout suffix=".">
+      <group delimiter=" ">
+        <text macro="auteurs" prefix="  "/>
```
rmzelle: (@adam3smith, have you seen first-line indents before in other styles? Might be better to have an official "first-line-indent" toggle. From the examples:

image
)

rmzelle: (Those are `&#8195;` em spaces, by the way, which the formatter currently doesn't escape; http://www.codetable.net/decimal/8195)

vita-latina.csl
```diff
@@ -0,0 +1,305 @@
...
+Remarques :
+	- pour les thèses, Zotero ne permet pas de spécifier le directeur ; la feuille de style laisse donc cette indication à compléter manuellement, entre crochets ;
+	- à ce stade, le langage CSL ne permet pas de définir certains éléments de formatage, tels que la police, la taille de la police ou l'alignement de texte (ainsi, forcer la bibliographie à être justifiée). Ces éléments doivent être mis en place manuellement. 
+ -->
+  <!-- INFORMATIONS RELATIVES A LA FEUILLE DE STYLE -->
+  <info>
+    <title>Vita Latina (French)</title>
+    <title-short>VL</title-short>
+    <id>http://www.zotero.org/styles/vita-latina</id>
+    <link href="http://www.zotero.org/styles/vita-latina" rel="self"/>
```
rmzelle: @Oxil-2, before we accept this, could you add a "documentation" link that points to instructions to authors? (probably to http://vitalatina.upv.univ-montp3.fr/page-3/, I think). See https://github.com/citation-style-language/styles/wiki/Style-Requirements#10---documentation-link

And, if you based this style on an existing CSL style from the repository, please also add a "template" link. See https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link

Finally, can you add the electronic ISSN as well? Looks like that's "1783-1830". See https://github.com/citation-style-language/styles/wiki/Style-Requirements#9---issn-and-eissn

rmzelle: I reindented your submission, by the way, with http://formatter.citationstyles.org/, according to our standard format. That removed some empty lines, among other things.

vita-latina.csl
adam3smith: I think I've seen them once before, but it's quite rare, obviously. Not sure this makes sense as a separate attribute, given that a perfectly fine solution exists (I think tab would also work)

vita-latina.csl
Oxil-2: Done, thanks for the reindentation!

rmzelle:
Thanks!

PR #2584, opened 2017-03-24
Create vita-latina-auteurs-anciens.csl

Oxil-2:
This style goes along with vita-latina.csl (#2583). It is specific to the citation of ancient authors in the French journal "Vita Latina".


adam3smith:
What's the vision for how people would use this? I assume you normally have both classical authors and contemporary works in an article, but you can only use one citation style.

Oxil-2:
This stylesheet is only meant to edit the final bibliography for classical authors (which is separated from the main bibliography and with a different formatting). I've written a small guide to explain this: People would use the main style (vita-latina) in Word, and just past the specific ancient authors bibliography from Zotero, formatted with this specific style. I hope this is somewhat clearer (obviously English is not my mother tongue ^^).

rmzelle:
Can you give some examples of how the regular and classical author citations differ? Do you use the same item types for both?

Oxil-2:
Hi,

yes, the problem is that we use the same item types for both classical authors and modern authors (academic works), e.g. a book, but the bibliography (not the citations) differs largely. The font of the classical author names stay regular (lowercase), while the contemporary authors are put in small-caps, the order isn't the same, etc. Attached is an example, where you can see the difference between the ancient works bibliography (first part of the bibliography page 2) and the secondary literature (second part).
I don't see any other solution for this problem than using two different stylesheets, and I've written, as said, a small guide to explain how to use this second style (this document will be put on the journal homepage). If there's a problem for you on this point (which I do not see), let just put the main style in the repository, and let the second one available on the website of the journal!
[FichierTest.pdf](https://github.com/citation-style-language/styles/files/869818/FichierTest.pdf)

rmzelle:
> Attached is an example, where you can see the difference between the ancient works bibliography (first part of the bibliography page 2) and the secondary literature (second part).

Just to confirm: is it typical that you group all the ancient works together, and all the current works together?

> I don't see any other solution for this problem than using two different stylesheets

There doesn't appear to be any clear different between current and ancient works, is there (i.e. some metadata fields that have a value for one type, but not for the other)? If that's the case, yes, I don't see how a single CSL style could discriminate between the two types of works.

Oxil-2:
Well it's typical for this journal (but generally speaking in Classics, we usually do this, yes). No, there isn't any metadata or fields-based difference between current and ancient works in Zotero, so I don't see neither how a single CSL style could distinguish between the two types of books. That's why I've created a general CSL style ("vita-latina"), which I've submitted 3 days ago, and which is meant to be used in Word/OpenOffice as the document style, and this secondary style ("vita-latina-auteurs-anciens") just for the bibliography containing the ancient authors. As I've written in the small guide, which will be put on the journal website, this secondary style is to be imported in Zotero, and used to edit the bibliography **from there** (and then copy/paste it into the word document).

Is it clearer now?

bwiernik:
Do classic works typically have Archive and Archive Location fields that could be used to distinguish them from contemporary items?

@rmzelle 
Another option that involves changes to CSL would be add something like a Classic item type. @fbennett's CSL-m added such an item type to accommodate this sort of situation (http://fbennett.github.io/z2csl/diffMap.html). It has essentially the same fields as Book.

adam3smith:
I'm fine taking this. Since most classical works are traditionally published by modern publishers, I do think we'll need an item type for this. I actually thought we had a ticket open for that, but it definitely has come up repeatedly on Zotero.

rmzelle:
Yes, I'm fine with taking this too, and agree we probably need a dedicated item type. I'll review the submissions in more detail later this week.

rmzelle:
> I actually thought we had a ticket open for that, but it definitely has come up repeatedly on Zotero.

(there is an old closed ticket at https://github.com/citation-style-language/schema/issues/22; I'll reopen it)

Oxil-2:
Ok, thanks for reopening this ticket. It would be actually a great possibility to be able to distinguish between the ancient works and the contemporary (academic) works. I just want to clear something up: ancient authors are edited in normal, contemporary books, it doesn't have anything to do with archive. So what we need is to create a item type parallel to "normal" books.

rmzelle:
> So what we need is to create a item type parallel to "normal" books.

And would a classical book item type cover all of your cases? Or do people often cite other types of classical works that can't be described as books?

Oxil-2:
No, a classical book item type would be perfect, but it has to contain a field to be able to distinguish in the CSL file between "normal" books and ancient authors. 
Classical works should contain in the reference at least the ancient author's name, the editor and the translator (which can be the same).

bwiernik:
It would be a completely separate item type in a future version of CSL, so a style like yours would simply specify different formatting for type="classic" versus type="book". Are there other specific fields that a Classic type would need that Book doesn't currently have?

Oxil-2:
I understand.
Another important modification for this "classic" type would be to add a field "book", which would be different from the field "volume". "Volume" would the refer to the modern collection, "book" to the ancient division of the work.
An example: Tacitus wrote a work called "The Annals", and splitted it in 16 books. But modern edition (for example the french one) usually publish the _Annals_ in 4 volumes, the first one containing books 1 to 3, the second one books 4 to 6, etc.

rmzelle:
> Tacitus wrote a work called "The Annals", and splitted it in 16 books. But modern edition (for example the french one) usually publish the Annals in 4 volumes, the first one containing books 1 to 3, the second one books 4 to 6, etc.

Do you know if people usually cite "The Annals" as a whole (one multi-volume work) or if they cite individual books? (I have this question both for the classic work and modern editions)

Oxil-2:
In the first part of the bibliography (the one citing the "sources", including the ancient authors), they usually refer to the different volumes (and not books, cf. my last post). If one works on the first part of the _Annals_ (ie on the first roman emperors), he will just cite Tacitus, _Annals_, vol. I books I-III.

rmzelle:
Thanks!

rmzelle:
> In the first part of the bibliography (the one citing the "sources", including the ancient authors), they usually refer to the different volumes (and not books, cf. my last post). If one works on the first part of the Annals  (ie on the first roman emperors), he will just cite Tacitus, Annals, vol. I books I-III.

I'm just wondering how well a "book" item type would fit such citations (Zotero and CSL currently don't really handle multi-volume works very well).

adam3smith:
yeah, we want to be thorough in thinking through the possible cases here. You also have classic works where multiple works with different titles (e.g. Plato dialogs) are in the same physical book, which in turn may be part of a multi-volume work (i.e. Plato's collected works). Pretty sure there are sample citations on the Zotero forums.
I'm pretty optimistic we can find the right fields for the item type, but I don't think we should necessarily just think about it in terms of a book item.

rmzelle:
> I'm pretty optimistic we can find the right fields for the item type, but I don't think we should necessarily just think about it in terms of a book item.

Yes, the question is more whether a single item type would suffice. Anyway, more discussion on this topic can take place in the issue I linked to above.

Oxil-2:
You're both perfectly right, one should be careful with those questions; i just want to point out that the works edited in a single book (eg Platos dialogs or Ciceros orations) can differ from one edition to the other, as can differ the grouping of the books of an historian like Tacitus (some editions group together books IV-V, and then VI to XI for example)

PR #2586, opened 2017-03-26
Create american-journal-of-primatology.csl

damnation333:
via https://forums.zotero.org/discussion/63979/style-error-american-journal-of-primatology
I did not review this due to lack of access to the journal. If somebody could check that would be great. Thanks :)


damnation333:
ah dammit. I should have just edited the independent file. Sorry.
Can somebody now tell me how I add a commit to the here PR to also delete the americaan-journal-of-primatology independent style?

rmzelle:
> I did not review this due to lack of access to the journal. If somebody could check that would be great. Thanks :)

You should be able to access http://onlinelibrary.wiley.com/doi/10.1002/ajp.v79.1/issuetoc (via "Get Sample Copy" in the side bar of http://onlinelibrary.wiley.com/journal/10.1002/(ISSN)1098-2345).

> Can somebody now tell me how I add a commit to the PR here to also delete the american-journal-of-primatology independent style?

Go to the branch from which your pull request originates (https://github.com/damnation333/styles/tree/patch-1), look up the style you want to change (https://github.com/damnation333/styles/blob/patch-1/american-journal-of-primatology.csl), and click the trash icon at the top of the file.

damnation333:
Thanks Rintze.
That ain't APA at all, although that's what they say in the guidelines or show in their examples.
The actual papers follow this: [Campbell & Pedersen, 2007] and Campbell JL, Pedersen OK. 2007. The varieties of capitalism and hybrid success. Comparative Political Studies 40:307–332. (florida-entomologist is a great match). I'll work on that in the next days.

rmzelle:
> That ain't APA at all, although that's what they say in the guidelines or show in their examples.

(the sample issue is from January 2017, so it's probably still current)

damnation333:
can be closed. the independent style on the repo, made by adamsmith, is correct. Just Wiley's guidelines are competely different and wrong.

adam3smith:
See my comment on the Zotero thread. I'm pretty sure they changed this

adam3smith:
http://onlinelibrary.wiley.com/doi/10.1002/ajp.22595/epdf
is open access and follows the new format

adam3smith:
@damnation333 trying to follow this -- is the dependent or the independent style correct? Obviously if we can do this via the dependent, that'd be nice.

damnation333:
@adam3smith
The independent style code in here is the right one to go with. It needs its own independent style and is not an APA dependent as initially thought. Thanks

adam3smith:
I haven't looked at this in detail, but wouldn't it be easier to just copy over APA style and change the couple of things that are different?

damnation333:
i think we've gone full circle on this now. Just made it a dependent of APA.

PR #2587, opened 2017-03-26
Make geology.csl into dependent style.

damnation333:
As discussed here, we actually got the "The Geological Society of America" style in the repo with a few dependents already.
https://forums.zotero.org/discussion/comment/270234/#Comment_270234

We also got environmental-and-engineering-geoscience.csl in the repo as an independent style (from 2014). Can I make that into a dependent, too? Then we would have all of the GSA journals as dependents.


rmzelle:
> We also got environmental-and-engineering-geoscience.csl in the repo as an independent style (from 2014). Can I make that into a dependent, too?

Sure, if it uses the GSA style format. It was actually last touched in 2014, and its history predates our switch to GitHub in 2011, and nobody has made any specific edits to it since 2011, so it might very well be regular GSA.

By the way, did you double-check whether the disambiguation rules are the same between Geology and the other GSA styles? Charles Parnot (ex-Papers) specifically made it an dependent because of differing disambiguation rules: https://github.com/citation-style-language/styles/commit/8802efea27233617d4a86c0a09fab48687844482

dependent/geology.csl
```diff
@@ -0,0 +1,15 @@
...
+    <link href="http://www.zotero.org/styles/geology" rel="self"/>
+    <link href="http://www.zotero.org/styles/the-geological-society-of-america" rel="independent-parent"/>
+    <link href="http://geosociety.org/GSA/Publications/Journals/Geology/GSA/Pubs/geology/home.aspx" rel="documentation"/>
+    <category citation-format="author-date"/>
+    <category field="geology"/>
+    <issn>0091-7613</issn>
+    <eissn>1943-2682</eissn>
+    <updated>2017-03-26T22:06:38+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
```
rmzelle: @damnation333, missing `</style>` end tag here. The other two errors are complaints that the "template" link in two other styles are now orphaned (you can change those to point to "http://www.zotero.org/styles/the-geological-society-of-america" instead).

damnation333:
@rmzelle 
Re disambiguation:
there is nothing in the style guideline and actually both Charles' Geology and the GSA style just use disambiaguate-add-year-suffix="true".

adam3smith:
@damnation333 if you could change the template link for `palaios` and `terra-nova` in your patch-2 branch, I think we can merge this

adam3smith:
Thanks!

PR #2588, opened 2017-03-27
Update universita-pontificia-salesiana-it.csl

xavi-mat:
- Minor bug solved with parenthesis
- added post-weblog


rmzelle:
Thanks!

PR #2589, opened 2017-03-27
Update universita-pontificia-salesiana.csl

xavi-mat:
- Solved minor bug with parenthesis
- added post-weblog type


adam3smith:
Thanks!

PR #2591, opened 2017-03-27
Update karger-journals.csl

damnation333:
Found a mistake just by chance.
Open access article here: http://www.karger.com/Article/FullText/461586


adam3smith:
thanks!

PR #2592, opened 2017-03-28
Update quaternary-international.csl

damnation333:
As discussed here: https://forums.zotero.org/discussion/64974/style-error-quaternary-international#latest
I made this independent, changed the in-text sorting to chronoligcal and the In: bit for chapters.
@adamsmith Can you review this if it needs anything else changed?

Style guidelines: https://www.elsevier.com/journals/quaternary-international/1040-6182/guide-for-authors


quaternary-international.csl
```diff
@@ -0,0 +1,221 @@
...
+  </macro>
+  <macro name="author-short">
+    <names variable="author">
+      <name form="short" and="text" delimiter=", " initialize-with=". "/>
+      <substitute>
+        <names variable="editor"/>
+        <names variable="translator"/>
+        <choose>
+          <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
+            <text variable="title" form="short" font-style="italic"/>
```
adam3smith: since we're not using italics and quotes anywhere else in the style, I think we can just do `<text variable="title" form="short"/>` here instead of this complicated `choose`, no?

adam3smith:
One question/comment. Everything else lgtm.

adam3smith:
Thanks!

PR #2593, opened 2017-03-28
Update Royal Society journals

mdoube:


adam3smith:
Could you say more about the new ACS style?

mdoube:
Hang on, some stuff just got included in this PR that shouldn't be there...

adam3smith:
(we generally don't like new PRs to fix things, but I won't begrudge you throwing the towel on the rebase here.

mdoube:
New way of working here for me, getting used to the workflow, sorry. Should be all good now.

proceedings-of-the-royal-society-b.csl
```diff
@@ -33,7 +33,12 @@
   </info>
   <locale>
     <terms>
+      <term name="accessed">accessed on</term>
       <term name="edition" form="short">edn</term>
+      <term name="page">
```
adam3smith: you don't need to add these terms; they're the default ones

mdoube: OK, deleted that term definition

adam3smith:
check out if that change makes sense: only labels are context dependent; terms are always singular.
Otherwise we're good to merge I think.

mdoube:
Sorry - that last commit breaks the change. Point is to use pp. when there's a multi-page citation and p. when it's on a single page. This change was a literal copy from http://docs.citationstyles.org/en/stable/specification.html. If it's meant to be in a different format, please show me what to do.

adam3smith:
Do you mean my last commit or yours? Because mine does exactly what you want and it's using the code from http://docs.citationstyles.org/en/stable/specification.html#label (with form="short" added)

mdoube:
Right - seems like two different ways to achieve the same effect. Yours works fine.

adam3smith:
I don't think your version could have ever made the automated p/pp distinction, though. That requires the use of `label` not `term` as per the specs linked above.

mdoube:
You're right. It only fixed one case :/

PR #2595, opened 2017-03-28
Create social-cognitive-and-affective-neuroscience.csl

iobergfeld:
When preparing a manuscript for Social Cognitive and Affective Neuroscience using Mendeley, I noticed no style file for this journal was available. I created one and used it to create references and a bibliography in my manuscript. As far as I can see, no errors occurred. 

Since I already created the csl file, it could be useful for other users, so they don't have to create one themselves.

### DIFF
--- a/social-cognitive-and-affective-neuroscience.csl
+++ b/social-cognitive-and-affective-neuroscience.csl
@@ -1,0 +1,77 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" default-locale="en-US" name-delimiter=";" and="text" et-al-min="3" et-al-use-first="1" name-as-sort-order="first" demote-non-dropping-particle="never" xmlns="http://purl.org/net/xbiblio/csl">
+  <info>
+    <title>Social Cognitive and Affective Neuroscience</title>
+    <title-short>SCAN</title-short>
+    <id>http://www.zotero.org/styles/social-cognitive-and-affective-neuroscience</id>
+    <link href="http://www.zotero.org/styles/social-cognitive-and-affective-neuroscience" rel="self"/>
+    <link href="https://academic.oup.com/scan/pages/General_Instructions" rel="documentation"/>
+    <issn>1749-5016</issn>
+    <eissn>1749-5024</eissn>
+    <author>
+      <name>Isidoor Bergfeld</name>
+      <email>i.o.bergfeld@gmail.com</email>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thank you!
We generally recommend starting styles from existing (ideally high quality) ones. Any reason you didn't here? There are a lot of redundancies in the code here, no use of macros, no use of groups, which makes the style quite fragile when any expected field is missing.

iobergfeld:
Dear Sebastian,

Thank you for your reply!

I actually started out with an existing 'author-date' style, but I got
errors when adapting them (sometimes when I really couldn't see why). Since
I needed the style on short notice, I decided to start from scratch (turned
out to be faster that way). I realize my 'coding' (as far as you can call
it that) is very sluggish. But given the style didn't exist yet, I thought
it would be better than to have nothing.

The journal is in the field of neuroscience, in which generally only
referrals to journal articles, books or chapters are made. This is the
reason I specified only these here. As a matter of fact, these are the only
ones specified by the journal instructions itself. I used this style to
prepare my manuscript and it turned out to work without errors.

When expected fields are missing, I personally prefer the reference to
'fail' in my document instead of it being replaced with another field
instead. If fields are missing, usually (if not always) the reference in
the reference manager needs updating in my experience. It is not something
that needs to be fixed from the style file (I actually really dislike this).

I understand if you think the style file to be of insufficient quality. If
so, hopefully someone else can create one of better quality in the near
future!

Kind regards,

Isidoor

2017-03-29 3:23 GMT+02:00 Sebastian Karcher <notifications@github.com>:

adam3smith:
@rmzelle -- what's your take here? The style is correct for the most common item types and I'd guess people would appreciate that, but it does need some clean-up,  and even then it's very fragile. E.g. for books w/o a publisher, references end on a colon and outside the 4 defined item types the current version doesn't print anything.

rmzelle:
@POBrien333, can I interest you in creating a style for Social Cognitive and Affective Neuroscience? Per @adam3smith comments above, the version in this pull request has some structural issues, so it would be better to base it on something else.

Assuming you have the Mares and Pedersen items stored in Mendeley, you should be able to use this style to create the formatted references to search by example.

rmzelle:
@iobergfeld, we created (https://github.com/citation-style-language/styles/pull/2814) a slightly more robust style for this journal by modifying an existing style from our repository. Feedback on that style is always welcome.

PR #2596, opened 2017-03-28
Update science-and-technology-for-the-built-environment.csl.

mitchpaulus:
Fix subsequent inline citation problem.
Fix author name ordering.

### DIFF
--- a/science-and-technology-for-the-built-environment.csl
+++ b/science-and-technology-for-the-built-environment.csl
@@ -3,7 +3,7 @@
   <info>
     <title>Science and Technology for the Built Environment</title>
     <title-short>STBE</title-short>
-    <id>http://csl.mendeley.com/styles/213838531/science-and-technology-for-the-built-environment-2</id>
+    <id>http://www.zotero.org/styles/science-and-technology-for-the-built-environment</id>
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks. I assume you based this on published articles? The instructions to authors only say:

> All authors are listed in a Reference list entry (no “et al.”).

but don't say anything about the use of "et al." in in-text citations.

rmzelle:
I was able to confirm the `et-al-min="3" et-al-use-first="1"` setting for in-text citations in http://www.tandfonline.com/doi/pdf/10.1080/23744731.2016.1208538, by the way. But according to the instructions, the style shouldn't use "et al." at all in the bibliography, so we'll have to change that as well.

rmzelle:
@mitchpaulus, I'll merge this once you give the go-ahead.

adam3smith:
(I was just looking over this, too; I think it's good to go)

mitchpaulus:
Thank you all for your help! Obviously, the journal does not have a fully specified format, and I did my best to reverse engineer out what they prefer. 

You can go ahead and merge in.

adam3smith:
Thanks!

PR #2597, opened 2017-03-29
Update aerosol-science-and-technology.csl

damnation333:
Fixes made as requested here: https://forums.zotero.org/discussion/comment/272932/#Comment_272932

Something odd is happening on the CSL editor. 
Campbell, J.L. and Pedersen, O.K. (2007). The varieties of capitalism and hybrid success. Comp. Polit. Stud., 
40(3):307–332.
Borges, J.L. (1999). Selected non-fictions. Viking, New 
York.
There is a new line for some citations. Any clue why?

### DIFF
--- a/aerosol-science-and-technology.csl
+++ b/aerosol-science-and-technology.csl
@@ -13,7 +13,7 @@
     <category citation-format="author-date"/>
     <category field="chemistry"/>
     <issn>0278-6826</issn>
-    <updated>2017-03-27T14:38:28+00:00</updated>
+    <updated>2017-03-29T13:20:23+00:00</updated>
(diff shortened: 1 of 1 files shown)


adam3smith:
We've come across that newline problem in the editor before. I believe @zuphilip helped fix it at the time -- which browser is this on?

adam3smith:
Thanks!

damnation333:
I'm editing in Chrome (although I use FF for all my browsing).
As long as I know it's just a bug, I don't mind.

adam3smith:
yeah, definitely nothing to do with the style(s), so nothing to worry about for you.

zuphilip:
The fix was for firefox only, because it worked okay with Chrome, but maybe they changed something now. See here for the issue: https://github.com/citation-style-language/csl-editor/issues/161

PR #2599, opened 2017-03-29
Update and rename african-zoology.csl to african-zoology-sabinet.csl

damnation333:
via https://forums.zotero.org/discussion/comment/272507/#Comment_272507


rmzelle:
(replaced by #2600)

PR #2600, opened 2017-03-29
Update african-zoology.csl

damnation333:
via https://forums.zotero.org/discussion/comment/272507/#Comment_272507
continued from https://github.com/citation-style-language/styles/pull/2599


rmzelle:
Thanks!

PR #2603, opened 2017-03-30
Update and rename dependent/journal-of-structural-geology.csl to jour…

damnation333:
…nal-of-structural-geology.csl

via https://forums.zotero.org/discussion/65043/style-error-journal-of-structural-geology


adam3smith:
I think we're still waiting for map formatting here, so waiting to merge.

damnation333:
Thank you. How long does it take to "see" these things? 
I do understand why and how to use groups now and do it loads more, but i still don't go over the code line by line yet and I spot weird things.

adam3smith:
I actually spend some time on this because it didn't work quite right when I tested some items (one of the biggest advantages of groups is that they work better when some fields are missing). Generally, whenever add/change code, I try to do everything with delimiters and only resort to affixes as a last resort. That sometimes does lead to groups nested a couple of levels deep, but it works beautifully down the line -- those styles are really easy to tweak.

PR #2606, opened 2017-03-31

damnation333:
via https://forums.zotero.org/discussion/65073/style-error-journal-of-visualized-experiments#latest


adam3smith:
Thanks!

PR #2608, opened 2017-03-31
Update applied-spectroscopy-reviews.csl

damnation333:
via https://forums.zotero.org/discussion/comment/273173/#Comment_273173


adam3smith:
Looks good to me, let's see how it goes.

PR #2610, opened 2017-04-01
Turn Harvard1 into dependent

adam3smith:
Turn Harvard1 into dependent of Harvard Cite Them Right. Still not getting rid of it, but at least
it's following a clear guide. Redirect all templates.

Follow-up of https://github.com/citation-style-language/styles/pull/2514#issuecomment-280897069.


adam3smith:
@rmzelle -- since this one is larger, I'm leaving it for you to merge, but gtg from my side.

rmzelle:
@adam3smith, I'm kicking the ball back to you. If you agree with my changes feel free to squash-merge. I added some explanation in the style's summary. I also removed the "author-date" in the title (replacing it by "retired") and removed the "generic-base" category to put as little emphasis as possible on the style.

rmzelle:
Also, slightly off-topic, the "documentation" link for https://github.com/citation-style-language/styles/blob/master/harvard-cite-them-right.csl is a bit weird. Should we link to https://he.palgrave.com/page/detail/Cite-Them-Right/?K=9781137585042 instead. I also noticed that the 10th edition was released in May 2016, while we're still on the 9th.

adam3smith:
Looks good. Two questions:
1) How about "deprecated" rather than "retired"? The latter sounds odd to me
2) For the style guide link: The current one won't do, obviously, since the link doesn't lead anywhere. We can link to either the book or the website for it (http://www.citethemrightonline.com/) but both are obviously not universally accessible. Alternatively, we could link to a freely available summary: https://www.canterbury.ac.uk/library/docs/harvard.pdf I'd be inclined to do the latter or both. Thoughts?
3) I'll update to the 10th edition. Fortunately, changes are minimal.

rmzelle:
Great, thanks. Maybe link to the "online" website and the summary, then?

PR #2611, opened 2017-04-01
Create water-sa.csl

damnation333:
via https://forums.zotero.org/discussion/65090/request-style-for-water-sa-journal

### DIFF
--- a/water-sa.csl
+++ b/water-sa.csl
@@ -13,7 +13,7 @@
     <category citation-format="author-date"/>
     <category field="science"/>
     <issn>0378-4738</issn>
-    <updated>2017-04-01T15:01:19+00:00</updated>
+    <updated>2017-04-02T17:51:23+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
@@ -47,7 +47,7 @@
   <macro name="secondary-contributors">
     <choose>
       <if type="article-journal chapter paper-conference entry-dictionary entry-encyclopedia" match="none">
-        <group delimiter=", " prefix=" (" suffix=")">
+        <group delimiter=", " prefix=" (" suffix="). ">
           <names variable="container-author" delimiter=", ">
             <name and="text" initialize-with=". " delimiter=", "/>
             <label form="short" prefix=", " text-case="title"/>
@@ -142,7 +142,7 @@
           </else>
         </choose>
       </if>
-      <else-if type="book" match="any">
+      <else-if type="book report" match="any">
         <group delimiter=" ">
           <text variable="number-of-pages"/>
           <label variable="number-of-pages" form="short"/>
@@ -188,7 +188,7 @@
             <text variable="title"/>
           </if>
           <else>
-            <text variable="title" font-style="italic"/>
+            <text variable="title" suffix="."/>
           </else>
         </choose>
       </if>
@@ -218,7 +218,7 @@
         </choose>
       </else-if>
       <else>
-        <text variable="title"/>
+        <text variable="title" suffix="."/>
       </else>
     </choose>
   </macro>
@@ -226,11 +226,11 @@
     <text macro="title"/>
     <choose>
       <if type="report thesis" match="any">
-        <group prefix=" (" suffix=")" delimiter=", ">
+        <group prefix=". " delimiter=", ">
           <group delimiter=" ">
             <choose>
               <if variable="genre" match="any">
-                <text variable="genre"/>
+                <text variable="genre" suffix=","/>
               </if>
               <else>
                 <text variable="collection-title"/>
@@ -291,15 +291,15 @@
           <choose>
             <if type="article-journal article-magazine" match="none">
               <group delimiter=" ">
+                <text variable="publisher" suffix=","/>
                 <choose>
                   <if variable="publisher-place">
-                    <text variable="publisher-place" suffix=","/>
+                    <text variable="publisher-place"/>
                   </if>
                   <else>
                     <text variable="event-place"/>
                   </else>
                 </choose>
-                <text variable="publisher" suffix="."/>
               </group>
             </if>
           </choose>
@@ -469,7 +469,7 @@
             <else>
               <group>
                 <text term="volume" form="short" plural="true" text-case="capitalize-first" suffix=" "/>
-                <number variable="number-of-volumes" form="numeric" prefix="1&#8211;"/>
+                <number variable="number-of-volumes" form="numeric" prefix="1–"/>
               </group>
             </else>
           </choose>
@@ -528,7 +528,7 @@
         <text variable="container-title" font-style="italic" text-case="title"/>
       </if>
       <else-if type="bill legal_case legislation" match="none">
-        <text variable="container-title" font-style="normal"/>
+        <text variable="container-title" font-style="normal" suffix=". "/>
       </else-if>
     </choose>
   </macro>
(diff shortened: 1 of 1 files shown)


rmzelle:
@damnation333, can you leave a note here when this should be good to merge?

water-sa.csl
```diff
@@ -0,0 +1,631 @@
...
+  </macro>
+  <macro name="title-plus-extra">
+    <text macro="title"/>
+    <choose>
+      <if type="report thesis" match="any">
+        <group prefix=". " delimiter=", ">
+          <group delimiter=" ">
+            <choose>
+              <if variable="genre" match="any">
+                <text variable="genre" suffix=","/>
```
libora: It's necessary to add space to the suffix (suffix=", ")

water-sa.csl
```diff
@@ -0,0 +1,631 @@
...
+        <text macro="container-title"/>
+      </group>
+    </group>
+  </macro>
+  <macro name="container-title">
+    <choose>
+      <if type="article article-journal article-magazine article-newspaper" match="any">
+        <text variable="container-title" font-style="italic" text-case="title"/>
+      </if>
+      <else-if type="bill legal_case legislation" match="none">
```
libora: name of book for chapter or cenference proceedings have to be in italics and finished by dot. So here can be add new code:
     `<else-if type="chapter paper-conference" match="any">`
        `<text variable="container-title" font-style="italic" suffix=". "/>`
      `</else-if>`

water-sa.csl
```diff
@@ -0,0 +1,631 @@
...
+    </sort>
+    <layout prefix="(" suffix=")" delimiter="; ">
+      <group delimiter=", ">
+        <text macro="author-short"/>
+        <text macro="issued-year"/>
+        <text macro="citation-locator"/>
+      </group>
+    </layout>
+  </citation>
+  <bibliography et-al-min="11" et-al-use-first="10" entry-spacing="0" line-spacing="2" hanging-indent="true">
```
libora: add delimiter-precedes-last="never"

water-sa.csl
```diff
@@ -0,0 +1,631 @@
...
+              </else>
+            </choose>
+          </else>
+        </choose>
+      </else>
+    </choose>
+  </macro>
+  <macro name="title">
+    <choose>
+      <if type="book graphic manuscript motion_picture report song speech thesis" match="any">
```
libora: Title of report can not be in italic --> delete "report" from line 186 (response from editor Water SA journal)

water-sa.csl
```diff
@@ -0,0 +1,631 @@
...
+        <choose>
+          <if variable="version" type="book" match="all">
+            <text variable="title"/>
+          </if>
+          <else>
+            <text variable="title" font-style="italic" suffix="."/>
```
libora: add space to suffix (suffix=". ")

water-sa.csl
```diff
@@ -0,0 +1,631 @@
...
+      <else>
+        <text variable="title" suffix="."/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="title-plus-extra">
+    <text macro="title"/>
+    <choose>
+      <if type="report thesis" match="any">
+        <group prefix=". " delimiter=", ">
```
libora: delete dot from prefix=". "  This prefix makes error in documents type Thesis with other changes in template.

libora:
version with all suggested changes is available on https://drive.google.com/file/d/0B9G2tT5B0K0WbmlrbEhOcGptQlE/view?usp=sharing
(I am not sure if I did not forget comment any changes)
This template creates correct outputs for examples in the guide for authors, but we still have tried accuracy for other documents.

water-sa.csl
adam3smith: why? I don't think so. There's a space in the prefix in title-plus-extra. I'd generally encourage to not use spaces in suffixes pretty much ever. Use delimiters where possible and put spaces in prefixes otherwise.

libora: I agree with the role of prefixes/suffixes and delimiters, but I only add a second string to existing suffix` even I think all interpunction should be realised by delimiters. I suggest only changes which I've tested on examples from the guide for authors. My suggestions really can be done in a different way and @damnation333 can use a different coding. I've created a public group for testing this template https://www.zotero.org/groups/water_sa_style/ and everybody can test output from this template on the same documents.

adam3smith: I'm just wondering why it's necessary. Since citeproc-js removes duplicate spaces, I don't think it matters greatly, but it's better to avoid.

libora: If there is not this space then the output for books is incorrect "title.publisher"

adam3smith: I see -- I'd argue that's better handled with a prefix on the group containing event and publisher in the bibliography. For one, it's easier to read, but also event doesn't have a space before it either.

libora:
I have short free time so I check the output for reports:
11. If the report has only Title, author, péalce and institution then there is missing space between Title and Publisher: CZSO (2014) Farm Structure Survey 2013.Czech Statistical Office, Prague.
12. If the report is in the series but has item Report type empty, then there are missing interpunction and space between  Series title and Publisher
13. Duplicit or unnecessary "No." when ty number of report is not numerical - KOK K, GRAMBERGER M, ZELLMER K, SIMON K-H, JÄGER J and OMANN I (2013) Report on the new methodology for scenario analysis and based on an analysis of past scenario exercises. No. CLIMSAVE Deliverable 3.1. Retrieved from http://www.climsave.eu/climsave/outputs.html.
14. Unnecessary comma between report type and report number: KOK K, VAN VLIET M, BÄRLUND I, SENDZIMIR J and DUBEL A (2009) First (‘first-order’) draft of pan-Europen storylines - result from the second pan-European stakeholder workshop. SCENES Deliverable, No. 2.6. Wageningen University, Wageningen.

libora:
I've  got the response from editor of Water SA for some points from yesterday and some other notes:
Point 5 (URL) - It is not essential, but is also not a problem. However, add ‘URL:’ before the URL.
15. How to write volume and edition of books? - I put the Volume as part of the title, and so also in italics. The edition follows in brackets after the title (no italics, and in the form ‘2nd edn’. (nd in 2nd is upper index) - example: Alcamo J (ed.) _Environmental Futures: the Practice of Environmental Scenario Analysis. Volume 2._ (2nd edn).
16. The chapter title should be in sentence case, not title case.
17. Book title should be in title case
18. If the type Book has editor as author then should be write with small cases "(ed.)" (current output from style is "(Ed.)"
19. for the edition (currently ‘ed.’) should be ‘edn’

water-sa.csl
```diff
@@ -0,0 +1,634 @@
...
+    <category citation-format="author-date"/>
+    <category field="science"/>
+    <issn>0378-4738</issn>
+    <eissn>1816-7950</eissn>
+    <updated>2017-04-04T15:00:30+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="en">
+    <terms>
+      <term name="editortranslator" form="short">
```
libora: add definition for term edition = edn 
`      <term name="edition"  form="short">`
`        <single>edn</single>`
`        <multiple>edn</multiple>`
`      </term>`
solution for erro No. 19)

water-sa.csl
```diff
@@ -0,0 +1,634 @@
...
+        </group>
+      </if>
+    </choose>
+  </macro>
+  <macro name="author">
+    <names variable="author">
+      <name and="text" initialize-with="" name-as-sort-order="all" sort-separator=" ">
+        <name-part name="family" text-case="uppercase"/>
+      </name>
+      <label form="short" prefix=" (" suffix=")" text-case="capitalize-first"/>
```
libora: delete  text-case="capitalize-first" (solution of error no. 18)

water-sa.csl
```diff
@@ -0,0 +1,634 @@
...
+      </group>
+    </group>
+  </macro>
+  <macro name="container-title">
+    <choose>
+      <if type="article article-journal article-magazine article-newspaper" match="any">
+        <text variable="container-title" font-style="italic" text-case="title"/>
+      </if>
+      <else-if type="chapter paper-conference" match="any">
+        <text variable="container-title" font-style="italic" suffix=". "/>
```
libora: add text-case="title" (error no. 17)

libora:
I've completely redesign template for Water SA. My template is based on ISO690 style. Now I communicate with the editor of Water SA to checking the accuracy of the template. Unfortunately, there are a lot of rules of Water SA's style which are not documented in the Guide for authors.

damnation333:
Ok. @libora 
I can close this commit then and you can open a new one then and we continue there. Easier that way.

libora:
@damnation333 Yes, you can.

2017-04-15 15:00 GMT+02:00 damnation333 <notifications@github.com>:

PR #2612, opened 2017-04-01
Create the-british-journal-for-the-philosophy-of-science.csl

damnation333:
via https://forums.zotero.org/discussion/65092/style-request-the-british-journal-for-the-philosophy-of-science

[_ BJPS-House-Style-01-2016.pdf](https://github.com/citation-style-language/styles/files/887880/_.BJPS-House-Style-01-2016.pdf)


damnation333:
Can somebody explain what's going on here? Something wrong with the et-al?

rmzelle:
> Can somebody explain what's going on here? Something wrong with the et-al?

CSL is rather picky about where to put the `<et-al/>` element (it should directly follow `<name/>`), which was done on purpose because "et al." always directly follows the abbreviated name list. Per http://docs.citationstyles.org/en/stable/specification.html#et-al:

> the optional cs:et-al element, which must follow the cs:name element (if present)

rmzelle:
(I asked the journal via Twitter to fix their author guidelines link, but haven't heard back yet: https://twitter.com/rintzezelle/status/849301328211980289. If we do, we should be able to update the "documentation" link)

PR #2613, opened 2017-04-01
Add test for "sentence" case on title and container-title

rmzelle:
(follow-up of https://github.com/citation-style-language/styles/pull/2609)

### DIFF
--- a/spec/styles_spec.rb
+++ b/spec/styles_spec.rb
@@ -71,6 +71,16 @@ shared_examples "style" do |basename, (filename, path, style, error), in_depende
         end
       end
 
+      it "must not use sentence case for title and container-title variables" do
+        style.descendants!.each do |node|
+          if node.matches?("text")
+            if node[:'text-case'] == 'sentence'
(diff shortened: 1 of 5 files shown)


spec/styles_spec.rb
```diff
@@ -71,6 +71,16 @@
...
+      it "must not use sentence case for title and container-title variables" do
+        style.descendants!.each do |node|
+          if node.matches?("text")
+            if node[:'text-case'] == 'sentence'
+              expect(node[:'variable']).not_to match(/^title|container-title$/)
+            end
+          end
+        end
+      end
+
```
rmzelle: @adam3smith, the logic is not the most elegant, but RSpec is still mostly a mystery to me. It seems to work and doesn't seem to be particularly slow, so I guess it's okay.

rmzelle:
@adam3smith, I'll leave this for you to merge. Looks like your earlier find-and-replace didn't catch all the cases.

adam3smith:
Looks good to me. Thanks -- yes, I used an overly simplistic `sed`.

PR #2614, opened 2017-04-02
Create clinical-radiology.csl

damnation333:
via https://forums.zotero.org/discussion/64304/style-suggestion-for-clinical-radiology
style guidelines differ slightly from actual papers (bold volume, issue in brackets)

It's probably going to say it doesn't validate due to the et-al in line 28 again.


rmzelle:
Thanks!

PR #2616, opened 2017-04-03

JamieIainMacpherson:
A client noticed that the title was missing from journal references in ACS style. Based on my reading of the guide the title should be added as proposed.

### DIFF
--- a/american-chemical-society.csl
+++ b/american-chemical-society.csl
@@ -128,7 +128,10 @@
       <choose>
         <if type="article-journal review" match="any">
           <group delimiter=" ">
-            <text variable="container-title" font-style="italic" form="short"/>
+            <group delimiter=". ">
+              <text variable="title"/>
+              <text variable="container-title" font-style="italic" form="short"/>
+            </group>
             <group delimiter=", ">
               <text macro="issued" font-weight="bold"/>
               <choose>


adam3smith:
I'm pretty sure there are still journal's that don't include the title. We do have various ACS styles with title, e.g. https://github.com/citation-style-language/styles/blob/master/american-chemical-society-with-titles-doi-no-et-al.csl

JamieIainMacpherson:
@adam3smith Fair enough, my thoughts were that some journals do include the title so it should be in the primary ACS style but I see your point. I'll feedback to my client to use one of the (with titles) styles instead. Thanks for your help.

adam3smith:
we try to keep the journal styles linked to the correct ACS style, so if you know which journal you're writing for, just picking that (dependent) style should work -- and where that is wrong, we're happy to fix it.

rmzelle:
@JamieIainMacpherson, you're at Chegg? (it often helps us to know what software people are using the styles with)

For the whole set of ACS style variants we have, see https://www.zotero.org/styles?q=american%20chemical. We've been in contact with an ACS editor about their styles, but they're a little disorganized, so these are just the variants we identified ourselves.

(and while I was stalking you on GitHub I came across https://github.com/easybib/styles/pull/1; we just want to let you know that upstream submissions (i.e. here) are always welcome)

JamieIainMacpherson:
@rmzelle Hi, yes I'm at Chegg. Thanks for the feedback about the Bath style, at the moment there's still a few things that need changing but if I get a chance to make those changes I'll submit it.

rmzelle:
@JamieIainMacpherson, okay, great.

PR #2619, opened 2017-04-06
Update quaternary-international.csl

damnation333:
@adamsmith
via https://forums.zotero.org/discussion/comment/273637/#Comment_273637


adam3smith:
cool, thanks!

PR #2620, opened 2017-04-06
Create journal-of-agricultural-science.csl

damnation333:
via https://forums.zotero.org/discussion/64181/style-request-journal-of-agricultural-science
adapted for journal, book, chapter.
open access example: https://doi.org/10.1017/S002185961600099X


rmzelle:
Great, thanks!

PR #2622, opened 2017-04-11
Update idojaras-quarterly-journal-of-the-hungarian-meteorological-ser…

leeladam:
…vice.csl

Corrected bibliography ordering of references from the same author.


adam3smith:
Thanks (checked on recent issue & changes look right, although journal isn't 100% consistent)

PR #2623, opened 2017-04-11
Update springer-mathphys styles

damnation333:
via https://forums.zotero.org/profile/3932165/annadci
- added to include DOI for journals.
- removed the dot at the end

### DIFF
--- a/springer-mathphys-author-date.csl
+++ b/springer-mathphys-author-date.csl
@@ -115,7 +115,7 @@
       <key macro="author-in-citation"/>
       <key macro="issued-sort" sort="ascending"/>
     </sort>
-    <layout >
+    <layout>
       <text macro="author" prefix=" " suffix=": "/>
       <choose>
         <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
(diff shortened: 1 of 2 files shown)


damnation333:
@adam3smith 
I'll never get this GitHub thingamajig....
I've edited in my damnation333 repo and then clicked on commit to master branch. i assume that was wrong? https://github.com/damnation333/styles/blob/master/springer-mathphys-brackets.csl

adam3smith:
yeah, you need to make the change in your patch-22 branch

damnation333:
I think that was unsuccesfull: https://github.com/damnation333/styles/pull/1

adam3smith:
no sorry, that went the wrong way. If you'd like to figure it out, I can walk you through that. Otherwise I'm also just happy to make the change for you.

damnation333:
NO, I need to learn this. Where the heck do I find branch-22 of my repository? I see it mentioned up there, but you can't click on it.

adam3smith:
Two ways:
1) Go to https://github.com/damnation333/styles and at the top left, where it says "Branch: Master" click on the arrow next to it and select patch-22

2) Go to https://github.com/damnation333/styles/tree/patch-22 directly. Change the last part of the URL for different branches, obviously.

damnation333:
ahhhhhhhhhhh!!
Thanks. That was the missing link. would've never looked there for the patches.

springer-mathphys-author-date.csl
```diff
@@ -114,7 +115,7 @@
       <key macro="author-in-citation"/>
       <key macro="issued-sort" sort="ascending"/>
     </sort>
-    <layout suffix=".">
+    <layout>
```
adam3smith: None of the references have a period at the end or just not the ones with DOI? For the latter, we'd need to find a better solution. I also wonder if it wouldn't make sense to put the access macro in there independent of item type now that we increasingly see books & chapters with DOI.

damnation333: None of them. Must have been a mistake in the old style (or they changed it).

adam3smith:
Thanks!

PR #2624, opened 2017-04-11
Update the-open-university-harvard.csl

damnation333:
via https://forums.zotero.org/discussion/65192/style-error-the-open-uinversity-harvard
I feel like I made this the complicated way? Could I have gotten rid of one of the if loops?


rmzelle:
Let's go with this. I assume a DOI is always preferable to an URL, so no reason to limit it to journals.

@adam3smith, please squash and merge if you're happy with this. (I think we can do without the XML comments that are being dropped.)

PR #2625, opened 2017-04-11

damnation333:
via https://forums.zotero.org/discussion/65214/style-update-royal-society-of-chemistry


rmzelle:
Great, thanks!

PR #2626, opened 2017-04-12
Create journal-of-the-american-heart-association.csl

damnation333:
via https://forums.zotero.org/discussion/65182/style-request-journal-of-the-american-heart-association
Adapted for journals, books, chapters and webpages.

- Is this sorting by appearance?


adam3smith:
Thanks!

PR #2627, opened 2017-04-12
Create developmental-neurobiology.csl

damnation333:
via https://forums.zotero.org/discussion/46858/style-request-developmental-neurobiology

didn't have to change much for this one as it's based on another Wiley journal style.
Added webpage though.
Wiley should really make their styles uniform. Way easier for them and other people then.


adam3smith:
Thanks!

rmzelle:
> Wiley should really make their styles uniform. Way easier for them and other people then.

I prodded them once in 2013, and they weren't very receptive. When I asked whether they standardize their citation styles at all, they gave the non-answer "Unfortunately, I regret to inform you that we already have our own way of standardizing and handling journal citation/reference styles.".

PR #2631, opened 2017-04-13
Update journal-of-biogeography.csl

damnation333:
via https://forums.zotero.org/discussion/45898/style-error-journal-of-biogeography


rmzelle:
Looks good, thanks!

PR #2633, opened 2017-04-13

damnation333:
via https://forums.zotero.org/discussion/65199/style-error-association-for-computing-machinery-acm

### DIFF
--- a/association-for-computing-machinery.csl
+++ b/association-for-computing-machinery.csl
@@ -13,7 +13,6 @@
     </author>
     <category citation-format="numeric"/>
     <category field="engineering"/>
-    <summary>The ACM Journal Reference Format (numeric) variant of the Chicago style.</summary>
     <updated>2017-07-03T15:22:45+00:00</updated>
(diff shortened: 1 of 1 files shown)


rmzelle:
@damnation333, in case you haven't figured this out yet, this is failing because the style has a lot of dependents, which are still all labeled as "author-date".

We can fix this by updating the template at https://github.com/citation-style-language/journals/tree/master/acm and regenerating the dependents, once we confirm that all ACM journals indeed use the numeric format. It seems to be the case (http://www.acm.org/publications/authors/reference-formatting), but perhaps you could check 2-3 of the individual journals to make sure?

damnation333:
Hi @rmzelle 
I see that and suspected that. At least the fix is just changing the template.
I've just tried to access a variety of these journals. But even with Unpaywall I don't have access to any papers.
And of course each individual links to the ACM style guide, so they should have all the same style. If somebody with library access to this could check a few journals, that would be most helpful.

rmzelle:
Covered by #2800.

PR #2634, opened 2017-04-13
Update american-journal-of-archaeology.csl

damnation333:
via https://forums.zotero.org/discussion/65262/style-error-american-journal-of-archaeology


rmzelle:
Thanks! (I tried to clean up the delimiters a bit more)

PR #2638, opened 2017-04-18
Update international-journal-of-exercise-science.csl

damnation333:
https://forums.zotero.org/discussion/comment/274378/#Comment_274378


libora:
I recomend turn-off "access" macro. URL is not shown in any articles which I control on the web of the IJES journal.

bwiernik:
@libora URLs/DOIs not being shown in the published version doesn't necessarily mean they should be in the style. Often, a publisher might request URLs/DOIs in order to embed links or supply citation information, etc., but not show the links in the PDF. If a style guide requests URLs/DOIs, we should err on the side of supplying them, even if they aren't showing up in published versions.

rmzelle:
> If a style guide requests URLs/DOIs, we should err on the side of supplying them

I don't see them requesting them (at http://digitalcommons.wku.edu/ijes/styleguide.html), though, so I'll merge this as is.

PR #2639, opened 2017-04-19
Update international-journal-of-exercise-science.csl

libora:
Add issue (https://forums.zotero.org/discussion/comment/274378/#Comment_274378)
Turn off macro "access", because the URL is not needed by the guide for authors (http://digitalcommons.wku.edu/ijes/styleguide.html) and I did not find URL link in any article available on the journal pages (I've tested only five articles).


rmzelle:
(redundant in view of #2638)

PR #2640, opened 2017-04-19
Update international-organization.csl

damnation333:
via https://forums.zotero.org/discussion/65185/edit-the-citation-style


libora:
damnation did you check your changes with a guide for authors? For example, I see on http://iojournal.org/wp-content/uploads/2014/02/2014-Guidelines-Final.pdf term  "Available at" not "Available from" as you add to the style.

adam3smith:
Thanks for doing this!

PR #1952, opened 2016-03-14
Create Hastings Center Report CSL style

modalrealist:
Begins with the Chicago Full Note CSL template, adapted to HCR submission guidelines at http://www.thehastingscenter.org/Publications/HCR/Guidelines.aspx.

PR #2641, opened 2017-04-19
Create the-hastings-center-report.csl

damnation333:
continued from https://github.com/citation-style-language/styles/pull/1952/files
I did not review this at all. These styles are crazy complicated!


adam3smith:
I didn't check this closely either, but these were the things that I noticed at one glance and the rest looked good enough to take. Thanks!

PR #2644, opened 2017-04-23
Create chinese-std-gb-t-7714-2005-numeric-chinese-yet-another-implemetation

JunhuiJi:
overcomes some problems of the csl file listed at https://github.com/citation-style-language/styles/blob/master/chinese-gb7714-2005-numeric.csl


rmzelle:
@JunhuiJi, sorry for the very late reply, but I hope you understand that we can't accept styles labeled "yet another implementation".

We're happy to accept corrections to the existing style, or to accept a variant if it's well-described in a published style guide. Can you elaborate a little on the "problems" that your submission fixes?

PR #2649, opened 2017-04-25
Update applied-spectroscopy-reviews.csl

damnation333:
via https://forums.zotero.org/discussion/comment/274746/#Comment_274746


damnation333:
Continued by requestor of style in https://github.com/citation-style-language/styles/pull/2652

Issue #2648, opened 2017-04-24
apa.csl ignoring post-weblog as needing access time


reagle:
apa.csl only specifies an accessed date (e.g., " Retrieved January 1, 2015, from ") for `webpage` types. `post-weblog` should have it too. And what about including it for any source that has a URL? At the least, it should be:

```
<if type="post-weblog webpage" match="any">
```

adam3smith:
I agree on post-weblog, but we're purposefully not including it for all URLs, since as per APA you're only supposed to add it when the source is likely to change (I'd have to check on the exact wording, but that's the gist of it).

reagle:
Okay, now in pull request #2650.

reagle:
now merged

PR #2650, opened 2017-04-25
apa.csl: add post and post-weblog to date accessed

reagle:
Fix from issue #2648.


adam3smith:
Closes #2648 
Thanks!

PR #2653, opened 2017-04-26
Add pamphlet pages for map type

inkenbrandt:
This is an attempted fix for the map type output in the bibliography. The style guide (http://files.geology.utah.gov/online/c/c-105.pdf) indicates that if a map includes a pamphlet, it should include  ", # p. pamphlet," after the series title and before the scale.

### DIFF
--- a/utah-geological-survey.csl
+++ b/utah-geological-survey.csl
@@ -152,7 +152,7 @@
         <else-if type="map">
           <text variable="title" prefix=" "/>
           <text variable="collection-title" prefix=": "/>
-          <text variable="note" prefix=", " suffix=" p. pamphlet"/>
+          <text variable="number-of-pages" prefix=", " suffix=" p. pamphlet"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
I really would prefer to not use "note" -- even in Zotero that requires awkward data entry and it straight-up breaks things for other reference managers. Can we really not think of a better way of doing this? Best fit would seem to be to include this as number of pages, maybe?

inkenbrandt:
I understand your concern, and I generally avoid using this field, but Zotero does not have a field for page numbers for the map type. Page numbers would be the most appropriate field, though.

adam3smith:
You can add this as 
number-of-pages: # p. pamphlet
in the Extra field in Zotero and it will get picked up as number-of-pages for the CSL. Let's do that instead, please.

adam3smith:
Thanks!

PR #2654, opened 2017-04-26
Update harvard-coventry-university.csl

Elbar914:
From EasyBib: A user reported that when she cites websites like https://www.nice.org.uk/guidance/cg181/chapter/1-Recommendations#lipid-modification-therapy-for-the-primary-and-secondary-prevention-of-cvd-2 in Harvard - Coventry University, "anon" is showing up as the author for each one. I'm not finding that listed in any author or contributor tags in the page source, and when I cite a different web URL in Harvard CU and remove the author, "anon" shows up. CU says this is not correct.

Fix: remove "anon" from website citations with no author in Harvard CU

### DIFF
--- a/harvard-coventry-university.csl
+++ b/harvard-coventry-university.csl
@@ -53,11 +53,6 @@
       <name and="text" sort-separator=", " initialize-with="." delimiter=", "/>
     </names>
   </macro>
-  <macro name="cite-editor">
-    <names variable="editor" delimiter=", ">
-      <name form="short" and="text" delimiter=", " initialize-with=". "/>
-    </names>
-  </macro>
   <macro name="bookauthor">
     <names variable="container-author">
       <name name-as-sort-order="all" and="text" sort-separator=", " initialize-with="." delimiter=", "/>
@@ -91,10 +86,10 @@
       <else>
         <names variable="author">
           <name name-as-sort-order="all" and="text" sort-separator=", " initialize-with="." delimiter=", " form="long"/>
-          <label form="short" prefix=" " strip-periods="false"/>
+          <label form="short" prefix=" (" suffix=")"/>
           <substitute>
-            <text macro="editor"/>
-            <text variable="title" text-case="title" quotes="true" font-style="normal" text-decoration="none"/>
+            <names variable="editor"/>
+            <text macro="title"/>
           </substitute>
         </names>
       </else>
@@ -119,8 +114,8 @@
         <names variable="author">
           <name form="short" and="text" delimiter=", " initialize-with=". "/>
           <substitute>
-            <text macro="cite-editor"/>
-            <text variable="title" text-case="title" quotes="true" font-style="normal"/>
+            <names variable="editor"/>
+            <text macro="title"/>
           </substitute>
         </names>
       </else>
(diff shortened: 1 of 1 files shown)


adam3smith:
I guess you've talked to the university, but here's what they actually write (guide linked to from http://www.coventry.ac.uk/study-at-coventry/student-support/academic-support/centre-for-academic-writing/support-for-students/academic-writing-resources/cu-harvard-reference-style-guide/ ):
image
It seems to me that's exactly what the style is doing.

Elbar914:
Thanks for getting back so quickly. They did not express the fact that there were multiple options for missing author scenarios.


Thanks again,


Elise

adam3smith:
If you could check back with them -- we can certainly change this (though it seems like we may want to go with "title" and not "container-title" given the text? Also, definitely not in quotes from the looks of it.

Elbar914:
I will check back with them and confirm what their preference is.


Thanks again,


Elise

Elbar914:
Hello, I have received confirmation that "Anon" should preferably be used only if the author is actually listed as "Anonymous." Due to user demand and frequent requests for a CSL update, if the author is simply unknown, then it should default to the "title."


Let me know what needs to be done in order to complete this change.



Thanks,


Elise

harvard-coventry-university.csl
```diff
@@ -98,7 +97,7 @@
           <label form="short" prefix=" " strip-periods="false"/>
           <substitute>
             <text macro="editor"/>
-            <text macro="anon"/>
+            <text variable="container-title" text-case="title" quotes="true" font-style="normal" text-decoration="none"/>
```
adam3smith: given the response you received, you'll want to use `<text variable="title"/>` here, not `container-title`. ditto below.

adam3smith:
You'll need to deleted the anon macro now that it's not used anymore.

Elbar914:
Thanks for your help.

adam3smith:
sorry, one more question -- the quotes you're putting around the title in those cases, are those based on something you're getting from the University, too? Because I don't see those in the styleguide.

Elbar914:
I believe it depends on the source type. Journal articles get single quotation marks, while books, websites, journals should be in italics.

harvard-coventry-university.csl
```diff
@@ -284,7 +280,7 @@
     <layout prefix="(" suffix=")" delimiter=", ">
       <choose>
         <if locator="page">
-          <group delimiter=":&#160;">
+          <group delimiter=": ">
```
rmzelle: @adam3smith, before you merge this, we might want to make certain these non-breaking spaces are still there, and escape them.

Elbar914:
Anything else you need from me?

rmzelle:
@Elbar914, I assume you didn't actually mean to make any changes to the non-breaking spaces in the original file, right?

Elbar914:
I did not, no.

rmzelle:
Okay, it was a bit hard to tell whether those special spaces were converted to regular spaces, or just 'unescaped', but I just put back the escaped non-breaking space codes. I think we were all set otherwise, thanks!

Elbar914:
Thanks so much!

PR #2656, opened 2017-04-27
Update modern-language-association.csl

Elbar914:
MLA 7 says for a work of visual art the medium of composition should be indicated. Our photograph citation leaves out "photograph."

Current: Cloyd, Allie. The Wedding. 1945. Museum of the City of New York, New York.
Expected: Cloyd, Allie. The Wedding. 1945. Photograph. Museum of the City of New York, New York.

I have added "graphic" to medium macro.

### DIFF
--- a/modern-language-association.csl
+++ b/modern-language-association.csl
@@ -33,7 +33,7 @@
     <category citation-format="author"/>
     <category field="generic-base"/>
     <summary>This style adheres to the MLA 7th edition handbook and contains modifications to these types of sources: e-mail, forum posts, interviews, manuscripts, maps, presentations, TV broadcasts, and web pages.</summary>
-    <updated>2014-07-06T20:05:10+00:00</updated>
+    <updated>2017-04-27T18:03:35+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
@@ -153,6 +153,9 @@
             <text value="Audio Recording" prefix=" "/>
           </else>
         </choose>
+      </else-if>
+      <else-if type="graphic" match="any">
+        <text variable="medium"/>
       </else-if>
       <else>
         <choose>
--- a/modern-language-association.csl
+++ b/modern-language-association.csl
@@ -154,8 +154,8 @@
           </else>
         </choose>
       </else-if>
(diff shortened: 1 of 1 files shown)


adam3smith:
Please have a look at the changes I made and see if they make sense (and please do test proposed edits: this one was missing a space between date and medium). Also, in my testing this looks like it's printing n. P. for artworks, which seems incorrect. Are you not seeing that?

Elbar914:
Your changes look correct, thank you. I tested this via the CSL validator here http://validator.citationstyles.org/. 

The n.P. did not come through in our testing, but is definitely incorrect.

rmzelle:
> The n.P. did not come through in our testing, but is definitely incorrect.

@Elbar914, coming back to this, have you been testing the style with Zotero, or another CSL-based reference manager? (Zotero currently doesn't have a "publisher" field for its artwork/graphic item type, so the current style will always print "N.p." for such items with Zotero; see https://aurimasv.github.io/z2csl/typeMap.xml#map-artwork)

Elbar914:
@rmzelle I will look into testing this further.

rmzelle:
@Elbar914, do you remember what program you used in the past?

Elbar914:
@rmzelle it was actually tested by someone who no longer works for us. I'm not sure what he used. 

Do you suggest a specific program?

rmzelle:
Zotero is generally considered to be the best-tested and most comprehensive CSL-implementation.

(can I also ask what your interest is in this? Are you affiliated with MLA?)

Elbar914:
@rmzelle thanks good to know. 

I work for Chegg, which owns/runs sites like EasyBib, Citation Machine, etc.

rmzelle:
Okay, thanks for letting us know. I think these sites usually use the same CSL processor as Zotero (citeproc-js), so yes, that would be my pick for you.

Elbar914:
Thanks, will look into that.

rmzelle:
Actually, since you're doing this for Chegg, a little background: while your websites and Zotero both use CSL (and even the same CSL processor software), there can be differences in which fields are available for the different item types. E.g. Zotero doesn't have a "publisher" field for the "graphic" item type, while it looks like your websites might have it. (CSL currently doesn't prescribe which item types should have which fields, so there is some variation between implementations)

Ideally our styles would work correctly with all implementations, but this might require some more coordination on the field mapping front. For Zotero, the full mapping is available at https://aurimasv.github.io/z2csl/typeMap.xml.

PR #2657, opened 2017-04-27

damnation333:
via https://forums.zotero.org/discussion/65446/style-error-imperial-college-london-harvard#latest

### DIFF
--- a/harvard-imperial-college-london.csl
+++ b/harvard-imperial-college-london.csl
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" class="in-text" demote-non-dropping-particle="sort-only" default-locale="en-GB" page-range-format="expanded">
   <info>
-    <title>Imperial College London - Harvard</title>
+    <title>Harvard - Imperial College London</title>
     <id>http://www.zotero.org/styles/harvard-imperial-college-london</id>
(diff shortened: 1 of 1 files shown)


adam3smith:
Note that the proposed solution would not have placed a comma between volume and pages with the issue number missing -- think of groups as logical units. Volume and issue are almost always good in a group, just like publisher and publisher place.

adam3smith:
Thanks!

damnation333:
Thanks for fixing this. :)

harvard-imperial-college-london.csl
```diff
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" class="in-text" demote-non-dropping-particle="sort-only" default-locale="en-GB" page-range-format="expanded">
   <info>
-    <title>Imperial College London - Harvard</title>
+    <title>Harvard - Imperial College London</title>
```
rmzelle: @damnation333, I rolled back the title change in https://github.com/citation-style-language/styles/commit/5ba1c6cb1fe16020e4551fd10f1ba864c8e6011a. We changed this for all university Harvard styles in https://github.com/citation-style-language/styles/commit/5def1b56828c5f3c7bb603afd56fb4a12e69546e#diff-00fbeb8bbff390e3f90235ae48c9843e back in March.

PR #2660, opened 2017-04-29
Change format to "numeric" for radiopaedia.csl

zuphilip:
This is an online resource and therefore (foot-)notes does not
make much sense, also it is a super-scripted number in the text.

HT @adam3smith which noted this "outliner" for medicine styles

### DIFF
--- a/radiopaedia.csl
+++ b/radiopaedia.csl
@@ -10,7 +10,7 @@
       <email>jeremy@radiopaedia.org</email>
       <uri>http://radiopaedia.org</uri>
     </author>
-    <category citation-format="note"/>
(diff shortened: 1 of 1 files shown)


zuphilip:
Here are some example urls:
* https://radiopaedia.org/articles/alzheimer-disease-1
* https://radiopaedia.org/articles/creutzfeldt-jakob-disease

adam3smith:
we also need to convert this into a numeric style in the citation section though, right?

zuphilip:
Yes, you are right. Moreover, the style seems to have changed in various other aspects which I tried to fix quickly...

adam3smith:
Thanks!

PR #2661, opened 2017-04-29
Update new-phytologist.csl

damnation333:
via https://forums.zotero.org/discussion/comment/275104/#Comment_275104


adam3smith:
Yeah, this looked great, thanks.

PR #2663, opened 2017-05-01
Update european-journal-of-international-law.csl

zuphilip:
Delete non-existing template link to "original-style", which maybe just copied from 
https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link


adam3smith:
that's funny. Thanks!

PR #2664, opened 2017-05-01
Update technische-universitat-munchen-controlling.csl

zuphilip:
The hwr-berlin style was renamed, cf.
https://github.com/citation-style-language/styles/blob/e09bfc04fe0dd4dd3090bb06125eca573f4da484/renamed-styles.json#L289

### DIFF
--- a/technische-universitat-munchen-controlling.csl
+++ b/technische-universitat-munchen-controlling.csl
@@ -7,7 +7,7 @@
     <id>http://www.zotero.org/styles/technische-universitat-munchen-controlling</id>
     <link href="http://www.zotero.org/styles/technische-universitat-munchen-controlling" rel="self"/>
     <link href="http://www.zotero.org/styles/universitat-freiburg-geschichte" rel="template"/>
-    <link href="http://www.zotero.org/styles/hwr-berlin" rel="template"/>
+    <link href="http://www.zotero.org/styles/hochschule-fur-wirtschaft-und-recht-berlin" rel="template"/>
     <link href="http://www.controlling.wi.tum.de/lehre/abschlussarbeiten/" rel="documentation"/>
     <author>
       <name>Tom Raley</name>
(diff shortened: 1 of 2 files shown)


zuphilip:
This search shows also that these are all occurences: https://github.com/citation-style-language/styles/search?utf8=%E2%9C%93&q=hwr&type=

adam3smith:
thanks!

technische-universitat-munchen-controlling.csl
```diff
@@ -7,7 +7,7 @@
     <id>http://www.zotero.org/styles/technische-universitat-munchen-controlling</id>
     <link href="http://www.zotero.org/styles/technische-universitat-munchen-controlling" rel="self"/>
     <link href="http://www.zotero.org/styles/universitat-freiburg-geschichte" rel="template"/>
-    <link href="http://www.zotero.org/styles/hwr-berlin" rel="template"/>
+    <link href="http://www.zotero.org/styles/hochschule-fur-wirtschaft-und-recht-berlin" rel="template"/>
```
rmzelle: @adam3smith, @zuphilip, only now seeing this, but I'm confused.

This template style was renamed in 2013 (see https://github.com/citation-style-language/styles/commit/d6754738d27fd528775648714ce36ef269373568#diff-2657503fcaf93aaf9836fa01c692099eR22 and https://github.com/citation-style-language/styles/commit/d6754738d27fd528775648714ce36ef269373568#diff-2657503fcaf93aaf9836fa01c692099eR22). Shouldn't our automatic tests have caught this way back? I'm pretty sure we're testing whether the template style exists in the current repository.

adam3smith: We definitely do, yes -- without actually having looked at the code, I'm guessing the test might just look if the first template exists. This style has two.

rmzelle: Right, that's it. Just discovered that independently :P.

(the code for checking the template link is at https://github.com/citation-style-language/styles/blob/a6646ba5657dd22f3f4a534f6190b4a45be2895b/spec/styles_spec.rb#L26)

rmzelle: Okay, (low-priority) issue created: https://github.com/citation-style-language/styles/issues/2802

PR #2665, opened 2017-05-01
Update iso690-author-date-cs.csl, iso690-numeric-brackets-cs.csl and iso690-numeric-cs.csl

krystof-k:
Add number of pages to book and thesis record in bibliography.

I hope that's right, it did the job. It should be compliant with the norm.

Would be happy about @libora and @kratec (authors of the original style) review.

### DIFF
--- a/iso690-author-date-cs.csl
+++ b/iso690-author-date-cs.csl
@@ -489,7 +489,7 @@
                 <text macro="edition"/>
                 <text macro="secondary-responsability"/>
                 <text macro="publisher-info"/>
-                <!--text macro="page"/-->
+                <text macro="page"/>
                 <text macro="collection"/>
               </group>
               <text macro="accessed"/>
@@ -598,6 +598,7 @@
               <text macro="publisher-place"/>
               <text macro="accessed"/>
             </group>
+            <text macro="page"/>
             <text macro="genre"/>
             <text macro="publisher"/>
           </group>
--- a/iso690-numeric-brackets-cs.csl
+++ b/iso690-numeric-brackets-cs.csl
@@ -449,6 +449,7 @@
                 </group>
                 <text macro="accessed"/>
               </group>
+              <text macro="page"/>
               <text macro="collection"/>
               <text macro="isbn"/>
             </group>
@@ -553,6 +554,7 @@
                   <text macro="accessed"/>
                 </group>
               </group>
+              <text macro="page"/>
               <text macro="genre"/>
               <text macro="publisher"/>
             </group>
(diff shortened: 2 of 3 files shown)


kratec:
Your suggestion is incorrect. Due to the current version of ISO 690 the pages is not requested to be written in a bibliographic reference of a book.

adam3smith:
That seems right. While I do see this in a fair number of instances claiming ISO 690, there's no indication in the actual standard to include the number of pages for books:
http://dominiodelasciencias.com/ojs/documentos/ISO690-2010.pdf (see table B1, p. 28)

krystof-k:
Thanks. Yep, I know, I understood that the style can contain even optional fields (and if anyone doesn't want to include some he can delete it from the references collection) – that's not like that, right?

I've just commited also a field for note for thesis (used for reference of thesis supervisor), that's the same case I guess…

adam3smith:
right, we don't want to encourage people to delete data from their records just because they don't want it cited in a given style.

krystof-k:
OK. I see it was even wrong (is it possible that the Czech version of ISO 690 is different? Because I didn't find this anywhere):

> If the whole of an item is being cited, its extent may optionally be stated in brackets.
> http://www.eursurveillance.org/ViewArticl.aspx [7pp]. 

And what is a correct approach if I want it there? Creating own style?

kratec:
I don´t recommend to include the pages because due to ISO 690 the location can be included "if a limited
number of copies are known to exist". Most of books are printed in many copies. Moreover most of citation styles don´t request the location in the references (besides ISO 690 I know only about NLM style). As Adam Smith wrote - it is better if users don´t delete any field in records...

kratec:
krysto-k: since me and my colleague created Czech versions of ISO 690 you are first person requesting this field. I recommend to make an offiline copy of the style where make changes and use it offline only for your purposes. After you make changes, open Zotero panel, choose configuration > Citing > Styles > click on symbol "+" and choose the changed CSL file from your computer. ;-)
Please, don´t maky changes in online version because me and my colleague prepared the style to be most consistent to the Czech official version of ISO 690 published by the Czech Office for Standards, Metology and Testing (http://www.unmz.cz/office/competence)

krystof-k:
I see, all right. I find as a bit weird approach, there should be something like *include optional fields* option, so that one style can contain both required and optional fields :)

I'll take a look at the Czech version of the norm itself :) Thanks for feedback!

kratec:
Czech version is a translation of PDF mentioned by Adam Smith, so you will find there same table with same sentence. ;-) Unfortunately ISO 690 is very complicated style, I regret that you have to follow it... :-(

libora:
@krystof-k 

> there should be something like include optional fields option, so that one style can contain both required and optional fields :)

Unfortunately, there is no way to do this.

PR #2666, opened 2017-05-02
Create behaviour.csl

damnation333:
via https://forums.zotero.org/discussion/46757/style-request-behaviour
adapted for journals, books and chapters.

### DIFF
--- a/behaviour.csl
+++ b/behaviour.csl
@@ -14,7 +14,7 @@
     <category field="biology"/>
     <issn>1568-539X</issn>
     <eissn>0005-7959</eissn>
-    <updated>2017-05-09T08:47:38+00:00</updated>
+    <updated>2017-05-09T09:02:11+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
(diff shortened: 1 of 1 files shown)


behaviour.csl
```diff
@@ -0,0 +1,378 @@
...
+          <name and="symbol" initialize-with=". " delimiter=", "/>
+          <label form="short" prefix=", " text-case="title"/>
+        </names>
+      </if>
+    </choose>
+  </macro>
+  <macro name="author">
+    <names variable="author">
+      <name and="symbol" initialize-with="." name-as-sort-order="all"/>
+      <label form="short" strip-periods="true" prefix=", "/>
```
adam3smith: I find it unlikely that it's ed/eds here and ed./eds. above (and Ed./Eds. for secondary contributors) likely all going to be identical.

behaviour.csl
```diff
@@ -0,0 +1,378 @@
...
+      <else>
+        <text variable="edition" suffix="."/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="locators">
+    <choose>
+      <if type="article-journal article-magazine" match="any">
+        <group delimiter=": " prefix=" ">
+          <group>
```
adam3smith: remove group (doesn't do anything)

behaviour.csl
damnation333: Thanks. good spot. Double checked that with some actual papers and fixed as necessary.

adam3smith:
Thanks!

PR #2667, opened 2017-05-03
Create epidemiology-and-infection.csl

damnation333:
via https://forums.zotero.org/discussion/36590/need-advice-on-style-not-in-depository#latest
Adapted for journal, book, chapter and web


adam3smith:
Thanks!

PR #2668, opened 2017-05-04
Create springer-vs-author-date-german.csl

damnation333:
via https://forums.zotero.org/discussion/65529/style-request-springer-vs-chicago-author-date
Guidelines: [Hinweise zur Manuskripterstellung_final.pdf](https://github.com/citation-style-language/styles/files/978822/Hinweise.zur.Manuskripterstellung_final.pdf)


adam3smith:
Thanks both!

PR #2669, opened 2017-05-04

damnation333:
via https://forums.zotero.org/discussion/65534/style-request-photochemical-photobiological-sciences-with-titles
Needs to be checked if all the RSC actually have adapted this new style including the title for journal articles.

### DIFF
--- a/royal-society-of-chemistry-with-title.csl
+++ b/royal-society-of-chemistry-with-title.csl
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
   <info>
-    <title>Royal Society of Chemistry</title>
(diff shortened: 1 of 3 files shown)


adam3smith:
From that thread it appears that is *not* the case so we need to turn this into an independent style, right?

damnation333:
@adam3smith I think it might be worth to give the RSC a shout directly and find out what's going on from the horse's mouth. I can do that unless somebody else already had been in touch with them in the past?

adam3smith:
please do. I don't think we have a direct contact there from the past. Feel free to invoke CSL ("I work with the Citation Style Language (CSL) project" or something along those line).

damnation333:
Great, that's sent. Let's see what they say (and when). :)

adam3smith:
@damnation333 I take it you never heard back? I'd suggest just going with a new indpendent style then as
royal-society-of-chemistry-with-title and changing the data in the journals repo to redirect Physical Chemistry Chemical Physics and Photochemical & Photobiological Sciences to that.

rmzelle:
Thanks!

PR #2670, opened 2017-05-04
Add article-newspaper to publisher exclusions

ultimatejimmy:
Newspapers also need to be excluded per the manual 6.30: 

> Periodical publisher names and locations are generally not included in references, in accordance with long practice.

### DIFF
--- a/apa.csl
+++ b/apa.csl
@@ -300,7 +300,7 @@
             </if>
           </choose>
           <choose>
-            <if type="article-journal article-magazine" match="none">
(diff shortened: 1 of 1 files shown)


bwiernik:
This change is correct per APA manual. Must have been an oversight when the style was written/revised and not noticed because most CSL clients don't have a publisher field for newspaper articles.

adam3smith:
Thanks -- indeed, we don't exclude publishers for periodicals in most citation styles.

PR #2672, opened 2017-05-05

libora:
We've discussed with Mr Kratochvíl probability of correct data input from Zotero users (especially from students). Because the current output from style is not fully correct according to ISO 690 (https://forums.zotero.org/discussion/57000/conditional-display-of-date-of-conference-meeting#latest) we've concluded do not display the date twice in the bibliography. If will be implemented the test of date parts into the CSL (https://github.com/citation-style-language/csl-evolution/issues/2) then it will be changed to full compliance with the ISO 690.

### DIFF
--- a/iso690-author-date-cs.csl
+++ b/iso690-author-date-cs.csl
@@ -283,17 +283,17 @@
       </if>
     </choose>
   </macro>
-  <macro name="date-day-month">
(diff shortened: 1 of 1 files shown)


adam3smith:
OK, thanks

PR #2673, opened 2017-05-05
Update us-geological-survey.csl

jbellino-usgs:
Added "Computer Program" biblio hack (taken from [APA 6th ed.](https://forums.zotero.org/discussion/32710/apa-6th-edition-computer-program-entries)) to the U.S. Geological Survey style. Not based on official guidance, subject to change.


adam3smith:
Great, thanks!

PR #2674, opened 2017-05-05
Fix deutsche medizinische wochenschrift

bovender:
DMW is a German medical journal. The style did not meet the [requirements](https://www.thieme.de/statics/dokumente/thieme/final/de/dokumente/zw_dmw/Autorenrichtlinien_Uebersicht_Dossier.pdf) (see page 5, "Literaturverzeichnis"). I've fixed the style for the three types of publications that are explicitly defined by DMW, journal article, book, and web page.

### DIFF
--- a/deutsche-medizinische-wochenschrift.csl
+++ b/deutsche-medizinische-wochenschrift.csl
@@ -153,15 +153,15 @@
     <choose>
       <if type="book" match="any">
         <group delimiter=". ">
+          <group suffix=". Aufl.">
+            <number variable="edition"/>
+          </group>
           <text macro="publisher"/>
-          <group delimiter=" ">
-            <number variable="number-of-pages"/>
-            <label variable="number-of-pages" form="short"/>
-          </group>
-          <group delimiter=" ">
-            <number variable="edition" form="ordinal"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
Looks good -- I did some code clean-up, but that shouldn't affect the style. The one thing I don't understand is why you're switching the default locale? Since you still have hard-coded German language in the style, it should presumably remain in German (the author guidelines also have citations in German)

bovender:
Thanks for the clean-up.

> The one thing I don't understand is why you're switching the default locale?

I did this to force the use of "et al.". With a "de" locale, "et al." is translated to "u. a.", which is not what the journal wants (and by the way, I've never seen "u. a." in the German literature, but that may depend on the field; in the life sciences/biomedical literature, "et al." is used even in German texts).

Switching the entire locale is admittedly a hack. Is there a better way to get "et al."?

adam3smith:
Yes, there's a to handle et al in German. Check the edited style.

I'd have to look but I think we're switching u.a. to et al. in about 1/3 of German styles, so it's indeed very common, but by no means universal.

adam3smith:
Thanks!

PR #2676, opened 2017-05-09
Update harvard-cardiff-university.csl

YoungZoe:
The Cardiff university version of the Harvard style has changed so i was tasked by the University Library Service to update it as i created it originally.  There are many changes including: removal of [online] from web items, adding DOI, correcting sort order on citations and adding some new types of materials.

### DIFF
--- a/harvard-cardiff-university.csl
+++ b/harvard-cardiff-university.csl
@@ -4,9 +4,9 @@
     <title>Harvard - Cardiff University</title>
     <id>http://www.zotero.org/styles/harvard-cardiff-university</id>
     <link href="http://www.zotero.org/styles/harvard-cardiff-university" rel="self"/>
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks!

YoungZoe:
Thanks for this. When will the updated style appear in the editor?
Also i have had a request for me to add a macro for patent within the style but i have been struggling, any pointers?
Reference should include: 
Author, Initials. Year of publication. Title of patent. Patent number [Patent].
For example: Robertson, D. et al. 1990. 3-aryloxy-3-substituted propanamines. US 4,956,388 [Patent].

adam3smith:
The updated style is online at Zotero now and should be available in Mendeley shortly if it isn't already. 
The styles only get updated in the visual editor every couple of months, as that is a manual process. 

For the patent, the patent number is CSL "number" -- does that help or what are you struggling with?

YoungZoe:
Yes that's great. I will submit a new pull request.

PR #2678, opened 2017-05-10

libora:
https://forums.zotero.org/discussion/comment/275825/#Comment_275825

### DIFF
--- a/harvard-imperial-college-london.csl
+++ b/harvard-imperial-college-london.csl
@@ -141,7 +141,7 @@
       </if>
     </choose>
   </macro>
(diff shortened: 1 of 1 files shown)


harvard-imperial-college-london.csl
```diff
@@ -143,9 +143,11 @@
   </macro>
   <citation et-al-min="4" et-al-use-first="3" disambiguate-add-year-suffix="true" collapse="year">
```
damnation333: this should also have et-al-use-first="1". Thanks.

libora: OK, thanks

adam3smith:
Great thanks!

PR #2679, opened 2017-05-10
Update geochimica-et-cosmochimica-acta.csl

mjayvizzle:
Set the disambiguate-add-names to false, as it was incorrectly set to "true" before.

### DIFF
--- a/geochimica-et-cosmochimica-acta.csl
+++ b/geochimica-et-cosmochimica-acta.csl
@@ -112,7 +112,7 @@
       <text variable="page"/>
     </group>
   </macro>
-  <citation and="text" et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true">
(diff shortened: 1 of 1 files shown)


adam3smith:
Do we have documentation or an example from a published article for that or are you with the journals editorial team?

mjayvizzle:
There's no documentation, but here's an example from a published article:
[http://www.sciencedirect.com/science/article/pii/S0016703717302247](http://www.sciencedirect.com/science/article/pii/S0016703717302247)

Schrader et al., 2016a
D. Schrader, J. Davidson, T. McCoy
Widespread evidence for high-temperature formation of pentlandite in chondrites
Geochim. Cosmochim. Acta, 189 (2016), pp. 359–376

Schrader et al., 2016b
Schrader D., Fu R. and Desch S. (2016b) Evaluating chondrule formation models and the protoplanetary disk background temperature with low-temperature, sub-silicate solidus chondrule cooling rates. In Lunar and Planetary Science Conference, p. 1180.

These are then cited as Schrader et al., 2016a etc. in the text.

adam3smith:
perfect, thanks

PR #2680, opened 2017-05-10
Update harvard-cardiff-university.csl

YoungZoe:
Added patent macro

### DIFF
--- a/harvard-cardiff-university.csl
+++ b/harvard-cardiff-university.csl
@@ -85,7 +85,7 @@
   <macro name="title">
     <choose>
       <if type="bill book graphic legal_case legislation motion_picture report song thesis patent" match="any">
-        <text variable="title" font-style="italic" prefix=" " suffix="."/>
+        <text variable="title" font-style="italic"/>
       </if>
       <else>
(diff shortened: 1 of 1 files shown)


harvard-cardiff-university.csl
```diff
@@ -84,11 +84,11 @@
   </macro>
   <macro name="title">
     <choose>
-      <if type="bill book graphic legal_case legislation motion_picture report song thesis" match="any">
-        <text variable="title" font-style="italic"/>
+      <if type="bill book graphic legal_case legislation motion_picture report song thesis patent" match="any">
+        <text variable="title" font-style="italic" prefix=" " suffix="."/>
```
adam3smith: The prefix/suffix for the title of all other elements is coded in the bibliography section. Please remove them here and add them to the patent there. This probably works for other item types because Zotero/Mendeley are pretty good at preventing duplicate punctuation or spaces, but we try not to rely on that too much.

adam3smith:
Please have a look, see if this makes sense.

adam3smith:
Thanks!

PR #2685, opened 2017-05-13
Create chinese-gb7714-2015-numeric.csl

Summerslyb:
first commit for GB/T 7714-2015

### DIFF
--- a/chinese-gb7714-2015-numeric.csl
+++ b/chinese-gb7714-2015-numeric.csl
@@ -1,0 +1,361 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" default-locale="zh-CN" demote-non-dropping-particle="sort-only" version="1.0" xmlns="http://purl.org/net/xbiblio/csl">
+    <!-- 文件信息 -->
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks! Some comments above. Also, you you clarify how this related to the 1987 version of this standard? Are both needed or is this an update that supercedes the old version. Finally, what exactly are the Ref Labels? I'm just surprised they'd be in Latin letters. Is that really part of the style?

Summerslyb:
Yes. GB/T 7714-2015 is an update of GB/T 7714-2005 and GB/T 7714-1987.
It should supercede the old version, but the old version is also widely used now.
It's not recommended to remove old version I think.
ref... er... just prefix... means reference... it is not part of the standard.
If Ref Labels means the macro of refType or refMedia, it is the part of standard for indicating the type and the media of reference. Just Acronym.
For example, M means monograph and C means Collection.

rmzelle:
@Summerslyb, do you have time to address @adam3smith's remarks? Just ask if anything is unclear or if you need help.

PR #2686, opened 2017-05-15
Create scientific-review-engineering-and-environmental-sciences.csl

libora:
SREaES style is based on "APA no ampersand" with this changes:
1. "et al." from three authors
2. italicise with "." (without space between italicised letters)
The guide for authors only states "use APA style", but the comparison of published articles with outputs from APA style leads to changes of APA style described above.  Unfortunately, available papers on web pages of journal do not use the same format for volume and issue; so the style template for SREaES does not change this. I also did not change prefix "https://doi.org/" for DOI, though the published articles used to prefix "doi:".

### DIFF
--- a/scientific-review-engineering-and-environmental-sciences.csl
+++ b/scientific-review-engineering-and-environmental-sciences.csl
@@ -621,7 +621,7 @@
       </if>
     </choose>
   </macro>
-  <citation et-al-min="3" et-al-use-first="1" et-al-subsequent-min="3" et-al-subsequent-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year" givenname-disambiguation-rule="primary-name">
+  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year" givenname-disambiguation-rule="primary-name">
     <sort>
       <key macro="author"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks!

rmzelle:
@libora, this journal is bilingual, right? I saw some Polish and English articles.

libora:
@rmzelle yes

rmzelle:
@libora, okay, thanks. I added an (XML) comment to that effect: https://github.com/citation-style-language/styles/commit/1be95ff8698bb6f2d6857e638f86f4917d9e47d9

libora:
Similar comments can be added to a lot of journals from East Europe
For example, journals Vodohospodářské technicko-ekonomické informace (vodohospodarske-technicko-ekonomicke-informace.csl for Czech and Slovak language and vodohospodarske-technicko-ekonomicke-informace-en.csl for English) and Podzemná voda (podzemna-voda.csl) published in Czech, Slovak and English language.

rmzelle:
> vodohospodarske-technicko-ekonomicke-informace.csl for Czech and Slovak language

That CSL style is currently fixed to Czech, though (see https://github.com/citation-style-language/styles/blob/master/vodohospodarske-technicko-ekonomicke-informace.csl). If this CSL style can also be used in Slovak, we should probably delete its `default-locale`.

> podzemna-voda.csl) publish in Czech, Slovak and English language.

Similarly, this one is fixed to Slovak currently (https://github.com/citation-style-language/styles/blob/master/podzemna-voda.csl).

PR #2687, opened 2017-05-15
Update revue-archeologique-du-centre-de-la-france.csl

FrancoisC37:
I've done changes for multiple citation eg : (Lardé 2007 ; Landreau 2007)


rmzelle:
Thanks! (I made a few small edits to simplify the code, and to restore the XML declaration on top)

PR #2691, opened 2017-05-16
Spaces, points and desambigationg fix

hildogjr:
Quando mais de uma citação em uma única vez, devem ser separadas por ";".
Quando as citações de mesmo autor puderem ser diferenciadas apenas pelo ano, colapsar os nomes.

### DIFF
--- a/associacao-nacional-de-pesquisa-e-ensino-em-transportes.csl
+++ b/associacao-nacional-de-pesquisa-e-ensino-em-transportes.csl
@@ -412,7 +412,7 @@
       </else>
     </choose>
   </macro>
-  <citation et-al-min="6" et-al-use-first="1" et-al-subsequent-min="3" et-al-subsequent-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year" givenname-disambiguation-rule="primary-name">
+  <citation et-al-min="4" et-al-use-first="1" et-al-subsequent-min="4" et-al-subsequent-use-first="1" disambiguate-add-names="true" disambiguate-add-givenname="true" givenname-disambiguation-rule="primary-name" disambiguate-add-year-suffix="false" year-suffix-delimiter="; ">
     <sort>
       <key macro="author"/>
       <key macro="issued-sort"/>
@@ -425,7 +425,7 @@
       </group>
     </layout>
   </citation>
-  <bibliography et-al-use-last="true" entry-spacing="0" hanging-indent="true">
+  <bibliography hanging-indent="false" entry-spacing="12" subsequent-author-substitute="______" subsequent-author-substitute-rule="complete-all">
     <sort>
       <key macro="author"/>
       <key macro="issued-sort" sort="ascending"/>


associacao-nacional-de-pesquisa-e-ensino-em-transportes.csl
```diff
@@ -425,7 +425,7 @@
       </group>
     </layout>
   </citation>
-  <bibliography et-al-use-last="true" entry-spacing="0" hanging-indent="true">
+  <bibliography hanging-indent="false" entry-spacing="12" subsequent-author-substitute="______" subsequent-author-substitute-rule="complete-all">
```
adam3smith: `entry-spacing="12"` can't possible be right. Entry spacing describes the number of rows between bibliography entries and there clearly shouldn't be 12. Most likely 1 is correct?

associacao-nacional-de-pesquisa-e-ensino-em-transportes.csl
```diff
@@ -412,7 +412,7 @@
       </else>
     </choose>
   </macro>
-  <citation et-al-min="6" et-al-use-first="1" et-al-subsequent-min="3" et-al-subsequent-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year" givenname-disambiguation-rule="primary-name">
+  <citation et-al-min="4" et-al-use-first="1" et-al-subsequent-min="4" et-al-subsequent-use-first="1" disambiguate-add-names="true" disambiguate-add-givenname="true" givenname-disambiguation-rule="primary-name" disambiguate-add-year-suffix="false" year-suffix-delimiter="; ">
```
adam3smith: are you sure you want to remove the `collapse="year"` property? That would give you (Smith 1991, Smith 1992) instead of (Smith 1991, 1992). 
Also, you can remove `et-al-subsequent-min="4" et-al-subsequent-use-first="1" ` which is redundant now.

associacao-nacional-de-pesquisa-e-ensino-em-transportes.csl
hildogjr: Here I use  entry-spacing="12pt"`. But if I use `pt`, do not pass in the automatic test.

rmzelle: @hildogjr, sorry for the late reply, but see http://docs.citationstyles.org/en/stable/specification.html#whitespace for additional explanation of how the "entry-spacing" attribute can be used in CSL 1.0.1 styles. It may only be set to an integer (whole) number to specify how many lines separate bibliographic entries.

If you need more precise spacing you might be able to do this by manipulating the final output, e.g. with CSS in HTML or document styles in Word.

rmzelle:
Also, @hildogjr, does the link http://www.anpet.ufsc.br/ssat/interface/content/autor/arquivos/formato_trabalhos.pdf still work for you? (it doesn't work for me)

Or did you find the style guide somewhere else?

hildogjr:
When I implemented (modified) this style I use the last revision of the ABNT 6032-2002, modifying the to represent more close the standard style.

rmzelle:
> When I implemented (modified) this style I use the last revision of the ABNT 6032-2002

@hildogjr, sorry we're replying so slowly. Does the journal specify anywhere that they follow the ABNT 6032-2002 style? (without modifications?)

hildogjr:
I just used this in my master dissertation (the university rules was to use the ABNT style), not for any journal.

rmzelle:
@hildogjr, ah, sorry, I misspoke. I see now that ANPET is not a journal.

I'm still a little confused, though. If your university tells you to use the ABNT style, wouldn't one of our ABNT styles be a better fit for you? See https://www.zotero.org/styles?q=abnt

We would only update the ANPET style if we can confirm that our current style doesn't follow the ANPET guidelines. And our old link to the ANPET guidelines, http://www.anpet.ufsc.br/ssat/interface/content/autor/arquivos/formato_trabalhos.pdf, no longer works. If we want to update the ANPET style, we first need a new link to the current guidelines.

PR #2692, opened 2017-05-17
Create citizen-science-theory-and-practice.csl

damnation333:
via https://forums.zotero.org/discussion/61206/style-request-citizen-science-theory-and-practice
Very nice guidelines.
Adapted for everything in there.
Not sure why in the bibliography it's only giving me 7 authors and not all of them?!

### DIFF
--- a/citizen-science-theory-and-practice.csl
+++ b/citizen-science-theory-and-practice.csl
@@ -1,0 +1,267 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+  <info>
+    <title>Citizen Science: Theory and Practice</title>
+    <id>http://www.zotero.org/styles/citizen-science-theory-and-practice</id>
+    <link href="http://www.zotero.org/styles/citizen-science-theory-and-practice" rel="self"/>
+    <link href="http://www.zotero.org/styles/sociology-of-health-and-illness" rel="template"/>
+    <link href="http://theoryandpractice.citizenscienceassociation.org/about/submissions/#References" rel="documentation"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
Do you need the et-al conditions in the author-short and editor macro? Those are somehow messing with the et-al for the bibliography. While that's a bug, may not be worth getting into. Alternatively, you can set explicit et-als for the bibliography and make them arbitrary large (99 or 999)

citizen-science-theory-and-practice.csl
```diff
@@ -0,0 +1,267 @@
...
+    <link href="http://theoryandpractice.citizenscienceassociation.org/about/submissions/#References" rel="documentation"/>
+    <author>
+      <name>Patrick O'Brien</name>
+      <email>obrienpat86@gmail.com</email>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="sociology"/>
+    <category field="social_science"/>
+    <category field="anthropology"/>
+    <category field="medicine"/>
```
rmzelle: @damnation333, based on http://theoryandpractice.citizenscienceassociation.org/about/, maybe we should just limit this to the "science" category?

citizen-science-theory-and-practice.csl
```diff
@@ -0,0 +1,267 @@
...
+          <text term="available at" text-case="capitalize-first"/>
+          <text variable="URL"/>
+          <group delimiter=" " prefix="[" suffix="]">
+            <text term="accessed" text-case="capitalize-first"/>
+            <date form="text" variable="accessed"/>
+          </group>
+        </group>
+      </if>
+      <else>
+        <text variable="DOI" prefix="DOI: http://dx.doi.org/"/>
```
rmzelle: https://doi.org/ here please, even if the guidelines say otherwise

rmzelle:
Thanks!

PR #2693, opened 2017-05-17
Update harvard-cite-them-right.csl

damnation333:
Per https://forums.zotero.org/discussion/62911/style-request-cite-them-right-10th-edition


rmzelle:
Thanks!

PR #2697, opened 2017-05-23

libora:
Add DOI
https://forums.zotero.org/discussion/65777/style-error-mdpi-and-cells


adam3smith:
Thanks!

PR #2701, opened 2017-05-23
Update medical-physics.csl

wodellUF:
From the journal publishers: 
> References: We require complete author lists. If there are more than 6 author names in a reference, the first 3 author 
> names are retained with "et al" . Please update your references for completeness and accuracy. 

I have made this correction in the medical-physics.csl file by replacing one line in the authors section with the amended line.

### DIFF
--- a/medical-physics.csl
+++ b/medical-physics.csl
@@ -19,7 +19,7 @@
   </info>
   <macro name="author">
     <names variable="author">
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks! I moved this into the bibliography section, to make the style easier to read, but same idea exactly.

PR #2703, opened 2017-05-24
Create journal-of-computer-applications-in-archeology.csl

damnation333:
VIa https://forums.zotero.org/discussion/65737/style-request-journal-of-computer-applications-in-archaeology-caa
As pointed out in the thread, the style is very similar to citizen-science-theory-and-practice as they are both on Ubiquity press.
This journal is new and has not published anything. One to watch. I can imagine the missing dots are mistakes and we can make this style a dependent at some stage.

### DIFF
--- a/journal-of-computer-applications-in-archaeology.csl
+++ b/journal-of-computer-applications-in-archaeology.csl
@@ -37,7 +37,8 @@
       <name and="text" delimiter-precedes-last="never" initialize-with=" " name-as-sort-order="all"/>
       <label form="short" prefix=" "/>
       <substitute>
-        <text macro="editor"/>
+        <names variable="editor"/>
+        <names variable="translator"/>
         <choose>
           <if type="report" match="any">
             <text variable="publisher"/>
--- a/journal-of-computer-applications-in-archaeology.csl
+++ b/journal-of-computer-applications-in-archaeology.csl
@@ -3,8 +3,8 @@
   <info>
     <title>Journal of Computer Applications in Archaeology</title>
     <title-short>JCAA</title-short>
-    <id>http://www.zotero.org/styles/journal-of-computer-applications-in-archeology</id>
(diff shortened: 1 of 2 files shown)


journal-of-computer-applications-in-archeology.csl
```diff
@@ -0,0 +1,265 @@
...
+            </group>
+            <group delimiter=", ">
+              <text macro="issued" suffix=", "/>
+            </group>
+            <text macro="pages"/>
+          </group>
+        </else-if>
+        <else-if type="book graphic" match="any">
+          <group prefix=" " delimiter=". ">
+            <text variable="note"/>
```
damnation333: That was from the template. No clue. I've removed it.

journal-of-computer-applications-in-archeology.csl
```diff
@@ -0,0 +1,265 @@
...
+      </substitute>
+    </names>
+  </macro>
+  <macro name="author-short">
+    <names variable="author" font-weight="normal">
+      <name form="short" and="text" delimiter-precedes-last="never"/>
+      <substitute>
+        <names variable="editor"/>
+        <names variable="translator"/>
+        <choose>
```
damnation333: thanks. all done.

journal-of-computer-applications-in-archaeology.csl
```diff
@@ -0,0 +1,267 @@
...
+    <names variable="author">
+      <name and="text" delimiter-precedes-last="never" initialize-with=" " name-as-sort-order="all"/>
+      <label form="short" prefix=" "/>
+      <substitute>
+        <text macro="editor"/>
+        <choose>
+          <if type="report" match="any">
+            <text variable="publisher"/>
+          </if>
+        </choose>
```
rmzelle: @damnation333, I assume it's better to add the `<names variable="translator"/>` here as well. It should always be possible to match the in-text citation against the bibliographic entry, so the author-macros almost always need the exact same substitution logic.

rmzelle:
Thanks!

PR #2704, opened 2017-05-24
Create usda-forest-service-pnw-research-station.csl

sngordon:
USDA Forest Service PNW Research Station

### DIFF
--- a/usda-forest-service-pacific-northwest-research-station.csl
+++ b/usda-forest-service-pacific-northwest-research-station.csl
@@ -1,14 +1,16 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" default-locale="en-US" name-as-sort-order="all" demote-non-dropping-particle="never">
   <info>
-    <title>USDA Forest Service PNW Research Station</title>
+    <title>USDA Forest Service - Pacific Northwest Research Station</title>
     <title-short>USFS PNWRS</title-short>
-    <id>http://www.zotero.org/styles/usda-forest-service-pnw-research-station</id>
-    <link rel="self" href="http://www.zotero.org/styles/usda-forest-service-pnw-research-station"/>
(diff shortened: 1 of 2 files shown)


adam3smith:
Thanks. Could you explain what exactly the style is for? Also, the link to the documentation doesn't work for me. Is that an intranet link? Any public version of it?

rmzelle:
> I'm part of the development team for a more specific repository for which we want to use the CSL to serve formatted references dynamically for authors and the public

We typically only accept styles into the repository for which wide distribution makes sense (since they become available to hundreds of thousands of users of the various reference managers that use CSL), and normally don't take styles that are solely for the back-end of a single repository. But if this CSL style can be used for report writing as well, we can still take it.

> Sorry the style guide seems to only be on their intranet (but interesting that the link I used came back from a a google search), so I uploaded a copy to my google drive here:

Do you think it's okay to attach a copy to this GitHub issue thread as well? That way we won't lose access as easily.

rmzelle:
Closing because of inactivity. @sngordon, if you get back to us (and answer my question above) we can just reopen this.

sngordon:
Sorry for the delay, I was looking into the possibility of implementing this entirely on our server. Unfortunately that has proven difficult, plus it would be good to make the style available via CSL for those not using our particular database. Scientists from a number of agencies and universities contribute to US Forest Service publications. Attaching the style guide as requested.
[UsfsPnwAuthorsGuide2012.pdf](https://github.com/citation-style-language/styles/files/1321859/UsfsPnwAuthorsGuide2012.pdf)

rmzelle:
> Scientists from a number of agencies and universities contribute to US Forest Service publications.

Okay, if they can use this style, then we can certainly accept this.

Do you remember which CSL style you based your style on? We always try to add a "template" link to track a style's pedigree (see also https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link).

rmzelle:
Thanks!

Note that I changed the style title and ID slightly to be more in line with the rest of our styles.

PR #2706, opened 2017-05-31
Add composer.json

stucki:
See also https://github.com/citation-style-language/locales/pull/161

Note: I don't understand the separation between https://github.com/citation-style-language/styles vs. https://github.com/citation-style-language/styles-distribution. If the file is synched to this repository, the name would probably need to be adjusted...

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,19 @@
+{
+    "name": "citation-style-language/styles",
+    "description": "Citation Style Language (CSL) Styles",
+    "type": "library",
+    "license": "Creative Commons Attribution-ShareAlike 3.0 License",
+    "homepage": "http://citationstyles.org/",
+    "authors": [
+        {
+            "name": "Citation Style Language (CSL) Team",
+            "homepage": "http://citationstyles.org/about/#Credits"
+        }
+    ],
+    "extra": {
+        "branch-alias": {
+            "dev-master": "1.0.x-dev"
+        }
+    },
+    "require": {}
+}
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -36,7 +36,7 @@ UNUSED_MACROS_FILTER = %w{
 # These files and directories are ignored when checking for extra files
 EXTRA_FILES_FILTER = [
   'CONTRIBUTING.md', 'Gemfile', 'Gemfile.lock', 'README.md',
-  'dependent', 'Rakefile', 'renamed-styles.json'
+  'dependent', 'Rakefile', 'composer.json', 'renamed-styles.json'
 ]
 
 # These directories and their contents are ignored when checking for extra files


stucki:
@seboettg This should be useful for https://packagist.org/packages/seboettg/citeproc-php

seboettg:
Hey @stucki! 
Yes, a `composer.json` file would be a great in order to add styles and locales as dependencies for php environments such as [citeproc-php](https://github.com/seboettg/citeproc-php).
+1

rmzelle:
> I don't understand the separation between https://github.com/citation-style-language/styles vs. https://github.com/citation-style-language/styles-distribution

We should probably clarify this a bit more, but see https://github.com/citation-style-language/styles-distribution/blob/master/README.md and http://citationstyles.org/developers/#Code_repositories. Basically, after each commit to the "styles" repository, if all the Travis CI checks clear, the changes get pushed to the "styles-distribution" repository. The styles in the "styles-distribution" repo also get updated timestamps in their style metadata based on their file modification date. This was set up with help from the Zotero developers using the code at https://github.com/citation-style-language/distribution-updater, to tie in with their own front-end of our repository at https://zotero.org/styles.

So "styles-distribution" is a little safer (since changes that break the Travis CI test shouldn't propagate into it) and has better style metadata. It usually lags less than an hour behind "styles".

seboettg:
@rmzelle, a `composer.json` file would be a useful supplement for CSL users in a PHP environment. Composer is a great tool for managing required packages on PHP. But every repo that should be available on [Packagist](https://packagist.org) needs a `composer.json` file in its project root, in order to include the repo as a required package in a PHP project. Then I would have the possibility to add the repos "styles-distribution" and "locales" as direct requirement of the citeproc-php project, which then can be used out of the box.

rmzelle:
Is there a standard way to make composer.json somewhat self-documenting (e.g. by adding a link to https://getcomposer.org/), so that people who come across the file know what's it for? Taking a quick look at the JSON schema, it looks like we could only add it in "extra", but that's not very elegant.

Otherwise we'll just mention what it is in the README, I guess.

stucki:
@rmzelle Doesn't look like this is possible, see https://github.com/composer/composer/issues/1988
However, isn't it somewhat easy to look this up if someone likes to know? Just similar to `Gemfile` & `Rakefile`...

Thanks for the explanation about the separation of styles vs. styles-distribution. I was wondering if this wouldn't be possible using branches in just one repository? However, it's probably out of scope to change this now.
But for practical reasons we need to decide if the project name in composer.json should be `citation-style-language/styles` or `citation-style-language/styles-distribution` - or if it can be set manually in the distribution repo...

Last question: What should be the version number of this project? Should it just stay at `1.0.x` forever? Actually there's a 1.0 branch, which was last updated 5 years ago...

Thanks for your feedback!

rmzelle:
> Doesn't look like this is possible

Okay, thanks for checking.

> Thanks for the explanation about the separation of styles vs. styles-distribution. I was wondering if this wouldn't be possible using branches in just one repository?

I guess, but I'm pretty sure that back when we set this up GitHub e.g. didn't yet have its [protected branches](https://help.github.com/articles/about-protected-branches/) feature yet, and I think we were a bit afraid somebody would accidentally modify such a branch. It's much harder to accidentally mess up a different repo.

Also, I just remembered that "styles-distribution" also deletes all testing files and most other cruft, so it's a bit cleaner.

> But for practical reasons we need to decide if the project name in composer.json should be citation-style-language/styles or citation-style-language/styles-distribution - or if it can be different in both repositories?

Ideally all PHP devs would pull their styles from "styles-distribution", I think, so I would just put the composer.json file there. It looks like we might have to adjust the script for this, though, since it looks like it starts from scratch and e.g. writes out the README anew on each commit: https://github.com/citation-style-language/distribution-updater/blob/master/scripts/styles_distribution.py#L25.

> What should be the version number of this project? Should it just stay at 1.0.x forever? Actually there's a 1.0 branch, which was last updated 5 years ago...

The current version of CSL is 1.0.1. We branched off "1.0" when 1.0.1 was released.

adam3smith:
> Ideally all PHP devs would pull their styles from "styles-distribution", I think, so I would just put the composer.json file there. 

+1

rmzelle:
So, @stucki, we'd be happy to take this, but for the CSL styles it is probably best to put the composer.json at https://github.com/citation-style-language/distribution-updater/tree/master/scripts and update the script in that folder to add that file to the script's output, so that the file ends up in "styles-distribution". We can do that ourselves, but I'm a bit swamped, so that might take a while. If you want take a shot at it yourself, please do.

stucki:
@rmzelle Thanks for your feedback. Unfortunately I have no time (and need) for this right now. If someone else comes up with a PR, that would be great!

Deyter:
>But for practical reasons we need to decide if the project name in composer.json should be citation-style-language/styles or citation-style-language/styles-distribution - or if it can be different in both repositories?
I think it can be citation-style-language/styles in both repositories. But package on Packagist can be only on styles-distribution repository. If anyone wants to use styles repository he will do it with git composer management but not package on Packagist.

Deyter:
>The current version of CSL is 1.0.1. We branched off "1.0" when 1.0.1 was released.

But in the repository must be the version not of CSL (it is for branches) but the version of style files (for commit tags).
Your update script can simple increment build (or maybe revision) number and add a tag to every successful commit.

PR #2710, opened 2017-06-01
Create lcgc.csl

damnation333:
via https://forums.zotero.org/discussion/65252/style-request-lcgc#latest
Adapted for journal articles and book.

### DIFF
--- a/lcgc.csl
+++ b/lcgc.csl
@@ -1,0 +1,118 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+  <info>
+    <title>LCGC</title>
+    <id>http://www.zotero.org/styles/lcgc</id>
+    <link href="http://www.zotero.org/styles/lcgc" rel="self"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks!

lcgc.csl
```diff
@@ -0,0 +1,118 @@
...
+    <link href="http://www.zotero.org/styles/lcgc" rel="self"/>
+    <link href="http://www.zotero.org/styles/nature-no-superscript" rel="template"/>
+    <link href="http://www.chromatographyonline.com/lcgc-author-guidelines" rel="documentation"/>
+    <author>
+      <name>Patrick O'Brien</name>
+      <email>obrienpat86@gmail.com</email>
+    </author>
+    <category citation-format="numeric"/>
+    <category field="chemistry"/>
+    <issn>1471-6577</issn>
```
rmzelle: @damnation333, sorry for chiming in late, but I noticed that this ISSN is for LCGC Europe (http://journalseek.net/cgi-bin/journalseek/journalsearch.cgi?field=issn&query=1471-6577), while there is e.g. also LC-GC North America (http://journalseek.net/cgi-bin/journalseek/journalsearch.cgi?field=issn&query=1527-5949).

In this case, it's probably not worth it to make dependents for these individual journals, but I'll remove the ISSN from the style since it doesn't describe a specific journal.

PR #2712, opened 2017-06-02
Create european-journal-of-public-health.csl

damnation333:
via https://forums.zotero.org/discussion/65899/style-request
adapted for journals, chapters, books, web.


adam3smith:
Thanks!

PR #2713, opened 2017-06-02
Update journal-of-medical-internet-research.csl

damnation333:
via https://forums.zotero.org/discussion/comment/276685/#Comment_276685
Add DOI
Add webcite function (via extra field)
remove et al
add ISBN

### DIFF
--- a/journal-of-medical-internet-research.csl
+++ b/journal-of-medical-internet-research.csl
@@ -17,16 +17,19 @@
     <contributor>
       <name>Matt Tracy</name>
     </contributor>
+    <contributor>
+      <name>Patrick O'Brien</name>
+    </contributor>    
     <category citation-format="numeric"/>
     <category field="medicine"/>
     <issn>1439-4456</issn>
     <eissn>1438-8871</eissn>
-    <updated>2012-09-10T01:58:08+00:00</updated>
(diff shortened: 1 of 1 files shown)


journal-of-medical-internet-research.csl
```diff
@@ -218,6 +238,16 @@
       </if>
     </choose>
   </macro>
+  <macro name="webcite">
+    <choose>
+      <if type="webpage post-weblog" match="any">
+        <group delimiter=": " prefix=". ">
+          <text value="archived at" text-case="capitalize-first"/>
+          <text variable="note"/>
```
adam3smith: I'm skeptical about this. This is one journal; we can't really expect users to change their data entry patterns for this and on the negative side this is going to add all sorts of incomprehensible clutter to the bibliography. I'd say comment this out

damnation333: Is there any other way of solving this? Can users manually somehow add the webcite link then? usually people don't cite many website anyway, so it wouldn't be the majority of entries anyway.

adam3smith: I'd say just add it as part of the final edit. I don't see how we'd reasonably automate this, no. We could use the archive field, but that's not available for webpages, so of little use there.

adam3smith:
Thanks!

PR #2716, opened 2017-06-04
Update APA and CPoS

bwiernik:
For APA:
Switch the priority of URL and Archive name/number for theses and reports. URL is used much more commonly in practice and both are acceptable. URL is more user-friendly.

For CPoS:
Add support for dictionary and encyclopedia entries.


adam3smith:
Thanks!

PR #2722, opened 2017-06-07

ibadger:
Hi

I've updated the style. The main changes are substituting the title for author if it is missing, also I've added terms for accessed and no date so that they aren't written out as values. I've added values for when publisher and place of publication aren't available, I couldn't find terms to use for this. Also made improvements to map and a few minor additions in different places.

Let me know if anything needs to be amended.

Ian

### DIFF
--- a/harvard-university-for-the-creative-arts.csl
+++ b/harvard-university-for-the-creative-arts.csl
@@ -14,7 +14,7 @@
     <category citation-format="author-date"/>
     <category field="generic-base"/>
     <summary>University for the Creative Arts Harvard style</summary>
-    <updated>2017-06-19T14:14:37+00:00</updated>
+    <updated>2017-06-24T11:51:19+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale>
@@ -97,11 +97,11 @@
                   </if>
                 </choose>
                 <text macro="edition-no"/>
-                <group>
+                <group delimiter=": " suffix=".">
                   <choose>
-                    <if match="any" variable="collection-title">
-                      <text term="in" text-case="capitalize-first" suffix=": "/>
-                      <text variable="collection-title" font-style="normal" suffix="."/>
+                    <if type="song" match="any">
+                      <text term="in" text-case="capitalize-first"/>
+                      <text variable="collection-title"/>
                     </if>
                   </choose>
                 </group>
@@ -151,8 +151,10 @@
       </else-if>
       <else-if type="speech personal_communication" match="any">
         <group delimiter=" " prefix="[" suffix="]">
+          <names variable="recipient" prefix="Email sent to " suffix=",">
+            <name/>
+          </names>
           <text variable="genre" suffix=" "/>
-          <text variable="note"/>
           <text variable="event-place" suffix=":"/>
           <date form="text" variable="issued"/>
         </group>
@@ -189,37 +191,30 @@
   </macro>
   <macro name="publisher">
     <choose>
-      <if type="thesis">
-        <group delimiter=", ">
-          <text variable="publisher" suffix="."/>
-        </group>
+      <if match="any" variable="publisher">
+        <text variable="publisher" prefix=" " suffix="."/>
       </if>
-      <else-if type="manuscript">
-        <group delimiter=", ">
+      <else-if type="manuscript" match="any">
+        <group delimiter=". ">
+          <text variable="archive_location"/>
           <text variable="archive"/>
-          <group delimiter=". ">
-            <text variable="archive_location"/>
-            <text variable="publisher-place"/>
-          </group>
         </group>
       </else-if>
+      <else-if match="any" variable="URL"/>
       <else>
-        <choose>
-          <if match="any" variable="publisher" type="motion_picture broadcast report musical_score song entry-dictionary entry-encyclopedia">
-            <text variable="publisher" suffix="."/>
-          </if>
-          <else>
-            <text value="s.n." prefix="(" suffix=")."/>
-          </else>
-        </choose>
+        <text value="s.n." prefix=" (" suffix=")."/>
       </else>
     </choose>
   </macro>
   <macro name="publisher-place">
     <choose>
-      <if match="any" variable="publisher-place" type="broadcast motion_picture report musical_score song entry-dictionary entry-encyclopedia manuscript">
+      <if match="any" variable="publisher-place">
         <text variable="publisher-place"/>
       </if>
+      <else-if type="report entry-dictionary entry-encyclopedia motion_picture chapter speech song paper-conference article-journal book" match="all" variable="URL">
+        <text variable="publisher-place"/>
+      </else-if>
+      <else-if type="manuscript" match="any"/>
       <else>
         <text value="s.l." prefix="(" suffix=")"/>
       </else>
@@ -267,7 +262,7 @@
       <else-if type="patent">
         <text variable="number" suffix="."/>
       </else-if>
-      <else-if type="broadcast">
+      <else-if type="broadcast motion_picture">
         <choose>
           <if match="any" variable="collection-title container-title">
             <text term="in" text-case="capitalize-first" suffix=": "/>
@@ -355,14 +350,14 @@
       </if>
       <else-if type="book chapter paper-conference entry-dictionary entry-encyclopedia motion_picture report manuscript article bill map song" match="any">
         <group>
-          <text variable="event" suffix=". "/>
-          <group>
-            <text macro="publisher-place" suffix=": "/>
+          <text variable="event" suffix=" "/>
+          <group delimiter=":" suffix=".">
+            <text macro="publisher-place"/>
             <text macro="publisher"/>
-            <group>
-              <label prefix=". " variable="page" form="short"/>
-              <text variable="page" suffix="."/>
-            </group>
+          </group>
+          <group>
+            <label prefix="  " variable="page" form="short"/>
+            <text variable="page" suffix="."/>
           </group>
         </group>
       </else-if>
@@ -458,13 +453,11 @@
         </group>
         <text macro="legal-detail"/>
         <text macro="locator"/>
-        <group>
-          <choose>
-            <if type="map" match="any">
-              <text variable="collection-title" prefix="(" suffix=")"/>
-            </if>
-          </choose>
-        </group>
+        <choose>
+          <if type="map" match="any">
+            <text variable="collection-title" prefix="(" suffix=")"/>
+          </if>
+        </choose>
       </group>
       <text macro="online-access"/>
     </layout>
--- a/harvard-university-for-the-creative-arts.csl
+++ b/harvard-university-for-the-creative-arts.csl
@@ -14,7 +14,7 @@
     <category citation-format="author-date"/>
     <category field="generic-base"/>
     <summary>University for the Creative Arts Harvard style</summary>
-    <updated>2017-06-07T14:59:58+00:00</updated>
+    <updated>2017-06-08T09:18:29+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale>
@@ -34,18 +34,16 @@
           <label form="short" prefix=" "/>
           <et-al font-style="italic"/>
           <substitute>
+            <names variable="editor">
+              <name and="text" delimiter-precedes-last="never" et-al-min="3" et-al-use-first="1" initialize-with="." name-as-sort-order="all"/>
+              <label form="short" plural="never" text-case="lowercase" prefix=" (" suffix=")"/>
+              <et-al font-style="italic"/>
+            </names>
             <text variable="title"/>
           </substitute>
         </names>
       </else>
     </choose>
-  </macro>
-  <macro name="editor-detail">
-    <names variable="editor">
-      <name and="text" delimiter-precedes-last="never" initialize-with="." name-as-sort-order="all"/>
-      <label form="short" text-case="lowercase" prefix=" (" suffix=")"/>
-      <et-al font-style="italic"/>
-    </names>
   </macro>
   <macro name="year-date">
     <choose>
@@ -229,7 +227,7 @@
           <text term="in" text-case="capitalize-first" suffix=": "/>
           <names variable="editor" delimiter="." suffix=" ">
             <name and="text" delimiter-precedes-last="never" et-al-min="3" et-al-use-first="1" initialize-with="." name-as-sort-order="all"/>
-            <label form="short" plural="never" prefix=" (" suffix=")"/>
+            <label form="short" plural="never" text-case="lowercase" prefix=" (" suffix=")"/>
             <et-al font-style="italic"/>
           </names>
           <group delimiter=", ">
@@ -284,7 +282,7 @@
       <else-if type="article" match="any"/>
       <else-if type="map" match="any">
         <text variable="genre" prefix="[" suffix="]"/>
-        <text variable="scale" suffix="."/>
+        <text variable="scale" prefix=" " suffix="."/>
       </else-if>
       <else>
         <choose>
--- a/harvard-university-for-the-creative-arts.csl
+++ b/harvard-university-for-the-creative-arts.csl
@@ -14,12 +14,12 @@
     <category citation-format="author-date"/>
     <category field="generic-base"/>
     <summary>University for the Creative Arts Harvard style</summary>
-    <updated>2017-06-08T09:18:29+00:00</updated>
+    <updated>2017-06-19T14:14:37+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale>
     <terms>
-      <term name="accessed">accessed on </term>
+      <term name="accessed">accessed on</term>
       <term name="no date" form="short">s.d.</term>
     </terms>
   </locale>
@@ -29,7 +29,7 @@
         <text variable="title" font-style="italic" suffix="."/>
       </if>
       <else>
-        <names variable="author" font-style="normal" font-weight="normal">
+        <names variable="author">
           <name font-style="normal" and="text" delimiter-precedes-last="never" et-al-min="3" et-al-use-first="1" initialize-with="." name-as-sort-order="all"/>
           <label form="short" prefix=" "/>
           <et-al font-style="italic"/>
@@ -90,6 +90,12 @@
             <group delimiter=" ">
               <group delimiter=" ">
                 <text variable="title" font-style="italic" suffix="."/>
+                <choose>
+                  <if match="any" variable="volume">
+                    <text term="volume" form="short" text-case="capitalize-first"/>
+                    <text variable="volume" text-case="uppercase" suffix="."/>
+                  </if>
+                </choose>
                 <text macro="edition-no"/>
                 <group>
                   <choose>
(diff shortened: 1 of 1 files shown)


ibadger:
Attempted to fix issue with Editor Macro, reason initial commit failed

harvard-university-for-the-creative-arts.csl
```diff
@@ -110,11 +125,11 @@
...
         </choose>
       </if>
-      <else-if type="manuscript song post-weblog post personal_communication speech" match="any">
-        <group delimiter=". ">
+      <else-if type="manuscript song post-weblog post" match="any">
+        <group delimiter=" ">
           <text variable="genre" prefix="[" suffix="]"/>
           <text variable="medium" prefix="[" suffix="]"/>
-          <text variable="note" prefix="[" suffix="]"/>
+          <text variable="note" suffix="."/>
```
adam3smith: remind me why we're using note here? I'm trying hard to get rid of that in as many styles as possible, given that it produces really weird output for many users.

ibadger: Hi

Thanks for getting in touch.

I had the note field there because I am not able to use the Podcast source type to recreate the podcast example from our referencing guide, as there is no date option in Podcast. Our example looks like this:

Conger, C. and Ervin, C. (2015) _How to skate like a girl, part 1_. [audio podcast, online] 29 June 2015. At: http://www.stuffmomnevertoldyou.com/podcasts/how-to-skate-like-a-girlpart-1/ (Accessed on 8 June 2016)

The podcast option doesn't put a year at the start or a date after the medium. So as a way around I was thinking of using Audio Recording as this does have a year at the start of the reference, but with the Extra field for the full date of podcast. I used the Extra field as I don't want the date ending up in other sources which use Audio Recording, e.g. songs which don't require a full date in the reference.

Is there another way to do this? 

Alternatively I can take it out and if you'd rather it wasn't there. 

Ian

harvard-university-for-the-creative-arts.csl
```diff
@@ -6,47 +6,51 @@
...
     <category citation-format="author-date"/>
     <category field="generic-base"/>
     <summary>University for the Creative Arts Harvard style</summary>
-    <updated>2017-02-07T11:38:09+00:00</updated>
+    <updated>2017-06-08T09:18:29+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
+  <locale>
+    <terms>
+      <term name="accessed">accessed on </term>
```
rmzelle: @ibadger, we prefer that all terms are defined without any separating punctuation, like the trailing space here. It's better to move that over to where the term is used.

harvard-university-for-the-creative-arts.csl
```diff
@@ -6,47 +6,51 @@
...
+    </terms>
+  </locale>
   <macro name="cite-author">
     <choose>
       <if type="broadcast motion_picture legislation bill map legal_case" match="any">
         <text variable="title" font-style="italic" suffix="."/>
       </if>
       <else>
-        <names variable="author composer" font-style="normal" font-weight="normal">
+        <names variable="author" font-style="normal" font-weight="normal">
```
rmzelle: @ibadger, these `font-style="normal"` and `font-weight="normal"` are almost always unnecessary (they're only needed if you e.g. italicized a whole group and want to de-italicize one of the elements within that group).

harvard-university-for-the-creative-arts.csl
rmzelle: > Is there another way to do this?

You're using Zotero, right? In that case, it does look like you can't use "song" for Podcast since it lacks any date variables other than "accessed" (https://aurimasv.github.io/z2csl/typeMap.xml#map-podcast). There is an old Zotero ticket (https://github.com/citation-style-language/zotero-bits/issues/2) to add the CSL "issued" date variable to Zotero's podcast item type. Not sure if there are any other workarounds right now.

> Alternatively I can take it out and if you'd rather it wasn't there.

I think so, yes. We try to avoid having any reference-manager-specific hacks (like this one for Zotero) in our CSL repository styles, and the use of "note" is problematic because users may already use it for other purposes.

harvard-university-for-the-creative-arts.csl
```diff
@@ -123,7 +138,18 @@
...
+          <text variable="archive_location" prefix=" "/>
+        </group>
+        <text variable="archive" prefix=": " suffix="."/>
+      </else-if>
+      <else-if type="speech personal_communication" match="any">
+        <group delimiter=" " prefix="[">
+          <text variable="genre" suffix=" "/>
+          <text variable="note"/>
+          <text variable="event-place" suffix=":"/>
+          <date form="text" variable="issued" suffix="]"/>
```
rmzelle: @ibadger, this square bracket suffix should be on the group element. In general, it's best to move as many delimiting punctuation to group elements.

harvard-university-for-the-creative-arts.csl
```diff
@@ -167,20 +198,25 @@
...
+      </else>
+    </choose>
+  </macro>
+  <macro name="publisher-place">
+    <choose>
+      <if match="any" variable="publisher-place" type="broadcast motion_picture report musical_score song entry-dictionary entry-encyclopedia manuscript">
+        <text variable="publisher-place"/>
+      </if>
+      <else>
+        <text value="s.l." prefix="(" suffix=")"/>
```
rmzelle: @adam3smith, I haven't looked, but I don't think we have a ticket to add "sine loco" and "sine nomine" as CSL terms, but it probably makes sense to add them. Agreed?

rmzelle: (see https://en.wikipedia.org/wiki/Sine_nomine and https://en.wikipedia.org/wiki/List_of_Latin_phrases_(S)#sine_loco)

harvard-university-for-the-creative-arts.csl
```diff
@@ -191,13 +227,14 @@
...
             <name and="text" delimiter-precedes-last="never" et-al-min="3" et-al-use-first="1" initialize-with="." name-as-sort-order="all"/>
-            <label form="short" plural="never" prefix=" (" suffix=")"/>
+            <label form="short" plural="never" text-case="lowercase" prefix=" (" suffix=")"/>
             <et-al font-style="italic"/>
           </names>
           <group delimiter=", ">
             <group font-style="normal" font-variant="normal">
               <text variable="container-title" font-style="italic" suffix="."/>
               <text variable="collection-title" font-style="italic"/>
+              <text variable="volume" prefix=" Vol " suffix="."/>
```
rmzelle: @ibadger, you can use the short form of the "volume" term here (see https://github.com/citation-style-language/locales/blob/96042867e58040158417e457e3cb78c73d36b029/locales-en-GB.xml#L227), with a new group to print the term conditionally. Let me know if you need help with this.

harvard-university-for-the-creative-arts.csl
```diff
@@ -223,19 +260,30 @@
         <text variable="number" suffix="."/>
       </else-if>
       <else-if type="motion_picture broadcast">
-        <text variable="medium" prefix="[" suffix="] "/>
+        <choose>
+          <if match="any" variable="collection-title container-title">
+            <text term="in" text-case="capitalize-first" suffix=": "/>
+            <text variable="collection-title" font-style="italic" suffix="."/>
+            <text variable="container-title" font-style="italic" suffix="."/>
```
rmzelle: @ibadger, do you know if there are cases in which the items may have both titles (collection and container)? Would you get correct output in that case?

harvard-university-for-the-creative-arts.csl
ibadger: I have removed this

harvard-university-for-the-creative-arts.csl
ibadger: Thanks for the tip, I have removed this

harvard-university-for-the-creative-arts.csl
ibadger: I will leave this until the CSL changes and then update to terms if that is ok?

harvard-university-for-the-creative-arts.csl
ibadger: Thanks for the tip. I've added the way you suggested, I think - could you let me know if there is an issue? I've also added it in Book as we quite a few books of multiple volumes where it may also be used

harvard-university-for-the-creative-arts.csl
ibadger: I have removed motion picture from this section. I have left broadcast with the variable though as we have instances of students referencing sections within broadcasts, e.g. a tv programme with different reports or sections on different topics.
If this isn't ok let me know

harvard-university-for-the-creative-arts.csl
ibadger: I have done this, thanks for the advice

harvard-university-for-the-creative-arts.csl
ibadger: I have removed these two examples. I found a couple of other uses of normal in the code but wasn't sure if they could be removed or not - on lines 33 and 104

harvard-university-for-the-creative-arts.csl
```diff
@@ -231,10 +236,15 @@
...
+            <group font-style="normal" font-variant="normal" delimiter=" ">
               <text variable="container-title" font-style="italic" suffix="."/>
               <text variable="collection-title" font-style="italic"/>
-              <text variable="volume" prefix=" Vol " suffix="."/>
+              <choose>
+                <if match="any" variable="volume">
+                  <text term="volume" form="short" text-case="capitalize-first"/>
+                  <text variable="volume" text-case="uppercase" prefix=" " suffix="."/>
+                </if>
+              </choose>
```
rmzelle: @ibadger, because `<group/>` acts as an implicit conditional (see http://docs.citationstyles.org/en/stable/specification.html#group), you can simplify

```xml
              <choose>
                <if match="any" variable="volume">
                  <text term="volume" form="short" text-case="capitalize-first"/>
                  <text variable="volume" text-case="uppercase" prefix=" " suffix="."/>
                </if>
              </choose>
```

to

```xml
              <group>
                <text term="volume" form="short" text-case="capitalize-first"/>
                <text variable="volume" text-case="uppercase" prefix=" " suffix="."/>
              </group>
```

(the group won't output anything if all the variables it calls have no value)

rmzelle: It's also better to put the delimiting punctuation on the group, i.e.:

```xml
                <text term="volume" form="short" text-case="capitalize-first"/>
                <text variable="volume" text-case="uppercase"/>
              </group>
```

ibadger: Thanks I didn't realise group acted like that, I think that I've done what you suggested.

really appreciate your help

harvard-university-for-the-creative-arts.csl
```diff
@@ -110,20 +131,30 @@
...
+          <text variable="dimensions" suffix="."/>
           <text variable="note" prefix=" " suffix="."/>
+          <text variable="archive_location" prefix=" "/>
+        </group>
+        <text variable="archive" prefix=": " suffix="."/>
+      </else-if>
+      <else-if type="speech personal_communication" match="any">
+        <group delimiter=" " prefix="[" suffix="]">
+          <text variable="genre" suffix=" "/>
+          <text variable="note"/>
```
rmzelle: @ibadger, do you still need "note" for "speech personal_communication" and "graphic" (above)?

harvard-university-for-the-creative-arts.csl
```diff
@@ -398,6 +458,13 @@
         </group>
         <text macro="legal-detail"/>
         <text macro="locator"/>
+        <group>
+          <choose>
+            <if type="map" match="any">
+              <text variable="collection-title" prefix="(" suffix=")"/>
+            </if>
+          </choose>
+        </group>
```
rmzelle: You can delete the group here, which doesn't do anything.

harvard-university-for-the-creative-arts.csl
rmzelle: > I will leave this until the CSL changes and then update to terms if that is ok?

Yes, definitely.

ibadger:
Sorry for the earlier one I made a mistake

harvard-university-for-the-creative-arts.csl
```diff
@@ -167,20 +203,25 @@
...
+          </else>
+        </choose>
+      </else>
+    </choose>
+  </macro>
+  <macro name="publisher-place">
+    <choose>
+      <if match="any" variable="publisher-place" type="broadcast motion_picture report musical_score song entry-dictionary entry-encyclopedia manuscript">
+        <text variable="publisher-place"/>
+      </if>
```
rmzelle: @ibadger, you can test this yourself, but I'm not sure this will work correctly. With `match="any"`, if-statement will report true whenever either a given item has a value for "publisher-place", or if it's of one of the item types you test for with `type`. Is that the logic you're after? E.g. a report without a value for "publisher-place" won't print "s.l.".

harvard-university-for-the-creative-arts.csl
```diff
@@ -82,9 +90,22 @@
...
+                </choose>
                 <text macro="edition-no"/>
+                <group>
+                  <choose>
+                    <if match="any" variable="collection-title">
+                      <text term="in" text-case="capitalize-first" suffix=": "/>
+                      <text variable="collection-title" font-style="normal" suffix="."/>
+                    </if>
+                  </choose>
+                </group>
```
rmzelle: @ibadger, can you rewrite this as well? You can keep the group, but get rid of the whole `<choose/>` construct.

harvard-university-for-the-creative-arts.csl
ibadger: It was another work around I'll take it out

harvard-university-for-the-creative-arts.csl
ibadger: I use the visual editor as I'm not very good with coding, so bear with me.

I only want s.l. to appear on a source where I would expect a publisher location and it doesn't have one.

At the moment the code will put in an s.l. on a book when the publisher field is blank, which is what I want.

Is there a better way of doing this?

harvard-university-for-the-creative-arts.csl
ibadger: I'm not too sure what I'm doing with this part, if I remove line 467 it won't validate. Do I need to remove more than that one line?

harvard-university-for-the-creative-arts.csl
ibadger: Hi

Is it the conditional that I need to remove, if so is there a way of making the collection title only appear for a map without a conditional before it? That is what I am trying to achieve here, the map collection appearing at the end of the reference after the publisher

harvard-university-for-the-creative-arts.csl
rmzelle: > I only want s.l. to appear on a source where I would expect a publisher location and it doesn't have one.
>
> At the moment the code will put in an s.l. on a book when the publisher field is blank, which is what I want.

Right. But what do you want to happen for e.g. reports, which have the publisher-place field (at least in Zotero: https://aurimasv.github.io/z2csl/typeMap.xml#map-report). With your current code, a report without a publisher-place will print nothing, instead of "(s.l.)", because all report items test positive for the if-statement.

harvard-university-for-the-creative-arts.csl
rmzelle: You should be able to replace

```xml
        <group>
          <choose>
            <if type="map" match="any">
              <text variable="collection-title" prefix="(" suffix=")"/>
            </if>
          </choose>
        </group>
```

with

```xml
        <choose>
          <if type="map" match="any">
            <text variable="collection-title" prefix="(" suffix=")"/>
          </if>
        </choose>
```

(you need to delete both the starting `<group>` and closing `</group>` elements)

rmzelle:
> Sorry for the earlier one I made a mistake

No problem. Let me know if I'm too demanding and giving you a headache. We just try to give our more frequent contributors like you some tips on how to code their styles better, but we understand if you are in a hurry and just want this accepted (in which case we can also make some of these changes ourselves).

harvard-university-for-the-creative-arts.csl
rmzelle: In

```xml
                <group>
                  <choose>
                    <if match="any" variable="collection-title">
                      <text variable="collection-title" font-style="normal" suffix="."/>
                    </if>
                  </choose>
                </group>
```

you're testing whether the "collection-title" variable has a value, and if so, you want to print it with a term. But as I mentioned in another comment, groups already act as implicit conditionals, so you can simplify this to:

```xml
                <group delimiter=": " suffix=".">
                </group>
```

> Is it the conditional that I need to remove, if so is there a way of making the collection title only appear for a map without a conditional before it? That is what I am trying to achieve here, the map collection appearing at the end of the reference after the publisher

Well, the only role of the `<choose/>` here is to suppress a dangling "in" when there is no value for "collection-title", but just using the group works as well. It should have no other impact.

harvard-university-for-the-creative-arts.csl
ibadger: Thanks for that

harvard-university-for-the-creative-arts.csl
ibadger: I understand now, thanks for that. I mainly edit on the visual editor. I need to get in the habit of looking at the differences I make to the code view when I edit something.

Thanks again

harvard-university-for-the-creative-arts.csl
ibadger: Thanks for your patience, I realise that what I was doing looked a bit confusing and I wasn’t explaining myself very well.

I have been had a thorough look at this, looking at the CSL-Zotero Mapping and at our referencing guide and what I am aiming to achieve is:

For the following CSL types: Thesis, Manuscript, Broadcast, article-newspaper
•	If the Place field is empty to not add s.l. as we don’t require a place field

For the following CSL types: Book, Entry-Dictionary, Entry-Encyclopedia, Motion Picture, Chapter, Speech, Song, Bill, Paper Conference, Report
•	If there is no url and the place field is blank I would like to see s.l. displayed
•	If there is a url and the place field is blank I would like to see nothing displayed (I wouldn’t like to see s.l.)
•	If there is a url and the place field is filled in I would like to see the place field displayed

Is this possible? Do you have any tips?


I really appreciate your help

ibadger:
Hi

I have implemented all of the changes that you suggested. I have been working through lots of edits, so if you think I missed something let me know

Also attempted to improve the section that you queried about what happens with missing publisher location.

Many thanks for all your help, I really appreciate it.

Ian

rmzelle:
Thanks! (just create another pull request for any followup changes, if you have any)

harvard-university-for-the-creative-arts.csl
rmzelle: So, you want the publisher-place variable printed, with a fallback to "s.l." unless the item, a) is of the type thesis, manuscript, broadcast, article-newspaper, or b) has a non-empty URL variable. I haven't tested it, but I think you could achieve that with the more compact:

```xml
  <macro name="publisher-place">
    <choose>
      <if variable="publisher-place">
      </if>
      <else-if type="thesis manuscript broadcast article-newspaper" variable="URL" match="none">
        <text value="s.l." prefix="(" suffix=")"/>
      </else-if>
    </choose>
  </macro>
```

ibadger:
Thanks for your help.
I'll add the code you suggested for s.l. the next time I push an update through

harvard-university-for-the-creative-arts.csl
rmzelle: > @adam3smith, I haven't looked, but I don't think we have a ticket to add "sine loco" and "sine nomine" as CSL terms, but it probably makes sense to add them. Agreed?

@adam3smith, bump. Only ticket I could find is https://github.com/citation-style-language/schema/issues/90#issuecomment-4329346, and there are at least a bunch of styles that that use the "sine loco" (https://github.com/search?q=org%3Acitation-style-language+loco&type=Code) and "sine nomine" (https://github.com/search?utf8=%E2%9C%93&q=org%3Acitation-style-language+nomine&type=Code).

rmzelle: (ticket created: https://github.com/citation-style-language/schema/issues/138)

PR #2723, opened 2017-06-08
Errors in layout of footnotes

guschartier:
Replacing macro at the end


rmzelle:
@guschartier, can we ignore this pull request? It looks like you're reverting the changes made in https://github.com/citation-style-language/styles/pull/2688.

Issue #2495, opened 2017-02-13
Plan for removal of Zotero's "Include URLs of paper articles in references" toggle


rmzelle:
Dan Stillman indicated he wouldn't mind removing the "Include URLs of paper articles in references" toggle in the "Styles" > "Cite" tab of the Zotero preferences (see https://forums.zotero.org/discussion/comment/269136/#Comment_269136 and later comments). It was added back in 2008, and is unchecked by default (https://github.com/zotero/zotero/blob/dcd1da70af223f41aa1b827387d0cc69ef828e69/defaults/preferences/zotero.js#L111). So by default, Zotero overrides the behavior specified in styles by suppressing the "URL" CSL variable when the "page" CSL variable isn't empty for the "article-journal", "article-newspaper", and "article-magazine" CSL item types (https://github.com/zotero/zotero/blob/36b6722c5ca4274e21c7679aad83be3d660a471b/chrome/content/zotero/xpcom/cite.js#L537):

> When this option is disabled, Zotero includes URLs when citing journal, magazine, and newspaper articles only if the article does not have a page range specified.

@adam3smith and I rather see the toggle (and its effects) removed. To minimize disruption, we probably first need to put the logic into styles where necessary.

rmzelle:
I guess the first step would be using some regex to identify how many styles print the "URL" variable, and determine how many of those do and do not print this conditionally based on a populated "page" field. For those without a conditional, we'll have to check the style guides to see if one is needed.

bwiernik:
During that effort, I think it would be a good and efficient idea to make the URL code construct a DOI-based URL if available and then only print the URL if no DOI is available.

bwiernik:
There currently are 1556 styles referring to the URL variable. Of these, there are four common calling structures:

1. Explicitly testing for page and including URL if it is absent (this is rare)
1. Including the URL only for webpage and post-weblog (this is most common)
1. Including the URL and adding the `internet` term in brackets to the title when URL is present (196 styles)
1. Including the URL unconditionally

For 1 and 2, nothing needs to change. For 3, I think it is reasonable to assume that the intended behavior is to include the URL if no page numbers are present (we have frequent complaints on the Zotero forums about this word being added). For 4, we could check style guides, but this appears to be a fairly small number of styles that I don't think it would be too disruptive to remove the option before addressing it. So the major step to take, I think, would be to adjust the 196 styles including the `internet` term if URL is present.

adam3smith:
[...]

PR #2730, opened 2017-06-11

ahahl:
Minor changes and improvements for some types (post, bill...)
I also had to make a workout for DOIs not to appear when pages variable exists (it would be nice if this could be included directly in Zotero's functionality concerning URLs though).
Thanks !

### DIFF
--- a/ens-de-lyon-centre-d-ingenierie-documentaire.csl
+++ b/ens-de-lyon-centre-d-ingenierie-documentaire.csl
@@ -19,8 +19,7 @@
     <category field="humanities"/>
     <category field="political_science"/>
     <updated>2017-01-25T13:45:07+00:00</updated>
-    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a
-      Creative Commons Attribution-ShareAlike 3.0 License</rights>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="fr">
     <terms>
--- a/ens-de-lyon-centre-d-ingenierie-documentaire.csl
+++ b/ens-de-lyon-centre-d-ingenierie-documentaire.csl
@@ -19,7 +19,8 @@
     <category field="humanities"/>
     <category field="political_science"/>
     <updated>2017-01-25T13:45:07+00:00</updated>
-    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a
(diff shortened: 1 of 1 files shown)


ens-de-lyon-centre-d-ingenierie-documentaire.csl
```diff
@@ -109,7 +109,7 @@
                 <name-part name="family" font-variant="small-caps"/>
               </name>
               <et-al font-style="italic"/>
-              <label form="short" prefix="&#160;(" suffix=".)"/>
+              <label form="short" prefix=" (" suffix=".)"/>
```
rmzelle: @ahahl, are you unescaping these non-breaking spaces on purpose (or replacing them by regular spaces)?

rmzelle:
> (it would be nice if this could be included directly in Zotero's functionality concerning URLs though)

We're actually considering asking Zotero to remove their special "don't-print-the-URL-if-there-are-pages" toggle, since it's logic can and probably should be expressed in the style itself, and because not all CSL-based reference managers have this toggle. (see https://github.com/citation-style-language/styles/issues/2495 for discussion)

ens-de-lyon-centre-d-ingenierie-documentaire.csl
rmzelle: (if you aren't, could you put the escape codes back? We very easily lose unescaped special spaces, which often turn into regular spaces when we copy and paste style code around)

ahahl: Hi rmzelle, thanks for your comments. My intention is to have normal spaces here, as I see no justification for having non-breaking spaces instead.

ahahl:
And considering the "don't-print-the-URL-if-there-are-pages" function, removing it is a solution indeed, which would avoid several problems. Although I know some users who can't edit CSL styles do like it, it is also misleading (especially as it does not manage the DOI in the same way as the URL).
About the ens-de-lyon-centre-d-ingenierie-documentaire.csl style, we are considering with coeditor smarchand creating a web version of the style (with all URL when available) and keep this one for more usual printed bibliographies.

ens-de-lyon-centre-d-ingenierie-documentaire.csl
rmzelle: Okay, thanks, just making sure this was intentional.

rmzelle:
Thanks!

> we are considering with coeditor smarchand creating a web version of the style (with all URL when available)

Note that we typically only accept CSL styles into the repository for which widespread distribution makes sense. If you make a web version for e.g. just the website of your organization, it might make more sense to distribute the style yourself as necessary.

ahahl:
Hi again :)
It seems the update hasn't been taken into account on Zotero style repository, did I miss a step somewhere?
Sorry for troubling, and thanks for your help!

rmzelle:
Looks like something is broken. We've pinged the Zotero devs, who usually fix stuff like this on the same day.

rmzelle:
@ahahl, the updated style is now in the Zotero repository as well. Thanks for letting us know something was wrong.

PR #2732, opened 2017-06-13
Improve utah-geological-survey.csl. 

inkenbrandt:
Series & series number in book section. This change will include the series and series number of the book from which the chapter was taken.  Many of the book sections that the UGS cites are from books that are in a collection.


rmzelle:
Thanks!

PR #2733, opened 2017-06-13
Create journal-of-veterinary-diagnostic-investigation.csl

damnation333:
via https://forums.zotero.org/discussion/65627/style-request-journal-of-veterinary-diagnostic-investigation
Adapted for journals, books, chapters, web.


rmzelle:
Thanks!

PR #2735, opened 2017-06-14
Create annals-of-the-new-york-academy-of-sciences.csl

damnation333:
via https://forums.zotero.org/discussion/65966/style-request-annals-of-the-new-york-academy-of-sciences
guidelines: https://www.nyas.org/Publications/Annals/AuthorGuidelines.aspx
sample paper: http://onlinelibrary.wiley.com/doi/10.1111/nyas.12785/epdf


rmzelle:
Thanks!

PR #2743, opened 2017-06-16
Create prehospital-emergency-care.csl

damnation333:
via https://forums.zotero.org/discussion/65961/style-request-prehospital-emergency-care#latest
Example article: http://www.tandfonline.com/doi/full/10.1080/10903127.2017.1315203
The style guides are not worthy a look. Style based on papers.


rmzelle:
Thanks as always!

PR #2744, opened 2017-06-16
Create ksce-journal-of-civil-engineering.csl

damnation333:
via https://forums.zotero.org/discussion/52844/style-request-korean-society-of-civil-engineering
adapted for journals, books, chapters

### DIFF
--- a/ksce-journal-of-civil-engineering.csl
+++ b/ksce-journal-of-civil-engineering.csl
@@ -132,8 +132,8 @@
             <text variable="genre"/>
             <text variable="publisher"/>
             <group delimiter=" ">
-              <label plural="never" variable="page" form="short"/>
-              <number/>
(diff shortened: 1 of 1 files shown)


rmzelle:
@damnation333, the title and documentation link are for the "KSCE Journal of Civil Engineering", but the request was for the Journal of the Korean Society of Civil Engineers (http://www.kscejournal.or.kr/plugins/sites/ksce/volume.html).

We can still take this, of course (looks like at least the ISSN and eISSN need to be updated, per http://www.springer.com/engineering/civil+engineering/journal/12205).

damnation333:
Thank you for spotting that. I've fixed that now and will get working on the requested style.

rmzelle:
Thanks 😆

PR #2745, opened 2017-06-16
Create cultural-geographies.csl

damnation333:
via https://forums.zotero.org/discussion/61978/style-request-humane-system-for-cultural-geographies
Adapted for journal-articles, book, chapter


rmzelle:
Thanks!

PR #2746, opened 2017-06-17
Create journal-of-the-society-of-civil-engineers.csl

damnation333:
https://forums.zotero.org/discussion/52844/style-request-korean-society-of-civil-engineering

(see also: https://github.com/citation-style-language/styles/pull/2744 for the creation of a style for the KSCE Journal of Civil Engineering)


rmzelle:
Little boo-boo with the style ID, but thanks!

Issue #2705, opened 2017-05-31
Style error: American Journal of Hypertension 


adam3smith:
Instructions for authors here: https://academic.oup.com/ajh/pages/general_instructions#FormatofManuscripts
they refer to both Vancouver and AMA, but the style, in particular for books and chapters, follows neither exactly. Journal articles look like AMA

rmzelle:
(this is currently a dependent style: https://github.com/citation-style-language/styles/blob/master/dependent/american-journal-of-hypertension.csl)

@damnation333, feel free to take this if you're interested :).

damnation333:
Cool. Hadn't seen this one. Will add it to the list.

PR #2750, opened 2017-06-19
Update and rename dependent/american-journal-of-hypertension.csl to a…

damnation333:
via https://github.com/citation-style-language/styles/issues/2705


american-journal-of-hypertension.csl
```diff
@@ -0,0 +1,223 @@
...
+                    <date-part name="month" suffix=" "/>
+                    <date-part name="year"/>
+                  </date>
+                </else>
+              </choose>
+              <text variable="volume" prefix=" "/>
+            </group>
+            <text variable="page" prefix=":"/>
+          </group>
+          <text variable="DOI" prefix=" &lt;" suffix="&gt;"/>
```
rmzelle: @damnation333, where did you find this? I e.g. found this open access article, which doesn't print DOIs at all (https://academic.oup.com/ajh/article/30/7/683/3739803/An-Assessment-of-the-Accuracy-of-Home-Blood), and the instructions also don't seem to mention them (https://academic.oup.com/ajh/pages/general_instructions#FormatofManuscripts).

Unless they're very clear about this, I'd just use the URL format here.

damnation333:
@rmzelle 
I had seen it here: Reference 4 https://academic.oup.com/ajh/article/30/6/587/3064943/Association-of-Blood-Pressure-Trajectory-With

rmzelle:
@damnation333, they still add the URL prefix there: `<http://dx.doi.org/10.1001/jama.2016.7050>`, in addition to the angle brackets. So I would go with `<https://doi.org/...>` here (especially since they already use ` DOI: https://doi.org/10.1093/ajh/hpx028` elsewhere on the page.

rmzelle:
Thanks!

PR #2751, opened 2017-06-20
Create indiana.csl

damnation333:
via https://forums.zotero.org/discussion/61615/style-request-indiana

If somebody can tell me how I can get the access-date to 20.06.2017 instead of 20/06/2017 for websites and how to get the bibliography into those blocks again, I would be very grateful. Like this:
Authors
  Year  other stuff

Also, I've left the locale unset, as I found an article in English. They're guidelines are in Spanish, so I assume they do publish in Spanish also.

### DIFF
--- a/indiana.csl
+++ b/indiana.csl
@@ -1,0 +1,358 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" and="symbol" delimiter-precedes-et-al="never" delimiter-precedes-last="never" demote-non-dropping-particle="never">
+  <info>
+    <title>Indiana</title>
+    <id>http://www.zotero.org/styles/indiana</id>
+    <link href="http://www.zotero.org/styles/indiana" rel="self"/>
+    <link href="http://www.zotero.org/styles/zeitschrift-fur-soziologie" rel="template"/>
+    <link href="http://www.zfs-online.org/index.php/zfs/information/authors" rel="documentation"/>
+    <author>
+      <name>Patrick O'Brien</name>
+      <email>obrienpat86@gmail.com</email>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="humanities"/>
+    <issn>0341-8642</issn>
+    <eissn>2365-2225</eissn>
+    <updated>2017-06-20T15:03:35+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <macro name="container-contributors">
+    <choose>
+      <if type="chapter paper-conference" match="any">
+        <names variable="editor" delimiter=", " suffix=": ">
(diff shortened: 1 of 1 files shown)


indiana.csl
```diff
@@ -0,0 +1,358 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" and="symbol" delimiter-precedes-et-al="never" delimiter-precedes-last="never" demote-non-dropping-particle="never">
+  <info>
+    <title>Indiana</title>
+    <id>http://www.zotero.org/styles/indiana</id>
+    <link href="http://www.zotero.org/styles/indiana" rel="self"/>
+    <link href="http://www.zotero.org/styles/zeitschrift-fur-soziologie" rel="template"/>
+    <link href="http://www.zfs-online.org/index.php/zfs/information/authors" rel="documentation"/>
```
adam3smith: this is still the old link

indiana.csl
```diff
@@ -0,0 +1,358 @@
...
+      </else-if>
+      <else-if type="bill legislation" match="any">
+        <date variable="issued" prefix=" (" suffix=")">
+          <date-part name="year"/>
+        </date>
+      </else-if>
+      <else-if type="webpage post-weblog" match="any">
+        <group delimiter=" ">
+          <text variable="URL" prefix=" &lt;" suffix="&gt;"/>
+          <date form="numeric" variable="accessed" prefix="(" suffix=")"/>
```
adam3smith: two options to get this as 20.06.2017:
hardcode

```
<date variable="accessed">
  <date-part name="day" form="numeric-leading-zeros" suffix="."/>
  <date-part name="month" form="numeric-leading-zeros" suffix="."/>
  <date-part name="year"/>
</date>
```
Alternatively, set default-locale to German or define the numeric format like this in cs:terms

adam3smith:
For your other question, setting a group around everything except the author and setting `display="block"` will do what you want.

rmzelle:
> Also, I've left the locale unset, as I found an article in English. They're guidelines are in Spanish, so I assume they do publish in Spanish also.

On the guidelines page, you can select a different language in the right-hand menu bar. E.g. the German guidelines say this journal publishes in Portuguese and German as well (I figure I don't have to translate this for you):

> INDIANA erscheint jährlich und veröffentlicht Originalbeiträge und Essays in spanischer, portugie­sischer, englischer und deutscher Sprache

rmzelle:
Thanks!

damnation333:
So we add the polyglot line when it's more than 1 language? I'll keep that in mind for next ones. :)

rmzelle:
Well, the idea is that we add a `default-locale` to all styles for journals that only publish in a single language. I've been adding the "polyglot" XML comment to styles for multi-lingual journals, so it's clear a) which languages the journal publishes in, and b) that the absence of a `default-locale` attribute is intentional.

rmzelle:
(see also https://github.com/citation-style-language/styles/search?utf8=%E2%9C%93&q=polyglot&type= for styles with the line)

PR #2752, opened 2017-06-20
Create cranfield-university-author-date.csl

searlesr:
Modifications to long author name style

### DIFF
--- a/cranfield-university-author-date.csl
+++ b/cranfield-university-author-date.csl
@@ -1,0 +1,387 @@
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" and="text" name-as-sort-order="all" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+  <info>
+    <title>Cranfield University (Author-Date)</title>
+    <title-short>Cranfield (Author-Date)</title-short>
+    <id>http://csl.mendeley.com/styles/454899671/cranfield-university-author-date</id>
+    <link href="http://www.zotero.org/styles/cranfield-university-author-date" rel="self"/>
+    <link href="http://www.zotero.org/styles/harvard-staffordshire-university" rel="template"/>
+    <link href="http://www.som.cranfield.ac.uk/som/dinamic-content/media/mirc/Referencing/AuthorDateReferencingGuide2014.pdf" rel="documentation"/>
+    <author>
+      <name>Matthew Upson</name>
+      <uri>http://www.mendeley.com/profiles/matthew-upson/</uri>
+      <email>ivyleavedtoadflax@gmail.com</email>
+    </author>
+    <contributor>
+      <name>Josh Cole</name>
+      <uri>http://www.mendeley.com/profiles/josh-cole7/</uri>
+    </contributor>
+    <contributor>
+      <name>Shannon Searle</name>
+      <email>s.searle@cranfield.ac.uk</email>
+    </contributor>
+    <category citation-format="author-date"/>
+    <category field="generic-base"/>
+    <summary>Cranfield Author-Date Style</summary>
+    <updated>2017-06-20T21:47:12+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <macro name="editor">
+    <names variable="editor" delimiter=", ">
+      <name and="text" delimiter-precedes-last="never" initialize-with=". " name-as-sort-order="all"/>
+      <label form="short" prefix=" (" text-case="lowercase" suffix=")"/>
+    </names>
+  </macro>
+  <macro name="cite-editor">
+    <names variable="editor" delimiter=",">
+      <name and="text" initialize-with=". " name-as-sort-order="first"/>
+    </names>
+  </macro>
+  <macro name="anon">
+    <choose>
+      <if type="book" match="any">
+        <text variable="title" font-style="normal"/>
+      </if>
+      <else>
+        <text term="anonymous" form="short" text-case="capitalize-first" strip-periods="true"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="author">
+    <names variable="author">
+      <name and="text" delimiter-precedes-last="never" initialize-with="." name-as-sort-order="all"/>
+      <label form="short" prefix=" " text-case="lowercase"/>
+      <substitute>
+        <choose>
+          <if type="chapter" match="any">
+            <text variable="publisher"/>
+          </if>
+        </choose>
+        <text macro="editor"/>
+        <text macro="anon"/>
+      </substitute>
+    </names>
+  </macro>
+  <macro name="author-short">
+    <names variable="author">
+      <name form="short" and="text" delimiter-precedes-last="never" et-al-min="4" et-al-use-first="1" initialize-with=". "/>
+      <substitute>
+        <names variable="translator"/>
+        <choose>
+          <if type="chapter" match="any">
+            <text variable="publisher"/>
+          </if>
+          <else>
+            <text macro="cite-editor"/>
+          </else>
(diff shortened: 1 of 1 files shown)


adam3smith:
could you clarify how this differs from the Cranfield University Harvard style?

searlesr:
Hi Sebastian,

The biblio section of the previous style is not displaying authors correctly (ie not showing all authors but truncating), the librarians here don’t want to use the word Harvard any longer in favour of simple author-date, there are more corrections/additions to be done to this style (citing blogs/wikis? Datasets? removing the template?) and so they want a clear delineation between the old style and the new one. I realise that normally we would just try and correct the changes in the current style, which have happened (I am assuming) due to changes in the referred template style. From what I can see something changed in September last year which ‘broke’ the style for us; and as we didn’t do any changes I can only assume it was the template. As this style is ‘broken’ the Library wants them to start using the newer style – I am assuming at some point we can request removal of the older one?

There doesn’t seem to be much in the way of tutorials/resources to learn csl  although I want/need to know more – the previous changes were made by Matt Upton who has since gone one to greater things…if I get way out of my depth - which is probable then is there any way to hire someone to do the changes for us or assist in reviewing them?

I am also a novice with Github as I’ve always used our svn repository internally with python/perl applications so hopefully I have done things correctly in the pull request. Thanks for your patience!

Shannon

rmzelle:
> From what I can see something changed in September last year which ‘broke’ the style for us; and as we didn’t do any changes I can only assume it was the template.

Could you elaborate a little bit on what "broken" means? Apart from a title change ("Harvard - Cranfield University" to "Cranfield University - Harvard" in March 2017, our repository style hasn't been changed in any way since March 2015 (see https://github.com/citation-style-language/styles/commits/master/harvard-cranfield-university.csl).

> As this style is ‘broken’ the Library wants them to start using the newer style – I am assuming at some point we can request removal of the older one?

We always try to avoid transition periods during which we have multiple versions of the same style. It sounds like you wouldn't mind if all students would switch right away to the new style, correct? In that case we can replace the old style by the new style in the same pull request (e.g. this one).

> There doesn’t seem to be much in the way of tutorials/resources to learn csl although I want/need to know more – the previous changes were made by Matt Upton who has since gone one to greater things…if I get way out of my depth - which is probable then is there any way to hire someone to do the changes for us or assist in reviewing them?

There is the introductory primer at http://docs.citationstyles.org/en/stable/primer.html. The CSL specification (http://docs.citationstyles.org/en/stable/specification.html) is a good reference during editing. See also http://citationstyles.org/styles/#Editing_Styles.

And yes, there are a few of us who are available for hire (you can reach out at http://citationstyles.org/contact/ if you want a quote, which depends on the amount of work). But if you just have a few quick questions you can just ask them here.

> I am also a novice with Github as I’ve always used our svn repository internally with python/perl applications so hopefully I have done things correctly in the pull request.

Everything looks good! And it's good to get some experience with GitHub, since it's so popular nowadays (and Git has largely replaced SVN, see e.g. https://trends.google.com/trends/explore?date=all&q=git,svn,github,sourceforge). The CSL styles repository itself moved from SVN to Git/GitHub in 2011, and it has made our work much easier.

searlesr:
Hi
I guess the word ‘broken’ isn’t right but I’m  not sure what else to call it. The bibliography should list all authors fully  in our style – from September 2016 I think, it started to truncate the author list to 4 instead of using the entire list. I know you didn’t change anything but I assumed that if there is a template file being used then that is a dependency; and when I followed changes to the template file back I noticed that there was a change about the same time our style ‘broke’. I tried to negate that change by altering ours to now show full authors list in the bibliography.

Hopefully I have sent you a pull request for the style that corrects that error. Editing these seems to spawn multiples of the file and references to id’s in Mendeley that I don’t understand – I am used to a proper versioning system that can be reverted/tracked but signing into Mendeley and editing csl files causes all kinds of files to be created without proper comments anywhere and a new id every time….most confusing!! I really wish the librarians would just use Zotero – so much easier and it’s open source.

As far as deleting the old style at the same time – so what happens when a student who has the old style installed and is not aware that there is a new one tries to apply the old one to a different computer? It won’t be there if they are on a different computer (as it stays installed in the profile) and it won’t be available to download? We have a lot of foreign students all over as I imagine most universities do, and our local environment doesn’t allow changes to the profile on the computer so they don’t see any styles except those on the repository or those installed in the Userprofile/Appdata/MendeleyDesktop folder by default…

I don’t like leaving the old one there either, it’s messy. Is there usually an issue with having multiple styles? As I said before ideally
I would like to remove all template dependencies on other styles, and there are at least 8 more style changes to be done to this one – here are some issues/changes needed…



1.       Journal citations should really have a resolvable doi path – currently just says 10.10222x/ instead of https://doi.org/10.10222x - the American Psych csl  does this properly so I will look there

2.       Inline citations should truncate to first author et al when four or more authors, list all if less (I can fix this as well)

3.       There are a few corrections in where things are placed or are missing (ie conference references aren’t putting locations/years in properly in proceedings

4.       No author citations don’t have a dot after Anon – I have fixed this on my testing csl file

5.       Remove any external style dependencies so it’s not inadvertently changed again (as stated)

6.       Better citations from VLE course lectures and online blogs – clear identification as a blog/vlog

7.       Does Mendeley even have a dataset citation format yet?

So this is a few of the changes I will want to send for pull over the next couple of months. I am unclear whether Mendeley uses your list for citation styles with Mendeley Desktop? When I edit with their editor I also get an id which points to their repository usually with my name so I realise I could send that link instead,  but I would rather keep it on this one since explaining how to install extra styles is difficult for some people to understand.
Sorry I meant to answer sooner but I had some questions to ask the IS staff about the doi format…should I fix/commit the ones from the test file that I can before you pull in? I am under great pressure to get the new style available even if all the bits aren’t immediately fixed…

Thanks so much for your help/suggestions!

Shannon

rmzelle:
> The bibliography should list all authors fully in our style

Our current style (at https://github.com/citation-style-language/styles/blob/master/harvard-cranfield-university.csl) does exactly that.

> Editing these seems to spawn multiples of the file and references to id’s in Mendeley that I don’t understand

You might find the general CSL editor at http://editor.citationstyles.org/about/ slightly easier to use in this respect. Note that the preloaded styles at http://editor.citationstyles.org/about/ and especially http://editor.mendeley.org/ can be quite outdated compared to the copies we maintain here in the CSL repository, so it's best to always start by copying and pasting the code style from here.

> As far as deleting the old style at the same time – so what happens when a student who has the old style installed and is not aware that there is a new one tries to apply the old one to a different computer?

It's my understanding that, if we update the existing style here, that Mendeley will automatically update the style for people who already have the style installed. (Zotero does that within 24 hours, not sure about Mendeley)

> Is there usually an issue with having multiple styles?

Well, no, because we usually don't allow it :).

> I am unclear whether Mendeley uses your list for citation styles with Mendeley Desktop? When I edit with their editor I also get an id which points to their repository usually with my name so I realise I could send that link instead

The repository here at https://github.com/citation-style-language/styles is the official repository for CSL styles, and both Mendeley and Zotero pull their styles from here. If you use Mendeley's CSL editor, it will generate a custom ID for the edited style and host it on the Mendeley servers, but those custom styles don't get automatically distributed to any users.

> I am under great pressure to get the new style available even if all the bits aren’t immediately fixed

It still sounds like we can just update the existing style at https://github.com/citation-style-language/styles/blob/master/harvard-cranfield-university.csl. It's not a problem for us if you want to make some changes now and some in the near future. It's probably easiest if you create a new pull request to update the existing style, though, while keeping the file name and style title and style ID the same. See https://github.com/citation-style-language/styles/blob/master/CONTRIBUTING.md#3b-submitting-changes-to-an-existing-style for details on how to edit existing styles.

rmzelle:
(let us know if you need help resubmitting this; you should be able to click the edit/pencil icon at https://github.com/citation-style-language/styles/blob/master/harvard-cranfield-university.csl to create a pull request to update the existing style)

PR #2754, opened 2017-06-21
Update invisu.csl

amendesd:
Original title, original publisher place, original publisher and original date added.


rmzelle:
Thanks!

PR #2756, opened 2017-06-21
Create university-of-zabol-farsi.csl

hamedheydari:


damnation333:
Hi @hamedheydari,

I see that there is plenty of translations in there.
You do know that Farsi can be automatically translated by setting the locale in the code? Or are all these terms different to what is in the translations?
See here: https://github.com/citation-style-language/locales/blob/master/locales-fa-IR.xml

hamedheydari:
@damnation333 Hi. Yes, I Know. and that locale file edited by me 6 months ago, but still not applied in Mendeley.
I can remove some terms with same translate. Or erase the whole terms after that Mendeley apply new locale file.

damnation333:
Ahhh. I see. Shame Mendeley is so slow on this.

rmzelle:
> And that locale file edited by me 5 months ago, but still not applied in Mendeley.

@cpina, are you back at Mendeley?

rmzelle:
@hamedheydari, a few questions/comments:

* I just asked another Mendeley employee whether they can update their CSL locales.
* Would it be acceptable to label this style as "Persian" instead of "Farsi"? I would like to use the same label for all current and future styles that rely on the "fa-ir" locale, and "Persian" seems like the more general term (per https://en.wikipedia.org/wiki/Persian_language).
* I noticed that we already have university-of-zabol.csl for English. It's possible to some extent to create CSL styles for more than one language (for these styles, we don't use a fixed style language and remove the style's `default-locale`), but since English and Persian/Farsi are so different that probably won't work in this case. If we do accept the style in this pull request as a second style for the University of Zabol, we should change the title of the English style from `<title>University of Zabol</title>` to `<title>University of Zabol (English)</title>`.
* I assume you based this style on the English variant? If so, we should point to that style via a "template" link (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link)

hamedheydari:
@rmzelle
1. Thank you. 
2. Yes. 
3. Ok. 
4. Yes. That's true. Please add that link.

rmzelle:
Okay, thanks!

We'll just take this as is, then. If you ever notice that Mendeley has updated their locales, please file another pull request to remove the locale section if it's no longer necessary.

hamedheydari:
@rmzelle Ok. Thanks. Also see here: https://github.com/citation-style-language/styles/pull/2755/commits/b91d4b0a74ec45fcee565b471965b418a9accf6a
(This is new)

rmzelle:
@hamedheydari, thanks. We've just been a little swamped here recently. The CSL project is run by volunteers, and we're all busy with other things. I still have some issues from March I need to start on, so you're getting better service than most :).

PR #2758, opened 2017-06-22
Create international-journal-of-circuit-theory-and-applications.csl

damnation333:
https://forums.zotero.org/discussion/51797/style-request-international-journal-of-circuit-theory-and-applications

Crappy guidelines. Basically adapted for article-journals.


rmzelle:
Thanks!

PR #2759, opened 2017-06-22
Create seed-science-research.csl

damnation333:
via https://forums.zotero.org/discussion/47555/style-request-seed-science-research
Adapted for journals. Chapters a bit weird in the guidelines. We'll see.


rmzelle:
Thanks!

PR #2760, opened 2017-06-22
Create cath-lab-digest.csl

damnation333:
via https://forums.zotero.org/discussion/62079/style-request-cath-lab-digest#latest
Adapted only for article-journal
I've emailed the style to the editors to be checked and included on their website. let's see if I get an answer.


rmzelle:
Thanks! (@adam3smith, intriguing that they use a modified version of the Campbell & Pedersen reference at http://www.cathlabdigest.com/author-guidelines!)

PR #2207, opened 2016-09-07
Et Al Tests

inukshuk:
The test does something similar to what citeproc-ruby does when rendering styles: looks at every `<names>` and fetches its `<name>` (or creates a `<name>` as a stand-in if there is none) then checks the attributes of that node inheriting from the closest `<bibliography>` or `<citation>` and `<style>`.

The tricky part is `<names>` inside a `<macro>` -- when rendering styles this is not a problem, because the processor will know in what context the macro is being rendered, but when checking the styles generically, I would argue it makes sense to run the checks twice, inheriting from both possible rendering nodes respectively.

The tests check that both `et-al-min` and `et-al-use-first` are defined (or that both are not defined) and that min is greater than or equal to use first. Both checks are repeated for the 'subsequent' variants.

I'm not 100% sure that this works. There are about 40 styles that fail the tests; I did a few spot checks and the failures seemed legit to me.

Let me know what you think!

(continuation of http://xbiblio-devel.2463403.n2.nabble.com/CSL-Specificaton-question-et-al-min-et-al-use-first-tp7579481p7579487.html)

PR #2767, opened 2017-06-25
Create journal-of-clinical-neurophysiology.csl

damnation333:
via https://forums.zotero.org/discussion/comment/229467#Comment_229467

(based on a style ready under review (https://github.com/citation-style-language/styles/pull/2766), hence it's not validating yet)

### DIFF
--- a/journal-of-clinical-neurophysiology.csl
+++ b/journal-of-clinical-neurophysiology.csl
@@ -6,21 +6,21 @@
     <id>http://www.zotero.org/styles/journal-of-clinical-neurophysiology</id>
     <link href="http://www.zotero.org/styles/journal-of-clinical-neurophysiology" rel="self"/>
     <link href="http://www.zotero.org/styles/journal-of-biological-regulators-and-homeostatic-agents" rel="template"/>
-    <link href="http://edmgr.ovid.com/jcnp/accounts/ifauth.htm" rel="documentation"/>
+    <link href="http://journals.lww.com/clinicalneurophys/_layouts/15/1033/oaks.journals/informationforauthors.aspx" rel="documentation"/>
     <author>
       <name>Patrick O'Brien</name>
       <email>obrienpat86@gmail.com</email>
     </author>
     <category citation-format="numeric"/>
     <category field="biology"/>
-    <eissn>1537-1603</eissn>
     <issn>0736-0258</issn>
+    <eissn>1537-1603</eissn>
     <updated>2017-06-25T12:04:18+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="author">
     <names variable="author" suffix=". ">
-      <name et-al-min="19" et-al-use-first="19" initialize-with="" name-as-sort-order="all" sort-separator=" "/>
+      <name et-al-min="7" et-al-use-first="3" initialize-with="" name-as-sort-order="all" sort-separator=" "/>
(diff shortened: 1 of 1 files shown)


journal-of-clinical-neurophysiology.csl
```diff
@@ -13,8 +13,8 @@
     </author>
     <category citation-format="numeric"/>
     <category field="biology"/>
-    <issn>0736-0258</issn>
     <eissn>1537-1603</eissn>
+    <issn>0736-0258</issn>
```
rmzelle: @damnation333, any reason you did this? I had already restarted the failing Travis build.

journal-of-clinical-neurophysiology.csl
```diff
@@ -6,21 +6,21 @@
...
-    <eissn>1537-1603</eissn>
     <issn>0736-0258</issn>
+    <eissn>1537-1603</eissn>
     <updated>2017-06-25T12:04:18+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="author">
     <names variable="author" suffix=". ">
-      <name et-al-min="19" et-al-use-first="19" initialize-with="" name-as-sort-order="all" sort-separator=" "/>
+      <name et-al-min="7" et-al-use-first="3" initialize-with="" name-as-sort-order="all" sort-separator=" "/>
```
rmzelle: @adam3smith, I thought we had automatic checks against invalid et-al settings, but this didn't trigger any Travis errors. I'm not misremembering, right? Looks like I should investigate.

adam3smith: I'm pretty sure you *are* misremembering this. We have talked about this, but I'm pretty sure we still have a pretty large number of styles with such settings.

rmzelle: > I'm pretty sure you are misremembering this. We have talked about this, but I'm pretty sure we still have a pretty large number of styles with such settings.

See https://github.com/citation-style-language/styles/pull/2207.

adam3smith: ah sorry,  you're right, I was wrong -- but the test checks for "greater or equal" when it should check for "greater", right?

rmzelle: Ah, yes, good catch! (we should fix this in a PR so we can fix any problematic styles before Travis starts to complain; I'll put it on my todo list)

rmzelle: (continued in https://github.com/citation-style-language/styles/pull/2777)

rmzelle:
@damnation333, thanks! (you were sloppy on the et-al settings, though 😬; the instructions say `Provide all authors' names when fewer than seven; when seven or more, list the first three and add "et al."`)

damnation333:
Yup. Sorry about that. I do indeed need to be careful. I've been pushing these out like a machine lately. Thanks for looking over them.
you also prefer eissn and then issn in that order, right?

rmzelle:
> I do indeed need to be careful. I've been pushing these out like a machine lately. Thanks for looking over them.

Just so you know, I usually only look carefully at the style metadata (title, documention link, ISSNs, etc.) and take a very brief look at the style code. I don't have the time to actually test and check the output of styles, so we rely mostly on you for that.

> you also prefer eissn and then issn in that order, right?

No, first `<issn/>` and then `<eissn/>`. When in doubt, just run your style through http://formatter.citationstyles.org/, or look at existing styles in the repository, which should both give you the desired order.

Issue #2222, opened 2016-09-14
Style for Journal of Evolution and Health


Noora99:
Journal of Evolution and Health uses style:
Freedman SB, Adler M, Seshadri R, Powell EC. Oral ondansetron for gastroenteritis in a pediatric emergency department. N Engl J Med. 2006 Apr 20;354(16):1698-705. PubMed PMID: 16625009.

It is similar to Physiology & Behavior citation style but instead of point after number in text uses brackets around number and instead of PMID uses doi.

rmzelle:
(instructions to authors at http://jevohealth.com/journal/policies.html#formatting)

rmzelle:
Style has been added to the repository (https://github.com/citation-style-language/styles/pull/2770).

PR #2770, opened 2017-06-26
Create journal-of-evolution-and-health.csl

damnation333:
via https://github.com/citation-style-language/styles/issues/2222

### DIFF
--- a/journal-of-evolution-and-health.csl
+++ b/journal-of-evolution-and-health.csl
@@ -124,7 +124,7 @@
             <text macro="year-date"/>
           </group>
         </else-if>
-        <else-if type="article-journal journal">
+        <else-if type="article-journal">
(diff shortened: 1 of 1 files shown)


rmzelle:
(the requestor wrote "It is similar to Physiology & Behavior citation style but instead of point after number in text uses brackets around number and instead of PMID uses doi.", but http://jevohealth.com/journal/policies.html#formatting shows an example with an PMID, so I'm a little confused)

damnation333:
I think she just made a mistake swapping the words DOI and PMID. The guidelines clearly show PMID. Consideringt he DOI is usually standard and also what the template style gives, it is highly likely she meant it the other way round.

rmzelle:
Thanks!

PR #2771, opened 2017-06-26
Create journal-of-the-history-of-collections.csl

damnation333:
via https://forums.zotero.org/discussion/55463/style-request-journal-of-the-history-of-collections
I only made the change to the issue number as pointed out by Adamsmith in the thread.

The template style links to a workaround to get the numbering as requested. See discussion here: https://forums.zotero.org/discussion/22752/#Comment_122034.
Can this be "fixed" now since CSL has been further developed?

(Rintze, the style as is gives me an error in the CSL editor. I've sent Adam a screenshot and he'll update Citeproc to see if it goes away.)

### DIFF
--- a/journal-of-the-history-of-collections.csl
+++ b/journal-of-the-history-of-collections.csl
@@ -1,0 +1,287 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" page-range-format="minimal" class="note" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+  <info>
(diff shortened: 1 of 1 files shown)


rmzelle:
> The template style links to a workaround to get the numbering as requested. See discussion here: https://forums.zotero.org/discussion/22752/#Comment_122034.
> Can this be "fixed" now since CSL has been further developed?

The CSL 1.0.1 spec is a little vague on the issue of whether page-range-format extends to the "locator" variable (see https://github.com/citation-style-language/documentation/issues/35), which we should eventually fix. I'm not sure what citeproc-js does nowadays in this respect.

rmzelle:
Thanks!

PR #2772, opened 2017-06-27
Update world-applied-sciences-journal.csl

damnation333:
FIx some issues with the delimiters/suffixes.
@adam3smith and @rmzelle 
Since a few weeks I finally get why the groups make it less possible for mistakes! Ha! Finally :) Hope that is reflected in my code.


rmzelle:
Thanks!

> Since a few weeks I finally get why the groups make it less possible for mistakes! Ha! Finally :)

:fireworks:

PR #2774, opened 2017-06-27
Create journal-of-political-philosophy.csl

damnation333:
via https://forums.zotero.org/discussion/58475/style-request-journal-of-political-philosophy

Quick'n'dirty one based on a very similar style, also in Wiley made by you guys.


rmzelle:
Thanks!

PR #2777, opened 2017-06-29
Require et-al-min to be greater than et-al-use-first (fix)

rmzelle:
Fix small bug in testing of et-al attribute values. (et-al(-subsequent)-min should have a higher value than et-al(-subsequent)-use-first)

(original code from https://github.com/citation-style-language/styles/pull/2207)

### DIFF
--- a/rtf-scan.csl
+++ b/rtf-scan.csl
@@ -44,7 +44,7 @@
       </else>
     </choose>
   </macro>
-  <citation et-al-min="1" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true">
+  <citation et-al-min="2" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true">
     <sort>
       <key macro="author-short"/>
       <key macro="title"/>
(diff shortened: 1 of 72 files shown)


rmzelle:
@adam3smith, just 70 styles! :)

rtf-scan.csl
```diff
@@ -44,7 +44,7 @@
       </else>
     </choose>
   </macro>
-  <citation et-al-min="1" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true">
+  <citation et-al-min="2" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true">
```
rmzelle: @adam3smith, I assume this doesn't break anything. Also, any idea whether people actually use this style?

rmzelle:
@adam3smith, I checked the instructions to authors as much as possible for these styles, although I had to guess some styles. Anyway, I think this is good to go. Please squash and merge.

rtf-scan.csl
adam3smith: yes, some people do use this and yes, the change is fine.

adam3smith:
Cool, thanks. No concerns from my end. Obviously I didn't double-check the actual et-al values, but I have all confidence that you did this as well or better than I'd be able to.

rmzelle:
Thanks!

ukyoi:
I'm sorry for my rudeness since I'm not a native English speaker, but you really shouldn't just change the file without knowing the details of the corresponding styles.

In Chinese GB/T 7714-1987 and GB/T 7714-2005, the standard specified that the value "et-al-use-first" is ought to be "3". I was happy to find that CSL and used the newest version on my thesis, then got a lot criticism. I know not checking the standard is my fault, but I remember it works fine around one year ago.

PR #2778, opened 2017-06-29
Update leiden-journal-of-international-law.csl

Danerad:
Assorted minor changes to fix small bugs and to increase flexibility.


rmzelle:
Thanks!

PR #2780, opened 2017-06-30

damnation333:
via https://forums.zotero.org/discussion/comment/278951#Comment_278951

Now, the chapters should give the DOI as should a speech, but what if a speech doesn't have a DOI and I want to give the URL. How would that work?


rmzelle:
> but what if a speech doesn't have a DOI and I want to give the URL. How would that work?

Well, you can just test for the presence of a DOI, right?

damnation333:
closed as already being dealth with here: https://github.com/citation-style-language/styles/pull/2633

PR #2781, opened 2017-07-01
Create journal-of-software.csl

carlsplace:
Journal of Software is a famous journal in China, I think it's helpful to share the csl file.

### DIFF
--- a/journal-of-software.csl
+++ b/journal-of-software.csl
@@ -1,0 +1,143 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>Journal of Software(Chinese)</title>
+    <title-short>JOS</title-short>
+    <id>http://www.zotero.org/styles/journal-of-software</id>
+    <link href="http://www.zotero.org/styles/journal-of-software" rel="self"/>
+    <link href="http://www.zotero.org/styles/elsevier-with-titles" rel="template"/>
+    <author>
+      <name>Chang YC</name>
+      <email>chang.1123@outlook.com</email>
+    </author>
+    <category citation-format="numeric"/>
+    <issn>1000-9825</issn>
+    <summary>A style for Journal of Software</summary>
+    <updated>2017-07-01T03:38:51+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <macro name="author">
+    <names variable="author">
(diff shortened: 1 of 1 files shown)


rmzelle:
@carlsplace, thanks!

Two questions:

* http://www.jos.org.cn/ch/index.aspx is the journal's website, right? Is there a page on that website that describes the journal's citation format? (usually found in the "Instructions to authors" section). We also try to add a "documentation" link to that page (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#10---documentation-link).
* The style you're submitting uses `default-locale="en-US"`, which means that English translations will be used for CSL terms like "in" and "accessed". Is that correct? Or should the style output be in Chinese?

journal-of-software.csl
```diff
@@ -0,0 +1,143 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>Journal of Software(Chinese)</title>
+    <title-short>JOS</title-short>
+    <id>http://www.zotero.org/styles/journal-of-software</id>
+    <link href="http://www.zotero.org/styles/journal-of-software" rel="self"/>
+    <link href="http://www.zotero.org/styles/elsevier-with-titles" rel="template"/>
+    <author>
```
damnation333: Can you add a link to the documentation of the style?

journal-of-software.csl
```diff
@@ -0,0 +1,143 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>Journal of Software(Chinese)</title>
```
damnation333: why does this title have (chinese)? From what I see they publish in English. Also, the default-locale is set to en-US, so this wouldn't appear in chinese in the first place.

journal-of-software.csl
rmzelle: (@damnation333, I appreciate the help, but I asked these exact same two questions directly above :P)

damnation333: ha. sorry. didn't see it in here as I was looking in the code view. Won't happen again. :)

carlsplace:
1.  I didn't find those information, but I  will try to fix that.
2.  This journal's citations are all in English even for Chinese papers, so it's fine to use `default-locale="en-US"`

rmzelle:
> I didn't find those information, but I will try to fix that.

Let us know if you were successful or not in finding something. How do authors know what citation format to use in their manuscripts? Almost all journals provide some guidance somewhere on their website.

rmzelle:
@carlsplace, closing because of inactivity. We can always reopen this if you answer my question above.

PR #2783, opened 2017-07-02
Thieme dependents

rmzelle:
Via https://github.com/citation-style-language/journals/pull/39


rmzelle:
(I just realized I didn't set up redirects in renamed-styles.json for the independent styles we're deleting that have different names in their dependent form; we could also change the metadata to keep the journal titles and style IDs the same)

adam3smith:
I did a quick check and the International Journal of Sports Medicine publishes in English with English abbreviations (i.e. ed instead of Hrsg.) and Röfo uses et al for >  authors, as does DMW. If we're in a rush here, I'd leave these independent and check back with @juliarojahn

rmzelle:
Sure. I can put them on the skip list for now and update the PR in the styles repo.

rmzelle:
> I'd leave these independent

@adam3smith, I updated both the "journals" and "styles" PRs to do just that. Should be good to merge.

juliarojahn:
Please include the Journal DMW in the style - it should use the same style, and the colleagues would be very happy to be included!

The journals with english citations (International Journal of Sports Medicine, RöFo etc.) I can tackle afterwards.
Thanks,
Julia

adam3smith:
OK, we'll fix this so it includes DMW and then merge it over the weekend. Sorry for the delay

rmzelle:
Hmz, looks like I should have done DMW instead of the European Journal of Ultrasound.

rmzelle:
> OK, we'll fix this so it includes DMW and then merge it over the weekend.

@adam3smith, just made this change. Feel free to merge (and the journals PR mentioned at the top of this PR as well).

adam3smith:
great, thanks!

PR #2785, opened 2017-07-03
Update transportation-research-record.csl

damnation333:
via https://forums.zotero.org/discussion/66324/formatting-issues-transportation-research-record-journal-of-the-transportation-research-board


adam3smith:
thanks!

PR #2786, opened 2017-07-03
Update collection-du-centre-jean-berard.csl

NMonteix:
Change in URL form (no more access mention) + minor corrections


rmzelle:
Thanks!

PR #2791, opened 2017-07-05
Create journal-of-periodontal-research.csl

damnation333:
via https://forums.zotero.org/discussion/61259/style-request-journal-of-periodontal-research


rmzelle:
Great, thanks!

PR #2792, opened 2017-07-06
Update and rename dependent/tetrahedron-letters.csl to tetrahedron-le…

damnation333:
…tters.csl

via https://forums.zotero.org/discussion/66363/style-error-tetrahedron-letters

Adapted for journal articles, chapters.st
Style has changed from Wiley to Elsevier, but none of the standard Elsevier styles matches or the ACS style.


rmzelle:
Thanks!

PR #2794, opened 2017-07-06
Update journal-of-korean-neurosurgical-society.csl

damnation333:
via https://forums.zotero.org/discussion/66370/style-error-journal-of-korean-neurosurgical-society-search


rmzelle:
Thanks!

PR #2798, opened 2017-07-07

cookie365:
Added new mediatypes: interview, motion-picture
Made changes paper-conference and speech
Added new contributor
Sorted macros and terms by their name Atttribut
Changed citation layout prefix and sufix should be a blank character instead of brackets
Changed the bibliography


rmzelle:
Thanks!

PR #2799, opened 2017-07-07
Create molecular-plant.csl

damnation333:
via https://github.com/citation-style-language/journals/pull/13#discussion_r126067609
adapted for article-journal and et al. settings.


adam3smith:
Before you fix this (need to remove the dependent style), could you check whether I was right with this?

> The style is actually identical to https://www.zotero.org/styles?q=The%20Plant%20Cell

in which case we can just switch the dependent style over the the plant cell instead of creating a new one?

damnation333:
No, this style is different. for journals, container-title is short, but with dots. there is also a space after the volume. I could've used that style as a base though.

rmzelle:
Thanks!

PR #2801, opened 2017-07-10
Update Patents in APA and align related versions

adam3smith:
Patents in APA are really weird:

Reference list entry:
Smith, I. M. (1988).*U.S. Patent No. 123,445*. Washington, DC: U.S. Patent and
Trademark Office.

Text citation:
U.S. Patent No. 123,445 (1988)

@bwiernik if you wouldn't mind taking a look at the changes to apa.csl -- I think you can ignore the rest, those just align similar styles to the "master" copy.
(U.S. Patent No. 123,445, 1988)

edit: a lot of the changes are indenting. Much easier to see what's actually new using `?w=1` at the end of the diff URL

### DIFF
--- a/apa-annotated-bibliography.csl
+++ b/apa-annotated-bibliography.csl
@@ -101,7 +101,7 @@
   <macro name="author-short">
     <choose>
       <if type="patent" variable="number" match="all">
-        <text variable="number"/>
+        <text macro="patent-number"/>
       </if>
       <else>
         <names variable="author">
@@ -141,6 +141,16 @@
         </names>
       </else>
     </choose>
+  </macro>
+  <macro name="patent-number">
+    <!-- genre: U.S. Patent number: 123,445-->
+    <group delimiter=" ">
+      <group delimiter=" ">
+	<text variable="genre"/>
+	<text term="issue" form="short" text-case="capitalize-first"/>
+      </group>
+      <text variable="number"/>
+    </group>
   </macro>
   <macro name="access">
     <choose>
@@ -236,7 +246,7 @@
         </choose>
       </else-if>
       <else-if type="patent" variable="number" match="all">
-        <text variable="number" font-style="italic"/>
+        <text macro="patent-number" font-style="italic"/>
       </else-if>
       <else>
         <text variable="title"/>
--- a/apa-annotated-bibliography.csl
+++ b/apa-annotated-bibliography.csl
@@ -99,41 +99,48 @@
     </names>
   </macro>
   <macro name="author-short">
-    <names variable="author">
-      <name form="short" and="symbol" delimiter=", " initialize-with=". "/>
-      <substitute>
-        <names variable="editor"/>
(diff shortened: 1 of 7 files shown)


rmzelle:
> Patents in APA are really weird:

What's so weird about it? The fact that they include the USPTO?

Also, do they provide any guidance on US patent applications and non-US patents? In my day job I e.g. deal a lot with WO (international) and EP (Europe) jurisdiction patents and patent applications.

(if only the patent community would adopt DOIs!)

adam3smith:
no, including the patent numbers makes perfect sense. The two things that are weird are that 1) you use the patent number in text but the inventor in the bibliography (i.e. the patents are a royal pain to find in the bib!) and 2) that you don't use the title of the patent in the bibliography -- that would seem rather important info... While they don't specify what to to with WO and EP explicitly, my reading would be that you cite them exactly the same way, ie. 
in-text:
(World patent application WO 2016/205406 A16, 201)

bib:
Beckett, K. D. R., & Tyc, G. (2016). *World patent application WO 2016/205406 A1*. Geneva: World Intellectual Property Organization.

bwiernik:
Looks right to me, assuming that the entire phrase "U.S. Patent No. 123,445" is entered as the patent number. This appears to be the expectation for Zotero. However, Mendeley and Juris-M both include genre/type fields mapped to `genre`. So it would probably be best to replace the flat `number` text with a conditional to print `genre` + `term:number` + `number` if genre is present or only `number` if not (ala: https://github.com/Juris-M/jm-styles/blob/master/jm-indigobook.csl#L154).

@adam3smith is correct. International patents are cited in the same way as US ones.

adam3smith:
Thanks @bwiernik -- did what you suggested. Let's see how this goes.

rmzelle:
(slightly related, I think we have been trying to phase it out, but there are still two styles with hard-coded "U.S. patent", which obviously only works for US patents:

the-optical-society.csl
106: `<text variable="number" prefix="U.S. patent "/>`

universidade-de-sao-paulo-instituto-de-matematica-e-estatistica.csl
169: `<text variable="number" prefix="U.S. Patent #"/>`
)

bwiernik:
Those should probably have the same format as just added to APA?

rmzelle:
> Those should probably have the same format as just added to APA?

Well, the current hard-coded "U.S. patent" is very limiting, so that would probably be an improvement, yes.

Also, do the Zotero translators (Espacenet, USPTO, Google Patents, FreePatentsOnline) already put the appropriate data in "genre" and "number"?

(does the CSL patent item type even have "genre" in Zotero? https://aurimasv.github.io/z2csl/typeMap.xml#map-patent suggests it does not)

bwiernik:
No, CSL patent doesn't have "genre" in Zotero, but it does in Mendeley and Juris-M. Zotero probably should get it at some point, rather than putting type and number into the same field.

PR #2803, opened 2017-07-11
Create applied-in-vitro-toxicology.csl

POBrien333:
via https://forums.zotero.org/discussion/66405/style-request-applied-in-vitro-toxicology
Adapted for article-journal, book, chapter.
Guidelines do not match actual publications regarding some smaller details. Open access publications here: http://online.liebertpub.com/toc/aivt/3/2


rmzelle:
Thanks!

PR #2804, opened 2017-07-11
Create journal-of-plant-ecology.csl

POBrien333:
via https://forums.zotero.org/discussion/47045/style-request-journal-of-plant-ecology
adapted for article-journal, chapter, webpage.
Mainly based off actual publications....
I'll email them this citation style and that their guidelines are not correct.


rmzelle:
Thanks!

PR #2805, opened 2017-07-11
Update european-journal-of-information-systems.csl

POBrien333:
via https://forums.zotero.org/discussion/66423/european-journal-of-information-systems
updated for books and chapters and a bit for article-journal (bold volume and issue).


rmzelle:
Thanks!

PR #2808, opened 2017-07-12
Test new Ubuntu image for Travis-CI

rmzelle:
Select Ubuntu Trusty (14.04) to upgrade from Precise (12.04), which should give us pre-installed ruby 2.4.1.

See https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming and https://docs.travis-ci.com/user/build-environment-updates/2017-07-12/.


rmzelle:
Hmmm, looks like Trusty has 2.4.0, but not 2.4.1 (which their blog post said it should have). Asked Travis-CI about it at https://twitter.com/rintzezelle/status/885206097396727808

rmzelle:
@adam3smith, this PR selects the new Ubuntu Trusty image offered by Travis. Don't really see a drawback, as Travis likely will stick with Ubuntu Trusty for several years, and it puts us a little ahead the curve (their roll-out of Trusty as default won't be complete until September).

We can now use the latest version of Ruby without a speed penalty (the old Ubuntu image had 2.2 as most recent Ruby, and it took ca. 30 s to install 2.4). In the past we've also seen some speedups with upgrading Ruby, so I rather run the latest version.

PR #2810, opened 2017-07-13
Create journal-of-value-inquiry.csl

damnation333:
https://forums.zotero.org/discussion/comment/266340#Comment_266340
Not sure if I did the Ibid. thing right.


journal-of-value-inquiry.csl
```diff
@@ -0,0 +1,254 @@
...
+                <text macro="pageref"/>
+              </group>
+            </if>
+            <else-if type="bill book graphic legal_case motion_picture song" match="any">
+              <group delimiter=", ">
+                <text macro="title"/>
+                <text macro="editor-translator"/>
+                <text macro="container"/>
+                <text macro="volumes"/>
+                <text macro="volref"/>
```
rmzelle: @POBrien333, there is no "volref" macro defined in this style (see the Travis report as well).

POBrien333: thank you :)

rmzelle:
Thanks!

PR #2811, opened 2017-07-14
Delete european-journal-of-ultrasound.csl

POBrien333:
as discussed here: https://github.com/citation-style-language/journals/pull/39 and https://github.com/citation-style-language/styles/pull/2783


rmzelle:
I'll update the other pull request, so everything is in one place.

PR #2813, opened 2017-07-14
Create apa-annotated-bibliography-abstract-annotation.csl

denlinkd:
Annotated bib with both abstract and extra field as requested here https://forums.zotero.org/discussion/64272/annotated-apa-bibliography-with-abstract-and-notes


adam3smith:
I appreciate this and you should feel free to use the link (https://github.com/denlinkd/styles/raw/aba008aa268d430778034c6fd6b0a5124b581d21/dependent/apa-annotated-bibliography-abstract-annotation.csl ) to point folks in the forum to, but I don't think we'll take this. These additional styles take time to maintain and there's no limit on the variations we could get..

PR #2814, opened 2017-07-15
Create social-cognitive-and-affective-neuroscience.csl

POBrien333:
to supersede https://github.com/citation-style-language/styles/pull/2595

@iobergfeld
Would be great if you could use this style and let us know if there is mistakes and if the editors say anything. Their guidelines are not consistent with their actual publications.


rmzelle:
Thanks!

PR #2815, opened 2017-07-15
Update and rename dependent/the-journal-of-nuclear-medicine.csl to th…

### LINKED ISSUE #1982: Jorunal of Nuclear Medicine
I'm using Journal of Nuclear Medicine csl. However, the format in the references is not correct, there should be no DOI information at the end. And the inline citations are incorrect as well. Instead of "1" in superscript format, it should be "(1)".

I tried to change the style by clicking on "Edit the style" on this site: **http://editor.citationstyles.org/about/**
However, nothing happens, the link doesn´t work. Can anyone please help to correct the style of this journal or give some instructions how I can change it?

Thank you

POBrien333:
…e-journal-of-nuclear-medicine.csl

Closes #1982.
adapted for article-journal.


rmzelle:
Thanks!

PR #2817, opened 2017-07-15
Create econometrica.csl

POBrien333:
via https://forums.zotero.org/discussion/66519/style-request-econometrica


rmzelle:
Thanks!

PR #2818, opened 2017-07-16
Add footpage

guschartier:
Ajout d'éléments dans les notes de bas de pages. Rectification suite à une précédente erreur.

### DIFF
--- a/seminaire-saint-sulpice-ecole-theologie.csl
+++ b/seminaire-saint-sulpice-ecole-theologie.csl
@@ -36,7 +36,7 @@
     <choose>
       <if variable="author">
         <names variable="author" font-variant="normal" suffix=", ">
-          <name font-style="normal" delimiter-precedes-et-al="always" delimiter-precedes-last="always" et-al-min="3" et-al-use-first="3" et-al-subsequent-min="3" et-al-subsequent-use-first="1" initialize="false" sort-separator=" ">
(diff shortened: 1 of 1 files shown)


seminaire-saint-sulpice-ecole-theologie.csl
```diff
@@ -498,7 +505,7 @@
     <label plural="never" suffix=" " variable="page" form="short"/>
     <text variable="locator" form="short"/>
   </macro>
-  <citation et-al-min="4" et-al-use-first="3" et-al-subsequent-min="3" et-al-subsequent-use-first="1">
+  <citation et-al-min="3" et-al-use-first="3" et-al-subsequent-min="3" et-al-subsequent-use-first="1">
```
rmzelle: @guschartier, "et-al-min" must always have a higher value than "et-al-use-first", since it doesn't make sense to print "et al." for unabbreviated name lists. We recently fixed this in this style (https://github.com/citation-style-language/styles/commit/e1e992968d45be5ccc92df5264beb41ce362c261#diff-2116f13782f3f36e81bd8b2277cae63b), but this pull request rolls those changes back. Can you confirm the right "et-al" settings?

rmzelle:
Thanks!

guschartier:
Thanks for your help !

rmzelle:
> Thanks for your help !

I just changed the et-al settings myself, so that they're at least logically consistent, but I don't know the actual guidelines, so let us know if the settings are correct.

Issue #1722, opened 2015-10-11
The Journal of Modern History


adam3smith:
https://forums.zotero.org/discussion/38095/style-error-the-journal-of-modern-history/#Item_3

Journal says it follows Chicago Manual and does for the most part, but no DOIs and no Publishers.

zuphilip:
This journal is currently a [dependent style](http://editor.citationstyles.org/styleInfo/?styleId=http%3A%2F%2Fwww.zotero.org%2Fstyles%2Fthe-journal-of-modern-history) of [Chicago Manual (note)](http://www.zotero.org/styles/chicago-note-bibliography) but I think it should first be corrected to Chicago Manual **full-note**. Or do I miss something here?

Howerver, the other differences, as you already pointed out, are:
 * no DOIs are used
 * publishers are not present (the style says "Note: publishers' names are included only in the initial source information for books under review.")

==> Therefore, we will need a new independent style started from CMoS full note style with these two changes. I will wait for @adam3smith's feedback and can then do the work.

PR #2821, opened 2017-07-17
Update and rename dependent/the-journal-of-modern-history.csl to the-…

POBrien333:
…journal-of-modern-history.csl

Closes https://github.com/citation-style-language/styles/issues/1722


rmzelle:
Thanks!

Issue #1887, opened 2016-02-14
Journal of Cachexia, Sarcopenia and Muscle


zuphilip:
The journal moved from Springer to Wiley:

http://onlinelibrary.wiley.com/journal/10.1007/13539.2190-6009/homepage/ForAuthors.html

 Is the reference style still `springer-vancouver-brackets`?

matejovo:
There are some changes...
**`springer-vancouver-brackets:`**
Slifka MK, Whitton JL (2000) Clinical implications of dysregulated cytokine production. J Mol Med 78:74–80. doi: 10.1007/s001090000086
**requested:** (no pp. year at the end)
Slifka MK, Whitton JL. Clinical implications of dysregulated cytokine production. J Mol Med. 2000; doi:10.1007/s001090000086

adam3smith:
I'm not sure what "article by DOI" means, but I'd not cite without the page number. But yeah, the citation style has changed entirely. We're basically looking at Vancouver (brackets) now, maybe a slight variant for online sources, but print sources look identical.

POBrien333:
solved in https://github.com/citation-style-language/styles/pull/2822

rmzelle:
Thanks, @POBrien333!

PR #2822, opened 2017-07-17
Create journal-of-cachexia-sarcopenia-and-muscle.csl

POBrien333:
via https://github.com/citation-style-language/styles/issues/1887
adapted for article-journal, book, chapter, web.
Superscript numeric style.


rmzelle:
Thanks!

PR #2827, opened 2017-07-18
Create journal-of-limnology.csl

POBrien333:
via https://forums.zotero.org/discussion/62824/style-request-journal-of-limnology

Adapted for book, chapter, article journal.


rmzelle:
Thanks!

PR #2828, opened 2017-07-18
Create the-journal-of-molecular-diagnostics.csl

POBrien333:
via https://forums.zotero.org/discussion/66551/style-csl-request-journal-of-molecular-diagnostics
Style posted in forum is a stub, so I made my own.
Can you tell me why it doesn't print the URL for webpages via the access macro?


the-journal-of-molecular-diagnostics.csl
```diff
@@ -0,0 +1,168 @@
...
+        </group>
+      </if>
+      <else>
+        <text variable="edition"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="access">
+    <choose>
+      <if variable="page volume" match="none">
```
adam3smith: since webpages and blogposts don't have page and volume fields, you will actually never get to the `else-if`. Easiest solution would be to flip this around: start with `<if type="webpage post-weblog" match="any">`

adam3smith:
Great, thanks!

Issue #2749, opened 2017-06-19
Style error: council-of-science-editors-alphabetical.csl


JeremyHorst:
The journal's issue number should appear if it exists, and there should be a period after the journal name abbreviation, for CSE 8.
This is either a problem with this format file, or Papers3... help!

damnation333:
Hi @JeremyHorst
Can you post an example article here or a copy of the guidelines or something? Just something we can base this on. Thanks.

JeremyHorst:
It currently leads to the output:
Horst JA, Ellenikiotis H, Milgrom PL. 2016. UCSF Protocol for Caries Arrest
Using Silver Diamine Fluoride: Rationale, Indications and Consent. J Calif
Dent Assoc 44:16–28.

Which should be:
Horst JA, Ellenikiotis H, Milgrom PL. 2016. UCSF Protocol for Caries Arrest
Using Silver Diamine Fluoride: Rationale, Indications and Consent. J Calif
Dent Assoc. 44(1):16–28.

Subtle but important difference... would lead to extensive unnecessary
manual curation.




Peace and Thanks,
Jeremy

---
*Jeremy Horst DDS, PhD*
Postdoctoral fellow, UCSF DeRisi Lab
<https://scholar.google.com/citations?user=CetD6iHud9IC&hl=en>
<https://www.linkedin.com/in/jeremyhorst>
<https://www.youtube.com/playlist?list=PLwbU_gwsxa0FmnREk9AYHlE7_kDL-I1dg>

<http://profiles.ucsf.edu/jeremy.horst>

adam3smith:
Thanks! The question was for the relevant passage from the CSE manual, though, so we can double-check. We're always making sure to go back to style guides before making changes to styles just to be safe.

rmzelle:
> It currently leads to the output: Horst JA, Ellenikiotis H, Milgrom PL. 2016. UCSF Protocol for Caries Arrest Using Silver Diamine Fluoride: Rationale, Indications and Consent. J Calif Dent Assoc 44:16–28.

I'm a bit confused here, because https://github.com/citation-style-language/styles/blob/master/council-of-science-editors-alphabetical.csl currently doesn't seem to print the journal title for articles (see @damnation333's example above).

adam3smith:
(note this is with papers 3 -- might just be an old version of the style?)

rmzelle:
[...]

PR #2829, opened 2017-07-19
Update council-of-science-editors.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/267127/#Comment_267127
and https://github.com/citation-style-language/styles/issues/2749
Added the doi bit.
It already had the URL and [accessed] bit.
I'll do the same for the other numeric style once you give me green lights and add it into this PR.


rmzelle:
@POBrien333, there is no real need for a DOI macro here, since it's just a one-liner, and part of the prefix (". " in `prefix=". doi:"`) belongs outside the macro anyways.

rmzelle:
(and can you take a look at https://github.com/citation-style-language/styles/blob/master/council-of-science-editors-alphabetical.csl as well? Per https://github.com/citation-style-language/styles/issues/2749#issuecomment-309550126, it is currently definitely wrong, since a journal article should always print the journal ("container-title"))

rmzelle:
Thanks!

PR #2830, opened 2017-07-20
Update council-of-science-editors-alphabetical.csl to 8th edition

POBrien333:
via https://forums.zotero.org/discussion/comment/267127/#Comment_267127
and https://github.com/citation-style-language/styles/issues/2749
Added the doi bit.
It already had the URL and [accessed] bit.
"Normal" numeric style dealt with here: https://github.com/citation-style-language/styles/pull/2829


rmzelle:
Thanks!

PR #2836, opened 2017-07-24
Create environmental-values.csl

POBrien333:
via https://forums.zotero.org/discussion/47623/style-request-environmental-values

Started with the wrong style, so there is some remnants in there that might not be ideal.
Adapted for article-journal, book, chapter, webpage.
Instructions: http://www.whpress.co.uk/EV/EVinst.html
Free publication examples: http://www.whpress.co.uk/EV/EVfree.html


rmzelle:
Thanks!

PR #2838, opened 2017-07-24
Create chemistry-education-research-and-practice.csl

POBrien333:
via https://forums.zotero.org/discussion/66697/style-change-request-chemistry-education-research-and-practice
goes hand in hand with this PR, removing the style from the automatic generating and skipping it in the list: https://github.com/citation-style-language/journals/pull/43


rmzelle:
@POBrien333, did you check this is an individual journal changing its citation format, or whether it extends to multiple RSC journals? (we currently have two variants: https://www.zotero.org/styles?q=royal%20society%20of%20chemistry)

POBrien333:
In their generic guidelines to all their journals they state this:
Bibliographic references & notes

We will format your content according to our house style before publication; however, it’s important you use Vancouver style (not Harvard style) for all journals except Chemistry Education Research and Practice, which requires the use of Harvard referencing.
Source: http://www.rsc.org/journals-books-databases/about-journals/chemistry-education-research-practice/#journal-specific-guidelines

Their generic style is NUMERIC, but for CREP they use an exceptional Harvard style.

rmzelle:
Thanks!

PR #2839, opened 2017-07-25
Create zeitschrift-fuer-allgemeinmedizin.csl

POBrien333:
via https://forums.zotero.org/discussion/66704/style-request-zfa-modified-vancouver-style
adapted for journals, books, chapters, web
Note, style is in German. although in the guidelines they mix Hrsg. and eds.


rmzelle:
Thanks!

PR #2841, opened 2017-07-26
Create journal-of-molecular-signaling.csl

POBrien333:
via https://github.com/citation-style-language/journals/issues/6
@zuphilip One to tick off your list ;)


rmzelle:
Thanks!

POBrien333:
How does one even notice these extra spaces?

rmzelle:
> How does one even notice these extra spaces?

I run most style submissions through http://formatter.citationstyles.org/ by hand to make sure the indentation is correct.

PR #2844, opened 2017-07-26
Update zeitschrift-fur-allgemeinmedizin.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/281386/#Comment_281386

### DIFF
--- a/zeitschrift-fur-allgemeinmedizin.csl
+++ b/zeitschrift-fur-allgemeinmedizin.csl
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
+  <!-- Polyglot; journal publishes in German and English -->
   <info>
-    <title>Zeitschrift für Allgemeinmedizin (German)</title>
+    <title>Zeitschrift für Allgemeinmedizin</title>
     <title-short>ZFA</title-short>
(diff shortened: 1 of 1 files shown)


zeitschrift-fur-allgemeinmedizin.csl
```diff
@@ -14,12 +14,13 @@
...
     <issn>1433-6251</issn>
     <eissn>1439-9229</eissn>
-    <updated>2017-07-25T14:09:51+00:00</updated>
+    <updated>2017-07-26T21:25:26+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="de">
     <terms>
       <term name="accessed">letzter Zugriff am</term>
+      <term name="et-al">et al</term>
```
rmzelle: @POBrien333, I haven't checked the guidelines, but should this be "et al."? It's an abbreviation ("et alia"), so a period is almost always used.

POBrien333: It actually has the suffix on the author macro, so that should've sufficed, no?

rmzelle: Even then the term should have a period. (term definitions should be correct independent of the context in which they're used, and CSL processors generally suppress duplicated punctuation to deal with e.g. extra periods)

rmzelle:
Thanks!

POBrien333:
Ah, dang. Thanks for fixing that! :)

rmzelle:
Nitpicking to the rescue!

image

PR #2845, opened 2017-07-28
Update el-profesional-de-la-informacion.csl

rzgairin:
Moved year behind authors.
Add line break before doi/url.


rmzelle:
Thanks!

I just made a few corrections and simplifications. The standard way to place text on a new line is to use `<group display="block"/>` elements within `<layout/>`, and your original code placed the DOI prefix ("https://doi.org") on the previous line, which can't have been what you were aiming for.

PR #2847, opened 2017-07-29
Update acta-orthopaedica.csl

gforge:
The Acta Orthopaedica is actually not just Vancouver-author-date but has its own quirks.


POBrien333:
You'll need to put this file in the normal foulder and not in the dependent folder. You can do this by editing the filename and just hitting backspace and it'll start deleting "dependent" (took me a while to realise that.
Then it'll validate.

acta-orthopaedica.csl
```diff
@@ -0,0 +1,267 @@
...
+      <date-part name="day"/>
+    </date>
+    <terms>
+      <term name="retrieved">available</term>
+      <term name="section" form="short">sect.</term>
+    </terms>
+  </locale>
+  <macro name="author">
+    <names variable="author">
+      <name delimiter-precedes-last="always" et-al-min="19" initialize-with="" name-as-sort-order="all" sort-separator=" "/>
```
adam3smith: delete the et-al-min here. If the setting in the bibliography (70/60) isn't right, move it to the bibliography.

acta-orthopaedica.csl
```diff
@@ -0,0 +1,267 @@
...
+          <text variable="publisher-place" prefix="[" suffix="]"/>
+        </if>
+        <else>
+          <text variable="publisher-place"/>
+        </else>
+      </choose>
+      <text variable="publisher"/>
+    </group>
+  </macro>
+  <macro name="access">
```
adam3smith: you're not using this macro -- delete it.

rmzelle:
Thanks!

gforge:
Thanks for the help fixing the remaining issues! You should by the way highlight more clearly the CSL language primer http://docs.citationstyles.org/en/stable/primer.html in the editing styles section. It is currently not directly linked and the CSL docs are hidden at the end of the paragraph. I'm a sloppy reader and despite reading the section several times I've failed to navigate to this section.

rmzelle:
> in the editing styles section.

Where is this section? http://citationstyles.org/ or here on the GitHub wiki?

PR #2848, opened 2017-07-29
Create planning-practice-and-research.csl

POBrien333:
via https://forums.zotero.org/discussion/62632/style-request-planning-practice-research-journal


rmzelle:
Thanks!

PR #2849, opened 2017-07-29
Create business-ethics-a-european-review.csl

POBrien333:
https://forums.zotero.org/discussion/60143/style-request-business-ethics-a-european-review
Wiley, aka guidelines don't match papers.
APA with slight modifications only.


rmzelle:
Thanks!

PR #2850, opened 2017-07-29
Create management-et-avenir

POBrien333:
via https://forums.zotero.org/discussion/47790/style-request-management-avenir/
APA adapted based on guidelines and the few comments in the thread. no access to journal.


rmzelle:
Thanks!

PR #2852, opened 2017-07-31
Create journal-of-social-philosophy.csl

POBrien333:
https://forums.zotero.org/discussion/50711/style-request-journal-of-social-philosophy
This Wiley journal doesn't even bother to put up any form of guidelines... Just an endnote and refman style...


rmzelle:
Thanks!

PR #2853, opened 2017-07-31
Create international-journal-of-management-reviews.csl

POBrien333:
via https://forums.zotero.org/discussion/51322/style-request-international-journal-of-management-reviews


rmzelle:
Thanks!

PR #2857, opened 2017-08-01
Create climate-change-economics.csl

POBrien333:
https://forums.zotero.org/discussion/66828/style-request-climate-change-economics

### DIFF
--- a/climate-change-economics.csl
+++ b/climate-change-economics.csl
@@ -105,7 +105,7 @@
       </group>
     </layout>
   </citation>
-  <bibliography subsequent-author-substitute="——" entry-spacing="0" hanging-indent="true">
+  <bibliography subsequent-author-substitute="&#8212;&#8212;" entry-spacing="0" hanging-indent="true">
     <sort>
       <key macro="author"/>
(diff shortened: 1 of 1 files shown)


climate-change-economics.csl
```diff
@@ -0,0 +1,186 @@
...
+        <text macro="author-short"/>
+        <text macro="issued"/>
+        <group delimiter=" ">
+          <label variable="locator" form="short"/>
+          <text variable="locator"/>
+        </group>
+      </group>
+    </layout>
+  </citation>
+  <bibliography subsequent-author-substitute="&#8212;&#8212;" entry-spacing="0" hanging-indent="true">
```
rmzelle: @POBrien333, did you find examples where `subsequent-author-substitute` is needed? The example article mentioned by the requestor (http://www.worldscientific.com/doi/pdf/10.1142/S2010007813400058) doesn't seem to have it.

2 dashes is also rather non-standard (3 is much more common)

POBrien333: Thanks for noting this. A remnant from the template style.

rmzelle:
Thanks!

PR #2859, opened 2017-08-01
Update environmental-health-perspectives.csl

POBrien333:
via https://forums.zotero.org/discussion/66836/style-error-environmental-health-perspectives


rmzelle:
Thanks!

PR #2860, opened 2017-08-02
Create international-management.csl

POBrien333:
via https://forums.zotero.org/discussion/56088/style-request-management-international


rmzelle:
Thanks!

PR #2861, opened 2017-08-03
Update and rename dependent/annals-of-surgery.csl to annals-of-surger…

POBrien333:
via https://forums.zotero.org/discussion/66550/style-request-annals-of-surgery


rmzelle:
Thanks!

PR #2862, opened 2017-08-03
Create acta-radiologica.csl

POBrien333:
via https://forums.zotero.org/discussion/66630/acta-radiologica-reference-style

Sage Journal, but doesn't follow their style at all.


rmzelle:
Thanks!

PR #2864, opened 2017-08-03
Quick fix: climate-change-economics.csl

POBrien333:
Fix minor error: https://forums.zotero.org/discussion/comment/282060/#Comment_282060


rmzelle:
Thanks!

PR #2866, opened 2017-08-03
Create smithsonian-institution-scholarly-press-author-date

POBrien333:
via https://github.com/citation-style-language/styles/issues/2867


POBrien333:
Not sure why I can't have a suffix on date-part.
Date should be as: "15 July 2015", so no comma after the month.

rmzelle:
Haven't checked the spec, but I'm pretty sure that it's not allowed to specify a suffix on date-part elements if you use a predefined date format. The reasoning is that predefined date formats automatically localize according to the style locale, and customizing suffixes is never going to be correct for the various locales.

So you can either redefine the predefined date format in the style's locale section, or just define the date format from scratch (and remove the `form` attribute).

```
                    <date form="text" variable="accessed">
                      <date-part name="day" suffix=" "/>
                      <date-part name="month"/>
                      <date-part name="year"/>
                    </date>
```

rmzelle:
Thanks!

PR #2868, opened 2017-08-04
Update harvard-london-south-bank-university.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/282160/#Comment_282160


rmzelle:
Thanks!

PR #2869, opened 2017-08-04
Update harvard-north-west-university.csl

POBrien333:
via https://forums.zotero.org/discussion/66898/csl-style-not-citing-pages

### DIFF
--- a/harvard-north-west-university.csl
+++ b/harvard-north-west-university.csl
@@ -13,7 +13,7 @@
     </author>
     <category citation-format="author-date"/>
     <category field="generic-base"/>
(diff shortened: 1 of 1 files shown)


rmzelle:
I don't think you're stupid (in this case, at least :) ). The original style print the contents of the "page" variable in in-text citations for journal articles, which is nonsense, so this should be "locator".

rmzelle:
And thanks!

POBrien333:
Ahhhh. Glad to hear. Was doing this in-between a job interview and catching a train. :)

PR #2870, opened 2017-08-08
Update and rename dependent/veterinary-microbiology.csl to veterinary…

POBrien333:
…-microbiology.csl

via https://forums.zotero.org/discussion/66916/veterinary-microbiology-style-incorrect
Just a quick independent style with one minor change to the in-text sorting which is different to the template style.


rmzelle:
Thanks!

PR #2873, opened 2017-08-08
Create journal-of-environmental-science-and-health-part-b.csl

POBrien333:
via https://forums.zotero.org/discussion/66937/style-request-journal-of-environmental-science-and-health-part-b

Part A seems to follow the same style, part C doesn't..... How does that even makes sense?!


rmzelle:
Thanks!

> How does that even makes sense?!

We're not really in a business that makes a lot of sense ;).

PR #2875, opened 2017-08-09
Create dependent for journal-of-environmental-science-and-health-part-a.csl

POBrien333:
via https://github.com/citation-style-language/styles/pull/2873/files
Sorry, should've added this to the PR yesterday. But with all the automatic "journals" generated ones I worked on recently, I had in my head I had to wait till this is merged.


rmzelle:
Thanks!

PR #2876, opened 2017-08-09
Create teaching-mathematics-and-its-applications.csl

POBrien333:
via https://forums.zotero.org/discussion/66918/style-for-ima-journals-such-as-teaching-mathematics-and-its-applications
Going to make independents of the other IMA journals once I hear back from the requestor.


rmzelle:
Thanks!

POBrien333:
Thanks! Also for telling me about the & (and not %) in the thread.

Issue #2874, opened 2017-08-08
Update Journal of General Virology


rmzelle:
I got the following request via citationstyles.org:

> The Journal of General Virology have changed their citation format, but CSL has still the old version. Can it be updated please.
> http://www.microbiologyresearch.org/authors/information-for-authors#references

Looks like 6 journals by the Microbiology Society use the same style.

@POBrien333, could you take a look?

POBrien333:
Sure, will do. :)

POBrien333:
Just made the style here: https://github.com/citation-style-language/styles/pull/2877
Who emailed you? They could link to the CSL repo from their website. They do that for Endnote...

rmzelle:
Covered by https://github.com/citation-style-language/styles/pull/2878. Thanks!

> Who emailed you?

An academic researcher. Didn't see a link to e.g. the editorial board of the journal.

PR #2877, opened 2017-08-09
Update society-for-general-microbiology.csl

POBrien333:
via https://github.com/citation-style-language/styles/issues/2874
Change style from author-date to numeric.


POBrien333:
@rmzelle 
Made a mess here by changing  some of these by hand rather than in the template.csl. Not sure how I can remove those commits again.

rmzelle:
> Made a mess here by changing some of these by hand rather than in the template.csl. Not sure how I can remove those commits again.

That's fine. I just built onto your pull request in #2878. Thanks!

PR #2879, opened 2017-08-10
Update Journal of Orthopaedic Research, add space between title and journal

swlinderman:


rmzelle:
Thanks! I made a few additional small changes as well (for e.g. chapters, per reference 34 in http://onlinelibrary.wiley.com/doi/10.1002/jor.23452/abstract).

PR #2886, opened 2017-08-14
Create bol-pediatr.csl

palantre:
Parentheses style strip periods for use in Boletín de Pediatría (http://www.sccalp.org/boletines). Language: Es-es


rmzelle:
Thanks!

PR #2889, opened 2017-08-15
Create journal-of-soil-and-water-conservation.csl

POBrien333:
via https://forums.zotero.org/discussion/64777/in-serach-for-a-style-for-journal-of-soil-and-water-conservation


rmzelle:
Thanks!

PR #2891, opened 2017-08-15
Update oecologia-australis.csl

gudryan:


oecologia-australis.csl
```diff
@@ -61,21 +61,17 @@
...
+            <label form="short" text-case="title" prefix=" "/>
           </names>
         </group>
       </if>
     </choose>
   </macro>
   <macro name="author">
     <names variable="author">
-      <name and="symbol" delimiter-precedes-last="always" initialize-with=". " name-as-sort-order="all"/>
+      <name and="symbol" delimiter-precedes-last="always" et-al-min="1" et-al-use-first="0" initialize-with=". " name-as-sort-order="all"/>
```
rmzelle: @gudryan, these values for the "et-al-..." attributes make no sense. Is "et al." even used in the bibliography, where the "author" macro is used?

See http://docs.citationstyles.org/en/stable/specification.html#name for a description of the attributes.

oecologia-australis.csl
```diff
@@ -142,38 +138,38 @@
...
       </if>
+      <else-if type="legislation webpage" match="any">
+        <text value="Retrieved in" prefix="("/>
+        <date form="text" variable="accessed" prefix=" " suffix=", ">
+          <date-part name="day" form="ordinal"/>
+          <date-part name="month"/>
+          <date-part name="year"/>
+        </date>
+        <text value="from " prefix=" "/>
+        <text variable="URL" strip-periods="false" quotes="false" font-variant="normal" suffix=")."/>
```
rmzelle: @gudryan, it's very fragile to split parenthesis over multiple elements (`<text value="Retrieved in" prefix="("/>` and `<text variable="URL" ... suffix=")."/>`). In this case, it's much better to put everything here in a group that use parentheses:

```xml
<group prefix="(" suffix=")">
   ...
</group>
```

oecologia-australis.csl
```diff
@@ -260,10 +259,15 @@
...
-      <if type="report" match="any">
-        <group delimiter=": ">
-          <text variable="publisher-place"/>
-          <text variable="publisher"/>
+      <if type="report" match="all">
+        <group delimiter=" " prefix=" ">
+          <choose>
+            <if match="any" variable="number-of-pages publisher">
+              <text variable="number-of-pages" prefix="p. " suffix="."/>
+              <text variable="publisher" prefix="ed. " suffix=", "/>
```
rmzelle: @gudryan, it's also better to use CSL terms whenever possible, instead of hardcoded translations. E.g. instead of `<text variable="number-of-pages" prefix="p. " suffix="."/>`, use:

```xml
  <label variable="number-of-pages" form="short" plural="never"/>
  <text variable="number-of-pages"/>
</group>
```

rmzelle: `<text variable="publisher" prefix="ed. " suffix=", "/>` is also rather odd, because "ed." usually stands for "editor", not "publisher".

oecologia-australis.csl
gudryan: I did not understand what is the problem. It was the unique way I got that all authors would be showed in bibliography. Trying now, if I {exclude et-al-use-first="0"} it also works. It´s possible by this way?

oecologia-australis.csl
gudryan: I think not, because only legislation or webpages are used between parenthesis. If I put it as group role, all it will be included between parenthesis. Furthermore, if someone forget these information, it would be easier to found an error. There is some better method to solve this?

oecologia-australis.csl
gudryan: Done!

oecologia-australis.csl
gudryan: Exclude the {exclude et-al-use-first="0"} is not possible, the test report says:

1) oecologia-australis: name nodes must have valid et-al-min and et-al-use-first attributes
     Failure/Error:
       expect(min.to_i).to be > first.to_i,
         "expected et-al-min (#{min}) to be greater than et-al-use-first (#{first})"
     
       expected et-al-min (1) to be greater than et-al-use-first (1)

So, what is the exactly problem with this? If I do not include anything here, when a reference have more than a X number of author it is automatically included "..., last author" later the 6th author, but I need that all author are included on all types of references. Could you help me?

rmzelle: > I need that all author are included on all types of references

"et-al" settings are a bit tricky, because you can specify them in different places in CSL styles. To always print all authors in the bibliography, delete `et-al-min="1" et-al-use-first="0"` here in the "author" macro, and also `et-al-min="8" et-al-use-first="6" et-al-use-last="true"` at the start of the bibliography section (in the line `<bibliography hanging-indent="true" et-al-min="8" et-al-use-first="6" et-al-use-last="true" entry-spacing="0" line-spacing="2">`).

oecologia-australis.csl
rmzelle: It's better to replace

```xml
        <else-if type="legislation webpage" match="any">
          <text value="Retrieved in" prefix="("/>
          <date form="text" variable="accessed" prefix=" " suffix=", ">
            <date-part name="day" form="ordinal"/>
            <date-part name="month"/>
            <date-part name="year"/>
          </date>
          <text value="from " prefix=" "/>
          <text variable="URL" strip-periods="false" quotes="false" font-variant="normal" suffix=")."/>
        </else-if>
```

with

```xml
        <else-if type="legislation webpage" match="any">
          <group prefix="(" delimiter=" " suffix=").">
            <text value="Retrieved in"/>
            <date form="text" variable="accessed" suffix=",">
              <date-part name="day" form="ordinal"/>
              <date-part name="month"/>
              <date-part name="year"/>
            </date>
            <text term="from"/>
            <text variable="URL"/>
          </group>
        </else-if>
```

> Furthermore, if someone forget these information, it would be easier to found an error.

We greatly prefer repository styles to be written as robust as possible.

rmzelle: Also, shouldn't "Retrieved in" be "Retrieved on"?

oecologia-australis.csl
gudryan: Thank you @rmzelle!
I have changed all that you improve, and I thank for you attention in explain my doubts.
Please, see the attached modifications.
Cheers

rmzelle:
Thanks!

PR #2892, opened 2017-08-16
Update database.csl

cwtung:


adam3smith:
huh - that was an odd error, thanks for the patch.

PR #2894, opened 2017-08-16
Update arts-university-bournemouth.csl

Dbelk:
Changed 'Available to' to 'Available from'.
Removed capitalisation on [Online].
Removed quote marks from titles.
Changed to a colon after place of publication instead of a comma.
Added a full stop to the date suffix.
Changed comma after the title to a full stop.
Changed '(' and ')' on 'Accessed' statement to '[' and ']'.

### DIFF
--- a/arts-university-bournemouth.csl
+++ b/arts-university-bournemouth.csl
@@ -1,253 +1,256 @@
-<?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" default-locale="en-GB">
-  <info>
-    <title>Arts University Bournemouth</title>
-    <title-short>AUB</title-short>
-    <id>http://www.zotero.org/styles/arts-university-bournemouth</id>
-    <link href="http://www.zotero.org/styles/arts-university-bournemouth" rel="self"/>
-    <link href="http://www.zotero.org/styles/the-open-university-harvard" rel="template"/>
-    <link href="http://www.aublibrary.info/index.php/referencing" rel="documentation"/>
-    <author>
-      <name>Duncan Belk</name>
-    </author>
-    <category citation-format="author-date"/>
-    <category field="generic-base"/>
-    <updated>2017-04-24T10:20:30+00:00</updated>
-    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
-  </info>
-  <macro name="editor">
-    <names variable="editor" delimiter=", ">
-      <name and="text" initialize-with=". " delimiter=", " sort-separator=", " name-as-sort-order="all"/>
-      <label form="short" prefix=" (" suffix=")" strip-periods="true"/>
-    </names>
-  </macro>
-  <macro name="anon">
-    <text term="anonymous" form="short" text-case="capitalize-first" strip-periods="true"/>
-  </macro>
-  <macro name="author">
-    <names variable="author">
-      <name name-as-sort-order="all" and="text" sort-separator=", " delimiter=", " delimiter-precedes-last="never" form="long" initialize-with=". "/>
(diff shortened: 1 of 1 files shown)


arts-university-bournemouth.csl
```diff
@@ -174,7 +174,7 @@
       <key macro="year-date" sort="descending"/>
       <key macro="author-short"/>
     </sort>
-    <layout prefix="(" suffix=")" delimiter="; ">
+    <layout prefix="(" suffix=")." delimiter="; ">
```
rmzelle: @Dbelk, the added period here in `<layout prefix="(" suffix=")." delimiter="; ">` for in-text citations is very odd, and should probably be deleted again. Did you test your style?

With the current code, you would get citations like `(Henningsen et al. 2015).`. While citations are often placed at the end of a sentence, that's not always the case, and the closing period of a sentence is not part of the citation.

arts-university-bournemouth.csl
```diff
@@ -183,8 +183,8 @@
           </group>
         </if>
         <else>
-          <group delimiter=", ">
-            <group delimiter=", ">
+          <group delimiter=". ">
+            <group delimiter=". ">
```
rmzelle: @Dbelk, results in `(D’Arcus. 2006).` and `(D’Arcus. 2006. p. 3).`. The use of periods here is very atypical. s this correct?

arts-university-bournemouth.csl
```diff
@@ -69,15 +70,14 @@
...
+      <text term="online" prefix="[" suffix="]."/>
       <choose>
         <if variable="DOI">
           <text variable="DOI" prefix="https://doi.org/"/>
         </if>
         <else>
           <group delimiter=" ">
-            <text term="available at" text-case="capitalize-first"/>
-            <text variable="URL"/>
+            <text variable="URL" prefix="Available from: "/>
```
rmzelle: @Dbelk, we can show you how, but it's better to change the English translation of the "available at" term than putting the text in a prefix.

arts-university-bournemouth.csl
```diff
@@ -12,13 +12,13 @@
...
     <category field="generic-base"/>
-    <updated>2017-04-24T10:20:30+00:00</updated>
+    <updated>2017-08-16T12:39:28+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="editor">
     <names variable="editor" delimiter=", ">
       <name and="text" initialize-with=". " delimiter=", " sort-separator=", " name-as-sort-order="all"/>
-      <label form="short" prefix=" (" suffix=")" strip-periods="true"/>
+      <label form="short" prefix=" (" suffix=")." strip-periods="true"/>
```
rmzelle: @Dbelk, another minor nitpick, but it's better to delete both the period and the `strip-periods="true"` here, which should give the same output.

arts-university-bournemouth.csl
Dbelk: We don't want to have the full stop after the ')' in our in-text citation, only in our full reference. I just didn't know which ')' was in the full reference (i can really read csl code), so i just put one after all of them and planned to figure it out after we had tested it. I'll retry...

arts-university-bournemouth.csl
Dbelk: Again, this is because i don't know what i'm doing. We just want the full stop after the elements of the full reference, not the in-text citation.

arts-university-bournemouth.csl
Dbelk: Sorry, this was the only way i could see of using 'Available from:'. We don't want it to say 'Available at'.

arts-university-bournemouth.csl
Dbelk: Again, sorry, this was because we wanted the full stop after the ')' in the full reference so i just put one after all of the ')'s to see how that worked. 
I'll try again.

PR #2897, opened 2017-08-16
Create european-journal-of-microbiology-and-immunology.csl

POBrien333:
via https://forums.zotero.org/discussion/51727/style-request-european-journal-of-microbiology-and-immunology-eujmi


rmzelle:
Thanks!

PR #2898, opened 2017-08-17
Create fom-hochschule-fuer-oekonomie-und-management.csl

POBrien333:
Hi guys,
Making this style for somebody. it's not to be published here, but I ran into a few problems. The user is also using Mendeley, which makes it all harder.

- macro named publisher should work, but then the o. O./o. V. don't show up.
- Webpages should look like
   > Kirmsse, Stefan, Schiele, Christian, Lindemann, Christoph (o. J.): Auswirkungen Von Basel III aus Sicht der Unternehmensberatung (keine Datumsangabe.), herbert’s url, (Zugriff 19-5-2015)
   >
   > but look like
   >
   > Kirmsse, Stefan, Schiele, Christian, Lindemann, Christoph Auswirkungen Von Basel III aus Sicht der Unternehmensberatung, (o. J.), herbert’s url, (Zugriff 19-5-2015)

the style is so complex, I'm going mad.

### DIFF
--- a/fom-hochschule-fuer-oekonomie-und-management.csl
+++ b/fom-hochschule-fuer-oekonomie-und-management.csl
@@ -11,7 +11,7 @@
     </author>
     <category citation-format="note"/>
     <category field="generic-base"/>
-    <summary>Chicago format adapted for FOM Hochschule with notes and bibliography. Adapted for use with Mendeley. Short titles for note references to be saved under &quot;Notes&quot; in Mendeley.</summary>
+    <summary>Chicago format adapted for FOM Hochschule with notes and bibliography. Adapted for use with Mendeley. Short titles for note references to be saved under "Notes" in Mendeley.</summary>
     <updated>2017-08-21T22:25:39+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
@@ -746,7 +746,7 @@
       <else-if type="article-newspaper">
         <text macro="issued"/>
       </else-if>
-      <else-if type="book bill chapter graphic legislation motion_picture paper-conference report song thesis" match="any" >
+      <else-if type="book bill chapter graphic legislation motion_picture paper-conference report song thesis" match="any">
         <group delimiter=", ">
           <choose>
             <if type="thesis">
@@ -798,6 +798,16 @@
       </if>
     </choose>
   </macro>
+  <macro name="locator-note">
+    <choose>
+      <if type="article-journal book chapter webpage post-weblog" match="any">
+        <text variable="page"/>
+      </if>
+      <else>
+        <text value="o. S."/>
+      </else>
+    </choose>
+  </macro>
   <macro name="case-pinpoint-subsequent">
     <choose>
       <if type="legal_case">
@@ -836,7 +846,7 @@
             <text macro="contributors-short"/>
             <group delimiter=" ">
               <group delimiter=", ">
-			    <text variable="note"/>
+                <text variable="note"/>
                 <choose>
                   <if variable="issued">
                     <text macro="issued"/>
@@ -845,6 +855,7 @@
                     <text value="o. J."/>
                   </else>
                 </choose>
+                <text macro="locator-note"/>
                 <text macro="case-locator-subsequent"/>
               </group>
               <text macro="case-pinpoint-subsequent"/>
@@ -859,7 +870,7 @@
       </choose>
     </layout>
   </citation>
-  <bibliography subsequent-author-substitute="———" entry-spacing="0" hanging-indent="true">
+  <bibliography subsequent-author-substitute="&#8212;&#8212;&#8212;" entry-spacing="0" hanging-indent="true">
     <sort>
       <key macro="contributors-sort"/>
       <key variable="issued"/>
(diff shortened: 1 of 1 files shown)


fom-hochschule-fuer-oekonomie-und-management.csl
```diff
@@ -0,0 +1,1008 @@
...
+              <if variable="issued">
+                <date variable="issued">
+                  <date-part name="day" suffix="-"/>
+                  <date-part name="month" form="numeric" suffix="-"/>
+                  <date-part name="year"/>
+                </date>
+              </if>
+              <else>
+                <text value="keine Datumsangabe"/>
+              </else>
```
rmzelle: You never hit this `<else/>` because you already tested for the presence of the "issued" variable at the start of the macro, so you always hit the 

```xml
      <else>
        <text term="no date" form="short"/>
      </else>
```

below instead if there is no date for webpages.

fom-hochschule-fuer-oekonomie-und-management.csl
```diff
@@ -0,0 +1,1008 @@
...
+          <group delimiter=". ">
+            <text macro="originally-published"/>
+            <group delimiter=", ">
+              <text macro="reprint"/>
+              <text macro="publisher"/>
+            </group>
+          </group>
+          <text macro="issued"/>
+        </group>
+      </else-if>
```
rmzelle: @POBrien333, your other question, from your DM:

> for books/chapters, when there is either no publisher or publisher-place, I've built this macro. If both those fields are left empty, I am not getting either o. O. or o. V.

I was able to reproduce this, and finally figured it out. I usually solve these weird issues by deleting as much of the style as possible, to simplify the style and locate which elements are actually relevant. In this case, I ended up with https://gist.github.com/rmzelle/b1d797f0453dd5729576c7ba9266e276, which I was live-testing in Zotero's Style Editor.

The answer turns out rather simple: for this `<else-if/>`, you test for the presence of either the "publisher-place" or "publisher" variables (`<else-if variable="publisher-place publisher" match="any">`). So, if neither variable has a value, the "publisher" macro is never called here.

fom-hochschule-fuer-oekonomie-und-management.csl
POBrien333: Oh thank you for finding this! Thanks! Got it working now!

fom-hochschule-fuer-oekonomie-und-management.csl
POBrien333: amazing! Thanks. I'll use that trick with the minimal styles :)

rmzelle:
(is this ready to merge from your side?)

POBrien333:
Nope. Still loads of work to do and then I have to ask the requestor if she wants it to be on the repository.

fom-hochschule-fuer-oekonomie-und-management.csl
```diff
@@ -0,0 +1,956 @@
...
+        <group delimiter=" ">
+          <text variable="volume"/>
+          <text variable="container-title"/>
+        </group>
+      </if>
+    </choose>
+  </macro>
+  <macro name="locator-note">
+    <choose>
+      <if type="article-journal book chapter webpage post-weblog " match="any">
```
POBrien333: Dear @rmzelle 
Can you tell me why books don't print the pages in the notes here? It works fine for chapters. (Note the user will use Mendeley and Mendeley doesn't have the variable="locator" function like Zotero)

rmzelle: @POBrien333, Mendeley's CSL mapping webpage seems to be offline, but my guess is that Mendeley, like Zotero, doesn't have the "page" variable mapped for the "book" item type. (Zotero only has "number-of-pages" for "book", see https://aurimasv.github.io/z2csl/typeMap.xml#map-book)

POBrien333: Grand. That's what I had thought.

rmzelle: @POBrien333, Mendeley's CSL mapping is back at https://service.elsevier.com/app/answers/detail/a_id/22175/supporthub/mendeley/ (for now at least)

POBrien333:
Hey guys,
Can I test a date if it contains a day and month? I need to ouput either a full date or "without date". Would that be possible? thanks :)

adam3smith:
You mean test whether, say  issued-->date-part name="month" exists? No, unfortunately not.

POBrien333:
Yup. That's what I meant. Thank you for answering this so swiftly.

POBrien333:
I've got one more.
I had to make the style that if a bookchapter does not have an author it also does not give the editors (don't ask me why...).
Somehow this effed up the whole thing regarding years and pages. Any idea? See below:

Mares, Isabela (2001): Firms and the Welfare State: When, Why, and How Does Social Policy Matter to Employers?, in: Hall, Peter A, Soskice, David (Hrsg.), Varieties of Capitalism. The Institutional Foundations of Comparative Advantage, 2001, S. 184–213

O. V. (2012): Die neue Liquiditätsrisiko-Rahmenvereinbarung der BIS, in: Modernes Liquiditätsrisikomanagement in Kreditinstituten, S. 304–365 2012,

adam3smith:
The wrong page/year formatting should apply to all chapters (and does for me):
You first call
`<text macro="locators-chapter"/>` which produces ", S. 304-365 and then `<text macro="issue-join-with-comma" suffix=", "/> which prints " 2012,"

(You've also moved the groups so that the labels don't make sense anymore, which makes the style extra hard to read: the "with comma" group now has a space and the "with space" group no delimiter.)

adam3smith:
@POBrien333  what the status of this? Do you still want to wrap this up and get it merged?

POBrien333:
As I said before, this will not be for merging. It was a style that was meant to be paid for. Sadly the person has gone cold on me now after spending many, many hours of work on it....

PR #2902, opened 2017-08-22
Replace American Geophysical Union with new style

bwiernik:
New style is APA, but using et-al on first citation with 3+ authors (rather than only subsequent).
https://forums.zotero.org/discussion/67246/style-request-american-geophysical-union-new

Would it be better to create a generic version of APA for this and to set AGU as a dependent style? This seems like it might be a somewhat common variation.

### DIFF
--- a/american-geophysical-union.csl
+++ b/american-geophysical-union.csl
@@ -5,13 +5,33 @@
     <title-short>AGU</title-short>
     <id>http://www.zotero.org/styles/american-geophysical-union</id>
     <link href="http://www.zotero.org/styles/american-geophysical-union" rel="self"/>
-    <link href="http://publications.agu.org/author-resource-center/text-requirements/reference-format/" rel="documentation"/>
+    <link href="http://www.zotero.org/styles/apa" rel="template"/>
+    <link href="http://publications.agu.org/brief-guide-agu-style-grammar/" rel="documentation"/>
     <author>
       <name>Julian Onions</name>
       <email>julian.onions@gmail.com</email>
     </author>
+    <author>
+      <name>Simon Kornblith</name>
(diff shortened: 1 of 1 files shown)


american-geophysical-union.csl
```diff
@@ -126,109 +471,226 @@
...
+          <!---This should be localized-->
+          <text value="Original work published"/>
+          <date variable="original-date" form="text"/>
+        </group>
+      </if>
+    </choose>
+  </macro>
+  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year" givenname-disambiguation-rule="primary-name">
     <sort>
-      <key macro="year-date"/>
```
adam3smith: I think they kept their old sorting rules:


    First author alone, chronologically (most recent first):

Smith, R. (2000a). ….
Smith, R. (2000b). ….
Smith, R. (2003). ….

    With one coauthor, alphabetically by coauthor and then chronologically:

Smith, R., & Frank, L. A. (1998). ….
Smith, R., & Allen, F. A. (2001). ….
Smith, R., & Frank, L. A. (2001). ….

    With two or more coauthors, chronologically only:

Smith, R., & Roberts, D. H. (2005). ….
Smith, R., Roberts, D. H., & Jones, J. (1998). ….
Smith, R., Allen, F. A., & Baker, T. L. (1999). ….
Smith, T. (1998). ….

bwiernik: Fixed.

adam3smith:
Thanks!
(given the sort order, I don't think a generic version of this makes sense)

PR #2907, opened 2017-08-26
Update Medical Physics

wodellUF:
I made the update to this journal's style in May, 2017, and it was correct at that time.  Since that time this journal changed their style to eliminate the quotations around the journal title, as demonstrated now about 2/3 of the way down this page:  http://aapm.onlinelibrary.wiley.com/hub/journal/10.1002/%28ISSN%292473-4209/about/author-guidelines.html


rmzelle:
Thanks!

PR #2916, opened 2017-08-30
Fixing the style for 'Austral Ecology' based on user feedback

cparnot:
See user feedback from the 2 commits. Here is the final comment from him on this version:

> I tried it and it seems to work perfect!


rmzelle:
Thanks!

PR #2917, opened 2017-08-31
Create genes-brains-and-behavior.csl

POBrien333:
via https://forums.zotero.org/discussion/67197/genes-brain-and-behavior-missing-citation#latest

### DIFF
--- a/genes-brains-and-behavior.csl
+++ b/genes-brains-and-behavior.csl
@@ -1,0 +1,231 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>Genes, Brain and Behavior</title>
+    <id>http://www.zotero.org/styles/genes-brains-and-behavior</id>
+    <link href="http://www.zotero.org/styles/genes-brains-and-behavior" rel="self"/>
+    <link href="http://www.zotero.org/styles/journal-of-zoology" rel="template"/>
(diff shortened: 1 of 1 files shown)


genes-brains-and-behavior.csl
```diff
@@ -0,0 +1,231 @@
...
+      <else>
+        <text variable="container-title" form="short" strip-periods="true" font-style="italic"/>
+        <text variable="collection-title" prefix=", "/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="author">
+    <names variable="author">
+      <name name-as-sort-order="all" and="symbol" sort-separator=", " initialize-with="." delimiter=", " delimiter-precedes-last="never"/>
+      <label form="short" prefix=" (" suffix=")" text-case="capitalize-first"/>
```
adam3smith: probably hard to find an example, but unless you have one to the contrary, make this lowercase & strip periods just like for editors in chapter citations.

genes-brains-and-behavior.csl
```diff
@@ -0,0 +1,231 @@
...
+      <else>
+        <text variable="edition" suffix="."/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="locators">
+    <choose>
+      <if type="article-journal article-magazine article-newspaper" match="any">
+        <group prefix=" " delimiter=", ">
+          <group>
```
adam3smith: no need for this group

adam3smith:
Thanks!

rmzelle:
("brains" in the style ID should be "brain"; will fix)

rmzelle:
Style ID fixed in https://github.com/citation-style-language/styles/pull/2928.

PR #2919, opened 2017-09-01

roaldfrosig:
'original-date' is now ignored, and the publiced year for book chapters are moved forward to just after the author name.


rmzelle:
Thanks!

PR #2922, opened 2017-09-03
default to long form title

michaelnetbiz:


adam3smith:
@bwiernik -- you did a more thorough review of this recently: can you see any reason why the style is using short titles?

bwiernik:
That was an error on my part copying the italics logic from the author-substitute Macro. The change is correct. It should also be made two lines above in the other branch of the choose node.

adam3smith:
Thanks both!

PR #2925, opened 2017-09-04
Create indian-dermatology-online-journal.csl

POBrien333:
via https://forums.zotero.org/discussion/64623/style-request-superscript-with-square-bracket-after-the-punctuation-marks


rmzelle:
Thanks!

PR #2926, opened 2017-09-04
Create clinical-anatomy.csl

POBrien333:
via https://forums.zotero.org/discussion/66430/clinical-anatomy-style-is-not-in-the-repository/p1


adam3smith:
Thanks!

POBrien333:
Thanks for fixing those things.

PR #2927, opened 2017-09-04

roaldfrosig:
Forgot to remove publication year after the editors name.


adam3smith:
thanks!

PR #2931, opened 2017-09-05
Create international-journal-of-urban-and-regional-research.csl

POBrien333:
via https://forums.zotero.org/discussion/47900/citation-for-international-journal-of-urban-and-regional-research


rmzelle:
Thanks!

PR #2948, opened 2017-09-14
create integrated-research-advances.csl

iraeditor:
This is the citation style for integrated research advances


rmzelle:
Thanks, but it looks like all ISP journals use the same citation format, so I created so-called dependent CSL styles for them: https://github.com/citation-style-language/styles/pull/2992

PR #2949, opened 2017-09-15
create integrated-science.csl

iraeditor:
This common citation style for all journals published by integrated science


adam3smith:
Thank you. Before I review this in greater detail, is there any better documentation about the journals' reference style you could provide? The page you currently link to doesn't have any information on formatting references I could find. Also, do you happen to work for/with the publisher?

rmzelle:
> Before I review this in greater detail, is there any better documentation about the journals' reference style you could provide? The page you currently link to doesn't have any information on formatting references I could find.

See e.g. http://www.pubs.iscience.in/journal/index.php/cbl/about/submissions#authorGuidelines

> Citation Manager style files: Authors are highly encouraged to use the citation manager programs for citing and formatting the references. Authors may use free Mendelay, Zotero, Paper 2 or Endnote program. If you are using any of these citation manager program for Bibliography (References and Notes), then you can download citation style file for journal **(common for ISP journals)**:
>
> Mendelay, Zotero, Paper 2 style file: ISP Journals CSL style file"

(emphasis mine; @iraeditor, "Mendeley" is misspelled as "Mendelay", by the way)

rmzelle:
@iraeditor, do you know if this CSL style was derived from another, and if so, which one? We always try to include a "template" link (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link).

rmzelle:
Thanks! (we're still curious about what this style was derived from)

PR #2954, opened 2017-09-18

mserifovic:
Added period after "(eds.)"
Updated et al settings


rmzelle:
Thanks! 🌍

PR #2955, opened 2017-09-18
Update biological-reviews.csl

POBrien333:
Per https://forums.zotero.org/discussion/comment/286593/#Comment_286593

- fix a double space issue
- adapt et al rules for bibliography to 16/15 as mentioned in guidelines.


rmzelle:
Thanks!

PR #1988, opened 2016-03-29
Create hiob-ludolf-centre-for-ethiopian-studies-web.csl

eu-genia:
A version of the bibliographic style of the Hiob Ludolf Centre for Ethiopian Studies for web publications (with DOIs/URLs)

PR #2956, opened 2017-09-18
changes to hiob ludolf

PietroLiuzzo:
added style with full names
added class encyclopaedia entry everywhere

This was elaborated by prof. bausi to obtain a result closer to the one desired for the Center publications with regard to encyclopaedia entry, e.g.

Bausi, Alessandro 2010. ‘Wängelä Wärq (Golden Gospel)’, in Alessandro Bausi and Siegbert Uhlig, eds, Encyclopaedia Aethiopica, IV (Wiesbaden: Harrassowitz, 2010), 1130b–1132a.
instead of what we got before
Bausi, A. 2010. ‘Wängelä Wärq (Golden Gospel)’, Encyclopaedia Aethiopica, IV (Wiesbaden: Harrassowitz, 2010), 1130b–1132a.

### DIFF
--- a/hiob-ludolf-centre-for-ethiopian-studies-long-names.csl
+++ b/hiob-ludolf-centre-for-ethiopian-studies-long-names.csl
@@ -2,7 +2,7 @@
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
   <info>
     <title>Hiob Ludolf Centre for Ethiopian Studies (long names)</title>
-    <title-short>HLCES (Hamburg) Names Long </title-short>
+    <title-short>HLCES (Hamburg) long names</title-short>
     <id>http://www.zotero.org/styles/hiob-ludolf-centre-for-ethiopian-studies-long-names</id>
     <link href="http://www.zotero.org/styles/hiob-ludolf-centre-for-ethiopian-studies-long-names" rel="self"/>
     <link href="http://www.zotero.org/styles/hiob-ludolf-centre-for-ethiopian-studies" rel="template"/>
--- a/hiob-ludolf-centre-for-ethiopian-studies-long-names.csl
+++ b/hiob-ludolf-centre-for-ethiopian-studies-long-names.csl
@@ -1,16 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
   <info>
-    <title>Hiob Ludolf Centre for Ethiopian Studies Names Long</title>
+    <title>Hiob Ludolf Centre for Ethiopian Studies (long names)</title>
     <title-short>HLCES (Hamburg) Names Long </title-short>
-    <id>http://www.zotero.org/styles/hiob-ludolf-centre-for-ethiopian-studies-names-long</id>
-    <link href="http://www.zotero.org/styles/hiob-ludolf-centre-for-ethiopian-studies-names-long" rel="self"/>
+    <id>http://www.zotero.org/styles/hiob-ludolf-centre-for-ethiopian-studies-long-names</id>
+    <link href="http://www.zotero.org/styles/hiob-ludolf-centre-for-ethiopian-studies-long-names" rel="self"/>
+    <link href="http://www.zotero.org/styles/hiob-ludolf-centre-for-ethiopian-studies" rel="template"/>
     <link href="http://www1.uni-hamburg.de/COMST/bulletin.html" rel="documentation"/>
     <author>
       <name>Eugenia Sokolinski</name>
       <email>eugenia.sokolinski@uni-hamburg.de</email>
     </author>
-<contributor>
+    <contributor>
       <name>Alessandro Bausi</name>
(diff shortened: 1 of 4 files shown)


adam3smith:
Thanks! The encyclopedia changes look good.
But is there no way to avoid having three different styles for a single research center? Multiplying the number of styles in the repository has real logistical costs.

PietroLiuzzo:
@adam3smith apparently not, we have hesitated about a year to add this third one, but it looked like the only way. We are happy to consider options of course! For example, is there a way to set options in the style for example for the full names instead of the initials? 

We try all to use the Zotero Group library and the centre main style, and we ask authors and contributors to use them trying to be as consistent as possible, but, I am told by the editors, the different publications produced by the centre do have traditionally differences in the way they do bibliography. 

We could circulate the new style with full names locally if it cannot be hosted here.

rmzelle:
> is there a way to set options in the style for example for the full names instead of the initials?

No, there is not.

> the different publications produced by the centre do have traditionally differences in the way they do bibliography.

Do you have a list available of which publications use which style variant?

PietroLiuzzo:
I forwarded the question to the author of the style and the publications editor. Thanks for your patience.
the DOI-URL was motivated here already https://github.com/citation-style-language/styles/pull/1988#issuecomment-203143158
The HLCES style is used by Aethiopica, Aethiopistische Forschungen and COMSt (see dependant styles) however Aethiopica publishes with each yearly issue a bibliography of linguistic studies and this one needs the full names.

PietroLiuzzo:
Please wait, I found out a number of other things which need to be fixed and I am going to update in the pull request. Thanks again for your patience.

PietroLiuzzo:
I have done all the changes I needed to do. I hope it is now ok. Thanks a lot!

PietroLiuzzo:
I apologise for the further update to the pull request. I was notified, unfortunately with bad timing of two further issues with the style which I am about to fix, namely the editor=translator role handling and the treatment of thesis. Thanks again

PietroLiuzzo:
Yes, now it should definitely be all, I hope it is ok. Thanks a lot

hiob-ludolf-centre-for-ethiopian-studies-long-names.csl
```diff
@@ -0,0 +1,347 @@
...
+  <locale xml:lang="en-GB">
+    <terms>
+      <term name="ordinal">th</term>
+      <term name="ordinal-01">st</term>
+      <term name="ordinal-02">nd</term>
+      <term name="ordinal-03">rd</term>
+      <term name="ordinal-04">th</term>
+      <term name="ordinal-11">th</term>
+      <term name="ordinal-12">th</term>
+      <term name="ordinal-13">th</term>
```
rmzelle: I reindented all three styles with http://formatter.citationstyles.org/, and changed the name of the new style slightly to be more consistent with the rest of our styles.

I think the "ordinal" terms in the locale section of all three styles can be removed. The "ordinal-04" term should be unnecessary for English, and all the other terms are already defined in the British English CSL locale file: https://github.com/citation-style-language/locales/blob/da68506f8f47089d28a8289db75f2f5b7f8e5000/locales-en-GB.xml#L71

PietroLiuzzo: Thanks a lot! I do not know why they were put there in the first place to be honest. might have been inherited from the style used to begin the editing of the new one. I am removing them now.

PietroLiuzzo:
Can I pose a further question, whilst I look forward to the approval of this updates? 

This style is now for us very good but I encounter some behaviours which I am not sure about. 

We have records like this in our group library
https://www.zotero.org/groups/358366/ethiostudies/items/itemKey/B66AKH8J

If I call the zotero api (although I assume there is still the old version of the style there, not the one in this PR), I get what I want, i.e. 

> Lepage, C. and J. Mercier 2011–2012. ‘Un tétraévangile illustré éthiopien à cycle long du XVe siècle. Codicologie et iconographie’, _Cahiers Archéologiques_, 54 (2011–2012), 99–174.

if instead with the same style I get the bibliography entry from the Zotero Standalone (latest) I get 

> Lepage, C. and J. Mercier 2011. ‘Un tétraévangile illustré éthiopien à cycle long du XVe siècle. Codicologie et iconographie’, _Cahiers Archéologiques_, 54 (2011), 99–174.

is it something I should change in the style which I am still in time to add to this PR?
Thanks a lot

rmzelle:
Thanks!

> If I call the zotero api (although I assume there is still the old version of the style there, not the one in this PR), I get what I want, i.e.
> 
>     Lepage, C. and J. Mercier 2011–2012. ‘Un tétraévangile illustré éthiopien à cycle long du XVe siècle. Codicologie et iconographie’, Cahiers Archéologiques, 54 (2011–2012), 99–174.
> 
> if instead with the same style I get the bibliography entry from the Zotero Standalone (latest) I get
> 
>     Lepage, C. and J. Mercier 2011. ‘Un tétraévangile illustré éthiopien à cycle long du XVe siècle. Codicologie et iconographie’, Cahiers Archéologiques, 54 (2011), 99–174.
> 
> is it something I should change in the style which I am still in time to add to this PR?

I'd ask this question on the Zotero forums. The Zotero API and Standalone program likely use slightly different versions of the citeproc-js CSL processor which could explain the different results. (the output of the API looks correct)

PietroLiuzzo:
Thanks a lot!

PR #2959, opened 2017-09-19
Create nuclear-technology.csl

POBrien333:
https://forums.zotero.org/discussion/67258/style-request-nuclear-technology


rmzelle:
Thanks!

I created an independent style for the American Nuclear Society and three dependents for their three journals, which appear to all use the same reference style.

POBrien333:
Great! I was wondering if I should do it that way or just make the dependents for the nuclear tech style.
Thanks for doing that :)

rmzelle:
(a Google search showed that the ANS style appears to be used beyond their journals, so it made sense to give the organization a style and add journal-dependents)

PR #2960, opened 2017-09-19
Create journal-of-molecular-recognition.csl

POBrien333:
https://forums.zotero.org/discussion/67489/style-request-journal-of-molecular-recognition-wiley-vch
basically AMA with minimal changes.


rmzelle:
Thanks!

PR #2962, opened 2017-09-19
Create chicago-17-fullnote-bibliography.csl

Elbar914:
CSL to accommodate specifications in 17th edition of Chicago Manual of Style.

### DIFF
--- a/chicago-17-fullnote-bibliography.csl
+++ b/chicago-17-fullnote-bibliography.csl
@@ -1,0 +1,1337 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="display-and-sort" page-range-format="chicago">
+  <info>
+    <title>Chicago Manual of Style 17th edition (note-bib)</title>
+    <id>http://www.zotero.org/styles/chicago-17-fullnote-bibliography</id>
+    <link href="http://www.zotero.org/styles/chicago-17-fullnote-bibliography" rel="self"/>
+    <author>
+      <name>Elise Barbeau</name>
+      <email>ebarbeau@chegg.com</email>
+    </author>
+    <category citation-format="note"/>
+    <category field="generic-base"/>
+    <summary>Chicago format with full notes and bibliography per 17th edition</summary>
+    <updated>2017-09-08T15:49:53+00:00</updated>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks for these -- you don't happen to have a list of changes you implemented versus the 16th edition, do you? I'll look at a diff otherwise, but I figured I'd ask.

Elbar914:
@adam3smith I don't have a list handy of exactly what was changed within the CSLs (changing variables, etc.) but the changes all had to do with the following updates to the style:

The changes I made to chicago-17-fullnote-bibliography had to do with: 
- Titles of websites that have a print counterpart should be in italics, while ones that are standalone resources should not.
- The use of "ibid" is no longer needed when repeating footnote citations. Instead, repeating the author's name and including the applicable page number if available is preferred.

For chicago-17-author-date:
- In an author-date reference list entry, the year may be repeated for sources that are also identified by month and day, such as journals.

adam3smith:
Thanks! I'll spend some time on this over the weekend to review and see exactly how we want to organize the 16-->17 edition transition.
@rmzelle What do you want to do about transitioning all the Chicago styles to 17th edition? I'm not too happy with the thought of having all the different copies of CMoS for both editions, but unclear what else we could do. Overall the changes are fairly small, but there are some other ones (e.g. proper DOI format)

Elbar914:
Great, thanks!

rmzelle:
> What do you want to do about transitioning all the Chicago styles to 17th edition? I'm not too happy with the thought of having all the different copies of CMoS for both editions, but unclear what else we could do.

Looks like the 17th edition was released earlier this month (https://www.amazon.com/Chicago-Manual-Style-17th/dp/022628705X lists September 5, 2017 for the print edition), so we probably shouldn't wait to long to update the CSL styles.

I think we should update all variants in-place to the 17th edition, and maybe just create 16th edition styles for the most popular variants (https://github.com/zotero/bundled-styles has chicago-author-date.csl, chicago-fullnote-bibliography.csl, and chicago-note-bibliography.csl; somebody recently mentioned what Mendeley styles are included by default (maybe on the Zotero forums? Can't remember), but https://blogs.cranfield.ac.uk/is/cranfield-harvard-mendeley and https://github.com/citation-style-language/styles/issues/1720#issuecomment-261983934 suggests Mendeley ships chicago-author-date.csl and that chicago-fullnote-bibliography is quite popular, so I'd just do those three).

adam3smith:
That sounds good -- I'll turn that into a pull request, hopefully over the weekend.

adam3smith:
supserseded by https://github.com/citation-style-language/styles/pull/3021

PR #2964, opened 2017-09-19
Update sodertorns-hogskola-harvard.csl

Ca7w3as3l:
Given name disambiguation
Remove [Online]

### DIFF
--- a/sodertorns-hogskola-harvard.csl
+++ b/sodertorns-hogskola-harvard.csl
@@ -186,13 +186,6 @@
       </if>
     </choose>
   </macro>
-  <macro name="online">
-    <choose>
-      <if variable="URL" match="any">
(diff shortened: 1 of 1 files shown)


rmzelle:
Continued in #2965.

@Ca7w3as3l, we generally prefer it if people update their original pull request if there are issues, instead of creating new ones. Each pull request has instructions on how to do so (see e.g. https://github.com/citation-style-language/styles/pull/2964#issuecomment-330692473 above)

Ca7w3as3l:
There is an update for this file.

Ca7w3as3l:
sodertorns-hogskola-harvard.csl and the related sodertorns-hogskola-harvard-ibid.csl are ready to be added to the repository. Do I just leave them as open until they are accepted? I am still unsure of how this works.

PR #2970, opened 2017-09-21
Create sodertorns-hogskola-harvard-ibid.csl

Ca7w3as3l:
Add ibid.
Remove online.

### DIFF
--- a/sodertorns-hogskola-harvard-ibid.csl
+++ b/sodertorns-hogskola-harvard-ibid.csl
@@ -1,0 +1,352 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
+  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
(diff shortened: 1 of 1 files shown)


rmzelle:
@Ca7w3as3l, are you planning to submit any more style variants after this one?

rmzelle:
(replaces #2966, #2967 and #2968)

Ca7w3as3l:
Oops.
Is this a poor practice? One file would be best if one could check a box in a Zotero menu for activating a feature.

I just replied to your mail with the question of practice. Use a language restricted style or allow users to select from the list? It is not my intention to pollute the database with variants.

rmzelle:
> Is this a poor practice?

Well, we just try to find the right balance between accepting styles for the wide variety of citation formats in use while keeping the repository from getting overly messy. My main question was whether you were planning to submit more variants like this "ibid" version, which is separate from the language issue.

> Use a language restricted style or allow users to select from the list? It is not my intention to pollute the database with variants.

Both Mendeley and Zotero nowadays allow users to select the language to use a style in, so there is no real need anymore to create language-specific styles for e.g. Swedish and English. I asked my fellow maintainer @adam3smith for his opinion as well.

PR #2947, opened 2017-09-14
Create sodertorns-hogskola-harvard.csl

Ca7w3as3l:
A working draft of a Harvard citation style meeting Södertörn University's criteria.
14 Sep 2017

PR #2972, opened 2017-09-22
Create sodertorns-hogskola-harvard-ibid.csl

Ca7w3as3l:
Added Swedish locale and plural page correction.
This was submitted previously and is now updated. If "new file" is not the way to do this, I apologise. Please correct me.
There will be only 2 version of this style. One using ibid and one without.

### DIFF
--- a/sodertorns-hogskola-harvard-ibid.csl
+++ b/sodertorns-hogskola-harvard-ibid.csl
@@ -7,6 +7,7 @@
     <id>http://www.zotero.org/styles/sodertorns-hogskola-harvard-ibid</id>
     <link href="http://www.zotero.org/styles/sodertorns-hogskola-harvard-ibid" rel="self"/>
     <link href="http://www.zotero.org/styles/sodertorns-hogskola-harvard" rel="template"/>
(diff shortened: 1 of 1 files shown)


Ca7w3as3l:
sodertorns-hogskola-harvard-ibid.csl and the related sodertorns-hogskola-harvard.csl are ready to be added to the repository. Do I just leave them as open until they are accepted? I am still unsure of how this works.

rmzelle:
(replaces #2970; no-ibid variant in #2964)

rmzelle:
> sodertorns-hogskola-harvard-ibid.csl and the related sodertorns-hogskola-harvard.csl are ready to be added to the repository. Do I just leave them as open until they are accepted? I am still unsure of how this works.

I added the no-ibid variant already in #2947 on September 16th, and I just added the ibid variant in this pull request. I did just accept #2964, after accepting #2965 on September 20th, both of which updated the no-ibid version.

> This was submitted previously and is now updated. If "new file" is not the way to do this, I apologise. Please correct me.

Well, you already figured out in this pull request on how to make additional changes, so in the future just try to do that. Having multiple pull requests per style makes things harder for us to track.

PR #2973, opened 2017-09-22
Create apa-everyurldatestamped.csl

alexmitrani:
The only change is to make sure each and every URL has an access date included in the bibliography.  See lines 167 - 174 in particular. 
I propose this variation on the existing style because anything on the Internet is subject to potentially being deleted in future.

### DIFF
--- a/apa-everyurldatestamped.csl
+++ b/apa-everyurldatestamped.csl
@@ -9,7 +9,7 @@
 	<link href="http://www.zotero.org/styles/apa" rel="template"/>
     <author>
       <name> Alex Mitrani</name>
-      <email>alex.mitrani@gmail.com</email>
+      <email>alex.mitrani@sdgworld.net</email>
     </author>	
     <category citation-format="author-date"/>
     <category field="psychology"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
you're obviously welcome to customize your styles, but we won't take this for the repo - APA is quite clear that they don't want access dates on most URLs.
Thanks for putting this up, though!

alexmitrani:
Hi Sebastian

Thanks for your reply.  I understand APA will not want it.  But it would
seem like copying if I called it something else as it is 99.9% APA.  Is
there a recommended way of sharing things like this?  I felt obliged to
share it due to the "sharealike" license.

Kind regards

Alex Mitrani


On 22 Sep 2017 8:39 p.m., "Sebastian Karcher" <notifications@github.com>
wrote:

you're obviously welcome to customize your styles, but we won't take this
for the repo - APA is quite clear that they don't want access dates on most
URLs.
Thanks for putting this up, though!

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/citation-style-language/styles/pull/2973#issuecomment-331583707>,
or mute the thread

adam3smith:
ah, no, you don't have to share it actively. You just can't put a non-free license on it, _if_ you share it. You can point interested folks to a copy e.g. here if you want

alexmitrani:
OK, thanks for clarifying that.

Alex

PR #2974, opened 2017-09-23
Create microbial-cell.csl

s-hofer:

PR #2975, opened 2017-09-23
Create cell-stress.csl

s-hofer:


rmzelle:
Is this for a journal? If so, please provide a link to the journal's website.

(and see https://github.com/citation-style-language/styles/wiki/Style-Requirements in general; the style is e.g. missing "documentation" and "template" links, and doesn't validate)

rmzelle:
(see also #2974)

PR #2976, opened 2017-09-24
MLA 8th edition update

### LINKED ISSUE #2763: Update MLA from 7th to 8th edition
Per https://github.com/citation-style-language/styles/issues/2700#issuecomment-304764851, we currently have "modern-language-association-8th-edition.csl" (8th edition) and "modern-language-association.csl" (7th edition). Eventually we should rename "modern-language-association-8th-edition.csl" to "modern-language-association.csl" (and set up a redirect in renamed-styles.json for this), and "modern-language-association.csl" to  "modern-language-association-7th-edition.csl".

@adam3smith indicated in the thread above that we still need to double-check the DOI behavior in the 8th edition. See also https://forums.zotero.org/discussion/comment/261399/#Comment_261399

adam3smith:
I think this should be right.

Closes #2763.


rmzelle:
@adam3smith, I assumed you forgot to rename "modern-language-association-underline" to "modern-language-association-7th-edition-underline"? And we'll just leave the "with-url" and "underline" at the 7th edition?

(I also sorted renamed-styles.json, which caught a few entries that weren't alphabetical)

adam3smith:
yeah, I think I wondered if we can just delete it. It's 2017. But as long as we don't, yes, we should rename.

rmzelle:
If we don't get any requests to update either variant to the 8th edition, we can drop them when we move to the 9th edition :).

I was also thinking we could maybe drop the 6th edition, which was released in 2003 according to https://en.wikipedia.org/wiki/MLA_Handbook#Editions, with the 7th edition being released in 2009. (we did get a stray PR for the 6th in 2015, though, so maybe it's still in use somewhere)

rmzelle:
(also of interest, but I assume you already saw this: https://www.mla.org/MLA-Style/What-s-New-in-the-Eighth-Edition)

rmzelle:
Thanks! :)

adam3smith:
The 6th edition style is only for the note style, which was last included in the 6th edition and so may be worth keeping.

rmzelle:
> The 6th edition style is only for the note style, which was last included in the 6th edition and so may be worth keeping.

Yeah, was thinking that too.

Also just tweeted about the change, since it's a bundled Zotero style, so quite a few people might notice: https://twitter.com/csl_styles/status/913604122573774853

PR #2978, opened 2017-09-25
Update ecology-letters.csl

POBrien333:
via https://forums.zotero.org/discussion/50635/style-error-ecology-letters#latest


adam3smith:
Thanks!

PR #2981, opened 2017-09-25

POBrien333:
Make style sort by appearance rather than alphabetically according this guidelines.
Via: https://forums.zotero.org/discussion/comment/287268/#Comment_287268


adam3smith:
Thanks!

PR #2982, opened 2017-09-25
Update scandinavian-journal-of-medicine-and-science-in-sports.csl

POBrien333:
via https://forums.zotero.org/discussion/67806/style-request-update-scandinavian-journal-of-medicine-science-in-sports


rmzelle:
Thanks!

PR #2986, opened 2017-09-27
Create medecine-intensive-reanimation.csl

POBrien333:
via https://forums.zotero.org/discussion/67839/stle-request-modify-intensive-care-medicine-style


rmzelle:
Thanks!

PR #2996, opened 2017-10-02
Update clinical-gastroenterology-and-hepatology.csl

POBrien333:
Found some minor errors when double checking the style needed by a user: https://forums.zotero.org/discussion/67910/including-all-citations-in-one-bracket-instead-of-multiple-individual-brackets


adam3smith:
Thanks!

PR #2999, opened 2017-10-02
Create pan-african-medical-journal.csl

POBrien333:
kinda via https://forums.zotero.org/discussion/67896/pamj-style

Made my own style. The one on their website is full of mistakes.


rmzelle:
Thanks! Renamed the style since e.g. http://www.panafrican-med-journal.com/content/article/28/91/full/ says "Cite this: The Pan African Medical Journal. 2017;28:91. doi:10.11604/pamj.2017.28.91.10986", and https://www.ncbi.nlm.nih.gov/nlmcatalog/101517926 also uses "The".

And would you mind sending a quick email to inform them that you made a more robust CSL style for their journal, and that they can start using that? The instructions to authors page says "Assistance with the process of manuscript preparation and submission is available from the customer support team (services@panafrican-med-journal.com).", so you could try there.

PR #3005, opened 2017-10-04
Properly support personal communications in APA

bwiernik:
APA specifies that personal communications should be cited with a
unique in-text style and should be omitted from the bibliography. This
change supports that.

The in-text format is:
(J. P. Campbell, personal communication, August 16, 2010)

To get the “personal communication” term, I applied a locale to
“interview”, which fits conceptually and is not otherwise used in APA.
This could be changed if a “personal-communication” term is added in a
future CSL update (which has been suggested). If there is a better way
to do that, let me know.

If this works, I can make parallel changes to the other APA styles.

### DIFF
--- a/apa.csl
+++ b/apa.csl
@@ -44,6 +44,7 @@
         <single>trans.</single>
         <multiple>trans.</multiple>
       </term>
+      <term name="interview">personal communication</term>
     </terms>
   </locale>
   <macro name="container-contributors">
@@ -650,17 +651,38 @@
       </if>
     </choose>
   </macro>
+  <macro name="personal_communication">
+    <group delimiter=", ">
+      <names variable="author">
+        <name and="symbol" sort-separator=", " initialize-with=". " delimiter=", " delimiter-precedes-last="always"/>
+      </names>
+      <text term="interview"/>
+      <date variable="issued">
+        <date-part name="month"/>
+        <date-part name="day" prefix=" "/>
+        <date-part name="year" prefix=", "/>
+      </date>
+      <text macro="citation-locator"/>
+    </group>
+  </macro>
   <citation et-al-min="6" et-al-use-first="1" et-al-subsequent-min="3" et-al-subsequent-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year" givenname-disambiguation-rule="primary-name">
     <sort>
       <key macro="author"/>
       <key macro="issued-sort"/>
     </sort>
     <layout prefix="(" suffix=")" delimiter="; ">
-      <group delimiter=", ">
-        <text macro="author-short"/>
-        <text macro="issued-year"/>
-        <text macro="citation-locator"/>
-      </group>
+      <choose>
+        <if type="personal_communication interview">
+          <text macro="personal_communication"/>
+        </if>
+        <else>
+          <group delimiter=", ">
+            <text macro="author-short"/>
+            <text macro="issued-year"/>
+            <text macro="citation-locator"/>
+          </group>
+        </else>
+      </choose>
     </layout>
   </citation>
   <bibliography hanging-indent="true" et-al-min="8" et-al-use-first="6" et-al-use-last="true" entry-spacing="0" line-spacing="2">
@@ -670,22 +692,26 @@
       <key macro="title"/>
     </sort>
     <layout>
-      <group suffix=".">
-        <group delimiter=". ">
-          <text macro="author"/>
-          <text macro="issued"/>
-          <text macro="title-plus-extra"/>
-          <text macro="container"/>
-        </group>
-        <text macro="legal-cites"/>
-        <text macro="locators"/>
-        <group delimiter=", " prefix=". ">
-          <text macro="event"/>
-          <text macro="publisher"/>
-        </group>
-      </group>
-      <text macro="access" prefix=" "/>
-      <text macro="original-date" prefix=" "/>
+      <choose>
+        <if type="personal_communication interview" match="none">
+          <group suffix=".">
+            <group delimiter=". ">
+              <text macro="author"/>
+              <text macro="issued"/>
+              <text macro="title-plus-extra"/>
+              <text macro="container"/>
+            </group>
+            <text macro="legal-cites"/>
+            <text macro="locators"/>
+            <group delimiter=", " prefix=". ">
+              <text macro="event"/>
+              <text macro="publisher"/>
+            </group>
+          </group>
+          <text macro="access" prefix=" "/>
+          <text macro="original-date" prefix=" "/>
+        </if>
+      </choose>
     </layout>
   </bibliography>
 </style>


adam3smith:
I'm pretty sure this will throw an error in the bibliography -- CSL (or rather: citeproc-js) doesn't like empty references.

bwiernik:
It works fine in Zotero 5.0.22-beta.6+f7e2cd634 (with and without Propachi enabled). I think that Frank changed citeproc-js's behavior at some point to ignore empty bibliography entries, rather than throwing an error.

rmzelle:
> I think that Frank changed citeproc-js's behavior at some point to ignore empty bibliography entries, rather than throwing an error.

But even if citeproc-js can handle this, it's not behavior that's addressed at all in the CSL 1.0.1 spec.

Also, documents can currently quite reliably switch between numeric and author-date styles. But in my experience, numeric styles still want authors to use full in-text descriptions of personal communication, which we (currently) can't support. (see e.g. http://aem.asm.org/site/misc/journal-ita_org.xhtml#04: "In the text, references are cited parenthetically by number in sequential order. Data that are not published or not peer reviewed are simply cited parenthetically in the text")

bwiernik:
If the spec doesn’t support it, then this change probably shouldn’t be merged obviously (Though I think we should consider it for a future CSL version).

I’m not sure I follow your second comment. Seems to me the solution would be for a similar change to be made to numeric styles with such a requirement? I’m not sure how this change here impacts changing between styles in a meaningful way?

adam3smith:
The problem with numeric styles v would be that they then skip a number in the reference list where the blank citation is (since every item has a citation - number

bwiernik:
I see, so supporting reference-free citations in numeric styles would involve a bigger change to CSL.

rmzelle:
> I see, so supporting reference-free citations in numeric styles would involve a bigger change to CSL.

Yes. It's worth leaving this pull request open if people have thoughts on how to best implement this, but to properly support this for numeric styles we'd need quite big changes to CSL. And we don't want styles that break documents in a weird way when users switch from an author-date to a numeric format.

> The problem with numeric styles v would be that they then skip a number in the reference list where the blank citation is (since every item has a citation - number

Yeah, for numeric styles these in-text personal-communication entries are typically entirely detached from the bibliography.

bwiernik:
At present, I think most users who want to enter in-text personal-communication entries just manually type them, but I think it would be nice for these citations of this sort to be handled by CSL. When you say this style would break documents in a weird way, that is because a person-communication item would go from being cited in text as (J. Campbell, personal communication, 2017) with no bibliography entry to a number with a bibliography entry (rather than staying as parenthetical in-text note)?

adam3smith:
The choices you have for numeric styles currently are
a) Print a reference number and an entry in the reference list, which is wrong or
b) Print a parenthetical citations and no entry in the reference list, which leads to a skipped number in the reference list.

Both fall under "weird failure," I'd say, the latter more so than the former.

bwiernik:
It seems like there would need to be an option to specify "no-number" for these types of citations in numeric styles (which might also be needed for history/religious studies--not sure how common numeric styles are in those fields?).

bwiernik:
It seems like adding a "no-number" property to the citation layout (which skips number for the citations it is applied to) would be a relatively straightforward way to address these types of citations without much in the way of side effects.

bwiernik:
@adam3smith Do we want to move this issue about no-Number items somewhere else?

fbennett:
@bwiernik could you clarify the no-number proposal? By property on the citation layout, do you mean an attribute somewhere within the scope of `cs:citation`? Or do you mean a value passed as a property on the JSON citation object fed to a CSL processor?

bwiernik:
So, for example, a journal might require that unpublished data or personal communications be cited in text as “(Author, Year, unpublished data)” and not included in the bibliography. For author-date styles, this produces no real problem—the item can just be be excluded from the `cs:bibliography` layout use `choose`. 

For numeric styles, it is more complex. The items need to both be excluded from the bibliography and not take up a citation number in the list. 

Thinking about this issue, I think a general solution might be best. Another common request is to be able to have separate bibliography layouts for different types of items (primary vs secondary, cases vs articles, etc.). Both the no-number issue and the separate bibliography issue could be addressed by allowing multiple layouts to be specified. `choose`-like conditionals can be used to allocate items to the correct layout, and there can be argument given to `layout` to specify what numbering sequence to use (e.g., maybe these could take arbitrary user-specified names `number-sequence="primary"`, `number-sequence="secondary"`, `number-sequence="none"`, with `number-sequence="default"` if none is specified). Each `number-sequence` would get its own numbering.

fbennett:
> Both the no-number issue and the separate bibliography issue could be addressed by allowing multiple layouts to be specified. choose-like conditionals can be used to allocate items to the correct layout, and there can be argument given to layout to specify what numbering sequence to use

That sounds pretty complicated, but setting aside the specifics, it sounds like the idea would be to implement citation-only items in the CSL of the style. Got that - it makes sense (since in a numeric style a non-bib citation would need to be something descriptive instead of a number).

Thinking about implementation, though, this would take a lot of work. To skip item numbers, the processor would need to maintain a separate numbering sequence for display purposes, which is doable, but not trivial. We would also have a problem with styling on `cs:layout`, since we may not want descriptive refs to come out superscripted or inside square braces. I guess that would be a point in favor of multiple layouts, but that would be very hard to implement in a flexible way.

There would also be the possibility that a non-bib descriptive item ends up placed in the middle of a series of numeric-ref items in a single citation. That would raise some tough problems. If the layout switched mid-citation, which affixes should be applied, and where should they go? Should the items be reordered by layout? If so, in what order should they be presented?

I see the attraction of managing and citing non-bib items in the usual way, but an attempt at full automation would hit a bunch of tough issues.

As an second-best alternative, though, perhaps the target type (like `personal_communication`) could be given a descriptive form, and be included in the bibliography, but applied by the user by clipboard-copy, pasting it into the manuscript as text. Would that suffice?

fbennett:
After sleeping on it, I've come up with a workaround for the number-sequence issue, by pushing to-be-excluded items to the bottom of the bibliography before applying `citation-number`. Not a complete solution, but it could be combined with an option to, say, suppress items of a particular type in the bibliography. You would set up a macro like this:
```
  <macro name="category">
    <choose>
      <if type="personal_communication">
        <text value="1"/>
      </if>
      <else>
        <text value="0"/>
      </else>
    </choose>
  </macro>
```
Then the `cs:sort` node would look like this:
```
    <sort>
      <key macro="category"/>
    </sort>
```
With that code, the `personal_communication` items will appear at the end of a numbered bibliography, assigned with the highest `citation-number` values. Those entries can then be removed at the last stage of manuscript preparation. If a `no-number` (or `no-bib`?) option were added to CSL, the entries could be suppressed automatically. [1]

In many numeric styles, you would still have the problem that formatting and affixes set on `cs:layout` would be applied equally to the descriptive citation form used for the "skipped" entries. But it's at least a partial solution.

-------------------

[1] In testing, I get an error string for entries with no content in the bibliography. If no-output entries are just disappearing in Zotero, there's probably something else going on there.

bwiernik:
Sure will do.

PR #3007, opened 2017-10-04
Create cancer-translational-medicine.csl

POBrien333:
via https://forums.zotero.org/discussion/67980/bibliography-as-superscript


rmzelle:
Thanks!

PR #3008, opened 2017-10-05

Irazall:
There was a problem with the citation of literature with more than two authors with the same year. It is fixed now.


rmzelle:
Thanks!

PR #3012, opened 2017-10-07
Update dataset formatting in APA

bwiernik:
Title should be italicized:
http://blog.apastyle.org/apastyle/2013/12/how-to-cite-a-data-set-in-apa-style.html

Additionally, it seems reasonable that many implementations/users might use `genre`
rather than `medium` to specify dataset types


adam3smith:
Yes, that works for me. If you can apply this to all other APAs, that'd be great. Thanks!

bwiernik:
Will do

PR #3014, opened 2017-10-07

ahahl:
- minor correction to avoid repetitions
- added spaces within abbreviated first names.
Thanks !


rmzelle:
Thanks!

PR #3016, opened 2017-10-08
Fix dataset formatting in other APA styles

bwiernik:
Also remove errant ampersand from apa-no-ampersand.csl


adam3smith:
Thanks!

PR #3020, opened 2017-10-08
Update sodertorns-hogskola-harvard-ibid.csl

Ca7w3as3l:
Fixed patent, volume(issue) and minimised url usage. With exception to ibid, now identical to non-ibid file.


rmzelle:
(replaced by #3033)

PR #3022, opened 2017-10-09
Create international-body-psychotherapy-journal.csl (waiting for reply from editors)

POBrien333:
via https://forums.zotero.org/discussion/comment/288471/#Comment_288471

apa.csl might not be the right version of APA. Let's see.


rmzelle:
Are we certain we need an independent for this?

POBrien333:
Yeah, as far as I understand he needs apa without DOI, but WITH issue.

bwiernik:
I think we should contact the journal first. Given that the journal frequently includes DOIs in their papers, I would guess the poster is mistaken.

POBrien333:
@bwiernik 
I have contacted them now. :)

rmzelle:
@POBrien333, heard back yet?

POBrien333:
No. I'm afraid no.

POBrien333:
I'll close this for now. If I ever get an answer, we can reopen/merge.

PR #3023, opened 2017-10-09
Create harvard-southern-cross-university.csl

POBrien333:
via https://forums.zotero.org/discussion/64998/style-request-scu-havard


rmzelle:
@POBrien333, did you check whether this style is the same as https://www.zotero.org/styles?q=id%3Astyle-manual-australian-government? http://libguides.scu.edu.au/harvard says they're using that:

> This guide follows the standards described in the 'Style Manual for Authors, Editors and Printers'  6th edition,  often referred to as the 'AGPS Manual' as it was formerly published by the Australian Government Publishing Services. It is also referred to as the 'Snooks manual', because the 6th edition was prepared by the consultants Snooks and Co.  This guide is also compatible with the Harvard SCU style used in Endnote software.  

(also, for new styles, I'm preferring southern-cross-university-harvard.csl over harvard-southern-cross-university.csl)

POBrien333:
Great spot! Thanks for that! Found a little mistake in the parent style and fixed that.

style-manual-australian-government.csl
```diff
@@ -50,15 +50,17 @@
...
-          <date variable="accessed">
-            <date-part name="day" prefix=" "/>
-            <date-part name="month" prefix=" "/>
-            <date-part name="year" prefix=" "/>
-          </date>
+        <group delimiter=", ">
+          <group>
+            <text value="viewed"/>
+            <date variable="accessed">
+              <date-part name="day" prefix=" "/>
```
rmzelle: @POBrien333, this prefix should also be deleted, I think.

rmzelle: Actually, it should be moved to the parent group as a delimiter.

rmzelle:
Thanks!

PR #3026, opened 2017-10-09
Create the-canadian-journal-of-psychiatry.csl

POBrien333:
via https://forums.zotero.org/discussion/68057/style-request-canadian-journal-of-psychiatry


rmzelle:
Thanks!

PR #3027, opened 2017-10-10
Create concepts-in-magnetic-resonance-part-a.csl

POBrien333:
via http://support.mendeley.com/customer/en/portal/questions/17161550-can-t-install-new-style?new=17161550

Guidelines don't match actual papers.


rmzelle:
Thanks!

PR #3032, opened 2017-10-11
Update sodertorns-hogskola-harvard.csl

Ca7w3as3l:
Update


rmzelle:
Thanks! (I again deleted any commented-out XML code)

PR #3034, opened 2017-10-11
Create monash-university-harvard.csl

POBrien333:
via https://forums.zotero.org/discussion/62833/style-request-monash-harvard


rmzelle:
@POBrien333, looks like this could potentially be another dependent of https://github.com/citation-style-language/styles/blob/master/style-manual-australian-government.csl. http://guides.lib.monash.edu/citing-referencing/harvard says:

> This guide is based on Commonwealth of Australia 2002, Style Manual for authors, editors and printers, 6th edn, rev. by Snooks & Co, John Wiley & Sons, Australia, Milton; and a wide range of online Harvard Guides were also consulted.

Could you check if the format is the same?

POBrien333:
I did see that this time round, but it's different. :)

rmzelle:
Great, thanks for checking!

Issue #2906, opened 2017-08-25
Wrong Brain, behavior and evolution.csl style


egisholt:
Sorry I am new here, I was trying to format my Bibliography according to the journal and the csl style is completely different. I have been trying to edit the style but I can´t sort the Bibliography alphabetically instead of appereance. Checking the editor I saw that it refers to this http://www.karger.com/Journal/Guidelines/254338#11

But that corresponds to the Guidelines of Acta Cytologica journal instead of Brain, Behavior and Evolution.

I am not sure how to change that.

Thanks for any help

adam3smith:
Generally we prefer if you report errors to & through your reference manager. Which are you using?

egisholt:
Hi I am sorry I just found your page. I use Mendeley

Thanks

adam3smith:
You should be able to use the style "Karger journals" which looks correct to me while we fix this.

egisholt:
Hello Sebastian

Thanks for all your help. Unfortunately Karger journals style is still different than the one that they require in Brain,Behavior and Evolution. I have modified it using the editor in Mendeley but I can´t find were to change the order of the Bibliography. Karger Journals sorts references by appearance while Brain, Behavior and Evolution sorts bibliography by alphabetical order. 

https://www.karger.com/Journal/Guidelines/223831#06 <https://www.karger.com/Journal/Guidelines/223831#06>

I am sorry that the publishers use different styles

Hope you can help me otherwise I have to do it manually

Thank you

Eduardo

adam3smith:
How quickly do you need this?

egisholt:
I just need this to submit the paper, but don't worry otherwise I can do it manually 

Thanks for all the help

Eduardo Garza Gisholt

rmzelle:
We now have a corrected style for Brain, Behavior and Evolution. Sorry for the wait, and thanks to @POBrien333 for doing the work!

PR #3038, opened 2017-10-12
Update karger-journals-author-date.csl

POBrien333:
Closes https://github.com/citation-style-language/styles/issues/2906

Generally for these Karger author-date journals:
Bibliography: citations are definitely sorted by author name. That was completely missing.
found some additional mistakes.

The brain, behavior and evolution journal: need to make an independent style for it. different style.


rmzelle:
Thanks!

Issue #3039, opened 2017-10-12
Taylor & Francis journal style requests


rmzelle:
We had a user request via the CSL website for:

International Journal of Polymeric Materials and Polymeric Biomaterials
http://www.tandfonline.com/action/authorSubmission?show=instructions&journalCode=gpom20

and

Materials Technology
http://www.tandfonline.com/action/authorSubmission?show=instructions&journalCode=ymte20#References

(this one seems to use standard https://www.zotero.org/styles/taylor-and-francis-national-library-of-medicine?source=1 )

@POBrien333? If it saves you some time I can prepare a full style request for the first journal.

rmzelle:
(and I can generate the second journal style via the "journals" repo)

POBrien333:
I'll make the independent style for the first one. Great if you can add the 2nd one for the automatic generation. :)

rmzelle:
Thanks @POBrien333!

PR #3042, opened 2017-10-13
Create international-journal-of-polymeric-materials-and-polymeric-bio…

POBrien333:
…materials.csl

via https://github.com/citation-style-language/styles/issues/3039


rmzelle:
Thanks!

PR #3044, opened 2017-10-13
Add Archives of Industrial Hygiene and Toxicology style

makso1212:
This is the style of the journal Arhiv za higijenu rada i toksikologiju (0004-1254), which is largely based on the Vancouver style.


rmzelle:
@makso1212, the style is currently fixed to US English. Is that okay? We can also remove the preset locale and let people pick their own when they use the style.

https://hrcak.srce.hr/aiht?lang=en says:

> Scientific papers should be written in English, and manuscripts in other categories may be in either Croatian or English. Announcements, book reviews and meeting reports are also accepted, and may be written in Croatian, English or Slovene.

rmzelle:
Thanks!

PR #3047, opened 2017-10-14

CharlesQc:
1. Added dot at the end of each cite group, as requested by the department. 
2. A cite can now have no locator and will display the full page range or number of pages as a reference. It implies that both op. cit. and ibid. have to be modified:
 - op. cit. are now also used for two subsequent cites without locator, as to prevent a situation where a cite group containing a cite without locator is followed by that single cite without locator and that it only shows "Ibid" without mention of which cite of the cite group is included in the ibid. This was confirmed by a member of the department as being acceptable.
 - op. cit. are now also used for subsequent cite without locator immediately following the same cite with a locator, as to prevent any confusion about the subsequent use of that locator.


rmzelle:
Thanks!

PR #3049, opened 2017-10-16
Create Springer Basic (Note)

bwiernik:
A footnote style, requested here:
https://forums.zotero.org/discussion/62377/is-there-a-way-to-put-springer-basic-into-footnotes

### DIFF
--- a/springer-basic-author-date-no-et-al.csl
+++ b/springer-basic-author-date-no-et-al.csl
@@ -67,13 +67,13 @@
   <macro name="title">
     <choose>
       <if type="book">
-        <group delimiter=", " suffix=". ">
+        <group delimiter=", ">
           <text variable="title"/>
           <text macro="edition"/>
         </group>
       </if>
       <else>
-        <text variable="title" suffix=". "/>
+        <text variable="title"/>
       </else>
     </choose>
   </macro>
@@ -111,7 +111,7 @@
              In: Smith J (ed) The rise of modern genomics, 3rd edn.
              Wiley, New York, pp 230-257 -->
         <if type="chapter paper-conference" match="any">
-          <group delimiter=" " suffix=". ">
+          <group delimiter=" " prefix=". ">
             <text term="in" text-case="capitalize-first" suffix=":"/>
             <names variable="editor">
               <name sort-separator=" " initialize-with="" name-as-sort-order="all" delimiter=", " delimiter-precedes-last="always"/>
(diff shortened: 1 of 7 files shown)


bwiernik:
My last commit (which includes final requested changes from requester/publisher and all styles validate) isn't showing up here. Should I close and re-open the PR?

adam3smith:
Looks like this solved itself?

bwiernik:
Yeah, I guess so. Weird, never had GitHub take more than a few seconds to update a PR. Ready for review.

rmzelle:
I touched up the submission a little bit. @bwiernik, can you add a "template" link to the new style? It would also be nice if we could find a URL for the style guide that is hosted on the springer.com domain (Springer has some maintenance going on right now, so a lot of pages aren't loading for me, but I can check this later).

(also, in general, it's handier to create separate pull requests for submitting new styles and touching up existing ones)

bwiernik:
Added a template link.

With Springer's site down, I wasn't able to find a link to the Law style instructions hosted on Springer's page. I asked the requester to supply one or to contact Springer for one.

(Sorry about mixing new/updated styles. I'll keep that in mind for the future.)

bwiernik:
Okay, I can't find a copy of these specific guidelines on the Springer site anywhere (all of their legal book author guidelines just point to the same general manuscript guidelines page), but this is the style they use with law books. What documentation link should I use?

rmzelle:
> What documentation link should I use?

Well, we can use the link to the copy on https://www.ligue.org, but it would be good to save a copy to GitHub by adding it as an attachment in a GitHub comment here in case the link goes dead. And I saw you already asked on the forums whether there is a Springer-hosted copy online, which is good. The current document is also from 2010, which seems a bit old, so maybe Springer has something newer too.

bwiernik:
Okay, here is a copy of the documentation as an archive. I haven't gotten a response from the style requester about a more current link to documentation. This style does appear to still be in use by Springer's law book, following its standard Springer Basic style, but as notes and with abbreviated citations for articles, books, and chapters as implemented in this style. The link to an example open access chapter reflects that.

[Springer Instructions for Authors_ Law.pdf](https://github.com/citation-style-language/styles/files/1413896/Springer.Instructions.for.Authors_.Law.pdf)

rmzelle:
I split these changes to existing Springer styles and the new style addition into two new pull requests, so I'll close this.

rmzelle:
(accepted all changes in #3078 and #3079)

PR #3056, opened 2017-10-16
Create uoy-american-psychological-association-6th-edition.csl

AVWoman:
University of York APA 6th style


bwiernik:
What are there differences between this style and the standard APA style (I didn’t do a full comparison)?

PR #3059, opened 2017-10-17
Update iica-catie.csl

danirv91:
I updated the style to the 5th edition (from 2016). So I included changes related with online papers, books, articles...

### DIFF
--- a/iica-catie.csl
+++ b/iica-catie.csl
@@ -3,7 +3,7 @@
   <info>
     <title>IICA-CATIE (Spanish)</title>
     <id>http://www.zotero.org/styles/iica-catie</id>
-    <link href="http://www.zotero.org/styles/iica-catie" rel="template"/>
+    <link href="http://www.zotero.org/styles/iica-catie" rel="self"/>
     <link href="http://www.iica.int/es/publications/redacci%C3%B3n-de-referencias-bibliogr%C3%A1ficas-normas-t%C3%A9cnicas-para-ciencias-agroalimentarias" rel="documentation"/>
     <author>
       <name>Marcelo Wall</name>
--- a/iica-catie.csl
+++ b/iica-catie.csl
@@ -1,18 +1,18 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="es-ES">
   <info>
-    <title>IICA-CATIE 5 ed. (Spanish)</title>
-    <id>http://csl.mendeley.com/styles/476532851/iica-catie-2</id>
+    <title>IICA-CATIE (Spanish)</title>
+    <id>http://www.zotero.org/styles/iica-catie</id>
     <link href="http://www.zotero.org/styles/iica-catie" rel="template"/>
-    <link href="http://repiica.iica.int/docs/B4013e/B4013e.pdf" rel="documentation"/>
+    <link href="http://www.iica.int/es/publications/redacci%C3%B3n-de-referencias-bibliogr%C3%A1ficas-normas-t%C3%A9cnicas-para-ciencias-agroalimentarias" rel="documentation"/>
     <author>
-      <uri>http://www.mendeley.com/profiles/daniela-rodrguez7/</uri>
-      <name>Daniela Rodriguez</name>
-      <email>daniela.rodriguez@iica.int</email>
-    </author>
-    <contributor>
       <name>Marcelo Wall</name>
       <email>wa.marc@yahoo.com</email>
+    </author>
+    <contributor>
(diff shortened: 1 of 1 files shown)


rmzelle:
(identical submission to #3060, so let's continue here and close the other pull request)

rmzelle:
Thanks for submitting a pull request! I switched back Marcelo Wall as author (since he originally contributed the style), and switched back the title and "self" link.

I noticed a few things that still need a closer look before we can accept this.

iica-catie.csl
```diff
@@ -59,24 +58,22 @@
...
-                <text macro="volume" prefix="Año " suffix=", "/>
+                <text macro="volume" prefix=" "/>
               </if>
-              <else>
-                <text value="s.d. "/>
-              </else>
             </choose>
-            <text variable="issue" prefix="N° "/>
+            <text variable="issue" prefix="(" suffix=")"/>
+            <text variable="page" form="short" strip-periods="false" font-variant="normal" display="block" prefix=":"/>
```
rmzelle: `strip-periods="false" font-variant="normal" display="block"` should be removed here.

iica-catie.csl
```diff
@@ -208,6 +199,9 @@
       <name form="short" and="text" delimiter=", " delimiter-precedes-last="never" initialize-with=". "/>
     </names>
   </macro>
+  <macro name="Editor">
+    <text term="editor" form="short" suffix="(ed.)."/>
+  </macro>
```
rmzelle: This is incorrect. You always want to print the names of the "editor" variable through `<names/>`, not `<text/>`. See the "author-citation" macro directly above as an example (or look at the "editor" macro of the original style).

iica-catie.csl
```diff
@@ -414,21 +438,76 @@
...
+            </if>
+          </choose>
+          <text macro="publisher-place" suffix=", "/>
+          <text macro="publisher" suffix=". "/>
+          <text variable="number" suffix=". "/>
+          <date variable="issued">
+            <date-part name="day" suffix=" "/>
+            <date-part name="month" form="short"/>
+          </date>
+          <text variable="page" display="block" suffix=" p. "/>
```
rmzelle: `display="block"` should be deleted here as well.

iica-catie.csl
```diff
@@ -406,6 +429,7 @@
               </else>
             </choose>
           </group>
+          <text value="(En línea, sitio web)." suffix=" "/>
```
rmzelle: This line is also a little strange, in view of `<text value=" (en línea)"/>` on line 428. For a "webpage" item with no URL but a value for "genre", you will get " (en línea). (En línea, sitio web). " here.

iica-catie.csl
danirv91: ok

iica-catie.csl
danirv91: ok

iica-catie.csl
danirv91: This is ok, this is because Mendeley has the possibility of import a citation for a website, and our style has a very specific way to cite a website. (en línea), applies to any online document

iica-catie.csl
danirv91: ok

danirv91:
Hello, I've changed a link in "documentation". The link that I put is an absolute link to the norm.

iica-catie.csl
```diff
@@ -234,118 +228,147 @@
...
             <text macro="author-short"/>
             <text macro="year-date"/>
             <text macro="title"/>
+            <choose>
+              <if match="any" variable="URL">
+                <text term="online" prefix="(" suffix=")"/>
+              </if>
+            </choose>
             <text macro="translator"/>
             <text macro="editor"/>
```
rmzelle: @danirv91, the style also doesn't validate right now. One reason is that you call the "editor" macro twice (here and somewhere further down), but you removed this macro.

rmzelle: The other issue is that the style defines a number of macros that aren't used, which we don't allow for repository styles (unused macros should just be deleted). These macros are: "Editor", "container-author", "date", "issue".

danirv91: Hello, if those macros are not being used, so they can be deleted. Regarding the macro "editor", I'm afraid of removing that calls, I don't know If that could affect the style. Sorry I don't know much about this language, I developped the style with the visual editor, and I only saw that it works good.

rmzelle: @POBrien333, do you have time to look this over? (you could also restart from the current repository version and check the new version of the style guide, if that's any easier)

If not, let me know, and I'll fix up the current submission.

POBrien333:
Sure. I'll look at it

POBrien333:
@rmzelle 
this one can be closed as it was dealt with in #3066.

PR #3062, opened 2017-10-19
Fixed citation sorting and missing book suffix.

dehling:
Added the missing sorting for citations (oldest first) and added the missing suffix "." for book citations.

### DIFF
--- a/european-journal-of-information-systems.csl
+++ b/european-journal-of-information-systems.csl
@@ -5,7 +5,7 @@
     <title-short>EJIS</title-short>
     <id>http://www.zotero.org/styles/european-journal-of-information-systems</id>
     <link href="http://www.zotero.org/styles/european-journal-of-information-systems" rel="self"/>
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks! Are these corrections based on the instructions at http://www.palgrave.com/gp/journal/41303/authors/presentation-formatting? (it looks like we need to update the "documentation" link in the style to this URL)

dehling:
Yes. They are based on https://www.palgrave.com/gp/journal/41303/authors/presentation-formatting

"In a list of references within the text put oldest first."

"Books

WALSHAM G (1993) Interpreting Information Systems in Organizations. Wiley, Chichester."

rmzelle:
Thanks!

I updated the "documentation" link, and I removed the second sort on the "issued" variable, which was rather redundant. If you see any proof that the journal sorts by full date (including month and day) instead of just by year, let me know, and I'll change the style to just sort by the "issued" variable.

PR #3063, opened 2017-10-20

Irazall:
Changed the suffix for contributers from ":" to ","


rmzelle:
Thanks!

(I replaced a few prefixes and suffixes by group delimiters, which are a little bit more robust; none of the changes should negatively affect the style)

Irazall:
@rmzelle Thank you very much!

PR #3069, opened 2017-10-25
Raffles Bulletin of Zoology

EverLearner:
It is a Mendeley citation style for a Singapore journal Raffles Bulletin of Zoology, published by Lee Kong Chian Natural History Museum. Loosely based on APA 6th style.

Reference: https://lkcnhm.nus.edu.sg/publications/raffles-bulletin-of-zoology/raffles-bulletin-zoology/

### DIFF
--- a/raffles-bulletin-of-zoology.csl
+++ b/raffles-bulletin-of-zoology.csl
@@ -9,6 +9,7 @@
     <link href="http://lkcnhm.nus.edu.sg/nus/index.php/instructions" rel="documentation"/>
     <author>
       <name>Kin Guan Wee</name>
+      <email>wee.kinguan@nus.edu.sg</email>
     </author>
     <category citation-format="author-date"/>
     <category field="zoology"/>
@@ -86,7 +87,7 @@
       </if>
       <else>
         <names variable="author">
-          <name form="short" and="symbol" initialize-with=". " et-al-subsequent-min="3" et-al-min="3"/>
+          <name form="short" and="symbol" initialize-with=". "/>
           <substitute>
             <names variable="editor"/>
             <names variable="translator"/>
--- a/raffles-bulletin-of-zoology.csl
+++ b/raffles-bulletin-of-zoology.csl
@@ -623,7 +623,7 @@
       </if>
     </choose>
   </macro>
-  <citation et-al-use-first="1" et-al-min="6" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year" givenname-disambiguation-rule="primary-name">
+  <citation et-al-use-first="1" et-al-min="3" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year" givenname-disambiguation-rule="primary-name">
     <sort>
       <key macro="issued-sort"/>
       <key macro="author"/>
--- a/raffles-bulletin-of-zoology.csl
+++ b/raffles-bulletin-of-zoology.csl
@@ -1,0 +1,661 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
+  <info>
+    <title>Raffles Bulletin of Zoology</title>
+    <title-short>RBZ</title-short>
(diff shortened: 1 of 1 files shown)


raffles-bulletin-of-zoology.csl
```diff
@@ -0,0 +1,661 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
+  <info>
+    <title>Raffles Bulletin of Zoology</title>
+    <title-short>RBZ</title-short>
+    <id>http://csl.mendeley.com/styles/476141641/rbz</id>
```
POBrien333: It's not validating as the id needs to be in the same style as the self-link in line 7. It also needs to match your file name.

EverLearner: Thanks for the comment! I will change it soon, supposedly http://www.zotero.org/styles/raffles-bulletin-of-zoology

raffles-bulletin-of-zoology.csl
```diff
@@ -0,0 +1,661 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
+  <info>
+    <title>Raffles Bulletin of Zoology</title>
+    <title-short>RBZ</title-short>
+    <id>http://www.zotero.org/styles/raffles-bulletin-of-zoology</id>
+    <link href="http://www.zotero.org/styles/raffles-bulletin-of-zoology" rel="self"/>
```
POBrien333: also, please add a template link like in line 7 with "template" with the style name for the style you based it on.

raffles-bulletin-of-zoology.csl
```diff
@@ -0,0 +1,661 @@
...
+            <name delimiter=", " and="symbol" initialize-with=". "/>
+            <label form="short" prefix=", " text-case="title"/>
+          </names>
+        </group>
+      </if>
+    </choose>
+  </macro>
+  <macro name="author">
+    <names variable="author">
+      <name and="symbol" initialize-with="" sort-separator=" " name-as-sort-order="all" et-al-subsequent-use-first="100" et-al-subsequent-min="101" et-al-use-first="100" et-al-min="101" delimiter-precedes-last="never"/>
```
POBrien333: It's not validating because of these et-al rules here.
If you want always all authors shown, just remove those 4 et-al rules here.

raffles-bulletin-of-zoology.csl
```diff
@@ -0,0 +1,661 @@
...
+    <choose>
+      <if variable="original-date">
+        <group delimiter=" " prefix="(" suffix=")">
+          <text value="Original work published"/>
+          <date form="text" variable="original-date"/>
+        </group>
+      </if>
+    </choose>
+  </macro>
+  <citation et-al-subsequent-use-first="1" et-al-subsequent-min="3" et-al-use-first="1" et-al-min="6" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year" givenname-disambiguation-rule="primary-name">
```
POBrien333: it's better to have the et-al rules here in citation and bibliography respectively. what exactly do you want to achieve?

raffles-bulletin-of-zoology.csl
EverLearner: is it something like this?
<link href="http://www.zotero.org/styles/apa" rel="template"/>

raffles-bulletin-of-zoology.csl
EverLearner: Yes I want to show all authors. Thanks for the tip!

raffles-bulletin-of-zoology.csl
EverLearner: In citation, I want to show that if less than 3 authors are present, show the full list of authors. If greater or equal to 3, show the first author followed by et al.

In bibliography, there is no such limit.

raffles-bulletin-of-zoology.csl
POBrien333: yes

rmzelle:
Thanks! (I updated the documentation link and corrected the et-al settings)

PR #3070, opened 2017-10-25
Added Event-Place to Conference-Papers

Irazall:
Event-Place was missing in Conference-Papers, hence it was added


rmzelle:
Thanks!

PR #3072, opened 2017-10-26
Setted an order for inline citation

Irazall:
Setted an order for inline citation for more than one citation


rmzelle:
@Irazall, it looks like you removed the sorting from the bibliography when you added it to the inline citations. I assume that's a mistake?

I also can't automatically merge this because the changes overlap with those from https://github.com/citation-style-language/styles/pull/3070. Could you submit a new pull request?

rmzelle:
(and for the new pull request, please start from the latest version in the repository, e.g. by editing https://github.com/citation-style-language/styles/blob/master/german-council-of-economic-experts.csl)

PR #3074, opened 2017-10-27
 Setted an order for inline citation

Irazall:
Setted an order for inline citation for more than one citation


rmzelle:
Thanks, that looks better!

PR #3075, opened 2017-10-28

POBrien333:
quick fix via https://forums.zotero.org/discussion/68358/style-error-journal-of-visualized-experiments
this style would need major reworking. Very simplistic style prone to errors.


rmzelle:
Thanks as always!

PR #3079, opened 2017-10-30
Add Springer Basic Note

rmzelle:
Replaces https://github.com/citation-style-language/styles/pull/3049


rmzelle:
@bwiernik, thanks! If we get a better documentation link we can also replace that later.

PR #3082, opened 2017-10-31
Create chemie-ingenieur-technik.csl

POBrien333:
via https://forums.zotero.org/discussion/68385/chemie-ingenieur-und-technik-style
I made my own style.


rmzelle:
Documentation link wasn't updated, but otherwise good. Thanks!

PR #3083, opened 2017-10-31
Added issue-number to journal articles

Irazall:
Now the issue-number is seen in parenthesis after the volume of an article

### DIFF
--- a/german-council-of-economic-experts.csl
+++ b/german-council-of-economic-experts.csl
@@ -15,7 +15,7 @@
     <category citation-format="author-date"/>
     <category field="political_science"/>
     <summary>Style for the German Council of Economic Experts</summary>
-    <updated>2017-10-27T07:31:47+00:00</updated>
+    <updated>2017-10-31T15:17:20+00:00</updated>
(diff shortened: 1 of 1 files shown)


german-council-of-economic-experts.csl
```diff
@@ -293,6 +294,9 @@
       </if>
     </choose>
   </macro>
+  <macro name="doi">
+    <text variable="DOI" form="short" prefix=", doi:"/>
```
adam3smith: do you have any input on the style format? DOIs should ideally be displayed as URLs (see e.g. [here](http://www.crossref.org/02publishers/doi_display_guidelines.html)), i.e. `prefix=", https://doi.org/"`. We'll take it like this if these are the style requirements and you can't change them, but I though I'd flag it.

rmzelle:
Thanks!

I improved the style a little. It's usually better to get rid of one-line macros that are only used once. I also used some group delimiters to replace some of the prefix and suffixes, which should be more robust (the "issued" macro is a little complicated in that regard). I just hand edited this without testing, but I'm pretty sure there should be no changes to the output.

PR #3086, opened 2017-11-01
Create chicago-fullnote-bibliography-with-ibid-indonesia

husnulfikr:
Translate Chicago Manual of Styles 17th edition in Indonesia. The translation in locales-id-ID.xml latest file is not applied in zotero yet. So I create this translation with little change from translation in locales-ID.xml.


rmzelle:
Thanks.

Did you base your translation on any official Indonesian Chicago style guide, or is this more of a personal translation of the style into Indonesian?
Also, did you have to change more than just the locale translations to get a style that is correct in Indonesian?

husnulfikr:
I did not find any official Indonesian Chicago style. I 'll be glad if I
find it.

So, I just put the code from local-ID.xml to the official Chicago style
(english) with little change.

I am a lecturer in my college and I want my student to use zotero in their
papers. But, the translation used in zotero is not complete.

The first think I wanted to do is to contribute to the translation in
local-ID.xml. But, I noticed that the file already has much better
translation that the one used in zotero. Although I made litle change.

So, basically, I just put the code from local-ID.xml to the official
chicago style with little changes to it's translation.

And I think it can be used not just by students but also anyone in my
country who want to use zotero in Indonesian language. So I uploaded it to
the repository.

On Nov 3, 2017 07:09, "Rintze M. Zelle" <notifications@github.com> wrote:

Thanks.

Did you base your translation on any official Indonesian Chicago style
guide, or is this more of a personal translation of the style into
Indonesian?
Also, did you have to change more than just the locale translations to get
a style that is correct in Indonesian?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/citation-style-language/styles/pull/3086#issuecomment-341594216>,
or mute the thread

adam3smith:
@husnulfikr -- this sounds like something we should just be able to solve with Zotero rather than having to create a whole new style. Which version of Zotero have you tested this with? What's an example of an incorrect translation that's correct in the locale?

rmzelle:
> I noticed that the file already has much better translation that the one used in zotero.

Somebody just contributed some changes to our Indonesian locale file earlier this week: https://github.com/citation-style-language/locales/pull/166. These changes should show up in the next Zotero release.

We usually only accept language-specific version of our main style guide styles like Chicago, if the automatic translation of the regular styles is incorrect and can't be fixed. In this case, it sounds like it might be enough to wait until Zotero updates its locale files?

husnulfikr:
> Somebody just contributed some changes to our Indonesian locale file earlier this week: citation-style-language/locales#166. These changes should show up in the next Zotero release.

I did not pay attention to the local-ID update time. Sorry for my silly mistake: D
I agree, it's better to wait for the next zotero release. 
Thank you very much.

rmzelle:
Well, it's just a coincidence that somebody contributed some translations so recently. The previous update to the Indonesian locale was from more than a year ago.

Let us know if there are still any more improvements we can make to the Indonesian CSL locale file at https://github.com/citation-style-language/locales/blob/master/locales-id-ID.xml.

Otherwise, Zotero usually updates its CSL locale files with every release, so our current version of the locale file should turn up in the latest version of Zotero in the next few weeks (releases have been quite frequent lately: https://www.zotero.org/support/5.0_changelog).

PR #3087, opened 2017-11-01
Update journal-of-plant-nutrition-and-soil-science.csl

POBrien333:
https://forums.zotero.org/discussion/comment/290617/#Comment_290617
Quick easy fix. :)


adam3smith:
quick easy merges, too. Thanks!

PR #2993, opened 2017-09-30
Update uludag-universitesi-sosyal-bilimler-enstitusu-full-note.csl

mtarakci:
So sorry for such a quick and near update.

PR #3089, opened 2017-11-03
Update uludag-universitesi-sosyal-bilimler-enstitusu-full-note.csl

mtarakci:
Follow-up of https://github.com/citation-style-language/styles/pull/2993#issuecomment-341628367


rmzelle:
Thanks!

PR #3090, opened 2017-11-03
Update uludag-universitesi-sosyal-bilimler-enstitusu-full-note-no-ibi…

mtarakci:
…d.csl

Follow-up of https://github.com/citation-style-language/styles/pull/2993#issuecomment-341628367


mtarakci:
This is "with ibid" version. Its old name was "no-ibid". So there appears an error at the Travis CI. 
I do not know what I should do at this stage.

rmzelle:
Thanks!

PR #3094, opened 2017-11-03
kerala-agricultural-university-format.csl

kaulis:


rmzelle:
Thanks. Is there a webpage by Kerala Agricultural University where this citation format is described?

PR #3095, opened 2017-11-05
BES journals now all use (old-DOI format) APA

adam3smith:
See also https://github.com/citation-style-language/journals/commit/e089bb9ee97712ed6a2e1db4cf024c2e5e1c696b


rmzelle:
(added a link to the matching "journals" repo commit in your opening post)

How did you find out this was the case? And are you sure the locale should be "en-US", not "en-GB"? Otherwise looks good to merge.

adam3smith:
I got an error report from Mendeley about the BES style, then started looking at them in detail. en-US vs. GB doesn't make a difference with APA styles, but I figured since they use APA, en-US made sense.

rmzelle:
Okay, thanks! Just trying to make sure we have a record.

rmzelle:
Could you add the dependents for "Journal of Applied Ecology" and "Methods in Ecology and Evolution"? I think you just missed them (they generate correctly from the "journals" metadata for BES).

rmzelle:
Just realized you did this in a repository branch I can commit to myself (duh), so easy to fix myself. Thanks!

PR #3099, opened 2017-11-06
Update university-of-gothenburg-apa-swedish-legislations.csl

bjornekstrom:
DOI link fixed. Removed month and day from citations.

### DIFF
--- a/university-of-gothenburg-apa-swedish-legislations.csl
+++ b/university-of-gothenburg-apa-swedish-legislations.csl
@@ -129,7 +129,7 @@
       <if type="thesis report" match="any">
         <choose>
           <if variable="DOI" match="any">
-            <text variable="DOI" prefix="https://dx.doi.org/"/>
+            <text variable="DOI" prefix="https://doi.org/"/>
           </if>
           <else-if variable="archive" match="any">
             <group>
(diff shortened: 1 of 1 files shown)


university-of-gothenburg-apa-swedish-legislations.csl
```diff
@@ -129,7 +129,7 @@
       <if type="thesis report" match="any">
         <choose>
           <if variable="DOI" match="any">
-            <text variable="DOI" prefix="doi: "/>
+            <text variable="DOI" prefix="https://dx.doi.org/"/>
```
rmzelle: @bjornekstrom, do you know if https://doi.org/ would also be acceptable? "https://dx.doi.org/" as a prefix is actually already outdated, and the shorter https://doi.org/ is now the officially recommended format. See https://www.zotero.org/support/kb/doi_in_apa for details.

university-of-gothenburg-apa-swedish-legislations.csl
```diff
@@ -151,7 +151,7 @@
       <else>
         <choose>
           <if variable="DOI">
-            <text variable="DOI" prefix="doi: "/>
+            <text variable="DOI" prefix="https://dx.doi.org/"/>
```
rmzelle: Idem.

university-of-gothenburg-apa-swedish-legislations.csl
bjornekstrom: @rmzelle Yes, this works better. Thanks for your reply. I have made the necessary amendments now.

rmzelle:
Thanks!

PR #3101, opened 2017-11-07
Update Springer Note

bwiernik:
The requester heard back from the publisher. Springer says the only official style documentation is their Key Points file. I added a link to this file (also used in other Springer styles). Should I remove the other documentation link or retain it to describe what is actually published for legal/note publications?

Also fixed formatting for legal cases and removed citation collapsing (which doesn't make sense given that this is a note style).


rmzelle:
Let's keep both documentation links. Thanks!

PR #3104, opened 2017-11-07
Create current-alzheimers-research.csl

POBrien333:
https://forums.zotero.org/discussion/68516/style-request-current-alzheimer-research


rmzelle:
Typo in style ID, but otherwise good. Thanks!

PR #3106, opened 2017-11-08
Update ib-tauris-note.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/291033/#Comment_291033


rmzelle:
Thanks!

PR #3111, opened 2017-11-13
Create Gallia-Prehistoire.csl

tomax31:
Hi
This style is the one linked with the journal Gallia Préhistoire (http://www.mae.u-paris10.fr/gallia-prehistoire/) from which I'm the editor in chief and which is one of the two main archaeological french journals. There wasn't yet any bibliographic style for it, so I hope this one will worked and will be accepted.
Cordially
Thomas Perrin

### DIFF
--- a/gallia-prehistoire.csl
+++ b/gallia-prehistoire.csl
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" delimiter-precedes-et-al="never" et-al-use-first="1" initialize-with="." page-range-format="expanded" demote-non-dropping-particle="sort-only" default-locale="fr-FR">
+<style class="in-text" version="1.0" delimiter-precedes-et-al="never" et-al-use-first="1" initialize-with="." page-range-format="expanded" demote-non-dropping-particle="sort-only" default-locale="fr-FR" xmlns="http://purl.org/net/xbiblio/csl">
   <info>
     <title>Gallia Préhistoire (French)</title>
     <title-short>GP</title-short>
     <id>http://www.zotero.org/styles/gallia-prehistoire</id>
     <link href="http://www.zotero.org/styles/gallia-prehistoire" rel="self"/>
-    <link href="http://www.persee.fr/collection/galip" rel="documentation"/>
+    <link href="http://www.persee.fr/collection/galip" rel="independent-parent"/>
     <link href="http://www.mae.u-paris10.fr/gallia-prehistoire/" rel="documentation"/>
     <author>
       <name>Thomas Perrin</name>
@@ -18,7 +18,8 @@
     <issn>0016-4127</issn>
     <eissn>2109-9642</eissn>
     <summary>Author-date style for Gallia Préhistoire - Préhistoire de la France dans son contexte européen</summary>
-    <updated>2017-11-14T12:51:08+00:00</updated>
+    <published>2017-11-14T13:33:50</published>
+    <updated>2017-11-15T07:14:36+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="fr">
@@ -42,13 +43,13 @@
     <choose>
       <if match="any" variable="author">
         <names variable="author" font-variant="normal" delimiter=", ">
-          <name form="short" font-variant="normal" font-weight="normal" and="text" delimiter-precedes-last="never" et-al-min="3" et-al-subsequent-min="0" initialize-with="." name-as-sort-order="all" sort-separator=" "/>
+          <name form="short" font-variant="normal" font-weight="normal" and="text" delimiter-precedes-last="never" et-al-min="3" et-al-use-first="1" et-al-subsequent-min="3" et-al-subsequent-use-first="1" initialize-with="." name-as-sort-order="all" sort-separator=" "/>
           <et-al font-style="italic"/>
         </names>
       </if>
       <else-if match="any" variable="editor">
         <names variable="editor">
-          <name form="short" and="text" delimiter-precedes-last="never" et-al-min="3" name-as-sort-order="all" sort-separator=" "/>
+          <name form="short" and="text" delimiter-precedes-last="never" et-al-min="3" et-al-use-first="1" et-al-subsequent-min="3" et-al-subsequent-use-first="1" name-as-sort-order="all" sort-separator=" "/>
           <et-al font-style="italic"/>
         </names>
         <text term="editor" form="short" prefix=" "/>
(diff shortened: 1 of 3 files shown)


POBrien333:
You'll also need to change the file name to match what you got under is/self link.

tomax31:
I didn’t saw the name over the window… hoping it will work now! Thank you for that.
I don’t understand what it wants about dependant/independant styles: as what I understood, all the « children » elements are there…

tomax31:
In fact no, it didn’t work. I really don’t understand. The names are the sames: 
 <info>
    <title>Gallia Préhistoire (French)</title>
    <title-short>GP</title-short>
    <id>http://www.zotero.org/styles/Gallia-Prehistoire</id>
    <link href="http://www.zotero.org/styles/Gallia-Prehistoire" rel="self"/>
As I saw on other styles, the title could be different, see for example : https://github.com/citation-style-language/styles/blob/master/aix-marseille-universite-departement-d-etudes-asiatiques.csl

POBrien333:
Have a look at my PR here and compare.

https://github.com/citation-style-language/styles/pull/3066

rmzelle:
@tomax31, don't worry about the failing tests for now. I'll take a look at this in the next few days and fix what I can correct myself.

tomax31:
OK thank you very much, that’s very kind. Wait just for the next version, because I’m actually making some changes and will send a new better version this night.

rmzelle:
> Wait just for the next version, because I’m actually making some changes and will send a new better version this night.

Okay. Please update the current pull request instead of a new one.

gallia-prehistoire.csl
```diff
@@ -0,0 +1,347 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" delimiter-precedes-et-al="never" et-al-use-first="1" initialize-with="." page-range-format="expanded" demote-non-dropping-particle="sort-only" default-locale="fr-FR">
+  <info>
+    <title>Gallia Préhistoire (French)</title>
+    <title-short>GP</title-short>
+    <id>http://www.zotero.org/styles/gallia-prehistoire</id>
+    <link href="http://www.zotero.org/styles/gallia-prehistoire" rel="self"/>
+    <link href="http://www.persee.fr/collection/galip" rel="documentation"/>
```
rmzelle: @tomax31, did you base this on an existing CSL style? If so, we'd like to have a template link (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link).

gallia-prehistoire.csl
```diff
@@ -0,0 +1,347 @@
...
+      <term name="accessed">consulté le</term>
+      <term name="cited">op.&amp;#160;cit.</term>
+      <term name="page">p.</term>
+    </terms>
+  </locale>
+  <macro name="author">
+    <choose>
+      <if match="any" variable="author">
+        <names variable="author" font-variant="normal" delimiter=", ">
+          <name form="short" font-variant="normal" font-weight="normal" and="text" delimiter-precedes-last="never" et-al-min="3" et-al-subsequent-min="0" initialize-with="." name-as-sort-order="all" sort-separator=" "/>
```
rmzelle: The "et-al" settings here are incorrect. "et-al-min" and "et-al-use-first" should always be defined together. "et-al-min" determines the minimum number of authors before "et al."-abbreviation kicks in, and "et-al-use-first" defines the number of authors to print before "et al.". "et-al-subsequent-min" and "et-al-subsequent-use-first" do the same for any additional time that a given item is cited.

In this case you have "et-al-min" and "et-al-subsequent-min", which is not a correct combo. The value of the "et-al" attributes should also always be higher than 0.

gallia-prehistoire.csl
```diff
@@ -0,0 +1,347 @@
...
+          <et-al font-style="italic"/>
+        </names>
+      </if>
+      <else-if match="any" variable="editor">
+        <names variable="editor">
+          <name form="short" and="text" delimiter-precedes-last="never" et-al-min="3" name-as-sort-order="all" sort-separator=" "/>
```
rmzelle: Missing "et-al-use-first" here.

gallia-prehistoire.csl
```diff
@@ -0,0 +1,347 @@
...
+    <names variable="container-author" delimiter=", ">
+      <name font-variant="normal" initialize-with="." name-as-sort-order="all" sort-separator=" "/>
+      <et-al font-style="italic"/>
+    </names>
+  </macro>
+  <macro name="author-bib">
+    <choose>
+      <if match="any" variable="author">
+        <names variable="author">
+          <name font-variant="normal" font-weight="bold" et-al-min="10" name-as-sort-order="all" sort-separator=" ">
```
rmzelle: Idem.

gallia-prehistoire.csl
```diff
@@ -0,0 +1,347 @@
...
+    <choose>
+      <if match="any" variable="author">
+        <names variable="author" delimiter=", ">
+          <name form="short" and="text" delimiter-precedes-last="never" et-al-min="3" et-al-use-first="1" initialize-with="." sort-separator=" "/>
+          <et-al font-style="italic"/>
+        </names>
+      </if>
+      <else-if match="any" variable="editor">
+        <names variable="editor" delimiter=", ">
+          <name and="text" delimiter-precedes-last="never" et-al-min="3" initialize-with="." sort-separator=" "/>
```
rmzelle: Idem.

gallia-prehistoire.csl
```diff
@@ -0,0 +1,347 @@
...
+          <et-al/>
+        </names>
+        <text term="editor" prefix=" "/>
+      </else-if>
+    </choose>
+    <text term="cited" font-style="italic" prefix=", "/>
+    <text variable="locator" prefix="p.&amp;#160;"/>
+  </macro>
+  <citation and="text" et-al-min="3" et-al-use-first="1" sort-separator=" ; ">
```
rmzelle: Note also that the "et-al" attributes on `<citation>` (and `<bibliography`>) trickle down to the `<names>` elements. The latter override the former.

tomax31:
No and yes… I took a look on 3 or 4 different styles but I did it completely with the CSL editor.

rmzelle:
> If you have an idea about how resolve it!-)

Well, there were still some errors in your style, which I just fixed, so you should have better luck now. Two quick comments:

* please always start from the latest repository version whenever you're making changes (you accidentally reverted some fixes I made, so I had to reapply those)
* I removed all the "et-al-subsequent-..." attributes. They're only necessary if they have values that are different from the regular "et-al-..." attributes.

rmzelle:
And I went ahead and merged this. If you wish to submit any more changes, please see our instructions at https://github.com/citation-style-language/styles/blob/master/CONTRIBUTING.md#3b-submitting-changes-to-an-existing-style.

tomax31:
Wonderful !! That’s great, and it’s working: thank you very very much. I tried it in my Zotero, and saw that I still have some little changes to do. I will follow your instructions.

PR #1228, opened 2014-11-07
Create polish_for_humanities.csl

milekpl:
This is a standard Polish style in humanities. It's available online but not in via repositiory, which makes it difficult to use:

http://www.doktoranci.uj.edu.pl/start/-/blogs/polski-styl-cytowan-w-programie-mendeley?_33_redirect=http%3A%2F%2Fwww.doktoranci.uj.edu.pl%2Fstart%3Fp_p_id%3D33%26p_p_lifecycle%3D0%26p_p_state%3Dnormal%26p_p_mode%3Dview%26p_p_col_id%3Dcolumn-2%26p_p_col_count%3D1%26_33_struts_action%3D%252Fblogs%252Fview

PR #3112, opened 2017-11-13
Create ISO-690 Numeric Polish

kgacaELS:


adam3smith:
Here is the original discussion, I believe: https://github.com/citation-style-language/styles/pull/1228
The concerns are still the same -- naming a style "Polish Humanities" seems fraught (if Polish academics are anything like other academics, there's bound to be dispute around that) and we really would like a link to where the style is described, even if that's not available online. Particularly since you are not the author of the style, that would be important: how else can we decide on corrections? 
Similarly for ISO 690 -- the standard leaves a lot of things open for interpretation, so it'd be ideal to have  a link that describes the Polish interpretation the style follows. 
Is there any chance of finding out that information?

PR #3113, opened 2017-11-13
Create Polish Humanities Style

kgacaELS:


adam3smith:
Thank you. Before we fix the test failures here, could you tell us a little more about the style? In particular, what style guide is it based on? We typically require for styles to link to a style guide either online or in book form, so that other people can check on possible errors in the style.

In addition, is there any reason you're setting all the polish terms? We do have Polish translations available https://github.com/citation-style-language/locales/blob/master/locales-pl-PL.xml that look like they're in very good shape. Finally, did you start from scratch or did you base this on any existing style?

kgacaELS:
Hi,

Thank you for reaching out to me.

The reason why I’m trying to get these two styles into the CSL repository is that we get A LOT of comments from Polish community using Mendeley (reference manager which uses styles from the CSL repository) that there isn’t ISO-690 in Polish, and there isn’t a style used by the Polish arts and humanities’ journals.

The styles I’ve tried to upload were created by a former PhD student, who granted me permission to host them in the library (he apparently failed to do so). I’m not sure whether he used a template or did them from scratch. What I do know, though, is that if I upload them locally to Mendeley Desktop, they work fine.

Can you help me further with this, or would it be wiser to find a programmer to clean up the code?

All the best,
Kasia

adam3smith:
Here is the original discussion, I believe: #1228
The concerns are still the same -- naming a style "Polish Humanities" seems fraught (if Polish academics are anything like other academics, there's bound to be dispute around that) and we really would like a link to where the style is described, even if that's not available online. Particularly since you are not the author of the style, that would be important: how else can we decide on corrections?
Similarly for ISO 690 -- the standard leaves a lot of things open for interpretation, so it'd be ideal to have a link that describes the Polish interpretation the style follows.
Is there any chance of finding out that information?

PR #3115, opened 2017-11-14
Create universitetet-i-oslo-rettsvitenskap.csl

hgsl:
This style was created based on the University of Oslo Law Library's recommendations for legal students and professionals: https://www.ub.uio.no/skrive-publisere/referere/fag/rettsvitenskap.html

The style's language is in Norwegian Bokmål, its primary audience being everyone working with Norwegian legal sources.


hgsl:
One small note: csl-bot initially stopped me because I had added the following to <citation>: et-al-min="1" et-al-use-first="1"

The alleged error consisted in et-al-min not having at least one higher value than et-al-use-first.

However, this was actually done intentionally, since our style dictates that for footnote citations only the surname of the first author should be displayed at all times (with no "et al."-specification). The solution above was used because I couldn't see any other way of specifying this in the style. Would be happy to hear that I am wrong though!

adam3smith:
I don't think `et-al-min` 1/1 would actually give you that -- certainly not reliably so. My recollection is that it actually doesn't behave differently from 2/1 in most implementations.

I haven't tested, but I think the way to go for just one name for in-text citations to do something like
`<term name="and others"> </term>` in the locale/terms section and then use 

```
<names>
<name/>
<et-al term="and others"/>
</names>
```
(and use `et-al-min="2"`)

hgsl:
Thanks for the reply!
I tried the solution you're describing earlier, and made this macro:

  <macro name="author-short">
    <names variable="author">
      <name form="short"/>
      <et-al term="and others"/>
      <substitute>
        <names variable="editor"/>
      </substitute>
    </names>
  </macro>

Combined with the et-al-min="2" this yields the following output, e.g.: 
Engelschiøn med flere (2017)
Med flere = and others in Norwegian

The trouble is that I only want this output, without the "and others":
Engelschiøn (2017)

et-al-min="1" gives me exactly this, but you may be right that it isn't reliable - although so far in my testing with various reference types it looks okay.

Anyway, it's a minor thing - we could probably alter the style a little bit to accommodate the et al.

And thanks for your help so far! I'm fairly new at this so I may certainly be overlooking something. :)

hgsl:
Thanks a lot! That works perfectly. I experimented with this solution yesterday, but a small syntax error had snuck in. Now I got it to work.

Your help is greatly appreciated. :)

rmzelle:
Since this style will be set to Norwegian, wouldn't it make more sense to change the title to Norwegian as well? E.g. `<title>Universitetet i Oslo - Rettsvitenskap (Norwegian - Bokmål)</title>` instead of `<title>University of Oslo - Law Library (Footnote with bibliography) (Norwegian Bokmål)</title>`? (in which case we'd change the style name to "universitetet-i-oslo-rettsvitenskap.csl")

(we also usually don't mention things like "(Footnote with bibliography)" unless we have multiple variants with an otherwise identical title)

rmzelle:
And if you based this on an existing CSL style, we'd like to have a "template" link in the style (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link).

hgsl:
That absolutely makes sense, will make sure to change it. Thanks for the feedback!

I didn't base it on any existing csl-style, wrote everything from scratch - although I studied other styles in conjunction with the documentation in order to get the hang of the syntax, ideas for macros, etc.

universitetet-i-oslo-rettsvitenskap.csl
```diff
@@ -0,0 +1,295 @@
...
+            <text prefix=", " suffix=": " variable="publisher-place"/>
+            <text variable="publisher"/>
+            <text prefix=", " macro="issued-no-parenthesis"/>
+            <text prefix=", " variable="URL"/>
+            <text prefix=" " macro="accessed-date"/>
+          </group>
+        </else-if>
+        <else-if type="legislation bill" match="any">
+          <group suffix=".">
+            <text suffix="            " macro="issued-no-parenthesis"/>
```
rmzelle: @hgsl, what are you trying to achieve here with the `suffix="            "`? I'm wondering if there isn't a more elegant way.

hgsl: This is actually done intentionally and is in line with our style. 

The purpose is, for certain reference types such as legislation, court cases, and certain other types of legal sources, to have them in separate registers apart from the literature. In these registers the bibliographic information is separated into two seperate columns, for ease of reference. This is also why I have incorporated the type sorting, so that it is easier for the user to separate the registers from the main bibliography.

You can see examples of this in the law library's specifications: https://www.ub.uio.no/skrive-publisere/referere/fag/rettsvitenskap.html

E.g. Norske rettskilder > Lover og forskrifter

adam3smith: That won't work well with spaces, though, a) because they'll align unevenly where dates aren't 4 characters (e.g. because of disambiguation) and b) because newlines will start at the front of the line. You should instead use `display` properties as explained here: http://docs.citationstyles.org/en/stable/specification.html#display

hgsl: Yep, you're definitely right about that. Thanks for the tip, I'll work on it some more when I have the time. Again, I really appreciate the feedback!

hgsl:
I have updated the style using the display properties for certain types of references, e.g. legislation. Ideally I would have liked some more space (preferably tab-sized) in between the left-aligned and right-aligned group elements for these reference types. Not sure if it's possible to do this elegantly though, and in any case it's a minor concern. Any further comments are welcome.

universitetet-i-oslo-rettsvitenskap.csl
```diff
@@ -0,0 +1,299 @@
...
+      <term name="editortranslator" form="verb-short">
+        <single>red. og overs.</single>
+        <multiple>red. og overs.</multiple>
+      </term>
+      <term name="editortranslator" form="verb">
+        <single>Redigert og oversatt av</single>
+        <multiple>Redigert og oversatt av</multiple>
+      </term>
+      <term name="translator" form="short">overs.</term>
+      <term name="and others"> </term>
```
rmzelle: @hgsl, did you test whether you get the correct output if you use an empty term here? (i.e. `<term name="and others"></term>` instead of `<term name="and others"> </term>`)

rmzelle: @adam3smith, other that this I have no comments.

hgsl: It appears to yield the same output, so I'll use the empty term instead - thanks!

rmzelle:
Thanks! Sorry for letting this sit for so long!

PR #3116, opened 2017-11-14
Propagate from 5e36aa1 and 210399d (see #2266)

tantignac:
ieee.csl has been updated but the changes have not been propagated to its derivative ieee-with-url.csl. This is a fix.

### DIFF
--- a/ieee-with-url.csl
+++ b/ieee-with-url.csl
@@ -28,7 +28,7 @@
     <category citation-format="numeric"/>
     <category field="engineering"/>
     <category field="generic-base"/>
-    <updated>2012-10-18T22:38:43+00:00</updated>
+    <updated>2017-11-14T12:30:00+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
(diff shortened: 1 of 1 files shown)


adam3smith:
Great, thanks. So currently IEEE and this port only use et al for authors/editors at the beginning of the bibliography entry, but not for editors of book chapters. Do you happen to know what the rules are on this? If -- as I suspect -- we want the same rules applied, we should move et al to bibliography rather than setting it on names.

tantignac:
My understanding is it should not be propagated to editors as well. Indeed, in https://github.com/citation-style-language/styles/files/514202/IEEE_style_manual.pdf (page 34, section A) the rule to use `et el.` is indicated in the paragraph explaining how papers should be cited in the text (not in the bibliography). Just after, in Section B, the `et al.` rule is also applied to author names only. This does not seem to be a mistake since another rule is applied to author *and* editor names just before. 

This seems quite sensible as limiting the length of the author list makes sense not to interrupt the reading flow (and editors do not appear in citations). However, in the bibliography, nothing prevents from having more names appearing.

adam3smith:
I'm not sure it's quite as clear, given that "If there are more than six names listed, use et al. after the
first author." applies to the reference list. But having more than six editors is also rather unlikely, so we might as well take it like this.

adam3smith:
Thanks!

PR #3118, opened 2017-11-16
Update scandinavian-journal-of-medicine-and-science-in-sports.csl

libora:
change sorting of bibliography - https://forums.zotero.org/discussion/comment/292220/#Comment_292220


adam3smith:
Thanks!

rmzelle:
(@libora, in general it's better to just delete unused sections of styles instead of commenting them out; we can always look at older versions using the Git history, so there is no need to keep unused code around)

PR #3119, opened 2017-11-16
Update fish-and-fisheries.csl

karendejong:
Thanks for creating this, I am using jabref and am very happy this exists. However, apparently the style of Fish and Fisheries changed and now it is close to, but not completely apa 6th (so please ignore the description in the file, I saved it before I checked the apa 6th guidelines). I thought it would be easiest to use the apa 6th and adapt that, and then call it fish-and-fisheries.csl, will that work in github?

https://github.com/citation-style-language/styles/compare/master...karendejong:patch-1?expand=1#diff-16a30fe6e9699e4af3617aef97ee54d7

reference for Fish and Fisheries style:
http://onlinelibrary.wiley.com/journal/10.1111/(ISSN)1467-2979/homepage/ForAuthors.html

Journal article
Beers, S. R. , & De Bellis, M. D. (2002). Neuropsychological function in children with maltreatment-related posttraumatic stress disorder. The American Journal of Psychiatry, 159, 483–486. doi:10.1176/appi.ajp.159.3.483

Book
Bradley-Johnson, S. (1994). Psychoeducational assessment of students who are visually impaired or blind: Infancy through high school (2nd ed.). Austin, TX: Pro-ed.

Chapter in an Edited Book
Borstrøm, I., & Elbro, C. (1997). Prevention of dyslexia in kindergarten: Effects of phoneme awareness training with children of dyslexic parents. In C. Hulme & M. Snowling (Eds.), Dyslexia: Biology, cognition and intervention (pp. 235–253). London: Whurr.

Internet Document
Norton, R. (2006, November 4). How to train a cat to operate a light switch [Video file]. Retrieved from http://www.youtube.com/watch?v=Vja83KLQXZs


rmzelle:
> (from the commit message) Thanks for creating this, I am using jabref and am very happy this exists. Just a small update: As far as I have seen, in apa 6th, volume should not be italic and issue number should only be mentioned when page numbers start at 0 with every issue, which only very few journals have. So it would be less work to have to add it in those cases, than having to remove it for all.

Thanks! For the repository styles, we always err on the side of including the issue number, so it looks like this style should just become a dependent of the regular APA 6th edition style. I'm also pretty confident that we can use the latest DOI format ("https://doi.org/" prefix). @adam3smith, thoughts? (see http://onlinelibrary.wiley.com/doi/10.1111/faf.12214/epdf for a recent free article)

@karendejong, how good is JabRef support for CSL styles nowadays, by the way? Initially it seemed to only use CSL for in-app reference previews. Can you use CSL for actual writing now too?

adam3smith:
Dependent seems right to me, yes. 
Not sure on the DOI format (as DOIs aren't listed in the reference list -- that's just an example citation at the end in a completely different style, oddly enough), probably OK, but who knows?

I'll note that the article you link to doesn't do firstname disambiguation for primary authors, even though it should according to APA (for Taylor).

rmzelle:
Thanks! Sorry for the long wait.

PR #3120, opened 2017-11-17
Mla8 accessed date

lightman76:
MLA 8 (page 53) says:

> Since online works typically can be changed or removed at any time, the date on which you accessed online material is often an important indicator of the version you consulted.

I've added an else-if clause so that accessed date is added if a URL or DOI is present.  At present, it only adds accessed date if no published date is present.  I was only going to do this on URL, but it appears that the DOI standard recommends a new DOI if significant changes are made, but does not require it.  Can pull out DOI if you prefer.


adam3smith:
I don't think I agree with this. The passage you cite is in a section titled "optional element" and most MLA citations including URLs in the handbook do not have an access date.

lightman76:
Ok - I'm working with some teachers who wanted it included.  I'll keep this in my personal branch then.

PR #3121, opened 2017-11-17
Mla8 support for interviewer

lightman76:
The MLA 8 format did not include interviewer in the other-contributors section.  Added interviewer to the other contributors macro and with this change it appears to now match Purdue OWL's MLA8 guide for published interviews: https://owl.english.purdue.edu/owl/resource/747/09/


adam3smith:
Yes, I think this works well. Thanks!

PR #3122, opened 2017-11-17
Mla8 2nd container not including source without a URL or DOI

lightman76:
In MLA 8, found that the source variable was not being displayed (in the container2-location macro) unless a URL or DOI was also present.  Believe that it should always be shown.

### DIFF
--- a/modern-language-association.csl
+++ b/modern-language-association.csl
@@ -185,7 +185,15 @@
       </group>
       <choose>
         <if variable="source" match="none">
+<<<<<<< HEAD:modern-language-association.csl
           <text macro="URI"/>
+=======
+          <choose>
+            <if variable="URL DOI" match="any">
+              <text macro="URI"/>
+            </if>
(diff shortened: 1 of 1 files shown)


adam3smith:
If something is coded that deliberately, probably better to start an issue for discussion first. This is on purpose. Sources for items without URLs are typically not the actual sources of an item but merely the catalog, i.e. the library catalog, google scholar, or CrossRef and so shouldn't be used here. 

IIRC the style has code in place to use archive fields to include the source of non-digital items.

adam3smith:
(something weird going on with your fork, btw. You may want to rebase or otherwise make sure you start from a clean commit)

lightman76:
Ok - will start an issue for that.  Will also try to get my branch cleaned up and update the PRs.

lightman76:
Moved discussion to #3123 .   Closing for now as seems current behavior is ok.

PR #3124, opened 2017-11-20
Update development-and-change.csl

POBrien333:
via https://forums.zotero.org/discussion/68839/development-and-change-style
Not sure why LiborA not just spent another 5min fixing the style when he already started doing it.


rmzelle:
> Not sure why LiborA not just spent another 5min fixing the style when he already started doing it.

You just established yourself as the style editing expert. Now everybody starts relying on you! Sorry! :)

rmzelle:
Thanks!

PR #3125, opened 2017-11-21
Update revue-archeologique.csl

NMonteix:
Final dot in citation removed
Double comma in ibid-with-locator removed

### DIFF
--- a/revue-archeologique.csl
+++ b/revue-archeologique.csl
@@ -150,7 +150,7 @@
     </choose>
   </macro>
   <citation et-al-min="4" et-al-use-first="1" disambiguate-add-year-suffix="true" collapse="year">
-    <layout delimiter="&#160;; ">
+    <layout suffix="." delimiter="&#160;; ">
       <choose>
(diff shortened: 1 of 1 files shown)


adam3smith:
we do generally set periods at the end of citations unless you're saying the period wouldn't be required in a single footnote.

NMonteix:
Some authors told me it was problematic when they wrote after the citation.

adam3smith:
Sorry, we're not going to introduce that level of inconsistency between citation styles. This works fine when using prefix/suffix, which is what's recommended to add text after the citation.
Without the period, a simple task like converting an in-text to a footnote citation doesn't work properly anymore (and it breaks compatibility with other foonote styles that include the period).

NMonteix:
@adam3smith 
Period reinstated

adam3smith:
Great, thanks!

NMonteix:
To you!

Le 21/11/2017 à 18:40, Sebastian Karcher a écrit :

PR #3126, opened 2017-11-21
Update and rename dependent/genome-biology.csl to genome-biology.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/284905/#Comment_284905


rmzelle:
@POBrien333, thanks!! (I guess it's pretty safe to assume that we don't need to modify the original independent-parent style instead, and that these tweaks are specific to this journal, right?)

POBrien333:
Actually.... This journal was BMC, then Springer and is back to BMC. It looks like that style. Just checked 2 random BMC journals.

rmzelle:
So it can remain a dependent then?

genome-biology.csl
```diff
@@ -0,0 +1,259 @@
...
+    <layout>
+      <text variable="citation-number" suffix=". "/>
+      <group delimiter=". " suffix=". ">
+        <text macro="author"/>
+        <text macro="title"/>
+      </group>
+      <group delimiter=" " suffix=". ">
+        <text macro="editor"/>
+        <text macro="container-title"/>
+        <text macro="publisher"/>
```
rmzelle: (the person who asked about this reported that this version prints the publisher for journal articles (in Paperpile), which doesn't seem to be correct)

POBrien333:
Wasn't there a thing with BMC that it all became Springer or something?
Anyway, adapted the style and fixed a bunch of things. Thanks for making me look over it again.

rmzelle:
> Wasn't there a thing with BMC that it all became Springer or something?

I'm not sure what it was, but it was something :).

> Anyway, adapted the style and fixed a bunch of things. Thanks for making me look over it again.

You're welcome! :)

rmzelle:
Actually, sorry to keep asking this, but are you sure this shouldn't remain a dependent? https://genomebiology.biomedcentral.com/submission-guidelines/preparing-your-manuscript/research says "Examples of the BioMed Central reference style are shown below.", which suggests the journal uses a general BMC style.

POBrien333:
yeah. I first adapted the style and then remembered your question.... haha
I asked the OP in the forum thread to double check if this is not biomed-central.csl.
Waiting to hear back and then switch it back to dependent on the outcome.

POBrien333:
Finally took the time to check this. Indeed, we can make this a dependent of biomed-central.
Guidelines are here: https://genomebiology.biomedcentral.com/submission-guidelines/preparing-your-manuscript/research

Can you get that sorted? it's an automatically-generated style. Thanks :)

Can't figure out how

rmzelle:
@POBrien333, do you mind giving this another look?

https://genomebiology.biomedcentral.com/submission-guidelines/preparing-your-manuscript/research now says:

> References
> Examples of the Vancouver reference style are shown below.

so maybe the citation format changed since we last reviewed this?

POBrien333:
@rmzelle 
Sure. I'll look into this in the coming days.

rmzelle:
> They want Vancouver style, yes, but in-text with brackets.

And that's different from https://www.zotero.org/styles?q=id%3Aspringer-vancouver-brackets ?

POBrien333:
Yes, already has differences for in-text.
Genome Biology wants: [1, 5]
Springer-vancouver-brackets: [1,5]

PR #3130, opened 2017-11-23
Removing sorting in citation

bwiernik:
https://forums.zotero.org/discussion/comment/293361/#Comment_293361


adam3smith:
Thanks!

PR #3132, opened 2017-11-24
Create the-journal-of-veterinary-medical-science.csl

shojitakahashi:
The Journal of Veterinary Medical Science
Publisher: Japanese Society of Veterinary Science
Journal homepage: http://jsvetsci.jp/jvms/
Instructions for Authors: http://jsvetsci.jp/jvms/instructions-for-authors/


rmzelle:
(replaced by #3133)

PR #3134, opened 2017-11-24
Fix misordering of author and date

bwiernik:
https://forums.zotero.org/discussion/68999/year-and-author-reversed-for-in-text-citation


adam3smith:
Thanks!

PR #3135, opened 2017-11-24
Create wirtschaftsuniversitat-wien-institute-for-managing-sutainabili…

georgd:
…ty.csl

Depending on another WU Vienna style.

### DIFF
--- a/dependent/wirtschaftsuniversitat-wien-institute-for-managing-sustainability.csl
+++ b/dependent/wirtschaftsuniversitat-wien-institute-for-managing-sustainability.csl
@@ -14,7 +14,7 @@
     </author>
     <category citation-format="author-date"/>
     <category field="social_science"/>
-    <summary>Citation style for WU Vienna’s Intitute for Managing Sustainability. This is an alias for the style Wirtschaftsuniveristät Wien - Master Management.</summary>
+    <summary>Citation style for WU Vienna’s Institute for Managing Sustainability. This is an alias for the style Wirtschaftsuniversität Wien - Master Management.</summary>
     <updated>2017-11-24T16:34:19+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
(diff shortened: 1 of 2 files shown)


adam3smith:
Before we address the tests -- since those are both styles at the same university, couldn't you just point students to the other style? The main purpose of dependent styles is to help users discover them -- is that really needed here? I'm a bit concerned about adding dependent styles for different faculties of the same university, but maybe there's a compelling reason here. Or might it make sense to rename the original style into something more generic? (feel free to respond in German)

georgd:
Please believe me, I'd prefer a generic solution. I'm maintaining this style for four different systems and in two languages (which for EndNote and Citavi means extra style files) and I'd be happy if it were a bit easier. But the current situation is due to a recent agreement with the university leadership about how to proceed with institutional styles. You may view it as a political problem. In the end it was decided that it should be unambiguously clear from the style name for whom a style might apply.

Also, at the moment I can't even tell if the requirements by the two institutes won't diverge in the near future. One of them has mostly German theses, the other one mostly English ones. And as you surely know, German and Austrian university institutes in social sciences and humanities tend rather to create their individual solutions than to consolidate on common styles. Therefore I'm a bit pessimistic.

rmzelle:
> I'm maintaining this style for four different systems

So right now this would be the last one to add beyond the three styles we already have for your university? (https://www.zotero.org/styles?q=wirtschaftsuniversitat-wien)

(I agree with @adam3smith that it's not ideal to add dependents for a single university, but I also understand it's not your personal (or library's) call to make, so if it stays with 4 styles total for now that would be okay with me)

georgd:
There seems to be a misunderstanding. I express that I have to maintain this very style in four different citation style languages. It's a coincidence that it's also the fourth style of our university in this repo. 

I don't think that this would be the last style of our university that we'd like to see in this repo but to save us the trouble I won't push any more dependent institutional styles.

georgd:
@adam3smith @rmzelle, please don't merge this yet, I'll talk to my colleagues about this tomorrow.

georgd:
Just to make sure: if we distribute dependent styles locally, will they automatically fetch the generic style from the Zotero style repository? If so, we will go this way and rename the Master Management style accordingly and locally distribute whatever dependent style seems necessary.

rmzelle:
> if we distribute dependent styles locally, will they automatically fetch the generic style from the Zotero style repository?

Yes. Just keep the "independent-parent" link of the dependent style intact, and things should work.

georgd:
Great, thank you! I'll close this request then and open a new one for the renaming.

rmzelle:
(renaming PR at #3155)

PR #3136, opened 2017-11-25
Update of gallia-prehistoire.csl

tomax31:
Here is the update of my gallia-prehistoire.csl style which corrects several important mistakes.

### DIFF
--- a/gallia-prehistoire.csl
+++ b/gallia-prehistoire.csl
@@ -244,12 +244,12 @@
       </if>
       <else-if type="chapter entry-dictionary entry-encyclopedia" match="any">
        <group delimiter=", ">
-       <text macro="publisher-place" font-style="italic"/>
+       <text variable="publisher-place" font-style="italic"/>
        </group>
       </else-if>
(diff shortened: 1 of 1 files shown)


gallia-prehistoire.csl
```diff
@@ -239,6 +242,16 @@
...
         <text macro="issue" prefix=", "/>
       </if>
+      <else-if type="chapter entry-dictionary entry-encyclopedia" match="any">
+       <group delimiter=", ">
+       <text macro="publisher-place" font-style="italic"/>
+       </group>
+      </else-if>
+      <else-if type="manuscript thesis report" match="any">
+       <group delimiter=", ">
+       <text macro="publisher-place" font-weight="normal"/>
```
rmzelle: @tomax31, here, and at line 247 above, you call a non-existing "publisher-place" macro, which is why our automatic testing is giving an error. (if you just want to print the "publisher-place" variable, you can use `<text variable="publisher-place" ... />` instead of `<text macro="publisher-place" ... />`)

rmzelle:
Thanks!

PR #3138, opened 2017-11-26
Create colorado-state-university-school-of-biomedical-engineering.csl

POBrien333:
via https://forums.zotero.org/discussion/68950/modified-chicago-citation-style


POBrien333:
@rmzelle 
Did I do the et-al settings and locale term settings correct for the et al output here (just one author)?
In the CSL editor this gives an error even though the style validates: "Citeproc exception: TypeError: Cannot read property 'length' of undefined"

adam3smith:
This looks right to me (and works in Zotero). We haven't updated the citeproc instance in the editor for a while, so I'd suspect a bug there.

adam3smith:
@rmzelle looks good to me. Leaving for you to merge in case you have naming/metadata concerns.

rmzelle:
Thanks!

PR #3140, opened 2017-11-26
Update international-journal-of-geriatric-psychiatry.csl

POBrien333:
Found a whole bunch of mistakes just by random. Should be a bit better now.

### DIFF
--- a/international-journal-of-geriatric-psychiatry.csl
+++ b/international-journal-of-geriatric-psychiatry.csl
@@ -2,6 +2,7 @@
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
   <info>
     <title>International Journal of Geriatric Psychiatry</title>
+    <title-short>JGP</title-short>
     <id>http://www.zotero.org/styles/international-journal-of-geriatric-psychiatry</id>
     <link href="http://www.zotero.org/styles/international-journal-of-geriatric-psychiatry" rel="self"/>
     <link href="http://www.zotero.org/styles/harvard-university-of-wolverhampton" rel="template"/>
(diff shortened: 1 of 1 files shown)


international-journal-of-geriatric-psychiatry.csl
```diff
@@ -45,6 +45,7 @@
   <macro name="author-short">
     <names variable="author">
       <name form="short" and="text" delimiter=", " delimiter-precedes-last="never" initialize-with=". "/>
+      <et-al/>
```
adam3smith: why are you adding this? This shouldn't change anything.

international-journal-of-geriatric-psychiatry.csl
```diff
@@ -20,7 +19,7 @@
     <issn>0885-6230</issn>
     <eissn>1099-1166</eissn>
     <summary>Harvard (author-date style) for International Journal of Geriatric Psychiatry</summary>
```
adam3smith: we should add a note that this follows published articles and not the (useless) author guidelines

international-journal-of-geriatric-psychiatry.csl
```diff
@@ -160,7 +159,7 @@
       </group>
     </layout>
   </citation>
-  <bibliography>
+  <bibliography delimiter-precedes-et-al="always" et-al-min="4" et-al-use-first="3">
```
adam3smith: I think this is 6/3 or 7/3 I see at least 5 listed authors here: http://onlinelibrary.wiley.com/doi/10.1002/gps.4447/epdf I think 7/3 is the most common rule, so I'd go with that.

international-journal-of-geriatric-psychiatry.csl
```diff
@@ -146,7 +145,7 @@
       </if>
     </choose>
   </macro>
-  <citation et-al-min="4" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year">
+  <citation delimiter-precedes-et-al="never" et-al-min="4" et-al-use-first="1" disambiguate-add-names="true" disambiguate-add-givenname="true" disambiguate-add-year-suffix="true" collapse="year">
```
adam3smith: Looking at the Crellin et al. reference in http://onlinelibrary.wiley.com/doi/10.1002/gps.4447/epdf I think this should be 3/1

international-journal-of-geriatric-psychiatry.csl
POBrien333: Not sure where that is coming from. thanks for spotting it.

adam3smith:
Thanks!

PR #3144, opened 2017-11-27
Create aipi.csl

romansos:


rmzelle:
@romansos, would it make sense to name this style after the Civiltà Italiana series, or is this style used much wider within the AIPI? And does the Civiltà Italiana series have an ISSN?

romansos:
@rmzelle Yes, Civiltà Italiana it is the series of AIPI (Associazione Italiana dei Professori d'Italiano) published with Cesati editore. It is maybe more correct (formally) to call it Civiltà Italiana. AIPI does not publish other series, so also simply AIPI is unique.

rmzelle:
Okay, thanks. I merged this under the name "Civiltà Italiana", then.

PR #3148, opened 2017-11-29
Update the-historical-journal.csl

ntdmiller:
Style changed to more closely reflect the description of the style in The Historical Journal instructions for contributors (assets.cambridge.org/HIS/HIS_ifc.pdf).

### DIFF
--- a/the-historical-journal.csl
+++ b/the-historical-journal.csl
@@ -24,7 +24,7 @@
   <macro name="author">
     <names variable="author">
       <name and="text" delimiter=", "/>
-      <label form="short" prefix=", " suffix=""/>
+      <label form="short" prefix=", "/>
       <substitute>
         <names variable="editor"/>
         <names variable="translator"/>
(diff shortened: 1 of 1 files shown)


the-historical-journal.csl
```diff
@@ -77,7 +94,6 @@
   </macro>
   <macro name="publisher">
     <group delimiter=": ">
-      <text variable="publisher"/>
       <text variable="publisher-place"/>
```
rmzelle: @ntdmiller, was this an intentional change? It's a bit uncommon to print the "publisher-place" but not the "publisher" itself.

adam3smith: Yes, that's right:
M. Cowling, 1867: Disraeli, Gladstone and revolution: the passing of the
second Reform Bill (Cambridge, 1967), pp. 41-5, 140-7.  
and other examples.

rmzelle:
Okay, thanks!

PR #3153, opened 2017-11-29

cookie365:
Added own terms for language en
Two new date forms (one for de and one for en)
Changed edition macro from numeric to ordinal
Added text term page instead of prefix S.


fachhochschule-vorarlberg-author-date.csl
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="de-DE">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
   <info>
-    <title>Fachhochschule Vorarlberg (author-date, German)</title>
+    <title>Fachhochschule Vorarlberg (author-date, German, English)</title>
```
adam3smith: same question/comment about language as for the note style

cookie365: Same answer/comment, i have no objection about it.

adam3smith:
thanks!

PR #3154, opened 2017-11-29

cookie365:
Added own terms for language en
Two new date forms
Changed edition macro from numeric to ordinal
Added text term page instead of prefix S.


fachhochschule-vorarlberg-note.csl
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" default-locale="de-DE">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only">
   <info>
-    <title>Fachhochschule Vorarlberg (note, German)</title>
+    <title>Fachhochschule Vorarlberg (note, German, English)</title>
```
adam3smith: if you have no objection, we'd simply remove the language entirely (technically the style would also work in French, e.g., even if you don't have customized terms).

cookie365: That's fine by me.

adam3smith:
Thanks!

PR #3155, opened 2017-11-29
Update and rename wirtschaftsuniversitat-wien-master-management.csl t…

georgd:
…o wirtschaftsuniversitat-wien-apa-based.csl

### DIFF
--- a/wirtschaftsuniversitat-wien-apa-based.csl
+++ b/wirtschaftsuniversitat-wien-apa-based.csl
@@ -15,7 +15,7 @@
     <category citation-format="author-date"/>
     <category field="social_science"/>
     <summary>APA 6th ed. based citation style used at WU Vienna by the Institute for Managing Sustainability, the Master Management programme and the PMBA Sozialmanagement programme.</summary>
-    <updated>2017-12-17T21:53:19+00:00</updated>
+    <updated>2017-11-29T21:53:19+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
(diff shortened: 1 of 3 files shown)


wirtschaftsuniversitat-wien-apa-based.csl
```diff
@@ -2,20 +2,20 @@
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" name-delimiter="/" initialize="false" initialize-with="." demote-non-dropping-particle="never">
   <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
-    <title>Wirtschaftsuniversität Wien - Master Management</title>
-    <id>http://www.zotero.org/styles/wirtschaftsuniversitat-wien-master-management</id>
-    <link href="http://www.zotero.org/styles/wirtschaftsuniversitat-wien-master-management" rel="self"/>
+    <title>Wirtschaftsuniversität Wien - APA Based (English and German)</title>
```
rmzelle: @georgd, would `<title>Wirtschaftsuniversität Wien - APA</title>` be acceptable? We never include languages in the title for multi-lingual styles, and "Based" is a bit redundant.

georgd:
OK, we can drop the languages.

With the "Based" part I wanted to indicate that it's a quite customized APA. Authors formats don't look like APA at all. At WU we have other institutes that use APA variants that are by far closer to the original. Do you have a better idea?

georgd:
OK, we can drop the languages.

With the "Based" part I wanted to indicate that it's a quite customized APA. Authors formats don't look like APA at all. At WU we have other institutes that use APA variants that are by far closer to the original. Do you have a better idea?

rmzelle:
> With the "Based" part I wanted to indicate that it's a quite customized APA. Authors formats don't look like APA at all. At WU we have other institutes that use APA variants that are by far closer to the original. Do you have a better idea?

If "APA" is not a good descriptor it might be better to stick with the department naming.

rmzelle:
Thanks!

PR #3156, opened 2017-11-30
Create cell-structure-and-function.csl

nitchy2008:
This is a newly made citation style for the journal Cell Structure and Function.


rmzelle:
@nitchy2008, thanks! Did you base this on an existing CSL style? If so, we'd like to know which one, so we can add a "template" link (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link)

nitchy2008:
@rmzelle thanks for you efforts on managing this kind of valuable resource. I created the style based on following one : http://www.zotero.org/styles/evolutionary-ecology-research
I'd be happy if you'd add it as a template link. thanks.

rmzelle:
Thanks!

PR #3159, opened 2017-11-30
Fix errors in Springer-SocPsych de locale

bwiernik:
https://forums.zotero.org/discussion/69160/style-error-publizistik

See uses of "et al." and "zugegriffen:" here:
https://link.springer.com/content/pdf/10.1007%2Fs11616-016-0308-2.pdf
https://link.springer.com/article/10.1007/s11616-016-0256-x

I'm not sure if specifying the de accessed like this in `locale` will override the `text-case` argument when `term="accessed"` is called in the style (will the colon will prevent case changes?), but I don't think it's much of a concern if not.

### DIFF
--- a/springer-socpsych-author-date.csl
+++ b/springer-socpsych-author-date.csl
@@ -40,8 +40,8 @@
   </locale>
   <locale xml:lang="de">
     <terms>
-      <term name="et-al">et al</term>
-      <term name="accessed" form="long">Zugegriffen:</term>
+      <term name="et-al">et al.</term>
+      <term name="accessed" form="long">zugegriffen:</term>
     </terms>
   </locale>
   <macro name="container-contributors">


adam3smith:
OK on the et-al, but I'm not following the change on Zugegriffen. Best I can tell this shouldn't change anything at all?

bwiernik:
In the published German articles, they use "zugegriffen:" rather than "Zugegriffen:"

I don't think the style as changed here will fix that, but I figured that storing terms in lowercase is better style in any event. I don't think there would be any way to get English to print "Accessed" but German to print "zugegriffen:", correct?

adam3smith:
That doesn't seem to be consistent:
image
And given that it's preceded by a period, capitalizing seems more consistent to me. 

The way to have lowercase German/uppercase English would be to not use text-case and then define the term capitalized in the English locale, so it's possible, just not convinced it should be done.

bwiernik:
Okay, if it's inconsistent, I agree that it would be better capitalized. I don't see the harm in leaving the term as defined in the PR now. Do you want me to change that one back?

adam3smith:
yes, agree, let's leave the change in, thanks

PR #3160, opened 2017-11-30
Create environmentalchemistry-harvard.csl

fvmolina:
New style for the CSIRO Environmental Chemistry journal. This journal had a numeric style, for which I create a csl file some years ago, but have just changed to a Harvard-type style. This proposal covers the new style. Note that the existing environmentalchemistry.csl is being deprecated.
The style has been validated

### DIFF
--- a/environmentalchemistry-harvard.csl
+++ b/environmentalchemistry-harvard.csl
@@ -1,0 +1,221 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="expanded" default-locale="en-AU">
+  <info>
+    <title>Environmental Chemistry - Harvard</title>
+    <id>http://www.zotero.org/styles/environmentalchemistry-harvard</id>
+    <link href="http://www.zotero.org/styles/environmentalchemistry-harvard" rel="self"/>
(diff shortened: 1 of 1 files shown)


fvmolina:
Dears,
The issue seems to be the ISSN. Because it is a new style for an existing journal, I create a different one. Should I resend the new style with the existing name? If so, how to proceed?

adam3smith:
If the journal changed its style, I don't see why the old one would still be needed? I'd replace it with this one, keeping name/id as is. You can do that using the pencil icon here: https://github.com/citation-style-language/styles/blob/master/environmental-chemistry.csl

fvmolina:
OK. The fact is that the two are right now coexisting, but for a short
time. I will go ahead and replace the old one. 

Thanks 

El 2017-11-30 13:55, Sebastian Karcher escribió: 

> If the journal changed its style, I don't see why the old one would still be needed? I'd replace it with this one, keeping name/id as is. You can do that using the pencil icon here: https://github.com/citation-style-language/styles/blob/master/environmental-chemistry.csl [1] 
> 
> --
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub [2], or mute the thread [3].
 

Links:
------
[1]
https://github.com/citation-style-language/styles/blob/master/environmental-chemistry.csl
[2]
https://github.com/citation-style-language/styles/pull/3160#issuecomment-348250862
[3]

rmzelle:
Continued in #3161.

PR #3161, opened 2017-11-30
Update environmental-chemistry.csl

fvmolina:
Big change of journal style, from number to autor-date format, so it is in fact a complete replacement.


fvmolina:
Corrected

rmzelle:
@fvmolina, thanks! (I removed the "Harvard" form the title, which we'd only include if we have multiple variants of a style in the repository that need to be distinguished)

PR #3162, opened 2017-12-01
Create eurointervention.csl

naritatsu:


rmzelle:
@naritatsu, looks like you based this one on "Elsevier - Vancouver" as well, so I went ahead and added the "template" link. Thanks!

PR #3165, opened 2017-12-03
Create juristische-zitierweise-schweizer-stil.csl

stepsch-z:
This style is based on the "Juristische Zitierweise (Stüber)" style by Oliver Wolf and Philipp Zumstein (https://www.zotero.org/styles/juristische-zitierweise). The citation is adapted for the style of articles in Swiss law publications. It's similar to the style of the course book Ryser Büschi/Schlegel/Pflaum, Juristische Arbeiten erfolgreich schreiben und präsentieren, 2nd ed., Zurich 2017 (ISBN 978-3-7255-7612-8). Legal commentaries and handbooks should use publication type "entry-encyclopedia". 
See more details in the comments.

### DIFF
--- a/juristische-zitierweise-schweizer-stil.csl
+++ b/juristische-zitierweise-schweizer-stil.csl
@@ -79,18 +79,6 @@
       </substitute>
     </names>
   </macro>
-  <macro name="locator-with-label">
-    <group delimiter=" ">
-      <label variable="locator" form="symbol"/>
-      <text variable="locator"/>
-    </group>
-  </macro>
-  <macro name="firstpage-locator">
-    <group delimiter=" ">
-      <text variable="page-first"/>
-      <text variable="locator" prefix="(" suffix=")"/>
-    </group>
-  </macro>
   <macro name="issued-without-zero">
     <date variable="issued">
       <date-part name="day" form="numeric" suffix="."/>
--- a/juristische-zitierweise-schweizer-stil.csl
+++ b/juristische-zitierweise-schweizer-stil.csl
@@ -1,0 +1,390 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" default-locale="de-DE">
+  <info>
+    <title>Juristische Zitierweise - Schweizer Stil (German)</title>
+    <id>http://www.zotero.org/styles/juristische-zitierweise-schweizer-stil</id>
+    <link href="http://www.zotero.org/styles/juristische-zitierweise-schweizer-stil" rel="self"/>
+    <author>
+      <name>Stephan Schlegel</name>
+      <email>stephan.schlegel@unibas.ch</email>
+    </author>
+    <category citation-format="note"/>
+    <category field="law"/>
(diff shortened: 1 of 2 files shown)


rmzelle:
> The citation is adapted for the style of articles in Swiss law publications. It's similar to the style of the course book Ryser Büschi/Schlegel/Pflaum, Juristische Arbeiten erfolgreich schreiben und präsentieren, 2nd ed., Zurich 2017 (ISBN 978-3-7255-7612-8)

Is this CSL style similar enough to the Ryser Büschi et al. book (http://www.worldcat.org/oclc/1013162746) that we can name the style after that? We generally don't accept styles that aren't based directly on a published style guide.

stepsch-z:
Ok, I can make some changes so the style is nearly identical. Actually the problem is that there is no specific style for Swiss law publications. It would be nice if you can name the style such as "Juristische Zitierweise Schweiz (Ryser Büschi/Schlegel/Pflaum)". So other people can easy find a style for Swiss law publication and use it for own modifications.

rmzelle:
> Ok, I can make some changes so the style is nearly identical. Actually the problem is that there is no specific style for Swiss law publications. It would be nice if you can name the style such as "Juristische Zitierweise Schweiz (Ryser Büschi/Schlegel/Pflaum)". So other people can easy find a style for Swiss law publication and use it for own modifications.

Yes, that'll work. We just want to anchor each of our styles to a published citation format, so that it's clear what the style should look like. This e.g. allows us to confirm that any outside corrections submitted to us are actually improvements.

juristische-zitierweise-schweizer-stil.csl
```diff
@@ -0,0 +1,388 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" default-locale="de-DE">
```
rmzelle: `default-locale="de-DE"`

Note that we have a dedicated CSL locale style for Swiss: https://github.com/citation-style-language/locales/blob/master/locales-de-CH.xml (locale code "de-CH")

rmzelle:
I change the style title and ID a little to be more in line with our other styles.

rmzelle:
You can see the differences between the "de-DE" and "de-CH" locale files at https://www.diffnow.com/?report=5yjkw. Let me know if it makes sense to switch, and whether you're okay with the changes I made. Otherwise this looks good to merge.

stepsch-z:
Thank you for your modifications! I totally agree with with all of your changes.

stepsch-z:
I'm sorry, I don't want close this.

rmzelle:

juristische-zitierweise-schweizer.csl
```diff
@@ -0,0 +1,391 @@
...
+                      </if>
+                    </choose>
+                  </group>
+                  <text variable="title" form="short" font-style="italic" prefix="&#8211; "/>
+                </group>
+              </else>
+            </choose>
+            <text variable="locator"/>
+          </group>
+          <text variable="note" prefix=" "/>
```
adam3smith: there's not way to find an alternative here? "note" is very erratically mapped from different CSL applications and we generally try to avoid it in citation styles.

stepsch-z:
Thanks for merging!

PR #3171, opened 2017-12-05
Create oscola-no-subsequent-author.csl

JTC-librarian:
This version of OSCOLA gives repeated authors' names in full in the bibliography, rather than using the em-dashes.

### DIFF
--- a/oscola-no-subsequent-author.csl
+++ b/oscola-no-subsequent-author.csl
@@ -1,0 +1,681 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="never" default-locale="en-GB">
+  <info>
+    <title>OSCOLA (Oxford University Standard for Citation of Legal Authorities) (no subsequent author)</title>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks. We're trying to limit the proliferations of slight variations of the same style. It makes maintenance very burdensome. What is the use case for this more specifically?

JTC-librarian:
Hi,

I work for Southampton Solent University in the UK, and the style will be used with our Talis Aspire Reading List system, which pulls in styles from Zotero. This style will correct an error in this system whereby em-dashes are displayed instead of author names, even when the references are displayed individually, rather than in a bibliography.

Thanks,
James

adam3smith:
Thanks -- we generally don't accept styles for this purpose. See the second bullet under "might not accept" here: https://github.com/citation-style-language/styles/wiki/Criteria-for-Accepting-Styles 
No way to have Talis use custom styles?

rmzelle:
Yes, sorry, we don't accept styles into the repository that will only be used for individual websites. If Talis doesn't support the use of CSL styles from outside of our repository, I recommend you put in a feature request with them for that.

PR #3174, opened 2017-12-07
Create springer-basic-numeric-brackets-alphabetical.csl

enBiter:
This style should correspond to the referencing requirements of Journal of Physiology and Biochemistry http://www.springer.com/biomed/journal/13105
I made it for myself but decided to share. I've left original author since I only shuffled some lines of code around.


enBiter:
Somehow I thought it was dependent, sorry.

PR #3176, opened 2017-12-07
Update springer-basic-brackets-no-et-al-alphabetical.csl

enBiter:
Removes the space between page numbers and the dot before "doi:"

### DIFF
--- a/springer-basic-brackets-no-et-al-alphabetical.csl
+++ b/springer-basic-brackets-no-et-al-alphabetical.csl
@@ -125,9 +125,9 @@
                   <text variable="container-title" form="short" strip-periods="true"/>
                   <group delimiter=":">
                     <text variable="volume" suffix=":"/>
-                    <text variable="page"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
I don't quite understand this change. You're going from 123-456. doi:10.1234/5678 to 
123-456.doi:10.1234/5678. That seems almost certainly wrong.

enBiter:
In a previous version of the style I was getting 98:123-456 . doi:10.123456
This fix makes it 98:123-456. doi:10.123456
I'm using Office 365 ProPlus version 1710 and Zotero 5.0.30

adam3smith:
ah yes, seeing that -- but problem remains that this puts a period after the page numbers when there's no DOI, which I don't think is right. I'll see if there's a quick fix.

enBiter:
I've tested new changes and they work. Thanks!

adam3smith:
Great, thanks for spotting this & testing

PR #3177, opened 2017-12-07
Create pharmacology-and-therapeutics.csl

POBrien333:
via https://forums.zotero.org/discussion/69218/style-request-pharmacology-therapeutics#latest


adam3smith:
I think the report was that there were errors in the existing style -- did you not spot any? The Elsevier list can be wrong.

POBrien333:
ah ok. didn't read that.
I'll have a look again.

PR #3178, opened 2017-12-08
Create american-journal-of-enology-and-viticulture.csl

POBrien333:
via https://forums.zotero.org/discussion/69219/style-request-american-journal-of-enology-viticulture

The template style was made by me. Must be old. Jesus... I need to fix the old, too. So many weird things in there. I've gotten the hang of this now... Just not Chicago styles.... haha.

### DIFF
--- a/american-journal-of-enology-and-viticulture.csl
+++ b/american-journal-of-enology-and-viticulture.csl
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
   <info>
     <title>American Journal of Enology and Viticulture</title>
     <title-short>AJEV</title-short>
@@ -12,7 +12,8 @@
       <email>obrienpat86@gmail.com</email>
     </author>
     <category citation-format="author-date"/>
-    <category field="medicine"/>
+    <category field="science"/>
+    <category field="botany"/>
     <issn>0002-9254</issn>
     <updated>2017-12-08T19:33:34+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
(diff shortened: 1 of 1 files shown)


american-journal-of-enology-and-viticulture.csl
```diff
@@ -0,0 +1,207 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
```
rmzelle: @POBrien333, can you check the locale? The guide says American-English, but I don't know if "en-US" would e.g. mess up the quotes.

american-journal-of-enology-and-viticulture.csl
```diff
@@ -0,0 +1,207 @@
...
+    <id>http://www.zotero.org/styles/american-journal-of-enology-and-viticulture</id>
+    <link href="http://www.zotero.org/styles/american-journal-of-enology-and-viticulture" rel="self"/>
+    <link href="http://www.zotero.org/styles/friedrich-schiller-university-jena-faculty-of-medicine" rel="template"/>
+    <link href="http://www.ajevonline.org/content/authors" rel="documentation"/>
+    <author>
+      <name>Patrick O'Brien, PhD</name>
+      <email>obrienpat86@gmail.com</email>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="medicine"/>
```
rmzelle: @POBrien333, wine is not really a medicine ;).

adam3smith: well... https://doi.org/10.4103%2F0975-3583.74259

rmzelle: "All Breathlessness In Mitral Stenosis Is Not Pulmonary Edema- A Case of Mitral Stenosis with Cardio Respiratory Arrest" ? Don't see the wine connection.

adam3smith: oops copied the DOI from PMC. Was supposed to be this one: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3023893/

rmzelle: Better, but my complaint stands :).

(I reported the DOI problem at https://support.crossref.org/hc/en-us/requests/new. It should resolve to http://old.jcdronline.org/article/145 but resolves to http://www.jcdronline.org/article/145. Note the identical article no.)

american-journal-of-enology-and-viticulture.csl
POBrien333: Thanks. Checked and indeed, en-US is better a fit.

rmzelle:
Thanks!

PR #3180, opened 2017-12-09

POBrien333:
https://forums.zotero.org/discussion/comment/295191/#Comment_295191

### DIFF
--- a/medicine-and-science-in-sports-and-exercise.csl
+++ b/medicine-and-science-in-sports-and-exercise.csl
@@ -18,7 +18,7 @@
     <category field="medicine"/>
     <issn>0195-9131</issn>
     <eissn>1530-0315</eissn>
-    <updated>2017-12-10T10:19:18+00:00</updated>
+    <updated>2017-12-13T13:21:47+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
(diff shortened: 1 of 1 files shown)


POBrien333:
Actually. This needs a bit more work. Some conditionals for url and doi for journals. only when there is no pages.

POBrien333:
Could you both check the changes I made in macro="access" to match the guidelines?

medicine-and-science-in-sports-and-exercise.csl
```diff
@@ -66,27 +71,58 @@
...
-          <date-part name="day" prefix=" "/>
-        </date>
-      </group>
-      <group>
-        <text value="Available from: "/>
-        <text variable="URL"/>
-      </group>
-    </group>
+    <choose>
+      <if type="article-journal" match="any">
```
adam3smith: I'm afraid this doesn't quite work: If you look at
Vickers AJ. Time course of muscle soreness following different types of exercise. BMC Musculoskeletal Disorders [Internet]. 2001 [cited 2001 May 31];2(5). Available from: http://www.biomedcentral.com/1471-2474/2/5. doi:10.1186/1471-2474-2-5.

The volume and issue information is between access date and URL/DOI. The way you'd have to do this is to the [cited 2001 May 3] onto the date macro and then have only URL/DOI in access all the way at the end.

POBrien333:
thanks. That made sense. This should work.

adam3smith:
perfect, thanks!

PR #3184, opened 2017-12-12
Acta Neuropathologica update

libora:
https://forums.zotero.org/discussion/69330/andern-eines-zitationsstils
- add new Springer basic numeric et-al alphabetical
- change parent style for Acta Neuropathologica journal to new Springer style

### DIFF
--- a/dependent/acta-neuropathologica.csl
+++ b/dependent/acta-neuropathologica.csl
@@ -6,7 +6,7 @@
     <title-short>Acta Neuropathol</title-short>
     <id>http://www.zotero.org/styles/acta-neuropathologica</id>
     <link href="http://www.zotero.org/styles/acta-neuropathologica" rel="self"/>
-    <link href="http://www.zotero.org/styles/springer-basic-brackets-no-et-al-alphabetical" rel="independent-parent"/>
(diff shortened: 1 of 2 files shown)


adam3smith:
@rmzelle I think this is likely the right approach. Please merge if you agree.

rmzelle:
Do we really need this? I see that http://www.springer.com/medicine/pathology/journal/401 says "(Up to six authors, all authors should be listed. If there are more than six, list only the first six, followed by et al.)", but further down it reads "Ideally, the names of all authors should be provided, but the usage of “et al” in long author lists will also be accepted". Wouldn't it be simpler to just stick to the second rule and leave things as they are?

(if we take this, we'd also have to put the style on the skip list of the Springer metadata in the "journals" repo)

libora:
From my point of view as an editor of a scholarly journal, use "et al." in the submissions is not welcome. I can easily change it in the editorial process.

adam3smith:
good call @rmzelle -- especially since the journal also doesn't appear to be using et al. (or at least not consistently) in publication, I think it makes sense to keep the number of styles limited.

PR #3189, opened 2017-12-13
Create turkish-journal-of-emergency-medicine.csl

drhaldun:
We have created the style for Turkish Journal of Emergency Medicine, a modified version of AMA


adam3smith:
What's wrong with the existing style? Is AMA not actually correct for the journal?
https://github.com/citation-style-language/styles/blob/master/dependent/turkish-journal-of-emergency-medicine.csl

libora:
drhaldun created 2 changes:

1. line 107 - use non-superscripted numbers with brackets for in-text citation - I do not see anything about non-superscripted in-text citation and use brackets in the instruction for authors.  In the articles is used superscripted numbers without brackets. I think this change is incorrect.
2. line 115 - only 3 authors and then et-al is compliant with instruction for authors: "_References should be numbered consecutively in the order in which they are first mentioned in the text, and should be formatted in AMA style (3 authors then "et al")._ "

adam3smith:
Given @libora  's comment, we'll leave the style as a dependent using standard AMA. Thanks!

PR #3190, opened 2017-12-14
Create ephemerides-theologicae-lovanienses.csl

carljohanberglund:

PR #3191, opened 2017-12-14
Create bibliotheca-ephemeridum-theologicarum-lovaniensium.csl

carljohanberglund:


rmzelle:
(I combined this with #3190)

PR #3195, opened 2017-12-16
Create laser-and-photonics-reviews.csl

POBrien333:
via https://forums.zotero.org/discussion/69427/style-request-laser-photonics-reviews

### DIFF
--- a/laser-and-photonics-reviews.csl
+++ b/laser-and-photonics-reviews.csl
@@ -1,0 +1,156 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>Laser &amp; Photonics Reviews</title>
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks!

laser-and-photonics-reviews.csl
```diff
@@ -0,0 +1,156 @@
...
+      <else>
+        <text variable="edition"/>
+      </else>
+    </choose>
+  </macro>
+  <citation collapse="citation-number">
+    <sort>
+      <key variable="citation-number"/>
+    </sort>
+    <layout vertical-align="sup" delimiter="," prefix="[" suffix="]">
```
qwerty: @rmzelle There should be no superscript in the "citation-number":
` <layout delimiter="," prefix="[" suffix="]">`

POBrien333: I've just checked this paper for example and the number is superscripted: http://onlinelibrary.wiley.com/doi/10.1002/lpor.201700082/epdf

qwerty: Yes you are right, my apologies: it seems they have changed the layout since July of last year.

PR #3197, opened 2017-12-18
Update circulation-journal.csl

naritatsu:
I changed line 2 : page-range-format="minimal" to page-range-format="expanded" , according to the journal citation style.


naritatsu:
Do I need to wait until this change is reflected?

rmzelle:
Thanks! (it just takes us a while sometimes to respond)

PR #3199, opened 2017-12-18
Update and rename dependent/methods-in-ecology-and-evolution.csl to m…

POBrien333:
…ethods-in-ecology-and-evolution.csl

via https://forums.zotero.org/discussion/comment/296090/#Comment_296090


adam3smith:
Thanks

rmzelle:
(we still need to put this on the skip list of the "bes" metadata at https://github.com/citation-style-language/journals)

PR #3205, opened 2017-12-20
jcn-no-parenthesis.csl

sarake:
Modified Journal of Compartive Neurology citation style. Without the parentheses so it can be more customisable within the text.

### DIFF
--- a/jcn-no-parenthesis.csl
+++ b/jcn-no-parenthesis.csl
@@ -3,13 +3,11 @@
   <info>
     <title>The Journal of Comparative Neurology (without parentheses)</title>
     <id>http://www.zotero.org/styles/jcn-no-parenthesis.csl</id>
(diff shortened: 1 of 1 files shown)


rmzelle:
@sarake, I don't think we'll accept this. With most CSL applications, the expectation is that the user writes out the author names by hand and suppress the authors in the in-text citation if they want to write something like "As demonstrated by Jones et al. (2007)" (only "(2007)" would be rendered by the style here). If you want to write something like `(i.e., independent of trophic factors; Southwell et al., 2012)`, you would use a prefix on the cited item.

sarake:
Fair enough. I apologise for the inconvenience.

Best regards,

--
Sara
+56 9 9332 1622

rmzelle:
No problem! And you can of course use this style yourself if it works better for you.

PR #3208, opened 2017-12-20
Create Speculum style

adunning:
Style for Speculum: <http://www.journals.uchicago.edu/journals/spc/style-sheet>.

### DIFF
--- a/speculum.csl
+++ b/speculum.csl
@@ -269,23 +269,6 @@
       </if>
     </choose>
   </macro>
-  <macro name="edition">
-    <choose>
-      <if type="book chapter graphic motion_picture paper-conference report song" match="any">
-        <choose>
-          <if is-numeric="edition">
-            <group delimiter=" ">
-              <number variable="edition" form="ordinal"/>
-              <text term="edition" form="short"/>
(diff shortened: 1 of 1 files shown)


adunning:
I seem to have broken the test bot, as it is complaining about macros that I have removed.

rmzelle:
@adunning, since this is a University of Chicago Press journal (at least per 2016: http://www.journals.uchicago.edu/journals/spc/about), and since http://www.journals.uchicago.edu/journals/spc/style-sheet says "For usage issues not found in Speculum, authors should consult the 16th edition of The Chicago Manual of Style (Chicago, 2010)", did you check whether this style is the same as either the "note" or "full note" Chicago 16th edition CSL styles we have (https://www.zotero.org/styles?q=16th)?

If so, we can just create a "dependent" style for this journal.

adunning:
Not the same; see the style guide.

rmzelle:
Okay, thanks! It looks like there was a temporary problem with the test bot infrastructure (Travis CI) that resolved itself.

adunning:
Thank you!

PR #3222, opened 2017-12-24
Update petit-chicago-author-date.csl

adunning:
Fix missing point.


rmzelle:
Thanks. I rewrote the style a little to move the ". " outside of the macro (CSL styles are generally easier to work with if you write macros that don't contain any separating punctuation for the context in which they're used).

PR #3223, opened 2017-12-24
Create pedagogicka-orientace.csl

libora:
Journal published by Masaryk University in Brno


rmzelle:
@libora, thanks! (it looks like this journal publishes primarily in Czech, but I guess our standard APA style doesn't cleanly localize to Czech, so let's stick with a default-locale of en-US for now)

PR #3225, opened 2017-12-24
Create trendy-v-podnikani.csl

libora:


rmzelle:
Thanks! (renamed the style to the English name since that's what the locale is set to)

PR #3226, opened 2017-12-25
Create reproduction-fertility-and-development.csl

okingfisher:
A new CSL for the journal Reproduction, Fertility and Development (http://www.publish.csiro.au/rd/AbouttheJournal). Based in the Journal of Biogeography CSL. Not sure about Communication references correctness.

### DIFF
--- a/reproduction-fertility-and-development.csl
+++ b/reproduction-fertility-and-development.csl
@@ -119,7 +119,7 @@
       </choose>
     </group>
   </macro>
-  <citation collapse="year-suffix" et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true">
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks -- I'd actually suggest you go back and modify https://github.com/citation-style-language/styles/blob/master/international-journal-of-wildland-fire.csl instead to get this style. It's by the same publisher and almost correct already (only think I see that's wrong are the authors). There are a number of unorthodox choices (and some things that aren't quite right) and rather than fixing this all up in this version, you'll save time starting with a better template.

okingfisher:
Redone from international-journal-of-wildland-fire.csl and resubmitted.

okingfisher:
OK, looks right this time.

okingfisher:
Looks right now...

okingfisher:
OK, OK, now it seems right! some minor mistakes I overlooked...

adam3smith:
Great, thanks!

PR #3229, opened 2017-12-26
Create michael-langlois-chronobibliographie-desc.csl

mlanglois:


rmzelle:
@mlanglois, can you explain what this CSL style will be used for? Is it based on a published style guide?

(this pull request continues #3227, #3228)

mlanglois:
It's used for bibliographies ordered by date and page number when listing several contributions from the same volume. I use it for CVs. But I decided not to make it public after all. Thanks!

rmzelle:
> I use it for CVs. But I decided not to make it public after all.

Okay, thanks for explaining. We generally don't accept personal styles here, just so you're aware. See also https://github.com/citation-style-language/styles/wiki/Criteria-for-Accepting-Styles

PR #3230, opened 2017-12-27
Update american-medical-association.csl

fredolin73:
AMA has changed the handling of DOI´s- There is no period after a DOI anymore.
See here:
http://www.amamanualofstyle.com/page/updates;jsessionid=C9614A8B750AFA437F92204D8DF415AC#2016_09_01

I have used the editor to spot the period and simply deleted it. Hope this will be ok;-)
Feedback welcome!

### DIFF
--- a/american-medical-association.csl
+++ b/american-medical-association.csl
@@ -242,7 +242,7 @@
           </group>
         </else>
       </choose>
-      <text prefix=" " macro="access" suffix="."/>
+      <text prefix=" " macro="access" suffix=""/>
     </layout>
   </bibliography>
 </style>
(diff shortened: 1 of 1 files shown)


fredolin73:
I have contacted AMA a while ago and asked them about this. They gave me the link to the update page where the DOI question is adressed. 

They also told me that after a URL is still a period but that this shall be adapted for the new edition. Till then, a period after URL is correct.
Fredolin

rmzelle:
Okay, thanks for confirming.

I edited your original submission which would also have removed the period after the accessed date.

fredolin73:
The period still seems to be present after the DOI.

PR #3232, opened 2017-12-29
Update university-of-cambridge-faculty-of-history.csl

ntdmiller:
Corrected several errors: changed to long title instead of short title in initial citations; corrected thesis and dictionary format.


ntdmiller:
I also introduced an unintentional title change from "University of Cambridge - Faculty of History" to "Cambridge University Faculty of History." The former is preferred.

PR #3233, opened 2017-12-29
Update the-historical-journal.csl

ntdmiller:
Corrected several errors: Changed from short title to full title on first entries; corrected thesis and dictionary format.


adam3smith:
thanks -- sorry for the delay

PR #3235, opened 2018-01-02
Create indian-journal-of-physics.csl

POBrien333:
via https://forums.zotero.org/discussion/69566/style-request-indian-journal-of-physics


rmzelle:
Thanks!

PR #3237, opened 2018-01-02
Update international-review-of-the-red-cross.csl

scarnecchia:
Updating to add proper "ibid" and "above note" format per documentation: https://www.icrc.org/data/files/publications/en/2015-guidelines-for-authors-irrc.pdf


rmzelle:
Thanks!

PR #3238, opened 2018-01-03
Update the-journal-of-egyptian-archaeology.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/297043/#Comment_297043


rmzelle:
Thanks!

PR #3244, opened 2018-01-05
Update international-review-of-the-red-cross.csl

scarnecchia:
Fixed two additional errors: 
1. delimiters in the "above note" format were broken.
2. initialization for authors is now turned off, per documentation.

I'm doing this on the fly as I bust through citations for a paper. After this I will hold off and push one large update if I discover more formatting issues.


rmzelle:
Thanks!

PR #3250, opened 2018-01-07
Update and rename dependent/molecular-oncology.csl to molecular-oncol…

POBrien333:
…ogy.csl

This was the one requested via email, Rintze.
Only adapted for article-journal. Couldn't find any book or chapter citations.


rmzelle:
Thanks! (should put this on the Elsevier skip list)

PR #3254, opened 2018-01-11
Create cambridge-university-press-note.csl

POBrien333:
via https://forums.zotero.org/discussion/69540/style-request-cambridge-university-press-short-title-system-law

### DIFF
--- a/cambridge-university-press-note.csl
+++ b/cambridge-university-press-note.csl
@@ -13,7 +13,7 @@
     </author>
     <category citation-format="note"/>
     <category field="generic-base"/>
-    <updated>2018-01-11T14:53:30+00:00</updated>
+    <updated>2018-01-15T14:34:39+00:00</updated>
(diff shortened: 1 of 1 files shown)


rmzelle:
@POBrien333, this is just for the field of law, right? Maybe we should rename the style to `<title>Cambridge University Press - Law (note)</title>`.

And since https://authornet.cambridge.org/information/academic/guidelines_for_authors.asp links to many different documents, a second "documentation" link to https://authornet.cambridge.org/information/academic/downloads/Additional%20style%20guidelines%20for%20Law.pdf would be useful (if that's the only file you used).

POBrien333:
I am not sure that this is only for law. the style is generic and has an extra .pdf for law specific style questions.

rmzelle:
Thanks!

POBrien333:
Turns out we'll need two styles. One generic note style and one for law specifically. there is a few small differences.

PR #3256, opened 2018-01-12
Sbl2 updates

tmykkanen:
Several updates to bring the style more in compliance with SBL 2.

Most significantly:
* Handling for signed and unsigned dictionary entries (SBL 6.3.6 & 6.3.7; also this series of posts: https://sblhs2.com/2017/03/28/citing-reference-works-1/)
* Handling for multivolume works (6.2.20-6.2.23)
* Using short titles rather than ibid per CMS 17 recommendations (CMS 14.34 specifically).

NB: Current changes should only affect footnote citations. Bibliography updates are coming once I get a chance to check the changes locally.

### DIFF
--- a/society-of-biblical-literature-fullnote-bibliography.csl
+++ b/society-of-biblical-literature-fullnote-bibliography.csl
@@ -596,6 +596,16 @@
       <else-if type="article-journal">
         <text variable="locator" prefix=": "/>
       </else-if>
+      <else-if type="entry-dictionary entry-encyclopedia">
+        <choose>
+          <if position="first" variable="volume" match="all">
+            <text macro="point-locators-subsequent" prefix=" "/><!--SBL 6.3.6 prefers space in first entry before vol info.-->
+          </if>
+          <else>
+            <text macro="point-locators-subsequent" prefix=", "/><!--SBL 6.3.6 prefers comma in subsequent entries before vol info.-->
+          </else>
+        </choose>
+      </else-if>
       <else>
         <text macro="point-locators-subsequent" prefix=", "/>
       </else>
--- a/society-of-biblical-literature-fullnote-bibliography.csl
+++ b/society-of-biblical-literature-fullnote-bibliography.csl
@@ -34,8 +34,8 @@
       <term name="translator" form="short">trans.</term>
       <term name="editortranslator" form="verb">edited and translated by</term>
       <term name="collection-editor">
-          <single>ed.</single>
-          <multiple>eds.</multiple>
+        <single>ed.</single>
+        <multiple>eds.</multiple>
       </term>
       <term name="collection-editor" form="verb">edited by</term>
     </terms>
@@ -98,7 +98,8 @@
     </names>
   </macro>
   <macro name="collection-editor">
-    <group delimiter=", " prefix=". "><!--TODO: Refactor? Bib uses prefix whereas CSL style prefers groups with delimeters.-->
+    <group delimiter=", " prefix=". ">
+      <!--TODO: Refactor? Bib uses prefix whereas CSL style prefers groups with delimeters.-->
       <names variable="collection-editor" delimiter=", ">
         <label form="verb" prefix=" " text-case="capitalize-first" suffix=" "/>
         <name and="text" delimiter=", "/>
@@ -334,19 +335,21 @@
   <macro name="collection-title">
     <choose>
       <if variable="volume number-of-volumes" match="any">
-      <!-- Presence of Volume info indicates multivolume work rather than series -->
-          <choose>
-            <!--This section required here rather than in title-note because the group delimiter inserts an unneeded comma otherwise-->
-            <if type="chapter" variable="container-title" match="all"> <!--Filter SBL 6.2.23-->
-            </if>
-            <else-if type="chapter" variable="collection-title" match="all"> <!--Filter SBL 6.2.22-->
-              <group>
-                <text term="volume" form="short" suffix=" "/>
-                <number variable="volume" form="numeric"/>
-                <text value=" of "/>
-              </group>
-            </else-if>
-          </choose>
+        <!-- Presence of Volume info indicates multivolume work rather than series -->
+        <choose>
+          <!--This section required here rather than in title-note because the group delimiter inserts an unneeded comma otherwise-->
+          <if type="chapter" variable="container-title" match="all">
+            <!--Filter SBL 6.2.23-->
+          </if>
+          <else-if type="chapter" variable="collection-title" match="all">
+            <!--Filter SBL 6.2.22-->
+            <group>
+              <text term="volume" form="short" suffix=" "/>
+              <number variable="volume" form="numeric"/>
+              <text value=" of "/>
+            </group>
+          </else-if>
+        </choose>
         <text variable="collection-title" form="short" text-case="title" font-style="italic"/>
         <text variable="collection-number" prefix=" "/>
       </if>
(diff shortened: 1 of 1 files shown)


society-of-biblical-literature-fullnote-bibliography.csl
```diff
@@ -656,42 +828,118 @@
...
-          </group>
-        </if>
-        <else-if position="ibid">
-          <text term="ibid"/>
-        </else-if>
-        <else-if position="subsequent">
+  <macro name="unsigned-dictionary-note">
+    <group delimiter=", ">
+      <group>
+        <text variable="container-title" font-style="italic" text-case="uppercase"/>
```
adam3smith: really uppercase? What's an example here?

society-of-biblical-literature-fullnote-bibliography.csl
```diff
@@ -656,42 +828,118 @@
...
             <text macro="contributors-short"/>
-            <text macro="title-short"/>
-            <text macro="disambiguate-chapter"/>
-            <text macro="point-locators-subsequent"/>
-          </group>
-        </else-if>
-        <else>
+            <choose>
+              <if position="ibid" match="none">
+                <!-- Excludes Title for ibid entries per CMS 14.34.-->
```
adam3smith: Why do we thing SBL is following CMS 17th edition on this? I'm worried this will cause a lot of confusion.

society-of-biblical-literature-fullnote-bibliography.csl
```diff
@@ -656,42 +828,118 @@
...
+              <!-- Signed Lexicon/Dictionary/Encyclopedia -->
+              <text macro="signed-dictionary-note"/>
+            </else>
+          </choose>
+        </if>
+        <!-- Not Lexicon/Dictionary/Encyclopedia -->
+            <else-if position="subsequent">
+                  <group delimiter=", ">
+                    <text macro="contributors-short"/>
+                    <!-- Implements conditional check to exclude title in ibid entries (see CMS 14.34) -->
```
adam3smith: 1) I'd prefer to to nest the ibid check into subsequent simply for readability
2) As per above, I'm worried about just following CMS in their new rule unless we have a very clear indication that's what they want.

society-of-biblical-literature-fullnote-bibliography.csl
```diff
@@ -313,20 +332,76 @@
...
   </macro>
   <macro name="collection-title">
-    <text variable="collection-title" form="short"/>
-    <text variable="collection-number" prefix=" "/>
+    <!--TODO: Macro is identical to collection-title-note. Seperate macros not needed?-->
+    <choose>
+      <if variable="volume number-of-volumes" match="any">
+      <!-- Presence of Volume info indicates multivolume work rather than series -->
+          <choose>
+            <!--TODO: Is this needed? Creates double "in".-->
```
adam3smith: what's the double "in" referring to? We'd almost certainly not want that, so when does it occur?

society-of-biblical-literature-fullnote-bibliography.csl
```diff
@@ -313,20 +332,76 @@
     </choose>
   </macro>
   <macro name="collection-title">
-    <text variable="collection-title" form="short"/>
-    <text variable="collection-number" prefix=" "/>
+    <!--TODO: Macro is identical to collection-title-note. Seperate macros not needed?-->
```
adam3smith: If the macros are identical, we'd not want to duplicated them.

tmykkanen: This comment is no longer relevant as I had to add some volume handling to the collection-title macro that is different from the collection-title-note macro. I'll remove comments which are no longer relevant via a new commit below.

society-of-biblical-literature-fullnote-bibliography.csl
tmykkanen: I forgot to remove this comment once I fixed the error.  I'll remove comments which are no longer relevant via a new commit below.

society-of-biblical-literature-fullnote-bibliography.csl
tmykkanen: The only thing  that gets to this macro should be an unsigned dictionary article. The formatting is from https://sblhs2.com/2017/03/30/citing-reference-works-2-lexica/.

I am fine removing the uppercase and just entering dictionary titles as uppercase in Zotero if that is a better way to handle it. Let me know and I can change it and submit a new commit.

society-of-biblical-literature-fullnote-bibliography.csl
tmykkanen: Good point! I'm basing this off guidance from my seminary professors at Bethlehem College and Seminary (mostly Andy Naselli). I think it is permitted currently with SBL, but I don't know that it has been officially adopted. I'll double check and see if we have any documentation confirming they are adopting the style. If not, I can split this preference off and make a non-ibid style to share with my seminary classmates while not including this change in the official pull request.

society-of-biblical-literature-fullnote-bibliography.csl
tmykkanen: 2. See above re: CMS rules.

1. I'm sorry, I'm not following what you mean. The ibid check is nested inside the subsequent check as far as I can tell, which makes me think I'm not understanding what you mean.

tmykkanen:
Thanks for the feedback @adam3smith. Let me know what else I can do to move this along. I'm eager for feedback as this is my first dive into CSL.

society-of-biblical-literature-fullnote-bibliography.csl
adam3smith: If you check how we do this e.g. in CMS:
https://github.com/citation-style-language/styles/blob/master/chicago-fullnote-bibliography.csl#L1310
We first check if ibid/ibid-with-locator, then check if subsequent. What you do is to check for if is subsequent and then differentiate for ibid within that. Both are functionally the same, but all other styles using ibid follow the way this is currently done in CMS, so I'd prefer to keep it that way just for simplicy.

tmykkanen: Haha, no problem. :) If I understand correct, what I'd need to do is essentially:

If ibid => contributors | locators
If subsequent => contributors | title | locators

This would filter out the ibid exceptions and pass the rest to subsequent. The code would be less DRY, but more readable. Correct?

adam3smith: yes, exactly, thanks!

tmykkanen: Thanks @adam3smith. Fixed this with commit 2969a2a67.

society-of-biblical-literature-fullnote-bibliography.csl
tmykkanen: Fixed with commit f3269e6.

society-of-biblical-literature-fullnote-bibliography.csl
tmykkanen: Fixed with commit f3269e6.

adam3smith:
Cool, thanks -- this looks good to me, but I didn't do any testing, so let's see if we get any complaints.

adam3smith:
Hi @tmykkanen we've received a complaint about the handling of series, related to the multivolume works fixes above here: https://forums.zotero.org/discussion/71221/style-error-for-sbl-handbook-of-style-2nd-ed-series-should-output-in-roman#latest 
would you be willing to chime in & explain the current logic and, if necessary, adapt?

tmykkanen:
@adam3smith, I took a look and I couldn't reproduce the OP's problem. I posted on the zotero forums asking for more info. I may not be correctly understanding the nature of the problem. Series titles look to be working fine to me.

adam3smith:
much appreciated!

PR #3257, opened 2018-01-12
Create twentieth-century-music.csl

POBrien333:
via https://forums.zotero.org/discussion/69836/style-request-twentieth-century-music

Likely that there is some mistakes in there.


rmzelle:
Thanks! (looks like this still has some issues, but let's accept what you have now)

PR #3259, opened 2018-01-12
Create iso690-author-date-no-abstract-en.csl

paezm:
I change the style iso-690-author-date-en.csl. In this version, the abstract has been removed. And instead of "Accessed 9 November 2016"  I changed by "Viewed 9 November 2016".

### DIFF
--- a/haute-ecole-de-gestion-de-geneve-iso-690-eng.csl
+++ b/haute-ecole-de-gestion-de-geneve-iso-690-eng.csl
@@ -2,7 +2,7 @@
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" default-locale="en-US">
   <info>
     <title>Haute école de gestion de Genève - ISO-690 (English)</title>
-    <id>http://www.zotero.org/styles/haute-ecole-de-gestion-de-geneve-iso-690</id>
+    <id>http://www.zotero.org/styles/haute-ecole-de-gestion-de-geneve-iso-690-eng</id>
     <link href="http://www.zotero.org/styles/haute-ecole-de-gestion-de-geneve-iso-690" rel="self"/>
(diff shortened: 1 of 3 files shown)


adam3smith:
What guide exactly is this based on? "ISO 690" is such a vague reference -- providing more general guidance than a specific citation format -- that I'd like to prevent the proliferation of styles. E.g., "Viewed" vs. "Accessed" certainly isn't something that the ISO standard specifies, is it?

paezm:
ISO 690:2010. My library school wrote a guide about this style https://www.hesge.ch/heg/infotheque/citations-et-references-bibliographiques.

adam3smith:
Great, thanks -- since this is only "based on" ISO 690:2010, I think we'll want to name this after the school rather than ISO 690. Looking briefly at the guide, I think it also makes sense to remove the note and archive macro and the block display?

paezm:
Sorry but I'm not sure I understand. It's better if I change the name ? And I add the School library name?
Is it neccessary to remove the note and archive macro and the block display ? :)

paezm:
It's better if we keep the name iso-690 because many schools use this standard. Et we have respected this standard. It's really specified "viewed" and not "accessed". How should I proceed to have the style validated? Sorry, this is the first time I'm correcting a style.. ;)

rmzelle:
(I renamed the style to what we'd typically do in these cases. We could add back the "author-date" descriptor if your institution uses more than one type of ISO 690 style (like the "numeric" or "note" versions))

paezm:
Thank you @rmzelle

paezm:
@rmzelle Now, my style is submitted ? 
How to display my style on www.zotero.org/styles?

paezm:
@adam3smith Hi ! How to display my style on www.zotero.org/styles?

adam3smith:
Thanks!

paezm:
Thank you!!! :D

PR #3263, opened 2018-01-15
Create the-canadian-journal-of-chemical-engineering.csl

POBrien333:
via https://forums.zotero.org/discussion/69670/style-request-the-canadian-journal-of-chemical-engineering


rmzelle:
Thanks!

PR #3267, opened 2018-01-16
Journal_IMTSP (2018).csl

galisteo:
New style for Mendeley


adam3smith:
This is for a journal, right? What's the exact name of the journal? Is the style documented online somewhere? Have you changed the style from the original Vancouver?

rmzelle:
Thanks! Sorry for the long wait!

PR #3271, opened 2018-01-17
Update for oecologia-australis.csl

gudryan:
I would like to propose update to oecologia-australis.csl
The new file is attached bellow.

[oecologia-australis.txt](https://github.com/citation-style-language/styles/files/1639353/oecologia-australis.txt)


rmzelle:
Thanks. Could you please follow our instructions at https://github.com/citation-style-language/styles/blob/master/CONTRIBUTING.md#3b-submitting-changes-to-an-existing-style and update the existing style at https://github.com/citation-style-language/styles/blob/master/oecologia-australis.csl? You're currently adding a second copy of the style to the repository (with the file name of "oecologia-australis" instead of "oecologia-australis.csl").

PR #3272, opened 2018-01-17
chichago edit 3

tamiastie:
i made some modif


adam3smith:
Thank you. If this is a personal customization -- as it appears to be -- is there any reason you'd need it on the CSL repository? If this is not just a personal version, could you explain what this is for more precisely?

PR #3273, opened 2018-01-18
Update oecologia-australis.csl

gudryan:
I am adding some new roles for Oecologia Australis style.
Thank you.
Gudryan

### DIFF
--- a/oecologia-australis.csl
+++ b/oecologia-australis.csl
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" default-locale="en-US" demote-non-dropping-particle="never">
-  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
     <title>Oecologia Australis</title>
     <title-short>OA</title-short>
     <id>http://www.zotero.org/styles/oecologia-australis</id>
     <link href="http://www.zotero.org/styles/oecologia-australis" rel="self"/>
     <link href="http://www.zotero.org/styles/apa" rel="template"/>
+    <link href="https://revistas.ufrj.br/index.php/oa/about/submissions#authorGuidelines" rel="documentation"/>
     <author>
       <name>Gudryan J. Barônio</name>
       <email>gudryan@gmail.com</email>
@@ -26,7 +26,7 @@
     <category citation-format="author-date"/>
     <category field="biology"/>
     <issn>2177-6199</issn>
-    <updated>2018-01-17T13:58:40+00:00</updated>
+    <updated>2018-01-21T10:43:04+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
@@ -620,7 +620,7 @@
         <text macro="locators"/>
         <group>
           <choose>
-            <if type="chapter book" match="all" variable="edition">
+            <if type="chapter book" match="any" variable="edition">
               <text variable="edition" prefix=" "/>
               <label prefix=" " variable="edition" form="short"/>
             </if>
--- a/oecologia-australis.csl
+++ b/oecologia-australis.csl
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" default-locale="en-US" demote-non-dropping-particle="never">
-  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
     <title>Oecologia Australis</title>
     <title-short>OA</title-short>
     <id>http://www.zotero.org/styles/oecologia-australis</id>
     <link href="http://www.zotero.org/styles/oecologia-australis" rel="self"/>
     <link href="http://www.zotero.org/styles/apa" rel="template"/>
+    <link href="https://revistas.ufrj.br/index.php/oa/about/submissions#authorGuidelines" rel="documentation"/>
     <author>
       <name>Gudryan J. Barônio</name>
       <email>gudryan@gmail.com</email>
--- a/oecologia-australis.csl
+++ b/oecologia-australis.csl
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" default-locale="en-US" demote-non-dropping-particle="never">
-  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
     <title>Oecologia Australis</title>
     <title-short>OA</title-short>
--- a/oecologia-australis.csl
+++ b/oecologia-australis.csl
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" default-locale="en-US" demote-non-dropping-particle="never">
+  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
     <title>Oecologia Australis</title>
     <title-short>OA</title-short>
     <id>http://www.zotero.org/styles/oecologia-australis</id>
     <link href="http://www.zotero.org/styles/oecologia-australis" rel="self"/>
     <link href="http://www.zotero.org/styles/apa" rel="template"/>
-    <link href="https://revistas.ufrj.br/index.php/oa/about/submissions#authorGuidelines" rel="documentation"/>
     <author>
       <name>Gudryan J. Barônio</name>
       <email>gudryan@gmail.com</email>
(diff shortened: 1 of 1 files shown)


oecologia-australis.csl
```diff
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" default-locale="en-US" demote-non-dropping-particle="never">
+  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
     <title>Oecologia Australis</title>
     <title-short>OA</title-short>
     <id>http://www.zotero.org/styles/oecologia-australis</id>
     <link href="http://www.zotero.org/styles/oecologia-australis" rel="self"/>
     <link href="http://www.zotero.org/styles/apa" rel="template"/>
```
adam3smith: why are you removing this? These seem like the most recent guidelines we have

oecologia-australis.csl
```diff
@@ -148,9 +148,9 @@
             </choose>
           </group>
         </if>
-        <else-if type="legislation webpage" match="any">
+        <else-if type="legislation webpage report" match="any">
           <group prefix="(" delimiter=" " suffix=").">
-            <text value="Retrieved on"/>
+            <text value="Accessed in"/>
```
adam3smith: "Accessed in" + date is not proper English -- the styleguides seems to suggest: Retrieved, `date`, from `URL` shouldn't we do that rather?

gudryan: @adam3smith in relation to "retrieved on/accessed in" I agree with you. However, I am only changing this style according to the required by Oecologia Australis in-chief editor. I tried to argue in order to follow style guides, but editor preferred on this form.

oecologia-australis.csl
gudryan: sorry, removing what? I didn´t understand your question.

adam3smith: The link works: https://www.dropbox.com/s/azm8ycbkk229uce/Diretrizes%20OA%202017_12julho2017.pdf?dl=0

oecologia-australis.csl
adam3smith: Can you maybe emphasize again that "Accessed in" is *not* a matter of taste. It is simply incorrect English. I'd be very reluctant to accept that. Is the style as requested by the editor documented anywhere?

oecologia-australis.csl
gudryan: Yes, but these are the old rules. I will include in some minutes one link to general journal guidelines. Thank you for calling my attention to this.

oecologia-australis.csl
```diff
@@ -61,9 +61,9 @@
     <choose>
       <if type="article-journal chapter paper-conference entry-dictionary entry-encyclopedia" match="none">
         <group delimiter=", ">
-          <names variable="editor translator" delimiter=", " suffix=" ">
```
adam3smith: So the new guide you posted still lists translators for books, albeit in consistently:
Borges, J. L. 1999. Selected non-fictions. E. Weinberger (Ed.), **E. Allen, S. J. Levine, & E. Weinberger, Trans.** New York: Viking: p. 559.

I think it would be odd to leave translators out of a citations, but I agree with the change of the label, let's change line 64 back to "editor translator" but keep the rest of the changes

oecologia-australis.csl
adam3smith: Please do. Both "Accessed + day" and "Accessed on + day" are fine. Accessed in would only work for months, not for specific days.

oecologia-australis.csl
```diff
@@ -268,42 +268,54 @@
...
-              <text variable="number-of-pages"/>
+              <text variable="number-of-pages" suffix="."/>
               <text variable="publisher" prefix="ed. " suffix=", "/>
               <text variable="publisher-place"/>
             </if>
           </choose>
         </group>
       </if>
       <else-if type="thesis" match="any">
+        <group suffix=".">
```
adam3smith: No need for the group here. just do `<text variable="genre" prefix=" " suffix="."/>`

oecologia-australis.csl
```diff
@@ -268,42 +268,54 @@
...
+          <text variable="genre" prefix=" "/>
+        </group>
         <group delimiter=", ">
           <text variable="publisher-place"/>
-          <text variable="publisher"/>
+          <text variable="publisher" prefix=" "/>
```
adam3smith: no need for the prefix here. It's already in the delimiter above, but I suspect you want to set in on the group (i.e. line 280)

oecologia-australis.csl
```diff
@@ -268,42 +268,54 @@
...
+                  </else-if>
                   <else>
                     <text variable="event-place"/>
                   </else>
                 </choose>
+                <text variable="publisher" font-variant="normal" prefix=" "/>
               </group>
             </if>
           </choose>
           <text variable="number-of-pages" prefix="p. "/>
```
adam3smith: The citation style is very weird here: Dunnett, N., & Kingsbury, N. 2008. Planting green roofs and living walls. 2nd ed. Portland, OR: Timber Press: p. 328.

reads like you're referring to page 328 of the book, while the variable `number-of-pages` refers to the total number of pages in the book. Are you sure you want that?

oecologia-australis.csl
```diff
@@ -268,42 +268,54 @@
...
+                  <if type="legislation"/>
+                  <else-if match="any" variable="publisher-place">
                     <text variable="publisher-place" font-style="normal" prefix=" "/>
-                  </if>
+                  </else-if>
                   <else>
                     <text variable="event-place"/>
                   </else>
                 </choose>
+                <text variable="publisher" font-variant="normal" prefix=" "/>
```
adam3smith: this prefix (and the one in line 300) should be moved to the group, too. The space is in the group delimiter.

oecologia-australis.csl
```diff
@@ -608,9 +620,9 @@
         <text macro="locators"/>
         <group>
           <choose>
-            <if type="chapter" match="all" variable="edition">
-              <text variable="edition" prefix=" " suffix="ª "/>
-              <label variable="edition" form="short"/>
+            <if type="chapter book" match="all" variable="edition">
```
adam3smith: this doesn't work. A test for `type="chapter book" match="all"` is _never_ true as an item can't be book and chapter at once But I don't think you need the test for edition here. just do 

```
<if type="chapter book" match="any">
    <text variable="edition"/>
  </group>
</if>
```

oecologia-australis.csl
gudryan: I am waiting for the editorial opinion about this.

oecologia-australis.csl
gudryan: Translator was excluded by new rules. As I said you before, this rules found in link are not updated, also by this I as reluctant in add this link.

oecologia-australis.csl
gudryan: I included in a group to be more visualizable during the editions, I think it is not prejudicial.

oecologia-australis.csl
gudryan: Isn´t it on a group? I didn´t understand the comment.

oecologia-australis.csl
gudryan: Ok, I got it, but due to an unknown reason the group prefix was not working in some software, I included it here once software excludes dual prefix when two ones work.

oecologia-australis.csl
gudryan: I agree with you, but how I inform you before I am only the programmer. I don´t have an opinion on journal rules to do any modification. I will contact the editor and decide better these changes.

oecologia-australis.csl
gudryan: The information about edition needs to be exhibited in both cases, when reference is chapter or a book. I will correct match by "any".

oecologia-australis.csl
adam3smith: No, sorry, that's just redundant code. Please remove.

oecologia-australis.csl
adam3smith: no you have `prefix=" "` on the publisher variable. But you already have a space between `publisher-place` and `publisher` because of the `delimiter=", "` in the group. 

On the other hand, you don't have, best I can tell, a space before `publisher-place` at all. So you should move `prefix=" "` to the group two lines above rather than keeping it on the variable to _both_ avoid the duplicate space (even if Zotero will strip that ) and to make sure the place of publication has a space in front of it when it's present. I'm suspecting you tested this with thesis data without a space.

oecologia-australis.csl
gudryan: Editorial board decided for "retrieved in". I will change it.

oecologia-australis.csl
gudryan: Ok, I will exclude it.

oecologia-australis.csl
gudryan: Ok, I will exclude it.

oecologia-australis.csl
gudryan: Ok, I will chang it.

oecologia-australis.csl
gudryan: Editorial board maintain "p. 328."

gudryan:
Please @adam3smith, could you merge it in order to make available the final version?

adam3smith:
Why did the documentation link get deleted again? While I'm very pleased that journals are eager to get their CSL styles up, this should still go hand-in-hand with publishing updated styleguides and we'd really want a documentation link included.

gudryan:
Sorry, I excluded without intention. See now that I included it again.

gudryan:
@adam3smith please, what we need to complete style update?

adam3smith:
Thanks. We merge these as quickly as we can, but CSL is run by volunteers, you'll sometimes have to allow a couple of days.

PR #3275, opened 2018-01-19
expand page ranges in bibliography

gedankenraum:
APA requires full (non-abbreviated) page ranges (e.g., 512-518) instead of shortened ones (e.g., 512-8). Because the latter is used in many medical journals, you often get shortened ranges pulled from online archives and have to edit by hand. The new option handles that automatically (tested and works in Mendeley).


adam3smith:
Thanks

PR #3276, opened 2018-01-19
Update chicago-fullnote-bibliography.csl

flatride:
I added the Date (issued) to the abbreviated subsequent citations

### DIFF
--- a/chicago-fullnote-bibliography.csl
+++ b/chicago-fullnote-bibliography.csl
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="display-and-sort" page-range-format="chicago">
   <info>
-    <title>Chicago Manual of Style 17th edition (full note)</title>
+    <title>Chicago Manual of Style 17th edition (full note) Subsequent Date</title>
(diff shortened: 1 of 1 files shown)


adam3smith:
Why? The Chicago Manual doesn't do this for short notes (with few exceptions such as letters, which would require a more elaborate fix)

flatride:
Thanks for your swift reply. 
I do not need to adhere to a specific style for my research. I can simply have it the way I find it useful.
Since it isnt possible to simply change the citation style locally without zotero always changing back to the original style, I had to do it this way.

adam3smith:
You can absolutely edit styles in Zotero locally and have changes remain, you just need to change style title and ID as explained here: https://www.zotero.org/support/dev/citation_styles/style_editing_step-by-step#change_the_style_title_and_id

flatride:
I did change the Style title and the ID but it still wouldnt let me. It always switches back!

adam3smith:
Rename the file, too, just to be sure, but I'm absolutely positive this works. Are you sure you've changed the ID?

flatride:
Habs rausgefunden! Musste im Word file in den document preferences den Zitierstil ändern, nicht bloss in Zotero auf meinen Stil klicken (das ändert den Zitierstil im Worddokument nämlich auch, allerdings nur so lange, bis Zotero wieder neu gestartet wird.
Vielen Dank Dir und schönes Wochenende!

In case you do not speak German – it works now. thanks!

PR #3280, opened 2018-01-22
Update journal-of-biogeography.csl

maoxin:
From the homepage of the journal, we can get that "When a work has two authors, cite both names every time the reference occurs in text. When a work has three, four, or five authors, cite all authors the first time the reference occurs; subsequent citations include only the surname of the first author followed by et al., (not Italicized and with a period after “al.”) and the year if it is the first citation of the reference within a paragraph.", "Please include up to seven authors in the list (use “&” before last author name). For eight or more authors please list the first six and then use ellipses followed by last author (do not use “&” before last author name)", and "The basic reference form for a journal paper is: Author (date).Paper title. Journal, Volume, page; and for a book citation: Author (date). Book title. Place of publication, publisher.". These features cannot be achieved by the old version, so I propose this version.


adam3smith:
Thank you -- I wonder, though, if it wouldn't make more sense to just link this style directly to APA style, as per the instructions: "References are styled according to the sixth edition of the Publication Manual of the American Psychological Association. "

rmzelle:
> I wonder, though, if it wouldn't make more sense to just link this style directly to APA style

Done. There was another style for a Wiley journal that was dependent on this one, but that refers to APA as well, so I switched that one over too.

rmzelle:
Thanks both! Sorry for the long wait.

PR #3283, opened 2018-01-22
Create historical-social-research.csl

POBrien333:
via https://forums.zotero.org/discussion/69233/style-request-historical-social-research#latest

### DIFF
--- a/historical-social-research.csl
+++ b/historical-social-research.csl
@@ -61,17 +61,6 @@
       </if>
     </choose>
   </macro>
-  <macro name="editor">
-    <names variable="editor">
-      <name name-as-sort-order="first" and="text" sort-separator=", " delimiter=", " delimiter-precedes-last="always"/>
-      <label form="short" prefix=", "/>
-    </names>
-  </macro>
-  <macro name="translator">
-    <names variable="translator">
-      <name name-as-sort-order="first" and="text" sort-separator=", " delimiter=", " delimiter-precedes-last="always"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
It looks like from forum comments that  we still need the "at" thing for some item types? -- I'm currently a bit confused on this, might require either going back to styleguide or asking requester for clarification.

historical-social-research.csl
```diff
@@ -0,0 +1,628 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="display-and-sort" page-range-format="chicago">
+  <!-- Polyglot; journal publishes in English and German -->
+  <info>
+    <title>Historical Social Research</title>
+    <title-short>HSR</title-short>
+    <id>http://www.zotero.org/styles/historical-social-research</id>
+    <link href="http://www.zotero.org/styles/historical-social-research" rel="self"/>
+    <link href="http://www.zotero.org/styles/historical-social-research" rel="template"/>
+    <link href="http://www.chicagomanualofstyle.org/tools_citationguide.html" rel="documentation"/>
```
rmzelle: @POBrien333, do you remember which Chicago variant you used here? The "template" link isn't right. The "documentation" link should also be updated (probably to https://www.gesis.org/en/hsr/faqs/faqs-for-authors/guidelines/).

PR #3285, opened 2018-01-23
Added style for L'homme - revue française d'anthropologie.

Marcool04:
Hello all,
I have taken the liberty to create what I believe is a relatively adequate copy of the citation style for one of the most renowned French language anthropology journals : _L'Homme_. My university in France uses this as a requirement for a number of our papers, and I wanted to offer it here for inclusion in the repository, if it seems acceptable. It has been based largely on the style for other French ethno/anthropo-logy journal _Ethnologie française_, so my acknowledgements go out to Viera Rebolledo-Dhuin for authoring that one.
For more information on the journal see http://www.editions.ehess.fr/revues/lhomme/
For an example article : http://www.persee.fr/doc/hom_0439-4216_1999_num_39_152_453664
(newer issues follow similar typography, but are behind paywalls).

### DIFF
--- a/l-homme.csl
+++ b/l-homme.csl
@@ -114,7 +114,7 @@
       <if is-numeric="edition">
         <group delimiter=" ">
           <number variable="edition" form="ordinal"/>
-          <text term="edition" form="short" suffix=", " text-case="lowercase"/>
+          <text term="edition" form="short" text-case="lowercase"/>
         </group>
       </if>
       <else>
@@ -154,18 +154,14 @@
       <key variable="page-first"/>
       <key variable="title"/>
     </sort>
-    <layout suffix="">
-      <group display="block">
-        <text macro="author" suffix=" "/>
-      </group>
+    <layout suffix=".">
+      <text macro="author" suffix=" " display="block"/>
       <group display="left-margin">
         <text macro="year-date" suffix=" "/>
         <choose>
           <if type="thesis">
-            <group suffix=".">
-              <text macro="title" font-style="italic"/>
-            </group>
-            <group delimiter=", " suffix=".">
+            <text macro="title" font-style="italic" suffix="."/>
+            <group delimiter=", ">
               <text variable="genre"/>
               <text variable="publisher"/>
               <text variable="publisher-place"/>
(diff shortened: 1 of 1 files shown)


l-homme.csl
```diff
@@ -0,0 +1,221 @@
...
+    </sort>
+    <layout suffix="">
+      <group display="block">
+        <text macro="author" suffix=" "/>
+      </group>
+      <group display="left-margin">
+        <text macro="year-date" suffix=" "/>
+        <choose>
+          <if type="thesis">
+            <group suffix=".">
```
adam3smith: no need for groups with a single element in them

l-homme.csl
```diff
@@ -0,0 +1,221 @@
...
+              <text variable="publisher"/>
+              <text variable="publisher-place"/>
+              <group>
+                <text variable="number-of-volumes" suffix="&#160;vol."/>
+                <text variable="number-of-pages" suffix="&#160;p."/>
+              </group>
+            </group>
+          </if>
+          <else-if type="webpage">
+            <group suffix=". ">
```
adam3smith: ditto

l-homme.csl
```diff
@@ -0,0 +1,221 @@
...
+              <text macro="title" font-style="italic"/>
+            </group>
+            <group>
+              <text variable="URL" prefix=", "/>
+              <text prefix=" (" macro="access" suffix=")"/>
+            </group>
+          </else-if>
+          <else-if type="article-journal article-magazine article-newspaper broadcast personal_communication thesis entry-dictionary entry-encyclopedia" match="any">
+            <group suffix=". ">
```
adam3smith: ditto

l-homme.csl
```diff
@@ -0,0 +1,221 @@
...
+    </layout>
+  </citation>
+  <bibliography subsequent-author-substitute="" entry-spacing="0">
+    <sort>
+      <key macro="author"/>
+      <key variable="issued"/>
+      <key variable="page-first"/>
+      <key variable="title"/>
+    </sort>
+    <layout suffix="">
```
adam3smith: I'd set `suffix="."` here and then remove it from all the individual item types. Simplifies things a good bit.

l-homme.csl
```diff
@@ -0,0 +1,221 @@
...
+              <text variable="URL" prefix=", "/>
+              <text prefix=" (" macro="access" suffix=")"/>
+            </group>
+          </else-if>
+          <else-if type="article-journal article-magazine article-newspaper broadcast personal_communication thesis entry-dictionary entry-encyclopedia" match="any">
+            <group suffix=". ">
+              <text macro="title" quotes="true"/>
+            </group>
+            <group>
+              <text variable="container-title" font-style="italic"/>
```
adam3smith: I'd just set a delimiter for this group to `", "` and then move the pages macro out of the group.

l-homme.csl
```diff
@@ -0,0 +1,221 @@
...
+              <text macro="title"/>
+            </group>
+            <group delimiter=", " suffix=". ">
+              <text macro="edition"/>
+              <text macro="publisher"/>
+            </group>
+          </else-if>
+          <else-if type="chapter paper-conference" match="any">
+            <text macro="title" quotes="true" suffix=". "/>
+            <group>
```
adam3smith: Similarly here: put "in" and "editor" in a group with `delimiter=" "`, set the delimiter for this whole group to `", "` and move pages out of the group. Will make this more robust for missing variable.

l-homme.csl
```diff
@@ -0,0 +1,221 @@
...
+        <text term="no date" form="short"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="edition">
+    <choose>
+      <if is-numeric="edition">
+        <group delimiter=" ">
+          <number variable="edition" form="ordinal"/>
+          <text term="edition" form="short" suffix=", " text-case="lowercase"/>
```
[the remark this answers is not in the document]
Marcool04: Indeed! Correct behavior requires the suffix in line 121. This, interestingly, is a bug inherited from the same element in the the base I used, [ethnologie française](https://github.com/citation-style-language/styles/blob/master/ethnologie-francaise.csl), [line 119](https://github.com/citation-style-language/styles/blob/f601d1af5d642b1a67fa066a948339472529c6f6/ethnologie-francaise.csl#L119).

l-homme.csl
Marcool04: Quite right…

l-homme.csl
Marcool04: +1

l-homme.csl
Marcool04: +2

l-homme.csl
Marcool04: Yes that is much better than my prefix=", blabla" attributes :/

l-homme.csl
Marcool04: Yup. Done.

l-homme.csl
```diff
@@ -0,0 +1,221 @@
...
+      <term name="no date" form="short">s.&#160;d.</term>
+      <term name="translator" form="short">trad.</term>
+      <term name="editor" form="short">dir.</term>
+      <term name="chapter" form="short">§</term>
+    </terms>
+  </locale>
+  <macro name="author">
+    <names variable="author" delimiter=", ">
+      <name and="text" name-as-sort-order="all" sort-separator=" " delimiter=", ">
+        <name-part name="family" suffix=", "/>
```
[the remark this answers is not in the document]
Marcool04: Indeed I don't understand properly, forgive me, I have very little experience with csl… What you're saying is that there is a "default" formatting for the author field? But what about the "substitute" that is set to "editor" here, is that also default?
I think I understand what you're saying, I'm just not clear which part of that code could be removed for the same result…

adam3smith:
Thanks. I fixed the names for you. How names are displayed depends on the combination of `name-as-sort-order`, `sort-separator`, and `initialize-with`. You don't need the name parts for any of that. Those only need to be put in specifically if you need to do things like uppercase or bold last names, for example.

Marcool04:
Great to see this merged :) Thanks for your patience in explaining all the details to me, and for the corrections.
Regards,
Mark.

PR #3289, opened 2018-01-25
Create letters-in-applied-microbiology.csl

POBrien333:
via https://forums.zotero.org/discussion/69476/letters-in-applied-microbiology-style#latest


adam3smith:
Thanks!

PR #3290, opened 2018-01-25
Update springer-basic-note.csl

POBrien333:
https://forums.zotero.org/discussion/comment/299234/#Comment_299234

### DIFF
--- a/springer-basic-note.csl
+++ b/springer-basic-note.csl
@@ -17,7 +17,7 @@
     <category citation-format="note"/>
     <category field="law"/>
     <summary>Springer note style for law and related disciplines.</summary>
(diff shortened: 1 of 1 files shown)


adam3smith:
@bwiernik do you know if the examples in comments are still useful? I don't want to throw out useful comments, but if these refer to a different Springer style, we might as well drop them.

bwiernik:
I might leave the comments on lines 11 and 80–these came from feedback from the requester rather than documentation anywhere. Other examples are from Springer Basic and can be dropped.

adam3smith:
Thanks -- sorry this took so long!

PR #3292, opened 2018-01-25
Update journal-of-wildlife-diseases.csl

woylunk:
Remove space between author initials as per journal formatting.


adam3smith:
Thanks! I also removed the period since the initials are AB not A.B.

PR #3293, opened 2018-01-26
Create henoch.csl

POBrien333:
via https://forums.zotero.org/discussion/70070/style-request-henoch#latest


adam3smith:
Thanks. Note the option I deleted from the locale section -- that caused the punctuation outside quotation marks

POBrien333:
Ah cool. Thanks for fixing that!

PR #3297, opened 2018-01-29
Update springer-vancouver-brackets.csl

POBrien333:
VIA https://forums.zotero.org/discussion/70132/current-infectious-dis-rep-and-springer-vancouver-styles-suppressing-periods-after-journal-abbrev

see manual p.12


adam3smith:
thanks!

PR #3298, opened 2018-01-30
Update development-and-change.csl

POBrien333:
Was emailed that the style has the wrong et-al settings for the bibliograpy.
I've double check with the guidelines where it states:
"list all the authors in the References (unless there are more than five)."


adam3smith:
thanks

PR #3300, opened 2018-01-31
Create the-optical-society-express

scycotic:
Similar to OSA, for Biomedical Optics Express, Optics Express, Optics Materials Express. Adapted from csl file provided for OSA by Barbara Williams and Theresa Miller (which I do not see in this depository either). Some errors were corrected in addition to making the changes required (issue number to be included, other OSA journals do not require this).


adam3smith:
Do you have a link to the OSA style you adapted this from?

scycotic:
Yes. From here: https://www.osapublishing.org/submit/templates/default.cfm
More specifically, the .csl file they provide is here:
https://www.osapublishing.org/submit/templates/wordwp/the-optical-society.csl
But their file is not listed in this repository nor in Mendeley's. I tried
setting their file as the template in the XML code but it was rejected
since it wasn't in the repository. There were a few errors (extra spaces),
and their "Express" journals also require the issue number which their
template does not include.

Express journals guidelines:
https://www.osapublishing.org/submit/style/oestyleguide.cfm

First time modifying citation list styles but a fellow student needed it
for an article he was submitting so I figured I may as well submit it here
since it's done.

Thanks,


Lawrence Yip

adam3smith:
Great, thanks. I wanted to make sure there was no proprietary claim on that file, but it has a CC license like all ours, so that's no problem. I'll review this as quickly as possible. We're a bit short-staffed, so might take up to a week, but I'll try to be faster than that.

rmzelle:
(the The Optical Society style at https://www.osapublishing.org/submit/templates/wordwp/the-optical-society.csl has a timestamp of 2017-04-03T22:54:26+00:00, and has clearly been derived from https://github.com/citation-style-language/styles/blob/master/the-optical-society.csl, which was last updated in 2015)

scycotic:
I admit I did not see the 2015 version you mentioned when I searched the repository earlier. However, the reason I was submitting this is still relevant because OSA's "Express" journals (Optics Express, Biomedical Optics Express, Optical Materials Express) have a slightly different reference format to the rest of OSA's journals. If you look at all the style guides at: https://www.osapublishing.org/submit/templates/default.cfm , you will see that these three journals require the issue number while the rest do not.

rmzelle:
Thanks. (I just created three styles for the different Express journals, one "independent" one that specifies the format and two "dependent" ones that rely on the independent style)

Issue #3268, opened 2018-01-16
Modern Pathology & other Nature journals


adam3smith:
Appear to follow a different style:
https://twitter.com/ModernPathology/status/953341659781230593

PR #3302, opened 2018-02-01
Update and rename dependent/modern-pathology.csl to modern-pathology.csl

POBrien333:
via #3268 (superseeds #3301)


adam3smith:
Thanks! Did you check if that is the same style as Lab Invest: https://www.nature.com/labinvest/ ? 
If so, I think we'd want to create nature-vancouver.csl with dependents.

POBrien333:
@adam3smith 
It's not the same style.
lab invest: et al. is in italics. and dot after citation-number in bibliography at a quick glance.

PR #3303, opened 2018-02-01

roaldfrosig:
Changing the citation output for legislation from law number to law title.


rmzelle:
Thanks!

PR #3305, opened 2018-02-02
Updates to sage-harvard.csl to conform more closely with style guide

tipichris:
This PR makes the following changes to sage-harvard.csl to conform more closely to the published style guide (https://uk.sagepub.com/sites/default/files/sage_harvard_reference_style_0.pdf).

* Website location should be 'Available at', not 'Available from'
* There is no indication in the style guide that website title should be included
* Thesis titles should be italic
* The place of publication of a thesis is formatted differently to other document types
* Blog titles should be preceded by 'In: '. (Note the style guide seems to indicate no italics, but as this is inconsistent with all other uses of In: I've assumed it's a typo and kept italics)
* Full dates are included for reports

### DIFF
--- a/sage-harvard.csl
+++ b/sage-harvard.csl
@@ -12,7 +12,7 @@
     <category citation-format="author-date"/>
     <category field="generic-base"/>
     <summary>The SAGE Harvard author-date style</summary>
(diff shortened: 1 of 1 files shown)


adam3smith:
Great, thanks!

PR #3307, opened 2018-02-04
Create SoRo.csl

debkalpa:
New addition for Soft Robotics journal


rmzelle:
@debkalpa, thanks! Sorry for the long wait.

PR #3310, opened 2018-02-05

roaldfrosig:
Changing legislation title in citations from italics to plain text.


rmzelle:
Thanks!

PR #3313, opened 2018-02-06
Updates to sage-harvard.csl

tipichris:
This PR makes a number of further changes to sage-harvard.csl.

* The handling of conferences has been improved, including using the event title if no proceedings title is present, not printing the 'In: ' if there is nothing to follow it, different form of listing editors to books, and including the date
* Some punctuation improvements, e.g. no comma following the list of editors of edited books; collection titles are followed by a full stop
* Added DOIs. The style guide specifies these only for online articles ahead of print, but they have started to be included more generally in some recent SAGE publications.

### DIFF
--- a/sage-harvard.csl
+++ b/sage-harvard.csl
@@ -29,6 +29,12 @@
       <label form="short" prefix=" (" suffix=")"/>
     </names>
   </macro>
+  <macro name="proceedings-editor">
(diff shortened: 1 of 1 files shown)


adam3smith:
This one is a bit harder to review without extensive testing, but best I can tell this all makes sense -- thanks!

PR #3317, opened 2018-02-07
Create MLA8-Annotated-Extra.csl

madduxl:
Uses the Extra field for user-generated annotation

### DIFF
--- a/MLA8-Annotated-Extra.csl
+++ b/MLA8-Annotated-Extra.csl
@@ -1,0 +1,295 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" page-range-format="minimal-two">
+  <info>
+<title>Modern Language Association 8th edition (Annotated, Extra Field)</title>
+    <title-short>MLA</title-short>
+    <id>http://www.zotero.org/styles/modern-language-association-annotated</id>
+    <link href="http://www.zotero.org/styles/modern-language-association" rel="self"/>
+    <link href="http://style.mla.org" rel="documentation"/>
+    <author>
+      <name>Sebastian Karcher</name>
+    </author>
+    <contributor>
+      <name> Keith Stanger</name>
+      <email>keith@stanger.com</email>
+    </contributor>
+    <category citation-format="author"/>
+    <category field="generic-base"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
Hi -- thank you for submitting these, but we're generally not accepting such variants of styles beyond what already exists. You're very welcome to host them on your own website. In fact, Keith Stanger at Eastern Michigan maintains a list of styles for annotated bibliographies on his page: https://guides.emich.edu/c.php?g=188045&p=1241422

madduxl:
Yes, but those styles are convoluted to add to Zotero, so we were trying to simplify for our users. WE will come up with some other way. Thanks!

L.

dconner3:
I request that you reconsider your decision not to add this style to the repository. Creating an annotated bibliography is one of the main reasons to use Zotero. Many existing annotated styles in the Zotero Repository require use of the abstract field, but that is illogical. An article abstract and annotation are not the same thing. Zotero users often need to keep the author-created abstract while generating their own annotation.  Users should not need to use a complicated workaround in order to generate annotated bibliographies from a citation manager tool.

adam3smith:
So here are the reasons against this (and why I have asked Keith several years back when he started to self-host these):
1) Creating annotated bibliographies is certainly a valid use of a reference manager, but having answered >20,000 user questions on reference managers, I'm positive it's *not* one of the main reasons.
2) There's no limit to the number of annotated bibliography styles on can come up with. Almost any style qualifies -- the fact that you're suggesting both 7th and 8th edition of MLA is pretty solid evidence of this. We only grudgingly created the 2 available ones. Keith has 8 on his site, adding those and the ones you're suggesting beyond them would get this to >10 with really no end in sight. We're a small group of volunteers and just can't keep up with them all. It's already quite tedious to make updates to popular styles like APA with 5 sub-styles.
3) The "note" field gets used differently by different reference managers. Using it for annotated bibliographies will create weird outcomes -- CSL is not just for Zotero 
4) The abstract version was created based on user requests, so it's certainly not illogical; people use their tools differently
5) My understanding is that the need for this is caused by the fact that linking to the custom styles is somehow broken in LibGuides? That, frankly, is a LibGuide issue, not ours. Even if you aren't able to serve CSL styles with a dedicated mimeType (which would allow for 2-click install just like on the Zotero repository), custom styles take exactly one extra click to install in Zotero (one to download, one to open, one to confirm installing). I don't think that's a complicated workaround by any stretch.

PR #3320, opened 2018-02-08
SAGE Harvard

tipichris:
More tweaks to SAGE-Harvard.csl

* suppress genre on websites. The connector often inserts a Website type of 'Text' into the database, which is not required in the output
* ensure numbers of reports etc are included
* sort order tweak addressing unexpected behaviour when two citations from the same author in the same year
* have magazines behave the same way as newspapers

Pretty much finished revising that paper, so this might be it now :)

### DIFF
--- a/sage-harvard.csl
+++ b/sage-harvard.csl
@@ -144,6 +144,13 @@
       </if>
     </choose>
   </macro>
+  <macro name="genre">
+    <choose>
+      <if type="webpage post-weblog" match="none">
+        <text variable="genre"/>
+      </if>
+    </choose>
+  </macro>
   <macro name="pages">
     <choose>
       <if type="chapter" match="any">
@@ -210,7 +217,7 @@
             <text variable="container-title" font-style="italic"/>
           </group>
           <choose>
-            <if type="article-newspaper" match="any">
+            <if type="article-newspaper article-magazine" match="any">
               <date variable="issued">
                 <date-part name="day" suffix=" "/>
                 <date-part name="month" form="long"/>
@@ -242,6 +249,7 @@
       <key macro="author-count" names-min="3" names-use-first="3"/>
       <key macro="author" names-min="3" names-use-first="1"/>
       <key macro="year-date"/>
+      <key variable="citation-number" sort="ascending"/>
       <key variable="title"/>
     </sort>
     <layout suffix=".">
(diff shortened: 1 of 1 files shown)


adam3smith:
Are you sure about the sort-order tweak? Why would you want items in the bibliography in the order in which they appear in the text rather than alphabetically by title? Do we have an example of this being handled like this in a journal?
This is rarely addressed in style guides, but those that do, like APA, keep bibliography sorting consistent.

tipichris:
It isn't specifically addressed in the style guide. What there is is an example of citations from the same author and year:
> 5. If there is more than one reference to the same author and year, insert a, b, etc. in both the text and the list:
> ... it was described (Author Last Name, yeara, yearb) ...

It's not explicit, but the implication seems to be that the 'a', 'b', 'c' should appear in the text in alphabetical order. The problem that I had was two references, same author, same year. The one that I cited first had a title that came second. So the reader encountered a citation for Bloggs 2016b before they encountered Bloggs 2016a. This seemed very odd, and I'm not sure I've ever read anything in which it happens.

Arguably from a reference list perspective this change is not inconsistent, since the sort order is author, year, title and the 'a' and 'b' form part of the year. It's an edge case, so would probably take a fair bit of searching to find an example. But if the paper I just revised gets accepted I guess we'll find out :)

tipichris:
Actually, there is an explicit example that could cause the same issue
> Et al. references in alphabetical order; in the event of more than one entry having the same date, they should be placed in alphabetical order of second (or third) author, and a, b, etc. must be inserted.

If feels slightly different, because the second author name appears before the date / letter combination, where as the title appears after. But it has the potential to have the same effect on the order of year suffixes in the main text, so it does at least suggest that such ordering is acceptable.

OK, what's the best way to deal with this? Should I make the change and push it to my branch?

adam3smith:
So I agree having Author (2000b, 2000a) is indeed weird, but I think having Author (2000b) yadda, yadda, Author (2000a) is perfectly fine. 

I don't buy that sorting by the year suffix in the bibliography is a thing because the year suffix is just a random artifact of where the item appears in the text, i.e. if I just look through at bibliography I can't predict it. This is no problem with a and b, but if you're thinking of a-h (think newspaper articles, say), you want these in a logical order in the bibliography.

To prevent the  wrong order of a/b in a single citation, i.e. the first case, let's just add a sort by title as the third key for citations.

adam3smith:
So for what to do, I'd remove the citation-number sort in the bibliography and add a title sort key as the final condition for citations.

tipichris:
Change done as requested

PR #3325, opened 2018-02-14
Update sodertorns-hogskola-harvard-ibid.csl

Ca7w3as3l:
Feb update
ibid
    et al.
    4 author correction


adam3smith:
Thanks! Note that you're commit message said something about ibid, but you didn't actually change any ibid-related settings for the style.

PR #3326, opened 2018-02-14
Change chinese-gb7714-* to fit the standard better.

ukyoi:
I made two changes. 

First, and the most important, **et-al-use-first** is ought to be **3** according to the standard, then I changed **et-al-min** to **4** to fit it.

Second, I removed **default-locale** in files, since the standard specified strings such as "et al." should fit the native language of the original documents.

I'm not sure whether there is a proper way to process mixed-language references in the same thesis. So please tell me if I made anything wrong.


adam3smith:
Thanks for the et-al fix. There's no way for mixed-language bibliographies in CSL at this time, so we're going to leave the default-locale as is.

PR #3328, opened 2018-02-14
Update bulletin-of-the-american-schools-of-oriental-research.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/301283/#Comment_301283


adam3smith:
I tested this & looks good to me. Spacing is a mess across different applications with the display attributes, but the style is solid as is.

POBrien333:
thanks :)

PR #3330, opened 2018-02-14
Fixed style for Acta Orthopaedica - month + removed access + URL

gforge:
The Acta Orthopaedica doesn't require months in the reference list. They also don't use the access/URL parts. Checked with one of the editors.


adam3smith:
Thanks & sorry for the delay!

PR #3333, opened 2018-02-15
Update American Society for Agronomy style

jasonjensen:
Fix accessed field and add doi field


adam3smith:
Thanks and sorry for the delay

PR #3336, opened 2018-02-15
small fixes for locator macro

alexo81:
- Added conditional for locator macro within article-journal for including issue if present.
- Also added conditional for including URL if article had ISSN or URL present.
- Changed from ordinal to numeric edition number
- Changed value "acceso" to term access in locators.
- Changed pages label for contextual plural.


alexo81:
Added some minor fixed I've found checking the outcome to the examples provided in journal's authors guidelines

PR #3338, opened 2018-02-15
Create serdica-journal-of-computing.csl

valeriina:
A mixed model of Elsevier and Chicago F styles


rmzelle:
@valeriina, thanks! Sorry for the long wait.

PR #3323, opened 2018-02-13
Create quaderni-materialisti.csl

POBrien333:
via https://forums.zotero.org/discussion/69875/how-to-cite-a-book-with-its-translation#latest

PR #3342, opened 2018-02-19
Update universita-di-bologna-lettere.csl

POBrien333:
as spotted by @adam3smith in another style based on this one in #3323


rmzelle:
Thanks!

PR #3314, opened 2018-02-06
Create revue-anthropologie-et-societes.csl

POBrien333:
Closes #3278

PR #3343, opened 2018-02-19

POBrien333:
To go with #3314 (more emails from requestor to fix some mistakes)


rmzelle:
Thanks!

PR #3347, opened 2018-02-20
Add iso690-numeric-brackets-no-abstract-da

heidivanparys:
Add Danish ISO 690 style with square brackets (instead of round brackets) and no abstract. The style is based on iso690-numeric-en.

### DIFF
--- a/iso690-numeric-brackets-no-abstract-da.csl
+++ b/iso690-numeric-brackets-no-abstract-da.csl
@@ -5,7 +5,7 @@
     <id>http://www.zotero.org/styles/iso690-numeric-brackets-no-abstract-da</id>
     <link href="http://www.zotero.org/styles/iso690-numeric-brackets-no-abstract-da" rel="self"/>
     <link href="http://www.zotero.org/styles/iso690-numeric-en" rel="template"/>
-    <link href="https://github.com/citation-style-language/styles/pull/3347" rel="documentation"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
Is there any style guide this is based on (ISO 690 is pretty vague)? We're a little wary of the proliferation of ISO 690 styles unless they correspond to something that's required/specified somewhere.

heidivanparys:
Yes there is, I updated the documentation link to https://www.iso.org/drafting-standards.html (also in #3348).

From https://www.iso.org/sites/directives/2016/part2/index.xhtml : 
> For other referenced documents and information resources (printed, electronic or otherwise), the relevant rules set out in ISO 690 shall be followed.

The documents on https://www.iso.org/drafting-standards.html provide more information on what method of citation to use (numeric, and square brackets in the citations and in the bibliography itself):

image

image

adam3smith:
Sorry for letting this linger, but I'm not quite sure what to do with it.
 We have all these "ISO 690" styles where the title has basically no meaning because ISO 690 is not really a citation style but rather a rough set of guidelines for references. I'd be happy to take a style specifically for the ISO Drafting Standards, which take one particular implementation of ISO 690, but I'd rather only take that either in one language (English presumably) or as a multilingual style. 
Alternatively, what are you actually using the style for if it's not for writing ISO standards?

PR #3349, opened 2018-02-21
Update the-american-journal-of-psychiatry.csl

POBrien333:
via https://forums.zotero.org/discussion/70516/style-error-the-american-journal-of-psychiatry
Didn't have access to the journal, so just went by the guidelines. There was some more mistakes for this style, esp. re the chapters. Would be great ifyou could check on it.


adam3smith:
Thanks -- checked against recent issue & looks good.

PR #3352, opened 2018-02-22
Use explicit year-suffix + minor code cleanup

fbennett:
The style Deutsche Gesellschaft für Psychologie (German) lacked an explicit year-suffix. This just adds it, and streamlines the date code a little. I haven't run it past the user for testing.

Issue was raised here:

https://forums.zotero.org/discussion/70521/disambiguation


adam3smith:
thanks!

PR #3355, opened 2018-02-23
Fix stray label issue

fbennett:
This addresses the SBL style issue here: https://forums.zotero.org/discussion/70547/sblhs-unexpected-output


adam3smith:
Thanks!

PR #3356, opened 2018-02-23
Update clinical-infectious-diseases.csl

libora:
https://forums.zotero.org/discussion/comment/302066#Comment_302066


adam3smith:
Thanks!

PR #3358, opened 2018-02-24

cookie365:
changed email address
removed prefix and suffix in citation

### DIFF
--- a/fachhochschule-vorarlberg-note.csl
+++ b/fachhochschule-vorarlberg-note.csl
@@ -23,12 +23,12 @@
     </contributor>
     <contributor>
       <name>Jasmin Zimmermann</name>
-      <email>zim3610@gmail.com</email>
+      <email>zim32079@gmail.com</email>
     </contributor>
     <category citation-format="note"/>
     <category field="generic-base"/>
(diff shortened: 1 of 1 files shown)


fachhochschule-vorarlberg-note.csl
```diff
@@ -383,7 +383,7 @@
     </choose>
   </macro>
   <citation et-al-min="4" et-al-use-first="1" et-al-subsequent-min="4" et-al-subsequent-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year">
-    <layout prefix="(" suffix=")" delimiter="; ">
+    <layout delimiter="; ">
```
rmzelle: @POBrien333, could you check if this change makes sense to you? As a note style, I guess parentheses aren't used, but I didn't check the style guide.

cookie365: Hi, in the FHV Style Guide on page 17 are examples, where you can see that there are no parentheses. I have marked them.

[FHV_Style_Guide.pdf](https://github.com/citation-style-language/styles/files/1817332/FHV_Style_Guide.pdf)

adam3smith:
Yes, this is right; thanks!

PR #3353, opened 2018-02-22
Create universite-catholique-de-louvain-histoire.csl

pbieswal:
Documentation site link added

PR #3360, opened 2018-02-26
Create universite-catholique-de-louvain-histoire.csl

pbieswal:


adam3smith:
Let's keep this in the original #3353

PR #3362, opened 2018-02-26
Update journal-of-tropical-life-science.csl

jtls:
Journal of Tropical Life Science added some new citation styles (ie: book, chapter in a book)


adam3smith:
Thanks -- I fixed this up some more following the examples on http://www.jtrolis.ub.ac.id/index.php/jtrolis/about/submissions#authorGuidelines
please take a look and let me know if the changes look OK

rmzelle:
@adam3smith, I assume we can just go ahead and merge this? Per above, fixed a bunch of errors that prevented validation.

rmzelle:
Thanks both!

PR #3365, opened 2018-03-05
first edition can be indicated with original-date

scarwire:
Added possibility for original publication date for books (yaml variable `original-date`) between [brackets].
Le style prend désormais en charge la date de première édition des livres (variable yaml `original-date`), rendue entre [crochets] après la date de publication.

See also [this issue](https://github.com/retorquere/zotero-better-bibtex/issues/922).

I also fixed a few little bugs having to do with nonbreaking spaces.
Quelques petites corrections aussi par rapport aux espaces insécables.

### DIFF
--- a/le-tapuscrit-note.csl
+++ b/le-tapuscrit-note.csl
@@ -13,7 +13,7 @@
     <category citation-format="note"/>
     <category field="social_science"/>
     <category field="generic-base"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
Sorry, this should have gotten reviewed a long time ago. I disagree with the changes removing NBSP, but original date should be in there -- I'll add a new commit, that's going to be faster.

PR #3369, opened 2018-03-06
Zeitschrift für Antikes Christentum

carljohanberglund:


carljohanberglund:
@rmzelle They differ in a few ways: Subsequent citations refer back to the number of the note where the first citation is located; editors, editions and series should be put within the same parentheses as the city and publisher; full page range for articles goes in a separate parenthesis rather than separated by a comma. So, both are needed.

rmzelle:
@carljohanberglund, thanks. Did you test this style with both English and German, or just English? I'm wondering whether we should fix the style to a single language.

The CSL style currently also has some French localization, so we should at least remove that (I assume the journal doesn't publish in French).

And do you know what CSL style you based this one on?

carljohanberglund:
I don’t write German, so I’ve only tested it in English. The journal publishes articles in German, English, French or Italian, concluded by either an English or German abstract.

I based it on my previous contribution, whichever that was.

Best regards,
Carl Johan

rmzelle:
Okay, thanks! I fixed the style to US English for now, then, and added the English journal name to the style title.

PR #3370, opened 2018-03-06
Journal of Early Christian Studies

carljohanberglund:


rmzelle:
Thanks! I fixed the style to US English since it doesn't appear to publish in other languages.

PR #3372, opened 2018-03-06
Create berghahn-books.csl

POBrien333:
via https://forums.zotero.org/discussion/70712/modifying-style-development-and-change


adam3smith:
@rmzelle if you check here: https://www.berghahnbooks.com/authors/ they actually have a US and a UK documentation (=references) guide and both have a "Harvard" and a note option. This is the UK "Harvard" one. What do you want to do about naming?

rmzelle:
> What do you want to do about naming?

I would use this (https://github.com/citation-style-language/styles/pull/3372/commits/86c0a519e651f54bcb384236b6378f07cc652e2d), which seems most consistent which what we've done in the past for such cases. Feel free to merge.

POBrien333:
Are we ready to merge this one?

rmzelle:
Thanks!

PR #3376, opened 2018-03-08
Create nature-sciences-societes.csl

POBrien333:
https://forums.zotero.org/discussion/70728/style-request-natures-sciences-societes#latest

### DIFF
--- a/nature-sciences-societes.csl
+++ b/nature-sciences-societes.csl
@@ -1,0 +1,301 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+  <info>
+    <title>Natures Sciences Sociétés</title>
+    <title-short>NSS</title-short>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks!

nature-sciences-societes.csl
```diff
@@ -0,0 +1,302 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+  <info>
+    <title>Natures Sciences Sociétés</title>
+    <title-short>NSS</title-short>
+    <id>http://www.zotero.org/styles/nature-sciences-societes</id>
```
rmzelle: Style ID is missing an "s" ("nature" should be "natures"). Will fix this up.

rmzelle: Fixed in https://github.com/citation-style-language/styles/pull/3420.

rmzelle:
@POBrien333, thanks! By the way, I assume you just checked this style for English for this bilingual journal? Or does it work for French as well?

PR #3378, opened 2018-03-10
Fix year-suffix delimiters etc

fbennett:
This fixes several issues with a contribution I made a couple of weeks ago, reported by the requesting user.

ref: https://forums.zotero.org/discussion/63141/zitierstil-deutsche-gesellschaft-fuer-psychologie#latest


adam3smith:
Thanks -- and sorry, I should have spotted that during code review

PR #3382, opened 2018-03-12
Create oecd.csl

POBrien333:
https://forums.zotero.org/discussion/70545/oecd-referencing-style#latest

### DIFF
--- a/oecd.csl
+++ b/oecd.csl
@@ -1,0 +1,293 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB" xmlns="http://purl.org/net/xbiblio/csl">
+  <info>
+    <title>OECD </title>
+    <title-short>OECD </title-short>
+    <id>http://www.zotero.org/styles/oecd</id>
+    <link href="http://www.zotero.org/styles/oecd" rel="self"/>
+    <link href="http://www.zotero.org/styles/deutsche-gesellschaft-fur-psychologie" rel="template"/>
+    <link href="https://www.oecd.org/site/eduimhe08/guidelinesforauthors.htm" rel="documentation"/>
+    <author>
+      <name>Patrick O'Brien</name>
+      <email>obrienpat86@gmail.com</email>
+    </author>
+    <category citation-format="author-date"/>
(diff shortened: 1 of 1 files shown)


oecd.csl
```diff
@@ -0,0 +1,293 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB" xmlns="http://purl.org/net/xbiblio/csl">
+  <info>
+    <title>OECD</title>
```
rmzelle: @POBrien333, we'll probably want to rename this to "Organisation for Economic Co-operation and Development".

oecd.csl
```diff
@@ -0,0 +1,293 @@
...
+    <title-short>OECD </title-short>
+    <id>http://www.zotero.org/styles/oecd</id>
+    <link href="http://www.zotero.org/styles/oecd" rel="self"/>
+    <link href="http://www.zotero.org/styles/deutsche-gesellschaft-fur-psychologie" rel="template"/>
+    <link href="https://www.oecd.org/site/eduimhe08/guidelinesforauthors.htm" rel="documentation"/>
```
rmzelle: @POBrien333, do you know if the style matches what is described in https://www.iea.org/publications/freepublications/publication/IEAOECD_style_guide.pdf. That seems to be a much better documentation link (that style guide seems to be both for the IEA and the OECD, by the way).

The current link is for a 2008 Institutional Management in Higher Education (IMHE) conference, and the IMHE program seems to have been canceled in 2016 (see http://www.oecd.org/education/imhe/: "In November 2015, the IMHE Governing Board agreed not to renew the programme's mandate which expired on 31 December 2016. As a result, the IMHE programme closed at the end of 2016.").

adam3smith: @POBrien333 see Rintze's questions above, please

POBrien333: Hi. I missed this one.
The two guidelines seem to be matching, but there is one small difference: in-text et al. is in italics for the "OECD" guidelines.

Since IMHE is dead now anyway and the IAE guidelines are way better documented, let's make the style for that then.

rmzelle:
Love the new style title. Very succinct and poetic. 😂

(it's probably just what we should call the style, though)

PR #3383, opened 2018-03-13

jjrondeau:
This new version is the official UQAM style developped by the UQAM Library Services. It is compliant with the new UQAM guidelines for the submission of dissertations and theses that were published September 2017. This version is intended for users of Zotero and Mendeley.


rmzelle:
Thanks!

PR #3385, opened 2018-03-15

ibadger:
Improved handling of editions, if the data contains words and not just a number. Fix a couple of minor problems: Adds missing full stop after conference event field. Titles appearing at the start of references when there is no author are now in italics.


adam3smith:
Thanks!

PR #3390, opened 2018-03-21
Create cambridge-journal-of-economics.csl

POBrien333:
via https://forums.zotero.org/discussion/70985/style-request-cambridge-j-econ-harvard#latest


adam3smith:
Thanks!

PR #3391, opened 2018-03-22
Update the-canadian-journal-of-chemical-engineering.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/299580/#Comment_299580


adam3smith:
Great, thanks!

PR #3393, opened 2018-03-24
fix the sort order of the bibliography

spearway:
There was a small problem in the sort order of the bibliography were the date was not taking into account


adam3smith:
apart from the sort order, you've also added a ton of other bits from the APA style -- was that on purpose?

rmzelle:
@adam3smith, https://onlinelibrary.wiley.com/page/journal/13652761/homepage/ForAuthors.html says "References should be prepared according to the Publication Manual of the American Psychological Association (6th edition).", so just making this a dependent of that style.

@spearway, thanks!

PR #3397, opened 2018-03-25
[CI] Test against Ruby 2.5

nicolasleger:


rmzelle:
@nicolasleger, thanks, but I don't think we'll take this.

I always try to use the newest Ruby version possible on Travis CI (since newer versions usually have performance improvements), but the Ubuntu Trusty images Travis CI uses doesn't yet have Ruby 2.5 preinstalled, and the installation of Ruby 2.5 will slow down the builds. (in the build of this pull request, installing Ruby 2.5 takes 23 seconds, whereas using the preinstalled Ruby 2.4.1 takes about 4 seconds; compare https://travis-ci.org/citation-style-language/styles/builds/358029616#L436 to https://travis-ci.org/citation-style-language/styles/builds/358096742#L436)

See also https://docs.travis-ci.com/user/reference/trusty/#Ruby-images

PR #3400, opened 2018-03-27

nchachereau:
This is a small update to a custom style for history students in Switzerland. See #234 for the original pull request, and #1675 for the last update.

Changes:
- name is now infoclio.ch
- Show DOIs as full URLs
- French: make abbreviations (et al., éds.) consistent with documentation
- whitespace clean-up

Thank you for taking the time to maintain the styles.


rmzelle:
Thanks! (also thanks for the clear pull request description!)

PR #3404, opened 2018-03-28
Update metallurgical-and-materials-transactions-a.csl

martok:
This patch removes an extra comma from article entries with no page reference. The comma is already provided by the enclosing `<group>`.


rmzelle:
Good catch. Thanks!

PR #3407, opened 2018-03-28
Fix for conditions with multiple type values without match="any"

### LINKED ISSUE #3405: Does `<if type="aaa bbb ccc">` without `match="any"` or `match="none"` make sense?
Hi,

Quick question:

I have noticed that some styles (around 85 of them) have multiple values in a `type` attribute without a `match="any"` or `match="none"`:

```xml
<if type="thesis manuscript personal_communication speech motion_picture">
```
Cf https://github.com/citation-style-language/styles/blob/master/ens-de-lyon-centre-d-ingenierie-documentaire.csl#L521

Would that ever evaluate to true though? My understanding is that the default `match` value is `all` and a reference can only have one type. (Cf https://github.com/citation-style-language/schema/blob/master/csl-data.json#L9)

Cheers

customcommander:
Fixes #3405

I found two other cases which are more difficult to fix. I didn't touch those as adding `match="any"` would perhaps not be what the authors of these styles meant. (Better to let them fix that IMO.)

Example [here](https://github.com/citation-style-language/styles/blob/8617a2b8fd3f7e0add85c6c75df1135d01c19aee/sodertorns-hogskola-harvard-ibid.csl#L112) and [here](https://github.com/citation-style-language/styles/blob/957c039ac5926df00c983e92e3166ff995d6083a/sodertorns-hogskola-harvard.csl#L112):

```xml
<if variable="URL" type="webpage post-weblog">
```

I also found these two other cases but I'm not entirely sure whether they are mistakes. (Just letting you know in case they are.)

In [university-college-lillebaelt-vancouver.csl](https://github.com/citation-style-language/styles/blob/5e38d2aabba638edabfb9603388ea3d425d63fd1/university-college-lillebaelt-vancouver.csl#L166):

```xml
<if type="legislation bill"/>
```

And in [uppsala-universitet-historia.csl](https://github.com/citation-style-language/styles/blob/8a1fc526de3e662520d93ce70dbce16b0cd259ce/uppsala-universitet-historia.csl#L328):

```xml
<if is-numeric="edition" type="book article"/>
```

### DIFF
--- a/anabases.csl
+++ b/anabases.csl
@@ -123,7 +123,7 @@
   </macro>
   <macro name="year-date">
     <choose>
-      <if type="webpage post-weblog">
+      <if type="webpage post-weblog" match="any">
         <text term="online" suffix=" "/>
         <date variable="issued" form="text"/>
       </if>
--- a/annals-of-joint.csl
+++ b/annals-of-joint.csl
@@ -68,7 +68,7 @@
   </macro>
   <macro name="access">
     <choose>
-      <if type="webpage post-weblog">
+      <if type="webpage post-weblog" match="any">
         <choose>
           <if variable="URL">
             <text variable="URL"/>
--- a/associacao-brasileira-de-normas-tecnicas-unirio-eipp.csl
+++ b/associacao-brasileira-de-normas-tecnicas-unirio-eipp.csl
@@ -215,7 +215,7 @@
   </macro>
   <macro name="genre">
     <choose>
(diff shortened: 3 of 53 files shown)


adam3smith:
Thanks so much!
Yes, the above examples are also all mistakes but fixing them (except for the lillebaelt one) requires nesting the conditions (in any order) so probably needs to be done by hand. If you want to handle those too, that'd be amazing, otherwise we can take those.

PR #3410, opened 2018-03-29
Create international-journal-of-computer-assisted-radiology-and-surgery.cls

dzenanz:


rmzelle:
@dzenanz, I'm sorry we never got back to you.

Do you remember why you submitted this style back in March? Did you find that the existing style for "International Journal of Computer Assisted Radiology and Surgery" (https://github.com/citation-style-language/styles/blob/master/dependent/international-journal-of-computer-assisted-radiology-and-surgery.csl), which uses the format of our "Springer - Basic (numeric, brackets)" style, was incorrect?

dzenanz:
I don't remember now. It could have been that I could not find it, or that it differed slightly from the described citation style.

rmzelle:
@POBrien333, would you mind adding this to your todo-list?

The question is whether the existing "International Journal of Computer Assisted Radiology and Surgery" style can stay with "Springer - Basic (numeric, brackets)" or whether it needs its own independent style (like the one provided here).

POBrien333:
Actually, just went through it.
This journal seems to print issues for article journal compared to the normal springer basic style.
There is however two mistakes in this style going by .pdfs of publications. No space between "volume(issue):pages". And dois are with https.
It would probably be worth taking the most recent springer-basic-brackets style as that one got an update in Sept-18.

rmzelle:
> It would probably be worth taking the most recent springer-basic-brackets style as that one got an update in Sept-18.

@POBrien333, sorry for reviewing this so slowly again. Are you suggesting to adapt the springer-basic-brackets style with the two differences you noticed? Do you have time to prepare that submission?

POBrien333:
Final check!
The style is just springer-basic-brackets and we already have it on the repo as a dependent Springer style.
In the guidelines they link to an Endnote style file, with the filename SpringerBasicNumber.zip.

Case closed.

rmzelle:
Great, thanks for checking.

PR #3419, opened 2018-04-03
Create chicago-fullnote-bibliography-tr

mtarakci:


mtarakci:
I could not understand what the problem is. Sorry.

mtarakci:
Changes: 
- Delimiter before last names were removed. 
- Editor and translator lebels (verbs) were transformed short ones (ed. for editors; çev. for translators)
- "in" term in chapter or paper conference was left null. 
- short dash was used for page-range-delimeter
- a colon after accessed term were added. 
---
And more notably, the titles of theses became the ones like books. I mean they are italik, and without quotation marks. This is much more common usage in Turkey. However, if you say it must be in quotes, it can be restored to its original.

adam3smith:
I think we've mentioned this before: Our criterion for accepting new styles is a written description of the style; in other words: 
If we receive a report that something in the style is wrong, are we able to check ourselves without reaching out to you?

Is the Turkish version of CMoS codified somewhere?

mtarakci:
Ok. Accept these corrections as feedback. 

a) In Turkish, among names, there is never a comma before "and". So, delimeter should be removed before last name. 
b) In Turkish, "in" term must be after "title", not before, or it should be removed. 
c) Long dash is almost never seen as page-range-delimeter. So it should be short in Turkish translation. 
d) Editor or translator lebels are never seen verbal. They are always abbreviated.
---
These are not particular or marginal interpretation, but the necessitates of a sound translation. So, if you like Zotero to have a stable Turkish version, and to reach to wider audience, I recommend these corrections be made, after twenty years of being in an academic environment in Turkey.

PR #3426, opened 2018-04-06
Officially supported by Institut Teknologi Bandung

ceefour:
This CSL is now officially supported by Institut Teknologi Bandung (I am currently assigned as the maintainer).


adam3smith:
Could you clarify why you're renaming the style? "SP" seems a lot less clear to me than "Tesis Magister" (and we generally try to avoid acronyms in style names).

ceefour:
@adam3smith SPs is the mother organization of all graduate programs in Institut Teknologi Bandung (everyone studying here knows it), including master and doctor. "Tesis Magister" only refer to the masters citation style, while now the styles are unified (used by both masters and doctoral).

rmzelle:
Thanks! (I replaced the "SPs" by the full "Sekolah Pascasarjana", since we avoid abbreviations in CSL repository style titles whenever feasible)

ceefour:
Thank you @rmzelle :)

PR #3432, opened 2018-04-10
Update deutsches-archaologisches-institut.csl

bellerophons-pegasus:
added locators for inline citations

### DIFF
--- a/deutsches-archaologisches-institut.csl
+++ b/deutsches-archaologisches-institut.csl
@@ -407,21 +407,18 @@
       </if>
     </choose>
   </macro>
-  <macro name="cited-page">
-    <text variable="locator" prefix=", "/>
-  </macro>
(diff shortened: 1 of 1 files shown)


deutsches-archaologisches-institut.csl
```diff
@@ -412,12 +412,13 @@
       <key variable="issued"/>
       <key macro="contributors-short-citation"/>
     </sort>
-    <layout prefix="" suffix="" delimiter="; ">
+    <layout delimiter="; ">
```
adam3smith: I'm actually wondering -- why aren't there parentheses here? That seems odd -- surely the style doesn't just put Author 1999 in the text?

adam3smith:
Thanks! You'll note that the style now automatically generates footnotes when a citation is inserted in the text. Citations inserted into footnotes will behave exact as before.

PR #3433, opened 2018-04-12
Update vita-latina.csl

Oxil-2:
Minor update (removing final dots in citation-mode)

### DIFF
--- a/vita-latina.csl
+++ b/vita-latina.csl
@@ -1,50 +1,59 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" default-locale="fr-FR">
-  <!--
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" default-locale="fr-FR" >
+
+<!--
 
 Feuille de style CSL 
 à destination du logiciel Zotero 
 comportant les instructions de formatage des références bibliographiques et de la bibliographie de littérature secondaire définies par la revue Vita Latina [http://www.persee.fr/collection/vita ; http://vitalatina.upv.univ-montp3.fr/]
-réalisée par Louis Autin (doctorant - Université Grenoble Alpes ; Universität Osnabrück)
+réalisée par Louis Autin (doctorant - Université Grenoble Alpes ; Universität Osnabrück - ATER Université de Strasbourg)
 
 Remarques :
 	- pour les thèses, Zotero ne permet pas de spécifier le directeur ; la feuille de style laisse donc cette indication à compléter manuellement, entre crochets ;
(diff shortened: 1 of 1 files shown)


Oxil-2:
Hi,

Ending the citation with a period was creating formatting problems when the end of the citation itself wasn't the end of the sentence (see attached picture). Such result wasn't appropriate with the citation style AND was inappropriate regarding the typographical rules in French.

I hope this makes sense to you!

2018-04-12 14:22 GMT+02:00 Sebastian Karcher <notifications@github.com<mailto:notifications@github.com>>:

Thanks -- the general expectation is that CSL will include all required punctuations for note styles. Are you saying that the citation style does not end footnotes with a period or why are you removing the period?

adam3smith:
(you can't use e-mail attachments. You'd need to upload screenshots on github)
Right, but this breaks compatibility with all other CSL styles and requires adding the period manually for every citation that stands by itself. 
If you're adding information to a citation, you can do so using the suffix field and keep the punctuation correct.

Oxil-2:
Mmmh, I don't really understand how I can distinguish in the CSL between a citation that stands by itself, and a citation put into a larger sentence in a footnote, like in the following example (I'm trying to put the image inside the e-mail). Thanks for your help anyway :-)

Example :
 [cid:ii_jfwizw3m0_162b9ec14724b420]


2018-04-12 14:46 GMT+02:00 Sebastian Karcher <notifications@github.com<mailto:notifications@github.com>>:

(you can't use e-mail attachments. You'd need to upload screenshots on github)
Right, but this breaks compatibility with all other CSL styles and requires adding the period manually for every citation that stands by itself.
If you're adding information to a citation, you can do so using the suffix field and keep the punctuation correct.

adam3smith:
you'll have to come to github to add images (click on "view on github" at the bottom of the e-mail.)
The distinction would not be in CSL but in how people insert the citation in Zotero. See https://www.zotero.org/support/word_processor_plugin_usage#customizing_cites

Oxil-2:
Here the image. 
ev_feuille_de_style

In this example, you see that the citation itself is a very small part of the footnote, which can contain a lot of information in addition to the citation, which may need some special formatting, etc. I don't find this solution (e.g. using Zotero to add information surrounding the citation) very useful, and I think it more convenient to manually add the period after the citations standing by themselves, since they are (in Humanities at least) the exception and not the rule : footnotes are normally used to cite but also to give large pieces of comment. But I'll understand if there is a compatibility problem.

adam3smith:
So in this case all you'd need to do for this to format correctly is to put ", en particulier p. 305-306." into the suffix of the Lemirre & Dubourdieu citation. I do think that makes sense both systematically (it really is part of the citation) and as a workflow. I've looked at a couple of articles in _Vita Latina_ and it's not at all unusual for footnotes to just contain a single citation. I don't think it's acceptable for authors to have to add a suffix "." to every citation (plus, as per above, it makes sense for us to have styles behaving uniformly), so I'm not going to take this, I'm afraid. 

Do you happen to have, btw., an updated link to the author instructions? The one in the style doesn't work anymore.

PR #3441, opened 2018-04-18
Add the journal title to the style to print by default

larsvilhuber:
The journal title (container-title) was not being printed in the default case. Added the else cause necessary to make this happen.


larsvilhuber:
This file validates when imported into Zotero.

adam3smith:
Thanks!

PR #3445, opened 2018-04-18
Create journal-of-surgery-and-medicine.csl

memu:
Journal of Surgery and Medicine
http://dergipark.gov.tr/josam


adam3smith:
Sorry for taking so long to review these. The style looks good in general. It looks like it's based on an existing style -- do you recall which? Same question for all the other styles you created.

memu:
Unfortunately, I don't remember the based style that I modified. I have searched a style that was conceptually similar and then I created all from that style. I checked each one in Zotero.

I will also prepare styles for other 70+ journals that exist in our database according to your response.

adam3smith:
@rmzelle could you take a quick look? The style looks sufficiently generic to me to not worry about the template link.

@memu for future reference, please do include the style you start from as a `rel="template"` link as per 8. here: https://github.com/citation-style-language/styles/wiki/Style-Requirements

rmzelle:
> I will also prepare styles for other 70+ journals that exist in our database according to your response.

@memu, sorry for letting this sit for so long. Before we merge this style, can you tell us a little bit more about these other 70+ journals? Do they all share the same citation format? (from reading the PoolText website, I guess the answer would be no?)

memu:
@adam3smith ok, next time i'll add `rel="template"` to citation styles. 

@rmzelle as you guessed the rest of journals' citation styles are different (of course some of them could share the same style)

PR #3447, opened 2018-04-18
Create journal-of-contemporary-medicine.csl

memu:
Journal of Contemporary Medicine 
http://dergipark.gov.tr/gopctd


rmzelle:
Thanks!

PR #3450, opened 2018-04-18
Create sanamed.csl

memu:
Sanamed
http://www.sanamed.rs


adam3smith:
Same thing with bolded authors -- I don't see the journal using or asking for those. The et-al change is just cosmetic. Also, same question on the original style you started from?

adam3smith:
@rmzelle ready for review

rmzelle:
Thanks!

PR #3452, opened 2018-04-19
Create journal-of-neolithic-archaeology.csl

MartinHinz:
Added style for the Journal of Neolithic Archaeology (http://www.j-n-a.org), based on the DAI style

### DIFF
--- a/journal-of-neolithic-archaeology.csl
+++ b/journal-of-neolithic-archaeology.csl
@@ -1,0 +1,465 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" delimiter-precedes-et-al="never" demote-non-dropping-particle="sort-only">
+  <info>
+    <title>Journal of Neolithic Archaeology</title>
+    <title-short>JNA</title-short>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks!

rmzelle:
@MartinHinz, sorry for the long wait for reviewing this, by the way.

I see that this journal publishes both in German and English. I was curious whether you tested the CSL style with both languages. The style also still contains some translations for Spanish, Italian, and French that we can probably delete.

MartinHinz:
@rmzelle , thank you for your feedback, and the detailed look to the journal! Actually, the journal articles from now on usually should be in english, but there might be some occasions when we accept german, or even articles in other languages. So I might check the translations, thank you for pointing this out. Will update the style soon if necessary.

rmzelle:
> Actually, the journal articles from now on usually should be in english, but there might be some occasions when we accept german, or even articles in other languages.

Okay, thanks. I added a note in the style that the journal publishes in multiple languages, so that it's clear in the future that the style shouldn't have a fixed "default-locale": https://github.com/citation-style-language/styles/commit/c7d187b8fa39b2367b247813040073457c7cfb25

PR #3453, opened 2018-04-19
Create mongolian-bibliography-standard

munkhmagnai:


munkhmagnai:
I did not understand this error at all. now what am I going to do?

adam3smith:
Is this style documented somewhere online?

POBrien333:
I assume this is to go with this Forum thread: https://forums.zotero.org/discussion/71500/need-help-in-submitting-mongolian-bibliography-style-standard

adam3smith:
Closing because of inactivity: @munkhmagnai -- we'd love to have this (or generally a Mongolian) style, so I'm happy to re-open this, but we do need some indication on which document(s) the style is actually based on, ideally in the form of a link.

rmzelle:
@munkhmagnai, I also left a note at https://forums.zotero.org/discussion/71500/need-help-in-submitting-mongolian-bibliography-style-standard, but translating https://github.com/citation-style-language/locales/blob/master/locales-mn-MN.xml would be a great step to improve Mongolian support in CSL. For some reason we already have a CSL locale file for Mongolian but it doesn't look like it was ever translated from English.

PR #3456, opened 2018-04-20

POBrien333:
https://forums.zotero.org/discussion/71441/improving-the-asa-american-sociological-association-style-citation-style-file


adam3smith:
Generally yes to adding edition, but I think we should check which one we're actually following.

PR #3457, opened 2018-04-22
Create bluebook-20e

JCBerger:


JCBerger:
Understood. I'm still getting a handle on things, but I can do that. Should I rename the old files to reflect the update? One thing I will admit, is that it is a work in progress — I tend to make changes or updates as I encounter the source types rather than all of it at once.

adam3smith:
no, don't change the filename, just add the edition number in the title (that's how we handle all current editions of styleguides)

adam3smith:
@JCBerger are you still working on this? I'm happy to leave it open if you are, just want to check in so we don't have abandoned tickets open.

JCBerger:
Right now it is on hold as I prepare for the bar exam.


_____________

rmzelle:
Let's just close this pull request. We'd still be more than happy to receive an update to bring our CSL style up to date with the 20th edition, but it's easier to just create a new pull request for that.

> don't change the filename, just add the edition number in the title (that's how we handle all current editions of styleguides)

Yes, so just make your changes directly to https://github.com/citation-style-language/styles/blob/master/bluebook-law-review.csl, per the instructions at https://github.com/citation-style-language/styles/blob/master/CONTRIBUTING.md#3b-submitting-changes-to-an-existing-style. Like @adam3smith said, just edit the existing style and keep the file name the same, but change the title to `<title>Bluebook Law Review 20th edition</title>`.

And success with your exam!

rmzelle:
(@adam3smith, by the way, instead of "Bluebook Law Review", the style should probably be renamed to "The Bluebook". I also don't remember what the story was behind https://github.com/citation-style-language/styles/blob/master/bluebook2.csl but it would be nice if we could get rid of that variant as well)

PR #3462, opened 2018-04-23
Update global-change-biology.csl

POBrien333:
via https://forums.zotero.org/discussion/71486/global-change-biology

No guidelines available. Papers available here: https://onlinelibrary.wiley.com/toc/13652486/current


rmzelle:
Thanks!

(I assume this can't be made a dependent of APA, as suggested by the OP?)

POBrien333:
@rmzelle 
The OP in the thread only answered after me making the edit (and then I was sleeping). 
Want me to make it a dependent?

rmzelle:
If you have a chance to confirm that the journal really uses APA, yes, please.

POBrien333:
Confirmed that it's just APA: https://github.com/citation-style-language/styles/pull/3468

PR #3463, opened 2018-04-23
Update resuscitation.csl

Blopo:
Adjustements with reduction of the number of authors when >6
Suppression of month in the date and '('')' as delimiters
All changes according to the specification of the journal

### DIFF
--- a/resuscitation.csl
+++ b/resuscitation.csl
@@ -120,7 +120,6 @@
       <else-if type="article-newspaper broadcast" match="any">
         <date variable="issued" delimiter=" ">
           <date-part name="year"/>
-          <date-part name="month" form="short"/>
           <date-part name="day"/>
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks! It looks like there are a few other issues with this style, in particular with the way the year is printed for books and chapters, if you'd like to make some more corrections.

I reverted your change to not show the month, since the way you changed it the style would print the year and day for the affected item types, which doesn't make sense. What type of items did you see the month for?

I also corrected the et-al settings.

Blopo:
Thanks! That's my first time that I contribute so I'm a bit lost. Thank you for correcting my mistakes

rmzelle:
Okay, we'll just take this as-is, then. Thanks for your contribution!

PR #3467, opened 2018-04-24
Create the-journal-on-efficiency-and-responsibility-in-education-and-…

SuperDomek:
Uploaded a new style for the ERIES journal.

### DIFF
--- a/the-journal-on-efficiency-and-responsibility-in-education-and-science.csl
+++ b/the-journal-on-efficiency-and-responsibility-in-education-and-science.csl
@@ -87,7 +87,7 @@
   </macro>
   <macro name="author-short">
     <names variable="author">
-      <name form="short" and="text" delimiter-precedes-last="never" et-al-min="4" initialize-with=". "/>
+      <name form="short" and="text" delimiter-precedes-last="never" initialize-with=". "/>
       <substitute>
         <names variable="editor"/>
         <names variable="translator"/>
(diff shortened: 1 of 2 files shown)


the-journal-on-efficiency-and-responsibility-in-education-and-science.csl
```diff
@@ -0,0 +1,636 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
+  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>The Journal on Efficiency and Responsibility in Education and Science</title>
+    <title-short>ERIES</title-short>
+    <id>http://www.zotero.org/styles/the-journal-on-efficiency-and-responsibility-in-education-and-science</id>
+    <link href="http://www.zotero.org/styles/the-journal-on-efficiency-and-responsibility-in-education-and-science" rel="self"/>
+    <author>
```
POBrien333: Can you also include a link to the template style you used?

POBrien333:
the style is not validating because of the et-al settings. :)

SuperDomek:
Oh, interesting, though. I validated it through the validator and created in the editor. :/ It doesn't look that hard to fix from the log. The original style is APA, I'll put it there. Thanks for your time.

adam3smith:
I fixed the et-al -- we have some tests on the settings sanity that go beyond pure validation and you had an `et-al-min` without an `et-al-use-first`,  which we don't allow.

rmzelle:
Thanks! Sorry for the wait!

PR #3469, opened 2018-04-25
Update oscola.csl

POBrien333:
via https://forums.zotero.org/discussion/71609/oscola-dates-and-issue-numbers-missing


adam3smith:
changes are exactly what I had in mind, but I'll want to keep the code comments, so might do this manually instead.

POBrien333:
Ah sorry. Should be fixed now. thanks for noting! :)

adam3smith:
Great, thanks. It'd be good to fix this in the visual editor, but since comments don't relate the GUI at all, it's a bit unclear how we would do this even if the XML parser allows it (i.e. where does the comment go when any surrounding text changes)

PR #3473, opened 2018-04-27
Create london-metropolitan-university.csl

POBrien333:
via https://forums.zotero.org/discussion/71625/style-request-based-on-cite-them-right-10th-edition-but-listing-all-authors-in-the-reference-list


rmzelle:
Thanks. Sorry for the wait.

PR #3476, opened 2018-04-27
Update university-of-york-harvard.csl

AVWoman:
Punctuation fix on publisher;
Forced exclusion of publisher from journal article.


rmzelle:
Thanks! Sorry for the wait.

PR #3478, opened 2018-04-28
Create al-jamiah-journal-of-islamic-studies-indonesia.csl

tonidjogja:
full note style used in Indonesia for Islamic studies


rmzelle:
(a style for this journal was added in #3492)

PR #3480, opened 2018-04-29
No lower case for thesis

NMonteix:


NMonteix:
@rmzelle 
I had a strange behavior testing this [commit](https://github.com/citation-style-language/styles/pull/3480/commits/244895442ec0fd80c305a21a2c7f2adda1497de4) on the validator. Instead of telling me there was a XML problem (a `/` is lacking at the end of line 50), it gave me strange errors (to me, even if they might make sense having lost a `/`). The screenshot of the errors is [there](https://ent.normandie-univ.fr/filex/get?k=qyYmqjWAH4rJnAeqahC).

rmzelle:
Thanks.

Note you can just add screenshots to GitHub comments (or you could give the validator URL if you tested an online style, like https://validator.citationstyles.org/?url=https%3A%2F%2Fraw.githubusercontent.com%2FNMonteix%2Fstyles%2F244895442ec0fd80c305a21a2c7f2adda1497de4%2Fcollection-du-centre-jean-berard.csl&version=1.0.1#source-code.

The validation errors are correct. Sometimes the messages can indeed be a little confusing, but usually they will at least pinpoint to where in the style there is a problem.

Issue #1619, opened 2015-07-13
apa-cv.csl not working with pandoc-citeproc


reagle:
`apa-cv.csl` generates parentheticals followed by references (standard format, not useful for CVs) with the following test files.

If you download the [test zip](http://reagle.org/joseph/2015/rtc/test.zip) you will see the following files:

```
apa-cv.csl  reviews-2.md  reviews.html  reviews.md  reviews.yaml
```

You can check that apa-cv.csl is the same as in [your repo](https://github.com/citation-style-language/styles/blob/master/apa-cv.csl). Then run it via pandoc:

```
> pandoc -s -w html -o reviews.html --csl=./apa-cv.csl --bibliography=reviews.yaml reviews-2.md
```

and confirm that review.html is not in the CV format. It works on a different file I have, so maybe I'm confusing myself somehow...?

rmzelle:
Looks like the main difference between regular APA and the CV-variant is the sort order in the bibliography: https://www.diffchecker.com/1h9dneu5

You can see how this is supposed to work at https://api.zotero.org/groups/4211/collections/A6BP5MVZ/items/top?format=bib&style=https://raw.githubusercontent.com/rmzelle/styles/master/apa-cv.csl. If you get different results with pandoc, it seems to be a problem with the latter.

reagle:
I also filed an issue there asking to see if folks can replicate.

reagle:
Yes, and I'm not asking for a different cite order, but a style which expands citations fully where they are invoked and without a post-bibliography, such as this [one](http://reagle.org/joseph/2003/cv/apa-cv.csl) which I use to generate my [cv](http://reagle.org/joseph/2003/cv/cv.html).

rmzelle:
Yeah, I agree that such a full-note style without a bibliography makes more sense than the current CV variant we have. @adam3smith, thoughts?

rmzelle:
Actually, I was thinking about the situation with the Zotero word processor plugins, where it's a bit convoluted to create a bibliography without citing items.

In general, though, a bibliography should be much more convenient, since you get your style output in a list, and you can benefit from bibliography-specific formatting options.

reagle:
To give you a sense of what I'm doing, here's the markdown (with the citation keys) and here's the result. It'd be nice to have a version of APA in the repo that permits this.

http://reagle.org/joseph/2003/cv/cv.md

http://reagle.org/joseph/2003/cv/cv.html

rmzelle:
[...]

PR #3487, opened 2018-04-30
Tweaking apa-cv to accord with discussion in https://github.com/citat…

reagle:
Here's the change to apa-cv.csl that I would like to use as discussed in #1619 . This renders the complete reference in place of the cite.

### DIFF
--- a/apa-cv.csl
+++ b/apa-cv.csl
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
   <info>
-    <title>American Psychological Association 6th edition (curriculum vitae)</title>
+    <title>American Psychological Association 6th edition (curriculum vitae, sorted by descending date)</title>
     <title-short>APA</title-short>
     <id>http://www.zotero.org/styles/apa-cv</id>
     <link href="http://www.zotero.org/styles/apa-cv" rel="self"/>
--- a/apa-cv.csl
+++ b/apa-cv.csl
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
   <info>
-    <title>American Psychological Association 6th edition (curriculum vitae, citation becomes reference, no bibliography)</title>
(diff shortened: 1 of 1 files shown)


reagle:
Someone's going to have to help me with the failed checks and what that means.

rmzelle:
There are errors for two things:

* Repository styles may not contain any unused macros. The offending macros are "author-short", "citation-locator", and "issued-year", so these need to be deleted.
* `<citation hanging-indent="true" et-al-min="8" et-al-use-first="6" et-al-use-last="true" entry-spacing="0" line-spacing="2">` uses a number of attributes that are only allowed on `<bibliography>`: "hanging-indent", "entry-spacing", and "line-spacing".

apa-cv.csl
```diff
@@ -1,15 +1,19 @@
...
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
   <info>
-    <title>American Psychological Association 6th edition (curriculum vitae, sorted by descending date)</title>
+    <title>American Psychological Association 6th edition (curriculum vitae, citation becomes reference, no bibliography)</title>
     <title-short>APA</title-short>
     <id>http://www.zotero.org/styles/apa-cv</id>
     <link href="http://www.zotero.org/styles/apa-cv" rel="self"/>
     <link href="http://owl.english.purdue.edu/owl/resource/560/01/" rel="documentation"/>
-    <author>
+    <autho>
```
rmzelle: @reagle, `<autho>` should read `<author>`. I always recommend running the style through https://validator.citationstyles.org/ after hand-editing to find typos like this.

reagle:
Okay, passed automated tests.

adam3smith:
Is there a reason you've removed the bibliography here? What I suggested on the ticket was to duplicate bibliography and citation. Removing the bibliography won't work e.g. for how the style is used on zotero.org -- but I also think generally the bibliography is the expected place for, well, the bibliography.

reagle:
@adam3smith, okay, I will restore the bibliography, which will then be redundant with the inline citations. Maybe there'll be a way to suppress them in pandoc.

reagle:
Okay, uploaded and it passed the tests. `suppress-bibliography: true` works with pandoc.

adam3smith:
excellent -- I was hoping there'd be a way to suppress the bib in Pandoc. Will review tonight & merge

adam3smith:
Thanks!

PR #3488, opened 2018-04-30
Update and rename associacao-brasileira-de-normas-tecnicas-ufjf.csl t…

Felipe-ferol:
…o ufjf.csl

I am the person who first uploaded this style in the repository, but I was using another email back then. We needed to change the title of the style because it was causing confusion amongst many others styles that already has "associação brasileira de normas tecnicas" in their titles.


rmzelle:
Thanks!

PR #3491, opened 2018-05-02
[SBL] add editortranslator locale term for verb-short

fbennett:
Adding the verb-short term should wake up the name collapsing.

(Per https://forums.zotero.org/discussion/71632/ed-and-trans-in-sbl-style)

### DIFF
--- a/society-of-biblical-literature-fullnote-bibliography.csl
+++ b/society-of-biblical-literature-fullnote-bibliography.csl
@@ -35,7 +35,7 @@
       <term name="editor" form="verb-short">ed.</term>
       <term name="translator" form="verb-short">trans.</term>
       <term name="translator" form="short">trans.</term>
-      <term name="editortranslator" form="verb">edited and translated by</term>
+      <term name="editortranslator" form="verb-short">ed. and trans.</term>
       <term name="collection-editor">
         <single>ed.</single>
         <multiple>eds.</multiple>
@@ -49,7 +49,7 @@
         <if variable="container-author reviewed-author" match="any">
           <group>
             <names variable="container-author reviewed-author" delimiter=", ">
-              <label form="verb-short"  suffix=" "/>
+              <label form="verb-short" suffix=" "/>
               <name and="text" delimiter=", "/>
             </names>
           </group>


society-of-biblical-literature-fullnote-bibliography.csl
```diff
@@ -35,7 +35,14 @@
...
       <term name="translator" form="short">trans.</term>
-      <term name="editortranslator" form="verb">edited and translated by</term>
+      <term name="editortranslator" form="verb-short">
+        <single>ed. and trans.</single>
+        <multiple>ed. and trans.</multiple>
+      </term>
+      <term name="editortranslator" form="verb">
+        <single>edited and translated by</single>
+        <multiple>edited and translated by</multiple>
+      </term>
```
rmzelle: @fbennett, is there a logic why this should be necessary?

The locales-en-US.xml locale file only defines a single form for these terms. See e.g. https://github.com/citation-style-language/locales/blob/2fddb0b00c15c04effe6fb46a3e74175f10896b7/locales-en-US.xml#L321.

adam3smith: @fbennett ping -- I'm happy to take this, but I'd also like to understand why it's necessary.

fbennett: The `editortranslator` trick works by calling a locale term of the same form with that name, when the call to `cs:names` includes exactly `editor` and `translator`. The `editor-translator` macro in SBL applies `verb-short`, which exists in `locales-en-US.xml`, so without anything more, it should work.

I don't remember the precise overloading hierarchy, but loading a term of a particular form will clobber its subordinates. Where SBL loads an `editortranslator` term in its `locale` section, I think that wipes out the `verb-short` version on which collapsing actually depends. Redefining it explicitly enabled the collapse.

If you want to trim this down, it looks like the `verb` form isn't used in the style, so I think that could be removed. Also, since the `single` and `multiple` forms are identical, defining the term value as a single static element should work fine. Would you like me to make those changes?

fbennett:
@adamsmith Do you need any further info or adjustments on this one?

fbennett:
@adam3smith @rmzelle The OP in the thread that prompted this pull request recently [posted a status query](https://forums.zotero.org/discussion/71632/ed-and-trans-in-sbl-style#latest).

adam3smith:
Sorry -- yes, if you could collapse the redundant single/multiple terms, that'd be great and we can take this.

fbennett:
@adam3smith Just so it doesn't slip off the radar, this item was not included in #3517.

PR #3497, opened 2018-05-08
ABNT style with small-caps and changes to song

jzeneto:
Optimization of ABNT (Brazilian Standards) style, with small-caps in author names (just in citations, not in bibliography), and improvements in song formatting (allowing it to be used also for sheet music).

### DIFF
--- a/associacao-brasileira-de-normas-tecnicas-small-caps.csl
+++ b/associacao-brasileira-de-normas-tecnicas-small-caps.csl
@@ -373,10 +373,6 @@
       </else-if>
     </choose>
   </macro>
-  <!--Resumo-->
-  <macro name="abstract">
-    <text variable="abstract"/>
-  </macro>
   <!--Mostra a ISBN-->
   <macro name="ISBN">
     <text variable="ISBN" prefix="ISBN "/>
@@ -482,8 +478,8 @@
             <text macro="translator"/>
             <text macro="edition"/>
             <text macro="publisher"/>
-            <!--Dados complementares "pagina, volume"-->
             <text macro="locators"/>
+            <text macro="ISBN"/>
             <text macro="access" suffix=". "/>
           </group>
         </else-if>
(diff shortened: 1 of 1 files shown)


rmzelle:
@jzeneto, do the ABNT standards require the use of small-caps?

jzeneto:
@rmzelle not explicitly. ABNT requires, ambiguously, "all-caps", but "small-caps" can also be used, and in some scenarios offer better reading (abnTex2 [has this option too](http://linorg.usp.br/CTAN/macros/latex/contrib/abntex2/doc/abntex2cite-alf.pdf#section.5)). And this style improves dramatically *song* formatting (which is almost inexistent in other ABNT styles), allowing its use for audio recordings and sheet music, following ABNT standards.

rmzelle:
@jzeneto, we greatly prefer to keep the number of CSL styles for style guides to a minimum, so unless there is an institute that specifically requires the use of small-caps, we're unlikely to accept this style variant.

We'd be more than happy to accept updates to our existing ABNT styles to have improved _song_ formatting.

@adam3smith, any opinion on whether it would be a good idea to switch over our standard ABNT styles (http://www.zotero.org/styles/associacao-brasileira-de-normas-tecnicas and http://www.zotero.org/styles/associacao-brasileira-de-normas-tecnicas-note) from all-caps to small-caps?

rmzelle:
Agree.

@jzeneto, would you mind creating a new pull request to update the two existing styles I listed above with your proposed changes?

jzeneto:
@rmzelle I made the changes just to one of the two (the [author-data version](http://www.zotero.org/styles/associacao-brasileira-de-normas-tecnicas)), because it's the style my changes were derived. The note version is entirely different in macros used, and unfortunately I don't have time to change it.
I'll make a new pull request with that change to author-data version. Thank you!

jzeneto:
@rmzelle I made the changes just to one of the two (the [author-data version](http://www.zotero.org/styles/associacao-brasileira-de-normas-tecnicas)), because it's the style my changes were derived. The note version is entirely different in macros used, and unfortunately I don't have time to change it.
I'll make a new pull request with that change to author-data version. Thank you!

PR #3501, opened 2018-05-09
Move colon out of group for conference papers

feldi-online:
The term colon inside the group with ", " delimiter resulted in a wrong rendering, always "[...], in:, [...]" for conference papers, see citation 3 in the picture:

image


rmzelle:
Good catch! Thanks!

rmzelle:
(I also checked the other Elsevier CSL styles, and none of them seem to suffer from the same issue)

PR #3502, opened 2018-05-09
Correct bibtex type for conference papers

feldi88:
The type paper-conference was used by mistake twice in the case distinction. Therefore, the correct bibtex mapping (inproceedings) was never chosen.


adam3smith:
Thanks. As a note, we're keeping this for legacy and in case it does some good, but the general recommendation is to rely on reference managers built-in bibtex capabilities which should generally be better than what a CSL style can achieve.

PR #3503, opened 2018-05-10
Update unified-style-linguistics.csl

azeini:
In books that are part of a series or collection, a period is inserted between book title and the series' title in the parentheses. I have fixed this following this discussion:
https://forums.zotero.org/discussion/71825/style-error-unified-style-sheet-for-linguistics-journals

I have also removed "hanging-indent" from `<bibliography>`, as it can cause problems while copy and pasting bibliographies.

### DIFF
--- a/unified-style-linguistics.csl
+++ b/unified-style-linguistics.csl
@@ -1,389 +1,388 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0"
-    demote-non-dropping-particle="display-and-sort" default-locale="en-US">
-    <info>
-        <title>Unified style sheet for linguistics journals</title>
-        <id>http://www.zotero.org/styles/unified-style-linguistics</id>
-        <link href="http://www.zotero.org/styles/unified-style-linguistics" rel="self"/>
-        <link href="https://www.linguisticsociety.org/sites/default/files/style-sheet_0.pdf" rel="documentation"/>
-        <author>
-            <name>Mark Dingemanse</name>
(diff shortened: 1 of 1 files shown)


adam3smith:
@azeini why did you remove the hanging indent? That looks like it's in the stylesheet you link to.

The way you fixed this only includes series for books. I'd at a minimum expect them to be required for chapters, no?

adam3smith:
@azeini did you see my questions above?

adam3smith:
@azeini did you see my questions above?

rmzelle:
@adam3smith, can you briefly review this? I hand-edited this without testing, but pretty confident it's correct. Should be good to merge.

I put back the hanging-indent, and the style should now print the "collection-title" for chapters as well. (which seems like a logical thing to include, even though there aren't examples for this case)

adam3smith:
Yes, looks good, thanks!

azeini:
Thank you very much for doing this. My apologies for the delayed reply, have been buried under work. Of course you are right that the same rule applies to chapters in a book that is part of a series. I am unsure about the indent, there are no rules in the guide, and am happy either way.

Unfortunately, the recent modifications seem to have introduced a new 'bug'. I now get the series information two times:

Amanat, Abbas. 2009. Apocalyptic Islam and Iranian shi’ism (Library of Modern Religion 4). (Library of Modern Religion 4). London: I. B. Tauris.

It's OK for the chapter in a series:

Surname, Name. 2018. Test title of a random chapter. This is the test title of the book (The Test Series Goes Here 5), 45--78. Berlin: Test Publisher.

azeini:
Works beautifully. Thank you very much for fixing it.

rmzelle:
Thanks for testing!

PR #3507, opened 2018-05-14

bwiernik:
Change `call-number` to `number` to be more consistent with CSL definitions.


adam3smith:
ah yes, thanks, I had meant to do this before merging.

PR #3508, opened 2018-05-16

CharlesQc:
-Added the prefix ", " before the accessed-date element. It prevents a situation like this : "www.website.comconsulté le 15 mai 2018" and becomes "www.website.com, consulté le 15 mai 2018"
-Added multiple types that should show a URL if locator is filled


rmzelle:
Thanks. You tested twice for the presence of the URL variable, and the "accessed-date" macro was only used once, so I just merged it into the "access" macro. It's also better to not put any punctuation meant to separate the output of macros within those macros themselves.

PR #3510, opened 2018-05-16
Create journal-of-agriculutral-and-applied-economics.csl

jmthomp:
I'm adding a CSL style that could be used for Zotero for the Journal of Agricultural and Applied Economics (JAAE).

### DIFF
--- a/journal-of-agricultural-and-applied-economics.csl
+++ b/journal-of-agricultural-and-applied-economics.csl
@@ -5,7 +5,8 @@
     <title-short>JAAE</title-short>
     <id>http://www.zotero.org/styles/journal-of-agricultural-and-applied-economics</id>
     <link href="http://www.zotero.org/styles/journal-of-agricultural-and-applied-economics" rel="self"/>
-    <link href="https://www.saea.org/wp-content/uploads/2013/08/reference-style-sheet.pdf" rel="documentation"/>
+    <link href="http://www.zotero.org/styles/apa" rel="template"/>
+    <link href="https://www.cambridge.org/core/journals/journal-of-agricultural-and-applied-economics/information/instructions-contributors" rel="documentation"/>
     <author>
       <name>Jada Thompson</name>
       <email>thompson.jadam@gmail.com</email>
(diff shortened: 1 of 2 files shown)


jmthomp:
Okay, I'm really new to this, so I'm not sure what this means. I validated it before submitting it and it works on my end via Zotero. Can you give me a better description of what you need.

journal-of-agricultural-and-applied-economics.csl
```diff
@@ -0,0 +1,664 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" and="text" et-al-min="3" et-al-use-first="1" page-range-format="expanded" demote-non-dropping-particle="never" default-locale="en-GB">
+  <info>
+    <title>Journal of Agricultural and Applied Economics</title>
+    <title-short>JAAE</title-short>
+    <id>http://www.zotero.org/styles/journal-of-agricultural-and-applied-economics</id>
+    <link href="http://www.zotero.org/styles/journal-of-agricultural-and-applied-economics" rel="self"/>
+    <link href="https://www.saea.org/wp-content/uploads/2013/08/reference-style-sheet.pdf" rel="documentation"/>
```
rmzelle: @jmthomp, do you know where you found these instructions at https://www.saea.org/wp-content/uploads/2013/08/reference-style-sheet.pdf?

When I went through the SAEA website, http://www.saea.org/jaae/manuscript-submission-guidelines/ directed me to https://www.cambridge.org/core/journals/journal-of-agricultural-and-applied-economics/information/instructions-contributors, which seems to list the same examples at the end of the page. That's probably a better documentation link.

rmzelle:
> Okay, I'm really new to this, so I'm not sure what this means. I validated it before submitting it and it works on my end via Zotero. Can you give me a better description of what you need.

Thanks for your submission! Our automated testing system found a typo in the style file name ("agriculutral" instead of "agricultural") which I fixed. This looks good and I'll accept it once we figure out the best documentation link for the style.

rmzelle:
Actually, another question. Did you base this on an existing CSL style? If so, we'd like to know which one, so we can add a "template" link (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link)

journal-of-agricultural-and-applied-economics.csl
jmthomp: The style sheet, is on http://www.saea.org/jaae/manuscript-submission-guidelines/ At the top there is a menu called JAAE --> then Reference Style Sheet. That was the pdf. The same references are in the contributors. As frustrating as it is these are the guidelines for this journal. So no other information on reference style other than those examples.

jmthomp: Whatever you feel is best. It's all the same information. :)Thanks for your help.

rmzelle:
> So, I guess the following should be added in?

Yes, exactly. I just went ahead and made the change. (and accepted the style; thanks!)

PR #3511, opened 2018-05-18
Adjust edition to style guide

pmelchor:
- replace short-form term with custom suffix
- change position in bibliography

### DIFF
--- a/the-open-university-harvard.csl
+++ b/the-open-university-harvard.csl
@@ -158,7 +158,10 @@
   <macro name="edition">
     <choose>
       <if is-numeric="edition">
-        <number variable="edition" form="ordinal" suffix=" edn"/>
+        <group delimiter=" ">
+          <number variable="edition" form="ordinal"/>
+          <text term="edition" form="short"/>
+        </group>
       </if>
       <else>
         <text variable="edition" suffix="."/>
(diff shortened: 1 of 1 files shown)


the-open-university-harvard.csl
```diff
@@ -158,10 +158,7 @@
   <macro name="edition">
     <choose>
       <if is-numeric="edition">
-        <group delimiter=" ">
-          <number variable="edition" form="ordinal"/>
-          <text term="edition" form="short"/>
-        </group>
+        <number variable="edition" form="ordinal" suffix=" edn"/>
```
adam3smith: why are you doing this? Always preferable to use terms (which you can define above if they need to differ from default) so that, e.g. people can more quickly change this style to a different locale

pmelchor: I didn't realize I could define terms for the default locale. Is my new commit the right way to do that? Thanks!

adam3smith:
Thanks!

adam3smith:
Yes it is. We typically put the `<locale>` right below `</info>` but that's just convention, functionally doesn't matter.

rmzelle:
> I saw that locale is listed after macro in the specification... but I didn't notice that citation and bibliography are listed before macro... which would have told me that the order in the specification is not the one to follow!

There is some logic to the order in the specification, by the way, since I listed the optional elements (`bibliography`, `macro`, and `locale`) after the required ones (`info` and `citation`).

PR #3514, opened 2018-05-20
Added website to MBoC

jpolka:
As described in https://www.molbiolcell.org/info-for-authors


rmzelle:
Thanks. Did you test this? I cleaned things up a little, but it doesn't look like the style would print the webpage title and website name.

jpolka:
Thanks for noticing that @rmzelle! I've added those fields and it seems to be working fine for me in Zotero.

rmzelle:
Thanks!

PR #3522, opened 2018-05-22
Isolate calls to editor + translator, provide for collapse of verb form

fbennett:
`editortranslator` collapse only works when editor and translator are the only two variables called on `cs:names`. This isolates them where they were mixed with other variables, preserving the original formatting. I've placed the three variants of the `editor translator` combination used in the style at the top. And I tested it this time!


fbennett:
@adam3smith Ready for review.

PR #3517, opened 2018-05-22
Fix redundant terms

fbennett:
In reference to #3491, there are 253 styles in the repository with similarly redundant terms. This pull request fixes them all. Here is the node script used to apply the fix (installed in the top level of the repo and run from there, after fetching its dependencies from `npm`):
```
var fs = require('fs');
var xml = require('xmldom').DOMParser;
var xpath = require('xpath');
var enc = require('encode-decode');
var select = xpath.useNamespaces({"cs": "http://purl.org/net/xbiblio/csl"});

var ret = {};

function EntityFixer() {
    this.dat = {};
    var lurkers = {
        "&lt;": "{LESS-THAN}",
        "&gt;": "{GREATER-THAN}",
        "&nbsp;": "{NO-BREAK-SPACE}"
    }
    this.memo = function(txt) {
        for (key in lurkers) {
            var rex = new RegExp(key, "g");
            txt = txt.replace(rex, lurkers[key]);
        }
        var m = txt.match(/(\&\#[0-9]+;)/g);
        if (m) {
            for (var memo of m) {
                this.dat[memo] = enc.htmlDecode(memo);
            }
        }
        return txt;
    }
    this.dememo = function(txt) {
        for (var key in this.dat) {
            var rex = new RegExp(this.dat[key], "g");
            txt = txt.replace(rex, key);
        }
        for (key in lurkers) {
            var rex = new RegExp(lurkers[key], "g");
            txt = txt.replace(rex, key);
        }
        return txt;
    }
}



for (var fn of fs.readdirSync('.')) {
    var writeToFile = false;
    var entityFixer = new EntityFixer();
    if (fn.slice(-4) !== ".csl") continue;
    var txt = fs.readFileSync(fn).toString();
    txt = entityFixer.memo(txt);
    var doc = new xml().parseFromString(txt);
    var termNodes = select("//cs:locale/cs:terms/cs:term", doc);
    for (var termNode of termNodes) {
        var single = select(".//cs:single", termNode)[0];
        var multiple = select(".//cs:multiple", termNode)[0];
        if (single) {
            single = single.textContent ? single.textContent : false;
            multiple = multiple.textContent ? multiple.textContent : false;
            if (single && single === multiple) {
	            for (var i=0,ilen=termNode.childNodes.length;i<ilen;i++) {
	                termNode.removeChild(termNode.childNodes[0]);
	            }
                var textNode = doc.createTextNode(single);
	            termNode.appendChild(textNode);
	            writeToFile = true;
            }
        }
[...]

PR #3524, opened 2018-05-23
Migrate common name attributes upward

fbennett:
Not sure how this one will be received ...

This is a more ambitious set of changes than #3517. It removes inheritable attributes from `cs:names` and `cs:name` nodes if they are the same in all calls originating from their `cs:citation` or `cs:bibliography` parent, and if they do not conflict across the two parents, and sets them on the parent instead.

The advantage of the reprocessed styles is a reduction in code replication. Also, since all inheritable attributes that remain on `cs:names` and `cs:name` nodes are consequential, it makes it easier to spot discrepancies, such as disparate application of `initialize-with`.

Here is the script used to apply the changes, with notes on the logic of the scan. (If this full pull request seems to ambitious, the script can be applied to individual styles as well.)

```
// Safely migrate cs:names and cs:name attributes upward

// (1) Walk all macros in each of cs:citation and cs:bibliography
// (2) Check inheritable attributes on all cs:names and cs:name nodes found in the walk
// (3) Flag all attributes with identical values across all calls in the scope for migration
// (4) Forego migration of attribute if its migration values differ in cs:citation and cs:bibliography
         scope, and there are cs:names-calling macros common to the two scopes
// (5) Forego migration of attribute if common value differs from an existing value set on parent
//       (cs:citation or cs:bibliography respectively)
//
// If run with no arguments as node ./migrateNamesAttributes.js, processes all independent styles.
// If run with style file name as sole argument, processes only that single file.

var fs = require('fs');
var xml = require('xmldom').DOMParser;
var xpath = require('xpath');
var enc = require('encode-decode');
var select = xpath.useNamespaces({"cs": "http://purl.org/net/xbiblio/csl"});

var singleFile = process.argv[2];

function EntityFixer() {
    this.dat = {};
    // Avoid chaos
    var lurkers = {
        "&lt;": "{LESS-THAN}",
        "&gt;": "{GREATER-THAN}",
        "&nbsp;": "{NO-BREAK-SPACE}"
    }
    this.memo = function(txt) {
        for (var key in lurkers) {
            var rex = new RegExp(key, "g");
            txt = txt.replace(rex, lurkers[key]);
        }
        var m = txt.match(/(\&\#[0-9]+;)/g);
        if (m) {
            for (var memo of m) {
                // Avoid chaos again. Don't touch these!
                if ("&#032;" === memo) continue;
                if ("&#009;" === memo) continue;
                this.dat[memo] = enc.htmlDecode(memo);
            }
        }
        return txt;
    }
    this.dememo = function(txt) {
        for (var key in this.dat) {
            var rex = new RegExp(this.dat[key], "g");
            txt = txt.replace(rex, key);
        }
        for (var key in lurkers) {
            var rex = new RegExp(lurkers[key], "g");
            txt = txt.replace(rex, key);
        }
        return txt;
    }
}

var AttributeCollector = function(doc) {
    this.doc = doc;
    this.parent_macro = null;
    this. acc = {
        citation: {
            names: {},
            name: {},
            seen_macros: {},
            macros: {}
        },
        bibliography: {
            names: {},
            name: {},
            seen_macros: {},
            macros: {}
        }
    };
    this.attrMap = {
        names: {
            "delimiter": "names-delimiter"
        },
        name: {
            "and": "and",
            "delimiter-precedes-et-al": "delimiter-precedes-et-al",
            "delimiter-precedes-last": "delimiter-precedes-last",
            "et-al-min": "et-al-min",
            "et-al-use-first": "et-al-use-first",
            "et-al-use-last": "et-al-use-last",
            "et-al-subsequent-min": "et-al-subsequent-min",
            "et-al-subsequent-use-first": "et-al-subsequent-use-first",
            "initialize": "initialize",
            "initialize-with": "initialize-with",
            "name-as-sort-order": "name-as-sort-order",
            "sort-separator": "sort-separator",
            "form": "name-form",
            "delimiter": "name-delimiter"
        }
    };
    this.processAttributes = function(seg, subseg, callback, node) {
        if (!node) {
            node = select("//cs:" + seg + "/cs:" + subseg, this.doc)[0];
            if (!node) return;
        }
        if (!this[seg + "Node"]) {
            this[seg + "Node"] = select("//cs:" + seg, this.doc)[0];
        }
        // Accumulate attributes on cs:names / cs:name
        var tags = ["names", "name"];
        for (var tag of tags) {
            var nxNodes = select(".//cs:" + tag, node);
            for (var nxNode of nxNodes) {
                callback.call(this, seg, subseg, tag, nxNode);
            }
        }
        // Drill down on either cs:key or cs:text macros
        // (the former will appear on cs:sort, and the latter never appear there)
        var macroNodes = select(".//cs:key[@macro]", node);
        if (macroNodes.length === 0) {
            macroNodes = select(".//cs:text[@macro]", node);
        }
        for (var macroNode of macroNodes) {
            var macroName = macroNode.getAttribute("macro");
            if (this.acc[seg].seen_macros[macroName]) continue;
            this.acc[seg].seen_macros[macroName] = true;
            var macroSourceNode = select(".//cs:macro[@name='" + macroName + "']", this.doc)[0];
            this.processAttributes(seg, subseg, callback, macroSourceNode);
        }
    }
    this.cleanup = function() {
        // Strip no-migrate values
        for (var seg of ["citation", "bibliography"]) {
            for (var tag of ["names", "name"]) {
                for (var key in this.acc[seg][tag]) {
                    if (!this.acc[seg][tag][key]) {
                        delete this.acc[seg][tag][key];
                    }
                }
            }
        }
        
        // Note that name-as-sort-order should be ignored
        // in sort context, since it is always forced to
        // "always".
        
        // In each segment, check each attribute marked for migration, to see if
        // (a) any macro is used in both segments, and (b) the migration value differs
        // from that in the other segment. If the answer to both is "yes," delete the
        // attribute migration marker from BOTH segments.

        var hasCommon = false;
        for (var key in this.acc.citation.macros) {
            if (this.acc.bibliography.macros[key]) {
                hasCommon = true;
                break;
            }
        }
        if (hasCommon) {
            for (var key in this.acc.citation.names) {
                var cVal = this.acc.citation.names[key];
                if (this.acc.bibliography.names[key] !== cVal) {
                    delete this.acc.citation.names[key];
                    delete this.acc.bibliography.names[key];
                }
            }
            for (var key in this.acc.citation.name) {
                var cVal = this.acc.citation.name[key];
                if (this.acc.bibliography.name[key] !== cVal) {
                    delete this.acc.citation.name[key];
                    delete this.acc.bibliography.name[key];
                }
            }
        }
        this.acc.citation.macros = {};
        this.acc.bibliography.macros = {};
        this.acc.citation.seen_macros = {};
        this.acc.bibliography.seen_macros = {};
    }
    this.finish = function(fn) {
        var writeFile = false;
        for (var seg of ["citation", "bibliography"]) {
            for (var tag of ["names", "name"]) {
                if (this[seg + "Node"]) {
                    for (var attr in this.acc[seg][tag]) {
                        writeFile = true;
                        var higherAttr = this.attrMap[tag][attr];
                        this[seg + "Node"].setAttribute(higherAttr, this.acc[seg][tag][attr])
                    }
                }
            }
        }
        if (writeFile) {
            var txt = entityFixer.dememo(this.doc.toString());
            //console.log(txt);
            fs.writeFileSync(fn, txt + "\n")
        }
    }
}

function collector(seg, subseg, tag, nxNode) {
    if (this.parent_macro) {
        this.acc[seg].macros[this.parent_macro] = true;
    }
    if (tag === "names" && nxNode.parentNode.tagName === "substitute" && nxNode.attributes.length === 1) return;
    for (var attr in this.attrMap[tag]) {
        if (subseg === "sort" && attr === "name-as-sort-order") continue;
        var val = nxNode.getAttribute(attr);
        var accVal = this.acc[seg][tag][attr];
        if (accVal === undefined) {
            this.acc[seg][tag][attr] = val;
        } else if (accVal !== val) {
            this.acc[seg][tag][attr] = false;
        }
    }
}

function editor(seg, subseg, tag, nxNode) {
    if (this.parent_macro) {
        this.acc[seg].macros[this.parent_macro] = true;
    }
    if (tag === "names" && nxNode.parentNode.tagName === "substitute" && nxNode.attributes.length === 1) return;
    for (var attr in this.acc[seg][tag]) {
        var higherAttr = this.attrMap[tag][attr];
        var higherVal = this[seg + "Node"].getAttribute(higherAttr);
        if (higherVal && higherVal !== this.acc[seg][tag][attr]) {
            delete this.acc[seg][tag][attr];
        } else {
            nxNode.removeAttribute(attr);
        }
    }
}

for (var fn of fs.readdirSync('.')) {
    var writeToFile = false;
    var entityFixer = new EntityFixer();
    if (fn.slice(-4) !== ".csl") continue;
    if (singleFile && fn !== singleFile) continue;
    
    console.log(fn);
    var txt = fs.readFileSync(fn).toString();
    txt = entityFixer.memo(txt);
    var doc = new xml().parseFromString(txt);
    var attributeCollector = new AttributeCollector(doc);

    // Collect common attributes
    attributeCollector.processAttributes("citation", "sort", collector);
    attributeCollector.processAttributes("citation", "layout", collector);
    attributeCollector.processAttributes("bibliography", "sort", collector);
    attributeCollector.processAttributes("bibliography", "layout", collector);

    // Reset for rerun of walk
    attributeCollector.cleanup();
    
    attributeCollector.processAttributes("citation", "sort", editor);
    attributeCollector.processAttributes("citation", "layout", editor);
    attributeCollector.processAttributes("bibliography", "sort", editor);
    attributeCollector.processAttributes("bibliography", "layout", editor);

    attributeCollector.finish(fn);
}
```


fbennett:
Oops. It seems there is a problem with validation ...

fbennett:
It's actually a problem with the Travis checks, not with the styles. Travis is demanding that if one of `et-al-min` or `et-al-use-first` is present on a node, its partner must also be present. That is not required by the schema. The first style caught by Travis, `gallia-prehistoire.csl`, sets a common value of `et-al-use-first="1"` in `cs:bibliography`, with values of `et-al-min="3"` or `et-al-min="10"` depending on the name variable rendered. With the `set-al-use-first` attribute only moved to `cs:bibliography`, the style still validates fine against the schema.

fbennett:
Oh! I jumped to conclusions there, as I so often do.

There is a specific problem with the `gallia-prehistoire.csl` style itself, in its current form. The style sets `et-al-use-first="1"` on the `cs:style` node, and there are bare `cs:name` elements in the style. It looks like Travis is correctly doing a tree-wise scan for the attribute, and failing when it hits one of those nodes, to find that there is no `et-al-min` value available. The style would have failed Travis checks anyway, if checked in in its current form.

So the automated edits seem to be triggering a failure only because they touch the style code, and attract the attention of Travis. Let's try this again with a revised version of that style ...

adam3smith:
The concern here is readability. We've been adding the `name` attributes to `cs:name` because it makes sense to think about them where they names are called (it's also where the visual editor takes you when you click on a name). So I'm wondering if there really is sufficient benefit in avoiding the relatively minor code duplication. 
This isn't a strong position, though, so I'm perfectly happy to be convinced. Curious what @rmzelle and @POBrien333 think

fbennett:
Yes---I'm not pushing at all. It's just something that I started tinkering with, and wanted to finish. When it's done and working, I can bundle the script in a project so that it's available for use if desired.

fbennett:
Well, that was an interesting journey. Tests finally pass!

The script touches 1604 styles in all. The repeated test failures above were triggered by a series of sloppy-coding bugs that failed to preserve the pairing of `et-al-min` and `et-al-use-first` on individual nodes. With that fixed, all is well.

Caution advised, but these should produce output identical to that of the originals. As to whether moving these attributes actually makes maintenance easier, that is one for active maintainers to call. I just wanted to give the script a proper test run once I had begun coding it.

(If this one isn't adopted, feel free to close the pull request at will!)

adam3smith:
I agree on the `et-al` attributes -- I prefer those on cs:citation and cs:bibliography, too. So I'd definitely be in favor of a modified version of this that only affects et als.

This goes further, though, and moves all inheritable elements there, i.e. also things like `initialize-with`, `and="text"` etc.

fbennett:
As a side-note, more styles would be affected if `et-al-min` and `et-al-use-first` were allowed to move independently&mdash;many styles apply a common value for `et-al-use-first', but differing values of `et-al-min`. Travis throws an error if the two attributes are not set as a pair on individual nodes, though, so the script currently treats them as a unit.

fbennett:
I've run a little poke-it-and-see test, and I was wrong about the Travis logic. Either of `et-al-min` or `et-al-use-first` can be set solo on a node, so long as the partner can be acquired by inheritance&mdash;so the Travis test is robust, and migrating either of those individually should be perfectly okay, barring bugs in the migration script. I'll try to fix that and make an additional checkin, so that you can see which arrangement you like best.

fbennett:
There we go. The third check-in touches just six additional styles, so there was a high degree of editorial consistency already (and no, that is not surprising!). Two styles were very slightly edited to avoid Travis errors: `lethaia.csl` and `paleobiology.csl`. These seem to differ only slightly, and both set a value on `cs:citation` that was overridden in the sole `cs:name` call within that scope. This blocked the migration of one element of the pair, and Travis blew up on that macro when (I think) inspecting it through a call in`cs:key`. Aligning the values allowed the migration to go through for both attributes, and all six styles pass with flying colors.

PR #3526, opened 2018-05-25
Create journal-of-porphyrins-and-phthalocyanines.csl

jakuhne:
Submitting new style by personal request and action

### DIFF
--- a/journal-of-porphyrins-and-phthalocyanines.csl
+++ b/journal-of-porphyrins-and-phthalocyanines.csl
@@ -3,7 +3,7 @@
   <info>
     <title>Journal of Porphyrins and Phthalocyanines</title>
     <title-short>J. Porphyrins Phthalocyanines</title-short>
-    <id>https://csl.mendeley.com/styles/493092071/journal-of-porphyrins-and-phthalocyanines</id>
+    <id>http://www.zotero.org/styles/journal-of-porphyrins-and-phthalocyanines</id>
     <link href="http://www.zotero.org/styles/journal-of-porphyrins-and-phthalocyanines" rel="self"/>
     <link href="http://www.zotero.org/styles/histopathology" rel="template"/>
     <link href="http://jpp.fontismedia.com/jpp/guides/Notice_to_Authors.pdf" rel="documentation"/>
@@ -14,8 +14,8 @@
     <contributor/>
     <category citation-format="numeric"/>
     <category field="organic chemistry"/>
-    <issn>0309-0167</issn>
-    <eissn>1365-2559</eissn>
+    <issn>1088-4246</issn>
+    <eissn>1099-1409</eissn>
     <summary>In agreement with JPP http://jpp.fontismedia.com/jpp/guides/Notice_to_Authors.pdf</summary>
     <published>5-24-2018</published>
     <updated>2018-05-25T03:40:07+00:00</updated>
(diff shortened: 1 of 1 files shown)


journal-of-porphyrins-and-phthalocyanines.csl
```diff
@@ -0,0 +1,251 @@
...
+      </if>
+    </choose>
+  </macro>
+  <macro name="title">
+    <choose>
+      <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
+        <text variable="title" text-case="title"/>
+      </if>
+      <else>
+        <text value=""/>
```
adam3smith: What's the goal here? If you actually don't want a title for some item types, you can just not have an `else` condition.

journal-of-porphyrins-and-phthalocyanines.csl
```diff
@@ -0,0 +1,247 @@
...
+  <macro name="editor">
+    <names variable="editor">
+      <name delimiter-precedes-et-al="never" delimiter-precedes-last="never" initialize-with="" name-as-sort-order="all" sort-separator=" "/>
+      <label form="short" prefix=", "/>
+    </names>
+  </macro>
+  <macro name="author">
+    <group suffix=".">
+      <names variable="author">
+        <name and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="19" et-al-use-first="18" et-al-use-last="true" initialize-with="" name-as-sort-order="all" sort-separator=" "/>
```
adam3smith: What are the et al rules here? You have 19/18 here and then 4/3 below and you also use the `et-al-use-last`, which would be unusual for chemistry.

journal-of-porphyrins-and-phthalocyanines.csl
jakuhne: Thank you for pointing these out. If you can't tell I'm new to this. I only have a little experience and that's with java. I started on this before I realized that I could ask someone else to make these. I used the visual editor so I was uncertain about just deleting functions.

adam3smith: I removed this already in the style, so only thing left is the et-al quesiton.

journal-of-porphyrins-and-phthalocyanines.csl
jakuhne: The 18/19 is the limit on Mendeley's visual editor.  I definitely didn't want a limit and I think it got cranky when I tried turning the et-al-use-last off. Something like that. Thank you.

adam3smith: So just to clarify, the style should never use et-al in the bibliography?

jakuhne: The guidelines don't explicitly say so but looking at a published review it looks like et-al is never used. 

"Journals should be cited in the following way: author surnames and initials (no dot after initial), insert “and” between the last two authors (with no comma), dot, journal name (italicized and abbreviated in accordance with the ACS Style Guide..."

jakuhne:
Thank you for cleaning up my crappy code @adam3smith!

adam3smith:
Thanks -- this actually looked very good from the start, nothing crappy about it.

POBrien333:
@jakuhne

Well done! And you got lucky your style god merged so quickly! (I was helping Jeremy via email as he contacted as he based the style on one I made. ) :)

PR #3527, opened 2018-05-27
Remove unused macros

fbennett:
Just a bit of tidying up. It looks like most of these were my doing from long ago.

### DIFF
--- a/chicago-annotated-bibliography.csl
+++ b/chicago-annotated-bibliography.csl
@@ -44,40 +44,9 @@
       <term name="translator" form="short">trans.</term>
     </terms>
   </locale>
-  <macro name="editor-translator">
-    <group delimiter=", ">
-      <group delimiter=" ">
-        <choose>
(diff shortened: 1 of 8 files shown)


rmzelle:
That's odd. Our Travis tests are supposed to check for any unused macros.

There is also some white space problem in this PR (a few newly inserted blank lines).

rmzelle:
Ah, apparently we are exempting the Chicago styles from those tests. I remember now.

https://github.com/citation-style-language/styles/blob/e7bf5fe64c40695cade64d24950cab7f55743cb2/spec/spec_helper.rb#L30

rmzelle:
Let's merge these changes in #3528.

rmzelle:
And obviously thanks!

adam3smith:
We're exempting these because it makes it much easier for me to handle adjustments of CMoS styles across different versions, so let's please keep the unused macros in.

PR #3528, opened 2018-05-29
Delete unused macros in Chicago styles

rmzelle:
Replaces #3527.

### DIFF
--- a/chicago-annotated-bibliography.csl
+++ b/chicago-annotated-bibliography.csl
@@ -44,9 +44,6 @@
       <term name="translator" form="short">trans.</term>
     </terms>
   </locale>
-  
-  
-  
   <macro name="secondary-contributors">
     <choose>
       <if type="chapter paper-conference" match="none">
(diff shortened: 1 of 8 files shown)


rmzelle:
@adam3smith, I'll leave this for you to merge. (I quickly glanced over the diffs and the only thing I see changed are indeed macros being deleted; summary stats are "+0 −2,088" too)

As long as Travis is happy I'm happy.

adam3smith:
So for some background -- when I make changes to a Chicago style I use diff viewer (Meld in my case) -- to merge over the changes to all styles. This involves some manual checking, so the fewer differences there are the better. Unless there are objections, I'd ask that we keep the macros in.

rmzelle:
> Unless there are objections, I'd ask that we keep the macros in.

I didn't check if there is a subset of unused macros common to all these styles. If there is, we could delete those.

adam3smith:
There isn't -- (the fullnote styles aren't affected by this PR because they don't contain any extra macros). I'm OK modifying the author-date Chicago styles, though; those don't follow the same template. Do you want to do this in this PR or in a new one.

rmzelle:
> Do you want to do this in this PR or in a new one.

It's fine to repurpose this one. Are you up for it?

PR #3530, opened 2018-05-29
Update journal-of-petrology.csl

POBrien333:
via https://forums.zotero.org/discussion/72083/style-error-journal-of-petrology


adam3smith:
Thanks

PR #3533, opened 2018-05-30
Create revista-portuguesa-de-arqueologia.csl

POBrien333:
via https://forums.zotero.org/discussion/71514/style-request-revista-portuguesa-de-arqueologia


rmzelle:
Thanks! (even though the example references use English translations, looks like the journal is multilingual and accepts articles in Portuguese, Spanish, French and English)

PR #3535, opened 2018-06-04
[Music Theory Spectrum] Title case:

adam3smith:
Reported https://forums.zotero.org/discussion/72059/title-case-not-working#latest and checked


rmzelle:
Thanks! :)

PR #3537, opened 2018-06-06
Create health-and-human-rights-journal.csl

POBrien333:
via https://forums.zotero.org/discussion/47216/style-request-health-human-rights

### DIFF
--- a/health-and-human-rights-journal.csl
+++ b/health-and-human-rights-journal.csl
@@ -1,0 +1,195 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US" xmlns="http://purl.org/net/xbiblio/csl">
+  <info>
+    <title>Health and Human Rights Journal</title>
+    <title-short>HHR</title-short>
+    <id>http://www.zotero.org/styles/health-and-human-rights-journal</id>
+    <link href="http://www.zotero.org/styles/health-and-human-rights-journal" rel="self"/>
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks!

rmzelle:
@POBrien333, I just got some private feedback about this style. It seems like this should be a "note" style instead of an "in-text" "numeric" one, no? Do you have time to take a look?

rmzelle:
https://cdn1.sph.harvard.edu/wp-content/uploads/sites/2469/2019/12/HHR-EDITORIAL-STYLE-GUIDE_2019-1.pdf, page 17: "The Center uses endnotes, not footnotes, for citations. Endnote numbers within the text should be placed at the end of the sentence, rather than within the sentence. Each sentence may include only one. Combine into this note all relevant reference material from the sentence."

adam3smith:
Yeah, this is a note style -- as I said [way back in 2015, ](https://forums.zotero.org/discussion/comment/218538/#Comment_218538) this will actually be a lot of work to implement, though -- it's not as simple as switching this to note.

rmzelle:
Yeah, that's what I wrote back.

lucianobottini:
Hi sorry to bother you guys (as I have no idea how you proceed) but anyone is in charge of updating styles or willing to fix bugs as the one observed? This style would be very useful now  - I have a couple of authors wanting to use it but don't have the skill to do it myself. If anyone is willing to help, I'd be most grateful.  Under COVID-19, health and human rights is a good journal to have it right.

POBrien333:
You can open an issue here: https://github.com/citation-style-language/styles/issues
Please be clear, provide examples (current vs needed) to make this easy to understand and we'll look at it. :)

PR #3548, opened 2018-06-11
Create american-journal-of-industrial-medicine.csl

POBrien333:
via https://forums.zotero.org/discussion/72303/how-to-import-endnote-styles-into-zotero


rmzelle:
Thanks!

PR #3549, opened 2018-06-12

CharlesQc:
-Changed my email address, as I do not use the previous one anymore. 
-Removed both "vol. " and "n° ", as requested by the department's Director of 2nd and 3rd grade in history and stated in the department's documentation under "Article dans une revue" (although it was confirmed to me that there should be a space between the first comma and the issue number, which does not appear in the documentation itself).


rmzelle:
Thanks!

PR #3550, opened 2018-06-15
Create acme-an-international-journal-for-critical-geographies.csl

POBrien333:
https://forums.zotero.org/discussion/72356/style-request-acme-an-international-journal-for-critical-geographies

I have some extra spaces after the title. Can't find where it's from.


rmzelle:
Thanks! (looks like you figured out the space issue by yourself)

PR #3551, opened 2018-06-15
Make harvard-university-of-gloucestershire.csl dependent

POBrien333:
…ent/harvard-university-of-gloucestershire.csl

https://forums.zotero.org/discussion/72381/style-error-university-of-gloucestershire-harvard


POBrien333:
harvard-stellenbosch-university harvard-university-of-cape-town  were dependent of the Gloucestershire style. Will need to check those first if they can be moved over to cite-them-right, too.

rmzelle:
Thanks!



> harvard-stellenbosch-university & harvard-university-of-cape-town have this as a template link. That shouldn't matter, right?

Well, actually we currently require that all "template" links link to existing CSL independent styles. I certainly helps catch typos, but the drawback is that we get cases like these. I just pointed the "template" links to the parent style.

PR #3555, opened 2018-06-19
Create mediterranean-journal-of-infection-microbes-and-antimicrobials…

dsleptsov:
….csl


dsleptsov:
What does the procedure require to proceed?

adam3smith:
as per the message above, we'll review and get back to you as soon as possible, but might take a little time.

rmzelle:
Perfect, thanks!

PR #3558, opened 2018-06-20
Fix scale macro for din-1505-2

heikojansen:
The contents of the "scale" macro were inserted no matter if there was any data on "scale". So while ```<text prefix=" " variable="scale"/>``` returned nothing, the unguarded ```<text term="scale"/>``` always inserted the German word "Maßstab" into my citations.
This patch request makes sure that the descriptive text is only inserted if there actually is some scale information present.

### DIFF
--- a/din-1505-2.csl
+++ b/din-1505-2.csl
@@ -210,10 +210,14 @@
     </group>
   </macro>
   <macro name="scale">
-    <group>
-      <text term="scale"/>
-      <text prefix=" " variable="scale"/>
-    </group>
+    <choose>
+      <if variable="scale">
+        <group>
+          <text term="scale"/>
+          <text prefix=" " variable="scale"/>
+        </group>
+      </if>
+    </choose>
   </macro>
   <macro name="recipient-show">
     <choose>


adam3smith:
Which reference manager is that? The if condition shouldn't be necessary: group includes an implicit "if" for exactly this case as explained here: http://docs.citationstyles.org/en/stable/specification.html#group

> cs:group and its child elements are suppressed if a) at least one rendering element in cs:group calls a variable (either directly or via a macro), and b) all variables that are called are empty.

heikojansen:
pandoc-2.1.2-19.1.x86_64 and ghc-pandoc-citeproc-0.14.3.1-1.2.x86_64 on openSUSE 42.3 (Repository: http://download.opensuse.org/repositories/devel:/languages:/haskell:/lts:/11/openSUSE_Leap_42.3/)

adam3smith:
Yeah, so I'd want to see if this is still an issue with the newest pandoc release (2.2.1 I believe) and then report to John at pandoc; the code you add should be redundant in a correctly working citeproc.

heikojansen:
Do you want to take this on yourself or do you expect me to test and perhaps report it upstream? If the latter: is pandoc-citeproc (I'm using the latest release here, 0.14.3.1) responsible for this or pandoc itself?

adam3smith:
if you wouldn't mind reporting this upstream, that'd be very helpful and yes, you'd report directly to pandoc-citeproc (and since you have the most recent version there, no need to test with something else) I think ideally with a minimal example.

PR #3559, opened 2018-06-20
Improve edition macro for din-1505-2

heikojansen:
Almost any edition i encountered is either plain numeric or - if not - already contains some (abbreviated) form of the word "edition". So in most cases my German citations have edition statements ending in "... Aufl. Aufl.", with the latter occurrence provided by the <text/> element in the style.
Unfortunately, I have no statistical data to prove that this solution is better. 
But at least the modifications in this patch request are based on how the "edition" is formatted in vancouver.csl.

### DIFF
--- a/din-1505-2-alphanumeric.csl
+++ b/din-1505-2-alphanumeric.csl
@@ -168,10 +168,17 @@
     </group>
   </macro>
   <macro name="edition">
-    <group>
-      <number variable="edition" form="numeric" suffix=". "/>
-      <text term="edition" form="short" plural="false"/>
-    </group>
+    <choose>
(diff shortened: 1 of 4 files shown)


adam3smith:
Yes, this is definitely better; we do this routinely in other styles.

heikojansen:
I just realized that - if this is considered an improvement -  I should apply the same modifications to din-1505-2-alphanumeric.csl, din-1505-2-numeric-alphabetical.csl and din-1505-2-numeric.csl
I'll update this pull request.

heikojansen:
Done

rmzelle:
Thanks!

PR #3560, opened 2018-06-21
Update mediterranean-journal-of-infection-microbes-and-antimicrobials…

dsleptsov:
….csl

Please update the following style


rmzelle:
Thanks!

PR #3556, opened 2018-06-20
Update cultural-geographies.csl

POBrien333:
https://forums.zotero.org/discussion/72439/problems-with-cultural-geographies-humane-style

PR #3563, opened 2018-06-22
Create law-culture-and-the-humanities.csl

POBrien333:
via https://forums.zotero.org/discussion/72439/problems-with-cultural-geographies-humane-style
In addition to style fixes here: https://github.com/citation-style-language/styles/pull/3556


rmzelle:
Thanks!

PR #3567, opened 2018-06-27
CSL for 'Genomics & Informatics'

johnwhere2go:
A CSL for 'Genomics & Informatics'.


rmzelle:
Thanks!

("&" is a special character in XML files such as these, so it needs to be 'escaped' as `&amp;`)

PR #3569, opened 2018-06-27
Create the-journal-of-nervous-and-mental-disease.csl

POBrien333:
via https://forums.zotero.org/discussion/72291/i-need-cls-file


rmzelle:
Thanks!

PR #3571, opened 2018-06-29
Update african-online-scientific-information-systems-harvard.csl

POBrien333:
some fixes needed via: https://forums.zotero.org/discussion/comment/311611/#Comment_311611


rmzelle:
Thanks!

PR #3576, opened 2018-06-29
Update accessed field for Incluenza and Other Respiratory Diseases style

jasonjensen:
They are now using the more standard "Accessed" now, as opposed to "cited".


rmzelle:
Confirmed. Thanks!

PR #3578, opened 2018-06-30
APA Fixes

adam3smith:
Fix Spanish "Retrieved from" and in-text author sorting.
(My original idea didn't work out because we're using "retrieved" and "from" split with the access date in the middle for webpages).

### DIFF
--- a/apa-annotated-bibliography.csl
+++ b/apa-annotated-bibliography.csl
@@ -44,6 +44,11 @@
       <term name="translator" form="short">trans.</term>
     </terms>
   </locale>
+  <locale xml:lang="es">
+    <terms>
+      <term name="from">de</term>
+    </terms>
+  </locale>
   <macro name="container-contributors">
     <choose>
       <if type="chapter paper-conference entry-dictionary entry-encyclopedia" match="any">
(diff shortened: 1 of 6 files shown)


rmzelle:
Thanks!

apa-annotated-bibliography.csl
```diff
@@ -679,7 +684,7 @@
   </macro>
   <citation et-al-min="6" et-al-use-first="1" et-al-subsequent-min="3" et-al-subsequent-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year" givenname-disambiguation-rule="primary-name">
     <sort>
-      <key macro="author"/>
+      <key macro="author" names-min="8" names-use-first="6"/>
```
rmzelle: @adam3smith, I assume you checked the rules here, but isn't it a bit strange that the sorting rules differ from the number of names shown?

adam3smith: Manual says explicitly same order in text as in the bibliography (and I recall vaguely that the reason this came up is a mis-ordered 2001a/2001b

PR #3207, opened 2017-12-20
universidade-federal-do-espirito-santo-abnt-autoria-abreviada.csl

vict0rg:

PR #3581, opened 2018-07-03
Add "Universidade Federal do Espírito Santo" ABNT styles

rmzelle:
Closes #3207 (original repository seems to have been deleted)

* Changed file names (my current preference is to use "abnt" as a label at the end, similar to what we've started doing with "harvard" and "apa")
* Added documentation link
* Removed "strip-periods" from titles
* Corrected template link


rmzelle:
Thanks, bud!

PR #3586, opened 2018-07-05
Create asia-pacific-journal-of-human-resources.csl

POBrien333:
via https://forums.zotero.org/discussion/72654/style-request-asia-pacific-journal-of-human-resources


rmzelle:
Thanks!

PR #3588, opened 2018-07-06
Update polish-legal.csl

Jack1982:
Various bug fixes. Making style compatible with most of publishing houses requirements in the field of law.

### DIFF
--- a/polish-legal.csl
+++ b/polish-legal.csl
@@ -66,7 +66,8 @@
     <!-- Tytuł. -->
     <group delimiter=", ">
       <text variable="title" text-case="capitalize-first" font-style="italic"/>
-      <names variable="translator" delimiter=", " prefix="tłum. ">
+      <names variable="translator">
+        <label form="short" suffix=" "/>
         <name and="text" delimiter=", " initialize-with=". "/>
       </names>
     </group>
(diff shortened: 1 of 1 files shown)


polish-legal.csl
```diff
@@ -62,16 +59,18 @@
       </substitute>
     </names>
   </macro>
-  <macro name="title">
-    <!-- Tytuł. -->
-    <text variable="title" text-case="capitalize-first" font-style="italic"/>
+  <macro name="title"> <!-- Tytuł. -->
+    <group delimiter=", ">
+      <text variable="title" text-case="capitalize-first" font-style="italic"/>
+      <names variable="translator" delimiter=", " prefix="tłum. ">
```
adam3smith: I'd prefer to see `prefix="tłum. "` done with a label rather than a prefix.

rmzelle: @adam3smith, can you check my commit, and merge if it looks okay? The translation in the Polish CSL locale file for the short form of the "translator" term is already "tłum." for singular and plural, so we just had to use the existing term (https://github.com/citation-style-language/locales/blob/b8c370c8a978790d2aeefa302f05f3bfb1478e75/locales-pl-PL.xml#L249).

adam3smith:
Thanks both!

PR #3589, opened 2018-07-06
Update african-online-scientific-information-systems-harvard.csl

POBrien333:
small fixes via https://forums.zotero.org/discussion/comment/312006/#Comment_312006


adam3smith:
Thanks!

PR #3590, opened 2018-07-06
APA uncertain date

bwiernik:
Add uncertain dates to APA and also format text-case on status variables

### DIFF
--- a/apa-annotated-bibliography.csl
+++ b/apa-annotated-bibliography.csl
@@ -526,7 +526,7 @@
       </else-if>
       <else>
         <text value="Original work published"/>
-      <else>
+      </else>
     </choose>
   </macro>
   <macro name="edition">
--- a/apa-annotated-bibliography.csl
+++ b/apa-annotated-bibliography.csl
@@ -393,29 +393,55 @@
       <if type="bill legal_case legislation" match="none">
         <choose>
           <if variable="issued">
-            <group prefix=" (" suffix=")">
-              <date variable="issued">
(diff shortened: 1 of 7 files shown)


adam3smith:
Thanks! Just looked over this quickly, but it looks to me like there are two large repeated code blocks. The one on line 399-416 in standard APA and on line 715-733. They're identical for uncertain + regular date and then uncertain puts [ca. ] around them and regular dates uses (). Could we pack those two in macros to reduce complexity & duplication?

bwiernik:
Okay. The lines you pointed out weren’t the same—they were for issued versus original-date. Pieces of the style use just year for both dates (the old issued-year macro) or conditionally all the parts (the old issues and original-date macros). I didn’t see a cleaner way to handle the combining circa with year and various placements of brackets versus parentheses needs than this.

adam3smith:
See the issued macro in vanilla apa.csl and then the change in the bibliography. Does that make sense? If so, we could do the same for original date.

adam3smith:
Perfect; exactly what I had in mind. 

Yes, this is enough complexity for one pull request, let's please put the other issues in a new one & thanks for working on that. 

Happy to follow your lead on the DOI/URL for books & chapters, although APA does indeed appear quite clear on omitting it: http://blog.apastyle.org/apastyle/2011/02/books-and-book-chapters-what-to-cite.html

bwiernik:
Thanks. I think I'll change to follow the DOI/URL rule and see what copyeditors at non-APA journals do.

PR #3598, opened 2018-07-10
Add the style file for the Quodlibet publisher

openp2pdesign:
https://www.quodlibet.it/

### DIFF
--- a/quodlibet.csl
+++ b/quodlibet.csl
@@ -1,0 +1,608 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" et-al-min="6" et-al-use-first="5" et-al-subsequent-min="6" et-al-subsequent-use-first="5" demote-non-dropping-particle="sort-only" default-locale="it-IT">
+  <info>
+    <title>Quodlibet</title>
+    <id>http://www.zotero.org/styles/quodlibet</id>
+    <link href="http://www.zotero.org/styles/quodlibet" rel="self"/>
+    <link href="http://www.zotero.org/styles/chicago-fullnote-bibliography" rel="template"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks. A couple of questions:
1) We'd really like a proper link to a description of the style. The current link in documentation goes to the thesis documentation for Bologna, but those actually say to contact your advisor for the rules for the bibliography
2) More specifically, I'm quite confused by the et-al rules in this styles. What are the actual rules for bibliography and footnotes (and are they actually different for sorting?)

rmzelle:
> We'd really like a proper link to a description of the style. The current link in documentation goes to the thesis documentation for Bologna, but those actually say to contact your advisor for the rules for the bibliography

@adam3smith, it looks like that is the documentation link of the original style (https://github.com/citation-style-language/styles/blob/405698fab86d8e86bcd16e5c5a473442eefb462b/universita-di-bologna-lettere.csl).

@openp2pdesign, can you give us a bit more information about this style? Who will be using the style, and is your style format described online anywhere?

openp2pdesign:
It is the style of the [Quodlibet publisher](https://www.quodlibet.it/), I created it for a book I wrote for this publisher. There is no online documentation for the style, I received the instructions from the publisher and found that it was basically a modified version of the [University of Bologna style](https://github.com/citation-style-language/styles/blob/405698fab86d8e86bcd16e5c5a473442eefb462b/universita-di-bologna-lettere.csl) so I created it for writing my book. Hope it can helps... I was not sure about including it, since it's not for a journal but for a book publisher.

rmzelle:
@openp2pdesign, coming back to this, we can only accept this style if we either have public access to the publisher style guidelines (available on the publisher's website or here as an attachment), or if we can link to the relevant document in some other unambiguous way.

I'll close the pull request for now.

PR #3600, opened 2018-07-10
Update canadian-journal-of-fisheries-and-aquatic-sciences.csl

minyanglee:
changed layout delimiter to semicolon to match http://www.nrcresearchpress.com/page/cjfas/authors


rmzelle:
Thanks! (confirmed with a recent article: https://doi.org/10.1139/cjfas-2017-0421)

PR #3601, opened 2018-07-10
Update to No Date abbreviation

fharnois:
I propose a small change to the abbreviation for "No date" from "n.d." to "s.d." whic is the correct format in french. The Provos Guideline does state that they use "n.d." but this is incorrect in french and all other APA-French versions guidelines that I could find use "s.d.". Since this style is used as a reference for many other APA-French style (e.g. Ecole-de-technologie-superieure-APA, UdeM, etc.), I propose the changes here. If it is impossible to change it here, could we make another APA-French style to which other styles could refer instead?

### DIFF
--- a/apa-fr-provost.csl
+++ b/apa-fr-provost.csl
@@ -26,7 +26,7 @@
       </term>
       <term name="editortranslator" form="short">éd. &amp; trad.</term>
       <term name="translator" form="short">trad.</term>
-      <term name="no date" form="short">n.d.</term>
+      <term name="no date" form="short">s.d.</term>
       <term name="in">dans</term>
       <term name="retrieved">repéré</term>
       <term name="from">à</term>


adam3smith:
>  Since this style is used as a reference for many other APA-French style (e.g. Ecole-de-technologie-superieure-APA, UdeM, etc.),

I don't quite understand this. There's a separate UdM APA style that uses s.d., for example? Why not just fix the existing styles that want s.d. I'm a bit reluctant to change this against the style's explicit prescription.

adam3smith:
@rmzelle thoughts? Not happy with either the option to change this for Provost against the guide or to create a new independent style just for this. 

FWIW, my understanding is that `non daté` (n.d.) is perfectly acceptable French,  so it's not wrong, even if sans date (s.d.) is more common and a better translation of the "no date" that n.d. stands for in English. @fharnois -- your school feels strongly about this?

fharnois:
@adam3smith, thank you for the information. I didn't know about the `non daté` in french. Nevertheless, we need to use s.d. since it is a requirement in the school's style guideline.

Would it be easier to just have a full style for Ecole-de-technologie-superieure-APA that would allow us to update it independantly for Provost's?

rmzelle:
Looks like an independent style is the way to go, then, if your school insists on "s.d.".

We wouldn't want to change our Provost CSL style to go against the Provost style guide, although you could always reach out to their editors and ask if they're willing to switch to "s.d." in their next edition.

fharnois:
That solution works for me. I will propose changes to the style ecole-de-technologie-superieure-apa to make it a style of its own without the reference then. 

Thank you.

fharnois:
What is the easiest way to proceed for you. Do you close this pull request and I make all changes to the [existing style](https://github.com/citation-style-language/styles/blob/47c5e9992f045329805d5326ac7d700d01053257/dependent/ecole-de-technologie-superieure-apa.csl) to make it non-dependant and make a new pull request?

adam3smith:
yes. Note that you'll also have to move the file out of the dependent folder. See here: https://help.github.com/articles/moving-a-file-to-a-new-location/ for instructions

PR #3603, opened 2018-07-11
Create own style and move out of /dependent

fharnois:
Create structure of its own for the style based on Provost APA and move it out of /dependant


rmzelle:
(replaces #3601)

rmzelle:
Thanks!

PR #3608, opened 2018-07-12
Create journal-of-jewish-studies.csl

kmpenner:
Edited Society of Biblical Literature csl to match the Journal of Jewish Studies style guide described at https://www.jjs-online.net/authors/style_guide


rmzelle:
Thanks!

PR #3609, opened 2018-07-13
Create ilahiyat-studies.csl

mtarakci:
A fullnote style for Ilahiyat Studies. 
Web page of the journal: http://ilahiyatstudies.org/index.php/journal/index
Link for style sheet: https://drive.google.com/file/d/1qNUR5UZ3QDJkxhJv0-gE39HA5-2I2n4E/view

### DIFF
--- a/ilahiyat-studies.csl
+++ b/ilahiyat-studies.csl
@@ -14,7 +14,7 @@
     <category citation-format="note"/>
     <category field="humanities"/>
     <summary>Chicago format with full notes and bibliography</summary>
-    <updated>2018-07-13T09:49:22+00:00</updated>
+    <updated>2018-07-28T12:49:32+00:00</updated>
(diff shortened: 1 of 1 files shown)


adam3smith:
(Helpful to say what you actually change if you submit a style that's 95% identical to Chicago Manual or another complex style). 

The changes to volume make sense to me, but I don't see why you're removing title case. The journal's style guide clearly applies title casing as in "History of Islamic Philosophy" etc.

adam3smith:
right, but as we discuss in this thread and as per https://www.zotero.org/support/kb/preventing_title_casing_for_non-english_titles there's a solution for Arabic (and if need be mixed) titles that doesn't require users to incorrectly enter data in title case (incorrectly because it will format incorrectly in styles requiring sentence case). Please revert that change.

mtarakci:
I changed titles to Title Case.

adam3smith:
Thanks!

mtarakci:
It's not used anymore. It would be better to delete it to avoid confusion.
See
https://www.ilahiyatstudies.org/journal/announcement/view/64)

PR #3612, opened 2018-07-14
Create universite-de-montreal-musique.csl

stevenB-MF:
This style describes the current citation guidelines used by the Université de Montréal's music department. It will appear on the department's website (http://guides.bib.umontreal.ca/disciplines/149-Musique?tab=620) as well as be linked and described in a printed & electronic book (title LIRE, ÉCOUTER, ÉCRIRE : Initiation à la recherche en musique à partir des méthodes des sciences humaines, ISBN 9782760639393) to be published this fall, used in their methodology class. The style is also expected to be used by other universities or students which will be using the book.


rmzelle:
Thanks!

PR #3614, opened 2018-07-17
Update council-of-science-editors-author-date.csl

bwiernik:
Add issue number for journal articles: https://forums.zotero.org/discussion/71192/only-the-old-version-of-the-style-of-the-journal-of-dental-research-is-available#latest


adam3smith:
Thanks!

PR #3618, opened 2018-07-21
Update Juristische Zitierweise (Stüber)

zuphilip:
- chapter: fix mismatched use of title instead of container-title
- chapter: add first page besides pages from locator, similar to journal article,
e.g. Knemeyer, in: von Mutius, FG für von Unruh, S. 209 (209).
- use non-breakable space between locator and label
- encyclopedia-aricles: fix problem when title is missing
- prefer short-title in footnote citations
- wegpage: add date in bibliography with prefix "Stand: "

### DIFF
--- a/juristische-zitierweise.csl
+++ b/juristische-zitierweise.csl
@@ -40,7 +40,7 @@
     </names>
   </macro>
   <macro name="locator-with-label">
-    <group delimiter="&amp;#160;">
+    <group delimiter="&#160;">
       <label variable="locator" form="symbol"/>
       <text variable="locator"/>
     </group>
@@ -112,7 +112,7 @@
             <text macro="author-note"/>
             <text macro="inbook"/>
             <group delimiter=" ">
-              <group delimiter="&amp;#160;">
+              <group delimiter="&#160;">
                 <text term="page" form="short"/>
                 <text variable="page-first"/>
               </group>
--- a/juristische-zitierweise.csl
+++ b/juristische-zitierweise.csl
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" class="note" default-locale="de-DE">
   <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
--- a/juristische-zitierweise.csl
+++ b/juristische-zitierweise.csl
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" class="note" default-locale="de-DE">
   <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
@@ -17,7 +17,7 @@
     <category citation-format="note"/>
     <category field="law"/>
     <summary>Juristische Zitierweise nach Stüber www.niederle-media.de/Zitieren.pdf</summary>
-    <updated>2017-01-29T14:33:14+00:00</updated>
+    <updated>2018-07-21T09:10:53+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="de-DE">
@@ -40,7 +40,7 @@
     </names>
   </macro>
   <macro name="locator-with-label">
-    <group delimiter=" ">
+    <group delimiter="&amp;#160;">
       <label variable="locator" form="symbol"/>
       <text variable="locator"/>
     </group>
@@ -60,7 +60,7 @@
         <names variable="editor">
           <name form="short" delimiter="/" initialize-with=""/>
         </names>
-        <text variable="title"/>
+        <text variable="container-title" form="short"/>
       </group>
     </group>
   </macro>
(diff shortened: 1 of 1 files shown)


juristische-zitierweise.csl
```diff
@@ -60,7 +60,7 @@
         <names variable="editor">
           <name form="short" delimiter="/" initialize-with=""/>
         </names>
-        <text variable="title"/>
+        <text variable="container-title" form="short"/>
```
zuphilip: I added manually a `container-title-short` but this did not show up here. Is that forbidden for a chapter?

rmzelle: No, CSL 1.0.1 puts no restrictions on which fields can be used with which item types. Might be some kind of bug in the editor. (if it's not too much work, it might help if you could give the exact steps to reproduce this problem)

juristische-zitierweise.csl
```diff
@@ -247,6 +255,7 @@
           <group delimiter=", ">
             <text macro="author"/>
             <text variable="title"/>
+            <date form="text" variable="issued" prefix="Stand: "/>
```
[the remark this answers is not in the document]
adam3smith: Stand would roughly translates to "as it stands as of [date]" so it's somewhat similar to accessed on, but not quite (which is why I assume zuphilip is using it with the issued date)

juristische-zitierweise.csl
zuphilip: Okay, I created an item of type chapter in Zotero and export it to CSL JSON, added there the field `container-title-short` with a value and then import that (without the outermost parenthesis for a set) into the CSL editor, where I don't see any effect of the short container title.

I used this CSL-JSON data
```json
{
	"id": "http://zotero.org/users/96641/items/TFNFUVLI",
	"type": "chapter",
	"title": "Endnotes",
	"container-title": "The World is Open",
	"container-title-short": "TWO",
	"publisher": "Wiley-Blackwell",
	"page": "427-467",
	"source": "Wiley Online Library",
	"URL": "https://onlinelibrary.wiley.com/doi/abs/10.1002/9781118269381.notes",
	"ISBN": "978-1-118-26938-1",
	"note": "DOI: 10.1002/9781118269381.notes",
	"language": "en",
	"issued": {
		"date-parts": [
			[
				"2012"
			]
		]
	},
	"accessed": {
		"date-parts": [
			[
				"2018",
				7,
				2
			]
		]
	}
}
```

Any idea?

juristische-zitierweise.csl
```diff
@@ -40,7 +40,7 @@
     </names>
   </macro>
   <macro name="locator-with-label">
-    <group delimiter=" ">
+    <group delimiter="&amp;#160;">
```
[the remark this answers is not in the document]
zuphilip: I added them by hand in the CSL editor and they still look fine within the editor:

csl-editor-nonbreaking-space
It might simply be that exporting the style from there is then escaping it wrongly.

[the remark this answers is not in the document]
zuphilip: Okay, I checked and it is also in the Code editor wrongly escaped. I will fix these issues w/o the editor.

juristische-zitierweise.csl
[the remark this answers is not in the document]
zuphilip: "Stand" is IMO a data which represents the version/edition of that work, and yes @adam3smith is right that I therefore choose here the variable issued. However, websites are only used rarely in law publications.

zuphilip:
Okay, I updated the style according to your remarks @rmzelle . That should be fine now.

juristische-zitierweise.csl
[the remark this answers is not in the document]
rmzelle: Maybe copy and pasting the non-breaking space character into the delimiter field works. But yeah, it looks like the code escapes the ampersand sign.

juristische-zitierweise.csl
rmzelle: This seems to work fine for me. I loaded this JSON into the visual editor, and edited the style to use

```xml
<text variable="container-title-short"/>
```

which prints "TWO" for the bibliography:

<hr/>

screen shot 2018-08-11 at 01 04 20

zuphilip: Okay, and what happens if you take instead
```
<text variable="container-title" form="short"/>
```
?

rmzelle: screen shot 2018-08-11 at 12 52 13

This looks like a citeproc-js mapping issue. `<text variable="container-title" form="short"/>` pulls in the (journal title) abbreviation for the first item, which is stored as "journalAbbreviation" in the CSL JSON (https://github.com/citation-style-language/csl-editor/blob/1319eeaebb5ca13f001d7773f8cc30a923718466/src/exampleData.js#L56). But "container-title-short" seems to be ignored.

This seems like a similar situation to what we had with "shortTitle" and "title-short". I can raise the issue with Frank.

zuphilip: > This seems like a similar situation to what we had with "shortTitle" and "title-short". I can raise the issue with Frank.

That would be great. Thank you!

Anything we should do in this PR or can it be merged now?

rmzelle: > Anything we should do in this PR or can it be merged now?

Well, `<text variable="container-title" form="short"/>` and `<text variable="container-title-short"/>` are both correct CSL in this case. And as Zotero currently doesn't offer a variable for chapters (https://aurimasv.github.io/z2csl/typeMap.xml#map-bookSection), it doesn't really matter which one we pick here.

zuphilip: Do both variants use `container-title` if the short form is not present?

Do you suggest to switch the form here? Maybe I am missing something else here...?

rmzelle: > Do both variants use container-title if the short form is not present?

I think only the `form="short"` call has fallback to the "long" form if unavailable. See http://docs.citationstyles.org/en/1.0.1/specification.html#text. Pretty sure the CSL specification is silent on fallback for "container-title-short".

zuphilip: Okay. How do you want to finish this PR then? (I would like to complete this soon as I have promised to some PhD candidates here...)

rmzelle:
Thanks!

PR #3619, opened 2018-07-23
Conditional url

bwiernik:
PR to address https://github.com/citation-style-language/styles/issues/2495


adam3smith:
It's still a bit unclear what BRAIN actually wants for internet sources, but this is clearly an improvement.

PR #3623, opened 2018-07-24
Update indian-journal-of-medical-research.csl

POBrien333:
Via https://forums.zotero.org/discussion/72668/indian-journal-of-medical-research-reference-style
Guidelines not updates (style is years old). Style based on recent papers.


indian-journal-of-medical-research.csl
```diff
@@ -237,8 +213,8 @@
     <sort>
       <key variable="citation-number"/>
     </sort>
-    <layout prefix="(" suffix=")" delimiter=",">
-      <text variable="citation-number"/>
+    <layout vertical-align="sup" delimiter=",">
```
adam3smith: what's going on here? Are these superscript or not? Presumably yes, right? If so, pls. remove the `vertical-align="baseline"` below - thanks!

rmzelle: Looks like this should be superscript, yes (checked in https://doi.org/10.4103/ijmr.IJMR_616_18). I'll go ahead and make the change and merge.

rmzelle:
Thanks!

POBrien333:
Thanks for the fix.

PR #3630, opened 2018-07-30
Update polish-legal.csl

Jack1982:
Replaces "section" with "thesis" (PL: "teza") as is customary for citing legal commentaries.

### DIFF
--- a/polish-legal.csl
+++ b/polish-legal.csl
@@ -12,7 +12,7 @@
     <category citation-format="note"/>
     <category field="law"/>
     <summary>A style for legal citations in Poland.</summary>
-    <updated>2018-07-06T08:12:22+01:00</updated>
+    <updated>2018-07-30T16:43:00+02:00</updated>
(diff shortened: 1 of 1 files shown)


polish-legal.csl
```diff
@@ -101,10 +102,20 @@
...
+    <choose>
+      <if locator="section">
+        <group delimiter=" ">
+          <label variable="locator" form="long"/>
+          <text variable="locator"/>
+        </group>
+      </if>
+      <else>
+        <group delimiter=" ">
+          <label variable="locator" form="short" suffix="."/>
```
adam3smith: the  suffix="." isn't really needed here, is it? The periods are already in the abbreviations. If that's right, we could just define teza as the short form of section above and the wouldn't have to have a separate if-else here but 
```
       <group delimiter=" ">
          <text variable="locator"/>
        </group>
```

would just work for everything.

Jack1982: I checked your proposal, works fine. Thank you!

adam3smith:
thanks

PR #3632, opened 2018-08-01
Create geophysical-journal-international.csl

POBrien333:
via https://forums.zotero.org/discussion/72988/style-request-geophysical-journal-international
Guidelines missing a lot, but publicatons are consistent.


rmzelle:
Thanks!

PR #3633, opened 2018-08-03
add note variable for citation and biblio

reagle:
In my CV I need to include venue stats.  This adds it in a parenthesis at end. Includes both citation and bibliography.

### DIFF
--- a/apa-cv.csl
+++ b/apa-cv.csl
@@ -653,6 +653,7 @@
           </group>
         </else-if>
       </choose>
+      <text variable="note" prefix=" (" suffix=")"/>
     </layout>
   </citation>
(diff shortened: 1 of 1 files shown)


adam3smith:
I understand why this is useful for you, but I don't think it generalizes well (starting with the fact that "note" is used differently by different tools). I'm not going to take this, though thanks for putting it up.
(And as a reminder for anyone who comes across this, the file can be downloaded from [here](https://github.com/reagle/styles/raw/0c7f1a5231939cbe071151fe36571d51323a1b11/apa-cv.csl)

reagle:
Could you point to how others use note differently? (I don't know.) Is there an alternative field I should use? Also, this is a CV specific sheet where it's important to provide reputation stats.

PR #3638, opened 2018-08-08
Remove duplicated issued in chicago (author-date)

bwiernik:
`issued` is rendered twice for webpage and post-weblog. This change aligns Chicago (author-date) with Chicago (note) and (fullnote), where these lines have already been removed.

Reported https://forums.zotero.org/discussion/73102/duplicated-date#latest

### DIFF
--- a/chicago-author-date.csl
+++ b/chicago-author-date.csl
@@ -163,11 +163,6 @@
         </else-if>
       </choose>
       <choose>
-        <if type="webpage post-weblog" match="any">
-          <date variable="issued" form="text"/>
-        </if>
-      </choose>
-      <choose>
         <if variable="issued" match="none">
           <group delimiter=" ">
             <text term="accessed" text-case="capitalize-first"/>


adam3smith:
It's a little trickier than this: Chicago Manual wants the full date, including year, repeated in the 17th edition. This used to be just month and day in the 16th edition.

E.g. 
Google. 2016. “Privacy Policy.” Privacy & Terms. Last modified March 25, 2016. http://www.google.com/policies/privacy/.

(This has "last modified" but the rule extends generally). Since we can't actually test for the presence of day/month, I'm not sure how to get this right across the board.

bwiernik:
Oh, okay, so the style is correct then, but just weird when there is no author or non-year date parts.

adam3smith:
I think the ability to test for date parts comes up frequently enough to consider that for the next "big" CSL release. Should be easy enough, I imagine.

bwiernik:
I agree. That would be helpful in a variety of ways.

PR #3639, opened 2018-08-09
Update clinical-physiology-and-functional-imaging.csl

libora:
https://forums.zotero.org/discussion/comment/314347/#Comment_314347

### DIFF
--- a/clinical-physiology-and-functional-imaging.csl
+++ b/clinical-physiology-and-functional-imaging.csl
@@ -33,7 +33,7 @@
     </names>
   </macro>
   <macro name="publisher">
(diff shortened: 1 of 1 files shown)


rmzelle:
@POBrien333, since you submitted this style less than a month ago, could you review these changes?

POBrien333:
Will do when back from holidays.
Thanks Libor for putting in the fix.

POBrien333:
The fixes are correct.
One more:
Line 36: group suffix needs to be a . and not a ;

rmzelle:
Thanks both!

PR #3640, opened 2018-08-09
Update american-journal-of-enology-and-viticulture.csl

curtisimo1:
Book citations were formatted to end with a semicolon.
Updated to end with a period per the journal requirements


rmzelle:
Thanks!

PR #3641, opened 2018-08-09
Webpage in MRM style

fsantini:
This change adds a macro to cite a webpage according to the Magnetic Resonance in Medicine guidelines.


adam3smith:
It will now (before it'd have shown the "anon" macro, which I've now deleted since there's not sign of it in the style nor in the AMA guide it refers to.)

adam3smith:
Thanks!

PR #3644, opened 2018-08-13
Create oxford-university-press-note.csl

POBrien333:
via https://forums.zotero.org/discussion/73005/style-request-oxford-university-press-numbered-notes

### DIFF
--- a/oxford-university-press-note.csl
+++ b/oxford-university-press-note.csl
@@ -11,7 +11,7 @@
     </author>
     <category citation-format="note"/>
     <category field="generic-base"/>
-    <updated>2018-08-13T12:47:55+00:00</updated>
+    <updated>2018-09-18T10:44:24+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
@@ -1110,7 +1110,7 @@
             </choose>
           </if>
           <else-if variable="volume issue" match="none">
-            <text macro="issue"/>
+            <text macro="issue" prefix="("/>
           </else-if>
         </choose>
       </if>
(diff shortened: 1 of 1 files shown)


POBrien333:
yes. need to look into it more. Note style...

POBrien333:
@adam3smith 
That should be it now. :)

rmzelle:
@POBrien333, if this is OUP's only note style, I'd just call it "note" instead of "full note".

And is the absence of a "default-locale" on purpose? https://global.oup.com/academic/authors/author-guidelines/reference-styles/numbered-notes/?lang=en&cc=ch says "We give examples using US spelling and punctuation here.", so I assume they have British and US versions of this style?

(our existing OUP styles are fixed to either en-US or en-GB: https://www.zotero.org/styles?q=oxford%20university%20press )

rmzelle:
> Locale: good spot. There are small differences. Access dates for example.

Where do you see these differences? Do they have a separate en-GB style guide?

rmzelle:
@POBrien333, can you answer my question above?

POBrien333:
Sorry.
Yes there are basically two guides that only differ in some punctuation regarding the access dates (see the Rambow examples in both guides).
UK: https://global.oup.com/academic/authors/author-guidelines/reference-styles/numbered-notes/?lang=en&cc=es
US (linked to at the bottom in the UK style guide): https://global.oup.com/academic/authors/author-guidelines/reference-styles/numbered-notes/?cc=es&lang=en&

rmzelle:
> Yes there are basically two guides that only differ in some punctuation regarding the access dates (see the Rambow examples in both guides).

Any thoughts on what to do here?

If you can address the date localization in the locale sections of this independent style (with "en-US" and "en-GB" locale sections), we can fix this style to one of those two locales, and create a dependent style for the other.

POBrien333:
@rmzelle 
Would it also be possible to not set a locale and then having the user choose in their document?

rmzelle:
> Would it also be possible to not set a locale and then having the user choose in their document?

Sure, that works too. Do you know whether

> you can address the date localization in the locale sections of this independent style (with "en-US" and "en-GB" locale sections)

?

POBrien333:
Yep. Will have a go at it. :)

POBrien333:
@rmzelle 
Just checked. The style does everything correct already once the correct locale is chosen, so this should be ready to go as it is. :)

rmzelle:
Thanks!

PR #3647, opened 2018-08-14
Create bulletin-of-the-seismological-society-of-america.csl

POBrien333:
via https://forums.zotero.org/discussion/73051/style-request-bulletin-of-seismological-society-of-america#latest

### DIFF
--- a/bulletin-of-the-seismological-society-of-america.csl
+++ b/bulletin-of-the-seismological-society-of-america.csl
@@ -1,0 +1,187 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>Bulletin of the Seismological Society of America</title>
+    <title-short>BSSA</title-short>
+    <id>http://www.zotero.org/styles/bulletin-of-the-seismological-society-of-america</id>
+    <link href="http://www.zotero.org/styles/bulletin-of-the-seismological-society-of-america" rel="self"/>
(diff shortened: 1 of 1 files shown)


bulletin-of-the-seismological-society-of-america.csl
```diff
@@ -0,0 +1,186 @@
...
+          <choose>
+            <if type="paper-conference">
+              <text variable="event-place"/>
+            </if>
+          </choose>
+        </group>
+      </if>
+    </choose>
+  </macro>
+  <macro name="mapinfo">
```
adam3smith: Is this from the template style? Can we do better here? I really don't like using `note` in styles.

bulletin-of-the-seismological-society-of-america.csl
```diff
@@ -0,0 +1,186 @@
...
+    <choose>
+      <if type="map">
+        <text variable="note"/>
+      </if>
+    </choose>
+  </macro>
+  <macro name="accessed">
+    <choose>
+      <if type="webpage post-weblog">
```
adam3smith: needs `match="any"` (otherwise defaults to `match="all"` even though Zotero/citeproc-js may not implement that correctly.

POBrien333: I wonder how that would've happened. I did this in the visual editor like always (click on plus and add post-weblog). thanks for spotting.

bulletin-of-the-seismological-society-of-america.csl
POBrien333: Sure. Guidelines not stating anything, but I guess this could be useful for seismologists.

adam3smith: what info is in there, though? We do, e.g., have "scale" for maps in Zotero&CSL?

PR #3653, opened 2018-08-21
Update wiesbaden-business-school.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/314709/#Comment_314709
Small fix for et al.


adam3smith:
thanks!

PR #3655, opened 2018-08-22
Create synthesis.csl and dependent/synlett.csl

POBrien333:
via https://forums.zotero.org/discussion/73022/style-request-synthesis-and-synlett


adam3smith:
thanks!

PR #3656, opened 2018-08-22
Create brazilian-oral-research.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/314220

basically adapted for article-journal. guidelines don't match perfectly with papers.


rmzelle:
Thanks!

PR #3657, opened 2018-08-22
Create universitat-bremen-lehrstul-fur-innovatives-markenmanagement.csl

POBrien333:
via https://forums.zotero.org/discussion/73057/style-request-lehrstuhl-marketing-uni-bremen


POBrien333:
@rmzelle 
Awesome tool. thanks for pointing that out. 

English locale bits removed.
Thanks.

rmzelle:
Great, thanks!

(there are a bunch of "slugify" tools available online, but I like this one best; the main thing it doesn't do is replace ampersands with "and")

PR #3658, opened 2018-08-23
Create photochemistry-and-photobiology.csl

POBrien333:
via https://forums.zotero.org/discussion/71922/citation-format-for-photochemistry-and-photobiology


rmzelle:
Thanks!

PR #3659, opened 2018-08-24
Create university-of-south-australia-2017-harvard.csl

lachlan-young:
There are existing styles for the University of South Australia institutional version of Harvard citation, from 2010 and 2013, but the most recent 2017 edition of the style has not been added.
This new style builds upon the 2013 version, with changes to the handling of publisher names and legal cases.


rmzelle:
Thanks!

Back when we introduced the 2013 version, we decided it would make sense to keep the 2011 version around as other institutions might still be using the older version. Do you agree it makes sense to keep both the 2011 and 2013 versions around for now?

lachlan-young:
Hi Rintze,

Yes, we’re happy to keep on the older versions for now.

Regards,
Lachlan Young
University of South Australia

PR #3663, opened 2018-08-27
Create Editorial de Imago Mundi

bwiernik:
Titled to differentiate from the Journal published by T&F. Waiting for testing and confirmation of publisher name/URL from requester.

https://forums.zotero.org/discussion/72891/apa-con-nombre-completo-del-autor


rmzelle:
Hi @bwiernik, we're closing this issue since it looks like the requestor never got back to you, and I don't think we'll accept this without a proper documentation link. Note that we can always reopen this issue if we get the requested information.

bwiernik:
Agreed.

PR #3665, opened 2018-08-28

jjrondeau:
The changes bring the style into line with the September 2018 edition of the Thesis Writing Guide. In addition, new document types are now supported by style.

### DIFF
--- a/universite-du-quebec-a-montreal.csl
+++ b/universite-du-quebec-a-montreal.csl
@@ -27,7 +27,7 @@
     <updated>2018-02-08T21:06:06+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
-    <locale xml:lang="fr">
+  <locale xml:lang="fr">
     <terms>
       <term name="editortranslator" form="short">dir. et trad.</term>
       <term name="editor" form="short">dir.</term>
@@ -52,9 +52,9 @@
         <multiple>éditions</multiple>
       </term>
       <term name="edition" form="short">éd.</term>
-      <term name="ordinal">ᵉ</term>
-      <term name="ordinal-01" gender-form="feminine" match="whole-number">ʳᵉ</term>
-      <term name="ordinal-01" gender-form="masculine" match="whole-number">ᵉʳ</term>
+      <term name="ordinal">&#7497;</term>
+      <term name="ordinal-01" gender-form="feminine" match="whole-number">ʳ&#7497;</term>
+      <term name="ordinal-01" gender-form="masculine" match="whole-number">&#7497;ʳ</term>
       <term name="number-of-volumes" form="short"> vol.</term>
       <term name="number">Numéro</term>
       <term name="circa" form="short">ca</term>
@@ -66,7 +66,7 @@
       <if type="chapter paper-conference entry-encyclopedia entry-dictionary map interview webpage" match="any">
         <names variable="editor container-author" delimiter=", " suffix=", ">
           <name and="text" delimiter-precedes-last="never" et-al-use-last="true" initialize-with=". "/>
-          <label form="short" prefix=" (" suffix=")" text-case="lowercase"/>
+          <label form="short" prefix=" (" suffix=")"/>
         </names>
       </if>
       <else-if type="broadcast" match="all" variable="container-title">
@@ -79,7 +79,7 @@
       <if type="article-journal book chapter paper-conference entry-encyclopedia entry-dictionary map" match="none">
         <names variable="editor" delimiter=", " prefix=" (" suffix=")">
           <name and="text" delimiter-precedes-last="never" et-al-use-last="true" initialize-with=". " name-as-sort-order="all"/>
-          <label form="short" prefix=" (" suffix=")" text-case="lowercase"/>
+          <label form="short" prefix=" (" suffix=")"/>
         </names>
       </if>
     </choose>
@@ -87,7 +87,7 @@
   <macro name="author">
     <names variable="author">
       <name and="text" delimiter-precedes-last="never" et-al-use-last="true" initialize-with=". " name-as-sort-order="all"/>
-      <label form="short" text-case="lowercase" prefix=" (" suffix=")"/>
+      <label form="short" prefix=" (" suffix=")"/>
       <substitute>
         <choose>
           <if type="entry-dictionary entry-encyclopedia webpage manuscript legislation legal_case bill treaty article map" match="any">
@@ -105,9 +105,7 @@
         <choose>
           <if type="report">
             <text variable="publisher"/>
-            <group suffix=".">
-              <text variable="title" font-style="italic"/>
-            </group>
+            <text variable="title" font-style="italic" suffix="."/>
           </if>
         </choose>
       </substitute>
@@ -137,7 +135,7 @@
                 <text variable="title" font-style="italic"/>
               </if>
               <else>
-                <text variable="title" quotes="false" font-style="italic"/>
+                <text variable="title" font-style="italic"/>
               </else>
             </choose>
           </substitute>
@@ -170,7 +168,7 @@
               <text term="retrieved" text-case="capitalize-first" suffix=" "/>
               <choose>
                 <if match="any" variable="URL">
-                  <text variable="source" font-style="normal" suffix=" : "/>
+                  <text variable="source" suffix=" : "/>
                 </if>
                 <else>
                   <text variable="source" suffix=". "/>
@@ -187,7 +185,7 @@
   <macro name="title">
     <choose>
       <if type="thesis treaty" match="any">
-        <text variable="title" quotes="false" font-style="italic"/>
+        <text variable="title" font-style="italic"/>
         <text variable="genre" prefix=" (" suffix=")"/>
       </if>
       <else-if type="report" match="any">
@@ -275,7 +273,7 @@
             <group delimiter=" ">
               <text variable="title" font-style="italic"/>
               <group delimiter=" " prefix="(" suffix=")">
-                <text term="version" text-case="lowercase"/>
+                <text term="version"/>
                 <text variable="version"/>
               </group>
             </group>
@@ -299,7 +297,7 @@
             <choose>
               <if match="any" variable="container-title">
                 <group delimiter=" ">
-                  <text variable="title" font-style="normal"/>
+                  <text variable="title"/>
                   <text variable="genre" prefix="[" suffix="]"/>
                 </group>
               </if>
@@ -376,7 +374,7 @@
       </else-if>
       <else-if type="article" match="any">
         <group delimiter=" ">
-          <text variable="title" font-style="normal"/>
+          <text variable="title"/>
           <text variable="number" prefix="(" suffix=")"/>
         </group>
       </else-if>
@@ -445,7 +443,7 @@
       <else-if type="post post-weblog" match="any" variable="container-title">
         <group delimiter=". ">
           <text variable="number" font-style="italic" prefix="Bill " suffix=": "/>
-          <text variable="title" font-style="normal"/>
+          <text variable="title"/>
         </group>
       </else-if>
       <else>
@@ -663,13 +661,13 @@
             </if>
             <else-if match="any" variable="status">
               <group>
-                <text variable="status" text-case="lowercase"/>
+                <text variable="status"/>
                 <text variable="year-suffix" prefix="-"/>
               </group>
             </else-if>
             <else-if type="manuscript graphic legislation" match="none">
               <group>
-                <text term="no date" form="short" text-case="lowercase"/>
+                <text term="no date" form="short"/>
                 <text variable="year-suffix" prefix="-"/>
               </group>
             </else-if>
@@ -781,7 +779,7 @@
         <group delimiter=", " prefix=" (" suffix=")">
           <text macro="edition"/>
           <group>
-            <text term="volume" form="short" text-case="lowercase" suffix=" "/>
+            <text term="volume" form="short" suffix=" "/>
             <number variable="volume" form="numeric"/>
           </group>
           <group>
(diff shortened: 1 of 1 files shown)


universite-du-quebec-a-montreal.csl
```diff
@@ -31,20 +31,33 @@
...
       <term name="presented at">Communication présentée à</term>
-      <term name="page" form="short">p.</term>
+      <term name="page" form="short">
+        <single>p.</single>
+        <multiple>p.</multiple>
+      </term>
       <term name="in press">sous presse</term>
       <term name="director" form="short">réal.</term>
       <term name="chapter" form="short">chap. </term>
+      <term name="scale">Échelle : </term>
```
rmzelle: @jjrondeau, when you define translations, you shouldn't include any separating delimiters (like " : " in this case). So you'd want to replace `<term name="scale">Échelle : </term>` by `<term name="scale">Échelle</term>`.

It also looks like you currently don't use this term yet. Instead of

```xml
<text variable="scale" prefix=". Échelle : "/>
```

it would be better to use:

```xml
<group prefix=". " delimiter=" : ">
  <text term="scale"/>
  <text variable="scale"/>
</group>
```

(the group here prevents the "scale" term from being printed when the "scale" variable is empty)

universite-du-quebec-a-montreal.csl
```diff
@@ -52,9 +52,6 @@
         <multiple>éditions</multiple>
       </term>
       <term name="edition" form="short">éd.</term>
-      <term name="ordinal">&lt;sup&gt;e&lt;/sup&gt;</term>
-      <term name="ordinal-01" gender-form="feminine" match="whole-number">&lt;sup&gt;re&lt;/sup&gt;</term>
-      <term name="ordinal-01" gender-form="masculine" match="whole-number">&lt;sup&gt;er&lt;/sup&gt;</term>
```
rmzelle: @jjrondeau, @adam3smith already removed these lines from your submission, but we don't allow HTML tags like this for superscript.

The best solution we currently have is to use superscript characters. See the French CSL locale for an example: https://github.com/citation-style-language/locales/blob/b8c370c8a978790d2aeefa302f05f3bfb1478e75/locales-fr-FR.xml#L74

rmzelle:
(see https://github.com/citation-style-language/styles/pull/3665/commits/4862ee76f20e2e54c4a46ab2f423878dbd4e2143#r214747469 as well)

universite-du-quebec-a-montreal.csl
```diff
@@ -31,20 +31,33 @@
...
+      <term name="number-of-volumes" form="short"> vol.</term>
     </terms>
   </locale>
   <macro name="container-contributors">
     <choose>
-      <if type="chapter paper-conference entry-encyclopedia entry-dictionary" match="any">
+      <if type="chapter paper-conference entry-encyclopedia entry-dictionary map interview webpage" match="any">
         <names variable="editor container-author" delimiter=", " suffix=", ">
           <name and="text" initialize-with=". " delimiter=", "/>
           <label form="short" prefix=" (" text-case="title" suffix=")"/>
```
adam3smith: this isn't your addition, but these two labels can't possibly be right?

universite-du-quebec-a-montreal.csl
```diff
@@ -55,7 +68,7 @@
   </macro>
   <macro name="secondary-contributors">
     <choose>
-      <if type="article-journal book chapter paper-conference entry-encyclopedia entry-dictionary" match="none">
+      <if type="article-journal book chapter paper-conference entry-encyclopedia entry-dictionary map" match="none">
         <names variable="editor" delimiter=", " prefix=" (" suffix=")">
           <name name-as-sort-order="all" and="text" sort-separator=", " initialize-with=". " delimiter=", " delimiter-precedes-last="never"/>
           <label form="short" prefix=", " text-case="title"/>
```
adam3smith: ditto here

universite-du-quebec-a-montreal.csl
```diff
@@ -74,6 +87,22 @@
             <names variable="author"/>
             <text variable="title"/>
           </if>
+          <else-if type="manuscript legislation bill legal_case treaty" match="any">
```
adam3smith: this seems way to complex. Wouldn't

```
<names variable="editor"/>
<text macro="title"/>
```

or something like that work?

universite-du-quebec-a-montreal.csl
```diff
@@ -82,15 +111,20 @@
...
-            <text macro="title"/>
+            <group suffix=".">
+              <text variable="title" font-style="italic"/>
+            </group>
           </if>
         </choose>
       </substitute>
     </names>
+    <names variable="composer">
+      <name and="text" delimiter-precedes-last="never" et-al-min="8" et-al-use-first="6" et-al-subsequent-min="8" et-al-subsequent-use-first="6" et-al-use-last="true" initialize-with="." name-as-sort-order="all"/>
```
adam3smith: we prefer the et al settings in the bibliography/citation line unless you really need to differentiate et al within citations or the bibliography.

universite-du-quebec-a-montreal.csl
```diff
@@ -82,15 +111,20 @@
...
+    </names>
   </macro>
   <macro name="author-short">
     <names variable="author">
-      <name form="short" and="text" initialize-with=". "/>
+      <name form="short" and="text" et-al-min="3" et-al-use-first="1" et-al-subsequent-min="3" et-al-subsequent-use-first="1" initialize-with=". "/>
```
adam3smith: ditto

universite-du-quebec-a-montreal.csl
```diff
@@ -147,29 +184,60 @@
                   <text variable="URL" prefix="de "/>
```
adam3smith: also not yours originally, but what is the "note" doing here? We like to avoid that.

universite-du-quebec-a-montreal.csl
```diff
@@ -147,29 +184,60 @@
...
-                  <text variable="URL" prefix="de "/>
+                  <group delimiter=" " prefix="de ">
+                    <text variable="source" suffix=" "/>
+                    <text variable="URL"/>
+                  </group>
                 </group>
               </else-if>
               <else>
                 <choose>
                   <if type="article-journal" match="all">
```
adam3smith: why not just continue these with else-if. Also, does the style really need _that_ many different URL forms? Can't we simplify this?

universite-du-quebec-a-montreal.csl
```diff
@@ -293,27 +407,101 @@
...
+              <text variable="genre" prefix="[" suffix="]"/>
+            </group>
+          </else>
+        </choose>
+      </else-if>
+      <else>
+        <choose>
+          <if type="post post-weblog" match="any" variable="container-title">
+            <group delimiter=". ">
+              <text variable="number" font-style="italic" prefix="Bill " suffix=": "/>
```
adam3smith: what is this for? Seems quite odd.

universite-du-quebec-a-montreal.csl
```diff
@@ -333,15 +521,41 @@
...
       <else>
-        <group delimiter=", ">
+        <group delimiter=". ">
           <choose>
-            <if variable="event" match="none" type="chapter patent post-weblog book article-journal post motion_picture broadcast">
+            <if variable="event" match="none" type="chapter patent post-weblog book article-journal post motion_picture broadcast map musical_score figure legislation dataset manuscript article webpage entry-encyclopedia">
               <text variable="genre" prefix="[" suffix="]"/>
+              <text variable="scale" prefix="Échelle : "/>
+              <text variable="number"/>
+              <text variable="annote"/>
```
adam3smith: now you're using both `note` and `annote`? You're really not doing users a favor by making them use a style that requires data entry that breaks everything else.

universite-du-quebec-a-montreal.csl
```diff
@@ -725,6 +1118,26 @@
...
+    <group prefix=" (" suffix=")">
+      <date form="text" variable="original-date">
+        <date-part name="year"/>
+        <date-part name="year"/>
+        <date-part name="year"/>
+      </date>
+      <text variable="year-suffix"/>
+    </group>
+  </macro>
   <citation et-al-min="3" et-al-use-first="1" et-al-subsequent-min="3" et-al-subsequent-use-first="1" disambiguate-add-names="true" disambiguate-add-givenname="true" disambiguate-add-year-suffix="true" givenname-disambiguation-rule="primary-name" collapse="year">
```
adam3smith: don't duplicate the subsequent et al here.

universite-du-quebec-a-montreal.csl
```diff
@@ -738,28 +1151,65 @@
       </group>
     </layout>
   </citation>
-  <bibliography hanging-indent="true" et-al-min="8" et-al-use-first="6" et-al-subsequent-min="8" et-al-subsequent-use-first="6" et-al-use-last="true" entry-spacing="0" line-spacing="1">
+  <bibliography et-al-min="8" et-al-use-first="6" et-al-subsequent-min="8" et-al-subsequent-use-first="6" et-al-use-last="true" entry-spacing="0" hanging-indent="true">
```
adam3smith: subsequent et al makes no sense for bibliography; remove.

universite-du-quebec-a-montreal.csl
jjrondeau: Yes, you're right. I will correct this.

universite-du-quebec-a-montreal.csl
jjrondeau: Thank you for the example. I will use it.

universite-du-quebec-a-montreal.csl
jjrondeau: Yes, I will delete one

jjrondeau: Yes, I will delete one.

universite-du-quebec-a-montreal.csl
jjrondeau: Yes, I will delete one.

universite-du-quebec-a-montreal.csl
jjrondeau: I will delete "annote" and make minimal use of "note".

universite-du-quebec-a-montreal.csl
jjrondeau: Understood.

universite-du-quebec-a-montreal.csl
jjrondeau: Understood.

universite-du-quebec-a-montreal.csl
jjrondeau: Correction made.

universite-du-quebec-a-montreal.csl
jjrondeau: I will avoid "note" as much as possible.

universite-du-quebec-a-montreal.csl
jjrondeau: Understood.

universite-du-quebec-a-montreal.csl
jjrondeau: OK

rmzelle:
@adam3smith, since you reviewed this initially, can you give this another look?

universite-du-quebec-a-montreal.csl
```diff
@@ -276,218 +337,342 @@
...
         <group delimiter=" : ">
           <text variable="publisher-place"/>
           <text variable="publisher"/>
         </group>
       </if>
-      <else-if type="thesis" match="any">
-        <group delimiter=": ">
+      <else-if type="speech" match="any">
+        <group delimiter=", ">
+          <text variable="abstract"/>
```
adam3smith: This is odd -- why would you add the whole abstract for a speech/presentation. I'd reconsider that.

universite-du-quebec-a-montreal.csl
```diff
@@ -276,218 +337,342 @@
...
-              <text variable="number" font-style="italic"/>
-              <text variable="genre" prefix="[" suffix="]"/>
+            <group delimiter=" ">
+              <text variable="title" font-style="italic"/>
+              <group delimiter="; " prefix="[" suffix="]">
+                <text variable="note"/>
+                <text variable="medium"/>
+                <text variable="genre"/>
+              </group>
+              <text variable="number" prefix="(épisode " suffix=")"/>
```
adam3smith: the episode label here seems highly specific and particularly odd for a musical_score type (I guess you have some sort of podcast in mind?) I don't think you're doing your users a favor by being overly specific with this.

universite-du-quebec-a-montreal.csl
```diff
@@ -564,51 +756,127 @@
...
           <group>
             <label suffix=" " variable="page" form="short"/>
             <text variable="page"/>
           </group>
         </group>
+        <choose>
+          <if type="song musical_score" match="any">
+            <choose>
+              <if match="any" variable="container-title">
+                <text variable="medium" prefix=" [" suffix="]"/>
```
adam3smith: the two consecurtive square brackets here seem wrong. What you have above, with the square brackets set on a group, ist almost certainly better.

universite-du-quebec-a-montreal.csl
```diff
@@ -564,51 +756,127 @@
...
+                <text variable="genre" prefix=" [" suffix="]"/>
+                <text variable="number" prefix=" (épisode " suffix=")"/>
```
adam3smith: ditto on the concern with episode

universite-du-quebec-a-montreal.csl
jjrondeau: I agree. This was for Mendeley as there is no field mapped to event variable. I will change this for publisher variable.

universite-du-quebec-a-montreal.csl
jjrondeau: Yes, this is quite specific. I have added these lines based on what is described in the APA style blog for podcast: see comment at the end of the page: https://blog.apastyle.org/apastyle/2012/10/how-to-cite-a-podcast.html#comment-6a01157041f4e3970b01b8d210f59a970c
I don't expect this will be used often, but as there is an Episode Number field in Zotero for podcast (but not for audio recording or musical score), I thought it could be a good idea.

universite-du-quebec-a-montreal.csl
jjrondeau: I agree. Will add a group there.

universite-du-quebec-a-montreal.csl
jjrondeau: As said above, this is based on APA style blog. I would like to keep it if you agree.

adam3smith:
Thanks & sorry for the slow review process!

PR #3667, opened 2018-08-30
Create minimal-grant-proposals.csl

antoncrombach:
A few years ago I created a bare-bones citation style for use in grant proposals. I made it available via Mendeley as I used their citation manager at the time. I regularly receive "thank you" comments from fellow scientists that use the style, and I think I could make their life even more comfortable if this citation style was also available in the Zotero style repository.

For some more info, please see http://anton.cromba.ch/2016/02/07/a-minimal-citation-stylefor-grant-proposals/

### DIFF
--- a/minimal-grant-proposals.csl
+++ b/minimal-grant-proposals.csl
@@ -1,0 +1,91 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+  <info>
+    <title>Minimal style for grant proposals</title>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks! I'm not inclined to take this for the official repository -- there are any number of formats that people may like for their grants and (except for grants with specific guidelines like NIH), we really have no way to say how a "grant proposal" style should look like and how to decide if e.g. requested changes should or shouldn't be made.

Obviously people can still use your style in Zotero: Simply download from https://raw.githubusercontent.com/citation-style-language/styles/69e904cb2977b23bd3c1224dc5f42af2bf732691/minimal-grant-proposals.csl via "save link as" and then double-click the downloaded file to install.

antoncrombach:
Fair enough, I'll turn it into a "gist". Thanks for the quick reply!

adam3smith:
you don't even have to -- you can just use the above link; it'll remain stable.

PR #3671, opened 2018-08-31
Update liverpool-john-moores-university-harvard.csl

POBrien333:
via https://forums.zotero.org/discussion/73400/liverpool-john-moores-university-harvard-referencing-style#latest


adam3smith:
Thanks!

PR #3672, opened 2018-09-01
Update australian-guide-to-legal-citation.csl

POBrien333:
fix blogposts via https://forums.zotero.org/discussion/73366/style-error-australian-guide-to-legal-citation-blog-post
Please review and suggest if this can be done better; not 100% sure on this one. :)


adam3smith:
thanks! Replacing this to keep the comments and also to just handle blogs by adding them to webpage all along (which I meant to do back when I created the style).

POBrien333:
Understood. thanks for taking this on. Quite a complex style, so I wasn't all too comfortable anyway.

PR #3677, opened 2018-09-05

POBrien333:
Was sent an email regarding some fixes which I did based on the guidelines.


rmzelle:
Merci beaucoup!

PR #3678, opened 2018-09-05
Fix locator delimiter in Unified Linguistics style

bwiernik:
https://forums.zotero.org/discussion/73473/unified-style-sheet-for-linguistics-page-number


rmzelle:
Thanks!

(the style's documentation link, https://www.linguisticsociety.org/sites/default/files/style-sheet_0.pdf, is silent on the issue, but it's addressed in the style guide mentioned in the Zotero forum thread (https://www.eva.mpg.de/lingua/pdf/GenericStyleRules.pdf), and this guide references the former)

PR #3680, opened 2018-09-11
Create springer-mathgeos-author-date.csl

fabien-ors:
This style is used by Springer Mathematical Geosciences publications.
Instructions to authors: https://www.springer.com/earth+sciences+and+geography/journal/11004?detailsPage=pltci_3160931

Changes made to the MathPhys original version:
- Added support for "Computer Program" type (known as 'book' type in Zotero considered as a computer program when the "version" variable has a defined value)
- Collapse citations by author
- Alphabetical order by author then ascending by year in citation
- Cosmetics in bibliography

### DIFF
--- a/mathematical-geosciences.csl
+++ b/mathematical-geosciences.csl
@@ -1,21 +1,20 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
   <info>
-    <title>Springer - MathGeos (author-date)</title>
-    <id>http://www.zotero.org/styles/springer-mathgeos-author-date</id>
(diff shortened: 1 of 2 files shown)


adam3smith:
The way I understand this, it's for a specific journal called _Mathematical Geosciences_ not for a whole series of Springer titles. Is that not right? If so, that's fine, of course, we'd just want to adjust the style information accordingly.

fabien-ors:
You are totally right. I try all existing springer styles from the Zotero repository and none of them where exactly corresponding to the Mathematical Geosciences journal recommandations. That's why I have "created" this one derived from MathPys. Maybe another (Springer) journal is using the same style, but I don't know !?

adam3smith:
Great, thanks!

adam3smith:
Thanks!

adam3smith:
@rmzelle -- any idea why this wasn't caught in the tests?

fabien-ors:
Thanks for the fixes.

PR #3681, opened 2018-09-11
universite-de-liege-droit.csl

graulich:
We made some improvements, especially concerning the reports (in the first version some variables were mentioned twice). We have also changed the way we refer to Internet URLs and the viewing date of web pages.
For chapters, the formatting of the "in" preceding the title of the work is now conditional: italics when there is an editor or director, roman type when there is neither editor nor director.
We have added support for references to case law, bill and legislation in the Belgian citation style.
Finally, we have reversed the initial and name of the editor and/or director in the bibliography.

### DIFF
--- a/universite-de-liege-droit.csl
+++ b/universite-de-liege-droit.csl
@@ -25,7 +25,7 @@
         <single>p.</single>
         <multiple>pp.</multiple>
       </term>
-	  <term name="paragraph" form="short">
+      <term name="paragraph" form="short">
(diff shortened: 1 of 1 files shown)


adam3smith:
Looks good to me (obviously I don't test a style like this, so I take your word the changes are correct; I just look for coherent syntax).

graulich:
@adam3smith Thank you for your follow-up, your syntax check and validation of our updated style !

PR #3682, opened 2018-09-11
universite-de-liege-droit-par-categorie.csl

graulich:
This style is dependent from "universite-de-liege-droit.csl" (currently beeing updated). The difference with the parent style is the classification of the bibliography : by document's type, then by author (alphabetical order).


adam3smith:
Dependent styles are identical  to the parent, so this isn't actually a dependent style; I fixed that here. 
@rmzelle - opinions on naming?

graulich:
Thank you very much for changing the category of our style variation as well as for correcting our syntax errors. We apologize for our misunderstanding of the "dependent styles". The difference in this style lies in the classification by document type and then in alphabetical order (author name), which is automatically done within the general bibliography (at the end of any document). Thank you again for your kindness and advice !

graulich:
@rmzelle Thank you very much for your feedback. The title could indeed mislead some people. Thus, we could  name it, as you've suggested it:
- "Université de Liège – Droit (French) (classement par catégorie)" 
- "universite-de-liege-droit-par-categorie" in the URL and 
- "ULiège - Droit - par catégorie" in the short title.
As for the usefulness of this style variation, some publishers allow or ask authors of 'small' contributions to publish a bibliography which, in order not to make the whole thing more cumbersome, is classified only by author (without subdivision according to whether it concerns books, parts of books, articles, etc.). They can then use the style already published. 
However, for larger contributions (as monographs or thesis), publishers ask the general bibliography to be subdivided (doctrine [monograph then parts of books and journal articles], legislation sensu lato and finally case law). In this case, the authors may use this style variation which operates the classification automatically. 
Thank you in advance for your advice and follow-up !

rmzelle:
@graulich, thanks. Those name suggestions seem much clearer to me. @adam3smith, happy with them as well?

rmzelle:
Thanks!

PR #3686, opened 2018-09-13

lachlan-young:
Updates to better align citations for video types with target style guide

### DIFF
--- a/university-of-south-australia-2017-harvard.csl
+++ b/university-of-south-australia-2017-harvard.csl
@@ -15,7 +15,7 @@
     <category citation-format="author-date"/>
     <category field="generic-base"/>
     <summary>University of South Australia citation style based on the January 2017 version of the style guide titled Harvard referencing guide UniSA</summary>
-    <updated>2018-08-20T03:25:46+00:00</updated>
+    <updated>2018-09-13T06:23:02+00:00</updated>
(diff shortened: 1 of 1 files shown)


adam3smith:
The code is fine, but I am curious for which reference manager/data entry you're optimizing this? I don't think Mendeley or Zotero map anything to CSL's director? 

If you're instructing students to use the Extra/Note field with director:, that's great, but in that case I'd only take this if you're actually associated with whoever at UniSA is responsible for the style to make sure that's passed on to students.

lachlan-young:
We're using this CSL within our reading list management software, which does support the director variable. Our style requires that directors of videos are marked, but that other authors (such as production companies) are not marked, so we need a way to distinguish the two kinds of names.
In practice for those reference managers which don't support director, the substitute should cascade over to using the author anyway.

adam3smith:
true, but with the old style, authors were marked as directors in movies and you're now losing that.

lachlan-young:
Ideally, there'd be a CSL type for movies (with directors) and a separate type for other videos cited under corporate authors. But as it is, we can only work with the existing types.
We'd rather have the director label not included for some directors, than to have it mistakenly included for some authors which are not directors.

adam3smith:
OK, thanks!

PR #3688, opened 2018-09-13
Update british-journal-of-political-science.csl

POBrien333:
via https://forums.zotero.org/discussion/73523/style-error-british-journal-of-political-science#latest
BJPS has a new style and moved from note to author-date.


adam3smith:
Thanks!

PR #3689, opened 2018-09-13
update Springer Basic Brackets CSL

krimhild25:
For journal articles the CSL added an extra space before the final period, if a doi was included. Also Springer now wants a "http://" before the doi, so linking out is easier. I added this for "Article by DOI". 
In order to get id of the space I used the code that had in the meantime implemented in the originating CSL (Springer - Basic (author date)).


adam3smith:
perfect, thanks!

PR #3690, opened 2018-09-14
Create aci-materials.csl

POBrien333:
https://forums.zotero.org/discussion/73577/style-request-aci-structural-journal


rmzelle:
@POBrien333, thanks! The ISSNs you used seemed to be for the sister journal "ACI Structural Journal" (see https://www.ncbi.nlm.nih.gov/nlmcatalog?term=0889-3241%5BISSN%5D and https://www.ncbi.nlm.nih.gov/nlmcatalog?term=0889-325X%5BISSN%5D). Also the "Journal" seems to be part of the title. Otherwise good! ;)

PR #3692, opened 2018-09-17
WCAG compliant AGLC (subsequent references)

shreeyaahrc:
Removes Ibid for subsequent references to ensure this version is WCAG compliant.

### DIFF
--- a/australian-guide-to-legal-citation.csl
+++ b/australian-guide-to-legal-citation.csl
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
   <info>
-    <title>Australian Guide to Legal Citation</title>
-    <title-short>AGLC</title-short>
-    <id>http://www.zotero.org/styles/australian-guide-to-legal-citation</id>
-    <link href="http://www.zotero.org/styles/australian-guide-to-legal-citation" rel="self"/>
-    <link href="https://forums.zotero.org/discussion/4841/new-australian-legal-citation-style/?Focus=20831#Comment_20831" rel="documentation"/>
+    <title>Accessibility compliant Australian Guide to Legal Citation</title>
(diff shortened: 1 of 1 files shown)


adam3smith:
Could you say a bit more about the context of this? CSL formats citation style according to existing style guides, so if there is a problem with using ibid for accessibility wouldn't the place to address this be the style guide, not an attempt to automate whatever it prescribes? (And AGLC just published a fourth edition, so this may have changed anyway?)

rmzelle:
@shreeyaahrc, yes, like @adam3smith said, we always do our best to have our citation styles match the official style guide requirements. Please let us know whether you're proposing this change because of some official guidance, by AGLC or another party, or whether this is a personal preference.

shreeyaahrc:
Thanks Adam, Rintze

So, here's the background. AGLC4 still abides by the same ibid rules as AGLC3. However, the Australian government's accessibility requirements mandate that ibid (and see note x above) cannot be used for accessibility reasons. I am working on helping a government agency use Zotero for their referencing needs, and we are trying to figure out how best to massage the AGLC style for Zotero to make this work. One work around that we have looked at is to create an in house style that is available on the agency's servers for download, but given this is a government requirement, there will be other government agencies who will have to comply. In light of that, I was wondering whether it was possible to have it available through the Central repository as AGLC WiCAG compliant.

Thanks

Shreeya

rmzelle:
> the Australian government's accessibility requirements mandate that ibid (and see note x above) cannot be used for accessibility reasons.

Thanks for getting back to us. Can you point us to those government requirements, in particular the section that discusses the use of "ibid."? And do those requirements specifically mention AGLC, or are they more general?

adam3smith:
(If we do this -- and it sounds like there may be a good case to -- I'd feel better if we just describe what the style doesn't do, i.e. no ibid, no supra. WCAG compliance, of course, is massively complex and I don't think we should suggest that using a given citation style will actually do more that a tiny, tiny bit towards compliance.)

rmzelle:
Hi @shreeyaahrc, we're closing this issue because we haven't heard from you in a while. Note that we can always reopen this issue if you get back to us.

PR #3700, opened 2018-09-21
Create university-of-bradford-harvard.csl

dzacca:
Created a style for Bradford's University.

This is based on the style used by University of Abertay Dundee from Gregory Goltsov


adam3smith:
@rmzelle Style looks good to me. For filename/id, what are we doing with Harvard styles now? harvard first or last?

rmzelle:
> For filename/id, what are we doing with Harvard styles now? harvard first or last?

Last. I think it makes most sense to use it as a descriptor for the type of style, which go always last.

rmzelle:
Thanks!

PR #3703, opened 2018-09-24
Fix for missing comma in last group, removal of DOI

Braintelligence:
Hello everyone!
We realized that we had a missing comma in the last group, so we ended up with wrongly styled citations. Also we realized that we don't want DOI in our citation, so we removed it for now.

I hope everything is fine, the CSL was validated properly before the PR.

### DIFF
--- a/german-council-of-economic-experts.csl
+++ b/german-council-of-economic-experts.csl
@@ -30,7 +30,7 @@
     <category citation-format="author-date"/>
     <category field="political_science"/>
     <summary>Style for the German Council of Economic Experts</summary>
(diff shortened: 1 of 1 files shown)


Braintelligence:
Oh, I forgot to change the updated date. 🐱

adam3smith:
I've adjusted how the comma is handled for robustness (with your code, you could have ended up with citations ending on a comma). I wish you would reconsider the DOI -- they're really useful...

Braintelligence:
Currently there are no plans to show the doi in citation. We do save it in Zotero, however.

adam3smith:
too bad... thanks for the style update.

PR #3708, opened 2018-09-26
Update pain.csl

POBrien333:
https://forums.zotero.org/discussion/73815/style-error-pain#latest

Technically the doi is wrong in this style. Shall we take it out?


adam3smith:
We're keeping DOIs in Elsevier styles unless we're being told explicitly to remove them. They appreciate them in copyedit.

adam3smith:
Thanks! I wish they had waited a month with the report though. PAIN and BRAIN just make for such good Halloween pull requests...

PR #3709, opened 2018-09-27
Fixed comma setting for books

Braintelligence:
Example: If we specified a book with an edition, we had a prefixing comma with whitespace between the prepending element and the book. 

We may run into other problems involving strange design choices concerning prefixes of the original citation style, but this is an atomic fix for now. I will submit a PR anytime we encounter actual problems and fix them.

### DIFF
--- a/german-council-of-economic-experts.csl
+++ b/german-council-of-economic-experts.csl
@@ -30,7 +30,7 @@
     <category citation-format="author-date"/>
     <category field="political_science"/>
     <summary>Style for the German Council of Economic Experts</summary>
-    <updated>2017-10-31T15:43:27+00:00</updated>
+    <updated>2018-09-24T18:00:00+02:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="de">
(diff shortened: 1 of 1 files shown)


adam3smith:
Something is off with the pull request. Could you redo this from an up to date fork of the repository?

Braintelligence:
> 
> 
> Something is off with the pull request. Could you redo this from an up to date fork of the repository?

Sure, let me redo real quick.

Braintelligence:
@adam3smith Is it better like this or should I destroy my repo and redo from scratch?

adam3smith:
Yes, that works, thanks. 
(Btw., you should always make pull requests from feature branches so you have a clean copy of master you can go back to)

adam3smith:
(I reopened since this looks good to me).

> It looks as if this line doesn't do anything in Zotero. I also don't understand the logic behind it. In my brain it should do something like <term=volume>.<variable=volume> which doesn't make any sense to me. That's why I investigated and realized it has no effect in Zotero at all. If you remove this line, Zotero still works.
Can someone please give me some insight to what could have been the intention behind this line? Sadly I didn't create the original CSL, but I'm tasked to fix bugs that we start to encounter now.

The line reference got lost here; which line are you referring to?

Braintelligence:
Man, I have a really bad timing 😛, sorry for closing, I just was about to destroy my repo and fork again.

The line reference should be visible here: https://github.com/Braintelligence/styles/commit/4452beb5234952923f8b49169344efeff6ab7ea9

adam3smith:
See my comment wrt. to the volume term. I don't think what you have here is right.

adam3smith:
great, thanks

PR #3710, opened 2018-09-27
Create isse.csl

sebknop:


adam3smith:
Thanks. So what I'm seeing is that the only differences between the German and the English style are
1) the access date format and
2) the , before "and" 

of these, 1) can be handled with a localized date and 2) I don't see in the styleguide, where the , appears before the "und" in the bibligraphy in all cases:

Becker, J., Knackstedt, R., Müller, O., und Winkelmann, A.
Stern, T., und Jaberg, H. 

so it seems to me we should be able to handle this with a single style?

rmzelle:
@adam3smith, can you merge this if this looks okay? (I skipped the "Fakultät für Maschinenbau" in the title for brevity)

adam3smith:
Thanks and sorry for the delay

PR #3713, opened 2018-09-28
Create international-security.csl

POBrien333:
via https://forums.zotero.org/discussion/73834/style-request-international-security#latest

Maybe this is a good opportunity to help me understand these extremely boxed styles better.
How do I achieve this (comma before Vol. 40 and : before doi):
Current: John L. Campbell and Ove K. Pedersen, “The Varieties of Capitalism and Hybrid Success,” <i>Comparative Political Studies</i> Vol. 40, No. 3 (March 1, 2007): pp. 307–32, doi:10.1177/0010414006286542.
Wanted: John L. Campbell and Ove K. Pedersen, "The varieties of capitalism and hybrid success," <i>Comparative Political Studies</i>, Vol. 40, No. 3 (2007), pp. 307, doi.org/10.1177/0010414006286542

### DIFF
--- a/international-security.csl
+++ b/international-security.csl
@@ -108,12 +108,6 @@
       <name and="text" delimiter=", "/>
     </names>
   </macro>
-  <macro name="interviewer">
-    <names variable="interviewer" delimiter=", ">
-      <label form="verb" text-case="capitalize-first" suffix=" "/>
-      <name and="text" delimiter=", "/>
-    </names>
-  </macro>
   <macro name="title-note">
     <choose>
       <if variable="title" match="none">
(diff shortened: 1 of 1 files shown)


adam3smith:
See changes above:
For the comma before the volume, you figure out that the volume is in a "join-with-space" macro for journal articles and then move it to the corresponding "join-with-comma" one, which in this case allows for significant simplifications.

For the comma before pages (not DOI): note that this is a locator, not a page range. The style already does a comma there. 

I also removed the bibliography; you'll still need to do some clean-up here after that since there will be unused macros.

POBrien333:
thanks, Sebastian. Let's see next time if I can figure this out. :)

PR #3719, opened 2018-10-02
"et all" authors was wrong.

luispuerto:
with 3 o more authors, inline citation should show first + et all + year.


luispuerto:
I've done further modifications... so I'll probably upload them in a few. I would like to do some more, but I'm really novel in this thing and I don't know really how to work with this files.

For example, the proceedings entry in the references is done wrong.

PR #3723, opened 2018-10-03
Fix for another problem with commas

Braintelligence:
Hello everyone, we had another problem with commas that I solved again by grouping. Thank you!

### DIFF
--- a/german-council-of-economic-experts.csl
+++ b/german-council-of-economic-experts.csl
@@ -245,11 +245,13 @@
   <macro name="issued">
     <choose>
       <if type="report">
-        <text variable="collection-title" prefix=", "/>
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks!

(and I see @adam3smith already suggested to you in #3709 to use feature branches instead of the "master" branch; if you follow our instructions at https://github.com/citation-style-language/styles/blob/master/CONTRIBUTING.md#3b-submitting-changes-to-an-existing-style, you'll automatically get an up-to-date feature branch for your changes whenever you create a pull request, and you'll usually never have to touch the "master" branch in your fork)

Braintelligence:
You're right. I forgot about that. I'll try to keep it in mind for next time!

PR #3724, opened 2018-10-03
Create antipode.csl

kaymmm:
Info: https://antipodefoundation.org/about-the-journal-and-foundation/a-radical-journal-of-geography/

Spec sheet: https://mc.manuscriptcentral.com/societyimages/anti/Final%20format%20(papers).doc


rmzelle:
@kaymmm, thanks!

kaymmm:
@rmzelle I apologize, it's been several months since I put it together using the visual csl editor and the journal's stylesheet. I believe that I started with the "search by example" tool but don't recall which template the visual editor found for me.

PR #3725, opened 2018-10-04
Create revue-francaise-de-gestion.csl

rpierronnet:
Proposition of a CSL style for french journal of management. 
This style has been developed in close cooperation with the journal board and its editor in chief


rmzelle:
Merci!

rpierronnet:
Thank you :-)

PR #3736, opened 2018-10-10

POBrien333:
as per https://forums.zotero.org/discussion/74023/citation-style-for-journal-is-incorrect


POBrien333:
thanks. didn't think it all the way through.

PR #3741, opened 2018-10-11
excli-journal.csl

guevaracodina:
Format for **EXCLI Journal** (ISSN: 1611-2156) [instructions](http://www.excli.de/instructions_to_authors.php)


adam3smith:
Thanks!

PR #3742, opened 2018-10-12
CMoS: month and year instead of exact date for journal articles

jrwiebe:
As discussed [in the Zotero forum](https://forums.zotero.org/discussion/comment/317879/), the CMoS CSL currently gives the exact publication date for a journal article. This is permitted by the style guide, but in practice it is rarely done. Usually the month and year or the year alone is provided. This PR adopts the month-year style.

### DIFF
--- a/chicago-fullnote-bibliography.csl
+++ b/chicago-fullnote-bibliography.csl
@@ -770,7 +770,7 @@
               </choose>
             </group>
           </if>
-          <if type="legal_case">
+          <else-if type="legal_case">
             <group delimiter=" ">
               <text variable="authority"/>
               <choose>
@@ -783,7 +783,7 @@
                 </else>
               </choose>
             </group>
-          </if>
+          </else-if>
           <else-if type="book bill chapter  legislation motion_picture paper-conference song thesis" match="any">
             <choose>
               <if is-uncertain-date="issued">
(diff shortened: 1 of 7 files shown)


chicago-fullnote-bibliography.csl
```diff
@@ -758,7 +758,19 @@
     <choose>
       <if variable="issued">
         <choose>
-          <if type="legal_case">
+          <if type="article-journal" match="any">
+            <group delimiter=" ">
```
adam3smith: no need for the group here since there's only one element (dates have spacing of their own)

adam3smith:
Thanks -- we'll take this, but it'd be great if you could implement it in all Chicago 17th edition note styles?

chicago-fullnote-bibliography.csl
rmzelle: Removed the group.

rmzelle:
> it'd be great if you could implement it in all Chicago 17th edition note styles?

@adam3smith, can you check whether the changes are correct like this, and squash merge this if it looks okay? I should have edited all the Chicago note styles (including the 16th edition ones), at least those starting with "Chicago" (I e.g. didn't do the T&F one: https://www.zotero.org/styles?q=chicago).

chicago-fullnote-bibliography.csl is the odd one out. It didn't have the `<if type="graphic report" match="any">` conditional in the "issued" macro, and I spotted a small double-space typo in the following `<else-if type="book bill chapter legislation motion_picture paper-conference song thesis" match="any">` line.

PR #3746, opened 2018-10-16
Fixing and changing (see comments)

Braintelligence:
Hello everyone!
So we ran into some problems again 😸, and then basically...
grafik

Currently the CSL seems to work like we want. If we find issues again, I'll push another PR. Sorry for the heavy frequency of PRs in the last weeks; we're in the middle of a project now and keep running into things we want to change. 😼 

Maybe I can refactor the whole thing next year.


rmzelle:
No problem. Thanks!

PR #3751, opened 2018-10-18
Updating

saufmvz:
Changing et-al configurations from bibliography


adam3smith:
Thanks! Question: neither of the documentation links works for me -- since you originally submitted the style, I trust this is correct, but it'd be great if there was a working link to the style -- is that available now (as always, feel free to respond in Portuguese if that's easier).

rmzelle:
Obrigado!

PR #3753, opened 2018-10-22
Create american-journal-of-sonography.csl

POBrien333:
via https://forums.zotero.org/discussion/36537/text-references-with-square-brackets


POBrien333:
Duplication due to GitHub glitch.
First submission here: https://github.com/citation-style-language/styles/pull/3752

PR #3754, opened 2018-10-22
Create journal-of-agricultural-and-resource-economics.csl

precisionedits:
New style file for Journal of Agricultural and Resource Economics


rmzelle:
Thanks!

PR #3755, opened 2018-10-23
Fix author formatting for MDPI

todoricMDPI:
Hi,

I work in MDPI and we notice that CSL don't follow MDPI citation rules.
You can find our rules [here](http://res.mdpi.com/data/mdpi_references_guide_v5.pdf).

Kind regards,
MDPI team.


adam3smith:
Thanks!

PR #3757, opened 2018-10-24
include Patents

joka-de:
added a section for somewhat proper inclusion of patents in citations

### DIFF
--- a/din-1505-2-numeric.csl
+++ b/din-1505-2-numeric.csl
@@ -86,13 +86,6 @@
       <!-- needed: Label should appear as suffix to EVERY name...!-->
     </names>
   </macro>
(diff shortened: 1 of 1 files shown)


rmzelle:
@adam3smith, can you take a look? The whole `<if variable="issued submitted" match="all">` conditional (https://github.com/citation-style-language/styles/pull/3757/files#diff-181ea5ab17da2889c3ac76e555189a10R330) looks a bit complicated, and can probably be removed if we hard-code "und" as a group delimiter instead of using the "and" term. (there are also three other DIN-1505-2 styles: https://www.zotero.org/styles?q=din-1505-2)

(PR didn't validate because of `term="und"`)

adam3smith:
Thanks!

PR #3758, opened 2018-10-24
Update annual-reviews-author-date.csl

POBrien333:
via https://forums.zotero.org/discussion/36537/text-references-with-square-brackets


adam3smith:
Thanks!

PR #3759, opened 2018-10-24
Update nature.csl (remove bold comma behind volume)

POBrien333:
via https://forums.zotero.org/discussion/74121/style-error-nature-and-possibly-subjournals


POBrien333:
I need to stop editing styles when my brain is not fried from work. Sorry.

rmzelle:
To non-boldly go where no one has gone before!

PR #3760, opened 2018-10-25
apa.csl: title case for conference paper (@inproceeding booktitle)

jedbrown:
The current apa.csl downcases the title of the proceedings, thus
requiring users to protect what is usually every word.  From the APA:

    Parsons, O. A., Pryzwansky, W. B., Weinstein, D. J., & Wiens, A. N.
     (1995). Taxonomy for psychology. In J. N. Reich, H. Sands, & A. N.
     Wiens (Eds.), Education and training beyond the doctoral degree:
     Proceedings of the American Psychological Association National
     Conference on Postdoctoral Education and Training in Psychology
     (pp. 45–50). Washington, DC: American Psychological Association.

  https://blog.apastyle.org/apastyle/2012/08/how-to-cite-materials-from-meetings-and-symposia.html

To obtain this one would currently need to protect case as in

  booktitle = {Proceedings of the {American} {Psychological} ...}

The more natural

  booktitle = {Proceedings of the American Psychological ...}

currently results in the title being lower case.  This commit provides
the desired behavior from either formatting.

### DIFF
--- a/apa.csl
+++ b/apa.csl
@@ -650,7 +650,7 @@
   </macro>
   <macro name="container-title">
     <choose>
-      <if type="article article-journal article-magazine article-newspaper" match="any">
+      <if type="article article-journal article-magazine article-newspaper paper-conference" match="any">
         <text variable="container-title" font-style="italic" text-case="title"/>
       </if>
       <else-if type="bill legal_case legislation" match="none">


adam3smith:
"Education and training beyond the doctoral degree:" is part of the proceedings title, though, and in sentence case. I'm pretty sure the rest is title cased only because it's all a proper name. (FWIW, our styles _never_ force lowercase; that's some interpretation of a processor that attempts to wrangle the differing conceptions of bibtex and CSL styles of how casing should be handled).

jedbrown:
@adam3smith Oh, I misinterpreted the behavior of "title" case here. It sounds like you are saying this should be investigated as a pandoc-citeproc issue rather than an apa.csl issue.

adam3smith:
Oh sorry, I had meant to post an update here:
If you are correct that proceedings titles should be in title case, we should indeed change the style (I think pandoc-citeproc's behavior is actually the best interpretation of the different logic followed by bibtex and CSL for casing titles.) But I'm not sure how to interpret the example above, given that "Education and training beyond the doctoral degree:" is in sentence case. @bwiernik do you happen to have any insight on this?

bwiernik:
The current style behavior is correct. Proceedings titles published like books are treated like book titles—only first letter, first letter after colon or em dash, and proper nouns in uppercase. 

To handle BibTeX’s convention of storing titles in Title Case and then down-casing for sentence case, the BetterBibTeX plugin in Zotero does indeed escape every uppercase letter. 

If you are using pandoc, I recommend using CSL-JSON, rather than .bib, to store your reference data to avoid this issue. 

(Tangential issue, proceedings published like journals should be formatted like journal articles with the proceedings title in title case—for example, the IEEE proceedings. I hadn’t addressed this in the major revamp of the APA csl style I submitted. @adam3smith Would good logic be that if the proceedings has an editor to treat it like a book, but otherwise to treat it like a journal?)

adam3smith:
> Would good logic be that if the proceedings has an editor to treat it like a book, but otherwise to treat it like a journal?

Can't say that I've every thought about this, but that makes sense to me.

jedbrown:
@bwiernik Switching all collaborative workflows (including those not using Pandoc) to CSL-JSON is rather burdensome. Do you think it's wrong to ask for Pandoc to process .bib files consistent with native Bib(La)TeX tools?

bwiernik:
There isn’t really “native” BibTeX tools. It’s extremely inconsistently implemented, and pandoc’s behavior isn’t unique. Frankly, the BibTeX convention of storing titles in Title Case and then lowercasing them to sentence case is always going to be fragile and a pain to work with. If you want to use BibTeX, I recommend using a tool like BetterBibTeX to manage it automatically. 

Regarding CSL-JSON versus .bib, keeping a shared set of database files in either format seems to me equivalently difficult and best done using Zotero, GitHub, or similar tools.

jedbrown:
My personal use cases are collaborating on different projects in different git repositories. Getting everyone to adopt the same bib management software is onerous. BibTeX is a de facto standard that everyone can use and that merges fine without special tools.

The use case that prompted opening this issue is slightly different: [JOSS papers](https://joss.readthedocs.io/en/latest/submitting.html#example-paper-and-bibliography) are processed using pandoc-citeproc. I think it's a rather big ask for each author to install software or manually convert their references to CSL-JSON because we are incapable of processing their BibTeX in a way that matches the behavior of their usual toolchain. Chances are that every other journal that these authors submit to will work with LaTeX source and Bib(La)TeX.

Is your position here (and in https://github.com/jgm/pandoc-citeproc/issues/358#issuecomment-436868867) that Pandoc should not support this workflow without requiring authors to manually protect case (not needed with any other toolchain that they use)?

jedbrown:
I don't know what processors were being compared, but there are large communities that have been using it for decades and do not observe this particular problem. So pandoc-citeproc + apa.csl is an outlier among the tools that we actually use. And if citation-js could actually reliable convert (it failed when I tried to run it on the pandoc-citeproc `tests/biblatex-examples.bib` and complains about numerous unrecognized fields when I extract single entries; it does not downcase with or without `langid = {french}` FWIW; the output is also not formatted to be human readable) then why insist that pandoc-citeproc should behave differently?

In short, I don't understand the fatalism or the principle by which there is no bug in the present toolchain, but the only way to get sensible behavior is to jump through unique hoops or switch to a different toolchain.

bwiernik:
I really don’t understand what your criticism is—APA style says to print proceedings/book titles in sentence case. Pandoc and apa.csl are producing that correctly. If you want certain words to remain uppercase, then you need to escape them. That is just how BibTeX works. Your suggestion in the other thread is incorrect—proceedings titles should be follow casing rules like every other item.

adam3smith:
Yeah, I think @jedbrown you're really just misunderstanding APA requirements and that cascaded down here. If proceeding titles need to be in sentence case (as really is the case; I double checked the manual), protecting proper nouns in bibtex is the only way to achieve that, regardless of whether you use bst or csl styles. You just picked the part of the proceedings that was chock-full of proper nouns and ommitted the first part of the proceedings title ("Education and training beyond the doctoral degree"), which would have clarified this, from your bibtex MWE. I'm not seeing a bug or even incompatibility anywhere.

I

jedbrown:
The issue is that many/most existing bibtex files (including those produced by commonly used tools and the example distributed with pandoc-citeproc) do not protect the proper nouns in the `booktitle` and the other tools they are likely to use do not require it either. I understand your position is that all those other tools and bibtex files, including pandoc-citeproc's example (but not its implementation), are broken. I can stipulate that, though I don't think it's useful.

bwiernik:
The examples in pandoc-citeproc escape all of the words in titles that should be uppercased. https://github.com/jgm/pandoc-citeproc/blob/master/tests/biblatex-examples.bib

It has always been the case with BibTeX that protecting uppercasing requires escaping the word. This simply is the expected behavior in BibTeX.

In any event, I’m not really interested in continuing to argue about this.

adam3smith:
Let me try to phrase this differently: My understanding is that the example citation that you yourself posted in your first post _cannot be rendered correctly without escaping in any bst or CSL style_. I'm happy to reconsider if I'm wrong, so maybe you could post the .bib that you think produces

>Parsons, O. A., Pryzwansky, W. B., Weinstein, D. J., & Wiens, A. N.
 (1995). Taxonomy for psychology. In J. N. Reich, H. Sands, & A. N.
 Wiens (Eds.), Education and training beyond the doctoral degree:
 Proceedings of the American Psychological Association National
 Conference on Postdoctoral Education and Training in Psychology
 (pp. 45–50). Washington, DC: American Psychological Association.

The key here being that "Education and training beyond the doctoral degree:
 Proceedings of the American Psychological Association National
 Conference on Postdoctoral Education and Training in Psychology" is capitalized exactly as in the APA example. The fact that there is a lot of bibtex (and other citation metadata) out there that is careless about capitalization is only relevant _if that data can be turned into proper citations_. My claim is that it cannot be.

jedbrown:
My apologies. My test using biblatex/biber did not use equivalent localization; with that fixed, its behavior is basically consistent with pandoc-citeproc/apa.csl (more escaping is needed). Understood that classic bibtex does not fully support APA.

PR #3762, opened 2018-10-28
Update invisu.csl

amendesd:
Mainly corrected the disambiquate macro and made other minor corrections.


adam3smith:
Thanks! (the change I made applies the term to other en, de, es locales, i.e. will work for both GB and US English, etc. I can't see any downside.

PR #3764, opened 2018-10-30
Create zeitschrift-fur-medien-und-kulturforschung.csl

tillheilmann:
Added style for Zeitschrift für Medien- und Kulturforschung (ZMK).

### DIFF
--- a/zeitschrift-fur-medien-und-kulturforschung.csl
+++ b/zeitschrift-fur-medien-und-kulturforschung.csl
@@ -455,7 +455,7 @@
       </else>
     </choose>
   </macro>
-  <citation delimiter-precedes-last="never" delimiter-precedes-et-al="never" and="text" et-al-subsequent-min="4" et-al-subsequent-use-first="1" disambiguate-add-names="true" disambiguate-add-givenname="true">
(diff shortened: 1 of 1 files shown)


adam3smith:
Is the style documented anywhere? Under the documentation link, I'm only seeing stylesheets for other journals

zeitschrift-fur-medien-und-kulturforschung.csl
```diff
@@ -0,0 +1,510 @@
...
+      </if>
+      <else>
+	<group prefix=", " delimiter=".&#160;">
+	  <label variable="locator" form="short"/>
+	  <text variable="locator"/>
+	</group>
+      </else>
+    </choose>
+  </macro>
+  <citation delimiter-precedes-last="never" delimiter-precedes-et-al="never" and="text" et-al-subsequent-min="4" et-al-subsequent-use-first="1" disambiguate-add-names="true" disambiguate-add-givenname="true">
```
adam3smith: Just to make sure I understand the goal here correctly: the idea is to not use et al ever on first citation? That seems risky? Would they really want 12 authors in a footnote?

adam3smith:
Thanks & looks good in general. Two questions above. Also see if my edits make sense. They shouldn't change any actual citation given any reasonably conceivable set of metadata input.

adam3smith:
Thanks!

PR #3766, opened 2018-10-30
Create universitat-basel-deutsche-sprachwissenschaft-v07.csl

wederm:
Citation style for German Linguistics, Departement of Linguistics and Literature, University of Basel (Mirjam Weder mirjam.weder@unibas.ch)

### DIFF
--- a/universitat-basel-deutsche-sprachwissenschaft-v07.csl
+++ b/universitat-basel-deutsche-sprachwissenschaft-v07.csl
@@ -1,0 +1,654 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
+  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>Universität Basel, Deutsche Sprachwissenschaft, v0.7</title>
+    <title-short>UniBas Linguistik</title-short>
+    <id>http://www.zotero.org/styles/universitat-basel-deutsche-sprachwissenschaft-v07</id>
+    <link rel="self" href="http://www.zotero.org/styles/universitat-basel-deutsche-sprachwissenschaft-v07"/>
+    <author>
+      <name>Sorin Marti</name>
+      <email>sorin.marti@stud.unibas.ch</email>
+    </author>
+    <contributor>
+      <name>Mirjam Weder</name>
+      <email>mirjam.weder@unibas.ch</email>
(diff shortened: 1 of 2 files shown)


adam3smith:
Thanks! A couple of questions:
1. Is the style documented somewhere, ideally online?
2. The et al settings that you have added caused the style to fail our tests. Could you describe the behavior you were trying to achieve for et al?

sorinmarti:
Hey adam3smith

Thank you for your work and questions.

1.) The style is not yet documented.

2.) The german "et-al" term in Zotero ist different than "et al." but I wanted it to be "et al." and not another term meaning the same (like "u. a." which means "and others" in german).

Second: In the short form I wanted the et-al to show up after 5 names when first mentioned and after 3 names when mentioned another time. In the long form, always all names should be mentioned.

Example:
In their book they mention it first. (Brown, Farmer, Green, Miller, Smith et al.) it is a really good book and they are explaining a lot of things twice. (Brown, Farmer, Miller et al.)

sorinmarti:
Follow up:

The part of the University of Basel which wants to use the submitted citation style has a document with guidelines on how to create written papers. The use of Zotero is described within this document. This is not a description of the style itself but a user guide on what style to use.

Is this what you are looking for?

The document can be found at:
https://germanistik.philhist.unibas.ch/fileadmin/user_upload/germanistik/Merkblaetter_und_Dokumente/Leitfaden_DeutscheSprachwissenschaft_HS18._Kopie.pdf

sorinmarti:
I was instructed to implement the "et-al" in the way I described. I will ask the responsible person why this requirement differs from the Leitfaden. I know they are in the process of updating the document, so this might be one of the changes. I will get back to you soon.

Most of the papers written with this style will be in German. Despite that, I would like to have a style which could work for all our official languages in Switzerland (German, French, Italian) and English. Also I would like to design the style in a way it can be used by other faculties (which might be writing papers mainly in English or, say, French). Does that make sense?

In my tests, French and English worked fine.

sorinmarti:
I just talked to the responsible person and we will update the Leitfaden and publish it under the same URL. This may take some weeks, though.

Et-al-rules are:
- When first mentioned, display all author names. If it are more than five, display five and "et al."
- Subsequently display a maximum of 3 authors if there is more than one.

I think this should be implemented in the csl file like so.

Thank you very much for your feedback and the work you put into this. It is highly appreciated.

sorinmarti:
Thank you!

Please let me know if any further steps are required from our side.

adam3smith:
Thanks! If you submit changes in the future, remember to always start with the current version of the style in the repository.

PR #3767, opened 2018-10-31
Update bursa-uludag-universitesi-saglik-bilimleri-enstitusu.csl

ozhansaglik:


bursa-uludag-universitesi-saglik-bilimleri-enstitusu.csl
```diff
@@ -635,7 +635,7 @@
       </group>
     </layout>
   </citation>
-  <bibliography delimiter-precedes-et-al="never" delimiter-precedes-last="always" et-al-min="6" et-al-use-first="1" subsequent-author-substitute="&#8212;&#8212;&#8212;" entry-spacing="0" line-spacing="2" hanging-indent="true">
+  <bibliography et-al-min="4" et-al-use-first="3" entry-spacing="0" line-spacing="2" hanging-indent="true">
```
adam3smith: So this seems very odd to me -- you'd list up to  authors in first citations, but no more than 3 authors in the bibliography?

ozhansaglik: Hi Adam,
On inline citations, if there are three or more authors, we are writing the first author and use "ve ark." (abbreviation of et-al in Turkish) like Saglik ve ark. 

But, in the bibliography if there are three or more authors we are writing 3 of them and use "ve ark." like Smith A., Saglik O., Bayraktar, A ve ark.

You can see the examples via link the page 16 and 17, but it is in Turkish. (https://www.uludag.edu.tr/dosyalar/saglikbilimleri/duyuruimages/2018%20Duyurular/SA%C4%9ELIK%20B%C4%B0L%C4%B0MLER%C4%B0%20ENST%C4%B0T%C3%9CS%C3%9C%20TEZ%20YAZIM%20KLAVUZU%202016-BURSA-edited.pdf)

ozhansaglik: Yes, you are right. Now, did you correct the codes as you mention?

adam3smith:
OK, thanks

PR #3770, opened 2018-11-06

cookie365:
Fixed problem if author is repressed, then there is a blank space before the year in the citation for example ( 2019), which should be (2019)


adam3smith:
Thanks!

PR #3775, opened 2018-11-08
Update invisu.csl

amendesd:
I realize that I forgot the locators in the subsequent citations. Here is the update. Sorry for the extra job.
Thank you very much!


adam3smith:
Thanks

PR #3776, opened 2018-11-09
Create sedimentology.csl as new style for journal 'Sedimentology'

andreas-paul:
Created a style for the journal 'Sedimentology', for which I wasn't able to find a replacement in the current range of styles. Link: https://onlinelibrary.wiley.com/journal/13653091. Happy to further edit and/or change. Adheres to style requirements and passed validation.

### DIFF
--- a/sedimentology.csl
+++ b/sedimentology.csl
@@ -7,7 +7,7 @@
     <link href="http://www.zotero.org/styles/sedimentology" rel="self"/>
     <author>
       <name>Andreas Paul</name>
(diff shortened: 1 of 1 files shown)


rmzelle:

andreas-paul:

Hi there,

It has been a long time ago since I created this style for personal use, at a time when I really only knew half of what I was doing in the CSL editor. I know that I used an existing style for this but I did not make note of which one, unfortunately.

rmzelle:
Okay, no problem. Thanks!

PR #3715, opened 2018-10-02
bursa-uludag-universitesi-egitim-bilimleri-enstitusu.csl

ozhansaglik:
Citation styles for "Bursa Uludag University Egitim Bilimleri Enstitusu"

PR #3778, opened 2018-11-12
Update bursa-uludag-universitesi-saglik-bilimleri-enstitusu.csl

ozhansaglik:
We need to exclude (Turkish) from the title.

### DIFF
--- a/bursa-uludag-universitesi-saglik-bilimleri-enstitusu.csl
+++ b/bursa-uludag-universitesi-saglik-bilimleri-enstitusu.csl
@@ -2,7 +2,7 @@
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" page-range-format="expanded" demote-non-dropping-particle="never" sort-separator=",  " default-locale="tr-TR">
   <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
-    <title>Bursa Uludağ Üniversitesi - Sağlık Bilimleri Enstitüsü (Turkish)</title>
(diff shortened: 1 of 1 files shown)


adam3smith:
All CSL styles with a non-English decault locale have the language in parentheses in the repository. Why would you want to remove it?

ozhansaglik:
Because our other styles do not include Turkish in parentheses. 

mendeley_desktop

rmzelle:
@ozhansaglik, like @adam3smith says, we always include the language in the title for CSL styles fixed to a non-English language.

In your recent update of the "Bursa Uludağ Üniversitesi - Sağlık Bilimleri Enstitüsü" style (https://github.com/citation-style-language/styles/pull/3767), you added `default-locale="tr-TR"` to that style, which fixes the style to Turkish. As a result, we added the "(Turkish)" to the style title.

Was the addition of `default-locale="tr-TR"` an intentional change? When you first contributed the styles for Bursa Uludağ Üniversitesi, you indicated that the styles could also be used in English (https://github.com/citation-style-language/styles/pull/3715#issuecomment-427739365), in which case we wouldn't want to fix any of the styles to Turkish.

ozhansaglik:
It is my fault sorry for the disturbance. I did not add "default-locale="tr-TR" intentionally, also I did not examine if there is a code like this.

Can you delete it or I have to delete it?

Sorry again.

Best

Özhan

rmzelle:
I just reverted the change: https://github.com/citation-style-language/styles/commit/41cb8d2df5c582f008703c01abe9833b18684a6b

Should show up in Zotero/Mendeley within a day.

PR #3781, opened 2018-11-14
Add "et al." after 10 authors

todoricMDPI:
Hi,

I work in MDPI and we notice that CSL don't follow MDPI citation rules.
You can find our rules [here](http://res.mdpi.com/data/mdpi_references_guide_v5.pdf).

Kind regards,
MDPI team.


rmzelle:
Hi, it doesn't matter in this case anymore since I already accepted #3782, but in the future it's handier to update your original pull request if you made a small mistake, instead of opening a new one. The comment at the top explains how: https://github.com/citation-style-language/styles/pull/3781#issuecomment-438620437

PR #3784, opened 2018-11-15
ikc.csl

omerfarukgorucu:
İzmir Katip Çelebi Üniversitesi Sosyal Bilimler Enstitüsü

### DIFF
--- a/IKC - SBE.csl
+++ b/IKC - SBE.csl
@@ -1,0 +1,1448 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" delimiter-precedes-et-al="never" delimiter-precedes-last="always" et-al-min="4" et-al-use-first="1" page-range-format="chicago" default-locale="tr-TR">
+  <info>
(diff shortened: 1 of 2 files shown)


adam3smith:
Why did you take out the diacritics and change it to lowercase? That doesn't make sense to me. Please ask questions if things aren't clear, we're happy to help, but don't just change things around randomly.

omerfarukgorucu:
adam3smith, i'm sorry. You wasted your time because of my fault. I'm new at this stuff. I was not sure whether some Turkish characters would be accepted. On the other hand i'm trying to figure out Github system. I need your help to save our csl style to citationstyles.org. What's the next step to make this possible?

omerfarukgorucu:
@rmzelle, #3783 and #3784 actually use same style. These are multiples. So we can use #3784 as just "Sosyal Bilimler Enstitüsü" and remove the other one. 

We based our style on "Chicago Manual of Style 17th edition (full note)" http://www.zotero.org/styles/chicago-fullnote-bibliography

rmzelle:
Thanks!

And to answer your earlier questions and remarks, special characters are allowed in style titles. They just need to be removed from the file name and style ID. And the easiest way to submit styles is by following our instructions at https://github.com/citation-style-language/styles/blob/master/CONTRIBUTING.md.

PR #3787, opened 2018-11-15
Create TUBITAK-proje-basvuru-2018-kasim-tasdemir.csl

aspirintr:
This style can be used in project proposals to be submitted to TÜBİTAK (The Scientific and Technological Research Council of Turkey). This guideline is followed: http://www.tubitak.gov.tr/ardeb-kaynakca  (last accessed: 15-11-2018)

### DIFF
--- a/TUBITAK-proje-basvuru-2018-kasim-tasdemir.csl
+++ b/TUBITAK-proje-basvuru-2018-kasim-tasdemir.csl
@@ -1,0 +1,746 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" page-range-format="expanded" demote-non-dropping-particle="never" default-locale="tr" xmlns="http://purl.org/net/xbiblio/csl">
+  <info>
+    <title>TÜBİTAK 2018 - Kasim Tasdemir</title>
+    <title-short>TÜBİTAK'18</title-short>
+    <id>https://csl.mendeley.com/styles/18815441/tubitak-2018-kasim-tasdemir</id>
+    <author>
+      <name>kasim tasdemir</name>
+      <email>kasimtasdemir@gmail.com</email>
+    </author>
+    <contributor>
+      <name>kasim tasdemir</name>
+      <uri>https://www.mendeley.com/profiles/kasim-tasdemir/</uri>
+      <email>kasimtasdemir@gmail.com</email>
+    </contributor>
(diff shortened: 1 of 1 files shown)


rmzelle:
@aspirintr, thanks. I changed the style title and file name a bit to be more in line with our guidelines. Let me know if it's correct like this.

And did you base this on an existing CSL style? If so, we'd like to know which one, so we can add a "template" link (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link). Was it https://github.com/citation-style-language/styles/blob/master/apa.csl ?

aspirintr:
@rmzelle Thank you for corrections. Yes, it is perfectly fine.

Yes, it is based on APA and now I have added a link to that style as "<link href="https://github.com/citation-style-language/styles/blob/master/apa.csl" rel="template"/>" See line 9. I had already mentioned APA in the summary but I did't know that I had to add a link to it. Sorry.

Thanks.

aspirintr:
@rmzelle I think that would be useful to keep TUBITAK (abbreviation for Turkiye Bilimsel ...) in the file name and title as well since it is more common than the full name.
Thanks

rmzelle:
> I think that would be useful to keep TUBITAK (abbreviation for Turkiye Bilimsel ...) in the file name and title as well since it is more common than the full name.

We try very hard to avoid acronyms in style titles (and file names). See also https://github.com/citation-style-language/styles/wiki/Style-Requirements#1---title-abbreviations.

In order to accept the style, we'll definitely want to keep the file name as "turkiye-bilimsel-ve-teknolojik-arastirma-kurumu.csl".

Regarding the title, we could use the title `<title>TÜBİTAK (Türkiye Bilimsel ve Teknolojik Araştırma Kurumu) (Turkish)</title>`. But the Zotero Style Repository also matches searches against `<title-short>`, so `<title-short>TÜBİTAK</title-short>` should already make the style easy to find for people who search for "TÜBİTAK". To make sure people who search for "tubitak" (without special characters) find the style, can we could change this to e.g. `<title-short>TÜBİTAK (TUBITAK)</title-short>`.

aspirintr:
@rmzelle If the search engine looks for the short title as well, then your suggestion perfectly makes sense. I have changed it accordingly now. 

But I wonder why American Psychological Association have apa.csl file name rather than full name?

PR #3789, opened 2018-11-19
Remove sudo: false from .travis.yml, update Ruby

rmzelle:
Removing sudo: false is suggested by Travis CI in preparation of upcoming container > VM migration. See https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial
 language: ruby
 cache: bundler
 rvm:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: false
 language: ruby
 cache: bundler
 rvm:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ dist: xenial
 language: ruby
 cache: bundler
 rvm:
-  - 2.5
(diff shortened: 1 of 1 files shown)


rmzelle:
@adam3smith, let me know if this looks alright.

Apart from removing sudo: false, which we should do regardless, we can save about 20 seconds per build by using a preinstalled copy of Ruby. I always have trouble with "rvm 2.4" or "rvm 2.5" (it stupidly never seems to pick the preinstalled version), so it looks like we just have to specify a preinstalled version with e.g. "rvm 2.5.3". Specifying the Xenial distribution allows us to use 2.5.3 (https://docs.travis-ci.com/user/reference/xenial/#ruby-support), whereas Trusty is at 2.4.1 (https://docs.travis-ci.com/user/reference/trusty/#ruby-images). Newer Ruby versions are usually a little faster.

rmzelle:
(now with Trusty with Ruby 2.4.1, which is a simpler config)

rmzelle:
Figured it out :P

adam3smith:
thanks for doing this!

rmzelle:
https://www.xkcd.com/149/

PR #3790, opened 2018-11-20
Update molecular-microbiology.csl

hschaefer:
Do not add names to disambiguate citations shortened with "et al." (disambiguate-add-names="false")

The [author guidelines](https://onlinelibrary.wiley.com/page/journal/13652958/homepage/ForAuthors.html) are not explicit about this, but during the review process I was asked to modify citations where a second author was shown for disambiguation.

### DIFF
--- a/molecular-microbiology.csl
+++ b/molecular-microbiology.csl
@@ -121,7 +121,7 @@
       </else>
     </choose>
   </macro>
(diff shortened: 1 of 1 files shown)


hschaefer:
@rmzelle Thank you!

PR #3388, opened 2018-03-20
Create wirtschaftsuniversitat-wien-ivm-interdisziplinares-institut-fu…

georgd:
…r-verhaltenswissenschaftlich-orientiertes-management.csl

PR #3791, opened 2018-11-20
Create wirtschaftsuniversitat-wien-institut-bwl-des-aussenhandels.csl

georgd:


rmzelle:
@adam3smith, since you did a thorough review of https://github.com/citation-style-language/styles/pull/3388, let me know if you also have time to review this, #3792, and #3793. Otherwise I'll just do a metadata check.

@georgd, I see we already have a few styles for Wirtschaftsuniversität Wien: https://www.zotero.org/styles?q=Wirtschaftsuniversit%C3%A4t%20Wien. Do you expect to submit more styles for this university in the future?

rmzelle:
Thanks!

georgd:
Thanks!

@rmzelle it’s quite probable that more styles might come. While we (that’s the team information competence of the university library) recommend to use existing styles when we are asked for advice, we don’t push the institutes to change their guidelines (we aren’t even in a position to do so).

PR #3801, opened 2018-11-26
Create La Trobe University APA 6th Edition

davidvenema:
There are slight changes unique to the La Trobe University implementation of APA 6.


davidvenema:
Was wondering if there was anything else I need to do for this style to be released?

rmzelle:
No, we're good. Just hadn't reviewed this yet. Thanks!

(I fixed the style to British English. Let me know if it should be US English instead. The two English locales mostly vary in date formats and the type of quotation marks)

rmzelle:
(PR replaced #3800)

PR #3802, opened 2018-11-27
Create trinity-theological-college-perth.csl

mnaah:
Based on the SBL first edition, which is no longer in the repository, plus it uses the 'extra' field for series abbreviations in footnotes (if the information is present), but the full series title in the bibliography.

### DIFF
--- a/trinity-theological-college-perth.csl
+++ b/trinity-theological-college-perth.csl
@@ -1,0 +1,806 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" page-range-format="chicago" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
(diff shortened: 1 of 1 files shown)


rmzelle:
> Based on the SBL first edition, which is no longer in the repository

The style currently says `<link href="http://www.zotero.org/styles/chicago-fullnote-bibliography" rel="template"/>`. Did you base your style on this style, or did you dig up the SBL first edition CSL style somewhere?

I was also wondering whether your college has any citation guidelines we could point to (ideally publicly available online).

adam3smith:
(I'd really like to avoid using "note" for random info -- series abbreviation here, e.g. -- in official CSL styles)

mnaah:
Hi Sebastian,

Sorry for the slow reply - I've been away at a conference.

To answer your questions:
- I dug up the CSL style somewhere and tinkered with it;
- our college uses the SBL first edition exactly. I've attached the pdf of
it.

I agree that using the note field for series abbreviation is not ideal; is
there another field that should be used instead?

thanks,
Dan

mnaah:
I'd be very happy with that solution, as long as there is some way to enter
the series abbreviation.

PR #3806, opened 2018-11-29
Hiobludolfcentre styles

PietroLiuzzo:
I have added locale for german because some publications in German want also to use the style and need different results from the default ones. Thank you very much


rmzelle:
Thanks!

PR #3812, opened 2018-12-03
Create representation.csl

POBrien333:
via https://forums.zotero.org/discussion/74192/style-request-for-the-journal-representation


rmzelle:
Thanks!

PR #3813, opened 2018-12-04
Create Dissertation-Medizin-Erlangen.csl

Stoevne:


adam3smith:
Thanks. Do you remember which style you started with to create this one?
Also, have you done more checking on this? I found some obvious discrepancies between this and the Zitierrichtlinie, but don't have time to check the whole thing.

PR #3814, opened 2018-12-05
Create Journal of Small Animal Practice

andywales:
Modification of Harvard style to conform with examples provided for JSAP. All bibliography entries with URL (for reference types 'webpage' and 'report') have [accessed {date}] element included.,


rmzelle:
Thanks!

PR #3817, opened 2018-12-05
Create ieee-Jangjou-2

mehrdadjangjou:
[Jangjou.txt](https://github.com/citation-style-language/styles/files/2650199/Jangjou.txt)


rmzelle:
@mehrdadjangjou, thanks. Can you explain how this style differs from our existing IEEE style?

mehrdadjangjou:
Hi,

Styled Sent Difference with IEEE style  In the case of separators of the authors' names, which are used in Persian styles.

thanks.

adam3smith:
@rmzelle that's specifically adding `delimiter="،" ` in row 86 (they also add it in 85 but it doesn't do anything there). 

I feel like localization should be able to pick that up, but it doesn't.

adam3smith:
@mehrdadjangjou and what specifically is this style for? An institute? Is it just the Persian version of IEEE?

mehrdadjangjou:
hi,
it is the Persian version of IEEE

adam3smith:
@rmzelle what do you think? There's no way to do this otherwise, but I'm also a bit disinclined to take a new variant of any style wherever our localization abilities fall short.

rmzelle:
Copied in the style from https://csl.mendeley.com/styles/529134391/ieee-Jangjou-2.

@mehrdadjangjou, I agree with @adam3smith that it's probably better to not accept this IEEE variant into the style repository. We're aware of the fact that commas need to be better localized, and hopefully CSL styles will be able to automatically adapt in the future. There issue for this is https://github.com/citation-style-language/schema/pull/144.

PR #3820, opened 2018-12-06
Update izmir-katip-celebi-universitesi-sosyal-bilimler-enstitusu.csl

omerfarukgorucu:
No ibid. No plural.


omerfarukgorucu:
Ibid were optional for this style so we prefered to use that. But now it is not wanted to use anymore. Plurality was my own mistake so i have just revised it.

rmzelle:
Thanks!

PR #3821, opened 2018-12-07
Create ibis.csl

carloliveros:


rmzelle:
@carloliveros, thanks. Did you base this on the citation format used in published articles? The instructions to authors (https://www.bou.org.uk/ibis/author-guidelines/submission/) suggest that the journal doesn't require a specific citation format ("References in the list should be in a neutral style").

carloliveros:
Yes. It is not clear what a neutral style means but anyone using a citation
manager will need to choose one format. So I thought that it would be
useful to start with a style that is closest to published articles.

rmzelle:
Do you perhaps know which CSL you started out from? We try to add "template" links to all our styles (https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link).

PR #3822, opened 2018-12-07
Create journal-of-leukocyte-biology.csl

POBrien333:
via 2012 request: https://forums.zotero.org/discussion/25258/style-request-the-journal-of-leukocyte-biology#latest

Style since then is with Wiley. Style doesn't match author guidelines. Style only adapted for article-journal.
I literally just changed a group delimiter, that's all.


rmzelle:
Thanks!

PR #3823, opened 2018-12-07
Adds date accessed to bibliography entry for motion_picture item type

datachampion:
Adds date accessed to bibliography entry for Video Recording / motion_picture item type, in line with London South Bank University referencing guidelines for online videos.


rmzelle:
Thanks!

PR #3825, opened 2018-12-10
Create pesquisa-agropecuaria-brasileira-abnt.csl

luisa-sandes:
This style will serve the Brazilian journal "Pesquisa Agropecuária Brasileira" from Embrapa. Changes have been maid in the ABNT style, especially based on the IPEA csl style. Changes include: spacing between authors in bibliography, between volume and page numbers, removal of "number" in journal articles, specific changes in the style of conference proceedings, reports, books and chapters, presenting all authors in the bibliography, presenting authors in citation in lowercase and with & between authors.

### DIFF
--- a/pesquisa-agropecuaria-brasileira-abnt.csl
+++ b/pesquisa-agropecuaria-brasileira-abnt.csl
@@ -1,0 +1,549 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" et-al-min="3" default-locale="pt-BR" demote-non-dropping-particle="display-and-sort">
+  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>Pesquisa Agropecuária Brasileira (PAB) - ABNT</title>
+    <title-short>PAB-ABNT</title-short>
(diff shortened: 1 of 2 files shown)


luisa-sandes:
Hi there,
I have seen the issues in my submission, the four of them, and I understand what I need to correct in the first 3, but the fourth one I can't really understand and I need your help. The error is:
  _4) pesquisa-agropecuaria-brasileira-abnt: name nodes must have valid et-al-min and et-al-use-first attributes
     Failure/Error:
       expect(min).to be_an_instance_of(first.class),
         "expected et-al-min (#{min}) and et-al-use-first (#{first}) to be of same type"
         expected et-al-min (3) and et-al-use-first () to be of same type_
In the style et-al-min is 3 and et-al-use-first is 1. On the other csl validated files using ABNT style I've checked is almost the same, but the et-al-min is 4. In our style et al should appear for documents with 3 or more authors and, if this is the case, the term et al should appear after the first author. So I believe it is correct to use et-al-min = "3" and et-al-use-first = "1". 
What should I change to correct this error?
Thank you very much for your attention.
Best regards, Luisa

adam3smith:
It looks like you spotted the issue? You can't have `et-al-min` without `et-al-use-first`

luisa-sandes:
> It looks like you spotted the issue? You can't have `et-al-min` without `et-al-use-first`

Hi adam3smith, I am still trying to figure out this issue, I believe in some point of the code I have et-al-min and no et-al-use-first. I'll check that and resend

rmzelle:
Thanks!

PR #3828, opened 2018-12-12
Create universite-du-quebec-a-montreal-litterature-et-semiologie.csl

jjrondeau:
This file is a note style that has rules specific to the Literature Department at Universite du Quebec a Montreal and should not be confused with the style Universite  du  Quebec a Montreal (French - Canada).


rmzelle:
Thanks!

PR #3834, opened 2018-12-18
Update forum-qualitative-social-research-english.csl

POBrien333:
via emails.
Now they actually want two styles, one for English and one for German since there is ONE difference. A comma.
Plus a few smaller changes.


rmzelle:
Thanks!

PR #3835, opened 2018-12-18
Create forum-qualitative-social-research-german.csl

POBrien333:
see https://github.com/citation-style-language/styles/pull/3834
make two separate styles for engl. and german.


POBrien333:
Well. How do I deal with the ISSN Problem. Just remove it for the German style?

rmzelle:
> Well. How do I deal with the ISSN Problem. Just remove it for the German style?

There is an exception list for ISSNs that are allowed to occur in multiple styles: https://github.com/citation-style-language/styles/pull/3835/files#diff-93830fa29d616f7c87903d08b5b1b29a

rmzelle:
Danke!

PR #3836, opened 2018-12-19
APA-G1.csl

DanaAlHabbal:


adam3smith:
Thanks -- could you say what the purpose of this version of APA is and how it differs from the existing APA style?

DanaAlHabbal:
> Thanks -- could you say what the purpose of this version of APA is and how it differs from the existing APA style?

Bold intext/inline citation, italic et al, and et al from 3 authors or more

adam3smith:
OK, and what specifically is it for? Those aren't, obviously, APA requirements.

DanaAlHabbal:
> OK, and what specifically is it for? Those aren't, obviously, APA requirements.

I am aware of that but some professors require these edits on theses even though they are not APA-approved edits

adam3smith:
There seems to be a misunderstanding here: You've submitted the style for inclusion in the official CSL style repository. For this to make sense, we need a way to describe the style and link it to existing requirements -- if this is a requirement for a specific university or degree, say, we could name it after that and people can find it. If that's the case, could you link to a description of these 
I'm concerned that with the current users wouldn't be able to tell what the style is/is for.

adam3smith:
Let's clear up the conceptual issue first; we can help you addressing the automated tests later.

adam3smith:
@DanaAlHabbal closing this because we haven't heard back from you, but we can reopen this at any time, please answer my questions above if you're still interested in getting this style into the repository.

PR #3837, opened 2018-12-19
Updates Elsevier dependents.

cpina:


rmzelle:
@cpina, thanks! Did you have to make any manual changes after running the script with the latest Elsevier journal metadata?

I see that a lot of journal styles only have their datestamp updated. If you run the script with the `--sync` option, it shouldn't touch the styles that are otherwise unchanged. See https://github.com/citation-style-language/journals#updating-timestamps

cpina:
Hi Rintze,

No, I didn't make any manual changes.
I'll generate them again with --sync, thanks! I think that I had used this in the past but I've forgotten, obviously...

cpina:
Actually doing the --sync it tries to delete many styles because are not in the report that I have, but I see them as existing journals. I'll follow this up why some journals are missing from the report.

rmzelle:
@cpina, should this be good to merge now?

PR #3839, opened 2018-12-19
Fixes for unified-style-sheet-for-linguistics.csl

bplimley:
Two simple corrections:

- Edition should be abbreviated "edn." as per the [style sheet](https://www.linguisticsociety.org/sites/default/files/style-sheet_0.pdf) bullet point 11
- Thesis title should be italicized as per the [style sheet](https://www.linguisticsociety.org/sites/default/files/style-sheet_0.pdf) bullet point 1 and style sheet example entries Stewart and Yu


rmzelle:
Thanks!

PR #3842, opened 2018-12-22
Create pure-and-applied-geophysics.csl

POBrien333:
via https://forums.zotero.org/discussion/74448/style-request-pure-and-applied-geophysics
Made a custom style since the Springer style mentioned didn't match.


rmzelle:
Thanks!

PR #3843, opened 2018-12-23
Update pesquisa-agropecuaria-brasileira.csl

luisa-sandes:
A small mistake was made in the original submission. I only modified the <style xmlns> line excluding the et-al-min and et-al-use-first attributes, since the journal does not require et al for the references, only for the citations, and these attributes were generating et al in the references.


rmzelle:
Thanks!

PR #3844, opened 2018-12-23
Enable Stale GitHub app

rmzelle:
See https://probot.github.io/apps/stale/

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
+# Comment to post when closing a stale issue. Set to `false` to disable
+closeComment: false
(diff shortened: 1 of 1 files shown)


rmzelle:
@adam3smith, maybe this would help us keep the number of open PR under control? Let me know if you would like to give it a try.

If so, any thoughts on the settings? How about:

```yaml
# Number of days of inactivity before an issue becomes stale
daysUntilStale: [30?]
# Number of days of inactivity before a stale issue is closed
daysUntilClose: [14?]
# Issues with these labels will never be considered stale
exemptLabels:
  [- pinned?]
# Label to use when marking an issue as stale
staleLabel: [waiting?]
# Comment to post when marking an issue as stale. Set to `false` to disable
markComment: >
  This issue has been automatically marked as stale because it has not had
  [activity in the past 30 days]. It will be closed if no further activity occurs [in the next two weeks]. 
  Thank you for your contributions.
# Comment to post when closing a stale issue. Set to `false` to disable
closeComment: false
```

adam3smith:
Yes! I had actually meant to suggest using this, too (I'm using their reminder for a different project)
Settings look good, except the exempt tag - we'd need something for big tickets that require more time from us.

rmzelle:
> Settings look good, except the exempt tag - we'd need something for big tickets that require more time from us.

I was proposing to just use the default "pinned" label, and we can add the existing repo labels "quality-control" and "repository". Do we need more at this point?

adam3smith:
> and we can add the existing repo labels "quality-control" and "repository". Do we need more at this point?

No, with those three that makes sense. I was worried about just using pinned since that (I think) actually pins the PR which we may not want. You'll set this up?

rmzelle:
I’m pretty sure the “pinned” label has no special status in GitHub. See
e.g.
https://help.github.com/articles/pinning-an-issue-to-your-repository/,
which had no mention of the label.

And yes, will set this up.

adam3smith:
Great, thanks!

PR #3846, opened 2018-12-26

POBrien333:
via https://forums.zotero.org/discussion/75195/royal-society-of-chemistry-presentation-reference-style-incorrect

Add item type "speech" to be properly dealt with in style.


rmzelle:
Thanks!

Issue #3701, opened 2018-09-21
Journal of Neurophysiology update


apstchristensen:
Hi, I copy edit [the _Journal of Neurophysiology_](https://www.physiology.org/journal/jn) for [the American Physiological Society (APS)](http://www.the-aps.org/). I probably worked on your manuscript in at least some capacity if you had one accepted by the journal in the last few years. The style posted on here is not quite correct. I believe [this](https://github.com/citation-style-language/styles/blob/master/journal-of-neurophysiology.csl) is the one being used. Would someone help us reorder the in-text reference citations alphabetically first, chronologically second within that? Some of the punctuation in the existing style may also not be correct, and some of these citations can be abbreviated by not repeating in full when only different years need to be given.

The following are a variety of "perfect" in-text citations to help illustrate my request.

> (Grinvald and Hildesheim 2004; Tsytsarev et al. 2009, 2010)

> (Avillac et al. 2005; Cohen and Andersen 2002; Fetsch et al. 2007; Maier and Groh 2009; Pouget et al. 2002)

> (Landis et al. 2012; Morris 2012; Wadman 2013)

> (Bremmer et al. 2002a; Yang et al. 2011)

> (Danner et al. 2009, 2011, 2013)

> (Caballero et al. 2014a, 2014b)

> Logan and Irwin (2000) proposed separate control signals for eye and hand, whereas Boucher et al. (2007b) proposed...

For the bibliography, here are some fresh examples that show our style, minus DOIs and PMIDs that we pull from CrossRef and PubMed, respectively. I notice sometimes we get manuscripts where the years are placed elsewhere, and we also do not use journal print issue numbers or abbreviate the last page number.

> Eide L, McMurray CT. Culture of adult mouse neurons. _Biotechniques_ 38: 99–104, 2005.

[...]

PR #3847, opened 2018-12-28
Update journal-of-neurophysiology.csl

POBrien333:
via https://github.com/citation-style-language/styles/issues/3701
changed in-text sorting as requested.
Made style more robust.


rmzelle:
Thanks!

PR #3849, opened 2018-12-30
Update monash-university-harvard.csl

POBrien333:
Kinda via here: https://forums.zotero.org/discussion/comment/323336/#Comment_323336
Fix two issues: final dot of bibliography and some extra comma.


rmzelle:
Kinda thanks!

Issue #3439, opened 2018-04-17
Style request: analytical chemistry


DimitriF:
Hello,
I used the american-chemical-society.csl for the journal analytical chemistry and was told that the issue number should not appear for journal articles.
I saw several ACS style having there own csl file in dependent, could it be the case for analytical chemistry?
here are the guideline:
http://pubs.acs.org/paragonplus/submission/ancham/ancham_authguide.pdf
from american-chemical-society.csl  only line 153 must be removed to do so, I can make a pull request.
Thanks
Dimitri

DimitriF:
I should have mention, editor comment is about references in the Supporting Information that do not follow the same editing path.

rmzelle:
@DimitriF, ACS has standardized on a single citation style, and this style requires the issue number for some special cases that CSL styles can't identify and which cannot easily be automated (such as journals that restart page numbering for each issue). For that reason we always include the issue number to err on the side of including more information than requested, instead of too little.

See #3850 for more details, and the text under "Publication Volume Field" in the ACS style guide (https://pubs.acs.org/doi/pdf/10.1021/bk-2006-STYG.ch014, page 296). Ideally the editors of Analytical Chemistry would be more permissive concerning supplementary information. Otherwise you'll just have to edit those references by hand to remove the issues for articles that don't need them.

Issue #1723, opened 2015-10-11
DOI for Advanced Publication in ACS styles


adam3smith:
At a minimum in Analytical Chemistry:
http://pubs.acs.org/paragonplus/submission/ancham/ancham_authguide.pdf
Possibly others.
https://forums.zotero.org/discussion/38084/add-style-request-acs-with-doi/#Item_2

damnation333:
I see there is several ACS styles which include the DOI from Sept 16. This has been dealth with therefore?

adam3smith:
not having looked closely, I don't think so, no: the styles that do _not_ include DOIs should have them added for items published early online. We usually test for the absence of volume and pages, i.e. `<if variable="page volume" match="none">` and then print the DOI for that purpose.

rmzelle:
@adam3smith, I think this is already covered by the current ACS style, which prints the DOI for all items if available. See https://github.com/citation-style-language/styles/blob/b55a77dab0382b7f289ba64b1b643116ee2ca957/american-chemical-society.csl#L266.

Please close this issue if you agree.

PR #3850, opened 2018-12-30

rmzelle:
Closes https://github.com/citation-style-language/styles/issues/3439.

@adam3smith, please merge if this looks okay. The main change is removing the issue for journal articles, which don't seem to be requested by the ACS style guide. I also simplified an `else-if` into an `else`, since the group already acts as a conditional.

For the style guide, Table 14-2 on page 6 of the PDF has examples without issue numbers: https://pubs.acs.org/doi/pdf/10.1021/bk-2006-STYG.ch014

Also, an editor from ACS wrote to me last year:

> American Chemical Society is finally going to a standard ref style in submission of articles. ... The style we will be using is the standard article titles included style, for example:
> 
> Popa, I.; Berkovich, R.; Alegre-Cebollada, J.; Badilla, C. L.; Rivas-Pardo, J. A.; Taniguchi, Y.; Kawakami, M.; Fernandez, J. M. Nanomechanics of HaloTag Tethers. J. Am. Chem. Soc. 2013, 135, 12762–12771."

### DIFF
--- a/american-chemical-society.csl
+++ b/american-chemical-society.csl
@@ -148,7 +148,10 @@
               <text macro="issued" font-weight="bold"/>
               <choose>
                 <if variable="volume">
-                  <text variable="volume" font-style="italic"/>
+                  <group delimiter=" ">
+                    <text variable="volume" font-style="italic"/>
+                    <text variable="issue" prefix="(" suffix=")"/>
+                  </group>
                 </if>
                 <else>
                   <group delimiter=" ">
--- a/american-chemical-society.csl
+++ b/american-chemical-society.csl
@@ -20,7 +20,7 @@
     <category citation-format="numeric"/>
     <category field="chemistry"/>
     <summary>The American Chemical Society style</summary>
-    <updated>2012-09-27T22:06:38+00:00</updated>
+    <updated>2018-12-30T12:00:00+00:00</updated>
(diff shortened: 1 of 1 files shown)


adam3smith:
We should also wrap DOI for advanced online pubs in there as per https://github.com/citation-style-language/styles/issues/1723

rmzelle:
> It's the standard problem with issues:

Ah, I missed that section of the style guide.

> We can decide that this is sufficiently rare for ACS citations to get rid of it, but it would be more consistent to leave the issue in.

In favor of the change:

* The ACS style uses DOIs, which (if provided and correct) already uniquely identify articles
* Since ACS accepts submissions in any style, it's clear that they replace references anyway (except, per the other issue, for supplementary info)
* From personal experience, I think it's rather uncommon for journals in the natural sciences to restart page counts with each issue, but I have no numbers on this

Against:

* The ACS style guide highlights another case whether the addition of the issue is probably more important:

> For publications that have supplements, the following form is recommended.
>
> Taylor, C. W.; Kumar, S. Eur. J. Cancer 2005, 40 (Suppl. 1), 781.
> Eur. J. Anaesthesiol. 2005, 22 (Suppl. S36), 1–35.

I'm not sure how reliably ACS would be able to identify such citations if the supplement info was missing.

Since we haven't gotten any other complaints, let's just leave it in, then.

PR #3851, opened 2018-12-31
Changes made on the IEEE.csl file

gsilano:
Changes made on:

* Added the "note" text field to the IEEE template "ieee.cls"
* Some changes in the date format, the maximum number of authors before it appears in the text "et al." Fixed bug when mentioning a chapter in a book.
* Added author info
* Added "status" value. The "status" allows to specify details about the status of the paper. This is particularly useful when you  use this css style in your own website (e.g., take a look at my website, giuseppesilano.net/publications.html).

### DIFF
--- a/ieee.csl
+++ b/ieee.csl
@@ -44,7 +44,11 @@
   </locale>
   <!-- Macros -->
   <macro name="status">
-    <text variable="status" text-case="capitalize-first" suffix="" font-weight="bold"/>
+    <choose>
+      <if variable="page issue volume" match="none">     
+        <text variable="status" text-case="capitalize-first" suffix="" font-weight="bold"/>
+        </if
+        </choose>
   </macro>
   <macro name="edition">
     <choose>
@@ -246,8 +250,6 @@
             <text macro="locators"/>
             <text macro="page"/>
             <text macro="issued"/>
-          </group>
-          <group delimiter=", " suffix=". ">
             <text macro="status"/>
           </group>
         </if>
@@ -258,8 +260,6 @@
             <text macro="issued"/>
             <text macro="locators"/>
             <text macro="page"/>
-          </group>
-          <group delimiter=", " suffix=". ">
             <text macro="status"/>
           </group>
         </else-if>
@@ -273,9 +273,6 @@
             </group>
             <text macro="issued"/>
           </group>
-          <group delimiter=", " suffix=". ">
-            <text macro="status"/>
-          </group>
         </else-if>
         <else-if type="thesis">
           <group delimiter=", ">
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks. Of these
- we won't take the addition of "note". That doesn't make sense for the general purpose IEEE style
- I don't think your change to the date (i.e. to always include the full date) is right. What are you basing this on?
- Not sure what you're trying to do with chapter? `chapter` isn't a CSL variable. Could you say what's incorrect?
- Where are you getting the et-al change from? I'm pretty sure 7 is in line with the IEEE guide
- The status change looks OK to me.

gsilano:
> Thanks. Of these
> 
> * we won't take the addition of "note". That doesn't make sense for the general purpose IEEE style

It is sometimes useful when you want to add something that can not be inserted using the standard fields in the IEEE model.

> * I don't think your change to the date (i.e. to always include the full date) is right. What are you basing this on?

I have made this change to include references to my papers on my website. Therefore, in my special cases, they are useful. But perhaps these are useful in general.

> * Not sure what you're trying to do with chapter? `chapter` isn't a CSL variable. Could you say what's incorrect?

I needed to include the name of the chapter and the book in my reference. So, I decided to add the chapter macro.

> * Where are you getting the et-al change from? I'm pretty sure 7 is in line with the IEEE guide

Yes of course. This is a typo. On my website, I made this change to show all the authors included in the paper.

> * The status change looks OK to me.

OK. If you agree with me, I can only keep this change by removing others.

ieee.csl
```diff
@@ -39,6 +43,9 @@
     </terms>
   </locale>
   <!-- Macros -->
+  <macro name="status">
+    <text variable="status" text-case="capitalize-first" suffix="" font-weight="bold"/>
```
adam3smith: In the styleguide, I'm seeing 
> [8] E. H. Miller, “A note on reflector arrays,” IEEE Trans. Antennas Propag..., to be published.*

which would suggest not using bold and not capitalizing the first letter (which seems weird given that comma before it anyway). Any reason for those two choices?

gsilano: They are style preferences, my preferences. I would like to emphasize that the paper is in print or about to be published. If you do not think they are useful, you can remove them, of course.

adam3smith:
Thanks. I've made this a good bit more restrictive for the official IEEE style than you had in your custom version. Status will only appear for journal articles and conference papers and only when there is no issue, volume, and page number. Given the IEEE guide, the conservative use of status seems right to me. Obviously other versions can make sense for other purposes (e.g. on a personal website such as in your case).

Issue #3379, opened 2018-03-12
[style request] Biblio 3W


albertovalzgris:
Hi all,

I'm contributing with an article to be published in this journal, currently learning my way through markdown and pandoc exports. I have been browsing the repository and can't find anything matching, yet it would be great to have a suitable .csl instead of hand editing things in the final .doc file in order to match their requirements.

Not sure whether a style request is the correct procedure in this case, so please correct me if I'm wrong. I'd be more than happy to contribute although my coding literacy is close to zero..let's see.

Title of Journal: Biblio 3W. Revista Bibliográfica de Geografía y Ciencias Sociales
ISSN: 1138-9796. 
Depósito Legal: B. 21.742-98
URL address for guideline for author: http://www.ub.edu/geocrit/nrm-b3w.htm
Journal language: English
The Journal URL : http://www.ub.edu/geocrit/bw-22.htm

Cheers!

adam3smith:
See here: https://github.com/citation-style-language/styles/wiki/Requesting-Styles

rmzelle:
@POBrien333, can you add this request to your list?

rmzelle:
@albertovalzgris, we just added a style for Biblio 3W to the repository (courtesy of @POBrien333).

albertovalzgris:
wow, thank you @rmzelle and @POBrien333! 🙏

POBrien333:
@albertovalzgris 
You can always buy me a coffee via https://www.ko-fi.com/damnation333 to show your appreciation ;)

PR #3858, opened 2019-01-04
Create biblio-3w.csl

POBrien333:
closes https://github.com/citation-style-language/styles/issues/3379#issuecomment-450751651
Started with a crap template. Potentially a bunch of mistakes in there I haven't seen.
Style adapted for article-journal, book, chapter, web.


rmzelle:
Thanks! The journal itself uses "Biblio3W" and "Biblio 3W" rather inconsistently, but other websites seem to generally use the latter.

PR #3860, opened 2019-01-06
Generic style rules for linguistics

bplimley:
Stylesheet following the [Generic Style Rules for linguistics](https://www.eva.mpg.de/linguistics/past-research-resources/resources/generic-style-rules.html).

This stylesheet is based on the [Unified Style Sheet for Linguistics](https://www.linguisticsociety.org/resource/unified-style-sheet), which [already has a CSL stylesheet](https://github.com/citation-style-language/styles/blob/master/unified-style-sheet-for-linguistics.csl). However there are several significant differences, described in footnotes 18-21 of the Generic Style Rules PDF:

1. Dissertations/Theses are formatted differently
2. Contributor names are always "Surname, Given Name" rather than alternating to "Given Name Surname" for non-first authors
3. Every contributor name in a list is separated by "&", rather than only the last two
4. Surnames with prefixes such as _van_ or _de_ are alphabetized by prefix instead of by the main part

Also, section 16.3 of the Generic Style Rules document describe placing some parts of the bibliography entry at the end in parentheses, which was unclear or unspecified in the Unified Style Sheet. I also cleaned up some things that seemed awkward.

Thank you!

### DIFF
--- a/generic-style-rules-for-linguistics.csl
+++ b/generic-style-rules-for-linguistics.csl
@@ -15,7 +15,7 @@
     <updated>2018-12-19T23:42:01+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
-  <locale>
+  <locale xml:lang="en">
     <terms>
       <term name="editor" form="verb-short">ed.</term>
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks! (it does seem a bit redundant to have a "unified" and "generic" style sheet for linguistics, though, especially if they're almost identical)

PR #3862, opened 2019-01-07
Update modern-humanities-research-association.csl

POBrien333:
via https://forums.zotero.org/discussion/75305/mhra-artwork-painting-problem
Add details for artwork.

(I might have butchered your citation style, Rintze. Ignore this PR if that's the case.)


adam3smith:
Thanks!

PR #3865, opened 2019-01-08
Create muscle-and-nerve.csl

POBrien333:
via https://forums.zotero.org/discussion/75046/style-request-muscle-nerve


rmzelle:
You've got some nerve muscling in like this! 😜

PR #3872, opened 2019-01-11
Update iforest.csl

luispuerto:


adam3smith:
Thanks! Most changes I made are cosmetic, but I did remove the `display="block"` for number of pages you had in there. I assume that was a typo; there's no indication of it in the styleguide and it makes little sense conceptually.

luispuerto:
Thanks!! I had some problems getting the number of pages to show so I was testing things. When I realized what was the problem I forgot to change that back.

adam3smith:
Thanks!

PR #3880, opened 2019-01-17
Create Scientia-Iranica

smvazirizade:


rmzelle:
@smvazirizade, thanks! There were several cases similar to `prefix=""" suffix="," "` in the style, which isn't valid XML, and which can be better handled with `quotes="true"`, so I made that change.

@adam3smith, since the style didn't validate we should probably briefly check the output against the style guide.

adam3smith:
(style is not quite right; don't merge yet, I'll fix)

rmzelle:
@adam3smith, good to merge now? Thanks already for the fixes.

adam3smith:
Yes, sorry, got interrupted before getting to merge

PR #3883, opened 2019-01-18

lachlan-young:
This update is designed to better match the style specifications for electronic access URLs, video content, titles of serials, legislation, and name particles.

### DIFF
--- a/university-of-south-australia-2017-harvard.csl
+++ b/university-of-south-australia-2017-harvard.csl
@@ -273,7 +273,7 @@
             <text term="accessed"/>
           </if>
           <else>
-            <text term="viewed"/>
+            <text term="retrieved"/>
           </else>
         </choose>
         <date variable="accessed" prefix=" ">
--- a/university-of-south-australia-2017-harvard.csl
+++ b/university-of-south-australia-2017-harvard.csl
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" default-locale="en-GB" version="1.0" demote-non-dropping-particle="sort-only">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" default-locale="en-GB" version="1.0" demote-non-dropping-particle="never">
   <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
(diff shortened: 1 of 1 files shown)


university-of-south-australia-2017-harvard.csl
```diff
@@ -283,8 +281,14 @@
           <date-part name="month" form="long" suffix=" "/>
           <date-part name="year" form="long"/>
         </date>
-        <text variable="URL" prefix=", &lt;" suffix="&gt;"/>
       </if>
+	</choose>
+  </macro>
+  <macro name="access-link">
+	<choose>
+	  <if type="article-journal book chapter" match="none">
```
adam3smith: Is this right, though? On https://www.library.unisa.edu.au/referencing-roadmap/electronic/e-book/ I'm seeing

> Author's family name, Initial(s) OR Authoring body year, Title of ebook, edition if applicable, Publisher, Place of publication if available, <URL>.

and a book with a URL is by definition an e-book.

university-of-south-australia-2017-harvard.csl
```diff
@@ -267,7 +265,7 @@
     <label variable="page" form="short" suffix=" "/>
     <text variable="page"/>
   </macro>
-  <macro name="access">
+  <macro name="access-date">
```
adam3smith: I didn't look at this in detail, but it looks like not all item types that get a URL should get an access date (see e.g. the e-book example) so maybe this needs to be restricted (to blogposts and webpages perhaps?)

university-of-south-australia-2017-harvard.csl
lachlan-young: As you can see on the style page you referenced, our style has multiple rules for displaying e-book URLs. Freely available e-books show the URL, but purchased or library e-books do not (and similarly for journals). Because we can't capture this nuance within the bounds of CSL, we've decided that as the majority of e-book and journal content available in our institution is purchased/subscribed content, the default should be to not display a URL for these content types.

university-of-south-australia-2017-harvard.csl
lachlan-young: Almost all types with a URL should have an access date. The exceptions will fall into the types discussed above where we don't want to show the URL anyway. I've repositioned the logic for these types to make it clear that they affect both the access date and the URL.

lachlan-young:
The proposed change has been updated to address the issues identified by @adam3smith.

rmzelle:
Thanks both!

PR #3884, opened 2019-01-19
Create Korean Journal of Dermatology.csl

mdwallnut:
Korean Journal of Dermatology csl form


rmzelle:
@mdwallnut, thanks. Could you point me to the instructions to authors for this journal? We try to include a "documentation" link in every style (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#10---documentation-link)

PR #3885, opened 2019-01-21
Changed et-al-min to 3 at first use

snauhaus:
JMS leaves this pretty ambiguous. If et al starts right away and at how many authors is not defined in their style guide. However, it was pointed out to me that all other articles in JMS use et al right away, at the 3rd author, so that's what I changed the style to.


adam3smith:
Thanks for documenting the rationale, too!

PR #3888, opened 2019-01-25
Create Chicago-author-date-Unit-test-nb-NO.csl

edunit:


adam3smith:
Thanks! Could you describe a bit what specifically this is for? I'm a bit confused by the mention of unit testing

edunit:
Unit is the Norwegian Directorate for ICT and Joint Services in Higher Education and Research.
All libraries needing this style in Norwegian are in the BIBSYS consortium so I don't imagine institutions not affiliated to Unit will want to use it.
Also I wrote test before i'm not sure all the changes I did are enough or work, but I'm obligated to upload it here in order to find the style in our library system.

rmzelle:
@edunit, could you explain in a bit more detail how this style will be used? Is it for rendering citations in library online catalogs?

rmzelle:
> Hi, It will be will to export reading lists from our reading list system Leganto. Chicago in English is available per now but we need a Norwegian version as well, which is what I tried to create here.

@adam3smith, thoughts? See https://www.diffchecker.com/L4mXqVCt for a diff against Chicago author-date. There are a few capitalization differences that we probably can't account for via style localization. (and Leganto is an Ex Libris product, FWIW)

PR #3889, opened 2019-01-25
Create chicago-author-date-unit-test-nb-no.csl

edunit:


rmzelle:
(supersedes #3888)

adam3smith:
So our concern with this is the 2nd bullet under reasons not to accept styles here: https://github.com/citation-style-language/styles/wiki/Criteria-for-Accepting-Styles
i.e.

> styles submitted solely for use with an API. (the Zotero Web API recently gained the ability to generate references using CSL styles hosted outside our repository) 

The way I understand you, this style is only going to be used to automatically generate citations from library catalogs. The expectation isn't that people will download it for use in Mendeley, Zotero, etc. We're discinclined to host styles for that purpose in the style repository, both because it's confusing for end-users and because that leaves us with their maintenance -- it's much easier to just have them in stored with the catalog directly.

Could be that I'm still misunderstanding the purpose here, though, so correct me if that's the case please.

edunit:
Hi,

Pretty much every research library in Norway uses the program Endnote, which I'm pretty sure is the reason why no Chicago style in Norwegian has been uploaded here before (the same kind of file in .enl format is quite easy to find).

Even though that's not the reason I uploaded the file, it doesn't mean Zotero users won't want to use it. In other words, it won't necessarily only be used to automatically generate citations from library catalogs.

The reason I uploaded it it to indeed generate citations from Leganto, which is not the library catalog itself but merely a reading list system. It seems to be a requirement that the style is stored here, in order to do so, and according to the customer service at Ex Libris', dutch librarians have done so before.
In this case, just storing the style in the catalog is not a simpler solution since it's is done something I can do. My hands are tied here so I hope you will reconsider.
Thank you.

edunit:
Hi again,
Any chance you can have a look at this and give us an answer?
Unit represent the BIBSYS-consortium, that gathers most university and research libraries in Norway.
We need to be able to provide citation styles aligned with the Norwegian citation rules and language.
Thank you.

Best regards,
Emmanuelle Dabin - System Expert
Unit – The Norwegian Directorate for ICT and Joint Services in Higher Education and Research
BIBSYS dpt.

adam3smith:
Having looked at the diff more closely, I just don't see why this couldn't be addressed simply by adding some specific Norwegian terms to the locale section of the main Chicago style(s). 
The only other changes you made are 
- removing a couple of `capitalize-first` from variables, which I don't think is right: those are variables rendered after periods and I'm pretty sure Norwegian capitalizes after periods.
- adding the editor and translator macro in the bibliography. That, too, I think is a mistake. Those macros are already called elsewhere and duplicating them is going to produce duplication of elements in the citation.

We have the ability to work with locales specifically to avoid as much as possible adding additional styles just to switch the language, which simply doesn't scale. 
I know Zotero, e.g., has a fair number of Norwegian users and I think the reason they haven't asked for this is that the standard styles with Norwegian locale work well. (Endnote doesn't have the same ability, hence the need for dedicated language versions of styles).

If it isn't possible already, ExLibris should be able to add the ability to specify a locale in generating citations in Leganto with relative ease, certainly for a large customer such as Bibsys. 

I'm sorry this is inconvenient for you, but we're having a hard enough time maintaining the number of styles there are as it is already, and we simply can't add more styles simply because an implementation is unable or unwilling to use features provided by CSL.

edunit:
Hi again,
I understand, and thank you for looking at it. 

How can we make sure that the capitalization is correct?
We've looked at the Norwegian locale file but we're still not sure how we "add it" to the main Chicago file.
Could be that the reason we don't know that is that it's just not possible (yet) in Alma.
Feel free to redirect me to the correct channel to ask this kind of questions.

adam3smith:
> How can we make sure that the capitalization is correct?

What are specific issues you have seen?

> We've looked at the Norwegian locale file but we're still not sure how we "add it" to the main Chicago file.

That depends: Is nothing in Norwegian currently? Then Alma is not picking the locale up at all and you'd need to talk to devs there.

If you get some Norwegian terms, but they're wrong, you can customize them in the style:
Look at the added `en` terms in the current Chicago style: https://github.com/citation-style-language/styles/blob/master/chicago-author-date.csl#L36 

You can create a separate locale section for Norwegian right next to it where you define any terms that differ from the regular Norwegian locale and we'd be happy to take that as a PR.

edunit:
> 
> 
> > How can we make sure that the capitalization is correct?
> 
> What are specific issues you have seen?
> 

So far, we've only been using the main Chicago style, with English capitalization rules. Capital letters for each word of the title is one of the issues we've seen. Is this solved by the use of locales (Norwegian language, Norwegian rules)? I couldn't see anything in the file that would confirm it.

> > We've looked at the Norwegian locale file but we're still not sure how we "add it" to the main Chicago file.
> 
> That depends: Is nothing in Norwegian currently? Then Alma is not picking the locale up at all and you'd need to talk to devs there.
> 

I was afraid of that, I'm already on it.

adam3smith:
> Is this solved by the use of locales (Norwegian language, Norwegian rules)? I couldn't see anything in the file that would confirm it.

Correct, with the locale set to Norwegian, anything that's not explicitly marked as English (by specifying "en" or "English" as the language) will not see any changes to capitalization.

adam3smith:
I'll close this here, but feel free to post if you need additional advice. Happy to try to help.

PR #3892, opened 2019-01-26
Make Resuscitation a dependent (again)

rmzelle:
Per https://github.com/citation-style-language/journals/pull/67

PR #3893, opened 2019-01-27
Revert "Make Resuscitation a dependent (again)"

pezhmanxxx:
Reverts citation-style-language/styles#3892


rmzelle:
@pezhmanxxx, can you explain why you're submitting this change? We determined at https://forums.zotero.org/discussion/comment/306994/#Comment_306994 that this style should use the Elsevier-Vancouver format after all. Do you think that's not right?

(cc @adam3smith, @POBrien333)

PR #3895, opened 2019-01-29
Update hochschule-pforzheim-fakultat-fur-wirtschaft-und-recht.csl

POBrien333:
Amendments as requested here: https://forums.zotero.org/discussion/75669/hochschule-pforzheim


rmzelle:
Thanks!

PR #3896, opened 2019-01-29
Create journal-of-transport-history.csl

POBrien333:
https://forums.zotero.org/discussion/73824/style-request-the-journal-of-transport-history


POBrien333:
This should be fine now.
Publisher moved from numeric to note style. Current publications (Jan-19) still don't follow that guide, but the guidelines from 2018 were announced as new.

adam3smith:
@POBrien333 check out the changes on the ibid/subsequent: because of how they evaluate, you always want `ibid-with-locator`, `ibid`, and `subsequent` in the order they are now. Everything that's `ibid-with-locator` is also `ibid` and everything that's `ibid` is also `subsequent`, so if you start with `subsequent`, you never get to the `ibid` conditions.

POBrien333:
Ah great! Thank you. Wasn't aware of that! Thanks for the fixes!

adam3smith:
Thanks!

PR #3897, opened 2019-01-29
Minor tweaks to bibliography

snauhaus:
Journal volume should be bold (line 171), there shouldn't be a DOI (73-89), and the et al min should be 3. The latter I already added in a previous pull request. I cannot figure out how to pull from the master to my fork on the website... Sorry for that. I hope it's ok that way. 

I got the adjustments from the JMS Style Guide (http://www.blackwellpublishing.com/pdf/JMS_Prep_of_Manuscripts.pdf). There are still some minor elements that could be tweaked in the CSL file, but I found it easier to just do it by hand in the end.

### DIFF
--- a/journal-of-management-studies.csl
+++ b/journal-of-management-studies.csl
@@ -71,7 +71,7 @@
   <macro name="access">
     <choose>
       <!-- <if variable="DOI">
-        <text variable="DOI" prefix=", doi: "/>
+      <!--   <text variable="DOI" prefix=", doi: "/>
       </if> -->
       <!-- <else> -->
         <group>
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks!

> I cannot figure out how to pull from the master to my fork on the website... Sorry for that. I hope it's ok that way.

It's generally easiest to follow the instructions at https://github.com/citation-style-language/styles/blob/master/CONTRIBUTING.md#3b-submitting-changes-to-an-existing-style, which will create a fresh branch in your fork, and which avoids the need to keep the "master" branch in your fork up to date.

PR #3898, opened 2019-01-30

nchachereau:
This is an update to a custom style for history students in Switzerland. See #234 for the original pull request, and #3400 for the last update.

Changes:
- When given, DOIs are now always preferred to URLs, even for webpages, weblog posts and forum posts; a new macro ensures that DOIs and URLs are displayed in a consistent manner (in the last update, a bug had crept in)
- The edition field is displayed for newspaper articles (important for historians citing articles from an era with two or even three daily editions for some newspapers)
- Added support for datasets

I look forward to your feedback and take this opportunity to thank you for the time invested in this repository! :pray:


rmzelle:
Looks good, thanks!

PR #3901, opened 2019-01-31
Create giappichelli.csl

POBrien333:
Quick and dirty style, as it's in Italian which is exhausting.
Via https://forums.zotero.org/discussion/75151/style-request-giappichelli-italian


rmzelle:
Thanks!

PR #3905, opened 2019-02-01
Update hochschule-pforzheim-fakultat-fur-wirtschaft-und-recht.csl

POBrien333:
small error: https://forums.zotero.org/discussion/comment/325417/#Comment_325417


rmzelle:
Thanks!

PR #3906, opened 2019-02-02
Update annales.csl

NicoBarr:
Made a few changes that correspond to the updated style that Annales is now using:
- corrected for articles: year comes after volume and issue
- updated: volume and issue for articles now noted "XX-X"
- updated: cited pages for articles (full citation) is: "ici p."
- updated: for subsequent citations, after the short title, "op. cit." is used for books but "art. cit." is used for articles.


rmzelle:
Thanks!

PR #3907, opened 2019-02-03
Create springer-humanities-author-date-no-parentheses.csl

Alexei-Kassian:
Identical to 'Springer - Humanities (author-date)', but without compulsory parentheses for in-line citations, e.g., Smith 2017, not (Smith 2017)


adam3smith:
We can help with addressing the automated checks, but could you clarify first what this is for? Is this actually required by a journal, or do you just prefer the no parentheses version for some reason?

adam3smith:
Great, thanks. Given that, I don't think we'll take this for the repository, though you're obviously more than welcome to share with other people (e.g. the [link to the raw file ](https://raw.githubusercontent.com/citation-style-language/styles/08f959307c5e7f630df935c506ef01683229b89d/dependent/springer-humanities-author-date-no-parentheses.csl) should be pretty stable if you want to share that).

PR #3908, opened 2019-02-04
Update invisu.csl

amendesd:
Modification of the "page macro" and the "accessed macro", and some other minor bugs.
Thanks a lot.


rmzelle:
Thanks!

PR #3913, opened 2019-02-06
Update diabetologia.csl

POBrien333:
Make independent style.
via https://forums.zotero.org/discussion/75482/diabetologia-style-needs-updating
Only changes is the et-al settings and that it wants the issue number of a article-journal.

Free publication from 2019: https://link.springer.com/content/pdf/10.1007%2Fs00125-018-4775-z.pdf


rmzelle:
Let's wait with this until we're certain this just affects Diabetologia, which might not be the case (per https://forums.zotero.org/discussion/comment/325921/#Comment_325921).

rmzelle:
Thanks. Looks like they just confirmed on the Zotero forums that this needs to be a custom style.

rmzelle:
(still need to put this on the skip list for the Springer journals metadata)

PR #3914, opened 2019-02-06
Align harvard-university-of-bath.csl with current guidance

alex-ball:
These changes bring the style as near as possible into line with the current University of Bath guidance:
https://library.bath.ac.uk/referencing/harvard-bath.

The CSL-YAML database against which this style has been tested is available for inspection:
https://github.com/alex-ball/bathbib/blob/master/csl/bath-csl-test.yaml

Tips for using this style are available here:
https://github.com/alex-ball/bathbib

(NB. I am a librarian at the University of Bath and also maintain the LaTeX implementations of the style.)

### DIFF
--- a/harvard-university-of-bath.csl
+++ b/harvard-university-of-bath.csl
@@ -90,7 +90,7 @@
   <macro name="archive">
     <choose>
       <if type="graphic">
-        <text variable="archive-place" prefix="At: " suffix=". " />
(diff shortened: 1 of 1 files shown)


adam3smith:
(for simplicity, we enforce a single license and the exact same text for it across styles)

rmzelle:
Thanks!

PR #3918, opened 2019-02-07
Update and rename dependent/journal-of-medicinal-chemistry.csl to jou…

POBrien333:
…rnal-of-medicinal-chemistry.csl

Make independent due to journal not wanting any et-al.
https://forums.zotero.org/discussion/75662/truncated-author-list-et-al-with-j-med-chem-style#latest


POBrien333:
OK. Apparently we DO have a no-et-al version of the ACS style (https://raw.githubusercontent.com/POBrien333/styles/master/american-chemical-society-with-titles-no-et-al.csl).
However, that doesn't appear on the repository?
https://www.zotero.org/styles?q=american-chemical-society-with-titles-no-et-al

rmzelle:
ACS recently standardized on a single citation style (https://axial.acs.org/2018/02/01/review-ready-submission/), so we got rid of all their variants. Apparently they shouldn't even be complaining about the reference style used at all since they wrote

>You may provide references in any style, but they must be complete including titles.

and

> To streamline reference styling, we will be updating the publicly available reference managers with a standard style, which will automatically include titles. Journal-specific reference styles will be applied by the journal production team after the journal accepts a paper for publication.

rmzelle:
> Apparently we DO have a no-et-al version of the ACS style (https://raw.githubusercontent.com/POBrien333/styles/master/american-chemical-society-with-titles-no-et-al.csl).
> However, that doesn't appear on the repository?

Just in case you're still wondering about this: your "master" branch is way behind ours (https://github.com/POBrien333/styles shows "This branch is 23 commits ahead, 1193 commits behind citation-style-language:master."), so this styles still exists in your fork even though we deleted it a while back.

We could revive the american-chemical-society-with-titles-no-et-al.csl variant (under the name of this journal), but it would be better if the editor would relax their requirements. I'll post in the Zotero thread.

POBrien333:
Ahhh. How do I go about getting that back in line with the master fork?

rmzelle:
There is typically no need to keep your "master" branch up to date, since it's much easier to just create fresh branches off the official repo.

Since you made some changes in it, you can either reset "master" (and lose all your changes in this branch) so you can more easily pull in any changes from upstream, or you can merge the newer commits into your branch.

PR #3920, opened 2019-02-08
Update molecular-biology-of-the-cell.csl

POBrien333:
Old request. Updated style to match guidelines. Checked a few publications and their style is all over the place.
via https://forums.zotero.org/discussion/70691/update-request-molecular-biology-of-the-cell

Free paper (Oct-18): https://www.molbiolcell.org/doi/pdf/10.1091/mbc.E18-02-0123


rmzelle:
Thanks!

PR #3930, opened 2019-02-11

mserifovic:
Removing extra space before "Available at" and adding "report" to types that should show the publisher.


mserifovic:
I also removed the hanging indent.

adam3smith:
Thanks!

PR #3933, opened 2019-02-15
fix: remove space after opening parenthesis

sarfata:
The style currently prints `( consulté le` for webpage and blog posts. I think it's much better without the extra space after the parenthesis.

Thanks a lot for an awesome and very useful `.csl`!


rmzelle:
Thanks! (I simplified it a little bit more, since the "suffix" was redundant)

PR #3938, opened 2019-02-16
Fixes details for edited volumes in TAPA style

diyclassics:
- removes prefix ', ' from secondary-contributors label, i.e. 'ed.'
 - adds suffix ' ' to same
 - moves locators-chapter to correct position; changes prefix to '.'
 - updates contributor


rmzelle:
Thanks!

PR #3939, opened 2019-02-17
Update marine-ecology-progress-series.csl

sgorgopa:
Format was not correct (the style that MEPS is dependant to is not very similar to the desired end result, also really out of date), made a new MEPS style using Addiction-Biology as a template.
Changed capitalization of titles, changed the format for the webpage access date (lowercase, added parentheses). Updated the ISSN and the link for author instructions.

### DIFF
--- a/dependent/marine-ecology-progress-series.csl
+++ b/dependent/marine-ecology-progress-series.csl
@@ -1,17 +1,164 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" default-locale="en-US">
-  <!-- Generated with https://github.com/citation-style-language/utilities/tree/master/generate_dependent_styles/data/int-res -->
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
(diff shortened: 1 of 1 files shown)


sgorgopa:
The checks appear to have failed because the MEPS style already exists? (issn duplicate)... also because the code probably is written as a standalone style, not dependent like the old version?  Please advise how to move forward. Thank you

sgorgopa:
Should i submit this as a new independent style and have the old (dependent) one removed?

rmzelle:
Don't worry about the checks for now. It's easy for us to fix.

Do you remember which guidelines you based your style on? It looks like all Inter-Research Science Center journals still use the same citation format (see https://www.int-res.com/journals/guidelines-for-authors/manuscript-formatting-requirements-and-preparation-tips/), so if that's the case we just would update that parent style (https://github.com/citation-style-language/styles/blob/master/inter-research-science-center.csl).

sgorgopa:
Okay. 

The guidelines you linked to are correct. So that sounds right, for the parent style to be updated so all Inter-Research Science Center journals would use the updated formatting. Thanks!

rmzelle:
> So that sounds right, for the parent style to be updated so all Inter-Research Science Center journals would use the updated formatting.

Done in https://github.com/citation-style-language/styles/pull/3941, using your style. Thanks!

PR #3932, opened 2019-02-13
Create journal-of-developmental-and-behavioral-pediatrics.csl

POBrien333:
via https://forums.zotero.org/discussion/75890/style-request-journal-of-developmental-and-behavioral-pediatrics#latest

PR #3940, opened 2019-02-17
Update journal-of-developmental-and-behavioral-pediatrics.csl

rmzelle:
Fix documentation link and use ampersand in title. Follow-up of #3932.


rmzelle:
Thanks! :P

PR #3943, opened 2019-02-19
Create JEMBE style

jbjharvey:
JEMBE = Journal of Experimental Marine Biology and Ecology, please add this style


adam3smith:
Pull requests like this are how you would contribute a style you have coded. To request a style see https://github.com/citation-style-language/styles/wiki/Requesting-Styles

PR #3953, opened 2019-02-25
Update brain.csl

### LINKED ISSUE #3949: Update brain.csl style
Brain (brain.csl) has updated their template since the last release. The letters following the year should now be italicized. Additionally only the letter should be highlighted. I have added a photo below of the current csl document (left) and what it should be (right)

img

Below I have uploaded the template which was emailed to me by the editor recently.
[Brain_template.docx](https://github.com/citation-style-language/styles/files/2894345/Brain_template.docx)

POBrien333:
closes #3949

### DIFF
--- a/brain.csl
+++ b/brain.csl
@@ -44,8 +44,8 @@
         <group>
           <date variable="issued">
             <date-part name="year"/>
-            <text variable="year-suffix" font-style="italic"/>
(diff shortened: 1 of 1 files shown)


rmzelle:
@POBrien333, thanks as always!

PR #3954, opened 2019-02-25
Create Mediaeval Studies style

adunning:
Style for the journal [_Mediaeval Studies_](http://www.pims.ca/publications/journal-mediaeval-studies).

### DIFF
--- a/mediaeval-studies.csl
+++ b/mediaeval-studies.csl
@@ -261,23 +261,6 @@
       </if>
     </choose>
   </macro>
-  <macro name="edition">
-    <choose>
-      <if type="book chapter graphic motion_picture paper-conference report song" match="any">
-        <choose>
-          <if is-numeric="edition">
-            <group delimiter=" ">
-              <number variable="edition" form="ordinal"/>
-              <text term="edition" form="short"/>
-            </group>
-          </if>
-          <else>
-            <text variable="edition" text-case="capitalize-first" suffix="."/>
-          </else>
-        </choose>
-      </if>
-    </choose>
-  </macro>
   <macro name="locators-note-join-with-space">
     <choose>
       <if type="article-journal" variable="volume" match="all">
@@ -407,22 +390,6 @@
   <macro name="event-note">
     <text variable="event"/>
   </macro>
-  <macro name="event">
-    <choose>
-      <if variable="title">
-        <group delimiter=" ">
-          <text term="presented at"/>
-          <text variable="event"/>
-        </group>
-      </if>
-      <else>
-        <group delimiter=" ">
-          <text term="presented at" text-case="capitalize-first"/>
-          <text variable="event"/>
-        </group>
-      </else>
-    </choose>
-  </macro>
   <macro name="originally-published">
     <group delimiter=", ">
       <group delimiter=": ">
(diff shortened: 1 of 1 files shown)


mediaeval-studies.csl
```diff
@@ -0,0 +1,1347 @@
...
+    <id>http://www.zotero.org/styles/mediaeval-studies</id>
+    <link href="http://www.zotero.org/styles/mediaeval-studies" rel="self"/>
+    <link href="http://www.pims.ca/publications/journal-mediaeval-studies" rel="documentation"/>
+    <link href="https://www.zotero.org/styles/chicago-fullnote-bibliography" rel="template"/>
+    <author>
+      <name>Andrew Dunning</name>
+      <uri>https://orcid.org/0000-0003-0464-5036</uri>
+    </author>
+    <category citation-format="note"/>
+    <category field="generic-base"/>
```
adam3smith: switch out for a less generic field

mediaeval-studies.csl
```diff
@@ -0,0 +1,1347 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="display-and-sort" page-range-format="chicago">
+  <info>
+    <title>Mediaeval Studies</title>
+    <id>http://www.zotero.org/styles/mediaeval-studies</id>
+    <link href="http://www.zotero.org/styles/mediaeval-studies" rel="self"/>
+    <link href="http://www.pims.ca/publications/journal-mediaeval-studies" rel="documentation"/>
+    <link href="https://www.zotero.org/styles/chicago-fullnote-bibliography" rel="template"/>
```
adam3smith: this should be http:// (no s) for legacy reasons

mediaeval-studies.csl
```diff
@@ -0,0 +1,1347 @@
...
+    </author>
+    <category citation-format="note"/>
+    <category field="generic-base"/>
+    <issn>0076-5872</issn>
+    <eissn>2507-0436</eissn>
+    <summary>Mediaeval Studies journal</summary>
+    <updated>2018-02-25T19:59:17+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="en-CA">
```
adam3smith: I'd probably to "en" here to catch all English locales

adunning:
According to the website, it's identical to Chicago, but without publisher names or URLs. Should we not bother with a separate style?

adam3smith:
no publisher names is a pretty significant change, we'll need a separate style.

adam3smith:
@adunning it'd be great if you could make the couple of adjustments as per the above.

adunning:
Have applied requested changes, if you can reassess.

adunning:
It's quite confusing that the online CSL Validator doesn't pick up the same errors as Travis.

adam3smith:
we run a set of additional tests beyond validation on styles for the repository -- naming, macros, etc. -- that aren't caught by just validation against the CSL schema.

adunning:
I've gone through the process five times now, and it simply finds more errors each time that it didn't mention before. How does one deal with this normally?

adam3smith:
I'm guessing that the macros you deleted contained the only call to other macros, which is why new issues came up. There's not a particularly easy way to handle this beyond checking this.

adunning:
Obviously it likes you better.

adam3smith:
Thanks!

PR #3955, opened 2019-02-26

jjrondeau:
I made some changes, mainly concerning the original publication date for books and the entry of the composers and performer's names in the Author field, thus avoiding the use of the Note field.

### DIFF
--- a/universite-du-quebec-a-montreal.csl
+++ b/universite-du-quebec-a-montreal.csl
@@ -326,7 +326,7 @@
           <group delimiter=" ">
             <text term="by"/>
             <names variable="reviewed-author">
-              <name and="text" delimiter-precedes-last="never" et-al-min="8" et-al-use-first="6" et-al-subsequent-min="8" et-al-subsequent-use-first="6" et-al-use-last="true" initialize-with="."/>
+              <name and="text" delimiter-precedes-last="never" et-al-use-last="true" initialize-with="."/>
             </names>
           </group>
         </group>
@@ -626,7 +626,7 @@
       <else-if variable="issued">
         <group prefix="(" suffix=")">
           <choose>
-            <if type="legal_case" match="all">
+            <if type="legal_case" match="any">
               <choose>
                 <if match="any" variable="references">
                   <text variable="authority" suffix=", "/>
@@ -703,7 +703,7 @@
     <choose>
       <if variable="issued">
         <choose>
-          <if type="book" match="all" variable="translator">
+          <if type="book" match="any">
             <date form="text" variable="original-date" suffix="/"/>
           </if>
         </choose>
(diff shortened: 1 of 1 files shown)


universite-du-quebec-a-montreal.csl
```diff
@@ -679,6 +702,11 @@
   <macro name="issued-year">
     <choose>
       <if variable="issued">
+        <choose>
+          <if type="book" match="all" variable="translator">
```
adam3smith: why are you testing for translator here? That seems odd for original date -- that's frequently for historic works (and you're not testing for translator below)

universite-du-quebec-a-montreal.csl
```diff
@@ -293,31 +305,30 @@
...
         <group delimiter=", " prefix="[" suffix="]">
           <text variable="title"/>
           <text variable="reviewed-title" font-style="italic" prefix="Compte rendu du livre "/>
-          <names variable="reviewed-author" prefix="par ">
-            <name and="text" delimiter-precedes-last="never" et-al-min="8" et-al-use-first="6" et-al-subsequent-min="8" et-al-subsequent-use-first="6" et-al-use-last="true" initialize-with="."/>
-          </names>
+          <group delimiter=" ">
+            <text term="by"/>
+            <names variable="reviewed-author">
+              <name and="text" delimiter-precedes-last="never" et-al-min="8" et-al-use-first="6" et-al-subsequent-min="8" et-al-subsequent-use-first="6" et-al-use-last="true" initialize-with="."/>
```
adam3smith: no need for the et al's here in `name` -- they'll get inherited from the bibliography element.

universite-du-quebec-a-montreal.csl
```diff
@@ -993,9 +1021,15 @@
...
+      <else-if type="book" match="all" variable="original-date">
+        <group delimiter=" " prefix=" (" suffix=").">
+          <text value="Oeuvre originale publiée en "/>
+          <date form="text" variable="original-date" prefix="&amp;#169;"/>
+        </group>
+      </else-if>
     </choose>
   </macro>
-  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-names="true" disambiguate-add-givenname="true" disambiguate-add-year-suffix="true" givenname-disambiguation-rule="primary-name" collapse="year">
+  <citation et-al-min="3" et-al-use-first="1" initialize-with="." disambiguate-add-names="true" disambiguate-add-givenname="true" disambiguate-add-year-suffix="true" givenname-disambiguation-rule="primary-name" collapse="year">
```
adam3smith: I'd remove the `initialize-with` here -- afaict you have this on all names (and where you don't, you can add it) and people don't expect this on `citation`, so this will lead to all sorts of confusion in editing the style.

universite-du-quebec-a-montreal.csl
jjrondeau: Our style guide distinguishes between translated and untranslated books and it should not in my humble opinion. The guide will be reviewed about this. For now, I'm going to remove the "translator" condition.

universite-du-quebec-a-montreal.csl
jjrondeau: OK, I will remove it.

universite-du-quebec-a-montreal.csl
jjrondeau: I was thinking about initials that appear in citations when two authors must be disambiguated. I will remove it.

adam3smith: You can add that to the `name` element as well (if it isn't already). They won't show up with `form="short"` unless for disambiguation.

adam3smith:
thanks!

PR #3958, opened 2019-02-27
Create Bulletin-des-GTV.csl

HyxeDay:
It's the same style as epidemiology-and-infection.csl ( https://github.com/POBrien333/styles/blob/69cc6e71e2810fc6039155a918cd8c7917c96a1e/epidemiology-and-infection.csl ) except for "citation-number", the brackets "[ ]" have been replaced by parentheses "( )".


adam3smith:
The guide for authors is here: https://www.sngtv.org/4DACTION/NS2013_INDEX/5#
Unfortunately the submitted style isn't actually very close at all.

PR #3962, opened 2019-03-01

POBrien333:
Just some minor changes as per https://forums.zotero.org/discussion/75447/style-request-american-journal-of-botany-changed-publisher-new-style
You'll notice I've changed the locale to en-GB as the dates are in British style.


rmzelle:
Thanks!

PR #3963, opened 2019-03-02
BiAMA-BAOnline

NMonteix:


NMonteix:
@adam3smith, is there anything to change on those two new styles?

bollettino-d-archeologia-online.csl
```diff
@@ -0,0 +1,237 @@
...
+                  <text macro="publisher"/>
+                </group>
+              </group>
+            </else-if>
+            <else-if type="entry-dictionary entry-encyclopedia" match="any">
+              <group delimiter=", ">
+                <text macro="title" quotes="true"/>
+                <choose>
+                  <if variable="title-short" match="any">
+                    <text variable="title" form="short" font-style="italic" prefix="in "/>
```
adam3smith: What's the expected output here? This seems odd

adam3smith:
Sorry for the wait. I have one question for the second style inline, everything else looks good.

bollettino-d-archeologia-online.csl
NMonteix: It would be a version that would function until container-title-short is recognized for abridged titles.
For instance, the _Realencyclopädie der classischen Altertumswissenschaft_ has to be shortened in _RE_ but it cannot be done yet.
I guess I should had an explanation about this in the info section?

NMonteix: Unless the extra field can be used to add a container-title-short field? But [there](https://www.zotero.org/support/kb/item_types_and_fields#citing_fields_from_extra), it doesn't seem possible (unless I can insert something like this:
`CSL Variable: Value`
`container-title-short: RE`)

adam3smith: Yes, exactly: `container-title-short: RE` will work and is the right way to do this.

NMonteix: Done. Travis seems to have difficulties to finish checking it though

adam3smith:
Yeah, seems stuck. I'll merge this, might just affect this PR.

NMonteix:
Thanks!

adam3smith:
yup, build went through fine once merged to master.

adam3smith:
(ignore this)

PR #3966, opened 2019-03-03
Create infomin

rcervantesb:


infomin.csl
```diff
@@ -0,0 +1,432 @@
...
+      <term name="anonymous">Anon.</term>
+      <term name="no date">n.d.</term>
+      <term name="retrieved">Available</term>
+      <term name="accessed">Consulted</term>
+    </terms>
+  </locale>
+  <macro name="author">
+    <names variable="author">
+      <name and="symbol" name-as-sort-order="all" sort-separator=", " initialize-with=". " delimiter="; " delimiter-precedes-last="never">
+        <name-part name="family"/>
```
adam3smith: no need for the name parts here.

infomin.csl
```diff
@@ -0,0 +1,432 @@
...
+        <name-part name="given"/>
+      </name>
+      <substitute>
+        <names variable="editor">
+          <name and="symbol" name-as-sort-order="all" sort-separator=", " initialize-with=". " delimiter="; " delimiter-precedes-last="never"/>
+          <label form="short" prefix=" (" suffix=")."/>
```
adam3smith: add the label just like this to the author (where it's empty), then use just `<names variable="editor"/>` (and translator, interviewer, container-author) in substitute.

infomin.csl
```diff
@@ -0,0 +1,432 @@
...
+  <macro name="event">
+    <text variable="event" font-style="italic"/>
+  </macro>
+  <macro name="title">
+    <choose>
+      <if type="book post webpage interview report patent thesis" match="any">
+        <text variable="title" prefix=". " font-style="italic"/>
+      </if>
+      <else-if type="article-journal article-magazine article-newspaper chapter entry-encyclopedia entry entry-dictionary paper-conference post-weblog" match="any">
+        <text variable="title" suffix="”" prefix=". “"/>
```
adam3smith: You'll want to use `quotes="true"` for that and write the quotes into the terms section

infomin.csl
```diff
@@ -0,0 +1,432 @@
...
+        <text variable="page" suffix=" p."/>
+      </else-if>
+      <else>
+        <label variable="page" form="short" suffix=" "/>
+        <text variable="page"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="number-of-pages">
+    <text variable="number-of-pages"/>
```
adam3smith: you'll want all of the term+variable combination in a group so that the term doesn't get printed if the variable isn't present.

infomin.csl
```diff
@@ -0,0 +1,432 @@
...
+      </else-if>
+      <else>
+        <text variable="title" prefix=". " font-style="italic"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="container-title">
+    <text variable="container-title" font-style="italic"/>
+  </macro>
+  <macro name="responsability">
```
adam3smith: should be "responsiblility" (obviously the code doesn't care about the typo, but makes it harder to fix).

infomin.csl
```diff
@@ -0,0 +1,432 @@
...
+    <choose>
```
adam3smith: No need for this choose/if here. The group takes care of that.

adam3smith:
Leaving this open

adam3smith:
Gracias

PR #3967, opened 2019-03-03
Create leonardo.csl

POBrien333:
via https://forums.zotero.org/discussion/75367/style-request-leonardo#latest

### DIFF
--- a/dependent/leonardo-music-journal.csl
+++ b/dependent/leonardo-music-journal.csl
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" default-locale="en-GB">
+  <info>
+    <title>Leonardo Music Journal</title>
+    <title-short>LMJ</title-short>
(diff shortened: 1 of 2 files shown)


adam3smith:
I'm not actually sure what to do with these requirements: This is, conceptually, an endnote style:

> Each reference number should appear ONLY ONCE in the text. Subsequent references to the same source should have a new number and refer back to the initial reference.
>
> Example:
>    12. See Jones [1] p. 32.

(and also asks for page numbers in the references) but they tell users explicitly not to use Word endnotes (because.... reasons). I think people are still going to be better of using endnotes, so would turn this into a note style.

adam3smith:
@POBrien333 have a look and see what you think. I don't think we can do this style 100% either way, but this seems much closer to the spirit of it, especially with the cross-referencing.

adam3smith:
@POBrien333 I'd be curious about your opinion on this -- not sure this is the right way to go.

POBrien333:
Looks good and makes sense. I hadn't seen how the bibliography was structured.
Good to go :)

adam3smith:
Thanks!

PR #3968, opened 2019-03-04
prefix doi witt https://doi.org/

mattluce:
This brings in line wither other styles (e.g. chicago)

### DIFF
--- a/american-medical-association.csl
+++ b/american-medical-association.csl
@@ -52,7 +52,7 @@
       <if type="article-newspaper" match="none">
         <choose>
           <if variable="DOI">
-            <text value="doi:"/>
+            <text value="https://doi.org/"/>
             <text variable="DOI"/>
           </if>
           <else-if variable="URL">


adam3smith:
I'm all for this conceptually, but do we know that's OK with AMA?

mattluce:
After doing some more research, I believe you are correct, AMA does not support this. Sorry for jumping the gun.

rmzelle:
I looked around a little bit, and it indeed looks like AMA still sticks with "doi:". AMA hasn't released a new print edition since their 10th edition from 2007, and while http://www.amamanualofstyle.com/page/About/about#about says their online version receives updates, I couldn't find any mention that their have adopted "https://doi.org/". I found a list with recent changes at http://www.amamanualofstyle.com/oxford/page/updates.

Their most recent tweet with a doi from January also still uses "doi:" (https://twitter.com/AMAManual/status/1088071180320423937), and the only blog post about DOIs I could find is from 2012 and also recommends "doi:" (http://amastyleinsider.com/2012/01/17/digital-object-identifiers-dois-for-electronic-references/). Crossref brought it up with them as well: https://twitter.com/AMAManual/status/842018162702991363

The documentation link of https://github.com/citation-style-language/styles/blob/master/american-medical-association.csl (http://westlibrary.txwes.edu/sites/default/files/pdf/ama_citation_style.pdf) is broken, by the way. Looks to have moved to https://westlibrary.txwes.edu/sites/default/files/pdf/AMACitationStyle.pdf.

rmzelle:
(updated the AMA documentation links in https://github.com/citation-style-language/styles/commit/5a44444b2294f8385c18752e690d7f90cda87c55)

PR #3969, opened 2019-03-05
Update Palaeontology and Journal of Paleontology

ms609:
These updates reflect edits that I have been asked to make by the journals' copyeditors, and bring the styles up to date.


journal-of-paleontology.csl
```diff
@@ -179,7 +179,7 @@
                 <choose>
                   <if is-numeric="volume">
                     <group delimiter=" ">
-                      <text value="Vol. "/>
```
adam3smith: I'd much rather do this properly with label so here `<label variable="volume" form="short"/>` and then define
```
<locale>
  <terms>
    <term name="volume" form="short">v.</term>
  </terms>
</locale>
```

after line 20. above

journal-of-paleontology.csl
```diff
@@ -152,10 +152,10 @@
                 <text variable="page"/>
               </else>
             </choose>
-            <text term="page" form="short"/>
+            <text term="page" form="short" prefix="p. "/>
```
adam3smith: this isn't right. The `term="page"` already gives you `p.` or `pp.` so you definitely don't want a `p. ` prefix on it. What's the intended end-result here?

journal-of-paleontology.csl
```diff
@@ -205,11 +205,11 @@
...
           </else-if>
           <else>
-            <text variable="title" suffix=". "/>
+            <text variable="title" suffix=": "/>
             <text variable="container-title" form="long" text-case="title" suffix=","/>
-            <group delimiter=":">
-              <text variable="volume" prefix=" "/>
-              <text variable="page"/>
+            <group delimiter=", ">
+              <text variable="volume" prefix="v. "/>
```
adam3smith: same here -- better to put the v. and the p. into labels. For the p.: this is never going to be pp.?

palaeontology.csl
```diff
@@ -142,14 +144,14 @@
...
+              <text term="in" text-case="capitalize-first" font-style="italic" suffix=" "/>
               <text macro="editor" suffix=" "/>
               <group>
                 <text variable="container-title" text-case="title" font-style="italic" suffix=", "/>
                 <choose>
                   <if is-numeric="volume">
-                    <group delimiter=" ">
-                      <text value="Vol. "/>
+                    <group delimiter=" " suffix=".">
+                      <text value="Vol."/>
```
adam3smith: let's use term here as you do in line 118 above

journal-of-paleontology.csl
ms609: This was an error. Thanks for spotting it.

journal-of-paleontology.csl
ms609: Apparently p. is to be used whether referring to one page or to a range.

journal-of-paleontology.csl
adam3smith: It's `form="short"`; you have `format="short"` which doesn't exist, so CSL reverts to the default, which is long.

ms609: Whoops! Thanks.  Looking good now.

ms609:
I think this should now be good to go -- did I miss anything, @adam3smith ?

adam3smith:
Great, thanks!

PR #3977, opened 2019-03-08
Create ceska-zemedelska-univerzita-v-praze-fakulta-agrobiologie-potra…

matejovo:

There are new rules for citaion. 

At our faculty we can create citation based on journal Conservation Biology.


rmzelle:
@matejovo, sorry for not reviewing this sooner.

Do I understand correctly that you would like to change the existing style for "Česká zemědělská univerzita v Praze - Fakulta agrobiologie, potravinových a přírodních zdrojů" (https://www.zotero.org/styles?q=id%3Aceska-zemedelska-univerzita-v-praze-fakulta-agrobiologie-potravinovych-a-prirodnich-zdroju) to use the exact same citation format of our Conservation Biology CSL style (https://www.zotero.org/styles?q=id%3Aconservation-biology)?

If so, we're almost there. We just need to delete the existing independent style, and we can probably remove the "2019" from this new dependent style. I can make those changes if you confirm that this is the change you want to make.

(I also noticed that the style you're submitting here is fixed to English, whereas the original style could be switched to other languages (like Czech) as well)

PR #3978, opened 2019-03-08
Delete ceska-zemedelska-univerzita-v-praze-fakulta-agrobiologie-potra…

matejovo:

this style is dependent on http://www.zotero.org/styles/conservation-biology

There are new rules for citaion.
At our faculty we can create citation based on journal Conservation Biology.


matejovo:
Any way how to move this style to a dependent style repository? And change it to style based on journal Conservation Biology?

PR #3979, opened 2019-03-08
Create contemporary-accounting-research.csl

POBrien333:
via https://forums.zotero.org/discussion/75460/style-request-contemporary-accounting-research#latest


adam3smith:
Thanks!

PR #3981, opened 2019-03-10
Correct et-al-min

tipichris:
Style guide is full list of names 'for up to three authors', so et-al-min needs to be 4


rmzelle:
Thanks!

PR #3982, opened 2019-03-12

lachlan-young:
This change corrects the display of author labels (they should include periods if these are in the labels) and corrects the et-al limit for the number of authors that are included in the bibliography (the limit should be 7, after which the elllipsis > last author format is used, see http://www.library.unisa.edu.au/siteassets/pdfs-powerpoint-files/referencing-roadmap/referencingroadmapfaqs.pdf pg. 8).


adam3smith:
Thanks! Note that we use 2spaces/level, not tabs, for XML indenting. You can use https://formatter.citationstyles.org to reformat styles before submitting (I've fixed this here)

PR #3986, opened 2019-03-13

POBrien333:
via https://forums.zotero.org/discussion/76383/style-error-acta-universitatis-agriculturae-sueciae


adam3smith:
Given pending update only took a quick look, but looks good in general. Thanks!

PR #3987, opened 2019-03-14
zeitgeschichte-wien-german

MertzG:


adam3smith:
Thanks! This looks like it is based on an existing style. Do you remember which?

MertzG:
Sorry I don't know anymore. Is there a way to find out?

zeitgeschichte-wien.csl
```diff
@@ -0,0 +1,393 @@
...
+            </if>
+            <else>
+              <text macro="contributors-long" suffix=", "/>
+              <text macro="title-long"/>
+              <text prefix=", " macro="container"/>
+              <text macro="issued"/>
+              <text prefix=", " macro="pages"/>
+              <choose>
+                <if variable="page" type="chapter" match="all">
+                  <text prefix=", hier " macro="citation-locator"/>
```
adam3smith: I'm not seeing the ", hier" for indidividual pages in the journal. Are you sure that's correct?

zeitgeschichte-wien.csl
```diff
@@ -0,0 +1,393 @@
...
+            <date-part name="year" prefix=" "/>
+          </date>
+        </group>
+      </if>
+    </choose>
+  </macro>
+  <macro name="citation-locator">
+    <text variable="locator"/>
+  </macro>
+  <citation et-al-subsequent-min="6" et-al-subsequent-use-first="5" disambiguate-add-names="true">
```
adam3smith: I don't see any mention of the et-al settings in the guide. You're sure about those?

zeitgeschichte-wien.csl
```diff
@@ -0,0 +1,393 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="note" version="1.0" names-delimiter="/" demote-non-dropping-particle="sort-only" default-locale="de-DE" xmlns="http://purl.org/net/xbiblio/csl">
+  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>Zeitgeschichte Wien (German)</title>
```
adam3smith: I'm seeing the journal listed as "Zeitgeschichte" everywhere. Why add the "Wien" part?

MertzG:
You are right with all three points: "Wien" can be left out. I checked back with the lecturer of Zeitgeschichte and "hier" can be left in but must not. I suggest to leave it out. For the et-al settings there should never be a "u.a." or "et. al", so I guess the limit should be high or no limit at all. Shall I try to correct or can you do it quickly?

adam3smith:
Great, thanks!

PR #3989, opened 2019-03-15
Update microbial-drug-resistance.csl

POBrien333:
via https://forums.zotero.org/discussion/76410/microbial-drug-resistance-journal-template-problem


adam3smith:
Thanks!

PR #3990, opened 2019-03-15
Added rules for online journals and newspapers

abeeken:


university-of-lincoln-harvard.csl
```diff
@@ -24,7 +24,7 @@
   </locale>
   <macro name="editor">
     <names variable="editor" delimiter=", ">
-      <name and="text"/>
+      <name/>
```
adam3smith: Why this change? I think the and should stay and you should add `initialize-with="."`, no? Or what are you trying to do?

university-of-lincoln-harvard.csl
```diff
@@ -223,31 +220,27 @@
           </group>
         </else-if>
         <else-if type="book graphic report" match="any">
-          <group delimiter=". " prefix=" ">
+          <group prefix=" " delimiter=". ">
             <text variable="note"/>
```
adam3smith: I know this isn't yours, but we want the note removed here. It makes no sense.

university-of-lincoln-harvard.csl
```diff
@@ -223,31 +220,27 @@
...
-            <text macro="title" prefix=" "/>
-            <text macro="edition" prefix=", "/>
-          </group>
+          <text macro="title" prefix=" " suffix="."/>
           <group prefix=" " delimiter=" ">
             <text term="in" text-case="capitalize-first" suffix=":"/>
             <text macro="editor"/>
             <text variable="container-title" font-style="italic" suffix="."/>
             <text variable="collection-title" suffix="."/>
+            <text macro="edition" prefix=" "/>
```
adam3smith: No need for the prefix given the group delimiter, though you may need a suffix to separate edition from date

university-of-lincoln-harvard.csl
abeeken: Hi Adam,

I've kind of inherited this work so I'm picking up a lot of things here as I'm going. Pretty much all of these aren't changes I'm sure I've made so I'm not entirely sure why they're needed or what implications they have! This being a prime example, I'm not sure what initialize-with would do for us.

university-of-lincoln-harvard.csl
abeeken: Will remove this.

university-of-lincoln-harvard.csl
abeeken: I see - I'll get rid of that and take a look at if we need the suffix.

university-of-lincoln-harvard.csl
adam3smith: `<name/>` gives you 
Adam Smith, Karl Marx, Andrew Beeken.

`<name and="text" initialize-with="."/>`
gives you A. Smith, K. Marx, and A. Beeken which I think is correct for editors in the style, no?

abeeken:
@adam3smith I've made the recommended changes - hopefully that's all okay!

PR #3991, opened 2019-03-15
Update pest-management-science.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/328338/#Comment_328338


adam3smith:
Thanks!

PR #4000, opened 2019-03-20

jjrondeau:
I modified some date parameters to correct displaying and add a few missing elements.


rmzelle:
Merci!

PR #4012, opened 2019-03-25
Update wiesbaden-business-school.csl with ibid

POBrien333:
add ibid via https://forums.zotero.org/discussion/comment/328859#Comment_328859


adam3smith:
Thanks!

PR #4013, opened 2019-03-25
Update the-journal-of-hand-surgery-asian-pacific-volume.csl

alfchong:
Added spaces in prefix and suffixes stripped by html-tidy


adam3smith:
@alfchong for future reference, you can use https://formatter.citationstyles.org/ instead of tidy, which formats the style according to our repository specs.

adam3smith:
Thanks!

PR #4016, opened 2019-03-25
Update journal-of-peptide-science.csl

POBrien333:
update of style as per: https://forums.zotero.org/discussion/comment/324708/#Comment_324708
See current pdfs of journal.


rmzelle:
Thanks!

PR #4019, opened 2019-03-27
Create journal-of-natural-products.csl

lfnothias:
Dear CSL community,

This style matches the requirements for Journal of Natural Products style.
I am not an expert in style, so please let me know if anything has to be adjusted.

Thank you,

LF

### DIFF
--- a/journal-of-natural-products.csl
+++ b/journal-of-natural-products.csl
@@ -7,6 +7,9 @@
     <id>http://www.zotero.org/styles/journal-of-natural-products</id>
     <link href="http://www.zotero.org/styles/journal-of-natural-products" rel="self"/>
     <link href="http://pubs.acs.org/paragonplus/submission/jnprdf/jnprdf_authguide.pdf" rel="documentation"/>
+    <issn>0163-3864</issn>
+    <eissn>1520-6025</eissn>
+    <updated>2019-03-04T12:00:00+00:00</updated>
     <contributor>
       <name>Julian Onions</name>
       <email>julian.onions@gmail.com</email>
--- a/journal-of-natural-products.csl
+++ b/journal-of-natural-products.csl
@@ -1,0 +1,268 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="expanded" default-locale="en-US">
+  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>Journal of Natural Products</title>
+    <title-short>J. Nat. Prod.</title-short>
+    <id>http://www.zotero.org/styles/journal-of-natural-products</id>
+    <link href="http://www.zotero.org/styles/journal-of-natural-products" rel="self"/>
+    <link href="http://pubs.acs.org/paragonplus/submission/jnprdf/jnprdf_authguide.pdf" rel="documentation"/>
+    <contributor>
+      <name>Julian Onions</name>
+      <email>julian.onions@gmail.com</email>
+    </contributor>
+    <contributor>
+      <name>Ivan Bushmarinov</name>
+      <email>ib@ineos.ac.ru</email>
+    </contributor>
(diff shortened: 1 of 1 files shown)


lfnothias:
I was doing some csl-validation with http://simonster.github.io/csl-validator.js/ and I got the following:

```
stdin:162:20: error: missing attributes of http://purl.org/net/xbiblio/csl^text
required:
	attribute ^macro
	attribute ^term
	attribute ^value
	attribute ^variable

stdin:269:0: error: junk after document element
error: invalid input
```

I am not sure how to interpret that though

adam3smith:
Before we get into why validation and the automated tests fail -- we do have a style for that journal already. Is that incorrect? If so, how?

lfnothias:
The format of the style available is basically the ACS general style.
However there are small differences (like no volume specified, ...)

adam3smith:
(and for future reference, we do have our own validator that produces slightly more helpful error messages: 
https://validator.citationstyles.org )

adam3smith:
Oh - didn't see this was an ACS journal. ACS has actually asked us to only have one single style for all their journals. They have a policy to accept submissions in any coherent citation style and adjust the style on publication.

lfnothias:
Right. What I know from experience with the Journal of Natural Products is that the style they required has some differences with the generic ACS format. It's fine for submission but they will ask the author(s) to make the changes themselves during the revisions, which can be painful if there is no dedicated style.

adam3smith:
Experience when though? ACS is very explicit about this:
> To streamline reference styling, we will be updating the publicly available reference managers with a standard style, which will automatically include titles. **Journal-specific reference styles will be applied by the journal production team** after the journal accepts a paper for publication.

https://axial.acs.org/2018/02/01/review-ready-submission/

lfnothias:
Thanks for your reply and informations.
It sounds likes they have a good intentions, but in practice it is not the case ^^, at least not for Journal of Natural Products. 

Some commercial reference manager like Paperpile have a dedicated style for JNP,  which is perfect. Myself, I started using paperpile for that very first reason. Now I am still using it because of google docs integration.

Unless, you recommend otherwise, I will simply provide my journal-of-natural-products.csl publicly on my own GitHub page.

adam3smith:
I'm pretty sure Paperpile uses our citation styles. 

But anyway, yes, by all means put the style up on your github page, but I would strongly recommend that you make sure it validates before you do so to ensure that it works with different reference managers.

lfnothias:
Thanks for your reply. I was investigating this further. You are right, Paperpile uses the CSL public style (Sorry I forgot that I made a custom style couple of years ago.)

Now looking at the Paperpile style for Journal of Natural Products, it proposes the ACS one (The same is observed for Zotero):

Journal of Natural Products (american-chemical-society.csl)
> (1) 	Nothias, L.-F.; Nothias-Esposito, M.; da Silva, R.; Wang, M.; Protsyuk, I.; Zhang, Z.; Sarvepalli, A.; Leyssen, P.; Touboul, D.; Costa, J.; et al. Bioactivity-Based Molecular Networking for the Discovery of Drug Leads in Natural Product Bioassay-Guided Fractionation. _J. Nat. Prod._ **2018**, _81_ (4), 758–767.

Expected style by the editors for Journal of Natural Products
> Nothias, L.-F.; Nothias-Esposito, M.; da Silva, R.; Wang, M.; Protsyuk, I.; Zhang, Z.; Sarvepalli, A.; Leyssen, P.; Touboul, D.; Costa, J.; Paolini J.; Litaudon M.; Dorrestein C. Bioactivity-Based Molecular Networking for the Discovery of Drug Leads in Natural Product Bioassay-Guided Fractionation. _J. Nat. Prod_. **2018**, _81_ (4), 758–767.

In comparison to the ACS style, JNP style requires for journal article:
1) No title.
2) No issue number.
3) No et al abreviation.

Also for Book/Chapter, only the book title is needed (not the chapter title).

Here is tree article from the last issue to illustrate the style (see the pdf):
https://pubs.acs.org/doi/10.1021/acs.jnatprod.8b00379
https://pubs.acs.org/doi/10.1021/acs.jnatprod.8b00567
https://pubs.acs.org/doi/10.1021/acs.jnatprod.8b00713

PR #4020, opened 2019-03-29
Update leonardo-music-journal.csl

rmzelle:
Correct ISSN(s), per https://www.mitpressjournals.org/loi/lmj


rmzelle:
(also changing locale to "en-US" since https://www.leonardo.info/preparing-your-materials-journals#refs says "Articles must employ proper use of academic written American English.", and this dependent was set to British English while the parent was already US English)

PR #4026, opened 2019-03-31
Update reproduction-in-domestic-animals.csl

okingfisher:
Corrections.


adam3smith:
The guide for authors just says to use APA style -- any reason we shouldn't just link this directly to the existing APA style?

okingfisher:
Sure, just found interesting to have a specific style for this journal.

adam3smith:
We can still have a specific style and link it to APA http://docs.citationstyles.org/en/stable/specification.html#dependent-styles -- we'll do this, no need for you to figure out the details. Thanks for pointing out the error in the existing style!

okingfisher:
Didn't know about that, good idea!

adam3smith:
Closed via https://github.com/citation-style-language/styles/commit/4898c4047a40a6b9359233a44d5bbb1b6c756a6b

PR #4033, opened 2019-04-03
Update universite-de-bordeaux-ecole-doctorale-de-droit.csl

fflamerie:
Tidied up macros `title` and `title-short` : text value "[microfiche]" is no longer displayed for books. The string "[microfiche]" should be inserted only for theses and does not apply to books.


adam3smith:
Thanks!

PR #4034, opened 2019-04-04
Create foodborne-pathogens-and-disease.csl

POBrien333:
via https://forums.zotero.org/discussion/75892/style-request-foodborne-pathogens-and-disease


adam3smith:
Thanks!

rmzelle:
Thanks! (add "medicine" as a category, since "pathogens" and "disease" are medical topics :) )

PR #4036, opened 2019-04-04
Update and rename perspectives-on-politics.csl to dependent/perspecti…

POBrien333:
…ves-on-politics.csl

via https://forums.zotero.org/discussion/75939/style-update-perspectives-on-politics


adam3smith:
Thanks!

PR #4037, opened 2019-04-04
Update international-journal-of-climatology.csl

POBrien333:
small fixes via https://forums.zotero.org/discussion/76455/style-error-international-journal-of-climatology#latest


adam3smith:
Thanks!

PR #4038, opened 2019-04-05
Create societes-contemporaines.csl

POBrien333:
via https://forums.zotero.org/discussion/76165/request-societes-contemporaines-style


rmzelle:
Thanks!

PR #4039, opened 2019-04-06
Change de Buck particle treatment to display-and-sort

fbennett:
See https://forums.zotero.org/discussion/76710/de-buck-dropping-particles-in-author-last-name-in-van-de-in-lastname


adam3smith:
I'll merge this as it seems to be the consensus on that thread. I've mostly given up on fully understanding van/Van in Dutch...

rmzelle:
Vantastic.

PR #4040, opened 2019-04-06
Update college-of-naturopathic-medicine.csl

taffod:
changes to et al format options


POBrien333:
the first et-al number needs to be higher than the 2nd number, that's why it's not validating.

adam3smith:
@taffod -- please always make changes to the latest version of the style on the repository. You're re-adding things that we had previously removed like the location of the college. Could you say specifically how you're trying to change the et al. behavior, ideally with examples?

adam3smith:
(we're happy to re-open this if you get back to us at any point in the future)

PR #4041, opened 2019-04-07
fix bugs in anglia.csl

icornelius:
I have corrected formatting for films and for secondary contributors to edited collections.


rmzelle:
Thanks!

PR #4044, opened 2019-04-07
Create philosophy-and-public-affairs.csl

POBrien333:
via https://forums.zotero.org/discussion/76330/style-request-philosophy-public-affairs#latest


POBrien333:
duplication of https://github.com/citation-style-language/styles/pull/4043

PR #4046, opened 2019-04-09
Update american-journal-of-obstetrics-and-gynecology.csl

POBrien333:
via https://forums.zotero.org/discussion/59526/style-error-american-journal-of-obstetrics-gynecology
I have not checked this, just submitted the PR.


adam3smith:
Yes, this is right, I did check the guide.

adam3smith:
Thanks!

PR #4055, opened 2019-04-20
Update iso690-full-note-sk.csl

kohafan:
Added a missing line (456).


adam3smith:
Thanks. FWIW, the `prefix=". " suffix="."` combinations will typically work, but it's not ideal to rely on the fact that the processor suppresses duplicate periods. Better to use groups with delimiters.

kohafan:
Thank you for the suggestion, I will try to improve the code.

PR #4057, opened 2019-04-23
Create humanistica-lovaniensia-journal-of-neo-latin-studies.csl

POBrien333:
via https://forums.zotero.org/discussion/76329/style-request-humanistica-lovaniensia#latest

### DIFF
--- a/humanistica-lovaniensia-journal-of-neo-latin-studies.csl
+++ b/humanistica-lovaniensia-journal-of-neo-latin-studies.csl
@@ -29,7 +29,7 @@
   <macro name="author">
     <names variable="author" suffix=", ">
       <name and="text" initialize-with="." delimiter=", "/>
-      <label form="short" prefix=" (" suffix=")" strip-periods="false"/>
+      <label form="short" prefix=" (" suffix=")"/>
       <substitute>
         <names variable="translator"/>
         <names variable="editor"/>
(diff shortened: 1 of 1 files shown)


humanistica-lovaniensia-journal-of-neo-latin-studies.csl
```diff
@@ -0,0 +1,298 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" page-range-format="minimal" demote-non-dropping-particle="sort-only">
+  <!-- Polyglot; journal publishes in English, French, German, Italian and Spanish. -->
+  <info>
+    <title>Humanistica Lovaniensia. Journal of Neo-Latin Studies</title>
```
adam3smith: @rmzelle Opinion on the subtitle? The journal consistently uses it & uses it with a period. In the US (e.g. in JSTOR) it's listed with a colon. I think any of the three possible options (period, colon, no subtitle) is fine, not sure we have super clear rules on this.

rmzelle: > I think any of the three possible options (period, colon, no subtitle) is fine, not sure we have super clear rules on this.

We never standardized on this, so I suggest to just follow whatever the journal uses.

rmzelle: (I do prefer replacing all dashes with hyphens)

rmzelle:
Thanks!

PR #4059, opened 2019-04-24
Update bibliotheque-universitaire-de-medecine-vancouver.csl

alerac:
* Added "note" variable to "webpage" document type to take
  into account the date of the webpage (not website) if any
  is present in the "Extra" field of Zotero.

### DIFF
--- a/bibliotheque-universitaire-de-medecine-vancouver.csl
+++ b/bibliotheque-universitaire-de-medecine-vancouver.csl
@@ -505,6 +505,7 @@
                   <text term="internet" text-case="capitalize-first" prefix="[" suffix="]. "/>
                   <date form="text" variable="issued" suffix=". "/>
                   <text variable="title" suffix="; "/>
+				  <text variable="note" suffix=" "/>
                   <group prefix="[" suffix="]. " delimiter=" ">
                     <text term="cited"/>
                     <date form="text" variable="accessed"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
No, I removed that for a reason - relying on note is generally a bad idea. People store all types of things in the Extra field in Zotero and once you move to other reference managers, you'll get all sorts of weird stuff. Could you clarify what date exactly this is supposed to capture? I couldn't quite follow the distinction.

alerac:
> No, I removed that for a reason - relying on note is generally a bad idea. People store all types of things in the Extra field in Zotero and once you move to other reference managers, you'll get all sorts of weird stuff. Could you clarify what date exactly this is supposed to capture? I couldn't quite follow the distinction.

Hi, Ok I see the problem. 

Actually in the Vancouver guide _(following the Vancouver recommendations)_ that we provide at the library ([link here](https://www.bium.ch/wp-content/uploads/2019/03/Vancouver_guide_BiUM_2019_0327.pdf)), P. 23 we say that if the date of the website _(globally)_ and the date of the webpage _(content of the website)_ are present then the students can mention it in the reference.

webpage_webiste_date

However, Zotero does not make a distinction between webpage and website _(there is only a "Webpage" document type)_ and only one "Date" field is available.

So I thought it might be worthwhile to manage the second date with the "Extra" field. But I really understand your point and it is better no to stray from the standard.

I will correct again and not take into account the webpage date.

Thank you for your explanation!

alerac:
Hi,
I removed the "note" variable and corrected the suffix for the "title" variable. It is ok on my side, no more corrections to do at this point.
Thanks for the help.
Best,
Alex

rmzelle:
Thanks!

PR #4060, opened 2019-04-25
Create universidad-de-leon-harvard.csl

buvmc:
* Changes in the format name and author
* Changes the citation of a work with three or more authors
* Changes in the "et al." expression
* After the DOI expression the blank space is eliminated

### DIFF
--- a/universidad-de-leon-harvard.csl
+++ b/universidad-de-leon-harvard.csl
@@ -6,7 +6,7 @@
     <id>http://www.zotero.org/styles/universidad-de-leon-harvard</id>
     <link href="http://www.zotero.org/styles/universidad-de-leon-harvard" rel="self"/>
     <link href="http://www.zotero.org/styles/harvard-university-of-greenwich" rel="template"/>
-    <link href="https://biblioteca.ucm.es/cps/recursos-para-tfg-tfm-citas-y-referencias-bibliograficas" rel="documentation"/>
+    <link href="http://buleria.unileon.es/handle/10612/5930" rel="documentation"/>
     <author>
       <name>Víctor Martínez</name>
     </author>
--- a/universidad-de-leon-harvard.csl
+++ b/universidad-de-leon-harvard.csl
@@ -1,0 +1,305 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+  <info>
+    <title>Universidad de León (España) - Harvard</title>
+    <title-short>ULE (España) - Harvard</title-short>
+    <id>https://csl.mendeley.com/styles/374856691/universidad-de-leon-harvard</id>
+    <link href="http://www.zotero.org/styles/universidad-de-leon-harvard" rel="self"/>
+    <link href="http://www.zotero.org/styles/harvard-university-of-greenwich" rel="template"/>
+    <link href="https://biblioteca.ucm.es/cps/recursos-para-tfg-tfm-citas-y-referencias-bibliograficas" rel="documentation"/>
+    <author>
+      <name>Víctor Martínez</name>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="generic-base"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks! The "changes" you describe in your commit message, are those relative to the template style or what are they referring to?

buvmc:
Hi. First of them is a change relative to the template style, and the other three are changes made in the style in order to adapt it to the specifications needed at the University of León (Spain)

universidad-de-leon-harvard.csl
```diff
@@ -0,0 +1,305 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+  <info>
+    <title>Universidad de León (España) - Harvard</title>
+    <title-short>ULE (España) - Harvard</title-short>
+    <id>http://www.zotero.org/styles/universidad-de-leon-harvard</id>
+    <link href="http://www.zotero.org/styles/universidad-de-leon-harvard" rel="self"/>
+    <link href="http://www.zotero.org/styles/harvard-university-of-greenwich" rel="template"/>
+    <link href="https://biblioteca.ucm.es/cps/recursos-para-tfg-tfm-citas-y-referencias-bibliograficas" rel="documentation"/>
```
rmzelle: @buvmc, did you enter this documentation link? It points to a website from the Universidad Complutense Madrid, not the Universidad de León. Does the Universidad de León have its own online style guide?

buvmc:
Hi rmzelle. Yes, I'd entered the documentation link from the Universidad Complutense from Madrid because I understand that is needed a link to a documentation manual and we still don't have any like that in the Universidad de León

universidad-de-leon-harvard.csl
buvmc: Hi @rmzelle. Yes, I'd entered the documentation link from the Universidad Complutense from Madrid because I understand that is needed a link to a documentation manual and we still don't have any like that in the Universidad de León

adam3smith: We do have a strong preference for the actual documentation link -- that's the only way we are able to tell what's right when someone reports an error in the style, for example. If the style guide is not online, you can also just attach it to a post here (note that you can't do that by email; you have to upload it on the github site).

buvmc: I have a little guide about the style, but it has been planned in order to work with the Mendeley software. If that is not a problem, here it is: http://buleria.unileon.es/handle/10612/5930

buvmc:
Hello. It has passed more than a month and I can see that my request to create the Universidad de León - Harvard style is still open. I just want to know if there is some kind of problem with the style in order to try to solve it, or if it is all well, why is not uploaded into the repository. Thank you very much for your patience

adam3smith:
Sorry for the delay. One thing I'm unclear about is the language for the style. You have it set to English, but when I look at your styleguide, I see 

> Vitale, A., Schnell, D., Raikhel, N. V. y Chrispeels, M. J. (2015) "Protein sorting and vesicle traffic", en Buchanan, B. B., Gruissem, W., y Jones, R. L. (eds.) Biochemistry and molecular biology of plants. 2.a
ed. Chichester: Wiley Blackwell, pp. 151-190.

which is clearly Spanish (y, en, 2a). What's correct? (it could also be both, but then we'd need to remove the default-locale entirely)

buvmc:
Hello @adam3smith. Concerning the language for the style, the default is English because it was that way as set in the default of the style that we took as a model (Harvard Cite Them Right 9th edition). For us it is right. We ask our students to set the language of the bibliography according with the language that the work is written, and teach them how to do that on the Mendeley software. The examples of the guide are in Spanish language because the majority of our students write their works in that language. We prefer that they learn to change the language according with what they need, instead of give them "the work done".

adam3smith:
OK, thanks. In that case removing the default-locale entirely is the way to go. We think the Mendeley implementation of this is a bit problematic -- styles with default locales shouldn't typically be used in other languages, and other reference managers, like Zotero, don't allow that. Without default locale, the style will be usable in any software supporting CSL.

PR #4072, opened 2019-04-30
Update american-phytopathological-society.csl

libora:
changed: use et al for first 3 authors insted of 4 (documentation https://forums.zotero.org/discussion/77104/new-american-phytopathology-citation)


adam3smith:
Thanks!

PR #4073, opened 2019-05-02

mserifovic:
Removed date published from conference paper type before place of publication.


adam3smith:
Thanks!

PR #4074, opened 2019-05-02
Update uppsala-universitet-institutionen-for-biologisk-grundutbildnin…

jakobsson-ibg:
…g.csl

Modifying our institution style that has not been updated for some years. Solving 2 problems that caused CSL errors when using the Mendeley client.

### DIFF
--- a/uppsala-universitet-institutionen-for-biologisk-grundutbildning.csl
+++ b/uppsala-universitet-institutionen-for-biologisk-grundutbildning.csl
@@ -11,10 +11,14 @@
       <name>Per Enström</name>
       <email>per.enstrom@ibg.uu.se</email>
     </author>
+    <contributor>
+      <name>Tobias Jakobsson</name>
+      <email>jakobsson@ibg.uu.se</email>
+    </contributor>
     <category citation-format="author-date"/>
     <category field="generic-base"/>
     <summary>Style for Swedish and English texts done at the Biology Education Centre (Institutionen för biologisk grundutbildning) at Uppsala University, Sweden.</summary>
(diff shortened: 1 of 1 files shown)


uppsala-universitet-institutionen-for-biologisk-grundutbildning.csl
```diff
@@ -50,24 +54,39 @@
...
+          <name sort-separator=" " initialize-with="" name-as-sort-order="all" delimiter=", " delimiter-precedes-last="always"/>
+          <label form="short" strip-periods="true" prefix=" (" suffix=")"/>
+          <substitute>
+            <names variable="editor"/>
+            <names variable="translator"/>
+          </substitute>
+        </names>
+      </if>
+      <else>
+        <text value=""/>
```
adam3smith: same here

rmzelle:
> Modifying our institution style that has not been updated for some years. Solving 2 problems that caused CSL errors when using the Mendeley client.

@jakobsson-ibg, sorry we responded so slowly to this. If you can describe the problem you were experiencing in some detail, maybe we can offer an alternative solution. What type of errors were you seeing?

jakobsson-ibg:
Sorry for the late reply from my end as well. High priority tasks and then vacations stopped me from replying sooner.

We had two frequent errors, where Mendeley threw errors. First one because it did not accept a line with "<span style="text-decoration:none;">et al.</span>", so this was removed. Second error was that it did not accept citations for web pages where there was no author name included. The if statements checked this, and provided an alternate solution by setting the string to null.

PR #4082, opened 2019-05-06
Create ausonius-editions-new.csl

ColineRuizDarasse:
Fixing document name.


adam3smith:
Could you say how this relates to the existing Ausonious Editions style? Is it based on it? What changed? Should this replace the existing style?

ColineRuizDarasse:
Hi, 
this is an updated version of the Ausonius Edition style. It presented bugs notably for book chapters and conference proceedings, as well as for dictionary entries. We have slightly modified it and we would like to make it available on the Zotero style depository.

ColineRuizDarasse:
It would replace the former one, indeed.

adam3smith:
reopening

rmzelle:
@adam3smith, can you review and merge these changes? I updated the original "ausonius-editions.csl" with the version in "ausonius-editions-new.csl" and made a few corrections (restored style title, author name). Should be good to go once we delete "ausonius-editions-new.csl" from the patch-4 branch, or we commit separately from this PR.

ausonius-editions.csl
```diff
@@ -134,7 +139,7 @@
       <if is-numeric="edition">
         <group delimiter=", ">
```
NMonteix: The comma should be removed, in order to avoid something like "2e, éd." for a second edition

adam3smith:
Thanks @NMonteix ! Done

adam3smith:
Thanks! (Note that the comma is actually outside the quotation mark; preview gets that wrong)

PR #4087, opened 2019-05-10
Update and rename molecular-biology-and-evolution.csl to molecular-bi…

semiramisCJ:
Editions made to comply with new guidelines from Journal (Molecular Biology and Evolution)
https://academic.oup.com/mbe/pages/General_Author_Guidelines#References


adam3smith:
I'm seeing this correctly that the only changes are about the display of online sources, right? Do we have any sense how, e.g. webpages _should_ be cited? The style probably overdid it with URLs, but it seems odd (and not in line with the CSE manual) to never cite them

adam3smith:
While the guide doesn't include URLs, published articles do, e.g. multiple examples here: https://academic.oup.com/mbe/article/36/7/1405/5378707#136841918

So we're keeping those in.

PR #4091, opened 2019-05-12
add line break trick to bibtex.csl

zhangchenkai:
In general, Citation Style Language do not support line break. So, the generated bibliography via "BibTeX generic citation style" is always mixed all keys together. 
But I found an answer from https://forums.zotero.org/discussion/comment/124644/#Comment_124644. And follow the mleoking's answer, I tested the code and it's worked.
Thus, I want to add this trick to repository and make it available for all users.

### DIFF
--- a/bibtex.csl
+++ b/bibtex.csl
@@ -143,7 +143,7 @@
     </sort>
     <layout>
       <text macro="zotero2bibtexType" prefix=" @"/>
-      <group prefix="{" suffix=" }" delimiter=", ">
+      <group prefix="{&#10; " suffix="&#10;}" delimiter=", &#10;">
         <text macro="citeKey"/>
         <text variable="publisher-place" prefix=" place={" suffix="}"/>
         <!--Fix This-->


adam3smith:
@rmzelle views on this? Given that this won't work reliably across processors, I'm not really sure whether to support this or not.

adam3smith:
@rmzelle I'm inclined to leave this, but please do take a look in case you disagree.

rmzelle:
If this is only to make the BibTeX output prettier, I agree it's probably better to not add special HTML codes. (@adam3smith, do you know of specific CSL processors that struggle with this stuff?)

People can also prettify their BibTeX with a secondary tool (e.g. https://flamingtempura.github.io/bibtex-tidy/).

If this change is required to make the BibTeX output validate and/or follow the official spec, it would be a different discussion.

zhangchenkai:
@adam3smith @rmzelle 
Thank you guys, I understand the concern of Rintze and it's indeed not elegant and safe to use special codes to implement line break. 
I will take you the tool you suggested, and I also hope CSL could handle these characters better in the future.

PR #4095, opened 2019-05-14
add a space between citation and the main text

yiluheihei:


rmzelle:
Hi, sorry for not reviewing and replying sooner, but it's customary in CSL to not include a leading space for "author-date" styles, so we won't accept this change.

PR #4096, opened 2019-05-14
Update journal-of-plant-nutrition-and-soil-science.csl

zsteinmetz:
According to the style guidelines of the journal (see https://wol-prod-cdn.literatumonline.com/pb-assets/assets/15222624/Instructions_for_Authors_August_2018-1533052683593_new-1541429205090.pdf), only author names should be in italic; "and" and "et al." go in normal font style.


zsteinmetz:
Hmm, strange .. the docs seem to allow it, don't they? See http://docs.citationstyles.org/en/stable/specification.html#et-al

adam3smith:
Thanks -- you just had it in the wrong place. the `et-al` element is a child of `names` not of `name`.

PR #4101, opened 2019-05-16
Update ruhr-universitat-bochum-lehrstuhl-fur-industrial-sales-and-ser…

sebknop:
…vice-engineering.csl

Fixed some inconsistencies with the style guide from the Chair for Industrial Sales and Service Engineering.


adam3smith:
Thanks!

PR #4106, opened 2019-05-18
Update chicago-author-date citation with collapse delimiter

gnmackenzie:
Include a semicolon as the delimiter for collapsed year citations. The use of the semicolon is consistent with non-collapsed citations. For example the current version creates (Schrijver 2015, 2018) whereas the proposed version creates (Schrijver 2015; 2018) which is consistent with the standard of (Doe 2011; Smith 2014).

### DIFF
--- a/chicago-author-date.csl
+++ b/chicago-author-date.csl
@@ -597,7 +597,7 @@
       </else>
     </choose>
   </macro>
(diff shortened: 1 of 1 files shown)


adam3smith:
Sorry for the delay. The change you propose shouldn't actually cause the change in output that you describe (if the `after-collapse-delimiter` is identical to the `delimiter` on `layout` it shouldn't change anything at all), but even if it did, (Schrijver 2015, 2018) is correct and (Schrijver 2015; 2018) as per CMoS 15.30:

> Additional works by the same author(s) are cited by date only, separated by commas

PR #4112, opened 2019-05-23
Update bursa-uludag-universitesi-fen-bilimleri-enstitusu.csl

ozhansaglik:
Editing et al options. Before this edition, users did edit the bibliography manually; but now it can be done autonomously.

### DIFF
--- a/bursa-uludag-universitesi-fen-bilimleri-enstitusu.csl
+++ b/bursa-uludag-universitesi-fen-bilimleri-enstitusu.csl
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" et-al-min="3" et-al-use-first="1" page-range-format="expanded" demote-non-dropping-particle="never">
-  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" et-al-use-first="1" page-range-format="expanded" demote-non-dropping-particle="never">
   <info>
-    <title>Bursa Uludağ Üniversitesi - Fen Bilimleri Enstitüsü</title>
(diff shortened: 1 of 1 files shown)


ozhansaglik:
Hi CSL team,

As you see in inline citations we use Saglik et al. ; but in the bibliography, we should write the name of the all authors, without using "..." or "&" etc.

Can you help?

biosainik:
> 
> 
> Hi CSL team,
> 
> As you see in inline citations we use Saglik et al. ; but in the bibliography, we should write the name of the all authors, without using "..." or "&" etc.
> 
> Can you help?

Can you please give a link to check your standard style of bibliography. Alternately, you may reply with a formatted citation (in your style) of a journal article?

biosainik:
> 
> 
> Hi CSL team,
> 
> As you see in inline citations we use Saglik et al. ; but in the bibliography, we should write the name of the all authors, without using "..." or "&" etc.
> 
> Can you help?

@ozhansaglik  Do you use "and" instead of "&"?

ozhansaglik:
@biosainik Goldmann, W., Hunter, N., Foster, J.D., Salbaum, J.M., Beyreuther, K., Hope, J. 1990. Alleles of a neural protein gene linked to scrapie in sheep. Proc. Natl. Acad. Sci., 87(7): 2476-2480. 

You can download the examples from here. http://www.uludag.edu.tr/fenbilimleri/default/konu/4861

We use "and" only for two authors and never use "&".

adam3smith:
@biosainik thanks for chiming in -- getting some help with style review/fixes would be great! Maybe wait, though, for the review on the style you're submitted so you become a bit more familiar with our procedures and standards before you start helping others (which, again, we would greatly appreciate)

biosainik:
@adam3smith thanks for the gentle advice! As you observe, I am a kind of beginner.
If I understand that right, I am expected to properly go through style fixes, by identifying/rectifying the errors in code with minimum efforts.

By the way, am I allowed to commit to issues/pull requests and edit the styles ?
Thanks!

ozhansaglik:
@adam3smith @biosainik Dear Stephen and Pragaadeesh thank you so much for your help.

PR #4114, opened 2019-05-23

fettingderek:
Fixed else clause in source macro.

### DIFF
--- a/bluebook-law-review.csl
+++ b/bluebook-law-review.csl
@@ -16,7 +16,7 @@
     <category citation-format="note"/>
     <category field="law"/>
     <summary>The Bluebook legal citation style for law reviews.</summary>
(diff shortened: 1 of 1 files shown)


mattluce:
Can this be merged?

fettingderek:
Can this be merged please?

adam3smith:
Sorry this took so long. 
I agree with the switch to italics, but I'm not sure moving volume to the back is right. My understanding of BB citations for books (which presumably is the most important category in `else` is that they should begin with the volume?

fettingderek:
Hello,
I've updated the style to display volume ahead of authors for books (instead of in the source macro), and to make book titles small caps.

adam3smith:
Great, thanks!

PR #4116, opened 2019-05-24
Changed the way the intext citations are displayed

davidvenema:


rmzelle:
@davidvenema, thanks! I made a small tweak to the code to make it a little neater.

Note that you can update a pull request after you've created it. See the instructions above (https://github.com/citation-style-language/styles/pull/4116#issuecomment-495470790).

(replaces #4111 and #4115)

PR #4123, opened 2019-05-29
Update inter-research-science-center.csl

POBrien333:
via https://forums.zotero.org/discussion/77580/updating-inter-research-science-center-style-need-to-remove-and-between-authors-in-bibliography


rmzelle:
Thanks!

PR #4124, opened 2019-05-29
Update university-of-bradford-harvard.csl

dzacca:
Changed the following after further revision and feedback to fix the below as per: https://www.bradford.ac.uk/library/find-out-about/referencing/Guide-to-referencing-using-the-Harvard-System--November-2017.pdf

Inline citation: 
- removed page number(s)

Bibliography:
- The ", " delimiter after each group was not correct. Replaced with " " and added ". " after titles and ", " after volume/issue of journals.
- Removed the "p./pp." label before page number(s).
- Changed the format of place and publisher as "Place: Publisher".
- Set the space between bibliography lines to "0" as per the EndNote style provided by the university.


dzacca:
There still is an issues with the bibliography format of book sections. I'll try to fix within the next few days.

PR #4127, opened 2019-06-02
Update university-of-bradford-harvard.csl

dzacca:
Changed the following after further revision to fix the below as per: https://www.bradford.ac.uk/library/find-out-about/referencing/Guide-to-referencing-using-the-Harvard-System--November-2017.pdf

Inline citation: 
- removed page number(s)
- Added page locator (if present).

Bibliography:
- The ", " delimiter after each group was not correct. Replaced with " " and added ". " after titles and ", " after volume/issue of journals.
- Removed the "p./pp." label before page number(s).
- Changed the format of place and publisher as "Place: Publisher".
- Set the space between bibliography lines to "0" as per the EndNote style provided by the university.
- Set italic for the title if publication is post or post-weblog.
- Fixed the separator between publisher and pages for chapters.
- Fixed date and url order for post post-weblog, and websites.
- Reverted back to "Accessed" for access date on post post-weblog, and websites. (There are contrasting directions between the brief citation guide and the full reference. The guidelines from the full reference was used here as it's the same used on EndNote's style from the university)

### DIFF
--- a/university-of-bradford-harvard.csl
+++ b/university-of-bradford-harvard.csl
@@ -19,7 +19,7 @@
     <category citation-format="author-date"/>
     <category field="generic-base"/>
     <summary>The Bradford version of the Harvard author-date style (based on University of Abertay Dundee style from Gregory Goltsov)</summary>
-    <updated>2019-05-24T14:31:50+00:00</updated>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks!

PR #4134, opened 2019-06-09
[fix] some papers failed to get doi & more authors

linhaigoo:
We found some papers failed to get doi because the access value is empty. And only one author showed when authors number is larger than 6.

### DIFF
--- a/science.csl
+++ b/science.csl
@@ -40,14 +40,7 @@
     </names>
   </macro>
   <macro name="access">
-    <choose>
-      <if variable="page" match="none">
-        <text macro="access-value"/>
-      </if>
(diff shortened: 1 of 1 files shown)


adam3smith:
What is the basis for these changes? The Science citation style 
https://www.sciencemag.org/authors/instructions-preparing-initial-manuscript isn't terribly well described, but I didn't see any mention of always including DOIs (the current style only includes it if items don't have page numbers). And while it's unclear to me whether there should be et al at all, the examples they do have use et al after the first, not after the third author.

adam3smith:
I'm going to close this -- happy to reopen or take a new PR if there's guidance related to DOIs from *Science*

PR #4135, opened 2019-06-10
Create mimbar-hukum.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/315058#Comment_315058


adam3smith:
@POBrien333 -- checkout the new feature we have for style review: rendered sample citations and bib in the csl-bot messages! 

You'll note, though, that these have missing commas & spaces that aren't correct in the style. It looks like citeproc-ruby is handling the delimiter differently. Will try to see if we can get Sylvester to adjust that.

POBrien333:
How cool is that?!?!
Amazing what you can do!
Very useful feature!

PR #4137, opened 2019-06-11
[Don't Merge] Testing new bots

adam3smith:


adam3smith:
@retorquere not quite there yet -- see above, somehow Sheldon is failing to load

retorquere:
I'd need to see the heroku log to see whether something gets delivered. Another thing that would be useful is if Sheldon-gem can be ran with the --verbose flag.

retorquere:
I can open my own PR to test the client side. But someone would either have to copy out the heroku logs to here or I'd need to have temporary access to Sheldon.

Another option is that I set up a heroku-18 stack and that I transfer it when I have it working, but someone would have to set secret vars to get it to work. I also don't know if I can migrate to the existing name.

adam3smith:
Happy to give you access. What email do you use?

retorquere:
It may not be required, see the new PR (sloppiness on my part, really sorry) -- but otherwise, emiliano.heyns@iris-advies.com

adam3smith:
added you for good measure. Feel free to also update the stack

PR #4141, opened 2019-06-11
[DO NOT MERGE] Sheldon test

retorquere:
Get the new shel-gem, trigger a report.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+.env
+spec/sheldon/travis.json
--- a/.rspec
+++ b/.rspec
@@ -1,3 +1,5 @@
 --format Fuubar
 --color
 --require spec_helper.rb
+--format json
+--out spec/sheldon/travis.json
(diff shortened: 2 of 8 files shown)


retorquere:
@adam3smith so this is better but still not what I would have expected.

retorquere:
Strange... when I render this locally, I still get `[,]` as the in-text citations. This can't be right. @adam3smith?

retorquere:
Is it possible that this style has no in-text citations? Is it a footnote style?

adam3smith:
Ah, interesting -- it's a numeric style and `citation-number` is generated dynamically by citeprocs and that doesn't seem to work here. We could try to just add citation-number in the JSON hardcoded. Very few cases where this would matter for our purposes.

retorquere:
*phew* I thought I had screwed something else up.

retorquere:
Aw yeah.

PR #4144, opened 2019-06-13
Correcting the 'and' separator for author-short

gaballench:
The instructions for authors indicate that the author separator should be ampersand for in-text citations instead of the 'and'; conversely, it should be textual (and) for the references section (both for authors and editors). The proposed change in the tag  in L42 produces the desired output.


adam3smith:
Thanks -- the old documentation, which is still up, does call for "and", but they clearly do want &

PR #4149, opened 2019-06-17
Create the-cancer-journal.csl

POBrien333:
via https://forums.zotero.org/discussion/76743/style-request-the-cancer-journal


rmzelle:
Thanks!

PR #4154, opened 2019-06-21
Create acta-ichthyologica-et-piscatoria.csl

POBrien333:
via https://forums.zotero.org/discussion/76794/style-request-acta-ichthyologica-et-piscatoria

### DIFF
--- a/acta-ichthyologica-et-piscatoria.csl
+++ b/acta-ichthyologica-et-piscatoria.csl
@@ -1,0 +1,164 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" default-locale="en-US" version="1.0" demote-non-dropping-particle="never">
+  <info>
+    <title>Acta Ichthyologica et Piscatoria</title>
+    <id>http://www.zotero.org/styles/acta-ichthyologica-et-piscatoria</id>
+    <link href="http://www.zotero.org/styles/acta-ichthyologica-et-piscatoria" rel="self"/>
+    <link rel="template" href="http://www.zotero.org/styles/wetlands"/>
+    <link href="https://aiep.pl/main/instr.php" rel="documentation"/>
+    <author>
+      <name>Sebastian Karcher</name>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="science"/>
+    <issn>0137-1592</issn>
+    <eissn>1734-1515</eissn>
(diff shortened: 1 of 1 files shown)


acta-ichthyologica-et-piscatoria.csl
```diff
@@ -4,10 +4,11 @@
     <title>Acta Ichthyologica et Piscatoria</title>
     <id>http://www.zotero.org/styles/acta-ichthyologica-et-piscatoria</id>
     <link href="http://www.zotero.org/styles/acta-ichthyologica-et-piscatoria" rel="self"/>
-    <link rel="template" href="http://www.zotero.org/styles/wetlands"/>
+    <link href="http://www.zotero.org/styles/wetlands" rel="template"/>
     <link href="https://aiep.pl/main/instr.php" rel="documentation"/>
     <author>
-      <name>Sebastian Karcher</name>
+      <name>Patrick O'Brien, PhD</name>
+      <email>obrienpat86@gmail.com</email>
```
rmzelle: @adam3smith, @POBrien333, I assume Patrick just forgot to update the name here.

acta-ichthyologica-et-piscatoria.csl
```diff
@@ -0,0 +1,164 @@
...
+      <key macro="author"/>
+      <key variable="author" sort="ascending"/>
+    </sort>
+    <layout>
+      <group delimiter=". ">
+        <group delimiter=". " suffix=".">
+          <text macro="author" font-weight="bold"/>
+          <text macro="year"/>
+          <text macro="title"/>
+          <choose>
```
adam3smith: Just an FYI if you're looking at the previews -- citeproc-ruby doesn't apply group delimiters within `choose` elements contained in the group. This is fine with other citeprocs and I think should be changed in citeproc-ruby (which is doing this technically correctly according to the specs) and it's definitely fine in the styles, so just mentioning this so you don't worry about it.

rmzelle: > This is fine with other citeprocs and I think should be changed in citeproc-ruby (which is doing this technically correctly according to the specs) and it's definitely fine in the styles, so just mentioning this so you don't worry about it.

My thinking about this has always been that style authors should be able to add `<choose/>` elements (e.g. to cover exceptions) without affecting the original functioning of the style, so yes, we should clarify this in the spec if needed and get citeproc-ruby to push the delimiter down into the `<choose/>` contents.

adam3smith:
Thanks!

PR #4155, opened 2019-06-25
Update berghahn-books-author-date-en-gb.csl

POBrien333:
small fixes:
1. add et-al settings for bibliography
2. fix a suffix


rmzelle:
Thanks!

PR #4156, opened 2019-06-25
Update american-meteorological-society.csl

POBrien333:
adjustments via https://forums.zotero.org/discussion/comment/334754/#Comment_334754


rmzelle:
Thanks!

PR #4157, opened 2019-06-25
Update r-and-d-management.csl

POBrien333:
Via emails:
- remove "note" field from citation style
- made it more robust for some item types


rmzelle:
Thanks!

PR #4160, opened 2019-06-30
Create nations-and-nationalism.csl

POBrien333:
via https://forums.zotero.org/discussion/77386/style-request-nations-and-nationalism-journal
(low-quality template style...)

PR #4161, opened 2019-06-30
Create nations-and-nationalism.csl

POBrien333:
via https://forums.zotero.org/discussion/77386/style-request-nations-and-nationalism-journal#latest
(crap template style...)


POBrien333:
duplication of #4160

Issue #3881, opened 2019-01-17
New Style: Science Advances


psychelzh:
Based on [the instructions](http://advances.sciencemag.org/content/information-authors) of this journal, it is a little different from the style of **Science**. Specifically,

> **Science Advances** uses a complete citation format that includes all authors, full titles of journal articles, the journal abbreviation, the volume, the first and last page, and the year of publication. The absolute formatting (what is bold and what is italic) is less important than having a complete citation for each journal article cited.

>The abbreviations for journal names are taken from the Bibliographic Guide for Editors and Authors (BGEA) or Serial Sources for the BIOSIS Data Base *(BIOSIS), a more recent publication. When in doubt, provide the journal’s complete name. Spell out cities that are listed after a journal name: *Acta Zool. (Stockholm). Do not use op. cit., ibid., 3-m dashes, en dashes, or et al. (in place of the complete list of authors’ names). For author names with Jr. or 2nd, etc. see example number 4 in the Journals section.

>In references to books or chapters in books, publisher names are given in shortened form. “Press” is usually dropped. Exceptions: Academic Press (“Academic” is an adjective), Cold Spring Harbor Laboratory Press, National Academies Press, University Park Press, CRC Press, MIT Press, and Cambridge Univ. Press (for university presses, to distinguish them from the university itself). Only one publisher city is needed. A few world-renowned cities (for example, Amsterdam, London, Philadelphia, Chicago, New York, Baltimore) can be listed without state or country; less well-known cities and those with names that could be confused take state abbreviations (Cambridge alone for the city in the U.K., but Cambridge, MA). Inclusive page numbers or a chapter number must be given when specific articles are referred to within an edited volume.

bwiernik:
Based on the description you are giving, the existing submission style for Science (which has titles and which is what the existing Science Advances style points to) fits all these requirements to the degree possible by CSL (e.g., CSL can't do things like drop "Press" or country names if they are supplied, but my understanding of Science's policies is that they aren't too concerned about that).

https://www.zotero.org/styles/science

psychelzh:
I do not know if I am correct. I think seemingly **Science** and **Science Advances** both requires to include all authors, just like:

[...]

PR #4163, opened 2019-07-02
remove et-al from science.csl

POBrien333:
via https://forums.zotero.org/discussion/78093/style-error-science-advances
also, see https://github.com/citation-style-language/styles/issues/3881


adam3smith:
Thanks!

PR #4167, opened 2019-07-03
Update zookeys.csl

tapani-hopkins:
Minor updates to match the current ZooKeys guidelines:
Citations should be disambiguated with e.g. 2019a, 2019b, not by adding names.
Bibliography sorted by first author, then number of authors, and only then year.

Also made minor changes to how the DOI and the editors are formatted. (The latter I'm unsure about. ZooKeys may not require that journal articles show the editors at all, one option would be to delete line 228:  <text macro="editor"/> )


rmzelle:
Great, thanks!

PR #4168, opened 2019-07-03
Create helvetica-chimica-acta

daenuprobst:
Was not yet available.

### DIFF
--- a/helvetica-chimica-acta
+++ b/helvetica-chimica-acta
@@ -1,0 +1,169 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>Helvetica Chimica Acta</title>
+    <id>http://www.zotero.org/styles/angewandte-chemie</id>
+    <link href="http://www.zotero.org/styles/helvetica-chimica-acta" rel="self"/>
+    <link href="http://www.helvchimacta.ch/downloads/Instructions_for_Authors.pdf" rel="documentation"/>
+    <author>
+      <name>Daniel Probst</name>
+      <email>daenuprobst@gmail.com</email>
+      <uri>http://danielprobst.net</uri>
+    </author>
+    <contributor>
+      <name>Daniel Probst</name>
+    </contributor>
+    <category citation-format="numeric"/>
+    <category field="engineering"/>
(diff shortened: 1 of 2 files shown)


helvetica-chimica-acta
```diff
@@ -0,0 +1,169 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>Helvetica Chimica Acta</title>
+    <id>http://www.zotero.org/styles/angewandte-chemie</id>
```
POBrien333: 1. change this line to match with your self link.
2. And I guess angewande-chemie was your template style. Please add that below the self link with rel="template".
3. add the .csl extension to your filename.

helvetica-chimica-acta.csl
```diff
@@ -0,0 +1,166 @@
...
+      <email>daenuprobst@gmail.com</email>
+      <uri>http://danielprobst.net</uri>
+    </author>
+    <category citation-format="numeric"/>
+    <category field="engineering"/>
+    <category field="chemistry"/>
+    <issn>0018-019X</issn>
+    <eissn>1522-2675</eissn>
+    <summary>A style for Wiley-VCH's journal "Helvetica Chimica Acta" based on "Angewandte Chemie International Edition".
+This style has many limits due to csl constraints, most notably the inability to include pages/page ranges for books, the wrong square brackets for two items cited together, and the lack of support for citing 1a)... b)...c)....</summary>
```
rmzelle: > This style has many limits due to csl constraints, most notably the inability to include pages/page ranges for books, the wrong square brackets for two items cited together, and the lack of support for citing 1a)... b)...c)....

@daenuprobst, can you describe the first two problems in more detail? We're aware of the third point, which is indeed currently not possible with CSL (https://github.com/citation-style-language/schema/issues/36).

adam3smith: 1) Refers to providing essentially a pinpoint to a specific passage in a book in a numeric style. CSL can kind of do this by using `text variable="page"` but that requires separate items every time you want to add a different part of a book, so isn't really feasible. That said, this isn't actually used in the style or in published articles, so removing it from the summary
2) The style using [1-2] instead of [1,2] which CSL can't do
3) The style explicitly does not use composit (1a) b) c)) references, so I'll take this out of the summary, too.

adam3smith:
Great, thanks!

PR #4169, opened 2019-07-04
Update nations-and-nationalism.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/335360/#Comment_335360
small fixes


rmzelle:
Thanks!

PR #4176, opened 2019-07-05
Create Revista-Agulhas-Negras-AMAN

rcleoni:
Normas ABNT para a Revista Agulhas Negras da Academia Militar das Agulhas Negras - AMAN


adam3smith:
I'm trying to understand the relationship between this style and the one proposed here: https://github.com/citation-style-language/styles/pull/4177
are they at all different? Is the style for the journal documented somewhere?

PR #4177, opened 2019-07-05
Update and rename associacao-brasileira-de-normas-tecnicas-eceme.csl …

rcleoni:
…to associacao-brasileira-de-normas-tecnicas-aman.csl


adam3smith:
@rmzelle I don't understand the Travis failure, I think it's a Sheldon bug. Could you double-check I'm not overlooking something?

Issue #4178, opened 2019-07-07
Structural breakage in many styles (cs:substitute + cs:choose)


fbennett:
While working through a CSL-M style bug with @denismaier, we traced the fault to use of `cs:choose` immediately under `cs:substitute`, without and enclosing `cs:group`. The style concerned was based on a style from the CSL repository, do I did a check for similar structures, and was alarmed to find that the structure, which can easily break formatting in ways that are difficult to debug when used, is quite common. A bare `cs:choose` element occurs under `cs:substitute` in 397 styles, and a condition node (`cs:if`, `cs:else-if`, or `cs:else`) contains multiple ungrouped elements (which is particularly bad) in 127 styles.

One of the ways that the latter of these structures will break formatting in `citeproc-js` is to grab the first-rendered element as the content of `display="block"`, and dump out the remaining elements in another context. In any case, the context is not properly closed in the processor, so all sorts of strange and wonderful things can happen. It's a bundle of styles that need attention, but they should really be fixed. The quick-and-dirty method that will work in most cases is just to add a `cs:group` wrapper around any `cs:choose` that occurs immediately under a `cs:substitute` element.

Here is the script I used for the audit. To run it, copy the code into the top of the styles repo, do `npm i xmldom` and `npm i xpath`, and run with `node ./NAME-OF-SCRIPT-FILE.js`.

```javascript
var fs = require("fs");
const dom = require('xmldom').DOMParser
const xpath = require("xpath");

console.log("#Styles with bare choose elements in cs:substitute");

var count = 0;
var noGroupList = [];
var countMultiElement = 0;
var multiElementList = [];

for (var fn of fs.readdirSync(".")) {
    if (fn.slice(-4) !== ".csl") {
        continue;
    }
    var xml = fs.readFileSync(fn).toString();
    xml = xml.replace("xmlns=\"http://purl.org/net/xbiblio/csl\"", "");
    var doc = new dom().parseFromString(xml)
    var nodes = xpath.select("//substitute/choose", doc);
    if (nodes.length > 0) {
        var hasGroup = false;
        var hasMultipleElements = false;
        for (var node of nodes) {
            var groups = xpath.select(".//group", node);
            if (groups.length > 0) {
                hasGroup = true;
            } else {
                var ifs = xpath.select(".//if", node);
                for (var ifNode of ifs) {
                    if (xpath.select("./*", ifNode).length > 1) {
                        hasMultipleElements = true;
                    }
                }
[...]

PR #4179, opened 2019-07-07
Wrap cs:choose children of cs:substitute in cs:group element

fbennett:
Not sure how this will be received, but for what it's worth. I tweaked the audit script quoted in https://github.com/citation-style-language/styles/issues/4178 a bit, and added `cs:group` wrappers to bare `cs:choose` children of `cs:substitute` everywhere. The group wrapper has no delimiter, but none would have been provided by the `cs:choose` parent anyway, so it should be okay. He said.

### DIFF
--- a/academy-of-management-review.csl
+++ b/academy-of-management-review.csl
@@ -73,14 +73,16 @@
       <substitute>
         <names variable="editor"/>
         <names variable="translator"/>
-        <choose>
-          <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
-            <text variable="title" form="short" font-style="italic"/>
-          </if>
-          <else>
-            <text variable="title" form="short" quotes="true"/>
-          </else>
-        </choose>
+        <group>
+          <choose>
+            <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
+              <text variable="title" form="short" font-style="italic"/>
+            </if>
+            <else>
+              <text variable="title" form="short" quotes="true"/>
+            </else>
+          </choose>
+        </group>
       </substitute>
     </names>
   </macro>
--- a/accident-analysis-and-prevention.csl
+++ b/accident-analysis-and-prevention.csl
@@ -60,14 +60,16 @@
       <substitute>
         <names variable="editor"/>
         <names variable="translator"/>
-        <choose>
-          <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
-            <text variable="title" form="short" font-style="italic"/>
-          </if>
-          <else>
-            <text variable="title" form="short" quotes="true"/>
-          </else>
-        </choose>
+        <group>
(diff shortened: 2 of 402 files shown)


fbennett:
Here is the script used for the conversion, in case it's useful for post-processing in future:
```javascript
var fs = require("fs");
const dom = require('xmldom').DOMParser
const xpath = require("xpath");
const format = require('xml-formatter');
const crlf = require('crlf');
 
const select = xpath.useNamespaces({"cs": "http://purl.org/net/xbiblio/csl"});

console.log("#Styles with bare choose elements in cs:substitute");

var count = 0;
var noGroupList = [];
var countMultiElement = 0;
var multiElementList = [];

for (var fn of fs.readdirSync(".")) {
    if (fn.slice(-4) !== ".csl") {
        continue;
    }
    var xml = fs.readFileSync(fn).toString();
    var doc = new dom().parseFromString(xml)
    var nodes = select("//cs:substitute/cs:choose", doc);
    if (nodes.length > 0) {
        var hasGroup = false;
        var hasMultipleElements = false;
        for (var node of nodes) {
            var groups = select(".//cs:group", node);
            if (groups.length > 0) {
                hasGroup = true;
            } else {
                var ifs = select(".//cs:if", node);
                for (var ifNode of ifs) {
                    if (select("./cs:*", ifNode).length > 1) {
                        hasMultipleElements = true;
                    }
                }
                var elseifs = select(".//cs:else-if", node);
                for (var elseifNode of elseifs) {
                    if (select("./cs:*", elseifNode).length > 1) {
                        hasMultipleElements = true;
                    }
                }
                var elses = select(".//cs:else", node);
                for (var elseNode of elses) {
                    if (select("./cs:*", elseNode).length > 1) {
                        hasMultipleElements = true;
                    }
                }
            }
        }
        if (!hasGroup) {
            noGroupList.push(fn);
            count++;
            if (hasMultipleElements) {
                multiElementList.push(fn);
                countMultiElement++;
            }
        }
        for (var choose of nodes) {
            var parent = choose.parentNode;
            var group = doc.createElement("group");
            var newChoose = choose.cloneNode(true);
            group.appendChild(newChoose);
            parent.replaceChild(group, choose);
        }
        xml = format(doc.toString().replace(/\>[\s\n]+\</g, "><"), {indentation: '  ', stripComments: false, collapseContent: true});
        fs.writeFileSync(fn, xml);

        crlf.set(fn, 'LF', function(err, endingType) {
            // file now uses LF
        });


    }
}

console.log("\n## Styles with bare cs:choose: "+count);
console.log("- " + noGroupList.join("\n- "));

console.log("\n## Styles with multiple elements under if, else-if, or else: "+countMultiElement);
console.log("- " + multiElementList.join("\n- "));
```

fbennett:
Closed because this was both unnecessary and wrong.

PR #4180, opened 2019-07-08
Added original dates of publication

JoshuaWeberFHNW:
I have added the original dates of publication in bibliography and in-line citation. Thanks in advance.


rmzelle:
Thanks!

PR #4183, opened 2019-07-08
Update karger-journals.csl

POBrien333:
via https://forums.zotero.org/discussion/78163/update-neuroendocrinology-style


adam3smith:
Thanks!

PR #4185, opened 2019-07-09
Buletin-Agrohorti.csl

tosptommy:
This is a citation style for Buletin Agrohorti, which is the academic journal from Agronomy and Horticulture Department, Faculty of Agriculture, IPB University, Bogor, Indonesia. The journal can be found in journal.ipb.ac.id/index.php/bulagron/index .


rmzelle:
@tosptommy, thanks. Did you start from an existing CSL style, and if so, do you remember which one? We try to add a "template" link to any derived CSL style (see also https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link).

adam3smith:
Thanks!

PR #4186, opened 2019-07-10
Update journal-of-the-american-ceramic-society.csl

POBrien333:
via https://forums.zotero.org/discussion/78190/style-error-journal-of-the-american-ceramic-society
Code from Hypolus and then overwritten with mine to see difference.


rmzelle:
Thanks!

PR #4083, opened 2019-05-07
Create aiaa-journal-2019.csl

keeganskateszero:
Updated based on most recent style requirements of AIAA Journal.

PR #4189, opened 2019-07-10
Update american-institute-of-aeronautics-and-astronautics.csl

rmzelle:
Per https://github.com/citation-style-language/styles/pull/4083 (after reindenting, and keeping the original cs:info block)


rmzelle:
@adam3smith, haven't verified it against the official style guide, but I trust it's an improvement. Feel free to merge.

adam3smith:
Yes, looks good, thanks!

Issue #4140, opened 2019-06-11
PLOS style references include publisher for journal articles


dhimmel:
[PLOS Computational Biology](https://journals.plos.org/ploscompbiol/s/submission-guidelines#loc-references) (and PLOS) show an example journal article reference of:

> 1. Hou WR, Hou YL, Wu GF, Song Y, Su XL, Sun B, et al. cDNA, genomic sequence cloning and overexpression of ribosomal protein gene L9 (rpL9) of the giant panda (Ailuropoda melanoleuca). Genet Mol Res. 2011;10: 1576-1588.
>
> 2. Devaraju P, Gulati R, Antony PT, Mithun CB, Negi VS. Susceptibility to SLE in South Indian Tamils may be influenced by genetic selection pressure on TLR2 and TLR9 genes. Mol Immunol. 2014 Nov 22. pii: S0161-5890(14)00313-7. doi: 10.1016/j.molimm.2014.11.005.

Note that publisher information is not included (at least when `container-title` is known). However, we find [here](https://greenelab.github.io/meta-review/v/040638d25b9a45ad3d060afc9689a17f4c3a3045/#ref-12sHvZy1a) that the publisher is included (in this case "Oxford University Press"):

> 4. Khare R, Good BM, Leaman R, Su AI, Lu Z. Crowdsourcing in biomedicine: challenges and opportunities. Brief Bioinform. Oxford University Press; 2016;17: 23–32. doi:10.1093/bib/bbv021

The CSL for this item is:

```json
  {
    "source": "PubMed",
    "accessed": {
      "date-parts": [
        [
          2019,
          4,
          4
        ]
      ]
    },
    "id": "12sHvZy1a",
    "title": "Crowdsourcing in biomedicine: challenges and opportunities",
    "author": [
      {
        "family": "Khare",
        "given": "Ritu"
      },
      {
        "family": "Good",
        "given": "Benjamin M"
      },
      {
        "family": "Leaman",
        "given": "Robert"
      },
      {
        "family": "Su",
        "given": "Andrew I"
      },
      {
        "family": "Lu",
        "given": "Zhiyong"
      }
    ],
    "container-title-short": "Brief Bioinform",
    "container-title": "Briefings in bioinformatics",
    "publisher": "Oxford University Press",
    "ISSN": "1467-5463",
    "issued": {
      "date-parts": [
        [
          2016,
          1
        ]
      ]
    },
    "page": "23-32",
    "volume": "17",
    "issue": "1",
[...]

PR #4191, opened 2019-07-12
Update plos.csl to exclude publisher from journals

POBrien333:
To close https://github.com/citation-style-language/styles/issues/4140.


adam3smith:
Thanks!

PR #4195, opened 2019-07-15
Update nations-and-nationalism.csl

POBrien333:
Small fixes via https://forums.zotero.org/discussion/comment/335973/#Comment_335973


rmzelle:
Thanks!

PR #4196, opened 2019-07-15
Create acta-zoologica-academiae-scientiarum-hungaricae.csl

POBrien333:
via https://forums.zotero.org/discussion/75958/style-request-acta-zoologica-academiae-scientiarum-hungaricae


rmzelle:
Thanks!

PR #4200, opened 2019-07-16
Disambiguation by add year suffix was not present - Added it to the s…

lorenzino:
…tyle

It was:
 <citation et-al-min="3" et-al-use-first="1" disambiguate-add-names="true" disambiguate-add-givenname="true" givenname-disambiguation-rule="primary-name">

Now is:  
<citation et-al-min="3" et-al-use-first="1" disambiguate-add-names="true" disambiguate-add-givenname="true" givenname-disambiguation-rule="primary-name" disambiguate-add-year-suffix="true">
disambiguate-add-year-suffix="true"


adam3smith:
Thanks! Sorry this took so long.

PR #4201, opened 2019-07-16
Create SU-DGES Style

OGMalan:
Creating a csl style based on the Stellenbosch University Department of Geography and Environmental Studies referencing guidelines.

### DIFF
--- a/stellenbosch-university-department-of-geography-and-environmental-studies-harvard.csl
+++ b/stellenbosch-university-department-of-geography-and-environmental-studies-harvard.csl
@@ -19,8 +19,8 @@
   </info>
   <macro name="editor">
     <names variable="editor">
-      <name font-style="normal" delimiter=" " and="symbol" initialize-with="" name-as-sort-order="all" sort-separator=" "/>
-      <label form="short" text-case="lowercase" strip-periods="true" font-style="normal" prefix=" (" suffix=")"/>
+      <name delimiter=" " and="symbol" initialize-with="" name-as-sort-order="all" sort-separator=" "/>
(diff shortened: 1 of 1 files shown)


OGMalan:
I would greatly appreciate any help in fixing the remaining type issue with the "stellenbosch-university-department-of-geography-and-environmental-studies-harvard: name nodes must have valid et-al-min and et-al-use-first attributes" error, as I do not understand why I am getting it.

adam3smith:
Is et al ever used in the bibliography?

OGMalan:
> Is et al ever used in the bibliography?

As far as I can tell, yes, but I am not sure. The style works as intended when I manually install it with Mendeley, or use it with Zotero and I really do not understand at all why I am getting that error.

Thank you for your help.

adam3smith:
Sorry, you're misunderstanding my question: according to the style guide, should et al. be used in the bibliography? If so, when exactly? 
The problem is that you currently that you have `et-al-use-first` without `et-al-min` but more important than fixing the automated test (which I can do for you) is to fix it _correctly_.

OGMalan:
> Sorry, you're misunderstanding my question: according to the style guide, should et al. be used in the bibliography? If so, when exactly?
> The problem is that you currently that you have `et-al-use-first` without `et-al-min` but more important than fixing the automated test (which I can do for you) is to fix it _correctly_.

Oh sorry for misunderstanding! No, et al is used in-text when a work has more than 3 authors. Check the last bullet of page 20 in the research report guidelines (the documentation link).

adam3smith:
@rmzelle style is good to go. If you could check metadata & merge, please?

PR #4205, opened 2019-07-17
Create journal-of-political-ideologies.csl

POBrien333:
via https://forums.zotero.org/discussion/75159/style-journal-of-political-ideologies


rmzelle:
Thanks!

PR #4207, opened 2019-07-18
Create journal-of-egyptian-history.csl

POBrien333:
via https://forums.zotero.org/discussion/78039/style-request-journal-of-egyptian-history

### DIFF
--- a/journal-of-egyptian-history.csl
+++ b/journal-of-egyptian-history.csl
@@ -465,9 +465,6 @@
       </if>
     </choose>
   </macro>
-  <macro name="event-note">
-    <text variable="event"/>
-  </macro>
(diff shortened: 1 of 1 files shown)


rmzelle:
@POBrien333, I assume you only tested this style for English? I guess we can just leave off the default-locale, mark the style as polyglot, and hope for the best.

https://brill.com/fileasset/downloads_products/Author_Instructions/JEH.pdf says "Manuscripts will be accepted in English(British or American spelling), German, and French."

POBrien333:
@rmzelle 
Thanks for spotting that. Hadn't seen that and the style is just optimised for English. :)

rmzelle:
Thanks!

PR #4209, opened 2019-07-19
Update university-of-gothenburg-apa-swedish-legislations.csl

david-kjellin:
Updated author abbreviation with more than 3 authors, so that the styles conforms with standard APA. This is an updated to line 656.


rmzelle:
Thanks!

(see also https://github.com/ub-digit/citation-styles/commit/2410f3843c1fdc987408e55fd380e98fce5bd44f)

PR #4214, opened 2019-07-22
Create ergo.csl

POBrien333:
via https://forums.zotero.org/discussion/78071/citation-style-for-open-access-philosophy-journal-ergo#

APA with small changes:
- in text locator delimiter
- bib: full names


rmzelle:
Thanks!

PR #4215, opened 2019-07-23
Update wiesbaden-business-school.csl

POBrien333:
As per: https://forums.zotero.org/discussion/comment/336542/#Comment_336542
- remove ibid
- change and adapt to polyglot style for both languages

Question:
Technically this style should show "o.V." for ohne Verfasser via "anonymous" for items without author, editor or an authority, right?
It doesn't do so on the VisualEditor for the Webpage item we have on there (CSL Search).

### DIFF
--- a/wiesbaden-business-school.csl
+++ b/wiesbaden-business-school.csl
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only">
-    <!-- Polyglot; journal publishes in English and German. -->
+  <!-- Polyglot; journal publishes in English and German -->
   <info>
     <title>Wiesbaden Business School</title>
     <id>http://www.zotero.org/styles/wiesbaden-business-school</id>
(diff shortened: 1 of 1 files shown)


rmzelle:
> Question:
> Technically this style should show "o.V." for ohne Verfasser via "anonymous" for items without author, editor or an authority, right?
> It doesn't do so on the VisualEditor for the Webpage item we have on there (CSL Search).

I only get it to work with Zotero 5.0.72 if I enclose the call to the "anonymous" term in a group, i.e.:

```xml
      <substitute>
        <names variable="editor"/>
        <text variable="authority"/>
        <group>
          <text term="anonymous" form="short"/>
        </group>
      </substitute>
```

@adam3smith, before we use the bat-signal to get Frank's attention, do you recall this being necessary for citeproc-js?

adam3smith:
> @adam3smith, before we use the bat-signal to get Frank's attention, do you recall this being necessary for citeproc-js?

Sorry for the late reply -- yes, this has been the case for a long time and is related to implicitly conditional in groups. Most styles handle this with a macro instead of a group, which I find more elegant.

rmzelle:
> yes, this has been the case for a long time and is related to implicitly conditional in groups. Most styles handle this with a macro instead of a group, which I find more elegant.

Can you elaborate a little? The following vanilla statement didn't work for the "anonymous" term. What's the problem here exactly?

```xml
      <substitute>
        <names variable="editor"/>
        <text variable="authority"/>
        <text term="anonymous" form="short"/>
      </substitute>
```

rmzelle:
Thanks!

PR #4216, opened 2019-07-23
Create british-journal-of-clinical-pharmacology.csl

POBrien333:
via https://forums.zotero.org/discussion/78000/style-request-british-journal-of-clinical-pharmacology

.pdfs show the citations in superscript. So apart from the DOI format the style just follows AMA as they say in the guidelines.
I´ve created this independent style, but let me know if a dependent is better here (and if we should changed the doi format for AMA)?


rmzelle:
> So apart from the DOI format the style just follows AMA as they say in the guidelines.
> I´ve created this independent style, but let me know if a dependent is better here (and if we should changed the doi format for AMA)?

Let's keep it as an independent. AMA was asked about the DOI format specifically and didn't indicate the new format was acceptable. (see the absence of DOI guidance updates at https://www.amamanualofstyle.com/page/updates, and their Twitter response at https://twitter.com/AMAManual/status/842018162702991363)

rmzelle:
Actually, since https://bpspubs.onlinelibrary.wiley.com/hub/journal/13652125/author-guidelines.html specifically says "References should be cited using the AMA (American Medical Association) style.", and doesn't give any examples of how DOIs should be formatted, I would just turn it into a dependent of the existing AMA CSL style and ignore the difference. Things will align themselves once AMA wakes up and adopts the current preferred DOI format.

rmzelle:
Thanks!

PR #4217, opened 2019-07-24
Update ergo.csl

POBrien333:
small fixes in addition to #4214 via https://forums.zotero.org/discussion/comment/336626/#Comment_336626.


rmzelle:
Thanks!

PR #4221, opened 2019-07-24

jjrondeau:
This version has been completely revised, starting from a leaner base. This is a shorter and better structured version.


rmzelle:
Thanks!

PR #4224, opened 2019-07-25
Create archaeologia-austriaca.csl

POBrien333:
via https://forums.zotero.org/discussion/73988/request-for-archa-citation-style


rmzelle:
Thanks! (since the journal is bilingual, I went ahead and removed the default-locale)

PR #4228, opened 2019-07-28
Create iubmb-life.csl

POBrien333:
via https://forums.zotero.org/discussion/75625/i-want-the-style-for-this-reference


rmzelle:
Thanks!

PR #4231, opened 2019-07-30
Create zilsel.csl

sebastien-plutniak:
Creation of a style for the Zilsel journal.

### DIFF
--- a/zilsel.csl
+++ b/zilsel.csl
@@ -3,10 +3,9 @@
   <info>
     <title>Zilsel. Science, technique, société</title>
     <title-short>Zilsel</title-short>
-	<id>http://www.zotero.org/styles/zilsel</id>
+    <id>http://www.zotero.org/styles/zilsel</id>
     <link href="http://www.zotero.org/styles/zilsel" rel="self"/>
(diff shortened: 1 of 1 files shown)


sebastien-plutniak:
I'm sorry but I don't understand why the checks failed =/ 
Someone might help?

rmzelle:
> I'm sorry but I don't understand why the checks failed =/
> Someone might help?

@sebastien-plutniak, there was an issue on our side that caused all tests to fail. Should be fixed now.

I assume this should be a "note" type citation style, with footnotes on each page? And is this CSL style for both French and English, or just French? We're also curious if you based this on an existing CSL style, in which case we'd like to include a "template" link (see also https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link).

rmzelle:
I was also wondering if the journal has an online description of the citation format (e.g. in a style guide).

sebastien-plutniak:
Thank you, I confirm that this is a “note” type citation style. I added a template reference to the style I used as a model. Unfortunately, the journal does not have yet an online description of the style.

rmzelle:
Thanks!

PR #4232, opened 2019-07-31
Create revue-d-histoire-des-sciences-humaines.csl

sebastien-plutniak:
Creation of the style.


rmzelle:
@sebastien-plutniak, similar questions here: is there an online style guide for this journal that's publicly accessible, and did you base this on an existing CSL style?

rmzelle:
Thanks!

PR #4233, opened 2019-08-01
Create research-institute-for-nature-and-forest.csl

florisvdh:
This is the citation style used at the Research Institute for Nature and Forest (INBO). www.inbo.be


rmzelle:
@florisvdh, thanks! Is the style guide on which you based this CSL style available online, so we can link to it? See also https://github.com/citation-style-language/styles/wiki/Style-Requirements#10---documentation-link.

And did you base this on an existing CSL style? If so, we'd like to know which, so we can add that information to the style as well. See also https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link.

florisvdh:
Hi @rmzelle , I've gathered some feedback from a few colleagues but it appears to be 'no' for both questions: not online, not derived from a particular CSL style as far as I could go back.

Thanks for updating the file! I'd rather suggest to switch Dutch and English in titles (for international context); I will push a commit to propose that.

research-institute-for-nature-and-forest.csl
```diff
@@ -0,0 +1,267 @@
...
+        <group prefix=". " delimiter=", ">
+          <text variable="container-title"/>
+          <text variable="collection-title"/>
+        </group>
+      </else>
+    </choose>
+  </macro>
+  <macro name="author">
+    <names variable="author">
+      <name and="symbol" delimiter-precedes-last="never" et-al-min="10" et-al-use-first="10" initialize-with="." name-as-sort-order="all" sort-separator=" "/>
```
rmzelle: `et-al-min="10" et-al-use-first="10"`

@florisvdh, `et-al-min` always needs to have a higher value than `et-al-use-first`. Per http://docs.citationstyles.org/en/1.0.1/specification.html#name:

> If the number of names in a name variable matches or exceeds the number set on et-al-min, the rendered name list is truncated after reaching the number of names set on et-al-use-first.

research-institute-for-nature-and-forest.csl
```diff
@@ -0,0 +1,267 @@
...
+      <substitute>
+        <names variable="editor"/>
+        <names variable="translator"/>
+        <text macro="title"/>
+      </substitute>
+    </names>
+  </macro>
+  <macro name="author-short">
+    <names variable="author">
+      <name form="short" and="symbol" et-al-min="3" initialize-with=". "/>
```
rmzelle: @florisvdh, we also require that `et-al-min` and `et-al-use-first` are always both defined together. `et-al-use-first` is missing here.

research-institute-for-nature-and-forest.csl
```diff
@@ -0,0 +1,267 @@
...
+        <text macro="author-short" font-style="normal"/>
+        <text macro="issued"/>
+        <group delimiter=" ">
+          <label font-style="normal" variable="locator" form="short"/>
+          <text variable="locator" form="long" font-style="normal"/>
+        </group>
+      </group>
+    </layout>
+  </citation>
+  <bibliography et-al-min="19" entry-spacing="0" hanging-indent="true">
```
rmzelle: @florisvdh, `et-al-use-first` is missing here too.

rmzelle:
@florisvdh, there are some minor errors in the style. I can help you fix them, but I'd have to know the desired et-al abbreviation rules.

florisvdh:
@rmzelle I have tried to fix this. It seems that the test succeeded.

rmzelle:
Great, thanks!

PR #4237, opened 2019-08-05
Update council-of-science-editors-author-date.csl

POBrien333:
fix via https://forums.zotero.org/discussion/comment/336766/#Comment_336766

Please review. Might be wrong.


rmzelle:
Thanks!

PR #4240, opened 2019-08-07
Update and rename dependent/molecular-psychiatry.csl to molecular-psy…

POBrien333:
…chiatry.csl

via https://forums.zotero.org/discussion/78577/style-request-molecular-psychiatry-currently-outdated


adam3smith:
Thanks! (and sorry for the long wait -- I think we should be able to get back to regular merge speeds now.)

PR #4246, opened 2019-08-09
Update psychological-medicine.csl

POBrien333:
fixes for chapter via https://forums.zotero.org/discussion/comment/337578/#Comment_337578


rmzelle:
Thanks!

PR #4248, opened 2019-08-11
Create organon.csl

sebastien-plutniak:
Creation of a style for papers in English to be published in Organon.


rmzelle:

sebastien-plutniak:
Indeed, one day far in the past I used another CSL style as a model (revue-francaise-de-sociologie.csl). I modified it, and then used the result as a model for the next style I wrote, and so on until today. So, we can add a template link to the initial style, but I am afraid it does not make sense. It is up to you.

rmzelle:
Thanks! Let's still include it.

PR #4249, opened 2019-08-13
Create science-china-chemistry.csl

POBrien333:
via https://forums.zotero.org/discussion/78299/style-for-science-china-chemistry


rmzelle:
Thanks!

PR #4251, opened 2019-08-14
Adapt o-bib.csl to current submission guidelines

denismaier:
o-Bib does not use the CMoS anymore. They've switched to infoclio.ch (Deutschschweiz). See https://www.o-bib.de/about/submissions


rmzelle:
Great, thanks!

PR #4259, opened 2019-08-19
Create seed-science-and-technology.csl

POBrien333:
via https://forums.zotero.org/discussion/78390/style-request-seed-science-and-technology#latest

Question for @adam3smith:
You had made the template style (like many others). I see that you tend to program the space " " in front of every title macro for every item time instead of how I did it here, after the year (2007) suffix=" ". Is there any particular reason for that?
Also, the title usually is included in every item type, so why include it in every single item type and not before the conditional?

Campbell, J.L. and Pedersen, O.K. (2007) The varieties of capitalism and hybrid success Comparative Political Studies, 40, 307–332. https://doi.org/10.1177/0010414006286542


POBrien333:
duplication of #4258.

PR #4264, opened 2019-08-21
Create springer-imis-series-migrationsgesellschaften-english.csl

Schrader311:
This stylesheet is for the ongoing book series called 'Migrationsgesellschaften' at the Institute for Migration Research and Intercultural Studies (IMIS) at the University of Osnabrück, GER and is in line with the requirements of the publisher (Springer). As the book series of both for submissions in English and German, there is a similar .csl, which is for German-speaking authors.


Schrader311:
The mistake which let to the failure is eliminated.

PR #4268, opened 2019-08-21

jjrondeau:
Some adjustements made to disambiguation rules, et-al suffix italicized, and a new macro created to improve the structure.


rmzelle:
Thanks!

Issue #4260, opened 2019-08-19
Mendeley CSL issue citing theses for ACS


tehH1035:
There are a few issues with Mendeley properly citing theses in American Chemical Society (ACS) format. Based on the ACS guidelines [here](https://pubs.acs.org/doi/pdf/10.1021/bk-2006-STYG.ch014), the format for citing theses in the bibliography should be as follows:

> Author. Title of Thesis. Level of Thesis, Degree-Granting University, Location of University, Date of Completion.

An example found in the linked PDF can also be found below here:

![Citing theses ACS format](https://user-images.githubusercontent.com/53947709/63287373-2beffe00-c288-11e9-97ed-b6ec052b264c.png)

Unfortunately, the way Mendeley provides the output, the citation format in the bibliography is as follows:

> Author. Title of Thesis, Level of Thesis, Degree-Granting University, Location of University, Date of Completion.

Note the period after "title of thesis" in the original, but not in Mendeley's format. 

Technically, the information for "level of thesis" is only provided by the "university" and does not make use of the "type" section at all, nor is there a section for the location of the university (see below), and so what I have done is incorporate the type of thesis and location of university all in the "university" section for Mendeley Desktop. (This issue is secondary to the other issue I brought up.)

![Mendeley Desktop format](https://user-images.githubusercontent.com/53947709/63287374-2beffe00-c288-11e9-8a33-11abc74a9465.png)

It seems that this issue is at least somewhat exclusive to Mendeley, as based on the same CSL, there's no issue in the format (see below). However, a minor note is that there is a colon after University of Cambridge and Cambridge (not sure if that's important).

![Editor Citation styles ACS format](https://user-images.githubusercontent.com/53947709/63287372-2beffe00-c288-11e9-9916-c08762e1cee5.png)

I'm quite unfamiliar with coding so if someone could help fix this, I would really appreciate it!

POBrien333:
1. I´ve put in a fix for the ":" between university and university-place.
2. Mendeley maps csl variables slightly different to Zotero. That could be why the type doesn't turn up.

tehH1035:
Thanks so much!

adam3smith:
Yes, for type use "Genre" in Mendeley (you need to add this to the availabile fields for Thesis in the preferences.) We've asked Mendeley to fix this a number of time but for they've not made that change.

tehH1035:
[...]

PR #4270, opened 2019-08-23

POBrien333:
Fix issue with publisher-punctuation for thesis in ACS style as mentioned here: #4260

(No idea how to deal with the other issue(s))


adam3smith:
Thanks!

PR #4274, opened 2019-08-26
Create knowledge-and-management-of-aquatic-ecosystems.csl

POBrien333:
via https://forums.zotero.org/discussion/78455/style-request-knowledge-and-management-of-aquatic-ecosystems


rmzelle:
Thanks!

PR #4275, opened 2019-08-26
Added DOI Field

inkenbrandt:
Changed the field that populates the report type, added a doi at the end of all types, and moved the ending period to the end of all types.


rmzelle:
(superseded by #4276)

PR #4283, opened 2019-08-28
Create australian-archaeology.csl

POBrien333:
via https://forums.zotero.org/discussion/78728/style-request-australian-archaeological-association

### DIFF
--- a/australian-archaeology.csl
+++ b/australian-archaeology.csl
@@ -1,0 +1,227 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB" xmlns="http://purl.org/net/xbiblio/csl">
+  <info>
+    <title>Australian Archaeology</title>
+    <id>http://www.zotero.org/styles/australian-archaeology</id>
+    <link href="http://www.zotero.org/styles/australian-archaeology" rel="self"/>
+    <link href="http://www.zotero.org/styles/ecology" rel="template"/>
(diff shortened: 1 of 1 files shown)


rmzelle:
@POBrien333, https://www.tandfonline.com/action/authorSubmission?journalCode=raaa20&page=instructions#refs suggests the journal uses "Taylor&Francis Reference Style V Harvard B" (which we don't have yet: https://www.zotero.org/styles?q=taylor).

rmzelle:
(if you could take a quick look to see if this Taylor and Francis style guide describes the same format as the documentation link you've been using, it would be better to create Taylor and Francis independent style and a dependent for this journal. Thanks!)

POBrien333:
Good find, but...
- The Taylor Harvard V we have is actually the B version (see that the documentation link is https://www.tandf.co.uk//journals/authors/style/reference/tf_V.pdf, which is the same actually given for the AA journal
- The T&F Harvard V B is different to the style I created based on the guidelines on the AA website and actual publications.

rmzelle:
Thanks!

POBrien333:
Anthropology! I was wondering what the best category would be!

rmzelle:
In this case I went to the Zotero Style Repository, searched for "archaeology", and checked what the most frequently used tag was for the matching styles.

adam3smith:
This is somewhat unique to the US, but "anthropology" here has three branches -- cultural anthropology (which is what we mostly think of under anthro), biological anthropology, and archaeology. They're all very different from each other, which makes of interesting/odd seminars.

PR #4284, opened 2019-08-28
Update bulletin-of-geosciences.csl

mcranium:
Changed "In:" for book chapters to regular font and the parentheses around the issue of articles to be in italics. I used the visual editor. Validated with https://validator.citationstyles.org/


rmzelle:
Thanks!

PR #4285, opened 2019-08-29
Update pnas.csl

POBrien333:
PNAS completely changed style. Please review. :)
Guidelines lack quite a bit of substance, so it's only really adapted for article-journal and chapter (and book extrapolated from that).
https://forums.zotero.org/discussion/78866/pnas-reference-update

### DIFF
--- a/pnas.csl
+++ b/pnas.csl
@@ -113,10 +113,12 @@
             <group delimiter=", " prefix=", ">
               <text variable="title" quotes="true" font-style="italic"/>
               <text macro="editor"/>
-              <text macro="publisher"/>
               <text macro="edition"/>
             </group>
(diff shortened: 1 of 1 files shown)


POBrien333:
Thanks for looking over it.

rmzelle:
> PNAS completely changed style.

No template used?

adam3smith:
> No template used?

Pretty sure this is based on the old style which didn't have a template in there.

rmzelle:
Yeah, looks like it. Thanks both!

(timely, see e.g. https://twitter.com/SamPassmore_/status/1166989872944623623)

POBrien333:
Template was just the old pnas.csl. :)

PR #4287, opened 2019-08-29
Update aids.csl

POBrien333:
via https://forums.zotero.org/discussion/78869/aids-reference-update#latest


adam3smith:
Duplicate of https://github.com/citation-style-language/styles/pull/4288

PR #4288, opened 2019-08-29
Update aids.csl: Journal article title in bold

bwiernik:
Per https://forums.zotero.org/discussion/78869/aids-reference-update#latest


adam3smith:
Thanks!

PR #4292, opened 2019-09-04

cookie365:
Removed "H." for journal articles. This change is necessary, because online magazines often have no booklet number but an article number and so the "H." does not fit for all cases.
The citation guidelines of this University of Applied Sciences require one point after each footnote.


adam3smith:
Thanks!

PR #4296, opened 2019-09-10
Create trends-in-glycoscience-and-glycotechnology.csl

kiyoko:
I'd like to submit a new format for TIGG.


rmzelle:

kiyoko:

I created this from scratch (I think)... I really don't remember now.  I used an online tool to generate the CSL file.

rmzelle:
Okay, that's fine. Thanks!

PR #4298, opened 2019-09-11
Update universitatsmedizin-gottingen.csl

POBrien333:
small fix via https://forums.zotero.org/discussion/79021/universitaetsmedizinn-goettingen


rmzelle:
Danke!

PR #4301, opened 2019-09-13
Create rever-revista-de-estudos-da-religiao.csl

POBrien333:
via https://forums.zotero.org/discussion/76051/rever-style


rmzelle:
Thanks! I dropped the "default-locale" since it's a polyglot journal.

POBrien333:
Thanks for spotting that!

PR #4302, opened 2019-09-13

alex-ball:
Changes:

- Fix punctuation in parenthetical citations: there should be a comma between the author and year.
- If the title is at the head of the reference, the [Online] label is now shown after the genre of a motion picture or the medium of a broadcast.


rmzelle:
Thanks!

PR #4303, opened 2019-09-14
Create helsingin-yliopisto-teologinen-tiedekunta-teologian-tyyli.csl

helmituuli:
A style for the Faculty of Theology at the University of Helsinki. (I'm aware that there's already a style for the faculty, but this style is quite different and some people still need to use the old one, so that is why I'm creating a new style instead of submitting changes.)


helsingin-yliopisto-teologinen-tiedekunta-teologian-tyyli.csl
```diff
@@ -0,0 +1,478 @@
...
+      <key macro="contributors"/>
+      <key macro="issued-year"/>
+      <key variable="title"/>
+    </sort>
+    <layout suffix=".">
+      <choose>
+        <if variable="author editor translator" match="none">
+          <group delimiter=". ">
+            <text macro="title"/>
+            <text macro="issued-year" display="left-margin" prefix="(" suffix=")"/>
```
adam3smith: What's the goal of the `display` element here? It doesn't really make sense nor do I see anything unusual about formatting in the style guide.

helmituuli: You're right, it's completely useless. I guess it was left over from the template style I used. I haven't done anything like this before, so the style overall probably isn't as neatly done as it could be 😀 I'll fix it!

rmzelle:
> A style for the Faculty of Theology at the University of Helsinki. (I'm aware that there's already a style for the faculty, but this style is quite different and some people still need to use the old one, so that is why I'm creating a new style instead of submitting changes.)

@helmituuli, sorry for the very long wait. Could you explain in more detail how the existing CSL style (https://www.zotero.org/styles?q=id%3Auniversity-of-helsinki-faculty-of-theology) differs from this one? I would like to determine what would be the best name for the old and new style (if the two CSL styles are for different versions the same style guide used by the faculty, we could label them by version and/or date).

helmituuli:
@rmzelle As far as I can tell, the main difference is the layout and other details in the bibliography: in the old one the author's name is on a separate line, the year of publication isn't in parentheses, the names of multiple authors are always as "surname, first name", the labels for editor etc. are always in the language of the work was written in. There might be other differences, but those seem to be the most noticeable. 

As for the naming, I'd suggest using just the title "Teologian tyyli" (roughly translated as "Style for theology") to differentiate the new style, since that is only used for the newer version. The old one didn't have a name/title as far as I can tell. 

Also, the old style is not going to be used for much longer, I think, since all new dissertations etc must be done using the new style. Besides, the old style is barely functional: it doesn't even use footnotes even though it should, so anyone using the old one has to do a lot of extra work to get their citations in order. So I'm not sure if it's even useful anymore to keep the old one hanging around, but I wouldn't want to mess up the work of anyone who might still be using the old style for whatever reason.

rmzelle:
@helmituuli, thanks. In that case, it's likely best to just update the existing style. I changed the style title to Finnish, considering the style is fixed to that language.

PR #4304, opened 2019-09-16
Create hakwolfsberg.csl

martinhak:
Changes:
- Citate Websites only with "URL: " and not with "Online verfügbar unter: URL: ")

### DIFF
--- a/hak-wolfsberg-diplomarbeit.csl
+++ b/hak-wolfsberg-diplomarbeit.csl
@@ -1,24 +1,19 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="never" default-locale="de-DE">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="never" default-locale="de-AT">
   <info>
     <title>HAK Wolfsberg Diplomarbeit</title>
-	<id>http://www.zotero.org/styles/hakwolfsberg</id>
-	<link href="http://www.zotero.org/styles/hakwolfsberg" rel="self"/>
+    <id>http://www.zotero.org/styles/hak-wolfsberg-diplomarbeit</id>
+    <link href="http://www.zotero.org/styles/hak-wolfsberg-diplomarbeit" rel="self"/>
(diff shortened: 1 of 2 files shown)


rmzelle:
Thanks! We try to include a "documentation" link in every style (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#10---documentation-link). Does HAK Wolfsberg have a webpage that describes your style format?

martinhak:
Hi! There is no actual documentation from our institution but there is one that we base it on. Can be found here: http://www.diplomarbeiten-bbs.at/sites/default/files/DA-Handreichung_14.4.2016.pdf  ( (page 26)
Our style is based on another one, I already added this to the code. Line shoud be:     <link href="http://www.zotero.org/styles/die-bachelorarbeit-samac-et-al-note" rel="template"/>

Nice regards!

rmzelle:
Ah, sorry, I missed the "template" link.

adam3smith:
Hmm - so this style seems quite different from the BBS style you link to (and the BBS style is quite thinly documented itself). So e.g. you use small caps/Kapitälchen, BBS doesn't. This style has quite specific et al settings, modelled after APA, BBS doesn't mention that. BBS also appears to use ebd. this style doesn't. 
Those sorts of things are the reasons we really want the actual document for the style. It's impossible for me to tell which of these things are done on purpose and which are just left from the Samac style.

martinhak:
We base our style on "Die Bachelorarbeit". There are no differences except citing a website ("Die Bachelorarbeit" says "Online Verfügbar unter: URL", in our style we only use "URL:"). The style "Die Bachelorarbeit" is documented within this book: https://www.amazon.de/Bachelorarbeit-P%C3%A4dagogischen-Hochschule-wissenschaftlicher-Lehrerbildung/dp/3708901886 . As already said there is no real documentation from our institution.

adam3smith:
OK, I guess we can do this, but it seems a bit odd to have a separate style just because of three words before the URL. Is that really needed? Couldn't we just create your style as a [dependent](http://docs.citationstyles.org/en/stable/specification.html#dependent-styles) to the Samac one, which would facilitate maintenance?

adam3smith:
@martinhak just wondering if you saw my question above. More different styles create more work in maintenance, so if possible I'd like to avoid that, but if this distinction is crucial for you style, we'd probably take it.

PR #4309, opened 2019-09-19
Update american-physiological-society.csl (change sorting)

POBrien333:
via https://forums.zotero.org/discussion/79133/style-error-american-physiological-society

set: demote-non-dropping-particle="never"


adam3smith:
Thanks!

PR #4312, opened 2019-09-22
Create soziologiemagazin.csl

nathanlesage:
Add the official Soziologiemagazin styleguide to the Zotero repository to finally enable Zotero et al. users to quickly cite in their articles for our magazine.

I've started from one that looked pretty similar to ours and hope that I didn't make any mistakes; at least the validator passes :)

Thank you very much for having a look at this and hopefully accepting it!


nathanlesage:
Hey, I changed all links back to HTTP, but due to security-concerns I would really appreciate it if the tests would also accept HTTPS-protocols, since all reference URLs also support HTTPS :) Thank you!

rmzelle:
Thanks!

Regarding the HTTP links, I don't have a quick solution, especially for the style ID and "self" link. We'll hopefully update all styles in the repository at some later point.

nathanlesage:
Perfect, thank you very much! :)

PR #4316, opened 2019-09-24
Create one-earth.csl

emblanchard:
Create CSL file specific to the journal One Earth.


adam3smith:
Thanks! We already have the style itself as cell-numeric-superscript, so this would just need to be a [dependent ](http://docs.citationstyles.org/en/stable/specification.html#dependent-styles) style for One Earth (we also need to pint the _Joule_ style to that while we're at it). We can handle this, nothing to do on your end.

adam3smith:
Added via https://github.com/citation-style-language/styles/pull/4317

emblanchard:
Thank you very much!

PR #4320, opened 2019-09-25
Update springer-lecture-notes-in-computer-science.csl to include accessed dates

POBrien333:
via https://forums.zotero.org/discussion/comment/324684#Comment_324684


adam3smith:
Thanks!

PR #4322, opened 2019-09-26
Update and rename immunological-reviews.csl to dependent/immunologica…

POBrien333:
…l-reviews.csl

via https://forums.zotero.org/discussion/comment/340115#Comment_340115
"All references should be numbered consecutively in order of appearance and followed by those references that appear in the figure legends and tables. References should be as complete as possible. In text citations should cite references in consecutive order using Arabic superscript numerals. For more information about AMA reference style please see the AMA Manual of Style."


adam3smith:
Thanks!

Issue #4138, opened 2019-06-11
PLOS style references include "[Internet]" for preprints and books


dhimmel:
_PLOS Computational Biology_ [specifies](https://journals.plos.org/ploscompbiol/s/submission-guidelines#loc-references) including `[Internet]` after the title for specific types of references:

+ New media (blogs, web sites, or other written works)
+ Databases and repositories (Figshare, arXiv)

For other types, examples do not include `[Internet]`, such as:

+ Accepted, unpublished articles
+ Online articles	
+ Books
+ Book chapters
+ Deposited articles (preprints, e-prints, or arXiv)
+ Published media (print or online newspapers and magazine articles)	
+ Masters' theses or doctoral dissertations
+ Multimedia (videos, movies, or TV shows)

Note that arXiv is listed under both "Databases and repositories" and "Deposited articles", although I think the later is a more accurate categorization for preprints.

Currently the CSL Style for _PLOS Computational Biology_, which inherits from PLOS, places `[Internet]` after many reference types that fall under the later categories above. It seems that the presence of a URL in the CSL JSON is sometimes used to determine whether `[Internet]` should be added:

https://github.com/citation-style-language/styles/blob/76e2ff8e518bd8623b48e31e1c70ff4ffb66217b/plos.csl#L163-L170

Note that most citeable works these days have a URL, but that doesn't mean they fall into the internet categories above.

We noticed this issue when creating [this reference list](https://greenelab.github.io/meta-review/v/040638d25b9a45ad3d060afc9689a17f4c3a3045/#references). Notice the following examples of how CSL JSON was converted to a reference.

[csl json](https://github.com/greenelab/meta-review/blob/040638d25b9a45ad3d060afc9689a17f4c3a3045/content/manual-references-2019-06-11.json#L23125-L23149): `type` is `entry`, but switching to `book` does not remove `[Internet]`:

> 1. Nielsen M. Reinventing Discovery [Internet]. Princeton University Press; 2011. doi:10.1515/9781400839452

[csl json](https://github.com/greenelab/meta-review/blob/040638d25b9a45ad3d060afc9689a17f4c3a3045/content/manual-references-2019-06-11.json#L1129-L1294). Type is `manuscript` for a bioRxiv preprint:

[...]

PR #4323, opened 2019-09-27
Update plos.csl (issues re. [internet/preprint])

POBrien333:
Fix issues regarding [internet] and [preprint] for PLoS as per: #4138 (and pointed out again here: https://github.com/citation-style-language/styles/issues/4140#issuecomment-535641750)

Please check carefully if I did this correctly. :)

@dhimmel, can you give this a test ride?

### DIFF
--- a/plos.csl
+++ b/plos.csl
@@ -16,7 +16,7 @@
     <category citation-format="numeric"/>
     <category field="science"/>
     <summary>Vancouver for PLoS Journals as of January 2015</summary>
-    <updated>2019-09-29T10:42:52+00:00</updated>
+    <updated>2019-09-29T19:19:07+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
@@ -117,14 +117,7 @@
         <group suffix="." delimiter=" ">
           <choose>
             <if type="article-journal report" match="any">
-              <group delimiter=" ">
-                <text variable="container-title" form="short" strip-periods="true"/>
-                <choose>
-                  <if match="none" variable="volume page">
-                    <text value="Preprint" prefix="[" suffix="]"/>
-                  </if>
-                </choose>
-              </group>
+              <text variable="container-title" form="short" strip-periods="true"/>
             </if>
             <else-if type="post-weblog webpage post" match="any">
               <group delimiter=" ">
--- a/plos.csl
+++ b/plos.csl
@@ -16,7 +16,7 @@
     <category citation-format="numeric"/>
     <category field="science"/>
     <summary>Vancouver for PLoS Journals as of January 2015</summary>
-    <updated>2019-09-27T08:33:01+00:00</updated>
+    <updated>2019-09-29T10:42:52+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
@@ -201,7 +201,7 @@
       <if type="article-journal" match="any">
         <choose>
           <if match="none" variable="volume page">
-            <group delimiter=" " suffix=". ">
+            <group delimiter=" ">
               <date date-parts="year" form="text" variable="issued"/>
               <text macro="accessed-date"/>
             </group>
--- a/plos.csl
+++ b/plos.csl
@@ -10,10 +10,13 @@
     <author>
       <name>Sebastian Karcher</name>
     </author>
+    <contributor>
+      <name>Patrick O'Brien</name>
+    </contributor>
     <category citation-format="numeric"/>
     <category field="science"/>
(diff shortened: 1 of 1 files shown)


dhimmel:
In https://github.com/greenelab/meta-review/pull/261, I evaluate the effect of the new PLOS style from https://github.com/citation-style-language/styles/pull/4323/commits/82df7b538c5471ab9806efadc0b2ecf131d24293 on our reference formatting. The result is available in [`manuscript-1.0.18-d0462c5.pdf`](https://github.com/citation-style-language/styles/files/3664081/manuscript-1.0.18-d0462c5.pdf).

I can confirm that `[Internet]` no longer appears for references 1, 10, 15, 25 as documented in https://github.com/citation-style-language/styles/issues/4138#issue-454851781.

`[Preprint]` does not appear for arXiv references, but also appears for references 23, 35, 45, 60, 69 which are not preprints (some are blog posts, but some are journal articles). Furthermore, preprint does not appear for bioRxiv references. Our [CSL JSON](https://github.com/greenelab/meta-review/blob/output/references.json) is located here, if that helps with debugging. I wonder if the method for detecting preprints is too imprecise? Given the current CSL JSON schema, perhaps it is best to not assume anything is a preprint (or perhaps to hardcode a list of preprint servers).

POBrien333:
Re [Preprint]:
The conditional tests for article-journal and reports.
It then only prints [preprint] if the item does NOT have a volume and a page.

Item 23 and 35 are blog articles and should NOT be article-journals.
The others a normal papers from journals and as far as I see it does that correctly, print [Preprint] if if there is no volume and or page assigned, meaning it is in preprint.

dhimmel:
> The others a normal papers from journals and as far as I see it does that correctly, print [Preprint] if if there is no volume and or page assigned, meaning it is in preprint.

Unfortunately the meaning of "preprint" differs from the obvious interpretation of an article prior to print publication. Instead in the life sciences, the term is [defined as such](https://asapbio.org/preprint-info/preprint-faq#qe-faq-637):

> A preprint is a complete scientific manuscript that is uploaded by the authors to a public server.  The preprint contains complete data and methodologies; it is often the same manuscript being submitted to a journal (see FAQ on submitting preprints).  After a brief quality-control inspection to ensure that the work is scientific in nature, the author’s manuscript is posted within a day or so on the Web without peer review and can be viewed without charge by anyone in the world.

The key distinctions is that preprints are posted at the author's discretion, independent of approval from peer review. They are usually hosted on preprint servers that are distinct from the journals, although some journals like _F1000 Research_ also make submitted articles available prior to peer review and acceptance and these could be called preprints (but this is a corner case).

The [PLOS guidelines](https://journals.plos.org/plosone/s/submission-guidelines) add `[Preprint]` to references under the category "Deposited articles (preprints, e-prints, or arXiv)`. 

> The conditional tests for article-journal and reports. It then only prints [preprint] if the item does NOT have a volume and a page.

I'd suggest never labeling an `article-journal` with `[Preprint]`. Labeling reports without a volume and page as a `[Preprint]` would probably be right more than not... although ultimately it's a tough workaround to CSL JSON's lack of an `article-preprint` type.

> Item 23 and 35 are blog articles and should NOT be article-journals.

Good catch. Will look into fixing my metadata.

POBrien333:
I see. In that case that conditional doesn't make sense. I've removed it again.
I don't think there is any other way we could test for with CSL.

adam3smith:
We have discussed using `article` for preprints and I'm still open for that, but it's obviously not in the specs. The alternative would be to just wait for CSL 1.1/1.2 and introduce a preprint item type.

dhimmel:
I rebuilt the manuscript with the style from https://github.com/citation-style-language/styles/pull/4323/commits/dfe2507a7851304fc8d3c08e9a1824045a7d5f49: 
[manuscript-1.0.19-468ace4.pdf](https://github.com/citation-style-language/styles/files/3667997/manuscript-1.0.19-468ace4.pdf).

The PR looks good on my end! An improvement for sure. Thanks @POBrien333.

adam3smith:
Thanks both!

bwiernik:
@adam3smith From my perspective, the spec doesn’t really clearly define the meaning of any item type, so conventions about type usage are important to consider.

The current typical conventional for preprints is to use `article-journal` or `report`. I think we agree that doesn’t make much sense. `article` is essentially unused, and I agree with your earlier comments that it’s a good conceptual fit. In CSL 1.2/2, the name could potentially be changed to `preprint`, but it seems like a waste to not move toward a convention of using it for this purpose now. That’s the assumption being used with APA, and `article` isn’t otherwise being printed at all with this style.

dhimmel:
> From my perspective, the spec doesn’t really clearly define the meaning of any item type, so conventions about type usage are important to consider.

Agreed although this should be fixed by updating the schema with definitions.

>  seems like a waste to not move toward a convention of using it for this purpose now.

I'm happy to use the `article` type for preprints in CSL JSON that I produce. However, users receive CSL JSON from many sources, and I don't think it's practical to expect all users of a given style will update CSL JSON they receive to use `article` for preprints. In other words, it's probably dangerous for a style to assume `article` means preprint.

PR #4329, opened 2019-10-02
Create journal-of-advertising-research.csl

POBrien333:
via https://forums.zotero.org/discussion/74531/style-request-journal-of-advertising-research


rmzelle:
Thanks!

PR #4330, opened 2019-10-02
Update nature.csl

POBrien333:
via https://forums.zotero.org/discussion/79336/style-error-nature-and-possibly-subjournals
Just submitting this. Your decision to take it or not :)

### DIFF
--- a/nature.csl
+++ b/nature.csl
@@ -15,7 +15,7 @@
     <category field="generic-base"/>
     <issn>0028-0836</issn>
     <eissn>1476-4687</eissn>
-    <updated>2019-10-02T13:55:54+00:00</updated>
+    <updated>2019-10-08T13:18:12+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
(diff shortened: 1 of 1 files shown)


adam3smith:
(I do want to take this, but unsure of the format. The semicolon seems odd, but is in the published articles. The Lake Forrest version also seems a bit off to me. I'll want to explore some more samples I think.)

adam3smith:
Looked around a bit e.g. at https://www.nature.com/articles/s41586-019-1619-z reference 42 and I think the format suggested by the LFC guide, i.e. title. URL (year) is the way to go. I'd be inclined to just change website to the same format. I didn't see an access date in any of the articles I looked at (LFC suggests using it for frequently changing sites)

nature.csl
```diff
@@ -111,9 +120,9 @@
             <text term="in"/>
           </if>
         </choose>
-        <text macro="container-title"/>
+        <text macro="container-title" suffix="."/>
```
adam3smith: I'm not seeing that for journals. Where is the period coming from?

POBrien333: I put this in for something, but can't remember now.
The way of doing this is wrong anyway. Not every container title has dots at the end (that's what I was thinking).

PR #4332, opened 2019-10-02
APA.csl - Change journal title case to capitalize-first

bondjimbond:
With the Container journal title case being "title", each word in journal titles (including conjunctions like And and Of) is capitalized. When your journal metadata is correct and it gets transformed on you, that can be quite vexing.

This pull changes the case to "capitalize-first" which better respects the existing value of the field.

### DIFF
--- a/apa.csl
+++ b/apa.csl
@@ -1347,7 +1347,7 @@
   <macro name="container-title">
     <choose>
       <if type="article article-journal article-magazine article-newspaper dataset" match="any">
-        <text variable="container-title" font-style="italic" text-case="title"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
title-case capitalization [should not be applied to](http://docs.citationstyles.org/en/stable/specification.html#title-case-conversion) most prepositions and conjunctions. What are you using to generate your bibliography?

(Title casing is used for container titles so that they are correct with import from, e.g., Pubmed, which has journal title in sentence case. )

bondjimbond:
Yeah, I'm finding that the problem is actually the module that's reading this, rather than the CSL file.

PR #4333, opened 2019-10-04
Update according to the new 2019 editorial standards

Clarisse-Lachat:
Update of the CSL style for the "Maison de l'Orient et de la Méditerranée" according to the new 2019 editorial standards.


rmzelle:
Thanks!

I removed the "-fr" from the CSL style, as we try to minimize changes in style file names (and IDs). We also typically don't include versions in the style title (like "2019") unless we have multiple versions of the same style in our repository.

PR #4334, opened 2019-10-04
Create maison-de-l-orient-et-de-la-mediterranee-2019-EN.csl

Clarisse-Lachat:
English version of the CSL style for the "Maison de l'Orient et de la Méditerranée" for articles or books written in English.


rmzelle:
Thanks!

PR #4335, opened 2019-10-06
Create les-nouvelles-de-l-archeologie.csl

sebastien-plutniak:
Creation of the style (no template).

### DIFF
--- a/les-nouvelles-de-l-archeologie.csl
+++ b/les-nouvelles-de-l-archeologie.csl
@@ -54,12 +54,8 @@
       <name and="symbol" name-as-sort-order="all" font-variant="small-caps" sort-separator=" " delimiter=", " delimiter-precedes-last="never" initialize-with="."/>
       <label form="short" prefix=" "/>
       <substitute>
-        <names variable="editor">
-          <name and="symbol" name-as-sort-order="all" font-variant="small-caps" sort-separator=" " delimiter=", " delimiter-precedes-last="never" initialize-with="."/>
-          <substitute>
-            <names variable="editorial-director" font-variant="small-caps"/>
-          </substitute>
-        </names>
+          <names variable="editor" />
+          <names variable="editorial-director"/>
       </substitute>
     </names>
   </macro>
--- a/les-nouvelles-de-l-archeologie.csl
+++ b/les-nouvelles-de-l-archeologie.csl
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" default-locale="fr-FR" page-range-format="expanded">
   <info>
-    <title>Les nouvelles de l'archéologie</title>
+    <title>Les nouvelles de l'archéologie (French)</title>
     <title-short>NDA</title-short>
     <id>http://www.zotero.org/styles/les-nouvelles-de-l-archeologie</id>
     <link href="http://www.zotero.org/styles/les-nouvelles-de-l-archeologie" rel="self"/>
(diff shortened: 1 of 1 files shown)


rmzelle:

sebastien-plutniak:
Let say I re-used and adapted the file I used to create the prehistoires-mediterraneennes.csl or the organon.csl files… but I am not sure I means I used them as a proper template.

les-nouvelles-de-l-archeologie.csl
```diff
@@ -0,0 +1,190 @@
...
+    <terms>
+      <term name="editor" form="short">
+        <single>(éd.)</single>
+        <multiple>(éds.)</multiple>
+      </term>
+      <term name="page" form="short">
+        <single/>
+        <multiple/>
+      </term>
+      <term name="no date">À paraître</term>
```
adam3smith: I don't think I like this. Forthcoming and no date are two very different concepts.

les-nouvelles-de-l-archeologie.csl
```diff
@@ -0,0 +1,190 @@
...
+  <macro name="journal-issue">
+    <choose>
+      <if variable="volume issue" match="all">
+        <text variable="volume" suffix=", "/>
+        <text variable="issue"/>
+      </if>
+      <else-if variable="volume">
+        <text variable="volume"/>
+      </else-if>
+    </choose>
```
adam3smith:
```suggestion
 <group delimiter=", ">
   <text variable="volume"/>
   <text variable="issue"/>
  </group>
```

adam3smith: this does basically the same

les-nouvelles-de-l-archeologie.csl
```diff
@@ -0,0 +1,190 @@
...
+    <names variable="author">
+      <name and="symbol" name-as-sort-order="all" font-variant="small-caps" sort-separator=" " delimiter=", " delimiter-precedes-last="never" initialize-with="."/>
+      <label form="short" prefix=" "/>
+      <substitute>
+        <names variable="editor">
+          <name and="symbol" name-as-sort-order="all" font-variant="small-caps" sort-separator=" " delimiter=", " delimiter-precedes-last="never" initialize-with="."/>
+          <substitute>
+            <names variable="editorial-director" font-variant="small-caps"/>
+          </substitute>
+        </names>
```
adam3smith:
```suggestion
        <names variable="editor"/>
        <names variable="editorial-director"/>
```

should do the same

adam3smith: actually it'll give you a label such as "eds." for editors, which I think you want.

adam3smith:
Is this a duplicate of https://github.com/citation-style-language/styles/pull/4293 ?

les-nouvelles-de-l-archeologie.csl
sebastien-plutniak: Ok, I changed it in the new version

les-nouvelles-de-l-archeologie.csl
sebastien-plutniak: Sorry, I don't understand this point (and haven't change these lines in the new version)

les-nouvelles-de-l-archeologie.csl
sebastien-plutniak: Ok, i changed it in the new version.

sebastien-plutniak:
Yes, indeed, it is a duplicate (I've forgot I submitted it early September) I close the pull request [#4293](https://github.com/citation-style-language/styles/pull/4293).

les-nouvelles-de-l-archeologie.csl
adam3smith: Why are you using this complicated construct with two nested substitutes rather than just
```
<substitute>
  <names variable="editor"/>
  <names variable="editorial-director"/>
</substitute>
```

which isn't just cleaner, it should also produce more accurate output. If you check the current generated samples, the edited volume (Hancke et al.) doesn't have an editor label for editors, which is almost certainly incorrect.

sebastien-plutniak: Thanks, now I got it. I modified the file as you suggest.

adam3smith:
Great, thanks!

PR #4336, opened 2019-10-07
Create european-union-interinstitutional-style-guide.csl

POBrien333:
I've whipped this together as I saw we have the note style, but not this one.

Kind of via here: https://forums.zotero.org/discussion/79389/eu-interinstitutional-style-guide#latest

### DIFF
--- a/europa-interinstitutional-style-guide.csl
+++ b/europa-interinstitutional-style-guide.csl
@@ -180,32 +180,6 @@
       </else>
     </choose>
   </macro>
-  <macro name="title-short">
-    <choose>
-      <if variable="title" match="none">
-        <choose>
-          <if type="interview">
-            <text term="interview"/>
-          </if>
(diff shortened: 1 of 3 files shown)


adam3smith:
In terms of naming, this should be `european-union-interinstitutional-style-guide-author-date` and the existing one should be `...-note`, right? "Europa" is just part of the URL?

rmzelle:
> In terms of naming, this should be `european-union-interinstitutional-style-guide-author-date` and the existing one should be `...-note`, right? "Europa" is just part of the URL?

Let's keep the style ID of the note version as is. I did change the title of the note version.

And let's keep "European Union" as a title prefix, as "Interinstitutional Style Guide" is a bit nondescript (I did add a hyphen to separate the institution from the style guide title).

Should be good to go from my side.

rmzelle:
Thanks!

PR #4337, opened 2019-10-07
Reference disambiguation in JASA style

floriandenk:
Bug fix: References to several papers from the same author and one year get disambigued (Author 2018a, 2018b) in citations, but not in the bibliography (a, b missing there). Fixed this issue by changing the appropriate macros in the bibliography generator.

### DIFF
--- a/the-journal-of-the-acoustical-society-of-america.csl
+++ b/the-journal-of-the-acoustical-society-of-america.csl
@@ -142,7 +142,7 @@
     <choose>
       <if type="report thesis" match="any">
         <text variable="title" font-style="italic"/>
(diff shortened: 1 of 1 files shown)


POBrien333:
@floriandenk 
There is a way easier solution that also doesn't break the style.
all you need to do is add <text variable="year-suffix" to the issued macro.
So reverse those changes you made and add that.

adam3smith:
(looking at the style guide, switching to the `issued-year` macro actually does seem like the better choice. No sign of full dates anywhere)

adam3smith:
Thanks!

PR #4338, opened 2019-10-08
Update elsevier-with-titles.csl

POBrien333:
via https://forums.zotero.org/discussion/79406/why-do-first-names-come-first-before-last-name-in-refense-list-fot-elsevier-style-with-numerics
Change to  name-as-sort-order="all" for authors/editors.

Confirmed in these papers: 
Journal of Hazardous Materials
   https://www.sciencedirect.com/science/article/pii/S0304389419309756
   https://www.sciencedirect.com/science/article/pii/S0304389419312373

However, I have checked a few other journals "Acta Austronautica", "Artificial Intelligence" and "Chemico-Biological Interactions" and none seem to have the new style. This PR will need some more research.

### DIFF
--- a/elsevier-with-titles.csl
+++ b/elsevier-with-titles.csl
@@ -18,12 +18,12 @@
     <category citation-format="numeric"/>
     <category field="generic-base"/>
     <summary>A style for many of Elsevier's journals that includes article titles in the reference list</summary>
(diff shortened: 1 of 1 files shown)


adam3smith:
No, I think the Journal of Hazardous Materials has just switched to Elsevier Harvard. The numeric style is right as is.

PR #4340, opened 2019-10-08
update apa style: no publisher place for book, only publisher name

nics:
https://apastyle.apa.org/style-grammar-guidelines/references/examples#whole
states that a whole book should have no publisher place

### DIFF
--- a/apa.csl
+++ b/apa.csl
@@ -834,6 +834,12 @@
           <text macro="archive"/>
         </group>
       </else-if>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks! That's for the 7th edition of APA, though (out for just a week) -- we're planning on adding support for it in the next weeks, but it'll be easier to do in one swoop rather than piecemeal.

PR #4344, opened 2019-10-10
Create iso690-author-date-de.csl

ApolloLV:
Starting translation to German

### DIFF
--- a/iso690-author-date-de.csl
+++ b/iso690-author-date-de.csl
@@ -241,18 +241,18 @@
   </macro>
   <macro name="issue">
     <group delimiter=", ">
-      <text variable="volume" prefix="Vol.&#160;"/>
+      <text variable="volume" prefix="Bd.&#160;"/>
       <choose>
         <if variable="volume">
-          <text variable="issue" prefix="no.&#160;"/>
-          <text variable="page" prefix="p.&#160;"/>
+          <text variable="issue" prefix="Nr.&#160;"/>
+          <text variable="page" prefix="S.&#160;"/>
         </if>
         <else-if variable="issue">
-          <text variable="issue" prefix="No.&#160;"/>
-          <text variable="page" prefix="p.&#160;"/>
+          <text variable="issue" prefix="Nr.&#160;"/>
+          <text variable="page" prefix="S.&#160;"/>
         </else-if>
         <else>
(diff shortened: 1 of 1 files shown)


ApolloLV:
I just tried this style on a thesis with Zotero, it looks fine to me.
I will create a new merge request if I stumble upon any remaining issues, so feel free to merge this already.

adam3smith:
I thought the ISO690-de norm wanted small-caps/Kapitälchen for last names, not uppercase? Also, did you start the style from scratch, or did you base this on an existing one?

ApolloLV:
I basically translated the existing ISO690-en.

ApolloLV:
The ISO690-de norm wants uppercase, not small-caps, as far as I can tell:
examples_iso690

adam3smith:
A couple of questions:
1. My reading of ISO 690 (and the German DIN ISO interpretations I've found) suggest that names should only by lastname, firstname for the first author, then firstname lastname, i.e. 

> HANCKÉ, Bob, Martin RHODES und Mark THATCHER

2. What's the deal with abstracts in ISO? I don't think they're normal part of a bibliography and people would expect them. Can we remove them?

3. I would expect editors to be included for chapters in edited volumes (MARES in the provided examples). Is that not the case?

Thanks again for your work on this!

ApolloLV:
> A couple of questions:
> 
>     1. My reading of ISO 690 (and the German DIN ISO interpretations I've found) suggest that names should only by lastname, firstname for the first author, then firstname lastname, i.e.
> 
Yes, the first author is to be given as lastname, firstname (5.2.1).
If the current style is incorrect in this regard, we may need to fix the version in other languages
> 
> > HANCKÉ, Bob, Martin RHODES und Mark THATCHER
> 
>     2. What's the deal with abstracts in ISO? I don't think they're normal part of a bibliography and people would expect them. Can we remove them?
As far as I can see, the standard does not even mention abstracts. They also don't show up in any of the examples.
> 
>     3. I would expect editors to be included for chapters in edited volumes (MARES in the provided examples). Is that not the case?
iso690
Sorry, I didn't check. They should be included.

> 
> 
> Thanks again for your work on this!

adam3smith:
great, thanks! Let us know if you find any issues with this.

PR #4345, opened 2019-10-10

ibadger:
Our Harvard guide has been updated over the summer, the amendments are to match that.
-change in how accessed dates are introduced and a full stop added after the brackets
-all dates expressed with / as a delimiter 
-minor amendments to archive, map, video, broadcast to match guide
-fixed a few minor errors (book chapters didn't display the edition)

### DIFF
--- a/harvard-university-for-the-creative-arts.csl
+++ b/harvard-university-for-the-creative-arts.csl
@@ -30,7 +30,7 @@
       </if>
       <else>
         <names variable="author">
-          <name font-style="normal" and="text" delimiter-precedes-last="never" et-al-min="3" et-al-use-first="1" initialize-with=". " name-as-sort-order="all"/>
+          <name and="text" delimiter-precedes-last="never" et-al-min="3" et-al-use-first="1" initialize-with=". " name-as-sort-order="all"/>
           <label form="short" prefix=" "/>
           <et-al font-style="italic"/>
           <substitute>
@@ -241,7 +241,7 @@
           <text term="in" text-case="capitalize-first" suffix=": "/>
           <text macro="editor"/>
           <group delimiter=", ">
-            <group font-style="normal" font-variant="normal" delimiter=" ">
+            <group delimiter=" ">
               <text variable="container-title" font-style="italic" suffix="."/>
               <group delimiter=" " suffix=".">
                 <text term="volume" form="short" text-case="capitalize-first"/>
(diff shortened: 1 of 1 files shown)


harvard-university-for-the-creative-arts.csl
```diff
@@ -426,9 +415,10 @@
       </if>
       <else>
         <names variable="author">
-          <name form="short" font-style="normal" and="text" delimiter-precedes-last="never" initialize-with="."/>
+          <name form="short" and="text" delimiter-precedes-last="never" initialize-with="."/>
           <et-al font-style="italic"/>
           <substitute>
+            <names variable="editor"/>
```
adam3smith: please check that this is right -- but it seems implausible to not list editors of edited books in in-text citations (see the Hancke/Beyond Varieties of Capitalism example)

harvard-university-for-the-creative-arts.csl
```diff
@@ -290,6 +276,7 @@
           <et-al font-style="italic"/>
         </names>
         <text variable="medium" prefix=" [" suffix="] "/>
+        <text variable="genre" prefix=" [" suffix="]"/>
```
adam3smith: The subsequent square brackets seem wrong -- shouldn't those be in a group (or have a test for one or the other between medium and genre)?

adam3smith:
Two questions above -- please check

harvard-university-for-the-creative-arts.csl
ibadger: That is probably my mistake. Usually you'd reference a chapter, so if referencing a edited book as a whole and not a chapter then yes I would like the editor to appear. Can you offer any guidance on what I need to change?

harvard-university-for-the-creative-arts.csl
ibadger: Looking at my test references I don't think that I need genre, there as for all sources that use Motion Picture the medium will be used to describe what the source is. I can take genre out, will that fix the problem?

harvard-university-for-the-creative-arts.csl
adam3smith: I did change that already, so nothing; I was just asking you to confirm.

adam3smith:
Great, thanks!

ibadger:
Hi

Sorry to be a pain but I hadn't noticed in the example refrrences on the thread from the newest version of the style - when eds. appears it isn't in brackets, see pasted reference below. 

It should be in brackets to match how it appears in the book chapter example edited by Hall further down the examples.


Hancké, B. et al. eds. (2007) ...

Should I amend the code and raise another pull request?

Thanks for all your help

Ian

PR #4354, opened 2019-10-11
Create weed-research.csl

POBrien333:
via https://forums.zotero.org/discussion/79431/add-new-citation-style-from-endnotes-to-zotero#latest


adam3smith:
Thanks!

PR #4359, opened 2019-10-15
Update elsevier-x.csl styles with https://doi.org/ format

### LINKED ISSUE #4341: Update DOI to correct link format for Elsevier
It looks like they've updated all their styleguides to use https:/doi.org/ instead of doi:

POBrien333:
closes #4341
All elsevier styles were checked. Half of them either already had the new format or do not use DOI.


adam3smith:
Great, thanks!

PR #4363, opened 2019-10-16
Create mbts-fullnote-bibliography.csl

gq9000:
Hello, I have created a new style for use at Midwestern Baptist Theological Seminary (https://www.mbts.edu/). It is based on Chicago. Thank you!

### DIFF
--- a/mbts-fullnote-bibliography.csl
+++ b/mbts-fullnote-bibliography.csl
@@ -1,0 +1,1515 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="display-and-sort" page-range-format="chicago">
+  <info>
+    <title>Midwestern Baptist Theological Seminary (full note, based on Chicago)</title>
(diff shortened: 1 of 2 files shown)


adam3smith:
The style guide you link to is for blogs and doesn't say anything about references -- is the citation style described somewhere you could link to? How is it different from Chicago Manual?

gq9000:
Hello,
thanks for your message. You're right, that's the wrong link. I just
checked and it looks like this style guide is not posted online. I just
have the PDF from the school. It's very similar to Chicago but there's some
minor differences for several types such as journal articles, encyclopedia
and dictionary entries, etc. Mostly punctuation and placement of some
locators. If you like, I can send you the PDF.

Sincerely,
J Brown

adam3smith:
Could you upload the PDF here so we have some point of reference? It's not ideal to have a style up where we can't access documentation in case someone reports issues.

gq9000:
Okay, I uploaded the file here:
https://github.com/gq9000/styles/blob/master/The-Midwestern-Style-Manual-Second-Edition-Revised-110718.pdf

Thanks.

rmzelle:
Thanks!

PR #4370, opened 2019-10-19
Create Style JSBM

spaetz:
This is the Style of the Journal of Small Business Management, as described here:
https://onlinelibrary.wiley.com/page/journal/1540627x/homepage/forauthors.html

Starting point was APA, no ampersand.
The style validates and seems to work well for me.


spaetz:
Will provide updated pull request soon.

PR #4372, opened 2019-10-21
Update developing-world-bioethics.csl and bioethics.csl

POBrien333:
Fix issues with DWB and make bioethics.csl a dependent.
via https://forums.zotero.org/discussion/79619/style-request-the-journal-bioethics


rmzelle:
Thanks!

PR #4373, opened 2019-10-21
Update and rename information-systems-journal.csl to dependent/inform…

POBrien333:
…ation-systems-journal.csl

Switch to a dependent of APA as mentioned in guidelines.
Via https://forums.zotero.org/discussion/79600/style-modification-information-sytems-journal


rmzelle:
Thanks!

PR #4374, opened 2019-10-21
Create institut-catholique-de-paris.csl

ddanangaw:


POBrien333:
Please have a look at the "Style Requirements" here.
Which style did you base this on? Please add that as a template.
Can you also link to the documentation for the style?

rmzelle:
@ddanangaw, thanks for your patience. Is the style guide for the Institut Catholique de Paris publicly available online? We try to always include a link to the style guide in every style.

ddanangaw:
> @ddanangaw, thanks for your patience. Is the style guide for the Institut Catholique de Paris publicly available online? We try to always include a link to the style guide in every style.

@rmzelle, no, it has not been publicly available online.

rmzelle:
@ddanangaw, in that case, can you provide us with the exact title and version of this style guide? Any other identifying information (like ISBN or other catalog information) would be very useful too.

And is the style guide used across the entire Institut Catholique de Paris, or only within certain departments?

ddanangaw:
@rmzelle, Sorry, i gave an inaccurate information. It turns out that this style has been publicly available online. Here is the address: https://icp.hypotheses.org/files/2018/07/Normes-%C3%A9ditoriales-Transversalit%C3%A9s.pdf

This style is commonly used at Institut Catholique de Paris, and specifically it become a writing guide in this university's scientific journal: Transversalités (https://icp.hypotheses.org/la-revue-transversalites).

rmzelle:
Merci! Since the style guide is for the journal, I took the liberty to rename the CSL style accordingly. We can create an additional "dependent" CSL style for the ICP if you think that would be useful (as explained at http://docs.citationstyles.org/en/stable/primer.html#independent-and-dependent-styles, such a dependent style would rely on another style for its citation format definition (the Transversalités CSL style in this case) and merely act as an alias.

ddanangaw:
Ok. Thank you @rmzelle.

PR #4375, opened 2019-10-22
Create transposition.csl

svictor9:
French style for the journal Transposition. French is the main language of the journal.


rmzelle:
Hi, thanks for your patience. I compared your French and English CSL styles for this journal, and they were similar enough that we should be able to cover both languages with a single style. So I made some changes for this and removed the fixed locale. Let us know if it doesn't work as expected.

Closes https://github.com/citation-style-language/styles/pull/4376.

PR #4378, opened 2019-10-22
Create gazeta-medica.csl

POBrien333:
via https://forums.zotero.org/discussion/79653/style-request-vancouver-superscript-with-doi

### DIFF
--- a/gazeta-medica.csl
+++ b/gazeta-medica.csl
@@ -52,7 +52,7 @@
     <choose>
       <if variable="issued">
         <date variable="issued">
-          <date-part name="year" form="long" range-delimiter="–"/>
+          <date-part name="year" form="long" range-delimiter="&#8211;"/>
         </date>
       </if>
       <else>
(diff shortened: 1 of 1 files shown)


gazeta-medica.csl
```diff
@@ -0,0 +1,196 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" page-range-format="minimal" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
```
rmzelle: @POBrien333, maybe we should remove the default-locale, even if you didn't test the style in both languages?

http://www.gazetamedica.pt/index.php/gazeta/about/submissions:

> LANGUAGE
> 
> The articles can be submitted in Portuguese or English.

POBrien333: Good spot!

rmzelle:
Thanks!

PR #4381, opened 2019-10-28
Create nauplius.csl

POBrien333:
via https://forums.zotero.org/discussion/79607/style-request-nauplius

### DIFF
--- a/nauplius.csl
+++ b/nauplius.csl
@@ -1,0 +1,489 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>Nauplius</title>
+    <id>http://www.zotero.org/styles/nauplius</id>
+    <link href="http://www.zotero.org/styles/nauplius" rel="self"/>
+    <link href="http://www.zotero.org/styles/water-alternatives" rel="template"/>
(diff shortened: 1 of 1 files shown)


nauplius.csl
```diff
@@ -0,0 +1,489 @@
...
+                <text variable="collection-title"/>
+              </else>
+            </choose>
+          </group>
+        </group>
+      </if>
+      <else-if type="chapter" match="any">
+        <group delimiter=", ">
+          <choose>
+            <if variable="ISBN ISBN" match="any">
```
adam3smith: I don't understand the test here -- don't see why a book chapter with an ISBN would be cited differently?

nauplius.csl
```diff
@@ -0,0 +1,489 @@
...
+            <date-part name="year" range-delimiter="/"/>
+          </date>
+        </group>
+      </else-if>
+    </choose>
+  </macro>
+  <macro name="editor">
+    <names variable="editor translator">
+      <name and="text" initialize-with="."/>
+      <label form="short" strip-periods="true" prefix=" (" suffix=")"/>
```
adam3smith: You'll probably want the strip-periods on all names labels then, no?

adam3smith:
Two quick comments -- this again looks a bit like a style where the template chose perhaps wasn't ideal, but might still be faster to just fix some of the weirdness in the style. Thanks!

POBrien333:
Yup, way too complex a style for what it does, and cumbersome at that.
Thanks for spotting the issues.

adam3smith:
Great, thanks!

PR #4384, opened 2019-10-29
Update cell-numeric-superscript.csl

emblanchard:
This update removes URLs from all non-webpage references, "Available at:" from all other URLs, access dates for all URLs, and publisher locations in accordance with Cell Press style.


emblanchard:
I'm noticing a few errors in citation sample 4 above: "et al." should not be italicized, and the URL should not be included because this is a journal article. How do we go about fixing those?

adam3smith:
Thanks for catching those -- that's what the previews are for. Should be all good now.

emblanchard:
Excellent, thank you!!

PR #4385, opened 2019-10-30
Update taylor-and-francis-council-of-science-editors-author-date.csl

POBrien333:
https://forums.zotero.org/discussion/79806/style-request-geografiska-annaler-series-a-physical-geography#latest

Needs proper review. Only made some changes for how article-journal is cited and how URLs are shown.


rmzelle:
Looks good, thanks!

PR #4387, opened 2019-11-01
Create dsh-harvard.csl

fedormyskin:
I created a style for the journal Digital Scholarship in the Humanities (DSH)


adam3smith:
(let me fix the basic validation and we can then discuss further)

adam3smith:
Thanks! Some questions:
1. Does the journal document the style online? If so, where?
2. Did you start from an existing style in creating this? Which?
3. Does the journal allow for different citation styles (e.g. a Harvard and a footnote style) or is this the only one (in which case we'd remove the "Harvard" label)

fedormyskin:
1. Documented here
https://academic.oup.com/dsh/pages/general_instructions#Typescript%20Text
2. I started from http://www.zotero.org/styles/griffith-college-harvard
3. Only this modified version of Harvard is allowed

adam3smith:
Thanks!

PR #4065, opened 2019-04-27
Create china-national-standard-gb-t-7714-2015-numeric.csl

buffalo-d:
"GB/T 7714-2015" is a national standard/guidelines of China for bibliographic references and citations to information resources. It is widely used in academic journals and papers in China mainland . It is drafted with reference to the international standard ISO 690:2010(E) and its previous version is "GB/T 7714-2005".

PR #4390, opened 2019-11-04
Create chinese-std-gb-t-7714-2015.csl

hktkzyx:
This style just partly implemented what the Chinese GB/T 7714-2015 requires and partly inherit from GB/T 7714-2005.


rmzelle:
@hktkzyx, thanks for your patience. We recently already added a CSL style for Chinese GB/T 7714-2015 (see https://www.zotero.org/styles?q=7714 and https://github.com/citation-style-language/styles/pull/4065).

PR #4391, opened 2019-11-05
Update catholic-biblical-association.csl

POBrien333:
via https://forums.zotero.org/discussion/79887/style-error-catholic-biblical-quarterly-and-catholic-biblical-association


rmzelle:
Thanks!

PR #4397, opened 2019-11-06
Create J_O_Ethnobio

jrw297:
Second improved csl style


adam3smith:
Actually, let's go with #4396 instead

PR #4398, opened 2019-11-07

nmueller18:
The category citation-format was defined twice which led to unwished behaviour (footnotes instead of in-text citations). In the proposed branch, the incriminating line is deleted.

### DIFF
--- a/journal-of-neolithic-archaeology.csl
+++ b/journal-of-neolithic-archaeology.csl
@@ -16,12 +16,16 @@
       <name>Journal of Neolithic Archaeology</name>
       <email>jna@ufg.uni-kiel.de</email>
     </contributor>
+    <contributor>
+      <name>Nils Müller-Scheeßel</name>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks. From a quick view of the styleguide, it'd seem like citations should also be in parentheses, right? Currently this does `Smith 1776` when I think is thould be `(Smith 1776)`

nmueller18:
Thanks for fixing the "in-text"-issue!

> Thanks. From a quick view of the styleguide, it'd seem like citations should also be in parentheses, right? Currently this does Smith 1776 when I think is thould be (Smith 1776)

Theoretically you are right. But I personally prefer to set the parentheses manually which gives you more flexibility to add information related to this citation.

adam3smith:
I'd rather see the parentheses here -- that's how all other such styles in CSL are implemented, so by removing that you break the ability to switch styles. 
You can obviously use a custom version without parentheses for your own purposes. Otherwise this is good to merge.

adam3smith:
Thanks!

PR #4400, opened 2019-11-08
Create diabetes-technology-and-therapeutics.csl

POBrien333:
via https://forums.zotero.org/discussion/79907/style-request-diabetes-technology-and-therapeutics

Minimal changes to the template style. Only adapted for article-journal.


rmzelle:
Thanks!

PR #4403, opened 2019-11-09
Update soziologiemagazin.csl

nathanlesage:
After some evaluations of the style, we have incorporated some changes so that it more closely matches what we expect for citations within the Soziologiemagazin.

Thank you very much!

### DIFF
--- a/soziologiemagazin.csl
+++ b/soziologiemagazin.csl
@@ -23,7 +23,7 @@
     <eissn>2198-9826</eissn>
     <summary>Official Citation Style for the German Soziologiemagazin</summary>
     <published>2019-09-22T12:00:00+00:00</published>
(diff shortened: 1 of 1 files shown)


nathanlesage:
For your notice: Apparently, the validator on the CI-build chokes when it encounters an `<if variable="url">`  within a `cs:choose` (I forgot to capitalise "URL") and then believes the `cs:locale`-element were superfluous — the validator at validator.citationstyles.org names the error correctly!

rmzelle:
Thanks!

And yes, https://validator.citationstyles.org/ depends on Jing (https://github.com/relaxng/jing-trang), which is renowned for giving clear validation error messages :).

nathanlesage:
Thank you very much! :)

PR #4407, opened 2019-11-13
Changed term on access macro, added rules for bill

abeeken:


university-of-lincoln-harvard.csl
```diff
@@ -61,9 +61,8 @@
   </macro>
   <macro name="access">
     <choose>
-      <if type="webpage article-journal article-newspaper" match="any">
-        <text term="available at" text-case="capitalize-first" suffix=" "/>
-        <text variable="URL"/>
+      <if type="bill webpage article-journal article-newspaper" match="any">
+        <text variable="URL" prefix="Available from "/>
```
adam3smith: I don't understand why you switched this? Using the term should give you exactly the same and is generally preferable (e.g., makes the style more robust for adaptation)

abeeken: Hi Sebastian,

We switched this as for some reason the term wasn't appearing for a "bill" type citation - using the prefix worked fine.

adam3smith: that really shouldn't matter and we don't like to work around bugs in citation managers. Was that in Zotero or a different tool?

abeeken: No, it was while we were building the style in the CSL editor: https://editor.citationstyles.org/visualEditor/

adam3smith: Could you walk me through steps to reproduce that? It really shouldn't happen (and I'm quite surprised it would, so I'm suspecting something else may have been going on), but if it does it's a bug in the editor we want to fix rather than to fix something in the style that isn't broken (e.g. I just tested in Zotero and using the term for the bill is fine).

abeeken: I've checked myself and I can't actually reproduce it - not sure what was going on when I was working on this but I've put the <text> tag back in and it's rendering fine now 🤷‍♂ I've committed the change again.

adam3smith:
(I'll fix validation; don't worry about it)

abeeken:
Hi! Just checking if we're okay to merge this? :)

rmzelle:
Thanks!

PR #4408, opened 2019-11-13

jjrondeau:
A few modifications to comply with Guide de présentation des mémoires et thèses and to be more compatible with citation informations downloaded from some databases and library catalogs.


adam3smith:
Looks good, thanks!

PR #4411, opened 2019-11-19
Add Presses universitaires de Strasbourg style

etnadji:
new file:   pus-note-fr.csl

According to [current instructions to authors](http://pus.unistra.fr/html/WYSIWYGfiles/files/Note-aux-auteurs_finale.pdf).

EDIT : Fixed conflicting ID -> pus-note-fr

### DIFF
--- a/pus-note-fr.csl
+++ b/pus-note-fr.csl
@@ -324,9 +324,6 @@
           <text macro="online"/>
         </group>
       </else-if>
-      <else>
-        <!-- <text macro="online" prefix="[" suffix="]"/> -->
-      </else>
     </choose>
   </macro>
   <macro name="date">
--- a/pus-note-fr.csl
+++ b/pus-note-fr.csl
@@ -1,0 +1,711 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="never" default-locale="fr-FR">
+
+  <!-- === Metadatas ======================================================= -->
+
+  <info>
+    <title>Presses Universitaires de Strasbourg (note, no abstract, ibid, French)</title>
+    <id>http://www.zotero.org/styles/pus-note-ibid-fr</id>
+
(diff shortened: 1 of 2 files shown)


pus-note-fr.csl
```diff
@@ -0,0 +1,711 @@
...
+      <term name="available at">disponible à l'adresse</term>
+      <term name="editor" form="short">dir.</term>
+      <term name="cited">, cité</term>
+      <term name="translator" form="verb-short">traduction de</term>
+      <term name="translator" form="verb">traduction de</term>
+      <term name="no date" form="short"></term>
+      <!-- Termes non-standards -->
+      <term name="opcit">op.&#160;cit.</term>
+      <term name="artcit">art.cité</term>
+      <term name="doiacronym">DOI&#160;:&#160;</term>
```
rmzelle: @etnadji, the style won't validate with non-standard CSL terms. You can use prefixes or verbatim text elements instead. E.g. replace

```xml
  <macro name="doi">
    <text term="doiacronym"/>
    <text variable="DOI"/>
  </macro>
```

by

```xml
  <macro name="doi">
    <text prefix="DOI&#160;:&#160;" variable="DOI"/>
  </macro>
```

or

 ```xml
  <macro name="doi">
    <group delimiter="&#160;:&#160;">
      <text value="DOI"/>
      <text variable="DOI"/>
    </group>
  </macro>
```

File: pus-note-fr.csl [hunk 2]
```diff
@@ -0,0 +1,711 @@
...
+            <if variable="medium">
+              <text variable="medium"/>
+            </if>
+          </choose>
+          <text macro="online"/>
+        </group>
+      </else-if>
+      <else>
+        <!-- <text macro="online" prefix="[" suffix="]"/> -->
+      </else>
```
rmzelle: Empty `<else/>` elements are also not allowed, and should just be deleted.

pus-note-fr.csl
etnadji: Thanks for this information ! It worked fine when I used it, maybe the CSL formatter was the one Zotero use.

rmzelle:
Thanks!

PR #4412, opened 2019-11-19

ahahl:
Added a few features (editorial-director ; better disambiguation ; "scale" field for maps ; and reviewed the URL output to avoid the conflict in Word processors between the URL and the following comma when automatically creating a weblink).
Thanks!


adam3smith:
(If you're having trouble with this, please just ask for help; don't close pull requests unless you're no longer interested in updating the citation style)

PR #4414, opened 2019-11-19
Create iso690-full-note-cs.csl

OTristanF:
I edited the Slovak full-note ISO-690 (iso690-full-note-sk) to create a CSL for the Czech version of the style (ČSN ISO-690). The full-note ČSN ISO-690 style is commonly used by Czech legal journals, so that's the reason why I changed the category to law.


rmzelle:
Thanks!

PR #4421, opened 2019-11-25
Update geistes-und-kulturwissenschaften-heilmann.csl

POBrien333:
via https://forums.zotero.org/discussion/80208/wer-kann-korrigieren


rmzelle:
Thanks. Did you strip the non-breaking spaces on purpose?

POBrien333:
Thank you for spotting my mistake.
I added them back in, although I don't completely understand if they actually make much sense.

rmzelle:
Thanks! :man_shrugging: better to leave them in :)

PR #4430, opened 2019-12-03
Create open-window.csl

justinhartman:
I have created a new style for the [Open Window](https://www.openwindow.co.za) academic institution. This file is based on the Harvard style but specific to the [Open Window Reference, Format and Style Guide for Academic Writing](https://drive.google.com/file/d/1Dg6RD92AArwhmGtfnfcLhamxoXdFvOhb/view) as per the academic institution's requirements.

As an aside, I am a lecturer at [Creative Academy](https://creativeacademy.ac.za) and we are affiliated with Open Window via shareholding between the two academic organisations. This may prove relevant to the acceptance of this style guide; I am not sure.


rmzelle:
Thanks!

PR #4429, opened 2019-12-02
Nutrition Research Reviews

justintondt:
Vancouver modified for parenthetical superscripts and et al. after 3 authors

PR #4428, opened 2019-12-01
Nutrition Research Reviews

justintondt:
Vancouver style modified for parenthetical superscript and et al after 3 authors

PR #4434, opened 2019-12-03
Nutrition Research Reviews

justintondt:
Vancouver modified for parenthetical superscripts and et al. for more than 3 authors

Continued from #4428 and #4429


POBrien333:
Your et-al rules are invalid. If you want to achieve an et al. for more than 2 authors set it to 3 and 1.

justintondt:
The et al. rules for the journal are to have the first 3 authors listed. If there are more authors, then use et al. after the 3rd author. Sorry for the confusion. It looks correct to me for this. Is there anything else I need to change?

justintondt:
This is an example citation of the et al.: 
Le HD, Meisel JA, de Meijer VE, et al. The essentiality of arachidonic acid and docosahexaenoic acid. Prostaglandins Leukot Essent Fat. Acids. 2009;81(2–3):165–70.

rmzelle:
Thanks!

PR #4436, opened 2019-12-06
JEMS.csl

burakkundakci:
Journal of ETA Maritime Science

### DIFF
--- a/JEMS
+++ b/JEMS
@@ -1,0 +1,349 @@
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
+<info>
+<title>Journal of ETA Maritime Science</title>
+<title-short>JEMS</title-short>
+<id>https://csl.mendeley.com/styles/484150931/JEMS</id>
+<link href="https://csl.mendeley.com/styles/484150931/JEMS" rel="self"/>
+<rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+<category citation-format="numeric"/>
+<category field="maritime"/>
+<updated>2019-12-05T11:29:34+00:00</updated>
+ <link href="http://www.zotero.org/styles/springer-socpsych-brackets" rel="template"/>
+ <issn>2147-2955</issn>
+ <eissn>2148-9386</eissn>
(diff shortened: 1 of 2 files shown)


JEMS
```diff
@@ -0,0 +1,349 @@
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
+<info>
+<title>Journal of ETA Maritime Science</title>
+<title-short>JEMS</title-short>
+<id>https://csl.mendeley.com/styles/484150931/JEMS</id>
```
POBrien333: change to: http://www.zotero.org/styles/journal-of-eta-maritime-science

JEMS
```diff
@@ -0,0 +1,349 @@
...
+<link href="https://csl.mendeley.com/styles/484150931/JEMS" rel="self"/>
```
POBrien333: also change to: http://www.zotero.org/styles/journal-of-eta-maritime-science

JEMS
```diff
@@ -0,0 +1,349 @@
...
+<rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+<category citation-format="numeric"/>
+<category field="maritime"/>
```
POBrien333: change to biology. 
It has to be one of these categories: http://docs.citationstyles.org/en/stable/specification.html#appendix-i-categories

JEMS
POBrien333: Also, change the filename to: journal-of-eta-maritime-science.csl
It has to match.

rmzelle:
@burakkundakci, could you change the setting on this pull request so we can make some corrections for you? There should be a checkbox in the right-hand column (see the example screenshot below):

image

burakkundakci:
> @burakkundakci, could you change the setting on this pull request so we can make some corrections for you? There should be a checkbox in the right-hand column (see the example screenshot below):
> 
> image

I change the setting. Thank you for your reply.

JEMS
burakkundakci: Thank you for your reply. I made the changes.

JEMS
burakkundakci: Thank you for your reply. I made the changes.

JEMS
burakkundakci: Thank you for your reply. I change it to engineering.

rmzelle:
Thanks!

burakkundakci:
> Thanks!

thank you for your help.

PR #4437, opened 2019-12-07
Create Phytotherapy Research.csl

Debashish-Mondol:
I am cordially requesting you to create the reference style as per the suggested journal as soon as possible.


adam3smith:
See here on style requests: https://github.com/citation-style-language/styles/wiki/Requesting-Styles

Pull requests are for style _contributions_

PR #4440, opened 2019-12-08
Title case titles in mcgill-en.csl

wohanley:
I've been following the [convention](https://citationstyles.org/authors/#/titles-in-sentence-and-title-case) of storing titles in sentence case so that they can be converted when necessary. The McGill Guide says to print titles as they appear on the documents in question, enforcing neither title case nor sentence case. The result is that titles come out in sentence case, which is not how titles appear on most English documents. This pull request enforces title case instead, which is also not correct but is generally much closer for those following the convention.

I think this is probably not a good way to address this problem. I include the code changes only because I made them already for my own convenience and they make it clear what I mean - they probably shouldn't be merged. I would prefer to update the documentation that recommends the "store in sentence case" convention to make it clear that it will cause problems with styles like McGill that don't enforce any specific casing. I'm not sure how to propose such a change, and these changes might be useful as a band-aid, so here I am. If someone with more knowledge about this project knows how I ought to proceed and could give me some advice, I would appreciate it!

### DIFF
--- a/mcgill-en.csl
+++ b/mcgill-en.csl
@@ -17,9 +17,13 @@
       <email>f.martin-bariteau@umontreal.ca</email>
       <uri>http://f-mb.github.io/cslegal/</uri>
     </contributor>
+    <contributor>
+      <name>William O'Hanley</name>
+      <email>me@wohanley.com</email>
+    </contributor>
     <category citation-format="note"/>
     <category field="law"/>
-    <updated>2014-06-28T19:23:00+00:00</updated>
+    <updated>2019-12-01T22:49:53+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks! I don't think we'll take this -- overall, I'd expect it to cause more problems than it causes and not specifying casing in the style is closer to the McGill prescription. I don't think I'll want to confuse people by discussing this rule in documentation because the bigger issue, of course, is that the McGill prescription simply makes no sense for English publication. To just take a random example off my bookshelf, the cover of the _Turabian_ guide is:
"A Manual for Writers OF RESEARCH PAPERS, THESES, AND DISSERTATIONS"
on opening the book on the inside, we find
"A Manual for Writers of Research Papers, Theses, and Dissertations"
and on the copyright page it's
"A manual for writers of research papers, theses, and dissertation"

Given that it's literally impossible to follow that rule, I'd ignore it and just use sentence case throughout.

wohanley:
That's a good point, thanks for explaining.

PR #4443, opened 2019-12-09
Update queen-margaret-university-harvard.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/344702/#Comment_344702
Not sure why the author names in the bibliography (author macro) are not in all caps. I thought that was just not shown in the editor due to a bug, but it seems not to show it.

### DIFF
--- a/queen-margaret-university-harvard.csl
+++ b/queen-margaret-university-harvard.csl
@@ -58,7 +58,7 @@
     <choose>
       <if type="motion_picture broadcast" match="none">
         <names variable="author">
-          <name font-variant="normal" and="text" initialize-with="." name-as-sort-order="all">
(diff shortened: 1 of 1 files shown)


rmzelle:
> Not sure why the author names in the bibliography (author macro) are not in all caps. I thought that was just not shown in the editor due to a bug, but it seems not to show it.

It was because `text-case="capitalize-all"` capitalizes the first letter of each word, not all letters of each word. You needed "uppercase" instead. See http://docs.citationstyles.org/en/1.0.1/specification.html#text-case.

rmzelle:
Thanks!

POBrien333:
Cheers!

PR #4447, opened 2019-12-12
Update canadian-geotechnical-journal.csl

POBrien333:
via https://forums.zotero.org/discussion/80471/canadian-journal-of-civil-engineering


rmzelle:
Thanks!

PR #4450, opened 2019-12-16
Create institut-catholique-de-paris.csl

ddanangaw:
Hai, 
This dependent style of Institut Catholique de Paris would rely the Transversalités CSL style.


rmzelle:
Thanks!

rmzelle:
(and if there is a webpage of ICP that mentions that they use the style guide of Transversalités, that would be good to document as well)

ddanangaw:
@rmzelle, Thanks for your help. I haven't found the specific webpage you requested, I'll let you know if I find it. As far as i experienced in my study at this university, the style guide used by Transversalités is the reference.

PR #4451, opened 2019-12-16
Create theologicum.csl

ddanangaw:
Hai, 
This dependent style of Theologicum (Faculté de Théologie et de Sciences Religieuses de l'Institut Catholique de Paris) would rely the Transversalités CSL style.


rmzelle:
@ddanangaw, thanks, but we wouldn't typically add dependent styles for departments if we already have a style for the entire institution (ICP).

ddanangaw:
@rmzelle, It's OK. Thanks for your information.

rmzelle:
Thanks!

PR #4452, opened 2019-12-16
Update kidney-international.csl

AlastairHeffernan:
Updated to match bibliography style of recent Kidney International papers (as per pdf versions of articles). 

Changes made to original Kidney International format are:
- Stripped full stops (periods) in abbreviated journal titles
- Full stop added after journal title.
- Volume number no longer bold
- "et al" no longer italicised
- Comma delimiter added before "et al"
- Space between year, volume and page numbers removed
- Hanging indents added

### DIFF
--- a/kidney-international.csl
+++ b/kidney-international.csl
@@ -17,7 +17,7 @@
     <category field="medicine"/>
     <issn>0085-2538</issn>
     <eissn>1523-1755</eissn>
-    <updated>2012-09-27T22:06:38+00:00</updated>
+    <updated>2014-08-28T01:36:46+00:00</updated>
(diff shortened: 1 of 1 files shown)


rmzelle:
@AlastairHeffernan, thanks. According to https://www.kidney-international.org/content/authorinfo#idp1849328, the journal uses standard AMA style (https://www.zotero.org/styles?q=id%3Aamerican-medical-association), so we should be able to turn this into a dependent style of that format.

AlastairHeffernan:
Hi @rmzelle - yes that looks right. Could have saved myself some work by just using AMA but yes making it a dependent style would be the most sensible thing.

adam3smith:
Made the change to Kidney Int'l here: https://github.com/citation-style-language/styles/commit/fad65b7341b0d93c1d3a50e684ed74100d3147cc

PR #4456, opened 2019-12-17

POBrien333:
Small fix for conference papers via https://forums.zotero.org/discussion/80550/acm-standard-reference-format


adam3smith:
This change looks good. I'm not quite clear why we have book-publisher and pages for speech (i.e. presentation) but leaving that in for now.

PR #4459, opened 2019-12-18

POBrien333:
add genre for thesis.
Via https://forums.zotero.org/discussion/comment/345228/#Comment_345228


adam3smith:
Thanks!

PR #4463, opened 2019-12-18
Update .travis.yml

rmzelle:
https://docs.travis-ci.com/user/reference/bionic/#ruby-support

See if we get better performance with Ubuntu Bionic Beaver and preinstalled Ruby 2.6.3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
-dist: xenial
+dist: bionic
 language: ruby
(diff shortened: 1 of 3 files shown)


rmzelle:
Apparently Travis docs weren't correct on preinstalled Ruby versions: https://github.com/travis-ci/docs-travis-ci-com/pull/2614

PR #4465, opened 2019-12-18
Create romanian-iso-690-full-note-with-ibid-romanian.csl

mihaiconstantinescu:
This style is based on the Inventaire général du patrimoine culturel - ISO-690 (full note, with Ibid., French) style and has been developed in order to fulfill the needs of the Romanian academic community.
I've made several changes to adjust the needs of the Romanian community (using the terms vol and nr in the print version instead of just the numbers for example, changing terms according to the Romanian version etc)


adam3smith:
Thanks! Is the Romanian part of this documented somewhere?

mihaiconstantinescu:
> 
> 
> Thanks! Is the Romanian part of this documented somewhere?

I'm not really sure what you mean by „documented somewhere” (I'm new to github and csl, srry). Do u mean if I created a document showing the translations that I used? if that, the answer is no but I want to (as soon as I can).

adam3smith:
> Do u mean if I created a document showing the translations that I used?

Basically, yes. You mention some other change like "using the terms vol and nr in the print version instead of just the numbers for example" -- we would very much like documentation these changes are based on so we have a point of reference e.g. if someone complains about (what they think is) an error in a style.

mihaiconstantinescu:
> 
> 
> > Do u mean if I created a document showing the translations that I used?
> 
> Basically, yes. You mention some other change like "using the terms vol and nr in the print version instead of just the numbers for example" -- we would very much like documentation these changes are based on so we have a point of reference e.g. if someone complains about (what they think is) an error in a style.

ok. I understand. Will upload a document detailing these changes in 24 hr or less.

mihaiconstantinescu:
> 
> 
> > Do u mean if I created a document showing the translations that I used?
> 
> Basically, yes. You mention some other change like "using the terms vol and nr in the print version instead of just the numbers for example" -- we would very much like documentation these changes are based on so we have a point of reference e.g. if someone complains about (what they think is) an error in a style.

I'm not sure where I should upload this document (and how this document should be structured or written) so I uploaded it here.  
hope everything is ok.
[changes for the Romanian ISO 690.docx](https://github.com/citation-style-language/styles/files/3991365/changes.for.the.Romanian.ISO.690.docx)

rmzelle:
Great, thanks!

PR #4466, opened 2019-12-19
Update stale.yml

rmzelle:
* Disable autoclosing of stale issues/PRs
* Limit stale bot to PRs

See also https://github.com/probot/stale#usage


adam3smith:
OK, let's experiment with no auto-closing and we'll see how that goes.

PR #4469, opened 2019-12-22
Update drug-and-alcohol-dependence.csl

katesklein:
Fixed a couple things that tidied up my reference list (but may not address all reference types) - hopefully it is helpful to others as well! Includes: ordering papers in ref list first by first author's name only then by ascending issued-year; added "(accessed on [date])" for web page citations; fixed author name abbreviation formatting in reference list (added periods after author initials, used only commas as the delimiter).


POBrien333:
Hi,
You are trying to overwrite a dependent style. DAD just uses the elsevier-harvard.csl style.
These styles are automatically generated.
elsevier-harvard already does all the things you are trying to fix. 
https://www.zotero.org/styles?q=id%3Aelsevier-harvard

adam3smith:
Thanks both

PR #4470, opened 2019-12-23
Update in-text citation for legal_case for BJPS

POBrien333:
via https://forums.zotero.org/discussion/80600/citing-cases-issue-bjps-style-inserts-anon-for-author


rmzelle:
Thanks!

PR #4477, opened 2019-12-29
Create informal-logic.csl

waleedmebane:
first version


rmzelle:
Replaced by #4478 (@waleedmebane, note that you can just update an existing pull request if you need to make any changes)

PR #4478, opened 2019-12-30
Add files via upload

waleedmebane:
informal-logic.csl, first version

--

"I am a former editorial assistant/copyeditor for the journal,
and I am submitting this file with the hope that it will be of help to authors.
To the best of my knowledge, it matches the journal's
citation style as shown at the website under information
for authors, except for any mistakes, which are solely my
own.  I have tested it using Zotero."


rmzelle:
Thanks!

PR #4479, opened 2019-12-30
Update budownictwo-i-architektura-pl.csl

lukboro:
This version is prepared both for Polish and English language (user have to change the language in the software e.g Mendeley). 
Fixed errors with DOI. Same small changes of style. 
The file was validated.


lukboro:
This version is prepared both for Polish and English language. So there is no need to have two versions of Budowncitwo i Archietektura style (PL and EN).

lukboro:
expected: "http://www.zotero.org/styles/budownictwo-i-architektura-pl"
     got: "http://www.zotero.org/styles/budownictwo-i-architektura-v2"

this looks ok.

rmzelle:
Thanks!

PR #4480, opened 2019-12-30
Update iso690-full-note-cs.csl

OTristanF:
Most importantly, added ibidem. Also fixed the missing italic in the subsequent citations.


rmzelle:
Thanks!

PR #4482, opened 2020-01-02
Create ahmad-ibrahim-kulliyyah-of-laws-international-islamic-universi…

POBrien333:
…ty-of-malaysia.csl

Style was created over the last months together with the requestor. Lots of back and forth, but the style is signed off by the curators of that university and now want to share it with all their students.
I do have the style guidelines (also some slightly newer ones) on file and there is a link to the older ones in this thread: https://forums.zotero.org/discussion/78653/style-request-aikol-iium-thesis-citation-style/p1


rmzelle:
Thanks!

PR #4483, opened 2020-01-02
Make ET&C dependent of CSE

POBrien333:
via https://forums.zotero.org/discussion/24240/new-citation-style-environmental-toxicology-and-chemistry#latest


rmzelle:
Thanks!

PR #4484, opened 2020-01-02
APA: Use "et al." in Dutch

bwiernik:
Also clean up the 6th edition locales a bit.

https://forums.zotero.org/discussion/80710/abriviation-of-et-alli#latest

Psychology journals currently publishing in Dutch seem mostly to use "et al." (e.g., https://dspace.library.uu.nl/bitstream/handle/1874/378701/GO_32_1_1.pdf?sequence=1), and several of the universities require Dutch-language theses to use "et al.".


rmzelle:
Bedankt!

PR #4485, opened 2020-01-03
Update weed-research.csl

vidotto:
According to current Weed Research citation style, in in-text citations two authors should be separated by "&" and not by "and".


rmzelle:
Thanks!

PR #4487, opened 2020-01-04
Create apa-with-abstract.csl

bwiernik:
There have been a few requests for this. Addresses https://github.com/citation-style-language/styles/pull/4419#discussion_r362609107


rmzelle:
Thanks!

PR #4488, opened 2020-01-06
Improve legal cites for british-journal-of-political-science

POBrien333:
via https://forums.zotero.org/discussion/comment/345980/#Comment_345980


rmzelle:
Thanks!

PR #4495, opened 2020-01-10
APA: Fix sort order for in-text citations

bwiernik:
https://forums.zotero.org/discussion/80814/style-error-apa-csl


rmzelle:
Thanks!

PR #4497, opened 2020-01-10
Further improve legal cites for BJPS

POBrien333:
via https://forums.zotero.org/discussion/comment/346237/#Comment_346237

Remove "anonymous" from style. Did not find any paper that shows "Anon". Must be from the template style.


rmzelle:
Thanks!

PR #4499, opened 2020-01-11
Create Publications du Groupe d'anthropologie et d'archéologie funéra…

FrancoisC37:
Hi,
I've done this style. I hope it can be integrate to the style repository
I don't have this link `<link href="http://www.zotero.org/styles/" rel="self"/>`
Thanks
François


rmzelle:

FrancoisC37:
Yes I use this one (the daf) one but I don't know how to put this line on this style...
<link href="http://www.zotero.org/styles/documents-d-archeologie-francaise" rel="template"/>
Thanks in advance!

rmzelle:
Perfect, thanks!

PR #4500, opened 2020-01-11
Create citation-compass-apa-note.csl

bwiernik:
https://forums.zotero.org/discussion/80599/request-new-style-apa-6th-for-footnotes

(Let's let the requester test it before merging.)

### DIFF
--- a/citation-compass-apa-note.csl
+++ b/citation-compass-apa-note.csl
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="never" page-range-format="expanded">
   <info>
-    <title>Citation Compass - APA (note)</title>
+    <title>Citation Compass (Kildekompasset) - APA (note)</title>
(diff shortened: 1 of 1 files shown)


bwiernik:
Okay, this should be good to go.

rmzelle:
@bwiernik, do you know if this style would only be used in English, only in one of the Norwegian locales (Bokmål, Nynorsk), or a combination? Even if we leave off a default-locale, we should probably strip the in-style translations of any other languages (like Danish, German, Spanish, and Dutch).

rmzelle:
And we could also name the style `<title>Citation Compass (Kildekompasset) - APA (note)</title>` (but leave the style ID as is), which we sometimes do in cases like these (see e.g. https://www.zotero.org/styles?q=id%3Aacta-universitatis-agriculturae-sueciae&dependent=0)

bwiernik:
Okay, made those changes. Should I add the other Norwegian locale for `et al` (and also in APA styles, versus just having it in `nn` but not `by`)?

rmzelle:
It's my understanding that "nb" is Norwegian Bokmål (see https://r12a.github.io/app-subtags/), which is already used in the APA styles. I'll put that back and merge the style.

rmzelle:
Thanks!

Issue #3243, opened 2018-01-04
Chicago Manual Presentation issues


adam3smith:
https://github.com/jgm/pandoc-citeproc/issues/45
and 
https://forums.zotero.org/discussion/69698/style-error-chicago-fullnote-bibliography-csl-presentation-does-not-output-type#latest

njbart:
@adam3smith – In https://github.com/jgm/pandoc-citeproc/issues/45#issuecomment-355379588 you said, “Chicago has so many moving parts that it always takes a bit longer to fix, but shouldn't be too hard here.” – Your work’s much appreciated, and I have no complaints at all about having to wait some time …

… but I do have my doubts whether the new stale bot indiscriminately closing issues after 30 days of inactivity is such a good idea – it seems to kind of defeat the actual purpose of an issue tracker.

njbart:
Great this is being dealt with. However, it seems that with the latest fixes to chicago-fullnote-bibliography.csl, `event` and `event-place` are omitted from the notes (they used to be included previously; see https://github.com/jgm/pandoc-citeproc/issues/45#issue-34230655).

CMOS (17e, 14.217) quite clearly does call for including this information in notes:

> 1. David G. Harper, “The Several Discoveries of the Ciliary Muscle” (Power­Point presentation, 25th Anniversary of the Cogan Ophthalmic History Society, Bethesda, MD, March 31,2012).

> 2. Viviana Hong, “Censorship in Children’s Literature during Argentina’s Dirty War (1976-1983)” (lecture, University of Chicago, Chicago, IL, April 30, 2015).

pandoc example (same as in https://github.com/jgm/pandoc-citeproc/issues/45#issue-34230655):

```
pandoc -s -F pandoc-citeproc -t plain <<EOT

Foo [@item1]. Bar [@item2].

# References {-}

---
csl: chicago-fullnote-bibliography.csl
references:
- title-short: Early violent death in delinquent youth
  author:
  - family: Teplin
    given: Linda A.
  - family: McClelland
    given: Gary M.
  - family: Abram
    given: Karen M.
  - family: Washburn
    given: Jason J.
  id: item2
  issued:
    date-parts:
    - - 2005
      - 3
  title: 'Early violent death in delinquent youth: A prospective longitudinal study'
  type: speech
  genre: paper
  event: Annual Meeting of the American Psychology-Law Society
  event-place: La Jolla, CA

- author:
  - family: "D’Erasmo"
    given: Stacy
  id: item1
[...]

PR #4503, opened 2020-01-12
Chicago fixes & add no-ibid variant

adam3smith:
- add chicago-fullnote-short-title-subsequent, which behaves like the 
16th ed. no-ibid style, printing short titles on all subsequent 
citations, including in ibid position
- remove all `number variable` calls on issue and volume unless they 
specifically reference number properties (see #3243)
- Fix the issue with event and event place for presentations noted in 
#3243

### DIFF
--- a/chicago-annotated-bibliography.csl
+++ b/chicago-annotated-bibliography.csl
@@ -516,10 +516,10 @@
       <if type="article-journal">
         <group delimiter=", ">
           <text macro="collection-title-journal"/>
-          <number variable="volume"/>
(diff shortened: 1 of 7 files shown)


adam3smith:
@rmzelle could you please check on the metadata of the new style?

rmzelle:
@adam3smith, should be good to go. (all of our "fullnote" styles have "-bibliography" in the file name, so let's do that here too)

rmzelle:
Thanks!

PR #4510, opened 2020-01-15

POBrien333:
Small fix for problem I introduced with last edit:
https://forums.zotero.org/discussion/comment/346663/#Comment_346663


rmzelle:
Thanks!

PR #4517, opened 2020-01-17
Create international-journal-of-population-data-science.csl

POBrien333:
https://forums.zotero.org/discussion/80753/style-request-ijpds

The style in papers is a bit mixed. I tried to make a "common denominator" style based on several .pdfs.


rmzelle:
Thanks!

PR #4518, opened 2020-01-17
Add (APA) to <title> tag

rmiessle:
This will make searching for the style in RefWorks easier, RefWorks doesn't search the <title-short> tag.


adam3smith:
If anything this would have to go into all APA styles, but I don't think we're inclined to take it. This is exactly what the `<short-title>` field is for and RefWorks should just support that - we only change styles to workaround downstream deficiencies in extraordinary cases.

rmzelle:
@rmiessle, thanks for the suggestion, but I also don't want to make this change. We've been trying very hard over the years to name styles consistently and concisely, and this includes avoiding redundancy in `<title/>` (like having both the abbreviated and unabbreviated name in it).

https://www.zotero.org/styles searches styles by title, abbreviated title, and file name. Like @adam3smith suggests, it would be better if RefWorks would be changed to work like that as well.

PR #4520, opened 2020-01-19
Create harvard-lancaster-university.csl

lbaajh:
Lancaster University (UK) Harvard referencing style


rmzelle:

lbaajh:

I initially started with harvard-imperial-college-london.csl, but ended up replacing 95% with original code, so not really based on any existing CSL.

rmzelle:
Thanks, let's still add it (we usually try to include a template link unless somebody starts from a clean slate).

PR #4521, opened 2020-01-21

PriscilaSJUFRGS:
Including new changes

### DIFF
--- a/associacao-brasileira-de-normas-tecnicas-ufrgs.csl
+++ b/associacao-brasileira-de-normas-tecnicas-ufrgs.csl
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" initialize="false" demote-non-dropping-particle="never" default-locale="pt-BR">
   <info>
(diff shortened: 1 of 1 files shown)


associacao-brasileira-de-normas-tecnicas-ufrgs.csl
```diff
@@ -207,14 +208,17 @@
...
     <choose>
       <if type="paper-conference" match="any">
-        <text variable="container-title" text-case="uppercase"/>
+        <text variable="event" text-case="uppercase"/>
         <text variable="number" prefix=", " suffix=". , "/>
         <text macro="issued-year" prefix=" " suffix=", "/>
         <text variable="publisher-place" suffix=". "/>
         <choose>
-          <if match="any" variable="title-short">
+          <else-if match="any" variable="container-title">
```
adam3smith: you can't start `<choose>` with `else-if`. It has to start with `if`

PriscilaSJUFRGS: thanks! corrected!

adam3smith:
Thanks! Since UFRGS is already in the short-title, no need to also include it in the main title. We try to keep that free of acronyms. If you search e.g. on zotero.org/styles, search results will include the short title.

PR #4533, opened 2020-01-28

JoshuaWeberFHNW:
Zotero assigns lowercase letters for publications with several authors in the same year, even if they are not the same authors, e.g. here (both cases are obviously simply processed as 'Ludwig-Mayerhofer et al.):
- Ludwig-Mayerhofer, Wolfgang/Behrend, Olaf/Sondermann, Ariadne (Hg.) (2007a). Fallverstehen und Deutungsmacht. etc.
- Ludwig-Mayerhofer, Wolfgang/Dölemeyer, Anne/Sondermann, Ariadne (2007b). Die neue Staatlichkeit: etc.


adam3smith:
Thanks!

PR #4534, opened 2020-01-28
Create south-african-actuarial-journal.csl

POBrien333:
Guidelines all over the place... Minimal effort style.
Via https://forums.zotero.org/discussion/81012/style-request-south-african-actuarial-journal


rmzelle:
Thanks!

PR #4538, opened 2020-01-28
Update clinical-and-experimental-optometry.csl

cxoadmin:
The citation style had a major upgrade in 2018 and this style is now obsolete. Please update as per the new guidelines https://onlinelibrary.wiley.com/page/journal/14440938/homepage/ForAuthors.html or use the current EndNote file for reference.


rmzelle:
@POBrien333, could you take a look?

POBrien333:
In-text: <sup>1</sup>
Bib: 
1. Campbell JL, Pedersen OK. The varieties of capitalism and hybrid success. _Comp Polit Stud_ 2007; 40: 307–332.

This looks like sage-vancouver.csl

Open papers: 
https://onlinelibrary.wiley.com/doi/epdf/10.1111/cxo.12948
https://onlinelibrary.wiley.com/doi/epdf/10.1111/cxo.12993

adam3smith:
Thanks @POBrien333 -- agree that's a good fit, including the 4/3 et al. for the bibliography. I think we should just go with that.

rmzelle:
Thanks!

PR #4550, opened 2020-02-05
Create ipb-ppki-3.csl

auriza:


rmzelle:
[PPKI_III.pdf](https://github.com/citation-style-language/styles/files/4163217/PPKI_III.pdf)

rmzelle:

auriza:

We use [visual editor](https://editor.citationstyles.org/visualEditor/) using the default template, based on [APA](http://www.zotero.org/styles/apa).

rmzelle:
Thanks!

auriza:
> Thanks!

You're welcome. Thank you very much for your assistance.

PR #4553, opened 2020-02-06

nmueller18:
Force "Author Year" even if a short title is supplied.


rmzelle:
Thanks!

PR #4554, opened 2020-02-10
Update journal-of-forensic-sciences.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/348224/#Comment_348224

### DIFF
--- a/journal-of-forensic-sciences.csl
+++ b/journal-of-forensic-sciences.csl
@@ -64,20 +64,6 @@
       </else-if>
     </choose>
   </macro>
-  <macro name="accessed-date">
-    <choose>
-      <if variable="URL">
-        <group prefix="[" suffix="]" delimiter=" ">
-          <text term="cited"/>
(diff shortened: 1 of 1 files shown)


journal-of-forensic-sciences.csl
```diff
@@ -27,51 +28,47 @@
...
-          <text term="cited"/>
-          <date variable="accessed">
-            <date-part name="year"/>
-            <date-part name="month" prefix=" " form="short" strip-periods="true"/>
-            <date-part name="day" prefix=" "/>
-          </date>
-        </group>
+      <if type="article-journal" match="any">
+        <choose>
+          <if match="any" variable="volume page">
```
adam3smith: I think you mean match="none" here? But reading quickly, so letting you double-check

POBrien333: Thank you for spotting that! :)

rmzelle:
Thanks!

PR #4557, opened 2020-02-10
Create CSL Styles: estudios_hispanicos.csl

editio:
The style is going to be used by the journal _Estudios hispánicos_ (Wroclaw, Poland) <http://www.estudios.uni.wroc.pl>. It would be great to include it in the repository. Thank you


editio:
perfect. so far so good.

rmzelle:
Thanks!

editio:
Thank you for your awesome work!

PR #4558, opened 2020-02-11
Create dermatology-online.csl

POBrien333:
via https://forums.zotero.org/discussion/81270/style-request-dermatology-online-journal

N.B.: editor labels for books and chapters are on purpose either short or long respectively.


rmzelle:
Thanks!

PR #4562, opened 2020-02-13

POBrien333:
via #4389
Note, the exact formatting that is request is not possible (insert version in-between publisher and publisher-place with dots), to the best of my knowledge.
I placed the Version between container-title and the publisher macro.

### DIFF
--- a/springer-basic-author-date.csl
+++ b/springer-basic-author-date.csl
@@ -68,7 +68,7 @@
       <date-part name="year"/>
     </date>
   </macro>
- <macro name="title">
+  <macro name="title">
     <choose>
       <if type="book">
         <group delimiter=". ">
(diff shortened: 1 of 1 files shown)


bwiernik:
No, it should definitely be possible to do that format (though its dumb; version more correctly belongs next to the title). In practice, it seems to be placed after the title: https://link.springer.com/article/10.1007/s00027-020-0701-9

adam3smith:
It's possible, but cumbersome. You'd do (in pseudo-code)
if version
delimit elements by ". "
else
delimit elements by ", "

But anyway -- I can't find any basis for the format suggested in the other thread and I completely agree with bwiernik (and the example he posts) that the version number belongs with the title, so merging this as is.

PR #4565, opened 2020-02-14
Create ayer.csl

POBrien333:
via https://forums.zotero.org/discussion/81228/style-request-revista-ayer-by-ahc-espana

### DIFF
--- a/ayer.csl
+++ b/ayer.csl
@@ -1,0 +1,365 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="note" version="1.0" page-range-format="minimal" demote-non-dropping-particle="sort-only" default-locale="es-ES" xmlns="http://purl.org/net/xbiblio/csl">
+  <info>
+    <title>Ayer</title>
+    <id>http://www.zotero.org/styles/ayer</id>
+    <link href="http://www.zotero.org/styles/ayer" rel="self"/>
+    <link href="http://www.zotero.org/styles/the-journal-of-transport-history" rel="template"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
@rmzelle could you check out the Hancké et al. citation above?
If Zotero/citeproc-js, it renders (correctly) as 
Bob Hancké, Martin Rhodes y Mark Thatcher (eds.): Beyond Varieties of Capitalism : Conflict, Contradiction, and Complementarities in the European Economy, Oxford and New York, Oxford University Press, 2007.

(disregarding italics; note the "en : " in the citeproc-ruby example above. Similar issues for the other item types). 
The difference appears to be that citeproc-js treats the macro like a group and omits the term, citeproc-ruby doesn't. Clearly, the -js behavior is desirable here, but I don't see that it's actually in the specs. Am I overlooking something? Thoughts?

adam3smith:
In the meantime, though, the style looks good, so merging here.

rmzelle:
> The difference appears to be that citeproc-js treats the macro like a group and omits the term, citeproc-ruby doesn't. Clearly, the -js behavior is desirable here, but I don't see that it's actually in the specs. Am I overlooking something? Thoughts?

Are we talking about the "editor-translator" macro acting as an implicit conditional here with citeproc-js?

The CSL 1.0.1 spec is clear that only `<group/>` acts as an implicit conditional (http://docs.citationstyles.org/en/1.0.1/specification.html#group), and I don't recall any discussion whether to extend this behavior to `<macro/>`. Personally I think I prefer the citeproc-ruby behavior as it makes `<macro/>` act more transparently. I.e., I think it's preferable that code would always behave exactly the same when it's moved into a macro.

(the style is also missing a language tag in the title, by the way, which I'll fix now)

PR #4567, opened 2020-02-15
Update american-journal-of-physical-anthropology.csl

bwiernik:
https://forums.zotero.org/discussion/81405/style-updated#latest


adam3smith:
Thanks!

PR #4571, opened 2020-02-18
Reorder/reindent

rmzelle:


rmzelle:
@adam3smith, looks like Travis emails still work. 👍

PR #4577, opened 2020-02-24
Update IEEE to use long Journal names

jaamulberry:
Page 13-14 of https://ieeeauthorcenter.ieee.org/wp-content/uploads/IEEE-Reference-Guide.pdf

> Virtual Journal
> II. Style—14
> Basic Format:
> Name(s) of Ed(s)., “Title of Issue,” in _Title of Journal_, Abbrev. month year. [Online]. Available: URL

Journal names should not be abbreviated

### DIFF
--- a/ieee.csl
+++ b/ieee.csl
@@ -34,7 +34,7 @@
     <category field="engineering"/>
     <category field="generic-base"/>
     <summary>IEEE style as per the 2018 guidelines, V 11.12.2018.</summary>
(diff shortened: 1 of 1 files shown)


adam3smith:
No, this is a misreading of the reference guidelines. What you cite refers to citing an entire virtual issue of a journal -- which would be tricky with CSL, but probably best be treated as an edited book. The standard format for journal articles continues to be abbreviated:
image
(p. 12)

PR #4578, opened 2020-02-24
chicago-fullnote-bibliography.csv: use reprint note if exists, cf. BibLaTeX 2.2.2 & CMoS 14.114

cacology:
## Expected Behavior

There should be a variable in chicago-fullnote-bibliography that inserts a user-generated reprint note.

## Actual Behavior

The CSL only writes "Reprint" or "rept." in certain special circumstances.

## Discussion

_Chicago Manual of Style_ (17th ed., CMoS) 14.114 allows free-text reprint descriptions, e.g.

> Emerson, Ralph Waldo. _Nature._ 1836. **Facsimile of the first edition, with an introduction by Jaroslav Pelikan.** Boston: Beacon Press, 1985.  (**bold** mine, p. 807)

or

> Fitzgerald, F. Scott. _The Great Gatsby._ New York: Scribner, 1925. **Reprinted with preface and notes by Matthew J. Bruccoli.** New York: Collier Books, 1992. [...] (**bold** mine, p. 807)

While the CSL standard suggests "note" is for commentary:

> (short) inline note giving additional item details (e.g. a concise summary or commentary) [link](https://docs.citationstyles.org/en/stable/specification.html#appendix-iv-variables)

The tradition inherited from BibLaTeX (and the crosswalks to CSL JSON) suggests that it's for:

> Miscellaneous bibliographic data which does not fit into any other field. The note field may be used to record bibliographic data in a free format. Publication facts such as “Reprint of the edition London 1831” are typical candidates for the note field. (Manual, 2.2.2) [link](https://www.ctan.org/pkg/biblatex)

I propose that we are to understand "a concise summary or commentary" as relating to the reprint preparation, thus restoring functionality present already in BibLaTeX.

In any case, I've edited my files locally, which solves my problem, but I thought someone else might find this useful too.

### Potential Side Effects

Zotero calls this variable "Extra" in its interface.  It too exports it to BibLaTeX "note", which suggests that the original purpose was enhanced publication info, but people could have used it in various weird ways that are as unpredictable as what "extra" might mean.

Respectfully submitted, open to tweaks/discussion/whatever

### DIFF
--- a/chicago-fullnote-bibliography.csl
+++ b/chicago-fullnote-bibliography.csl
@@ -729,22 +729,28 @@
   <macro name="reprint-note">
     <!--needs localization-->
     <choose>
-      <if variable="original-date issued" match="all">
+      <if variable="note" match="all">
+        <text variable="note"/>
+      </if>
+      <else-if variable="original-date issued" match="all">
         <choose>
           <!--for whatever reason in notes, when we have both original and new publishers, reprint doesn't appear-->
           <if variable="original-publisher original-publisher-place" match="none">
             <text value="repr."/>
           </if>
         </choose>
-      </if>
+      </else-if>
     </choose>
   </macro>
   <macro name="reprint">
     <!--needs localization-->
     <choose>
-      <if variable="original-date issued" match="all">
-        <text value="reprint" text-case="capitalize-first"/>
+      <if variable="note" match="all">
+        <text variable="note" text-case="capitalize-first"/>
       </if>
+      <else-if variable="original-date issued" match="all">
+        <text value="reprint" text-case="capitalize-first"/>
+      </else-if>
     </choose>
   </macro>
   <macro name="publisher">


adam3smith:
I understand the need, but for legacy reasons we definitely can't do this using `note`. It would lead to thousands of users suddenly ending up with unexpected content in their Chicago footnotes (I believe Mendeley actually puts user notes into `note`, so it wouldn't just be Zotero users).

Not sure there is a current viable solution in CSL -- my view has always been that at some point Chicago footnotes are sufficiently ideosyncratic as to require manual editing, but I understand that's not really feasible with pandoc-type workflows.

cacology:
Thanks for the quick response! I ran Mendeley and took an entry round-trip, that's deliberately a little ugly (don't judge--I wanted to see what would break ;-):

Imported:
```BibTeX
@inbook{GoldsmithAugustan,
  author =       "Oliver Goldsmith",
  title =        "An Account of the {A}ugustan {A}ge of {E}ngland",
  volume =       4,
  pages =        "283-292",
  booktitle =    "The Works of Oliver Goldsmith: With an Account of
                  His Life and Writings",
  edition =      "new ed.",
  address =      "London",
  publisher =    "J. Johnson, etc.",
  year =         1806,
  origlocation = {London},
  origdate = {1759},
  note =         "from \textit{The Bee}"
}
```

Exported:
```BibTeX
@incollection{Goldsmith1806,
address = {London},
annote = {from The Bee},
author = {Goldsmith, Oliver},
booktitle = {The Works of Oliver Goldsmith: With an Account of
His Life and Writings},
edition = {new ed.},
pages = {283--292},
publisher = {J. Johnson, etc.},
title = {{An Account of the Augustan Age of England}},
volume = {4},
year = {1806}
}
```

The UI says that it imports BibTeX, not BibLaTeX, which seems to be happening here.  BibTeX has a much vaguer definition of note, so Mendeley puts it in something better:

> **note** Any additional information that can help the reader. The first word should be capitalized.

Also from BibTeX
> **annote** An annotation. It is not used by the standard bibliography styles, but may be used by others that produce an annotated bibliography. [Using BibTeX](https://ctan.org/pkg/bibtex)

So since Mendeley neither ingests nor exports BibLaTeX, but BibTeX,  I don't think `note` fields should hit this CSL?  I'm not quite sure how Mendeley uses the CSL files, but if it's converting `note` to `annote` the problem is avoided.

Thoughts?

adam3smith:
> Maybe the thing to do is clarify the relationship between CSL and BibLaTeX?

Yeah, I think you got on the wrong track on this because of pandoc's ability to read BibLaTex and use it for CSL styles (by converting it to CSL JSON internally). There is no relationship at all between CSL/CSL JSON and BibLaTeX. They are completely separate bibliography and metadata formats and the metadata component can be, with some loss, translated into each other just like any other metadata formats. 

E.g. you're conversion exercise using Mendeley is completely irrelevant. What matters is the CSL JSON that Mendeley uses to generate citation styles, and that does, indeed, convert the content of "Notes" into CSL `note` -- so I'm sorry, but this solution won't work for CSL.

PR #4579, opened 2020-02-25
Update mcgill-en.csl - include render-legislation macro

gspanglett:
Initial checkin did not include render-legislation macro, now included.


rmzelle:
Thanks!

PR #4582, opened 2020-02-26
Update mercator-institut-fur-sprachforderung-und-deutsch-als-zweitspr…

ltwente:
...ache.csl

changed bibliography features: entry-spacing (0) and line-spacing (1) according to user requests.


rmzelle:
Thanks!

PR #4584, opened 2020-02-27
Update vodohospodarske-technicko-ekonomicke-informace.csl

libora:
Use of "et al." is not defined in the guidance for authors for this journal. The journal uses ČSN ISO 690 standard, but the standard is not strict in this and use of "et al." is not mandatory (can be published all authors OR only first three and "et al."). Published articles used all authors so I prepare this change of style. Also, Zotero style for ČSN ISO 690 does not use "et al.".


rmzelle:
Thanks!

PR #4586, opened 2020-02-28
Create archeologicke-rozhledy.csl

Minesurfer97:
I've made some rather cosmetic changes to Natures Sciences Sociétés template and it matches citation style of Archeologické rozhledy journal now.

### DIFF
--- a/archeologicke-rozhledy.csl
+++ b/archeologicke-rozhledy.csl
@@ -12,7 +12,8 @@
       <name>David Spacil</name>
     </author>
     <category citation-format="author-date"/>
-    <category field="science"/>
+    <category field="anthropology"/>
+    <category field="history"/>
     <issn>0323-1267</issn>
     <eissn>2570-9151</eissn>
     <updated>2020-02-28T06:25:16+00:00</updated>
--- a/archeologicke-rozhledy.csl
+++ b/archeologicke-rozhledy.csl
@@ -1,0 +1,301 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+  <!-- This style was edited with the Visual CSL Editor (https://editor.citationstyles.org/visualEditor/) -->
+  <info>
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks! I assume this style (which is currently fixed to British English) is only for English manuscripts? http://www.arup.cas.cz/?cat=227 says the journals publishes in Czech, Slovak, English and German.

If so, we would typically change the style title to `<title>Archeologické rozhledy (English)</title>`.

Minesurfer97:
I'm sorry but I'm not sure I understand. Is the style usable for one language only? The journal is mainly in czech. Only a few articles are in different languages.

rmzelle:
@Minesurfer97, CSL styles can behave in one two ways:

1. They can be fixed to a single language
2. They can automatically change their language based on the user's settings

The style you've submitted here uses option 1 and is fixed to British English (`default-locale="en-GB"`), so it will always use English translations.

The Czech instructions give the following examples (with words that CSL typically translates in bold):

> … tyto výsledky (Novák **et al.** 1994, 199-200) … archeologický výzkum (Bílá 2002, 38; 2006, 45; Bílá – Novák 1985, 56-59; Olivier – See **eds.** 2001).
> Olivier, L. – See, E. **eds.** 2001: La nécropole de tumulus du Bois de Voivre à Haroué (Meurthe-et-Moselle). Paris: CNRS.
Peck, M. 1976: Early Bronze Age settlement in Poland. **In**: E. White **ed.**, Bronze Age in Europe, London: Routledge, 11–52.

Maybe the journal uses the same translations ("in", "et al.", "ed.", "eds.") for articles in any language, in which case we can leave things as they are. I couldn't tell from the instructions if this is the case.

If these terms do need to change depending on the language of the article, we can either clarify in the CSL style title that it's locked to a particular language, or remove the `default-locale` from the style so the CSL style automatically adjusts its language (option 2).

Minesurfer97:
@rmzelle Thank you for explanation, I'm far from being an expert. The terms don't need to change based on author (article) language. This isn't even that strict. For example "et al." and its czech equivalent "a kol." are being interchanged at random. So I think it's fine if it is fixed on one language only. :)

rmzelle:
Okay, thanks! We'll just take it like this, then.

PR #4588, opened 2020-03-02

POBrien333:
…cs-today.csl

small fix via https://forums.zotero.org/discussion/comment/347477/#Comment_347477


adam3smith:
Looks like it's working correctly, thanks!

PR #4589, opened 2020-03-02
Create pediatric-infectious-disease-journal.csl

morksoldat:
Based on AMA style, minor adjustements from american-medical-association.csl
Journal site https://journals.lww.com/pidj/pages/default.aspx , citattion style as indicated in http://edmgr.ovid.com/pidj/accounts/ifauth.htm


adam3smith:
Thanks!

PR #4591, opened 2020-03-03
Create laboratory-animal-science-professional.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/349964#Comment_349964

### DIFF
--- a/laboratory-animal-science-professional.csl
+++ b/laboratory-animal-science-professional.csl
@@ -100,7 +100,7 @@
   </macro>
   <macro name="pages">
     <group>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks! I don't think the discrepancy between style guide & journal is quite as bad as you made it sound (though it really isn't CSE in any meaningful sense).

POBrien333:
Gotta exaggerate a bit, so they get their s*!t together. ;)

adam3smith:
Thanks!

PR #4592, opened 2020-03-03
Create apa-doi-arxiv-extra.csl

andreaiorio:
A style for those who uses the extra field to save the arXiv ID  publication. It prints out both the DOI url and the arxiv one.


adam3smith:
Thanks -- we generally don't host styles that use custom hacks to add information such as unmapped IDs etc. I completely understand how this is can be useful and you're (obviously -- not that you'd need our permission) welcome to host this yourself, but it creates too much confusion on the repository (not least, because CSL doesn't just serve Zotero but almost 50 other products)

PR #4595, opened 2020-03-05
journal title in italics in biophysical-journal.csl

POBrien333:
Via https://forums.zotero.org/discussion/81748/downloaded-biophysical-journal-citation-style-does-not-italicize-journal-titles


rmzelle:
Thanks!

PR #4596, opened 2020-03-05
Always show DOI for article-journal for journal-of-forensic-sciences.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/350125/#Comment_350125


adam3smith:
Thanks!

PR #4597, opened 2020-03-05
Update geistes-und-kulturwissenschaften-heilmann.csl

tillheilmann:
Refactored the code and added support for more types and variables.


rmzelle:
Thanks!

PR #4598, opened 2020-03-05

PriscilaSJUFRGS:
Update with correction for some document types.


rmzelle:
Thanks!

PR #4609, opened 2020-03-07
Create jiaptr.csl

Jesse-Heo:


jiaptr.csl
```diff
@@ -0,0 +1,266 @@
...
+  <info>
+    <title>JIAPTR(Journal of International Academy of Physical Therapy Research)</title>
+    <title-short>JIAPTR</title-short>
+    <id>http://www.zotero.org/styles/jiaptr</id>
+    <link href="http://www.zotero.org/styles/jiaptr" rel="self"/>
+    <link href="https://westlibrary.txwes.edu/sites/default/files/pdf/AMACitationStyle.pdf" rel="documentation"/>
+    <category citation-format="numeric"/>
+    <category field="medicine"/>
+    <issnl>2092-8475</issnl>
+    <summary>Same as the American Medical Association style as used in JAMA.</summary>
```
rmzelle: @Jesse-Heo, so this style is the same as https://www.zotero.org/styles?q=id%3Aamerican-medical-association?

Jesse-Heo: yes, exactly, thank you for the help!

Jesse-Heo: I was working on JIAPTR's style  in the Zotero Style Repository(https://www.zotero.org/styles) to make it easy for researchers to understand the paper and it was very difficult for me because it was my first time.

POBrien333:
In that case, you can just make a dependent style. See this for example:
https://github.com/citation-style-language/styles/blob/master/dependent/acta-biomaterialia.csl

Adapt your style to this by using the link to ama as your independent-parent.
https://github.com/citation-style-language/styles/blob/master/dependent/4or.csl

rmzelle:
@Jesse-Heo, thanks. I just changed this style to a "dependent" type that's just an alias for the standard AMA style.

(also, for the future, note that you can just update an existing pull request by following the instructions provided by @csl-bot above, so there is no need to create a new one every time an automated test fails: #4603, #4604, #4605, #4606, #4607, #4608)

PR #4614, opened 2020-03-09
Fix chapter for american-institute-of-aeronautics-and-astronautics.csl

POBrien333:
via https://forums.zotero.org/discussion/81825/citing-book-sections-in-aiaa-format-but-zotero-doesnt-include-year


rmzelle:
Thanks!

PR #4615, opened 2020-03-09
Remove et-al settings from american-chemical-society.csl

POBrien333:
As discussed here. Remove et-al from acs.csl. https://forums.zotero.org/discussion/comment/333791/#Comment_333791

Editors are coming back to writers again and again that they don't want any et-al in papers.

### DIFF
--- a/american-chemical-society.csl
+++ b/american-chemical-society.csl
@@ -145,7 +145,7 @@
       <text variable="citation-number"/>
     </layout>
   </citation>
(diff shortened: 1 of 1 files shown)


adam3smith:
@rmzelle I suggest we change this as proposed, but would be good to check with ACS

rmzelle:
https://pubs.acs.org/doi/pdf/10.1021/bk-2006-STYG.ch014 says

> Author Name Field
> Include all author names in a reference citation.

So let's just do this.

PR #4617, opened 2020-03-09
Create rivista-trimestrale-di-sanita-pubblica-veterinaria.csl

POBrien333:
Crappy template style. You might run into some artefacts.

via https://forums.zotero.org/discussion/79895/style-request-veterinaria-italiana-rivista-trimestrale-di-sanita-pubblica-veterinaria


rmzelle:
Thanks! (Rivista trimestrale di Sanità Pubblica Veterinaria seems to be the subtitle of this journal)

PR #4620, opened 2020-03-10
Create aporia.csl

POBrien333:
via https://forums.zotero.org/discussion/81846/how-to-write-the-full-name-of-the-journal-in-the-list-of-references#latest

Typeset papers: https://uottawa.scholarsportal.info/ottawa/index.php/aporia/issue/view/513/294

The Vancouver styles with all that [Internet] stuff always throw me off.


adam3smith:
Let's call this `Aporia: The Nursing Journal` -- there's also a philosophy journal called Aporia

rmzelle:
Thanks!

PR #4622, opened 2020-03-11
Update harvard-university-of-westminster.csl

POBrien333:
Reworked style, as per https://forums.zotero.org/discussion/81860/same-author-same-year-different-publications


rmzelle:
Thanks!

PR #4623, opened 2020-03-11
Update instituto-de-pesquisas-energeticas-e-nucleares.csl

amkawai:
Changed locators for type book (added chapter between volume and page numbers);
Changed volume macro suffix from dot to comma;
Changed order for type book section (added number between edition and publisher);
Changed edition macro (added comma after dot at "ed.").


rmzelle:
Thanks!

amkawai:
Thanks, Rintze!

PR #4624, opened 2020-03-11
Fix translator and accessed date for transversalites.csl

ddanangaw:
Update for the placement of translator and web page's access date. Thanks for your help.


rmzelle:
Thanks!

PR #4630, opened 2020-03-15
Create journal-of-breast-cancer.csl

POBrien333:
via https://forums.zotero.org/discussion/81027/new-style-request-journal-of-breast-cancer-j-breast-cancer


rmzelle:
Thanks!

POBrien333:
That was quick. Looks like you're already in lockdown conditions?

rmzelle:
Yes, pretty much (although living in rural Vermont we're always already quite isolated). Take care!

PR #4631, opened 2020-03-18
Create the-journal-of-trauma-and-acute-care-surgery.csl

POBrien333:
via https://forums.zotero.org/discussion/81964/journal-of-trauma-and-acute-care-surgery-jtacs


adam3smith:
Thanks!

PR #4632, opened 2020-03-19
Update unified-style-sheet-for-linguistics.csl for "in" in chapters

POBrien333:
as per https://forums.zotero.org/discussion/75241/missing-in-if-no-editors-unified-style-sheet-for-linguistics#latest

Please double check thoroughly. :)


adam3smith:
Thanks -- that's an elegant solution.

POBrien333:
Awesome!! :))

PR #4633, opened 2020-03-21
Scientific Reports style format update

austin-curnutt:
The journal reference format has changed since the previous style was uploaded in 2014. I created this based on a Nature style to fit the current requirements.

### DIFF
--- a/scientific-reports-updated
+++ b/scientific-reports-updated
@@ -1,0 +1,109 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+  <info>
+    <title>Scientific Reports</title>
+    <id>http://www.zotero.org/styles/scientific-reports-updated</id>
+    <link href="http://www.zotero.org/styles/scientific-reports-updated" rel="self"/>
+    <link href="http://www.zotero.org/styles/nature-no-superscript" rel="template"/>
+    <link href="https://www.nature.com/srep/author-instructions/submission-guidelines#references" rel="documentation"/>
+    <link href="http://fastart.nature.com/srep/Checklist_revised__submissions.pdf" rel="documentation"/>
+    <author>
+      <name>Austin Curnutt</name>
+      <email>austin.curnutt@alumni.ou.edu</email>
+    </author>
+    <category citation-format="numeric"/>
+    <category field="science"/>
+    <category field="generic-base"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
How does this differ from one of the regular Nature styles (don't worry about the Travis failure for now)

austin-curnutt:
It's for Scientific Reports, which is a Nature journal. I used the nature style as the template, but it's for Scientific Reports to match the current formatting guidelines (see the documents in the info section).

Austin Curnutt

adam3smith:
Right, I know what Scientific Reports is; I'm wondering how this is different from regular Nature style. The current CSL style for Scientific Reports just links to that and I think that's still the right approach given

> Scientific Reports uses standard Nature referencing style. 

If there are problems with the Nature style (which is totally possible) we can fix those there.

austin-curnutt:
The PDF file with guidance for references describes a different format than what is currently available from the Scientific Reports style in the repository. The guidance there asks for bracketed references and the DOI in the bibliography, which the current style does not have.

Austin Curnutt

adam3smith:
It doesn't ask for DOIs, no -- those are only for online only journals and datasets (note the Schott et al. example in both PDF and HTML instructions which as no DOI).

The square brackets are weird -- let me see if we can get some guidance from Springer Nature on that -- it's odd that journals with the same style in print would want something different in for submission (Scientific Reports uses superscripts for references in print).

austin-curnutt:
I see that now about the DOI. I noticed that point about the brackets when reading some of their other papers, but made the edit for ours to comply with that guidance in the revision document.

I’m working on submitting a paper now, so thank you for pointing that out about the DOI requirement.

adam3smith:
If you're submitting anyway, would you mind asking the managing editor about square brackets vs. superscript? If they're fine with superscript, we could leave this linked to Nature style; if they really want brackets, we need a separate style (i.e. yours) which is of course possible, but much less convenient.

austin-curnutt:
Definitely. We should be submitting tonight or tomorrow so I’ll have the submitting author include that question.

rmzelle:
@austin-curnutt, did you ever hear back?

austin-curnutt:
Hello,

I’m checking with the corresponding author to see how they accepted the paper formatting. I’ll let you know when I hear back.

austin-curnutt:
I just saw the proof for the paper. The in-line citations for Scientific Reports were done as superscripts and the references were done using numbers in the format: “1.”

adam3smith:
Didn't realized this actually got answered fully -- thanks for getting back to us & good to know we can leave things as they are

PR #4636, opened 2020-03-23
Update URL styling for ieee.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/351282#Comment_351282
and https://forums.zotero.org/discussion/80776/ieee-with-url


rmzelle:
Thanks!

PR #4645, opened 2020-03-27
Create journal-of-urban-technology.csl

POBrien333:
via https://forums.zotero.org/discussion/82080/citation-style-request-for-journal-of-urban-technology

### DIFF
--- a/journal-of-urban-technology.csl
+++ b/journal-of-urban-technology.csl
@@ -183,22 +183,6 @@
       </if>
     </choose>
   </macro>
-  <macro name="issued">
-    <choose>
-      <if type="webpage" match="none">
-        <choose>
-          <if variable="issued">
-            <date variable="issued">
-              <date-part name="year"/>
-            </date>
-          </if>
-          <else>
-            <text term="no date" form="short"/>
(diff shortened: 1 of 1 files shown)


journal-of-urban-technology.csl
```diff
@@ -0,0 +1,318 @@
...
+      <name>Patrick O'Brien</name>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="engineering"/>
+    <issn>1063-0732</issn>
+    <eissn>1466-1853</eissn>
+    <updated>2020-03-27T15:30:17+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="en-GB">
```
rmzelle: @POBrien333, can you check the locale section here? If the style gives correct output, I assume this locale section can be removed as it's specific to "en-GB", while the style is locked to "en-US".

rmzelle:
Thanks!

PR #4650, opened 2020-03-29
Create geografia-fisica-e-dinamica-quaternaria.csl

RicceR-unipi:
We upload the citation style file for the journal "[Geografia Fisica e Dinamica Quaternaria](http://gfdq.glaciologia.it/)"

### DIFF
--- a/geografia-fisica-e-dinamica-quaternaria.csl
+++ b/geografia-fisica-e-dinamica-quaternaria.csl
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
-  <!-- This style was edited with the Visual CSL Editor (https://editor.citationstyles.org/visualEditor/) -->
+  <!-- Polyglot; journal publishes in English and Italian -->
   <info>
     <title>Geografia Fisica e Dinamica Quaternaria</title>
     <title-short>GFDQ</title-short>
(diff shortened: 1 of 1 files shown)


rmzelle:

And do you know if it's common for this journal to have articles in languages other than English? http://gfdq.glaciologia.it/ says "The language of the journal is English, but papers can be written in other main scientific languages". We typically fix styles to a single locale, but not if they're for bilingual/polyglot journals. See also https://github.com/citation-style-language/styles/wiki/Style-Requirements#3---style-locale.

RicceR-unipi:
In my knowledge the new articles should be in English but the authors could add abstract and keyword in the language of the area where the research took place if they desire: "In addition the authors could insert an abstract and key words in the language of the Country in which the work has been carried out." ([Typing section](http://gfdq.glaciologia.it/submitting/)). However, once a year the Journal publishes the bulletin of the Italian glaciers both in English and Italian this is why I did not set the style to a single locale.

rmzelle:
> About the first point, I wrote the style by myself following the rules of the Journal and the instruction of the Editor-in-Chief, so I am relatively sure when I say that, in my knowledge, it is not based upon other existing style.

So you wrote the CSL style from scratch? Or did you start from another CSL style? (just in case I wasn't clear, we're curious whether this CSL style was based on an existing one. Our question isn't about the origin of the style guide of the journal itself)

RicceR-unipi:
> So you wrote the CSL style from scratch? Or did you start from another CSL style? (just in case I wasn't clear, we're curious whether this CSL style was based on an existing one. Our question isn't about the origin of the style guide of the journal itself)

Yes, I know. I looked some others style trying to understand and to learn the CSL code and then I wrote the style from the scratch using the visual editor. Probably it could be written better or more stylishly, but in our experience and in our knoledge it works well. If I had based this CSL on another style I would have no problem to set it in the template. The biggest inspiration that I took from other CSL are the names of the macros

rmzelle:
Okay, thanks. Just making sure I understood you correctly.

And thanks for the style submission!

RicceR-unipi:
Thank you for your replies and your effort to improve the style! It was a pleasure submit it to the repository!

PR #4655, opened 2020-03-30
Correct Acta Palaeontologica Polonica in-text citation format

benjaminmoon:
Modifications to in-text citation style to correct rendering of year-suffix and page references:
e.g. (Abel 2010*a*, *b*)
(Abel 2010: 544)


rmzelle:
Thanks!

PR #4658, opened 2020-03-31
Update freshwater-science.csl

POBrien333:
via https://forums.zotero.org/discussion/82205/style-correction-freshwater-science


POBrien333:
Can you guys help me here?
Book titles should be rendered with a capital for the first letter of a subtitle. I added the text-case="sentence" for this, but it seems I can't do this?
Any help would be appreciated :)

rmzelle:
> Can you guys help me here?
> Book titles should be rendered with a capital for the first letter of a subtitle. I added the text-case="sentence" for this, but it seems I can't do this?

Pretty sure CSL 1.0.1 doesn't have a concept of subtitles, so I don't think you can do anything in the style itself to address this. (although citeproc-js might have implemented some heuristics to automatically reformat subtitles; not sure)

PR #4661, opened 2020-04-01
Create optics-letters.csl

POBrien333:
via https://forums.zotero.org/discussion/82127/style-request-optics-letters#latest

Optics Letters uses a shorter version of the the-opitical-society.csl style.


rmzelle:
Thanks!

PR #4664, opened 2020-04-02
Delete dependent/biomedical-optics-express.csl and optical-materials-express.csl (for automatic style)

POBrien333:
Delete, to be replaced via automatically generated independent styles via https://github.com/citation-style-language/journals/pull/93


rmzelle:
@POBrien333, this isn't necessary, and the other PR has automatically overwritten these existing dependent styles.

POBrien333:
Ahhh. That's good to know. :)

PR #4665, opened 2020-04-02

ibadger:
Just fixing two errors:

editor macro wasn't being used

article-journal was missing from the volume, issue, page variable. Meaning the journal title wouldn't display if this info was missing.

If you need any further info let me know, not a rush at all with everything that is going on, just have more time as working from home


rmzelle:
Thanks!

PR #4667, opened 2020-04-03
Create proceedings-of-the-joint-international-grassland-and-international-rangeland-congress-2021.csl

cmoreno57:
The International Grassland Congress 2020 received over a 1,000 abstracts that will need their corresponding papers. The organizers provided a file template with examples of citation. However, there were not a matching citation style in the repository. I created it by editing an existing one that matched 94%.


rmzelle:
@cmoreno57, thanks! It looks like the conference has been postponed to 2021. And it also looks like the same format is used by the IGC and IRC.

Maybe we can name the style "Proceedings of the Joint International Grassland & International Rangeland Congress 2021"?

cmoreno57:
Hi! Yes, I didn't notice they already changed the name.
Sure, the name "Proceedings of the Joint International Grassland & International Rangeland Congress 2021" and the short name "IGC-IRC2021" should be more appropriate. Agreed with the edits.

rmzelle:
Thanks!

PR #4669, opened 2020-04-04
Update iso690-author-date-sk.csl

tferianc:
1. changed bibliography date format to shorter numeric
2. changed edition format from "2nd. vyd" to "2. vyd" which is much better in slovak language
3. removed comments


rmzelle:
Thanks!

PR #4674, opened 2020-04-06

POBrien333:
via https://forums.zotero.org/discussion/82337/help-amending-citation-presets-please


rmzelle:
Thanks!

PR #4675, opened 2020-04-07

POBrien333:
To undo #4674 
I mixed up the journals. equine-veterinary-journal.csl (which we have) has a numeric style. equinve-veterinary-education.csl has an author-date style that has the bibliographic style as evj.csl otherwise.

via https://forums.zotero.org/discussion/82337/help-amending-citation-presets-please#latest


rmzelle:
Thanks!

PR #4678, opened 2020-04-08
Create sylwan.csl

wkedziora:
Hi Guys!

I have created new style called "sylwan" that is used in journal of this name. I have based on APA style. I would like somebody more experienced to check it up, because I am new to this stuff and unsure about the quality. I would love to receive feedback and improve my work.

Regards

### DIFF
--- a/sylwan.csl
+++ b/sylwan.csl
@@ -565,7 +565,7 @@
       </if>
     </choose>
   </macro>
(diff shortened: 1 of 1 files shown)


rmzelle:
@wkedziora, thanks for your submission!

We don't have the manpower here to review all submissions in depth, but I didn't spot any significant issues with the style. I just made a few small tweaks to the style metadata (title and categories), and removed the English localization section in the style which will never be used in a Polish style.

wkedziora:
Thanks a lot! I am glad I have helped.

PR #4679, opened 2020-04-08
Create fhk-school-of-management.csl

stromi65:
new file


stromi65:
class="in-text"

rmzelle:
@stromi65, thanks for your submission. Is there an online style guide available for this style? Per https://github.com/citation-style-language/styles/wiki/Style-Requirements#10---documentation-link, we typically require each CSL style to include a link to the format it describes.

adam3smith:
@stromi65 -- completely understand that time may be particularly short in supply right now, but if you could briefly reply to @rmzelle 's question above about documentation?

PR #4680, opened 2020-04-09
Update trabajos-de-prehistoria.csl

vjjaimez:
Hi guys! I created this style a week ago or so, but I've discovered a couple of bugs. Inline citations where the same author(s) had two or more publications cited were not properly contracted. [e.g. Balsera et al. 2015; Balsera et al. 2016] when it should be [Balsera et al. 2015, 2016].
The other bug was in the sorting order in the bibliography. References are now ordered as they should: by name and, in the case of authors with more than one reference, by name and year-suffix.


rmzelle:
Thanks!

PR #4682, opened 2020-04-09
Update archeosciences.csl

nfrerebeau:
Changes:

* Fix inline citation sort order: first issued, then author;
* Use italic only for book titles;
* Add version for reports and datasets.

Tested against [CSL Validator](https://validator.citationstyles.org/).


rmzelle:
Thanks!

PR #4686, opened 2020-04-10
Create iso690-full-note-cs-pravnik.csl

OTristanF:
Created a style suitable for Pravnik – a journal published by the Institute of State and Law of the Czech Academy of Sciences


rmzelle:
Thanks!

PR #4690, opened 2020-04-11
Update pravnik.csl

OTristanF:
1. Added missing space in the prefix of (ed.)/(eds).
2. Removed the publisher's info of journal articles.
3. Removed the ISSN from the bibliography. (Although the bibliography itself is not required in the Pravnik journal.)


rmzelle:
Thanks!

PR #4694, opened 2020-04-11
Create iso690-author-date-pt

andreappel:
Proposing the creation of ISO 690 Brazilian Portuguese variation.
Took the opportunity to:
- implement et-al which was not available on the original/template;
- remove the hard-encoded spaces and block entries;

I was not sure if the <id> should include "pt-br" or just "pt", so I've included pt only.


rmzelle:
Thanks!

PR #4695, opened 2020-04-12
add DOI to mdpi.csl back

POBrien333:
via https://forums.zotero.org/discussion/82441/adding-doi-in-bibliography-of-code-in-csl-for-journal-named-sustainability


rmzelle:
Thanks!

PR #4698, opened 2020-04-13
Request for review and assessment of the repa.csl

repajp:
Dear reviewers,

After several attempts, hopefully would like to submit the new version of the repa.csl for review and validation. 
Thank you in advance for your time and for your review.
Note: The proposed style will be applied for 6-journals and also conference proceedings publications.

### DIFF
--- a/repa.csl
+++ b/repa.csl
@@ -13,7 +13,7 @@
     </author>
     <category citation-format="numeric"/>
     <category field="science"/>
-    <updated>2020-04-12T21:09:30+00:00</updated>
+    <updated>2020-04-13T07:17:55+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="editor">
@@ -24,7 +24,7 @@
   </macro>
   <macro name="author">
     <names variable="author">
-      <name delimiter=". " delimiter-precedes-last="always" et-al-min="19" initialize-with="" name-as-sort-order="all" sort-separator=" "/>
+      <name delimiter=". " delimiter-precedes-last="always" initialize-with="" name-as-sort-order="all" sort-separator=" "/>
       <label/>
       <substitute>
         <names variable="editor" suffix=" "/>
--- a/repa.csl
+++ b/repa.csl
@@ -3,9 +3,10 @@
   <!-- This style was edited with the Visual CSL Editor (https://editor.citationstyles.org/visualEditor/) -->
   <info>
     <title>REPA</title>
+    <title-short>Research and Education Promotion Association (REPA)</title-short>
     <id>http://www.zotero.org/styles/repa</id>
     <link rel="self" href="http://www.zotero.org/styles/repa"/>
-    <link rel="documentation" href="https://repa.jp/"/>
+    <link rel="documentation" href="https://repa.jp/journals/editorial-guide-and-citation-style/"/>
     <author>
       <name>Research and Education Promotion Association (REPA)</name>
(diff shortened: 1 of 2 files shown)


repajp:
Many thanks for your kind support and good news :). Since three days we are working on it.

rmzelle:
> The proposed style will be applied for 6-journals and also conference proceedings publications.

@repajp, could you point us to the "instructions to authors" webpages for these 6 journals and, if one exists, the conference proceedings?

repajp:
Sure we will do so, now the REPA journal websites are under maintenance and shortly will be online. So, what is the best option to point the CSL? Is there any standard quotation "    " ? 
Also, still the repa citation is not accessible through Zotero.

repajp:
Sure we will do so, now the REPA journal websites are under maintenance and shortly will be online. So, what is the best option to point the CSL? Is there any standard quotation " " ?
Also, still the repa citation is not accessible through Zotero.

rmzelle:
> Sure we will do so, now the REPA journal websites are under maintenance and shortly will be online. So, what is the best option to point the CSL? Is there any standard quotation " " ?

Some software products like Mendeley have their own interface for accessing our repository styles, but you can always point users to your style(s) in the Zotero Style Repository (e.g. https://www.zotero.org/styles?q=id%3Aapa).

> Also, still the repa citation is not accessible through Zotero.

We haven't accepted your CSL style yet, as I was hoping you could first link to your online style guide. Per https://github.com/citation-style-language/styles/wiki/Style-Requirements#10---documentation-link, we typically require each CSL style to include a link to the format it describes.

repajp:
I went through the style guide several times :) still, our journals' websites are public. However, please find the linked of author instructions that included and already mentioned citation use:
https://repa.jp/journals/paper-submission-instructions/

rmzelle:

rmzelle:
And if you can give us a spreadsheet of journal titles, and print/online ISSNs, we can also add 'dependent' styles for these. (see https://github.com/citation-style-language/journals/blob/master/apa/_journals.tab for an example)

repajp:
Dear Zelle,
Thank you for your cooperation. 
This is a standalone citation style based on the unique style such as grouping the vol. no. pp. and etc. without any similarity with existing styles in the same manner

repajp:
We already applied for the ISSN for our upcoming 6-journals, may we request your cooperation on the subject matter later after receiving the ISSNs

rmzelle:
> This is a standalone citation style based on the unique style such as grouping the vol. no. pp. and etc. without any similarity with existing styles in the same manner

Just to make sure we understand each other correctly: I was asking whether you based the CSL style file on an existing CSL style, not whether the REPA citation format is derived from something else. Did you write this CSL style from scratch, or did you start from an existing style?

rmzelle:
> We already applied for the ISSN for our upcoming 6-journals, may we request your cooperation on the subject matter later after receiving the ISSNs

Yes. If the ISSNs are not yet available, we can just accept the REPA CSL style by itself. You can create a new GitHub issue once you have the ISSNs and we'll generate the dependent CSL styles for the journals.

repajp:
For better understanding, our team inspired and analyzed existing styles especially (APA, IEEE, etc.) and in light (observing the structures, establishing relations, variables, and etc.) create a new style, confidently can claim that the REPA style is unique without similarity with the available styles. We think, as per the copyright, we do not imply to acknowledge or be dependent when we had a general observation in the public domain.  
For the ISSNs, we will do as per your advice, thank you.

rmzelle:
Thanks!

repajp:
Dear Zelle, thank you very much for your great cooperation towards the fulfillment of the endeavor. Take care, and stay safe, bye.

PR #4701, opened 2020-04-13
Change apa-fr-provost.csl to fr-CA locale

bwiernik:
Conform to fr-CA locale standards rather than fr-FR per the Provost guide.
https://forums.zotero.org/discussion/82278/apa-6th-edition-provost-french-canada-style#latest

(Also update title-short for apa-tr to be consistent with other APA styles.)


rmzelle:
Thanks!

PR #4703, opened 2020-04-13
Create isnad-dipnotlu.csl

osmndrmz:
The ISNAD Citation Style 2nd edition full notes and bibliography format.

### DIFF
--- a/isnad-dipnotlu.csl
+++ b/isnad-dipnotlu.csl
@@ -4,7 +4,7 @@
     <title>İSNAD Atıf Sistemi 2. Edisyon (dipnotlu)</title>
     <title-short>İSNAD</title-short>
     <id>http://www.zotero.org/styles/isnad-dipnotlu</id>
-    <link href="http://www.zotero.org/styles/isnad-note-bibliography" rel="self"/>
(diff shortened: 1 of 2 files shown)


rmzelle:
@osmndrmz, thanks. Do I understand correctly that this style can be used in both Turkish and English?

Would a Turkish style title ("İSNAD Atıf Sistemi 2. Edisyon") make more sense here? And is "İSNAD" an abbreviation of something?

osmndrmz:
You are right. This style is multilingual but it is more convenient to use Turkish style title. "İSNAD" is not an abbreviation of anything. İsnad, as an Arabic originated word, means “to convey the word to the one who says it and to inform the way of the coming of a rumor”. The word “sanad”, which comes from the same root as the word “isnad”, and as in the question “What is the sanad of your word?” is used to question and indicate the source of the word. Becuse of this, "İSNAD" (with capital letters) was chosen as a brand name for this style.

osmndrmz:
> @osmndrmz, thanks. Do I understand correctly that this style can be used in both Turkish and English?
> 
> Would a Turkish style title ("İSNAD Atıf Sistemi 2. Edisyon") make more sense here? And is "İSNAD" an abbreviation of something?

You are right.

rmzelle:
Thanks!

PR #4704, opened 2020-04-13
Create isnad-metinici.csl

osmndrmz:
The ISNAD Citation Style 2nd edition author date format.


rmzelle:
Thanks!

(companion of #4703)

PR #4705, opened 2020-04-13
Update pravnik.csl, masarykova-univerzita-pravnicka-fakulta.csl and iso690-full-note-cs.csl

OTristanF:
Since "pravnik.csl" and "masarykova-univerzita-pravnicka-fakulta.csl" has "iso690-full-note-cs.csl" as a template, I am opening a pull request for all three of them. For "pravnik.csl" and "iso690-full-note-cs.csl", I made sure that it ends with "." even without a locator. At the top of that, for all the three styles, I made the book formatting as default (by moving it from if to else branch), which allows citing more item types.


rmzelle:
Thanks!

PR #4708, opened 2020-04-15
remove space before in-text citation in le-tapuscrit-author-date.csl

POBrien333:
via https://forums.zotero.org/discussion/82514/personalized-style-space-added-before-the-quotation#latest

Not sure why there is a space there. If there is a reason for it, please close.


rmzelle:
Thanks!

(adding spaces before author-date citations makes it quicker to switch between author-date and numeric styles, but we historically haven't been doing this for repository styles, so it's better to remove it here)

PR #4710, opened 2020-04-17
Small fix for society-for-american-archaeology.csl

POBrien333:
via https://forums.zotero.org/discussion/82550/error-year-showing-up-at-beginning-and-end-of-citation


rmzelle:
Thanks!

PR #4711, opened 2020-04-17
Create the-open-university-harvard-s390-modules.csl

neallyon:
This style is based on "The Open University - Harvard" (the-open-university-harvard.csl) style and has been adjusted to meet the specific requirements of the S390 project modules. Specifically, the in-text citations are limited to 3 authors max, with 'et-al' in italic font. In the bibliography, the authors are limited to 10 authors and 'et-al' in italic font.  Credit is given to original authors through the "template" reference as per the style requirements here: https://github.com/citation-style-language/styles/wiki/Style-Requirements.

### DIFF
--- a/the-open-university-harvard-s390-modules.csl
+++ b/the-open-university-harvard-s390-modules.csl
@@ -1,0 +1,263 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+  <info>
+    <title>The Open University - Harvard - S390 Modules</title>
+    <id>http://www.zotero.org/styles/the-open-university-harvard-s390-modules</id>
+    <link href="http://www.zotero.org/styles/the-open-university-harvard-s390-modules" rel="self"/>
+    <link href="http://www.zotero.org/styles/the-open-university-harvard" rel="template"/>
+    <link href="https://learn1.open.ac.uk/mod/oucontent/view.php?id=2620" rel="documentation"/>
+    <author>
+      <name>Neal Lyon</name>
+      <email>neallyon34@gmail.com/</email>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="generic-base"/>
+    <summary>The Open University Harvard author-date style-this version includes et al font style in italics and bibliography limited to 10 (ten) authors-for use with S390 modules</summary>
+    <updated>2020-03-26T21:53:41+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <macro name="editor">
+    <names variable="editor" delimiter=", ">
+      <name and="text" initialize-with=". " delimiter=", " sort-separator=", " name-as-sort-order="all"/>
+      <label form="short" prefix=" (" suffix=")" strip-periods="true"/>
+    </names>
+  </macro>
+  <macro name="anon">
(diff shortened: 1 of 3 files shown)


adam3smith:
Thanks -- is there any way to make the description of the style publicly available (or link to a public version)? The current link requires login.

neallyon:
> Thanks -- is there any way to make the description of the style publicly available (or link to a public version)? The current link requires login.

Thanks - I think I have now set the link to a publicly available .pdf. Please check and report back if this does not work.

the-open-university-s390.csl
```diff
@@ -0,0 +1,263 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+  <info>
+    <title>The Open University - S390</title>
```
rmzelle, line 4: We could put back "Harvard" in the style title, but then we should update the other Open University course-specific Harvard variants as well (https://www.zotero.org/styles?q=open-university).

neallyon: Agree with the name change removing "Harvard" - 
(1) the "Harvard" part of the title is probably not necessary for a course-specific variant, 
(2) the variant refers to the original "Harvard"-based style brief (at line 8), and, 
(3) changing the other course styles to match these minor adjustments in presentation of the style seems overkill.
Finally, seeing as some of the other course specific variants have been around a few years, dropping "Harvard" from the title does not seem to have been an issue.

the-open-university-harvard-s390-modules.csl
```diff
@@ -5,7 +5,7 @@
     <id>http://www.zotero.org/styles/the-open-university-harvard-s390-modules</id>
     <link href="http://www.zotero.org/styles/the-open-university-harvard-s390-modules" rel="self"/>
     <link href="http://www.zotero.org/styles/the-open-university-harvard" rel="template"/>
-    <link href="https://learn1.open.ac.uk/mod/oucontent/view.php?id=2620" rel="documentation"/>
+    <link href="http://www.open.ac.uk/libraryservices/documents/Harvard_citation_hlp.pdf" rel="documentation"/>
```
rmzelle: @neallyon, just to make sure, it looks like http://www.open.ac.uk/libraryservices/documents/Harvard_citation_hlp.pdf describes the general Open University Harvard style, right? Does https://learn1.open.ac.uk/mod/oucontent/view.php?id=2620 link to the document that describes the specific requirements of S390?

neallyon: @rmzelle, 
(1) Yes, the linked pdf is a publicly available reference to the general style.
(2) No, the second link is to the same general style, only in a HTML format on a site requiring login. 

The specific referencing requirements for S390 (which is an 'umbrella term' for a small suite of science project modules of different disciplines achieving essentially the same objectives so managed under the same module rather than through the specific science faculties) are provided within module materials which are only available on a site requiring login, as far as I am aware. Unfortunately as the Open University is a pure distance-learning provider, a lot of their material is HTML-based and only available via their webpages which require login.  

These specific requirements are very minor; my adjustments simply ensure papers with more than 10 authors show one author and et al (in italics) in the citations and show only the first 10 authors and et al (in italics) in the bibliography - the original shows all authors in the bibliography and et-al is not in italics. Having adjusted the file myself last year, I have passed it to students this year, but thought it would be a good idea to make it an official .csl in the Zotero repository for future students.

adam3smith: @rmzelle  -- this sounds OK by me: will merge in ~48 hs unless you voice concerns, with a link to the above comment in the documentation.

rmzelle:
Thanks!

PR #4712, opened 2020-04-18

OTristanF:
1. The thesis on the faculty can be written in other languages, although Czech is the primary. I have received feedback that some students would want to use it for their thesis written in English. Therefore, I have removed the "default-locale" so users can choose other languages.
2. Because of the previous, I have modified the code to prevent it from using hardcoded values (such as s. for pages) instead of locale terms.
3. I have removed notes since the documentation does not permit them.
4. I have added new macros to prevent the code repetition.
5. I have modified the "thesis" type part to make the code more readable.

1–4 needs to be also done for pravnik.csl (the style can be used not just for Pravnik, but also for The Lawyer Quarterly, which is not in Czech, but in English), and 2–4 for iso690-full-note-cs.csl since it is their template style.

I don't know if the name (with "Czech" in it) should remain like that since I have removed the default-locale parameter. Students will use that primarily for their thesis written in Czech, but now, they can use it even if they decided to write the thesis in another language.

### DIFF
--- a/masarykova-univerzita-pravnicka-fakulta.csl
+++ b/masarykova-univerzita-pravnicka-fakulta.csl
@@ -14,7 +14,7 @@
     <category citation-format="note"/>
     <category field="law"/>
     <summary>Masaryk University, Faculty of Law</summary>
-    <updated>2020-04-13T10:00:00+00:00</updated>
+    <updated>2020-04-18T10:00:00+00:00</updated>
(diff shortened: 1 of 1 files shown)


rmzelle:
> I don't know if the name (with "Czech" in it) should remain like that since I have removed the default-locale parameter.

We always include the style locale in the title if it's set to a non-English locale. For styles without a fixed locale we don't include the language in the title.

rmzelle:
Thanks!

PR #4713, opened 2020-04-18
Update animal-welfare.csl

POBrien333:
via https://forums.zotero.org/discussion/82590/adding-urls-to-animal-welfare-references


rmzelle:
Thanks!

PR #4719, opened 2020-04-21
fixed bug with "near-note" position univ.-bordeaux-ecole-doctorale-de…

fflamerie:
…-droit

"near-note" position would create a bug for "legal_case" and display an error message. This has been fixed by creating specific instructions for "legal_case" in "near-note" position.

### DIFF
--- a/universite-de-bordeaux-ecole-doctorale-de-droit.csl
+++ b/universite-de-bordeaux-ecole-doctorale-de-droit.csl
@@ -661,15 +661,18 @@
         </else-if>
         <else-if position="near-note" match="any">
           <choose>
-            <if type="legal_case">
+            <if variable="author editor" match="none">
               <group delimiter=", ">
                 <text macro="authority"/>
                 <text macro="case-patent-ref"/>
+                <text macro="author"/>
                 <text macro="title-short"/>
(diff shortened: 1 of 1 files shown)


universite-de-bordeaux-ecole-doctorale-de-droit.csl
```diff
@@ -660,11 +660,26 @@
           <text term="ibid" text-case="capitalize-first" font-style="italic"/>
         </else-if>
         <else-if position="near-note" match="any">
-          <group delimiter=", ">
-            <text macro="author"/>
-            <text term="cited" font-style="italic"/>
-            <text macro="locator"/>
-          </group>
+          <choose>
+            <if type="legal_case">
```
adam3smith, line 664: while this fixes the reported issue, what about other items commonly without authors, e.g. bills, legislation, and often websites? Could you do something like
```
<else-if variable="author editor" match="none">
  <text variable="title-short"/>
</else-if>
```
(or something along those lines that works with the logic of the style)

fflamerie:
Thank you for your feedback, I have made the suggested changes.

adam3smith:
Thanks!

PR #4720, opened 2020-04-21
set strip-periods for journal container-title for the-febs-journal.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/354105/#Comment_354105


adam3smith:
Thanks!

PR #4721, opened 2020-04-21
APA: switch volume label to volume term for non-numeric volumes

bwiernik:
Accomodate some undesirable behavior in citeproc-js: https://github.com/Juris-M/citeproc-js/issues/124

If the non-numeric pluralization behavior for `label` changes in citeproc-js (or when volume-title is incorporated), we can change this back.


adam3smith:
Given the update to the citeproc-js issue, do you still want to do this?

bwiernik:
Nope. Wasn’t sure how quickly Frank would be able to get to it.

PR #4724, opened 2020-04-22
Fix comma before et al. for the-oncologist.csl

POBrien333:
via https://forums.zotero.org/discussion/82694/style-request-updating-the-oncologist


adam3smith:
FYI -- check out the changed code above @POBrien333 in case you weren't aware of that attribute. It's not super common

POBrien333:
Cheers

PR #4727, opened 2020-04-23
Update american-medical-association.csl to version 11

POBrien333:
via https://forums.zotero.org/discussion/82658/new-ama-format
https://www.amamanualofstyle.com/fileasset/AMAMOS/aaaAMWA%20presentation%20Nov%202019%20FULL.pdf

### DIFF
--- a/american-medical-association-10th-edition.csl
+++ b/american-medical-association-10th-edition.csl
@@ -1,0 +1,276 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US" page-range-format="expanded">
+  <info>
+    <title>American Medical Association 10th edition</title>
+    <title-short>AMA (10th ed.)</title-short>
+    <id>http://www.zotero.org/styles/american-medical-association-10th-edition</id>
+    <link href="http://www.zotero.org/styles/american-medical-association-10th-edition" rel="self"/>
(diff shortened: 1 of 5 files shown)


adam3smith:
Thanks -- I'm wondering if we want to a) specify the edition in the title and b) keep the 10th edition around as we do for APA/MLA/Chicago.
I don't know if people will require/specify old AMA versions, but at this point I'd be surprised if they didn't. 
Thoughts?

rmzelle:
> I'm wondering if we want to a) specify the edition in the title and b) keep the 10th edition around as we do for APA/MLA/Chicago.
> I don't know if people will require/specify old AMA versions, but at this point I'd be surprised if they didn't.

Agree. Do you recall what we typically do for style variants? Just update them all to the newest style edition. See https://www.zotero.org/styles?q=american-medical-association

adam3smith:
> Do you recall what we typically do for style variants? Just update them all to the newest style edition

We do some complicated dance, which depends both on the likelihood that some dependent styles will continue to want the old version and how much users will want the old versions. Given that these are very limited (and pretty straightforward) changes, yes, I'd just switch all the other styles to 11th edition and see if there's protest.

POBrien333:
Do you want me to prepare the others versions of the style within this PR?

rmzelle:
> Do you want me to prepare the others versions of the style within this PR?

Yes, please. So far I renamed the existing AMA style as 10th edition, and made your 11th edition version the new american-medical-association.csl.

rmzelle:
@POBrien333, @adam3smith, I diffed the 10th edition styles and re-applied the (main) differences to @POBrien333's 11th edition style. Should be good to merge, I think.

adam3smith:
Yes, looks great, thanks!

PR #4709, opened 2020-04-17
Create pontifical-university-lateranense-ius.csl

palhander:

PR #4728, opened 2020-04-23
Create pontifical_universiti_lateranenseIUI

palhander:


adam3smith:
Is this different from https://github.com/citation-style-language/styles/pull/4709 ? If so how?

rmzelle:
@adam3smith, this appears to be identical to https://github.com/citation-style-language/styles/pull/4709/commits/142c654a62732280debce3d9a255ebe3219b237b. @palhander, let's keep the discussion in #4709.

PR #4731, opened 2020-04-24
Create ifao-publications.csl

NMonteix:
This a style without ISSN, that would work with all the publisher's publications


ifao-publications.csl
```diff
@@ -0,0 +1,335 @@
...
+    <terms>
+      <term name="editor" form="short">éd.</term>
+      <term name="accessed">consulté le</term>
+    </terms>
+  </locale>
+  <locale xml:lang="en">
+    <style-options punctuation-in-quote="false"/>
+    <terms>
+      <term name="translator" form="short">trans.</term>
+      <term name="open-quote">“</term>
```
adam3smith, line 34: You should also define the inner quotes then (just copy from the en-US locale)

ifao-publications.csl
```diff
@@ -0,0 +1,335 @@
...
+      <key macro="author-short"/>
+      <key variable="issued"/>
+      <key variable="page-first"/>
+      <key variable="title"/>
+    </sort>
+    <layout suffix=".">
+      <group display="block">
+        <text macro="author-short"/>
+      </group>
+      <group display="left-margin">
```
adam3smith, line 217: You have tested this? I'd have expected that just setting `display="block"` on this group (and no display on author short) would give you better results.

rmzelle:
Thanks. Do I understand correctly that this style can be used in both French and English?

And is there an list/overview of which IFAO journals this is for?

ifao-publications.csl
NMonteix: Done

ifao-publications.csl
NMonteix: It works perfectly on the test pane, less well on Word. I remember we had the same conversation about gallia.csl long ago (either here or on Zotero's forum). To make it work on Word, I need to add `suffix="&#9;"` after author-short. Just done.

NMonteix:
@adam3smith as for the (complicated) form of the style, see e.g. [there](https://www.ifao.egnet.net/bifao/119/10/). And it also appears [there](https://www.ifao.egnet.net/uploads/publications/normes/x_anglais-mep.pdf).
@rmzelle there would be a sort of a list [here](https://www.ifao.egnet.net/publications/pub-presentation/), with links only in the French part. Roughly, one journal [Bulletin de l’Institut français d’archéologie orientale (BIFAO) ISSN : 0255-0962](https://www.ifao.egnet.net/publications/catalogue/BIFAO/) and at least two series but some do not appear on their site.
[Bibliothèque d’étude (BiEtud) ISSN : 0259-3823]()
[Fouilles de l’Institut français d’archéologie orientale (FIFAO) ISSN : 0768-4703](https://www.ifao.egnet.net/publications/catalogue/FIFAO/).
Should I prepare dependent styles?

rmzelle:
> Should I prepare dependent styles?

We can take this IFAO style already, but yes, feel free to submit dependent styles for these journals. It doesn't look like IFAO has enough publications for it to make sense to use our tools for automatically generating dependents (https://github.com/citation-style-language/journals).

@adam3smith, looks good on my part.

NMonteix:
> It doesn't look like IFAO has enough publications for it to make sense to use our tools for automatically generating dependents

In fact the dependent styles might be more than I thought: a dozen are gathered (here)[https://www.ifao.egnet.net/publications/catalogue/]
I will see with the publisher to create a proper tab.
Anyway, I have tried to create the main three dependent styles, but something went wrong, even after correcting the ISSN and recreating the styles in the dependent subdirectory.

rmzelle:
Merci!

> I will see with the publisher to create a proper tab.

Great. At minimum we need journal titles, ISSNs (online and print, if available). Subject matter field (https://docs.citationstyles.org/en/1.0.1/specification.html#appendix-i-categories) and journal URL are useful too. https://github.com/citation-style-language/journals#metadata-repository-for-dependent-csl-styles explains how we organize this metadata, but you can also just submit a regular spreadsheet.

> Anyway, I have tried to create the main three dependent styles, but something went wrong, even after correcting the ISSN and recreating the styles in the dependent subdirectory.

You had one of the dependent styles twice, once in the root directory, and once in the "dependent" subdirectory.

NMonteix:
Thanks to you!

And thanks for the explanation

Le 29/04/2020 à 02:56, Rintze M. Zelle a écrit :

PR #4734, opened 2020-04-27
Create current-opinion-in-endocrinology-diabetes-and-obesity.csl

POBrien333:
Just making this style as somebody published a paper citing my previous work in it.
I have the CSL disease now... I check what type of citation style a journal uses... haha


rmzelle:
> I have the CSL disease now...

We'll get to 10,000 CSL styles some day!

rmzelle:
Thanks!

PR #4738, opened 2020-04-29
Update the-geological-society-of-london.csl

MaxineAtGeolsoc:
changed category field to geology
changed and=symbol to text
deleted publisher location (only need publisher name)


rmzelle:
Thanks! I restored the original author and template style, which shouldn't be removed when a style is tweaked like this.

PR #4739, opened 2020-04-29
Update BBU-FTO style

nturcan:
The style was updated:
- no ibid.;
- all names are small-caps;
- the text for an accessed link was deleted.


rmzelle:
It looks like this is the same as #4750, right?

PR #4740, opened 2020-04-29
Update journal-of-management-studies.csl

tetyanaloskutova:
Changed to use "and" instead of & between 2 authors


rmzelle:
Thanks!

PR #4741, opened 2020-04-30
Update journal-of-the-geological-society.csl

MaxineAtGeolsoc:
Updated generated from and documentation link lines.


rmzelle:
Thanks for updating the documentation links. I took care of this via https://github.com/citation-style-language/journals/pull/96 and https://github.com/citation-style-language/styles/pull/4751 using our script to generate dependent CSL styles.

PR #4759, opened 2020-05-03
Update journal-of-applied-pharmaceutical-research.csl

joapreditor:
When citations are moved from one place to other in the documents, the citation numbers are to updated automatically after pressing "Refresh". Unfortunately this was not happening. Hope with this code, it will be done. Kindly comment


adam3smith:
The code you added has no effect on citation numbers in the text and likely none on the style at all. All numeric styles should update when you move citations and refresh, so this is an issue with your reference manager of choice and you should use their troubleshooting channels.

PR #4765, opened 2020-05-05
fix issue with DOI URL for unified-style-sheet-for-linguistics.csl

POBrien333:
via https://forums.zotero.org/discussion/82952/error-in-stylesheet


rmzelle:
Thanks! (caught another instance of this)

PR #4767, opened 2020-05-05
Create apa-6th-edition-no-ampersand.csl

bwiernik:
Gotten several requests from authors writing in Spanish for APA 6 without ampersand.

Would we want to just switch the locale for Spanish to use "y" for `and="symbol"`?

### DIFF
--- a/apa-6th-edition-no-ampersand.csl
+++ b/apa-6th-edition-no-ampersand.csl
@@ -1,0 +1,1548 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" page-range-format="expanded">
+  <info>
+    <title>American Psychological Association 6th edition (no ampersand)</title>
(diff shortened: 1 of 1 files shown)


adam3smith:
> Would we want to just switch the locale for Spanish to use "y" for and="symbol"?

Doing this from memory, but I don't think and="symbol" can be localized? I think that should be on the list for CSL 1.0.2 (or wherever that discussion ended up ;).

If I'm wrong, I'd be happy to see this localized, yes, I've gotten the same complaint about APA in Spanish a number of times; I think that's why we created the original option.

bwiernik:
Ah, yes, forgot about that. Once `and="symbol"` is localizable, we can change the locales and move the no-ampersand versions to dependent styles I think.

rmzelle:
So would we want to turn this style into a Spanish one, then?

adam3smith:
> So would we want to turn this style into a Spanish one, then?

I don't think so, no. It's just a particular common scenario for not wanting the &

bwiernik:
Can this be merged?

PR #4768, opened 2020-05-05
fix initials of revista-biblica.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/355314/#Comment_355314


rmzelle:
Thanks!

PR #4769, opened 2020-05-06
added missing whitespace prefix before URL

fedormyskin:


adam3smith:
I don't understand why this would ever be necessary -- the access macro is called in a group with delimiter=" " and the locator macro itself with prefix=" ". Is this based on an error report? For what citation?

fedormyskin:
Yes, I was getting a wrong citation format for web pages: no whitespace
before the URL. Like this:
*CSL Search by Example.* (2012). *Citation Style Editor*.
http://editor.citationstyles.org/searchByExample/ (accessed 15 December
2012)

adam3smith:
> Yes, I was getting a wrong citation format for web pages: no whitespace
before the URL. Like ths

with which software? It really shouldn't matter and you can see that it e.g. doesn't in our automated test.

fedormyskin:
Zotero plugin for Microsoft Word on MacOS, I didn't check with other configurations

rmzelle:
> Zotero plugin for Microsoft Word on MacOS, I didn't check with other configurations

Yes, sorry, but I can't reproduce this with the style editor of Zotero 5.0.87. Is your copy of Zotero up to date? I get:

> Beyond Contiguity – Assessing the Quality of Genome Assemblies with the 3 C’s. (2020). PacBio. https://www.pacb.com/blog/beyond-contiguity/ (accessed 8 March 2020).

with the current style (https://github.com/citation-style-language/styles/blob/cb4e7f70a05d3fddd69c5e4a817559d82b7c54b7/digital-scholarship-in-the-humanities.csl). Have you tried reinstalling the style from https://www.zotero.org/styles?q=id%3Adigital-scholarship-in-the-humanities, just to make sure we're testing the same version?

And, like @adam3smith, reading the style by eye confirms that there should already be whitespace before the URLs.

rmzelle:
> Yes, I was getting a wrong citation format for web pages: no whitespace before the URL. Like this: *CSL Search by Example.* (2012). *Citation Style Editor*.
http://editor.citationstyles.org/searchByExample/ (accessed 15 December 2012)

And your example here is a little unclear, since there is a new line before the URL. Do I understand correctly that you saw:

> *CSL Search by Example.* (2012). *Citation Style Editor*.http://editor.citationstyles.org/searchByExample/ (accessed 15 December 2012)

(with no space between "*Citation Style Editor*." and the URL)

adam3smith:
I'll close this for inactivity and because we can't replicate this. Thanks for submitting the fix and we're happy to re-open if we can get steps to replicate for the error with a current version of Zotero, but even then that'd more likely be a processor issue. The style is almost certainly fine.

PR #4770, opened 2020-05-06
small fixes for revista-biblica.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/355361/#Comment_355361


rmzelle:
Thanks!

PR #4774, opened 2020-05-08
Create antafr.csl

samarkar:
New style for _Antiquités africaines_ Journal.

### DIFF
--- a/antafr.csl
+++ b/antafr.csl
@@ -1,0 +1,631 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
+  <info>
+    <title>Antiquités africaines</title>
+    <title-short>AntAfr</title-short>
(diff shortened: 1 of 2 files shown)


rmzelle:
Thanks! Same questions as for https://github.com/citation-style-language/styles/pull/4775:

And does this journal only publish in French?

samarkar:
Thank you very much for those corrections. I based myself on another style I created for a journal published in the team where I worked before, more oriented on art history: invisu.csl.
But I have changed a lot of things for the needs of the archaeological journals I am in charge of now, in particular by changing the style from "note" to "author-date". So I don't think it's useful to mention the InVisu style, which is very different.
Antonio

samarkar:
I forgot: the journal publishes articles in several languages (French, English, Italian, Spanish, German...).

rmzelle:
> So I don't think it's useful to mention the InVisu style, which is very different.

Thanks. I'd still like to include that as a template link, since we can use these links for a bunch of different things (such as attribution and maintenance of style families).

rmzelle:
Thanks!

PR #4775, opened 2020-05-08
Create archaeonautica.csl

samarkar:
New style for _Archaeonautica_ Journal


rmzelle:

And does this journal only publish in French?

samarkar:
Thanks! Same comments as for [#4774 ](https://github.com/citation-style-language/styles/pull/4774)
Antonio

rmzelle:
Thanks!

PR #4778, opened 2020-05-10
fix demote-non-dropping-particle for equine-veterinary-education.csl

POBrien333:
via https://forums.zotero.org/discussion/83043/issues-with-prefixed-author-surname-in-equine-veterinary-education-format

checked via several papers.


rmzelle:
Thanks!

PR #4782, opened 2020-05-13
Update eksploatacja-i-niezawodnosc.csl

Brenkacz:
The publisher "Eksploatacja i Niezawodnosc Maintenance and Reliability" uses the sorting of bibliography by author name.

### DIFF
--- a/eksploatacja-i-niezawodnosc.csl
+++ b/eksploatacja-i-niezawodnosc.csl
@@ -75,7 +75,7 @@
   </macro>
   <citation collapse="citation-number">
     <sort>
-      <key variable="citation-number"/>
+      <key variable="author"/>
     </sort>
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks!

Brenkacz:
Hello Rintze,

 

Thank you for approving merge the style. When I checked the style in the repository (https://www.zotero.org/styles?q=id%3Aeksploatacja-i-niezawodnosc) There is visible a new version (from yesterday). I use Mendeley Program to manage the bibliography, and there is visible still the old version of style (from January). Update option is not available. The same is when I uninstall and instal it again. How long will take to update the style in the database of Mendeley? Can I do something to do this or speed up this?

 

Kind regards,

Lukasz Brenkacz

PR #4783, opened 2020-05-13

OTristanF:
1. Added a year to subsequent citations when there is no author.
2. Although the examples suggest otherwise, cl. 9 odst. 1 tells us not to use the period at the end. of the link. The dean confirmed to me there should not be a period. Therefore, I have removed it.
3. Removed a dot from the suffix of accessed shortened term, since it is already included in some locales.
4. The dean told me that the shortened term for editors should be without a period. Therefore, I have updated it for the Czech locale.
5. Removed DOI since the documentation does not support it.
6. Excluded legal texts from the bibliography since the documentation says that they should not be included.
7. Show [online] only for webpages.


rmzelle:
Thanks!

PR #4785, opened 2020-05-13
create csl for the journal Textual Cultures

icornelius:
Chicago author-date format, but with family names in small caps, per journal specifications.


rmzelle:
Thanks!

PR #4786, opened 2020-05-14
Update taylor-and-francis-national-library-of-medicine.csl

POBrien333:
via https://forums.zotero.org/discussion/83126/taylor-francis-nlm

I checked 3 papers in 3 random journals that use this style as a dependent.
 T&F doesn't seem to be suuuper strict, but the overall format seems to be with stripped periods.


rmzelle:
Thanks! (https://www.tandf.co.uk//journals/authors/style/reference/tf_NLM.pdf also seems to strip periods for journal titles)

PR #4788, opened 2020-05-14
Update RMIT Harvard to match Easy Cite

areographe:
Easy Cite referencing tool states that a report with a URL and 'accessed' date should have these included in the bibliography


rmzelle:
Thanks!

PR #4790, opened 2020-05-15
Update revista-brasileira-de-ciencia-do-solo.csl

guilherme-cptk:
Update of the RBCS CSL file to suit the 2020 citation format.


rmzelle:
Thanks!

PR #4791, opened 2020-05-15
Update angewandte-chemie.csl

POBrien333:
via https://forums.zotero.org/discussion/83161/angewandte-chemie-style

checked in guidelines and a paper.


rmzelle:
Thanks!

PR #4794, opened 2020-05-16
Update American Physical Society styles

bwiernik:
Used by Physical Review Applied.


rmzelle:
> Used by Physical Review Applied.

@bwiernik, is there any indication that multiple journals use this variant?

bwiernik:
As far as I can see, it's currently just Physical Review Applied, but the publisher is adding new journals fairly regularly it seems, and it seems plausible that others might adopt this variant instead. Seems easier to me to keep track of them with the generic independent.

bwiernik:
Actually, yes, several other journals use titles: 
> Please note that Physical Review X, Physical Review Applied, and Physical Review Fluids require all references to include titles.

bwiernik:
(Some other metadata on these journals is out of date. Updating.)

adam3smith:
The change of the APS id would need to be included in renamed.json, please.
Since we do have this whole list, could we add this to https://github.com/citation-style-language/journals (note that the "APS" there is for the Physiological Society)

bwiernik:
I already added the renamed ID to renamed-styles.json. I can add them to journals.

rmzelle:
Dan also strongly requests us not messing around with style IDs, so I'm not sure we should rename "american-physics-society" to "american-physical-society", even if it's wrong. Although the impact might be limited if most people have installed one of the journal styles instead of (just) the independent parent.

bwiernik:
I'm happy to leave the style id the same; I was just following existing practice until that issue was finally resolved. Should we still rename the style file/self link but leave the id the same?

bwiernik:
Or we could add "American Physical Society" as another title to this style or as a dependent style?

bwiernik:
@adamsmith @rmzelle 

Okay, I added APS to the journals repo and restored the style ids here.

bwiernik:
Do I need to remove the dependent styles from this PR?

jlaehne:
> Or we could add "American Physical Society" as another title to this style or as a dependent style?

I would favor correcting the naming. I am always confused when looking for the APS style and it is has the wrong name. Couldn't you rename the style (and use the correct name in the new variants!) and then add a dependent style 'American Physics Society' for backwards compatibility?

bwiernik:
I've changed the display title but left the style id

bwiernik:
Anything else need to be changed here @adam3smith @rmzelle ?

adam3smith:
Great, thanks

PR #4796, opened 2020-05-17
Create turabian-9-author-date

TCfish:
As far as I know, the Turabian 9th edition is very similar to Chicago 17th. The only difference I am aware of for the author-date variant is using "no. x" for journal issue numbers.
So instead of: "Random Journal 5 (1): 123-321."
You will see: "Random Journal 5, no. 1: 123-321."

My PhD program requires Turabian 9th edition, author-date. So here is the new file, which is based off Chicago 17th. If there are other differences, I am not aware of it. Others can modify this file to fill in what I missed. This addition is still better than nothing. The last Turabian author-date variant we have in the Zotero repository is in 2013.


TCfish:
I've tried many changes in a now-closed, older pull request. I have no idea why csl-bot is giving me issues with the "unused macro." The code is almost the same as Chicago 17th (author-date). This should mean that the chicago-author-date.csl has an unused macro too. How did it pass the csl-bot test?

Anyone want to contribute and help fix this? Again, this will be helpful not just to me but to everyone in my school.

rmzelle:
@adam3smith, can you review this?

> I have no idea why csl-bot is giving me issues with the "unused macro." The code is almost the same as Chicago 17th (author-date). This should mean that the chicago-author-date.csl has an unused macro too. How did it pass the csl-bot test?

In general, we don't allow unused macros in repository styles, but we make style-ID-based exemptions for the Chicago styles (for which there are quite a few variants) since retaining unused macros in those makes them more maintainable.

adam3smith:
(will look shortly)

TCfish:
So... any updates? Can we merge this pull request? @adam3smith, @rmzelle ?

The only change I made (beside the beginning contributor names and all that) is on line 260:

Instead of 
<group prefix=" (" suffix=")">

I changed it to
<group prefix=", no. " suffix="">

Everything else on my Turabian 9 author date is the same as the Chicago 17 author date file.

TCfish:
Hello? @adam3smith, @rmzelle. Did you see my comment? Can we merge this yet?

adam3smith:
@rmzelle OK to squash&merge from my end

rmzelle:
@adam3smith, do you agree it would be better to delete https://github.com/citation-style-language/styles/blob/master/dependent/turabian-author-date.csl (which points to the (17th edition) chicago-author-date), and use its style ID and file name for this independent style, i.e. "turabian-author-date" instead of "turabian-author-date-9th-edition"?

adam3smith:
ah yes, absolutely, good call. If people do want the old format, they can just use Chicago author-date 16th edition, which this used to point to.

Will you make that change or should I?

rmzelle:
Thanks!

TCfish:
You guys rock!

PR #4797, opened 2020-05-17
Create journal-of-plant-protection-research.csl

POBrien333:
via https://forums.zotero.org/discussion/83189/style-request-journal-of-plant-protection-research


rmzelle:
Thanks!

PR #4810, opened 2020-05-19
Update bulletin-de-la-societe-prehistorique-francaise.csl

DomieComme:
Minor modifications of formatting adapted to the evolution of the bulletin


rmzelle:
Thanks!

PR #4812, opened 2020-05-19
Create society-of-biblical-literature-author-date.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/352911/#Comment_352911


POBrien333:
@rmzelle 
This is as far as I want to take this style. It's pretty solid now and amendments can be made by others in the future.

rmzelle:
Thanks!

PR #4814, opened 2020-05-20
Update molecular-biology-and-evolution.csl

gonzalezibeas:
I have edited line 151. Journal name must be in italics.


rmzelle:
Thanks!

PR #4818, opened 2020-05-21
Create interdisziplinare-anthropologie.csl

POBrien333:
via https://forums.zotero.org/discussion/83221/style-request-interdisziplinaere-anthropologie


rmzelle:
Thanks!

PR #4819, opened 2020-05-21
Update journal-of-international-business-studies.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/356587/#Comment_356587

as per editors request adapted. (preprint journals, book titles, newspapers)


rmzelle:
Thanks!

PR #4822, opened 2020-05-21

OTristanF:
removed book editors and translators (inherited from the style template)


rmzelle:
Thanks!

PR #4823, opened 2020-05-21
Update epilepsia.csl

ebiggs53135:
Adjusted number of author names listed before "et al." to comply with 2019 guidelines for authors


rmzelle:
Thanks!

PR #4824, opened 2020-05-22
Update eksploatacja-i-niezawodnosc.csl

Brenkacz:
According to the publisher's requirements, titles should be full (not shortened).


rmzelle:
Thanks!

PR #4826, opened 2020-05-25
Update pravnik.csl, masarykova-univerzita-pravnicka-fakulta.csl, and iso690-full-note-cs.csl

OTristanF:
masarykova-univerzita-pravnicka-fakulta.csl
- removed edition since it is not supported

pravnik.csl
- removed default locale since it can be used for The Lawyer Quarterly too

pravnik.csl & iso690-full-note-cs.csl
- I have modified the code to prevent it from using hardcoded values (such as s. for pages) instead of locale terms
- I have removed notes
- I have added new macros to prevent the code repetition.
- added thesis support and removed legal texts from the bibliography
- Removed a dot from the suffix of accessed shortened term, since it is already included in some locales. 
- Show [online] only for webpages

### DIFF
--- a/pravnik.csl
+++ b/pravnik.csl
@@ -323,21 +323,12 @@
           </group>
         </if>
         <else-if position="ibid">
-          <text term="ibid" suffix="."/>
+          <text term="ibid"/>
         </else-if>
         <else-if position="subsequent">
           <text macro="contributors-short" suffix=". "/>
           <text macro="title-short" font-style="italic"/>
-          <choose>
-            <if variable="locator">
-              <text prefix=", " macro="citation-locator" suffix="."/>
-            </if>
-            <else-if variable="accessed URL DOI" match="any">
-              <text term="online" prefix=" [" suffix="] "/>
-              <text macro="quoted"/>
-              <text prefix=". " macro="identifier"/>
-            </else-if>
-          </choose>
+          <text macro="locator-or-period"/>
         </else-if>
         <else>
           <choose>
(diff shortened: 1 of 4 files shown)


pravnik.csl
```diff
@@ -14,22 +14,34 @@
     <category citation-format="note"/>
     <category field="law"/>
     <issn>0231-6625</issn>
-    <summary>Czech ISO-690, full note, suitable for Pravnik - a journal published by the Institute of State and Law of the Czech Academy of Sciences.</summary>
-    <updated>2020-04-13T10:00:00+00:00</updated>
+    <summary>Suitable for Pravnik and its English version The Lawyer Quarterly - journals published by the Institute of State and Law of the Czech Academy of Sciences.</summary>
```
rmzelle, line 17: @OTristanF, so Právník publishes in Czech, and The Lawyer Quarterly (https://tlq.ilaw.cas.cz/index.php/tlq) in English? If they're different journals, it would be better to keep this Právník style fixed to Czech, and create a "dependent" CSL style for The Lawyer Quarterly that points to the Právník style, but has its `default-locale` set to English. (see https://github.com/citation-style-language/styles/blob/master/dependent/nature-biotechnology.csl to see what a dependent style looks like)

OTristanF: @rmzelle, thanks for the suggestion. I have created a dependent style.

OTristanF:
I see there is a rendering failure of the dependent style, although the tests passed. Is that alright?

rmzelle:
> @rmzelle I see there is a rendering failure of the dependent style, although the tests passed. Is that alright?

Yes, dependent styles don't render and always give an error. Nothing to worry about here.

rmzelle:
Thanks!

PR #4827, opened 2020-05-25
Update gost-r-7-0-5-2008-numeric-alphabetical.csl

AlexanderSavenkov:
Added post-weblog type and corrected missing spaces after initialized names.


rmzelle:
Thanks!

PR #4833, opened 2020-05-28
Hotfix: pravnik.csl & iso690-full-note-cs.csl

OTristanF:
Hotfix of the issued macro in pravnik.csl & iso690-full-note-cs.csl.

### DIFF
--- a/iso690-full-note-cs.csl
+++ b/iso690-full-note-cs.csl
@@ -241,14 +241,10 @@
             <text term="volume" form="short"/>
             <text variable="volume"/>
           </group>
-          <choose>
-            <if variable="issue">
-              <group delimiter="&#160;">
-                <text term="issue" form="short"/>
-                <text variable="issue"/>
-              </group>
-            </if>
-          </choose>
+          <group delimiter="&#160;">
+            <text term="issue" form="short"/>
+            <text variable="issue"/>
+          </group>
         </group>
       </if>
       <else-if type="webpage" match="any">
--- a/iso690-full-note-cs.csl
+++ b/iso690-full-note-cs.csl
@@ -14,7 +14,7 @@
     <category citation-format="note"/>
     <category field="law"/>
     <summary>Czech ISO-690, full note.</summary>
-    <updated>2020-05-25T10:00:00+00:00</updated>
+    <updated>2020-05-30T10:00:00+00:00</updated>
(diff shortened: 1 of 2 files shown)


OTristanF:
The tests are quite strange because they claim that the output has not changed. This is the Zotero output:

The former pravnik.csl:
ETZIONI, Amitai. Social Norms: Internalization, Persuasion, and History. _Law & Society Review_. 2000, roč. , 34, č. 1. DOI: 10.2307/3115119

The fixed pravnik.csl:
ETZIONI, Amitai. Social Norms: Internalization, Persuasion, and History. _Law & Society Review_. 2000, roč. 34, č. 1. DOI: 10.2307/3115119

iso690-full-note-cs.csl
```diff
@@ -233,17 +233,15 @@
...
-        </group>
+        <date variable="issued">
+          <date-part name="year" range-delimiter="&#160;&#8211;&#160;"/>
+        </date>
+        <choose>
+          <if variable="volume">
+            <text prefix=", " term="volume" form="short" suffix="&#160;"/>
+            <text variable="volume"/>
+          </if>
+        </choose>
```
rmzelle, line 244: @adam3smith, this would be better, right?

```suggestion
        <group delimiter=", ">
          <date variable="issued">
            <date-part name="year" range-delimiter="&#160;&#8211;&#160;"/>
          </date>
          <group delimiter="&#160;">
            <text term="volume" form="short"/>
            <text variable="volume"/>
          </group>
        </group>
```

iso690-full-note-cs.csl
```diff
@@ -233,21 +233,23 @@
...
-        </choose>
-        <choose>
-          <if variable="issue">
-            <text prefix=", " term="issue" form="short" suffix="&#160;"/>
-            <text variable="issue"/>
-          </if>
-        </choose>
+          </group>
+          <choose>
+            <if variable="issue">
```
rmzelle, line 245: @OTristanF, did you test my suggestion as-is? You shouldn't need these `<choose><if>` conditionals, as `<group/>` is an implicit conditional (see https://docs.citationstyles.org/en/stable/specification.html#group).

rmzelle:
Thanks. I assume it's working as desired now?

OTristanF:
@rmzelle yes, now it is alright. 

I am still wondering about the tests. They claim that the output is not changed, but when I try the old and the current versions of the styles in Zotero, the output for journal articles is not the same. Why is it like that?

PR #4835, opened 2020-05-29
Create journal-of-clinical-neurology.csl

POBrien333:
via https://forums.zotero.org/discussion/83294/request-of-new-csl-form-journal-of-clinical-neurology


rmzelle:
Thanks!

PR #4837, opened 2020-05-31
Update journal-of-plant-ecology.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/356516/#Comment_356516
Guidelines: https://academic.oup.com/jpe/pages/General_Instructions#Manuscript%20style

One of the earlier styles I made, so defo needed an update.


rmzelle:
Thanks!

PR #4841, opened 2020-06-03
Update antiquity.csl

vjjaimez:
Modified the existing CSL style for Antiquity journal. Many changes have been made:

- Inline citations are now sorted by date; e.g. (Grayson 1983; Evans 2009), instead of (Evans 2009; Grayson 1983).
- Inline citations are now grouped by author; e.g. (Aubet 2012, 2019) instead of (Aubet 2012; Aubet 2019).
- Inline citations: N-dashes now separate page spans.
- In bibliography, page numbers now always appear as they should, contracted in every situation; e.g. 118-56.
- In bibliography, second and further authors are now followed by a comma, instead of a dot; e.g. "M.E. AUBET,",  instead of "M.A. AUBET.". 
- In bibliography, journal papers now include a dot after the title and before the journal name.
- In bibliography, book series and number now appear between parentheses; e.g. (British Archaeological Series 101).
- Bibliography: more than 10 authors are abbreviated using first author + et al.
- Bibliography: DOI is included by default where available.

### DIFF
--- a/antiquity.csl
+++ b/antiquity.csl
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style class="in-text" version="1.0" and="symbol" delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="3" et-al-use-first="1" et-al-subsequent-min="3" et-al-subsequent-use-first="1" page-range-format="chicago" demote-non-dropping-particle="never" default-locale="en-GB" xmlns="http://purl.org/net/xbiblio/csl">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" and="symbol" delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="3" et-al-use-first="1" et-al-subsequent-min="3" et-al-subsequent-use-first="1" page-range-format="chicago" demote-non-dropping-particle="never" default-locale="en-GB">
   <!-- This style was edited with the Visual CSL Editor (https://editor.citationstyles.org/visualEditor/) -->
   <info>
     <title>Antiquity</title>
@@ -80,6 +80,19 @@
           <name-part name="family" font-variant="small-caps"/>
         </name>
         <label form="short" prefix=" (" suffix=")" plural="never"/>
+        <substitute>
+          <names variable="editor"/>
+          <names variable="translator"/>
+          <text macro="title"/>
+        </substitute>
+      </names>
+      <text macro="recipient"/>
+    </group>
+  </macro>
+  <macro name="contributors-count">
+    <group delimiter=". ">
+      <names variable="author">
+        <name form="count"/>
         <substitute>
           <names variable="editor"/>
           <names variable="translator"/>
@@ -425,9 +438,11 @@
       </group>
     </layout>
   </citation>
-  <bibliography and="symbol" delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="10" et-al-use-first="1" et-al-subsequent-min="10" et-al-subsequent-use-first="1" subsequent-author-substitute="—" entry-spacing="0" hanging-indent="true">
+  <bibliography and="symbol" delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="10" et-al-use-first="1" et-al-subsequent-min="10" et-al-subsequent-use-first="1" subsequent-author-substitute="&#8212;" entry-spacing="0" hanging-indent="true">
     <sort>
-      <key macro="contributors"/>
+      <key macro="contributors" names-min="1" names-use-first="1"/>
+      <key macro="contributors-count" names-min="3" names-use-first="3"/>
+      <key macro="contributors" names-min="3" names-use-first="1"/>
       <key variable="issued"/>
     </sort>
     <layout suffix=".">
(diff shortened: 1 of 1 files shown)


vjjaimez:
There still two remaining bugs that I can't figure out how to fix myself:

Bibliography: References where the first author is the same, but the second and further authors are different. According to Antiquity's submission guidelines, references with two authors and the same first author must be sorted alphabetically, not by date. Example: 1) HAWKES , J. & M. CARVER. 2003; 2) HAWKES, J. & D. SMITH. 1999. However, if there are are multiple references with 3 or more authors, and hence et al. is used, then the criteria for sorting is not name of the second author, but date. Example: 
HAWKES, J., D. SMITH & M. CARVER. 1980
HAWKES , J., P. BAYNES & M. CARVER. 1989. 
Since they are both Hawkes et al., they should be sorted by date, not alphabetically. I don't know how to do this.

The other bug that I can't fix is also in the bibliography. Antiquity requires a dot at the end of the contributors block, before the date. The problem is that this needs to be included or not depending on the number of authors, because in references with just one author there is already one dot (the dot corresponding to name initialization), but in references with multiple authors the initializing dot is absent. Example (look at the dot right before the date):
Marzoli, D., J. Suárez-Padilla & C. León Martín 2020. Los Castillejos de Alcorrín (Manilva, Málaga): transformaciones locales e impacto colonial fenicio en el entorno del Estrecho de Gibraltar (siglos IX-VII a.C.). Actualidad de la investigación arqueológica en España 1: 305–23.
Mederos Martín, A. 2006. Fenicios en Huelva, en el siglo X AC, durante el reinado de Hîrãm I de Tiro. Spal 15: 167–88. https://doi.org/10.12795/spal.2006.i15.08.

Is there any way to make conditional actions based on the number of authors in a given reference?

rmzelle:
> Bibliography: References where the first author is the same, but the second and further authors are different. According to Antiquity's submission guidelines, references with two authors and the same first author must be sorted alphabetically, not by date. Example: 1) HAWKES , J. & M. CARVER. 2003; 2) HAWKES, J. & D. SMITH. 1999. However, if there are are multiple references with 3 or more authors, and hence et al. is used, then the criteria for sorting is not name of the second author, but date. Example:
> 
> > HAWKES, J., D. SMITH & M. CARVER. 1980
> > HAWKES , J., P. BAYNES & M. CARVER. 1989.
> 
> Since they are both Hawkes et al., they should be sorted by date, not alphabetically. I don't know how to do this.

This can be solved with something like:

```xml
  <macro name="author-count">
    <names variable="author">
      <name form="count"/>
    </names>
  </macro>
  <citation>
    <sort>
      <key macro="author" names-min="1" names-use-first="1"/>
      <key variable="issued"/>
    </sort>
    ...
  </citation>
```

Per the requirements at http://journal.antiquity.ac.uk/sites/default/files/downloads/Submission%20Guidelines.pdf, items are first sorted alphabetically by the first author. Then, ambiguously sorted items (with the same first author) are sorted by the number of names (1, 2, or >= 3). Then, items that are still ambiguous are sorted by the second name. Finally, remaining ambiguous items are sorted by date.

> The other bug that I can't fix is also in the bibliography. Antiquity requires a dot at the end of the contributors block, before the date. The problem is that this needs to be included or not depending on the number of authors, because in references with just one author there is already one dot (the dot corresponding to name initialization), but in references with multiple authors the initializing dot is absent. Example (look at the dot right before the date):
> 
> > Marzoli, D., J. Suárez-Padilla & C. León Martín 2020. Los Castillejos de Alcorrín (Manilva, Málaga): transformaciones locales e impacto colonial fenicio en el entorno del Estrecho de Gibraltar (siglos IX-VII a.C.). Actualidad de la investigación arqueológica en España 1: 305–23.
> > Mederos Martín, A. 2006. Fenicios en Huelva, en el siglo X AC, durante el reinado de Hîrãm I de Tiro. Spal 15: 167–88. https://doi.org/10.12795/spal.2006.i15.08.
> 
> Is there any way to make conditional actions based on the number of authors in a given reference?

It looks like you just need to remove the dot before the year. The dot for single-author items will come from the name initial.

vjjaimez:
Thank you! I'll try out the fix to the first bug you suggested.

However, the second one remains a problem. Again, the issue is that the journal doesn't treat single- and multiple-author journals the same way. I first thought exactly as you did, that removing the dot before the year would be enough. This is the solution for single-author records. In this journal, the first author is listed as FAMILY-NAME, Initial-of-the-first-name and a dot. So you're right: the dot for single-author items will come from the name initial. However, where does the dot come from in entries with multiple authors? In this journal, the second and subsequent authors must be listed in reverse order to that of the first author:  Initial-of-the-first-name followed by a dot and the FAMILY-NAME. Thus, any 'contributors block' where multiple authors are listed won't end in a dot, it will simply end with the family name of the last author. How do I add the dot then? 
I hope I was able to explain myself. It is much easier to see this on practice than putting it into words.

vjjaimez:
> items are first sorted alphabetically by the first author. Then, ambiguously sorted items (with the same first author) are sorted by the number of names (1, 2, or >= 3). Then, items that are still ambiguous are sorted by the second name. Finally, remaining ambiguous items are sorted by date.

Hi! You correctly identified what the problem is, but I tried implementing your proposed solution and unfortunately it doesn't seem to have the desired affect. Instead, what it does is simply sort references by date. I attach a screenshot of the resulting bibliography:
Screenshot_2020-06-06_13-58-46

vjjaimez:
In both bugs, the problem, as I see it, is that the specific format that a reference takes depends on the number of authors in the reference. There must be some conditional variable that allows this, but I'm too much of a noob to figure it out :-/

rmzelle:
@vjjaimez, I noticed a problem in my example code above. The bibliography sorting should (obviously) be specified for `<bibliography/>`, not `<citation/>`. I modified your style, and now get the following in Zotero, which seems correct:

> Christensen, S.J. 2040.
Christensen, S.J. & K.B.R.M. Bob 2017.
Christensen, S.J. & K.B.R.M. Dan 2016.
Christensen, S.J., K.B.R.M. Dib, N. Spodsberg, K. Borch & P. Westh 1920.
Christensen, S.J., K.B.R.M. Ast, N. Spodsberg, K. Borch & P. Westh 1940.
Zob, S.J., K.B.R.M. Krogh, N. Spodsberg, K. Borch & P. Westh 2023.

Can you give it a try?

vjjaimez:
Much better now, thank you @rmzelle! It works beautifully.
I haven't been able to fix the last remaining bug yet. Basically, in the bibliography a period is required before the date, but only for references where author >1. In references where author=1, the dot is already given by the initialization, so there's no need to add it.

rmzelle:
Try the latest version. (note that duplicate punctuation should be automatically suppressed)

vjjaimez:
It woks like a charm. Thank you!

rmzelle:
Thanks!

PR #4842, opened 2020-06-03
Create journal-of-clinical-ethics.csl

POBrien333:
via https://forums.zotero.org/discussion/83469/style-request-journal-of-clinical-ethics


rmzelle:
Thanks!

PR #4844, opened 2020-06-04
create independent style for new-zealand-journal-of-forestry-science

POBrien333:
via https://forums.zotero.org/discussion/83506/correction-required-to-nz-j-forestry-science-reference-style

they moved away from Springer in 2018: https://link.springer.com/journal/40490/volumes-and-issues


rmzelle:
Thanks! :evergreen_tree:

PR #4847, opened 2020-06-05
Update methods-of-information-in-medicine.csl

POBrien333:
via https://forums.zotero.org/discussion/83519/style-error-methods-of-information-in-medicine

Switched to Thieme Verlag. thieme.csl doesn't fit. It is an AMA variant with lots of changes. I just modified the old style.


rmzelle:
Thanks!

PR #4848, opened 2020-06-05
Create etasr.csl

dionojs:
the csl citation style file for Engineering, Technology & Applied Science Research
(www.etasr.com).
Largely based on the IEEE 2018 guidelines, V 11.12.2018 withour the use of abbreviations for containers.
I have tried to follow the instructions in https://github.com/citation-style-language/styles/wiki/Style-Requirements and apply all instructions to the csl file, but then the file could not be added to zotero (an error message popped up).

Thus, I uploaded the original version that worked in zotero without issues.


rmzelle:

dionojs:
ETASR style is largely based on IEEE 2018 guidelines, V 11.12.2018 without the use of abbreviations for containers.
I have also put that statement in the summary although I just noticed a typo (withour instead of without)

rmzelle:
Thanks!

PR #4849, opened 2020-06-06
Create style for Journal of Vertebrate Biology

kilkas:
I would like to add a style for the Journal of Vertebrate Biology. This style is based on the style for "Invertebrate Biology". The file passed XML validator. I am just not sure what is the difference between a dependent and an independent style.


rmzelle:
Thanks!

PR #4850, opened 2020-06-07
correct et-al settings for heart-rhythm.csl

### LINKED ISSUE #4845: Bug in CSL-style heart-rhythm.csl, et al behavior, 123/3 should be 7/3
Et al rule for the journal should be 7/3 and the comments state that it been accordingly adjusted from 13/3 but in reality the old 13/3 rule persists (line 262: <bibliography et-al-min="13" et-al-use-first="3" second-field-align="flush">.

(this being one of the reasons for a recent manuscript returned, I was not able to correct the simple bug in Github)

POBrien333:
closes #4845 
Correct et-al settings.

### DIFF
--- a/heart-rhythm.csl
+++ b/heart-rhythm.csl
@@ -259,7 +259,7 @@
       <text variable="citation-number"/>
     </layout>
   </citation>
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks!

PR #4854, opened 2020-06-09
show issue numbers in springer basic author-date style

fgnievinski:
forked springer-basic-author-date-no-et-al.csl

### DIFF
--- a/dependent/gps-solutions.csl
+++ b/dependent/gps-solutions.csl
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" default-locale="en-US">
-  <!-- Generated with https://github.com/citation-style-language/utilities/tree/master/generate_dependent_styles/data/springer -->
+  <!-- Springer, generated from "springer" metadata at https://github.com/citation-style-language/journals -->
   <info>
     <title>GPS Solutions</title>
     <title-short>GPS Solut</title-short>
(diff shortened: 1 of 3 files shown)


adam3smith:
Is this for a specific journal or other requirement?

fgnievinski:
Yes, it's for Springer's "GPS Solutions" journal:
https://www.springer.com/journal/10291/submission-guidelines#Instructions%20for%20Authors_References
GPS Solutions has its own CSL file, which simply inherits 
Springer basic author-year:
https://github.com/citation-style-language/styles/blob/master/dependent/gps-solutions.csl
I didn't update that CSL file directly because I thought 
that showing issue numbers might be helpful in general.

fgnievinski:
just pinging to avoid being automatically closed for no further activity.

adam3smith:
Sorry for the long wait. So I've confirmed that GPS journal consistently does use issue numbers. I'd want to double-check that some of the journals of the springer-basic-author-date-no-et-al.csl style _do_ use issue numbers next and then either merge this or replace the original style

adam3smith:
@rmzelle -- currently I'm not seeing any other journal that uses this style, but given the number of Springer journals, there's a good chance there is. Two options:
1. Take this style and redirect the gps-solutions independent style
2. Make this style just for gps-solutions and delete the dependent

I don't have a strong opinion either way.

rmzelle:
>     1. Take this style and redirect the gps-solutions dependent style

Let's go with this one (having many Springer independent variants isn't super problematic as I assume nobody looks for them, unlike APA/CMoS).

@fgnievinski, thanks!

PR #4859, opened 2020-06-11
Create alkoholizmus-a-drogove-zavislosti.csl

michalturcek:
Citation style for Slovak journal in addiction medicine entitled „Alkoholizmus a drogové závislosti“ (Alcoholism and drug addictions). 
Journal homepage: https://www.adzpo.sk
The style respects the guidance for authors (including citation examples): https://www.adzpo.sk/sk/adzpo-pokyny-pre-autorov. 
Example of published article: https://www.adzpo.sk/images/articles/adzpo-2018-53-4-171-180.pdf

### DIFF
--- a/alkoholizmus-a-drogove-zavislosti.csl
+++ b/alkoholizmus-a-drogove-zavislosti.csl
@@ -1,21 +1,21 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US" xmlns="http://purl.org/net/xbiblio/csl">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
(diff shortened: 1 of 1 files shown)


POBrien333:
The style is currently not validating for two errors.
You have a macro in your style that is not used, so delete that macro.
And the licensing message was changed, so you'll need to change that back to the one that is shown.

rmzelle:
Thanks!

michalturcek:
Hello, it seems the style was accepted since I can see it can be found in the Zotero repository. Thank you very much for the help with correcting the errors and final steps, you really made my day!

Regarding the question, wheter it is an original or template-based style: From my point of view it is original. First I reviewed four existing styles (Vancouver, APA 6th edition, British Journal of Pharmacology, Genetics and Molecular Biology) at the beginning, to get familiar with the environment, to see the structure and how macros are made. Then I started to build it piece by piece, using the CSL visual editor, considering various publication types and our Journal standards. With some breaks, it took me several months, since I am not a programmer...

rmzelle:
Okay, in that case we're all set. Thanks again for your contribution!

PR #4860, opened 2020-06-12
add missing cs:sort to cs:bibliography in universitat-bern-institut-fur-musikwissenschaft

denismaier:
@adam3smith This fixes the erroneous sorting reported at https://forums.zotero.org/discussion/comment/357476

I'm not sure if just using `<key macro="author_title-bibliography"/>` is a wise choice, or if using the content of that macro would have any benefits:

```xml
<sort>
  <key macro="primary-contributor"/>
  <key macro="title_volumes"/>
</sort>
```


adam3smith:
Thanks! I don't think using the whole macro versus its parts makes a difference, no.

PR #4863, opened 2020-06-12
Fixed page-range-format

luciomarinelli:


adam3smith:
Thanks (confirmed with instructions for authors)

PR #4866, opened 2020-06-14
Add "personal communication" to remaining APA locales

bwiernik:
Found examples of usage for these. No examples for Khmer, but the main locale file for that is basically empty anyway.


adam3smith:
Thanks!

PR #4867, opened 2020-06-14
Add article number to IEEE

bwiernik:
Also change the test for speech/conference papers being published to be editor/volume/issue/page instead of container-title.


rmzelle:
Thanks!

PR #4868, opened 2020-06-15

OTristanF:
I have received feedback for the style that blogs were treated differently than other webpages. This should solve it.


rmzelle:
Thanks!

PR #4871, opened 2020-06-16

anfourny:
…al-et-europeen.csl

These changes were made by a student after he received feedback from a group who used this style. He then submitted this updated version of the style to me. It's been so long since I worked on this code, I'd be very grateful if someone could help me and see if everything seems OK.

Thanks!


rmzelle:
@adam3smith, LGTM.

adam3smith:
Thanks! (I changed the code for bill/legislation to what I assume was intended: testing match="all" on more than one type will always return false, so that didn't make sense as written)

anfourny:
Thank you guys so much for your help!

PR #4874, opened 2020-06-17
Create ruhr-universitat-bochum-medizinische-fakultat-variante-a.csl

POBrien333:
via https://forums.zotero.org/discussion/83704/need-help-in-editing-style

As far as I understand this is for the Ruhr Universitaet Bochum and dissertations in the medical/life sciences. There is two variants: A = numeric, B = author-date.


adam3smith:
@POBrien333 -- just to make sure: There isn't actually a reason to define and as the English "and" right? Didn't see anything in the styleguide

rmzelle:
Thanks! (changed "Variante A" to "numerisch")

PR #4875, opened 2020-06-18
update et-al settings for stroke.csl

POBrien333:
https://forums.zotero.org/discussion/83735/changing-the-number-of-author-names-preceding-et-al


adam3smith:
Thanks!

PR #4883, opened 2020-06-24
Create electrophoresis.csl

prikryljan:
Modified Electrophoresis (Wiley) Zotero style
[https://onlinelibrary.wiley.com/journal/15222683](url)

Citations in text modified to be in square brackets without upper index.

### DIFF
--- a/electrophoresis.csl
+++ b/electrophoresis.csl
@@ -14,7 +14,7 @@
     <category field="chemistry"/>
     <summary>The Electrophoresis style.</summary>
     <updated>2020-06-24T12:49:56+00:00</updated>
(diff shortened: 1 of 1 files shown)


rmzelle:

rmzelle:
Thanks, looks like this was derived from an (old) ACS style.

I also noticed that https://onlinelibrary.wiley.com/page/journal/15222683/homepage/forauthors.html links to
https://csl.mendeley.com/styles/558732791/Electrophoresis, which seems to have been created by one of the journal's editors-in-chief. (I didn't compare which version matches the guidelines better)

Issue #4882, opened 2020-06-23
Question about combining multiple conditions


danmichaelo:
Hope this is the right place to ask questions, CONTRIBUTING.md seemed to suggest it is.

I'm working on an update to `universitetet-i-oslo-rettsvitenskap.csl` and struggle a bit with combinations of conditions. I'm also quite new to CSL. My goal is to format the locator in a special way if (locator is "section" OR "paragraph") AND (type is "legislation" OR "legal_case" OR "bill").

This matches too much:

```xml
<if locator="section paragraph" type="legislation legal_case" match="any">
``` 

(for instance locator=section, type=book)

While this matches too little / nothing(?):

```xml
<if locator="section paragraph" type="legislation legal_case" match="all">
```

Is there a way I can express this without resorting to multiple levels of <if>s and having to repeat the `<else>` section? Here's my full macro at the moment:

```xml
  <macro name="locator">
    <choose>
      <if locator="section paragraph" type="legislation legal_case">
        <text variable="locator" prefix=" § "/>
      </if>
      <else>
        <group delimiter=" ">
          <label variable="locator" form="short"/>
          <text variable="locator"/>
        </group>
      </else>
    </choose>
  </macro>
```

adam3smith:
(I'd generally put "how to" questions on our discourse forum discourse.citationstyles.org/ but it doesn't matter terribly).
There's no great solution to this, no; you can't combine AND and OR in a single condition, so some repetition is unavoidable. 

I think the shortest version is 
```
  <macro name="locator">
    <choose>
      <if l type="legislation legal_case" match="none">
        <group delimiter=" ">
          <label variable="locator" form="short"/>
          <text variable="locator"/>
        </group>
      </if>
     <else-if locator="section paragraph" match="any">
       <text variable="locator" prefix=" § "/>
      </else-if>
      <else>
        <group delimiter=" ">
          <label variable="locator" form="short"/>
          <text variable="locator"/>
        </group>
      </else>
    </choose>
  </macro>
```

danmichaelo:
Thanks a lot, it's better to know that it's not possible than to ponder if there is maybe some way that I just didn't find! And thanks for preparing a concrete suggestion! Will check out the discourse forum.

PR #4886, opened 2020-06-24
Update universitetet-i-oslo-rettsvitenskap.csl

danmichaelo:
The style has been updated to support locators. Sections and paragraph locators are prefixed using `§` for statutes and bills, as discussed in #4882 .


adam3smith:
Thanks!

PR #4887, opened 2020-06-25
Create rmit-university-modified-harvard-for-primo.csl

katie-finch:
Version of RMIT Harvard intended for use for back-end csl config in RMIT's Primo Library discovery system only.


rmzelle:
@katie-finch, sorry, but we don't accept custom styles into this repository meant solely for back-ends, as such styles basically only have a single user and would be irrelevant to everybody else.

The repository is primarily for sharing and discovery of styles where the style creator and users don't know each other, as is the case for general style guides and journal. See also https://github.com/citation-style-language/styles/wiki/Criteria-for-Accepting-Styles.

If Primo only supports use of repository styles (which appears to be case per e.g. https://knowledge.exlibrisgroup.com/Primo/Product_Documentation/Primo/Back_Office_Guide/110Additional_Primo_Features/Managing_Citation_Styles), the best solution here would be to ask them to add support for non-repository styles.

PR #4890, opened 2020-06-26
Update rmit-university-harvard.csl

katie-finch:
Amendments to correct an error of duplicate commas appearing in some systems (eg Ex Libris' Primo).


rmzelle:
Thanks!

PR #4891, opened 2020-06-26
Update universidade-do-porto-faculdade-de-engenharia-chicago.csl

serbernardo:
The quotes "" were missing in the the webpages titles.

So in the code, inside of `<macro name="title">` tag and then inside of `<else-if type="legal_case interview patent webpage" match="any">`
We have changed the line 209.

From:
`<text variable="title"/>`

to: 
`<text variable="title" quotes="true"/>`

### DIFF
--- a/universidade-do-porto-faculdade-de-engenharia-chicago.csl
+++ b/universidade-do-porto-faculdade-de-engenharia-chicago.csl
@@ -206,7 +206,7 @@
         </choose>
       </else-if>
       <else-if type="legal_case interview patent webpage" match="any">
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks!

Note that you don't need to describe the code changes you're making, as we can see this in the "Files changed" tab of the pull request, although it's always good to describe the reason behind the code changes.

XML fragments (like `<text variable="title"/>`) also need to be wrapped in backticks to render correctly (see the "Code" examples at https://guides.github.com/features/mastering-markdown/).

PR #4893, opened 2020-06-28
document origin of Springer Basic

fgnievinski:
note that Springer Basic is based on Harvard - CBE.


adam3smith:
Thanks -- since "Harvard style" isn't really a thing, I don't think this adds anything to understanding/specifying the style, so let's keep this out. I also prefer the simpler link to the PDF, which I assume is more likely to remain stable.

PR #4896, opened 2020-06-29
Create german-kunstwissenschaft.csl

malte-frey:
I'm relativly new to this and needed to get some specification for german citing style, fitting to my needs. The code could probably be simplified.

### DIFF
--- a/german-kunstwissenschaft.csl
+++ b/german-kunstwissenschaft.csl
@@ -6,13 +6,14 @@
     <title-short>DISS</title-short>
     <id>http://www.zotero.org/styles/german-kunstwissenschaft</id>
     <link href="http://www.zotero.org/styles/german-kunstwissenschaft" rel="self"/>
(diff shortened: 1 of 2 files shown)


rmzelle:
@malte-frey, thanks. Is your style based on any style guide in particular? We generally only accept styles in the repository that are based on a documented citation format. See also https://github.com/citation-style-language/styles/wiki/Style-Requirements#10---documentation-link

malte-frey:
Yes, it's wirtten according to the standards of The Academy of Fine Arts Münster, Germany. I put the documentation link in the file now, which leads to a PDF describing the standards.

rmzelle:
Thanks! That'll do nicely. We'll just name the style after the Kunstakademie Münster, then.

malte-frey:
Yes, I did. I put in the link now. Is this ok?

rmzelle:
Great, thanks!

malte-frey:
Hey guys, would it be possible to change the name of the style to german kunstwissenschaft.csl again? Due to some specifics in the code I could not 100% write it like the template link would advise. Additionally, I wrote some things on which are no specific instructions in the guidelines. To be sure that there is no legal trouble I'd rather youd'd change the name. Thank you.

PR #4898, opened 2020-07-01
Create geographische-zeitschrift.csl

POBrien333:
via https://forums.zotero.org/discussion/83866/style-request-geographische-zeitschrift-franz-steiner-verlag

Somehow the authors in the bib are NOT initialised. What's going on there?

### DIFF
--- a/geographische-zeitschrift.csl
+++ b/geographische-zeitschrift.csl
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
+  <!-- Polyglot; journal publishes in German and English -->
   <info>
     <title>Geographische Zeitschrift</title>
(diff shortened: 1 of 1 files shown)


geographische-zeitschrift.csl
```diff
@@ -0,0 +1,288 @@
...
+              <text macro="author-short"/>
+            </else>
+          </choose>
+          <text macro="issued"/>
+        </group>
+        <text variable="locator"/>
+      </group>
+    </layout>
+  </citation>
+  <bibliography et-al-min="3" et-al-use-first="1" initialize="false" entry-spacing="0" hanging-indent="true">
```
rmzelle, line 271: > Somehow the authors in the bib are NOT initialised. What's going on there?

`<bibliography ... initialize="false" ...>` is why. See https://docs.citationstyles.org/en/1.0.1/specification.html#name. It's an option to decorate existing initials with the suffix specified by `initialize-with`, but not convert full names to initials.

POBrien333: Ahhhh. Thanks for looking into it and explaining. :)

rmzelle:
Thanks!

PR #4904, opened 2020-07-04
Create apa-german.csl

daniel043:
Basic APA style for German, mainly for working with online sourrces. 
If available, always shows source URLs.

### DIFF
--- a/apa-german.csl
+++ b/apa-german.csl
@@ -1,0 +1,304 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="de-DE">
+  <info>
+    <title>APA German</title>
+    <id>http://www.zotero.org/styles/apa-german</id>
+    <author>
+      <name>Daniel Eggert</name>
+      <email>info@daniel-eggert.de</email>
+    </author>
+    <category citation-format="author-date"/>
+    <summary>Basic German APA style(Jul-2020).</summary>
+    <updated>2020-07-04T12:20:03+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="de">
+    <terms>
+      <term name="et-al">et al.</term>
+      <term name="retrieved">Zugriff am</term>
+    </terms>
+    <date form="text">
+      <date-part name="year"/>
+      <date-part name="month" prefix=", "/>
+      <date-part name="day" prefix=" "/>
+    </date>
+  </locale>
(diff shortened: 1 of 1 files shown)


rmzelle:
@daniel043, thanks. Is this based on a particular version/translation of APA? Are you not getting good results with the general version (https://github.com/citation-style-language/styles/blob/master/apa.csl)?

adam3smith:
Also, the style Deutsche Gesellschaft für Psychologie is essentially German APA. I'd be reluctant to duplicate: 
https://github.com/citation-style-language/styles/blob/master/deutsche-gesellschaft-fur-psychologie.csl

daniel043:
> Also, the style Deutsche Gesellschaft für Psychologie is essentially German APA. I'd be reluctant to duplicate:
> https://github.com/citation-style-language/styles/blob/master/deutsche-gesellschaft-fur-psychologie.csl

Yes, I saw this German one. I was not happy with how it handled URL sources...  unlike normal APA 7th style it is not showing all URLs in the blibliography (e.g. newspaper and journal articles URLs are missing). The beforementioned general APA 7th style does, but does not have the German translations. So after some time consuming search I decided to make this, to get the best German APA result and thought it might be nice to contribute.

bwiernik:
What exactly is different here versus using the regular apa.csl style set to German?

bwiernik:
Looking at the code, it looks like this wasn’t based on the APA 7 apa.csl. It’s missing most of that style. Can you take a step back and describe what doesn’t look right to you when you use the regular American Psychological Association (7th edition) style in German?

daniel043:
> Looking at the code, it looks like this wasn’t based on the APA 7 apa.csl. It’s missing most of that style. Can you take a step back and describe what doesn’t look right to you when you use the regular American Psychological Association (7th edition) style in German?

It is based on the style "Deutsche Gesellschaft für Psychologie"... as you said it is essentially German APA. Compared to APA 7, most importantly, that style has additional German words in several sections, e.g. for "added on %date", "seen on %date", etc. . So for Germans that style looks more sophisticated than the available APA 7 style. But it lacks a proper output of online resources (which the normal APA style has). So this contribution is a mix of those two, to get a better German style that is close to APA.

bwiernik:
APA style doesn't use most of those terms anymore in any language. If a new German APA style were to be made, you should start from apa.csl, which is the current version of APA style.

adam3smith:
I'm generally skeptical here. We can take a look at URL handling in the DGP style (which may be wrong?), but our general stance has been to only accept APA in a new language where an  guide for how it should be translated exists (like it does e.g. for French and Norwegian). The  guide in German is the DGP style -- where people don't like that, I think using generic translated APA 7 is fine.

bwiernik:
I'm not really sure why you don't just use the regular APA.csl with the language set to German? That would actually follow current APA guidelines correctly.

daniel043:
Alright... thanks for the clarification.

PR #4906, opened 2020-07-06
Add period between editor and translator (SBL styles)

johnhpercival:
As reported in forum: https://forums.zotero.org/discussion/comment/359784/#Comment_359784


rmzelle:
Thanks!

PR #4908, opened 2020-07-09
Create nspc-dprof.csl

boycott:
Middlesex uses cite-them-right for DProf students to validate citations.
The inline citation for the harvard-cite-them-right.csl didn't match the expected format:

`<Authors> (<Year>, location. <location>)`
e.g
Fanon (2008, p. 6)
or:
Fanon (2008, pp. 6-7)

I have run the csl through the validator.


adam3smith:
Could you say a bit more about the use case here? The CSL Cite Them Right style would print the correct  (Fanon, 2008, p. 6).

Also, just as a note, we wouldn't accept styles for use in an internal tool/API for the public repository, see https://github.com/citation-style-language/styles/wiki/Criteria-for-Accepting-Styles
They could obviously still be hosted & used locally.

PR #4910, opened 2020-07-10
Create common-market-law-review.csl

POBrien333:
via https://forums.zotero.org/discussion/83903/style-request-common-market-law-review


adam3smith:
Looks like this isn't quite ready yet, right? I'm seeing one more request by the OP for fixes and some of the output still seems a bit funky

rmzelle:
Thanks!

PR #4918, opened 2020-07-13
Update harvard-institut-fur-praxisforschung-de.csl

POBrien333:
via https://forums.zotero.org/discussion/84123/optimierung-im-stil-institut-fuer-praxisforschung-harvard-bahr-frackmann-german


rmzelle:
Thanks!

PR #4922, opened 2020-07-14
Create harvard-prifysgol-caerdydd.csl

YoungZoe:
This is a Welsh language version of the Cardiff University harvard style


rmzelle:
Thanks!

The only real difference was the style locale, so I turned this into a so-called dependent style (which are much more compact) that relies on the English version.

PR #4912, opened 2020-07-10
Update gewerblicher-rechtsschutz-und-urheberrecht.csl

CMvGR:
The original style contained a bug: For books and encyclopaedia articles the abbreviated second citation would create a double comma between the reference and the page number (e.g.: "[1] Schwab, (o. Fn. 1), , S. 15."). This has been addressed by slightly modifying two lines. The abbreviated second citation now reads correctly (e.g. "[1] Schwab, (o. Fn. 1), S. 15."). There are no adverse consequences to the proposed changes.

PR #4923, opened 2020-07-14
Update gewerblicher-rechtsschutz-und-urheberrecht.csl

CMvGR:
The original style contained a bug: For books and encyclopaedia articles the abbreviated second citation would create a double comma between the reference and the page number (e.g.: "[1] Schwab, (o. Fn. 1), , S. 15."). This has been addressed by slightly modifying two lines. The abbreviated second citation now reads correctly (e.g. "[1] Schwab, (o. Fn. 1), S. 15."). There are no adverse consequences to the proposed changes


rmzelle:
Thanks! (replaces #4912)

PR #4933, opened 2020-07-19
simplify usage and maintenance of IEEE style

monperrus:
Having two IEEE styles:

1. is confusing for users (it took me some time to realize that ieee.csl does not support URLs and to figure out how to fix it)
1. requires double maintenance

What about having a single, unified version of the IEEE style?


adam3smith:
Thanks, but these two exist for a reason. IEEE makes the use of a URL optional for anything other than webpages, which is why we need two versions. This isn't unusual for CSL styles and I don't see any way to avoid this while still giving users both options.

monperrus:
ack, thanks.

PR #4935, opened 2020-07-20
Remove issue from article-journal academy-of-management-review.csl

POBrien333:
via https://forums.zotero.org/discussion/84252/no-lines-of-codes-in-the-csl-style-file#latest


POBrien333:
Closing as the journals don't do this very consistently.

PR #4937, opened 2020-07-21
fix et-al & add DOI thyroid.csl

POBrien333:
via https://forums.zotero.org/discussion/84253/maximum-of-three-authors-in-bibliography


rmzelle:
Thanks!

PR #4938, opened 2020-07-22

POBrien333:
via https://forums.zotero.org/discussion/84279/issue-with-lancaster-university-harvard

For you to decide if we take this or not.
Guidelines just speak of "alphabetically" and no further explanation.


rmzelle:
Thanks!

PR #4939, opened 2020-07-23
Update health-services-research.csl

badgettrg:
Insert a comma between author and year for inline citations.


rmzelle:
> Insert a comma between author and year for inline citations.

Are you sure this is the case?

https://www.hsr.org/system/files?file=media/file/2019/12/HSR-Instructions-for-Authors-2019-Update.pdf says "HSR has recently changed to the AMA citation style. The AMA style uses numerical citations within the text".

That suggest we should just switch the style to use the format of https://github.com/citation-style-language/styles/blob/master/american-medical-association.csl.

badgettrg:
Yes, we should switch to the AMA style. Someone else caught that as well after my submission.

rmzelle:
Okay, great. Thanks!

PR #4940, opened 2020-07-24
Create agora.csl

POBrien333:
via https://forums.zotero.org/discussion/84097/style-request-agora


agora.csl
```diff
@@ -0,0 +1,266 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" page-range-format="expanded" demote-non-dropping-particle="sort-only" default-locale="sv-SE">
```
rmzelle, line 2: @POBrien333, did you set this to Swedish intentionally?

It looks like the journal publishes in multiple Scandinavian languages, but the requestor mentioned Norwegian/norsk bokmål and the journal's domain is Norwegian too.

POBrien333:
Yeah, I just set it to one of the Scandinavian languages and wanted to hear what the OP says. (Forgot to take it out at the end though).

rmzelle:
Thanks!

PR #4941, opened 2020-07-24
Update and rename dependent/chinese-medical-journal.csl to chinese-me…

POBrien333:
…dical-journal.csl

via https://forums.zotero.org/discussion/84331/add-doi-to-chinese-medical-journal#latest


rmzelle:
Thanks!

PR #4945, opened 2020-07-27
Update american-marketing-association.csl

POBrien333:
via https://forums.zotero.org/discussion/84369/journal-of-marketing-style-error


rmzelle:
Thanks!

PR #4946, opened 2020-07-27
add DOI preprint to american-society-for-microbiology.csl

POBrien333:
via https://forums.zotero.org/discussion/83976/change-style-for-journal-of-microbiology-and-biology-education-and-american-microbiology-society

Looks like some ASM journals are adding the DOI now.
Option 1:
Accept this PR and just have it added for all the journals and see if somebody moans about it.
Option 2:
Make two separate styles, one with and one without the DOI.

Let me know what you think and I prepare it.

### DIFF
--- a/american-society-for-microbiology.csl
+++ b/american-society-for-microbiology.csl
@@ -27,7 +27,7 @@
     <category citation-format="numeric"/>
     <category field="biology"/>
     <summary>Style for all American Society for Microbiology journals.</summary>
-    <updated>2015-03-13T23:14:05+00:00</updated>
+    <updated>2020-07-27T17:44:43+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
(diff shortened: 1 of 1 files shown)


adam3smith:
Hmmm: https://mmbr.asm.org/sites/default/files/additional-assets/MMBR-ITA.pdf

>  Note  that  DOIs  are  notneeded for most references. ASM copy editors will automat-ically  insert  DOIs  on  all  references  in  the  CrossRef  andPubMed  databases  during  copyediting. 

I didn't check if that's the same for all journals, but would guess it is?

POBrien333:
Oh, didn't check that file. Thanks for spotting that.
Maybe then let's just leave it as is, no?

MCB for example has the same guidelines file: https://mcb.asm.org/sites/default/files/additional-assets/MCB-ITA.pdf

adam3smith:
We could ask if they're getting told to include DOIs by the journals/editors, in which case I'm always happy to add them, but otherweise I wouldn't be inclined to.

rmzelle:
Thanks!

PR #4948, opened 2020-07-28

isabellypsilva:
Correction of text spacing on paper-conference to fit ABNT requirements.
Setting a condition for appearance of the text "E-book" for books and chapters.
Setting a condition for appearance of the text "Tradução" for books and chapters.

### DIFF
--- a/associacao-brasileira-de-normas-tecnicas-ufrgs.csl
+++ b/associacao-brasileira-de-normas-tecnicas-ufrgs.csl
@@ -96,11 +96,13 @@
     </choose>
   </macro>
   <macro name="translator">
-    <text value="Tradução "/>
-    <names variable="translator" delimiter="; ">
-      <name delimiter="; " sort-separator=" " delimiter-precedes-last="always"/>
-      <et-al font-style="italic"/>
-    </names>
+    <if variable="translator">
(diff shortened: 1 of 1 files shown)


associacao-brasileira-de-normas-tecnicas-ufrgs.csl
```diff
@@ -173,7 +175,7 @@
   </macro>
   <macro name="access">
     <choose>
-      <if type="book chapter">
+      <if type="book chapter" variable="URL" match="all">
```
rmzelle, line 178: @adam3smith, do you know if this generally works well (since an item can only be of a single type)? Or do we needed a nested conditional here?

adam3smith: no, this won't work, you need to test sequentially (the above shouldn't work either, btw., given that mach="all" is the default)

adam3smith:
If this is still relevant after the most recent update (https://github.com/citation-style-language/styles/pull/5058 ) it would need to be rebased (or just close and make a new pull request against the updated style)

adam3smith:
Closing this for inactivity -- happy to repopen (or take a new PR) if needed

PR #4949, opened 2020-07-28
Create journal-of-sport-science-and-medicine.csl

POBrien333:
via https://forums.zotero.org/discussion/84130/style-request-journal-of-sport-science-and-medicine


rmzelle:
Thankss! [sic]

POBrien333:
Thankss for checking the journal name. Just blindly copied from the thread without looking...

PR #4950, opened 2020-07-29
Add collection-editor to author subs deutsche-gesellschaft-fur-psychologie.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/361313/#Comment_361313


bwiernik:
No, this isn’t correct. APA/DGP don’t substitute collection-editor. The poster in that thread is just encountering that the Zotero Report type doesn’t have an editor field. They should enter that organization as an author instead of as an editor or, if they do for some reason want to label the organization as editor, they should enter it in Extra.

PR #4951, opened 2020-07-30
Create critical-reviews-in-solid-state-and-materials-science.csl

POBrien333:
via https://forums.zotero.org/discussion/84195/plz-add-style-critical-reviews-in-solid-state-and-materials-science#latest

.pdf guidelines currently down, so no idea what the other items should look like.


rmzelle:
@POBrien333: [ref_BSMS.pdf](https://github.com/citation-style-language/styles/files/5004947/ref_BSMS.pdf)

rmzelle:
Thanks!

PR #4952, opened 2020-07-31
fix small issue with anthropologie-et-societes.csl

POBrien333:
via emails.
Replace comma with dot after book title.


rmzelle:
Thanks!

PR #4953, opened 2020-07-31
Create korean-journal-of-internal-medicine.csl

POBrien333:
via https://forums.zotero.org/discussion/84251/style-request-korean-journal-of-internal-medicine-kjim


rmzelle:
Thanks!

PR #4955, opened 2020-08-03
Create historical-materialism.csl

POBrien333:
via https://forums.zotero.org/discussion/84092/style-request-historical-materialism-brill


rmzelle:
Thanks!

PR #4956, opened 2020-08-04
Create sorbonne-student-law-review.csl

p-heckler:
Hi,
This is a style proposal for the Sorbonne Student Law Review. It is intended to be a multilingual version of the French legal citation structure since the review accepts submission in a range of languages. This style has been in the making for a while and I am hoping that it is up to the CSL standards!

### DIFF
--- a/sorbonne-student-law-review.csl
+++ b/sorbonne-student-law-review.csl
@@ -11,7 +11,7 @@
       <name>Paul Heckler</name>
       <email>paul.heckler@univ-paris1.fr</email>
     </author>
-    <category citation-format="note">
+    <category citation-format="note"/>
     <category field="law"/>
     <issn>2647-4867</issn>
     <summary>Zotero style of the Sorbonne Student Law Review/Style zotero de la revue juridique des étudiants de la Sorbonne https://sorbonnestudentlawreview.org</summary>
--- a/sorbonne-student-law-review.csl
+++ b/sorbonne-student-law-review.csl
@@ -11,6 +11,7 @@
       <name>Paul Heckler</name>
       <email>paul.heckler@univ-paris1.fr</email>
     </author>
+    <category citation-format="note">
     <category field="law"/>
     <issn>2647-4867</issn>
     <summary>Zotero style of the Sorbonne Student Law Review/Style zotero de la revue juridique des étudiants de la Sorbonne https://sorbonnestudentlawreview.org</summary>
@@ -327,13 +328,7 @@
     </group>
   </if>
   </choose>
-  </macro>
-  <macro name="number-of-volumes">
-    <group delimiter=" ">
-      <text variable="number-of-volumes"/>
-      <text term="volume" plural="true"/>
-    </group>
-  </macro>
+    </macro>
   <macro name="number">
     <choose>
       <if type="legal_case report" match="any">
--- a/sorbonne-student-law-review.csl
+++ b/sorbonne-student-law-review.csl
@@ -1,0 +1,564 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" et-al-min="4" et-al-use-first="1" et-al-subsequent-min="4" et-al-subsequent-use-first="1" initialize-with="." demote-non-dropping-particle="never">
+  <info>
+    <title>Sorbonne Student Law Review - Revue juridique des étudiants de la Sorbonne</title>
+    <title-short>SSLR</title-short>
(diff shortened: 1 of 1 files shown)


p-heckler:
Somehow, the generated citations here do not match the output I obtain when I test the file. Since I have not specified a `name-as-sort-order`, the given name should come first. I also have no idea where the parenthesis is coming from. Should I declare `name` explicitly or can I leave it as is, relying on the default settings? Also, the delimiter that follows quoted titles appears within the quote even though `quote="true"` is applied directly to the title. Again, I don't have this problem when testing it locally and I'm not sure where the code goes wrong here.

sorbonne-student-law-review.csl
```diff
@@ -0,0 +1,575 @@
...
+  <macro name="author-biblio">
+    <choose>
+      <if type="legal_case" match="any">
+        <text variable="authority"/>
+      </if>
+      <else-if match="any" variable="author">
+        <names variable="author">
+          <name name-as-sort-order="all" sort-separator="&#160;">
+            <name-part name="family" text-case="capitalize-first" font-variant="small-caps"/>
+            <name-part name="given" text-case="capitalize-first" font-variant="normal" prefix="(" suffix=")"/>
```
rmzelle, line 94: @p-heckler, you're encapsulating given names with parentheses here, which is probably not what you want.

sorbonne-student-law-review.csl
```diff
@@ -0,0 +1,575 @@
...
+        </names>
+      </else-if>
+      <else-if match="any" variable="editor collection-editor">
+        <choose>
+          <!-- for chapter or paper conference, no author implies the editor is also the author. Name should remain in editor position-->
+          <if type="chapter paper-conference" match="none">
+            <names variable="editor">
+              <name name-as-sort-order="all" sort-separator="&#160;">
+                <name-part name="family" text-case="capitalize-first" font-variant="small-caps"/>
+                <name-part name="given" text-case="capitalize-first" font-variant="normal" prefix="(" suffix=")"/>
```
rmzelle, line 105: Same here.

sorbonne-student-law-review.csl
```diff
@@ -0,0 +1,575 @@
...
+  </macro>
+  <macro name="editor-biblio">
+    <choose>
+      <if match="any" type="chapter paper-conference entry-encyclopedia entry-dictionary">
+        <group delimiter="&#160;">
+          <text term="in" font-style="italic"/>
+          <names variable="editor collection-editor">
+            <name name-as-sort-order="all" sort-separator="&#160;">
+              <name-part name="family" text-case="capitalize-first" font-variant="small-caps"/>
+              <name-part name="given" text-case="capitalize-first" font-variant="normal" prefix="(" suffix=")"/>
```
rmzelle, line 129: And here.

sorbonne-student-law-review.csl
```diff
@@ -0,0 +1,575 @@
...
+      </if>
+    </choose>
+  </macro>
+  <macro name="translator-biblio">
+    <group delimiter="&#160;">
+      <text term="translator" form="short"/>
+      <names variable="translator">
+        <name name-as-sort-order="all" sort-separator="&#160;">
+          <name-part name="family" text-case="capitalize-first" font-variant="small-caps"/>
+          <name-part name="given" text-case="capitalize-first" font-variant="normal" prefix="(" suffix=")"/>
```
rmzelle, line 143: And here.

rmzelle:
> Somehow, the generated citations here do not match the output I obtain when I test the file. ... Again, I don't have this problem when testing it locally and I'm not sure where the code goes wrong here.

The citation previews here suffer from a few formatting bugs, and you shouldn't base your style on them. How are you testing your style locally? Zotero?

You probably want to fix the issue I highlighted, though, as I didn't see any example citations in https://sorbonnestudentlawreview.org/charters/charter_EN where given names are put in parentheses.

p-heckler:
Hi @rmzelle! 
Thank you very much for your input! 
I have fixed the given name encapsulation issue. Since the Sorbonne Law Review does not specify bibliographical entries, I was using the Sorbonne publishing guidelines, which the style mentions as a fallback, and the version I was using had given names encapsulated. That version is no longer online as of this week (they are migrating their website) and the only version I can find does not contain encapsulation, so I will abide to this last version.

> How are you testing your style locally? Zotero?

Yes, I occasionally use Juris-M, but Zotero is my primary testing method and a few people who are using it from my Github repo and giving me feedback are using Zotero as well.

rmzelle:
Thanks!

PR #4957, opened 2020-08-04
Create revista-materia.csl

POBrien333:
via https://forums.zotero.org/discussion/82560/style-request-revista-materia


rmzelle:
Thanks!

PR #4960, opened 2020-08-10
Update historical-materialism.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/361808/#Comment_361808


rmzelle:
Thanks!

PR #4963, opened 2020-08-11
Update ucl-university-college-denmark-vancouver.csl

ckan9445:
New institution name and documentation.


adam3smith:
@rmzelle what do you want to do about the style ID & filename here? How strictly are we trying to avoid changes?

rmzelle:
@adam3smith, since the institute changed names I'd still change the ID.

adam3smith:
Great, thanks!

PR #4974, opened 2020-08-17
Create american-physical-society-et-al

jlaehne:
First author only variant adapted from APS style in order to create short bibliographies.


jlaehne:
@adam3smith: this variant should meet the relevance criteria.

adam3smith:
Sorry, I still don't quite follow what this is for. There is a style guide by APS, mainly for their journals (see https://cdn.journals.aps.org/files/styleguide-pr.pdf ) -- they seem pretty opposed to et al, certainly to a first-author-only version. Where is that coming from? Is it used in any journal/publication?

I'm sorry if this seems like we're pedantic about this, but without some rules, we'll end up with 30 different versions to accommodate different preferences and that's unsustainable. Similar concern for #4975

jlaehne:
As this document covers all aspects of style for papers, I thought it is not specific enough to be mentioned, but the table on page 7 indeed lists exactly this type of citation: 
``Only  if  length  constrained and four or more authors: J. M. Smith _et al._, Phys.  Rev.  B **46**, 1 (1992).``

For publications in the APS journals it is indeed discouraged to use _et al._, but the APS style is quite accepted in physics also for other use cases than publications in the Physical Review journals. For use cases with more limited space - particularly in presentations, but partly also in project reports/applications - it is helpful to have an as short as possible version of the style. That was my motivation for adapting the style and why I think it might be useful also for others.

adam3smith:
How about we make this `american-physical-society-et-al.csl` (with respective changes in title & ID) then and do `et-al-min=4` as per the style guide. It's not the super short version you're suggesting, but it follows explicit guidance, so could also be used for journal publications e.g. where author lists are long, it's almost as short and I'd expect it's just as common as the extra short just first author variant.

jlaehne:
Well, I do still see a motivation for getting the shortest possible citation, but I could adapt this style. Personally, I hardly cite papers with less than 4 authors anyway, unless they are like old groundwork from the 50s.

adam3smith:
Thanks!

PR #4975, opened 2020-08-17
Create american-physical-society-publication-list

jlaehne:
Publication listing including DOI with newline after author, title and journal/publisher details. First author is printed with last name first. Based on the American Physical Society (APS) style where volume is bold and year comes last.


jlaehne:
@adam3smith: as #4974 a more generalized version. I did not find a proper description of the APS style, but at least in physics that is a well known citation style. I extended the summary to describe the main specifics of the style.

jlaehne:
As in #4974, the APS style is quite commonly used in physics also apart from publications in Physical Review journals. For publication lists in e.g. CVs, grant applications, progress reports, ... where there is no limit to the length, but all details should be included, it can be more clear to have additional newlines between authors, title, publisher and DOI. That is why I think the style might be of interest also to others.

adam3smith:
So, I understand how this is individually useful, but the details of the format (beyond the APS core) here still seem fairly arbitrary to me, so I'd lean against taking this for the repository. @rmzelle thoughts?

rmzelle:
> So, I understand how this is individually useful, but the details of the format (beyond the APS core) here still seem fairly arbitrary to me, so I'd lean against taking this for the repository. @rmzelle thoughts?

@jlaehne, I agree with @adam3smith here. If there is a published style guide that describes this kind of APS derivative, we would likely accept a style for that (and please point us to such a style guide if you are aware of one), but we generally don't accept styles formatted based on personal preference in order to keep the repository somewhat manageable. You can always distribute this style variant yourself via other channels. Sorry!

PR #4976, opened 2020-08-17
Create paul-drude-institut-all-authors

jlaehne:
Full author list style used in PDI reports. Adapted from APS style.


adam3smith:
Thanks! Could you clarify what this and #4974 and #4975 are for more precisely? If these are just for reports published by the PDI and not to be used by external authors, it may make more sense to host and maintain them locally rather than as part of the CSL repository. (See also the first bullet point here: https://github.com/citation-style-language/styles/wiki/Criteria-for-Accepting-Styles ). If they _are_ intended for wider use, can you link to a style guide?

jlaehne:
Primary intent is for internal use, but I thought that they might be of interest to others ... though then naming should probably be different to make them more easily findable. 
For example #4974 is basically the shortest possible variant of the American Physical Society style.

PR #4977, opened 2020-08-18
Create acta-ornithologica.csl

POBrien333:
via https://forums.zotero.org/discussion/84555/style-request-acta-ornithologica

### DIFF
--- a/acta-ornithologica.csl
+++ b/acta-ornithologica.csl
@@ -13,6 +13,7 @@
     <category field="biology"/>
     <category field="zoology"/>
     <issn>0001-6454</issn>
+    <eissn>1734-8471</eissn>
     <updated>2020-08-18T08:09:50+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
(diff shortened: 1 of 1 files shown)


acta-ornithologica.csl
```diff
@@ -0,0 +1,143 @@
...
+              <group delimiter=": ">
+                <text term="available at" text-case="capitalize-first"/>
+                <text variable="URL"/>
+              </group>
+            </group>
+          </else-if>
+          <else>
+            <group delimiter=". ">
+              <text macro="title"/>
+              <text variable="container-title" form="short" strip-periods="true"/>
```
adam3smith, line 132: I don't think this is right: The journal abbreviations are _with_ period but there is _no_ period after the journal name, see e.g. `Zool. Zhurnal 72: 85–97.` from the example article

adam3smith:
One quick thing, looks good otherwise - thanks!

acta-ornithologica.csl
POBrien333: Yeah, thanks for spotting that.
I looked at "Acta Ornithol." and didn't get that Acta is a full word...

rmzelle:
Thanks!

PR #4978, opened 2020-08-18
Create chinese-research-writing.csl

wingkei1001:
This CSL is modified on the basis of Turabian 8th full note CSL to support both English and Chinese citation at the same time. The English part basically follows the Turabian formatting standard. For the Chinese part, we comply with the Chinese formatting standard according to _A handbook for research writing : for biblical, theological, and pastoral ministry-related studies_ which is wildly used by Chinese academics in the field of theology. 

For the reference of the handbook, please visit the following link,  [https://www.sbc.edu.sg/sbc-publications/](url)

### DIFF
--- a/chinese-research-writing.csl
+++ b/chinese-research-writing.csl
@@ -1,0 +1,2821 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" 
+  class="note" version="1.0" 
+  default-locale="en-US"
+  demote-non-dropping-particle="sort-only">
+
+  <info>
+    <title>Chinese Research Writing 3rd edition (full note)</title>
+    <id>http://www.zotero.org/styles/chinese-research-writing</id>
+
+    <link href="http://www.zotero.org/styles/chinese-research-writing" rel="self"/>
+    <!--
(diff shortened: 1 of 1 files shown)


adam3smith:
I'm afraid CSL doesn't currently support styles with differently formatting by locale. While that's possible in Zotero, that's due to the fact that Zotero unofficially supports some syntax from the csl-m extension to CSL -- we don't accept these styles to the CSL repository, which only accepts valid CSL 1.0.1 styles. 

I don't know if it makes sense to reformat your style in that light (e.g. to only include the Chinese formatting following the SBC guide) or not.

wingkei1001:
> I'm afraid CSL doesn't currently support styles with differently formatting by locale. While that's possible in Zotero, that's due to the fact that Zotero unofficially supports some syntax from the csl-m extension to CSL -- we don't accept these styles to the CSL repository, which only accepts valid CSL 1.0.1 styles.
Thanks for your quick reply and let me know the issue about that. By the way, do you want is there any progress or roadmap that the official CSL specification is trying to support multi-language in one citation style?  

> I don't know if it makes sense to reformat your style in that light (e.g. to only include the Chinese formatting following the SBC guide) or not.
I'm afraid that it won't help in our case of Chinese users. We probably need to cite both Chinese and English materials in the same paper, so we urge to have a CSL style to support multi-language citation. 

> Edit: But just to be clear -- you don't need us to use or share this style. You should absolutely feel free to share it otherwise.
Yes, I know but it is nice to have an official way to distribute it. It seems that we can't do that at the moment.

adam3smith:
OK, I'll close this than. I think we're looking at multi-lingual layouts for CSL 1.2, hopefully not too far out (as in -- good chance we'll open it for public comment this year) -- is that right @bdarcus @bwiernik  ?

PR #4980, opened 2020-08-20
Changes to match Taxon guidelines for authors

garlandma:
Add "series" macro
Remove spaces between authors' initials
Change (Ed.) and Vol. to lowercase
Remove parentheses around Vol.
Change accessed date format in webpage bibliography entries
Precede page numbers in citations with : instead of p., pp.


rmzelle:
Thanks!

PR #4982, opened 2020-08-21
Update and rename university-college-lillebaelt-harvard.csl to ucl-un…

ckan9445:
…iversity-college-harvard.csl

New institution name and documentation.


rmzelle:
Thanks!

PR #4984, opened 2020-08-23
Don't demote particles in Equine Vet J

bwiernik:
https://forums.zotero.org/discussion/84783/issues-with-prefixed-surname-in-equine-veterinary-journal-format#latest


adam3smith:
Thanks!

PR #4985, opened 2020-08-23
Fix MLA editor translator

bwiernik:


adam3smith:
Great, thanks, that's elegant.

PR #4986, opened 2020-08-23
Update pravnik.csl

OTristanF:
I have added a missing period in ibid without the locator (see the [instructions](https://www.ilaw.cas.cz/casopisy-a-knihy/casopisy/casopis-pravnik/pro-autory/pokyny-pro-autory-citace.html)).

How it currently is:
> Ibidem

How it should be (the proposed change):
> Ibidem.

### DIFF
--- a/pravnik.csl
+++ b/pravnik.csl
@@ -322,7 +322,7 @@
           </group>
         </if>
         <else-if position="ibid">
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks!

PR #4989, opened 2020-08-24
Update taxon.csl a little more

garlandma:
More updates to match Taxon style guidelines:
-Update documentation link
-Change format of website without author: Bold title, remove parentheses around title
-Change thesis format: Change date format, remove parentheses around (Dissertation) statement, reverse publisher and publication place
-Make citation of chapter the same as citation of other locators except page
-Add page before "in:" in collections in bibliography


rmzelle:
Thanks!

garlandma:
Thank you!  I think that's about it for me.  I did try to set the term page-range-delimiter to two hyphens (--), as the Taxon style guide requires for manuscript submissions, but Zotero kept making them three hyphens (---).  That's the last thing that's obviously needed as far as I can tell.

PR #4991, opened 2020-08-25
Update and rename emerald-harvard.csl to emerald-harvard-2.csl

sgusmaob:
Italicized "et al." for inline citations.

### DIFF
--- a/emerald-harvard-2.csl
+++ b/emerald-harvard-2.csl
@@ -61,6 +61,7 @@
     <names variable="author">
       <name name-as-sort-order="all" and="text" sort-separator=", " initialize-with="." delimiter=", " delimiter-precedes-last="never"/>
       <label form="short" prefix=" (" suffix=")" text-case="capitalize-first"/>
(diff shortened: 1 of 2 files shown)


rmzelle:
> Italicized "et al." for inline citations.

Thanks. Is there a reason why you changed the file name of this style? Are there two versions in use? (one with italics, and one without)

If you're submitting a correction, we'd just update the existing style.

sgusmaob:
I was not sure if the original file should be kept. However, Emerald publications are required to have the “et al.” italicized, so, I think only one file should be used.

 

Thank you.

rmzelle:
Okay, thanks. Yes, they clearly require italicized "et al." now, so we should just update the existing style. Thanks!

sgusmaob:
Is it this citation style already updated?

 

Thank you,

 

Sara

rmzelle:
> Is it this citation style already updated?

@sgusmaob, yes, when a pull request is merged your contribution is accepted into the style repository.

PR #4994, opened 2020-08-27
Update circulation-journal.csl

naritatsu:
Minor change according to the "Circulation Journal" citation style (Citations and biblography).


rmzelle:
Thanks!

PR #4995, opened 2020-08-27
Create journal-de-la-societe-des-americanistes.csl

POBrien333:
via https://forums.zotero.org/discussion/84576/style-request-journal-de-la-societe-des-americanistes


rmzelle:
Thanks!

PR #4999, opened 2020-08-30
Update international-energy-agency-organisation-for-economic-co-opera…

POBrien333:
…tion-and-development.csl

via emails


rmzelle:
@POBrien333, thanks. (I was just wondering if this style should be renamed, since the new style guide is hosted by the OECD and doesn't seem to reference the IEA; not sure what the IEA uses now)

POBrien333:
That would make sense. Didn't realise it then, but I first just searched for OECD on the repo and wondered why I didn't get anything. Would certainly make sense then, yes.

PR #5000, opened 2020-08-31
Create chicago-fullnote-bibliography-without-DOI.csl

tubia:
Chicago Manual of style 17th edition (full note) without DOI both in the note and in the bibliography.

### DIFF
--- a/chicago-fullnote-bibliography-without-DOI.csl
+++ b/chicago-fullnote-bibliography-without-DOI.csl
@@ -1,0 +1,1431 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="display-and-sort" page-range-format="chicago">
+  <info>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks -- before we get into the automated validation, is this for something specific? I don't think we'd be very inclined to take this for the repository if it's just a personal preference.

tubia:
Hi @adam3smith thanks for your comment. I will not say that this is for something specific - it is not for a specific journal - but I can see many practical applications, especially in project proposals writing. For example, in the case of the Marie Curie Individual Fellowship, you have to add all the references as footnotes which have an impact on the final number of pages - the application has a page limit - and DOI are not a primary information - you can understand the reference without them.

What do you think?

adam3smith:
Thanks -- I don't think we'll accept this on the repository -- nothing wrong with it, but we have to draw the line somewhere on the types of styles we keep & maintain. There's has been some discussion and some custom styles on the Zotero forums related to the MC Fellowship and there is a range of preferences for this, I don't think it makes sense for us to maintain an inventory of these.

PR #5001, opened 2020-08-31
Create european-journal-of-entomology.csl

POBrien333:
via https://forums.zotero.org/discussion/82820/style-request-european-journal-of-entomology


rmzelle:
Thanks!

PR #5002, opened 2020-09-01
Create european-journal-of-anaesthesiology.csl

POBrien333:
via https://forums.zotero.org/discussion/84511/style-request-european-journal-of-anaesthesiology/


rmzelle:
Thanks!

PR #5003, opened 2020-09-01

libora:
Change was made after consultation with Mr. Kratochvíl. Some journals provide publisher infomation but not publisher place infomration.  Zotero fill it in the Extra field and final output is not correct according to the ČSN ISO 690.


rmzelle:
@libora, thanks! I also took out the commented-out code sections. It's generally better to just delete these, and if you ever wonder what the code used to look like, look into the version history of the file.

PR #5005, opened 2020-09-06
Create journal-of-soil-science-and-plant-nutrition

Tournolive:
I've just created this style by modifying springer-basic-author-date.csl. 
I sorted the inline citation to author, issued (instead of issued, author)
I changed the "et al." threshold in bibliography to 15 authors (there is no specific number in the author guidelines)


rmzelle:
Thanks!

Tournolive:
Thanks a lot!

PR #5008, opened 2020-09-07
Create clinical-management-issues.csl

cappelluti:
CSL Styles for Clinical Management Issues (CMI) journal, publishes by SEEd Medical Publishers


rmzelle:

cappelluti:



Hi, yes I based my CSL on the following one:
<link href="http://www.zotero.org/styles/journal-of-pediatric-gastroenterology-and-nutrition" rel="template"/>

rmzelle:
Thanks!

PR #5009, opened 2020-09-07
Update molecular-therapy.csl

ustervbo:
Fix missing space before 'doi:{url}'

Before:
Hawking, S (1966). Properties of expanding universesdoi:10.17863/CAM.11283.

After:
Hawking, S (1966). Properties of expanding universes doi:10.17863/CAM.11283.


rmzelle:
Thanks!

PR #5010, opened 2020-09-08
Update journal-of-australian-strength-and-conditioning.csl

POBrien333:
via https://forums.zotero.org/discussion/85045/difficulty-editing-style

This was one of my first styles in 2017. You, Rintze, had to tell me how to link to the .csl file to download so the OPs could download the files straight away. I think I learned a lot since then!


rmzelle:
> This was one of my first styles in 2017. You, Rintze, had to tell me how to link to the .csl file to download so the OPs could download the files straight away. I think I learned a lot since then!

🎂 🎆 🌟 

Thanks for all your efforts!

rmzelle:
Thanks!

PR #5011, opened 2020-09-09
Create amerindia.csl

POBrien333:
via https://forums.zotero.org/discussion/84690/style-request-amerindia


rmzelle:
I'm seeing a 403 Forbidden error for both those documentation links. Looks like they might have some maintenance going on (https://www.vjf.cnrs.fr says under construction).

adam3smith:
It's in the IA https://web.archive.org/web/20200726141707/https://www.vjf.cnrs.fr/sedyl/images/Divers_Doc/Instructions_authors.pdf -- let's assume it's coming back, no harm in accepting this for now.

PR #5015, opened 2020-09-11
Create law-technology-and-humans-journal-style.csl

aarontimo:
Removed 'et al' from appearing in footnotes with more than 3 authors but first author still appears.
Webpage is handled like journal-article

### DIFF
--- a/law-technology-and-humans-journal-style.csl
+++ b/law-technology-and-humans-journal-style.csl
@@ -69,20 +69,6 @@
       </names>
     </group>
   </macro>
-  <macro name="secondary-contributors-note">
-    <choose>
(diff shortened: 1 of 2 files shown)


rmzelle:

And it might make sense to fix this style to English. Shall we use British, Australian, or American spelling?

aarontimo:
Thanks @rmzelle for your help in finalising this. Yes, this style is based on http://www.zotero.org/styles/chicago-note-bibliography - apologies for overlooking this.

Please fix it US spelling as that spelling uses the correct double quotation marks.

rmzelle:
Thanks!

PR #5019, opened 2020-09-14
Update pravnik.csl & iso690-full-note-cs.csl

OTristanF:
add the missing period to container-contributors and container-contributors-full


rmzelle:
Thanks!

Issue #5021, opened 2020-09-14
Volume n of <blank> in SBL style


jcuenod:
https://github.com/citation-style-language/styles/blob/4c018d5ef359fb36657112da8f029c1b5c7ef1b2/society-of-biblical-literature-fullnote-bibliography.csl#L381

I'm not sure why I'm getting a footnote like `... vol. 2 of (Place: Publisher ...` when I have a volume number filled in. I believe the line cited is to blame.

I would expect that when the series is not filled out, it would say `... vol. 2. ...` (no of). It's possible I am misunderstanding something in the style guide but I feel like this is a problem with the CSL.

POBrien333:
I have submitted a fix, @jcuenod.
Can you download the style and give it a test:
https://github.com/citation-style-language/styles/raw/3d27fad7280b5d70cbebd672c92cfd8f27330210/society-of-biblical-literature-fullnote-bibliography.csl (right click, save as, install via menu or a double click)

jcuenod:
This appears to remove the volume number entirely from the footnote but then it still says "Vol. 2 of <Publisher>" in the bibliography (I'm testing with a book chapter right now).

I could be wrong but I would have thought the style would be to always show the volume number (assuming one exists) and, optionally, add "of" if there is a series.

adam3smith:
> I could be wrong but I would have thought the style would be to always show the volume number (assuming one exists) and, optionally, add "of" if there is a series.

Yes, this is how this is intended to work. I've tested the now updated style with both books and book chapter with and without series and it works as expected.

PR #5022, opened 2020-09-15

POBrien333:
closes https://github.com/citation-style-language/styles/issues/5021

@adam3smith, can you check I did this right?
I think it was just testing for the wrong variable in the conditional.


adam3smith:
Yes, I think that's right; same logic appears again, so adding the fix here. I have the SBL manual here, want to double-check tomorrow, but pretty sure this is right.

PR #5023, opened 2020-09-15
Delete wirtschaftsuniversitat-wien-interdisziplinares-institut-fur-ve…

georgd:
…rhaltenswissenschaftlich-orientiertes-management.csl

The style was deprecated by the institute.


adam3smith:
What are they using instead? Just deleting styles isn't ideal as it can lead to issues, especially with shared documents using this style.

georgd:
After staff changes, they don't specify one certain style any more. And they don't want a style bearing the institute's name.

If user requests arise, I'll be happy to respond.

adam3smith:
Removing the style is fine, but we'd still want this mapped to some reasonable replacement in `renamed-styles.json` -- maybe pick one of the other WU Wien ones?

georgd:
What does that mapping do? Does it lead to automatically installing a different style on the user’s PC, or even change existing documents to use the new style?

I think, this case is different from a journal switching styles. The institute is switching to a policy where the supervisor says which style the student should use. Students who are still working on a thesis with the current style will probably finish it without switching. For any other user, there’s no definite replacement. Therefore, I think it's best to just remove the style.

adam3smith:
@rmzelle -- I obviously don't have any objection against deleting this, but am worried what happens if a document using this style gets opened by someone without it. I don't remember the mechanism various tools use well enough, though.

adam3smith:
ping @rmzelle

rmzelle:
@georgd, thanks!

@adam3smith, I don't think we need to do anything else in this case. IIRC, I remapped IDs of dependent styles for discontinued journals to the independent parent style of the publisher in the past, but here is no obvious remap target.

PR #5024, opened 2020-09-16
Create cardiff-university-biosi-and-carbs-only-harvard.csl

YoungZoe:
This is new style for Cardiff University's Biosciences and buisness schools. i will also be submitting a welsh version.


adam3smith:
Am I understanding correctly that this should replace the existing BIOSI style https://www.zotero.org/styles?q=id%3Aharvard-cardiff-university-biosi ?

adam3smith:
@rmzelle this looks good to me in general. I'd remove the "only" in title and id but wasn't sure whether we're putting "harvard" at the end or beginning of ids currently.

rmzelle:
> whether we're putting "harvard" at the end or beginning of ids currently.

End.

YoungZoe:
> Am I understanding correctly that this should replace the existing BIOSI style https://www.zotero.org/styles?q=id%3Aharvard-cardiff-university-biosi ?

Yes please remove harvard-cardiff-university-biosi style

YoungZoe:
Why have the styles below been changed - they have nothing to do with Cardiff University?  If they based there styles on our old one they probably wanted it to remain as it was?
demographic-research.csl (modified style; unchanged output for sample items)
harvard-university-of-kent.csl (modified style; unchanged output for sample items)

adam3smith:
> Why have the styles below been changed - they have nothing to do with Cardiff University? If they based there styles on our old one they probably wanted it to remain as it was?

They pointed to the old style as a template, which is just a way of tracking which style the authors modified to create them, it has no impact on what the style does. Since we deleted the old style, that link was no longer valid.

YoungZoe:
> > Why have the styles below been changed - they have nothing to do with Cardiff University? If they based there styles on our old one they probably wanted it to remain as it was?
> 
> They pointed to the old style as a template, which is just a way of tracking which style the authors modified to create them, it has no impact on what the style does. Since we deleted the old style, that link was no longer valid.

Thank you, that makes sense.

adam3smith:
Thanks!

PR #5025, opened 2020-09-16
Create harvard-yn-unig-biosi-a-carbs-prifysgol-caerdydd.csl

YoungZoe:
This is a welsh version of cardiff-university-biosi-and-carbs-only-harvard.csl so is dependent but i have only just submitted the engilish version so wasn't sure how to link the two.


adam3smith:
Apart from the language, is this identical to the English style? Then we'd just publish one without default locale and users can select

YoungZoe:
> Apart from the language, is this identical to the English style? Then we'd just publish one without default locale and users can select

Yes it is identical to the English. In Wales we have a requirement to produce as much as possible in both languages. How would users know to select the language?

adam3smith:
> How would users know to select the language?

You'd set the language when selecting the citation style. It defaults to your OS language, but you can switch -- avoiding duplicate styles to maintain is a big advantage, so we'd much prefer that

YoungZoe:
> > How would users know to select the language?
> 
> You'd set the language when selecting the citation style. It defaults to your OS language, but you can switch -- avoiding duplicate styles to maintain is a big advantage, so we'd much prefer that

Okay that's fine. Thanks you.

adam3smith:
Folding into https://github.com/citation-style-language/styles/pull/5024

PR #5035, opened 2020-09-21
create new independent style for deutsches-arzteblatt.csl

POBrien333:
via https://forums.zotero.org/discussion/84813/style-error-deutsches-arzteblatt#latest


adam3smith:
Thanks!

PR #5040, opened 2020-09-25
Create physiologia-plantarum.csl

POBrien333:
via https://github.com/citation-style-language/journals/pull/75#event-3805768094

And https://forums.zotero.org/discussion/75030/style-request-physiologia-plantarum


adam3smith (changes requested):
One comment in-line. You'll want to accomodate standard work types without authors

rmzelle:
@POBrien333, thanks!

PR #5042, opened 2020-09-25
Create institut-de-recherches-archeologiques-preventives.csl

Clem-F:


rmzelle:
Thanks for your submission. Is there an online style guide available for this style? Per https://github.com/citation-style-language/styles/wiki/Style-Requirements#10---documentation-link, we typically require each CSL style to include a link to the format it describes.

Clem-F:
Thanks for your comments.
I added the two missing lines : 
  - \<link [...] rel="template"/>
  - \<link [...] rel="documentation"/>

adam3smith:
(The preview shown here is incorrect, so don't worry about that)

rmzelle:
Thanks!

PR #5049, opened 2020-09-28
Create cureus.csl

dryphi:
Vancouver style modified for Cureus journal. Modified from BMJ but may need to be independent.


dryphi:
I used the BMJ style as a template for this. Will try to edit the other pull request. This one may be redundant.

PR #3796, opened 2018-11-23
New style for University of Lausanne

nchachereau:
This style is developed mainly for the history department of the University of Lausanne, Switzerland. It is based on another style (infoclio), and adapted to fit the citation style that has been used for years in teaching and students' papers in the department. 

More information is available on our separate Github repository (French): https://github.com/LausanneCitationStyle/Lausanne, where we also write the [documentation](https://github.com/LausanneCitationStyle/Lausanne/blob/master/documentation/StyleDescription.md) intended for the department's website.

PR #5050, opened 2020-09-29
New version of the style for the History Department at the University of Lausanne

nchachereau:
The history department of the University of Lausanne has changed its bibliography guidelines, noticed a bug in the CSL style and created a better documentation.

Hence this update to the "universite-de-lausanne-histoire.csl" style (originally submitted in #3796).

We'd be greatful if these changes could be merged in the official repository.

Many thanks for keeping this great project alive and running!

### DIFF
--- a/universite-de-lausanne-histoire.csl
+++ b/universite-de-lausanne-histoire.csl
@@ -5,7 +5,7 @@
     <id>http://www.zotero.org/styles/universite-de-lausanne-histoire</id>
     <link href="http://www.zotero.org/styles/universite-de-lausanne-histoire" rel="self"/>
     <link href="http://www.zotero.org/styles/infoclio-fr-smallcaps" rel="template"/>
-    <link href="https://github.com/LausanneCitationStyle/Lausanne" rel="documentation"/>
+    <link href="https://lausannecitationstyle.github.io/support/" rel="documentation"/>
     <author>
       <name>Nicolas Chachereau</name>
       <email>nicolas.chachereau@unil.ch</email>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks! Could you elaborate on the "no comma" change -- that looks very odd to me (I'd expect readers to confuse first and last names as a result) and isn't reflected in the linked styleguide, which still shows a comma. How is a sample citation supposed to look?

grandjeanmartin:
Hi! We wanted to update the guide only after the style had been accepted and updated (to avoid updating the guide in vain if the update process was not successful or immediate). I've just updated the [PDF link](https://github.com/LausanneCitationStyle/Lausanne/raw/master/documentation/Guide_Bibliographique_UNIL_HIST.pdf) so that you can already access the new version. Regarding the comma question, the last name is displayed in small caps by the style anyway, so confusion with the first name is not possible. Besides, I don't think this is the only style not to put a comma? (It is even the case in the "note" format of the Chicago manual of style.)

nchachereau:
Thanks for the feedback! To elaborate a bit on Martin's answer, it is maybe useful to point out that in history, and especially in French-speaking historiography, there is no standard agreed upon, so that readers are quite flexible, so that the change would not bring confusion.

Why the change? The instructors in the department found that the comma was redundant because the small capitals already differenciated between first and last names. This style is also quite common in our field (see for instance the [guidelines of a Swiss editor specialized in historical studies](https://www.alphil.com/pdf/004_instructions_bibliographie_et_notes.pdf)).

adam3smith:
OK, thanks for the explanation, just wanted to make sure this was intentional.

adam3smith:
This, btw., isn't the case:
> It is even the case in the "note" format of the Chicago manual of style or in the bibliography entry when displaying co-authors

In Chicago Note, that's "Firstname Lastname" but "Lastname, Firstname" -- that's a fairly common convention and I wanted to make sure that's not what you were after.

nchachereau:
Thanks a lot!

PR #5051, opened 2020-09-30
Add new style china-national-standard-gb-t-7714-2015-author-lowercase.csl

8clock:
Add china-national-standard-gb-t-7714-2015-author-lowercase.csl

It was made by SLKun. I am requesitng to add it to styles.


rmzelle:
@8clock, thanks. Can you tell us what this style variant is for? Is there a journal or institute that requests references to be formatted with lowercase (or at least not all-uppercase) author names?

It also looks like this CSL style still capitalizes full names.

rmzelle:
Hi, we're closing this issue because we haven't heard from you in a long time. Note that we can always reopen this issue if you get back to us.

PR #5052, opened 2020-09-30
New style - german-yearbook-of-international-law.csl

mikerd09:
New style for the German Yearbook of International Law.


adam3smith:
Thanks! Did you base this on an existing style? 
Also, I removed sorting in footnotes since that'd be very unusual, but I mainly did that also to trigger the automatic check, which somehow didn't happen previously, so happy to put that back in.

mikerd09:
Hey thanks for the comments!

Yes this is based on the style _European Journal of International Law_ that is already in the depository, should I reference it anywhere?

Also, this being my first style, how do I avoid sorting in footnotes in the future? I'm not sure how I did that in the first place, so it would be nice to know how not to do that again haha.

Thanks!

adam3smith:
Thanks -- I've added the template. See my commit above: https://github.com/citation-style-language/styles/pull/5052/commits/c751b284796c03530f77fd75819fa065975ab756 on removing sorting from footnotes.

PR #5053, opened 2020-10-01
biotechnology-journal.csl

BiotechnologyJournal:


adam3smith:
Something went quite wrong here -- the current file you have in this pull request is 2-3 CSL styles mixed (to the extent that there is code, it also needs a lot of cleaning). 

This is for the Wiley journal? They say it's just APA with numeric citations, which we should be able to add quickly (CC @bwiernik )

BiotechnologyJournal:
Thanks, that would be great. The style is the same as in bioessays.csl



bwiernik:
I'll submit this style tomorrow. Traveling today. I'll add it for this journal and Bioassays.

adam3smith:
@BiotechnologyJournal  -- are you affiliated with the journal? When I look at their styleguide, it looks nothing like APA/Bioessays (see https://onlinelibrary.wiley.com/pb-assets/assets/18607314/BTJ_Instructions_to_authors_2018-1519640119293.pdf

bwiernik:
BioEssays did also just update their style to APA numeric, but haven't updated their style guide.

PR #5056, opened 2020-10-04
Update mcgill-en.csl

gspanglett:
Added dictionary and encyclopedia style types.
Added support for Codes - see s E-2.3 in the McGill Guide.
Corrections on some punctuation issues.

### DIFF
--- a/mcgill-en.csl
+++ b/mcgill-en.csl
@@ -42,11 +42,11 @@
         <single>at para</single>
         <multiple>at paras</multiple>
       </term>
-	  <term name="sub verbo" form="short">
+      <term name="sub verbo" form="short">
         <single>sub verbo</single>
         <multiple>sub verbis</multiple>
       </term>
-	  <term name="part" form="short">
+      <term name="part" form="short">
         <single>Art</single>
         <multiple>Arts</multiple>
       </term>
@@ -172,7 +172,7 @@ all the other tests should have been done... -->
     <text macro="series-info" prefix=" "/>
     <text macro="edition" prefix=", "/>
     <text macro="publisher-place-year"/>
-    <text variable="page-first" prefix=" "/>     
+    <text variable="page-first" prefix=" "/>
   </macro>
   <macro name="render-encyclopedia">
     <group delimiter=" ">
@@ -184,7 +184,7 @@ all the other tests should have been done... -->
     <text macro="series-info" prefix=" "/>
     <text macro="edition" prefix=", "/>
     <text macro="publisher-place-year"/>
-    <text variable="page-first" prefix=" "/>    
+    <text variable="page-first" prefix=" "/>
   </macro>
   <macro name="render-article-journal">
     <group delimiter=" ">
@@ -280,7 +280,7 @@ all the other tests should have been done... -->
           </group>
           <text variable="number-of-volumes"/>
           <group delimiter=" ">
-	    <text variable="number"/>
+            <text variable="number"/>
             <text macro="issued-long" prefix="(" suffix=")"/>
           </group>
         </if>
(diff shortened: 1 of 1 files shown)


mcgill-en.csl
```diff
@@ -38,13 +41,20 @@
       <term name="ordinal-11">th</term>
       <term name="ordinal-12">th</term>
       <term name="ordinal-13">th</term>
+	  <term name="open-quote">"</term>
```
adam3smith, line 44: actual straight quotes? That seems wrong. I don't have access to the guide, but the samples at Calgary and Carleton both have regular (typographic) quotes.

mcgill-en.csl
```diff
@@ -17,14 +17,17 @@
       <email>f.martin-bariteau@umontreal.ca</email>
       <uri>http://f-mb.github.io/cslegal/</uri>
     </contributor>
+
```
adam3smith, line 20: did something go wrong here? No reason to space these out otherwise.

mcgill-en.csl
```diff
@@ -275,18 +342,34 @@ all the other tests should have been done... -->
...
-      </group>
-      <group delimiter=" ">
-        <text variable="container-title"/>
-        <date form="text" variable="issued" date-parts="year"/>
-      </group>
-      <text variable="section"/>
-    </group>
+	  
+	    <choose>
+		  <if variable="references" match="any">
```
adam3smith, line 347: something weird is going on here -- you're checking for `references` but then not printing them but then you are calling them in the `else` where they can't appear

adam3smith (commented):
I started reviewing but I'm wondering if something went wrong with this PR? There are a lot of things included that aren't mentioned in the description, the indentation is all out of whack, some things that are obviously wrong, the inclusion of all the `note` stuff, etc. Could you double check and if this is indeed right, fix the formatting and explain some of your other changes?

mcgill-en.csl
gspanglett: Hi, and yes.  I will remove the spaces.

mcgill-en.csl
gspanglett: Hi, and yes - I should have removed this before finalizing.  Will remove.

mcgill-en.csl
gspanglett: Hi, this is because of a quirk of Canadian Law being both common law and civil law.  
Normally a statute looks  like "title year, section" (Criminal Code, RSC 1985, C 8, art 2).  But in civil law, "Codes" are structured completely differently: "section title (year)" (Art 2 CCQ (1985).)  
We need both citation styles, but I only have the one type.  It requires an if-then, and I opted to use references.  It can only check for the presence of a value and cannot operate on the values themselves.  I didn't have a lot of fields to work with and references was available.
When there is a value in references, it formats as a Code; otherwise it is as an english statute.
I hope this explains it.

However, I will remove the references variable from the else statement, since it must be null there.

S

gspanglett:
Hi, and thanks again for taking the time to review and comment - it is greatly appreciated.

As a general comment on the changes to the citation layout section - the old csl used to render the "pinpoint" and note variables at the end of all citation types; however, to accommodate civil law (Quebec) statute references - which are completely different from english statute citations - required that pinpoint appear in different positions.

The other change to the "render-legislation" macro is related - it is to handle to different styles used in Canada's bijural system.  I left a more detailed comment above, under your specific comment.

I hope this helps - I have made the changes and re-checked the file.  Thanks again.

rmzelle:
@gspanglett, thanks! For future pull requests, note that we recommend to always run your style through https://formatter.citationstyles.org/ to ensure the style follows our standard indentation.

PR #5059, opened 2020-10-06
Create apa-no.csl

haarbakke:
This is a Norwegian version of APA7 where comma before & has been removed, changes made to lines 4, 5, 6, and 7 adjust as appropriate. 
The comma has been removed before & by changing delimiter-precedes-last="never" in lines 331, 336, 360, 364, and 368.
Also in line 1474 delimiter-precedes-last="never" has been added.


adam3smith:
I'm wondering if we should call this "Norwegian" -- there's nothing Norwegian about this other than the lack of comma, which isn't particular to Norwegian*  so maybe a more generic title makes more sense?


\* as discussed over at Zotero, it's actually just a misunderstanding and they'd save us a bunch of trouble by getting rid of it, but the *stated* rationale applies to most European languages other than English

haarbakke:
No problem at all, could "(no comma before ampersand)" or something similar do? There is another style called "(no ampersand)"

adam3smith:
@bwiernik @rmzelle thoughts on the naming?

bwiernik:
Where are the style guidelines that call for this formatting?

Is this just for the Psychology Compass journal or a broader organization? I'd really rather not have a general "APA" style here given that it is a such a minor change (and misunderstanding of the function of the comma) if its use is limited to a journal.

adam3smith:
It's not a journal -- Citation Compass is a site that describes APA style that's used by a number of Norwegian universities (https://kildekompasset.no/en/home/) and there is an official APA 7th manual maintained by, if I understand correctly, the consortium of ICT services of Norwegian universities https://www.unit.no/tjenester/norsk-apa-referansestil which has this rule -- but my recollection is that we've also had people from other European countries asking about this?

haarbakke:
This is the official Norwegian guidelines for APA7, see ( in Norwegian): https://www.unit.no/en/node/560 and the style guide: https://www.unit.no/en/media/1886/download It is a national standard. So adam3smith is correct about the Norwegian part.

rmzelle:
> thoughts on the naming?

Maybe "Norsk APA-manual" (after https://www.unit.no/en/media/1886/download), with either " (Norwegian - Bokmål)" or " (Norwegian - Nynorsk)" as suffix, depending on the locale, if it's fixed to one of these two.

bwiernik:
That sounds good to me.

adam3smith:
OK, but what to do about the default locale? If we just set this to "no" it doesn't work in Zotero (not sure about other implementations), so we'd keep this without default locale?

bwiernik:
Yeah, I think so. Might be worth a discussion about permitting default-locale to give a list of allowed locales (e.g., many universities do both their local language and English)?

haarbakke:
Would a good name then be "Norsk APA-manual" without a suffix because we make it language neutral. Should I update the patch to reflect that?

bwiernik:
Yes, please do.

haarbakke:
@bwiernik or @adam3smith thoughts on what to do next?

bwiernik:
Looks good to me

haarbakke:
Thank you @bwiernik but I do not know whether I am allowed to push this so that it becomes available for Zotero users or how to do it, please guide me so that this patch and my other patch will not stay in pull request because of my ignorance.

bwiernik:
It's waiting for merging by @adam3smith or @rmzelle

haarbakke:
OK, thank you, then I know it is not dependent on me. Also, thank you for all your help.

rmzelle:
This should be good to merge on my end once the tests complete (Travis CI seems to be having a bit of a backlog recently: https://www.traviscistatus.com/#month).

@haarbakke, @bwiernik, my only question is whether it would make sense to eliminate most of the locale sections. Is it likely this style will ever be used beyond the two Norwegian locales and English? If not, we can make it a little bit more compact.

haarbakke:
adam3smith stated earlier in this tread that "the stated rationale applies to most European languages other than English" so I would assume that this means that the locales section should be as it is. We would need at least Norwegian, Danish and Swedish. adam3smith there also questioned the name of the style, but that is up to you. I can live fine with the style as it is now.

bwiernik:
I think we can assume that only Scandinavian locales will use the style, so let's remove the locales that aren't English/Norway/Danish/Swedish.

rmzelle:
> I think we can assume that only Scandinavian locales will use the style, so let's remove the locales that aren't English/Norway/Danish/Swedish.

Yeah, I think we make that assumption. Even if the rationale isn't specific to Norwegian, I don't think people will have the expectation that they can use a style named "Norsk APA-manual" for non-Norwegian locales (with the exception of English).

@haarbakke, thanks!

PR #5060, opened 2020-10-06
Create citation-compass-apa-note-no.csl

haarbakke:
This is a Norwegian version of Citation Compass (Kildekompasset) - APA (note) where comma before & has been removed, changes made to lines 4, 5, and 6 adjust as appropriate.
The comma has been removed before & by changing delimiter-precedes-last="never" in lines 65, 70, 94, 98, and 102.
Also in line 1208 delimiter-precedes-last="never" has been added.

### DIFF
--- a/citation-compass-apa-note-no.csl
+++ b/citation-compass-apa-note-no.csl
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="never" page-range-format="expanded">
   <info>
-    <title>Citation Compass (Kildekompasset) - APA (note) (Norwegian)</title>
+    <title>Norsk APA-manual (note)</title>
     <id>http://www.zotero.org/styles/citation-compass-apa-note-no</id>
     <link href="http://www.zotero.org/styles/citation-compass-apa-note-no" rel="self"/>
     <link href="http://www.zotero.org/styles/apa" rel="template"/>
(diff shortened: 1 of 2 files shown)


adam3smith:
Can the other Citation Compass style then be set to default to English? And how about this one -- because there are two Norwegian locales, we can't set a language default?

haarbakke:
We could probably choose Norsk bokmål which is used by the majority of Norwegian language users, but I should probably check the locales files that applies to styles and see if there are any differences. If not it doesn't matter which Norwegian locale is used. Can you give me a link to where I can find the necessary files?

haarbakke (commented):
I have updated the default locale to nb-NO which is the locale that the majority of Norwegian Zotero users will use. Just for the record, default locale does not mean that it is not possible to change the locale to e.g. Norsk nynorsk or Danish?

Secondly, I am OK with setting a default locale to English for the original Citation Compass style if that is OK for all the others.

haarbakke (commented):
Fixed an error committed in patch 2

haarbakke:
Fixed an error in patch-2

adam3smith:
> I have updated the default locale to nb-NO which is the locale that the majority of Norwegian Zotero users will use. Just for the record, default locale does not mean that it is not possible to change the locale to e.g. Norsk nynorsk or Danish?

It depends on the software, but in Zotero the default locale fixes the style to that language and can't be changed and I think that was our original vision, so if that doesn't work, we'd probably want to keep it language-neutral

haarbakke:
OK, I will update the csl file to be language neutral and let the name be "Citation Compass (Kildekompasset) - APA (note) Norwegian" if that is OK.

haarbakke:
@adam3smith thoughts on what to do next?

bwiernik:
We would probably want to name this style Norsk APA-Manual (note) to follow the other style from Citation Compass?

haarbakke:
That is OK for me. I made an attempt to rename earlier, but that attempt failed. Would you mind renaming the style or guide me how to do it?

PR #5066, opened 2020-10-11
Create offa.csl

nmueller18:
Style following the rules of the archaeological journal "Offa - Berichte und Mitteilungen zur Urgeschichte, Frühgeschichte und Mittelalterarchäologie".

### DIFF
--- a/offa.csl
+++ b/offa.csl
@@ -223,19 +223,20 @@
                 <text variable="publisher"/>
               </if>
               <else>
-                <choose>
-                  <if match="any" variable="original-date original-publisher-place">
-                    <text variable="original-publisher-place"/>
-                    <date date-parts="year" form="text" variable="original-date" prefix=" "/>
-                    <text value="; "/>
-                  </if>
-                </choose>
                 <text variable="publisher-place"/>
               </else>
             </choose>
             <text macro="date"/>
           </group>
           <text macro="edition"/>
+          <choose>
+            <if match="any" variable="original-date original-publisher-place">
+              <group prefix=" [Originalausgabe: " suffix="]">
+                <text variable="original-publisher-place"/>
+                <date date-parts="year" form="text" variable="original-date" prefix=" "/>
+              </group>
+            </if>
+          </choose>
         </group>
       </if>
     </choose>
--- a/offa.csl
+++ b/offa.csl
@@ -20,7 +20,7 @@
     <category field="anthropology"/>
     <issn>0078-3714</issn>
     <summary>Author-date style meeting citation specifications provided by Offa. Based on the JNA style.</summary>
-    <updated>2020-10-2T00:09:05+00:00</updated>
+    <updated>2020-10-12T00:09:05+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
--- a/offa.csl
+++ b/offa.csl
@@ -308,14 +308,7 @@
     </choose>
   </macro>
   <macro name="doi">
-    <choose>
-      <if match="any">
-        <group>
-          <text value=". DOI: https://doi.org/"/>
-          <text variable="DOI"/>
-        </group>
-      </if>
-    </choose>
+  <text variable="DOI" prefix=". DOI: https://doi.org/"/>
   </macro>
   <macro name="review">
     <choose>
@@ -417,6 +410,7 @@
             </else>
           </choose>
           <choose>
+          <if variable="DOI" match="any">
               <text macro="doi"/>
             </if>
             <else>
--- a/offa.csl
+++ b/offa.csl
@@ -90,6 +90,7 @@
     </choose>
   </macro>
   <macro name="contributors">
+    <group suffix=", ">
     <choose>
       <if variable="author">
         <names variable="author">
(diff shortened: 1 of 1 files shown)


nmueller18 (commented):
Correct wrong date format.

offa.csl
```diff
@@ -0,0 +1,438 @@
...
+    </layout>
+  </citation>
+  <bibliography hanging-indent="false" line-spacing="1" entry-spacing="1">
+    <sort>
+      <key macro="contributors-short-biblio"/>
+      <key variable="issued"/>
+    </sort>
+    <layout>
+      <group suffix=".">
+        <group display="left-margin" suffix=": ">
```
adam3smith, line 397: using left margin doesn't seem right. The bibliography just seems to start with the short author and have a hanging indent

offa.csl
```diff
@@ -0,0 +1,438 @@
...
+              </group>
+            </if>
+          </choose>
+        </group>
+      </if>
+    </choose>
+  </macro>
+  <macro name="doi">
+    <choose>
+      <if match="any" variable="DOI">
```
adam3smith, line 319: you test for this twice (here and in the bibliography) but don't actually need to at all: the group acts as a test and will only print if a DOI exists

offa.csl
```diff
@@ -0,0 +1,438 @@
...
+      </if>
+      <else-if variable="editor">
+        <text macro="editor"/>
+      </else-if>
+    </choose>
+    <choose>
+      <if match="any" variable="title">
+        <choose>
+          <if match="any" variable="editor author">
+            <text value=", "/>
```
adam3smith, line 107: Couldn't you just put a group with a comma delimiter around contributors and title in the bibliography? Much neater than this

offa.csl
```diff
@@ -0,0 +1,438 @@
...
+    <date variable="issued">
+      <date-part name="year"/>
+    </date>
+  </macro>
+  <macro name="place-date">
+    <choose>
+      <if type="book thesis chapter paper-conference" match="any">
+        <choose>
+          <if match="none" variable="edition">
+            <text value=""/>
```
adam3smith, line 222: this doesn't seem to do anything

offa.csl
```diff
@@ -0,0 +1,438 @@
...
+              <if type="thesis" match="any">
+                <text variable="genre" suffix=" "/>
+                <text variable="publisher"/>
+              </if>
+              <else>
+                <choose>
+                  <if match="any" variable="original-date original-publisher-place">
+                    <text variable="original-publisher-place"/>
+                    <date date-parts="year" form="text" variable="original-date" prefix=" "/>
+                    <text value="; "/>
```
adam3smith, line 237: use a group with a suffix. Generally avoid `text value` for most purposes.

offa.csl
```diff
@@ -0,0 +1,438 @@
...
+        </names>
+        <text variable="reviewed-title" prefix=", "/>
+      </if>
+    </choose>
+  </macro>
+  <macro name="encyclopedia-article">
+    <text variable="container-title"/>
+    <number prefix=" " variable="volume"/>
+    <choose>
+      <if match="none" is-uncertain-date="issued">
```
adam3smith, line 342: This is valid, but what's the purposes of this construct? Why would you not print uncertain dates?

adam3smith (changes requested):
Thanks, some questions, some change requests, mostly to tidy/simplify the code.

offa.csl
nmueller18: Corrected. Also corrected in style journal-of-neolithic-archaeology, as the offa-style derives from the latter.

offa.csl
nmueller18: Thanks for clarifying! Corrected. Also corrected in style journal-of-neolithic-archaeology, as the offa-style derives from the latter.

offa.csl
nmueller18: I am also not sure what to you with this ...

offa.csl
nmueller18: Again, I am also not sure what this was intended for ...

offa.csl
nmueller18: Modifications here are difficult to test for me as somehow "original-date" and "original-publisher-place" are not valid fields in my Zotero-Bib.

offa.csl
```diff
@@ -0,0 +1,430 @@
...
+  </macro>
+  <macro name="doi">
+    <choose>
+      <if match="any">
+        <group>
+          <text value=". DOI: https://doi.org/"/>
+          <text variable="DOI"/>
+        </group>
+      </if>
+    </choose>
```
adam3smith, line 318:
```suggestion
  <text variable="DOI" prefix=". DOI: https://doi.org/"/>
```

offa.csl
```diff
@@ -0,0 +1,430 @@
...
+              <text macro="webpage-date"/>
+              <text macro="place-date"/>
+              <text macro="journal-date"/>
+              <text macro="locators-chapter"/>
+              <text macro="locators-article"/>
+            </else>
+          </choose>
+          <choose>
+              <text macro="doi"/>
+            </if>
```
adam3smith, line 421:
```suggestion
          <choose>
          <if variable="DOI" match="any">
              <text macro="doi"/>
            </if>
```

sorry, you do need this here because of the else for the URL

adam3smith (commented):
See my changes here to fix the validation & simplify. Same idea for your other style

offa.csl
adam3smith: Presumably you mean Zotero, not Zotero-bib (i.e. zbib.org)? You can just add those to the extra field and they'll get picked up by Zotero (true for any CSL variable):
```
original-date: 1776
original-publisher: Glasgow University Press
```

nmueller18: I did not know this. Very handy. Many thanks for pointing this out!

adam3smith:
Great, thanks!

PR #5069, opened 2020-10-15

kaitlyndoesresearch:
Removed "DOI" from DOI prefix.

### DIFF
--- a/association-for-computing-machinery.csl
+++ b/association-for-computing-machinery.csl
@@ -88,7 +88,7 @@
   <macro name="access">
     <choose>
       <if variable="DOI">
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks! What guidance is this based on? Why I think this update makes sense, the guidance I'm seeing at https://www.acm.org/publications/authors/reference-formatting still has the `DOI: ` label.

kaitlyndoesresearch:
Ah, my bad! I suppose I overlooked it in the citations. Consider the
request rescinded.

Kaitlyn

adam3smith:
Thanks -- obviously it's a weird rule, but we'll keep it there until ACM removes it from the guidelines.

PR #5075, opened 2020-10-21
Update mcgill-fr.csl

gspanglett:
Minor correction to formatting of legislation - punctuation after statute title.  Also ran style through CSL Style Formatter for consistency.


adam3smith:
Thanks!

PR #5076, opened 2020-10-21
Update mcgill-en.csl

gspanglett:
Minor correction to formatting of legislation - punctuation after statute title.  Also ran style through CSL Style Formatter for consistency.


adam3smith:
Thanks -- I have a small preference for taking analogous updates for the fr and en version of the style in a single PR in the future: Easier to review and cleaner in the commit history.

PR #5079, opened 2020-10-25
Update geografia-fisica-e-dinamica-quaternaria.csl

RicceR-unipi:
DOI section is added to the end of the citation.

### DIFF
--- a/geografia-fisica-e-dinamica-quaternaria.csl
+++ b/geografia-fisica-e-dinamica-quaternaria.csl
@@ -19,7 +19,7 @@
     <category field="geology"/>
     <issn>0391-9838</issn>
     <eissn>1724-4781</eissn>
-    <updated>2020-03-28T09:24:47+00:00</updated>
+    <updated>2020-10-25T18:28:19+00:00</updated>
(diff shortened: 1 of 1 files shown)


adam3smith:
What is this based on? I'm not seeing DOIs in the sample issue they link to from the author information.
If they now want them (which would be great), why not leave the current group as it, with the suffix="." and just add the DOI macro outside of the group once, so it applies to all item types.

RicceR-unipi:
> why not leave the current group as it, with the suffix="." and just add the DOI macro outside of the group once, so it applies to all item types.

you're right, I didn't think about this solution yesterday. I will immediatelly upgrade the file following your suggestion

adam3smith:
Thanks! How about the reason for including the DOI in the first place? Where is that coming from?

RicceR-unipi:
Sorry, I forgot to reply to this part of your comment; the Editors didn't update the sample issue file yet but the DOIs have been begun to be inserted in the issues of the last volume (e.g. [first article](http://www.glaciologia.it/wp-content/uploads/FullText/full_text_42_2/02_GFDQ_42_2_DiRita_203-214.pdf) of the vol 42(1) and [second article](http://www.glaciologia.it/wp-content/uploads/FullText/full_text_42_1/02_GFDQ_42_1_Fierz_17-26.pdf) of the vol. 42(2))

adam3smith:
Great, thanks!

PR #5088, opened 2020-10-30
Create tu-dortmund-agvm.csl

millsptf:


adam3smith:
Thanks! Is there a public style guide you could link to? (Antwort auf Deutsch ist auch OK)

millsptf:
> 
> 
> Thanks! Is there a public style guide you could link to? (Antwort auf Deutsch ist auch OK)

Thanks for your comment. I have added a link to the homepage where the style will be found. A documentation isn't available so far.

technische-universitat-dortmund-ag-virtual-machining.csl
```diff
@@ -0,0 +1,141 @@
...
+          <text macro="authors"/>
+          <text macro="date" />
+        </group>
+        <text macro="point-locators"/>
+      </group>
+    </layout>
+  </citation>
+  <bibliography hanging-indent="false" et-al-min="11" et-al-use-first="7" subsequent-author-substitute="---" entry-spacing="0">
+    <sort>
+      <key macro="sort-date" sort="descending"/>
```
adam3smith, line 127: This seems very odd -- the bibliography is sorted by date, not by author names?
Also, the date as currently specified in the `date-sort` macro is day-month-year, leading to completely unpredictable sorting. Is that really what's expected?

adam3smith:
@millsptf -- Looks mostly good. Please check that the naming (which follows our standard conventions) is OK with you and see my one question about bibliography sorting above.

adam3smith:
@millsptf Thanks! One follow-up based on your last answer -- could you clarify what this is for? We're happy to host a style for students of your chair to use for term papers (or something along the lines), but we generally don't host styles used mainly for internal purposes such as formatting online bibliographies, interact with an API, etc. (those can typically be hosted locally without any issues).

millsptf:
> 
> 
> @millsptf Thanks! One follow-up based on your last answer -- could you clarify what this is for? We're happy to host a style for students of your chair to use for term papers (or something along the lines), but we generally don't host styles used mainly for internal purposes such as formatting online bibliographies, interact with an API, etc. (those can typically be hosted locally without any issues).

millsptf:
> 
> 
> > @millsptf Thanks! One follow-up based on your last answer -- could you clarify what this is for? We're happy to host a style for students of your chair to use for term papers (or something along the lines), but we generally don't host styles used mainly for internal purposes such as formatting online bibliographies, interact with an API, etc. (those can typically be hosted locally without any issues).

sorry for the late response. the style shall be used by our students in a seminar where they get publications provided by us. it makes sense for us to use this kind of citation-style because we can easily see which one they used.

rmzelle:
Okay, thanks!

PR #5089, opened 2020-10-31
Add APA numeric and update BioEssays

bwiernik:
https://forums.zotero.org/discussion/85312/style-request-updating-bioessays#latest


adam3smith:
@bwiernik do you want to add the styles on the exception list for unused macros?

bwiernik:
Actually, yeah, I'll do that instead.

adam3smith:
I think that should be `apa-numeric-superscript` in line with Vancouver, no?
(1) numeric
[2] numeric-brackets
<sup>1</sup> numeric-superscript
<sup>[1]</sup> numeric-superscript-brackets

bwiernik:
Got it okay.

adam3smith:
I hope the two journals are OK with this -- I've seen the provisional guidelines from the Biotechnology journals and they're a not-very-precise adaptation of APA 6th... let's hope they're actually OK with correct APA 7

adam3smith:
Thanks!

bwiernik:
BioEssays indicated that APA 7 was fine, and Biotechnology Journal said they were the same as BioEssays, so I’m thinking they aren’t too concerned about the details.

PR #5103, opened 2020-11-13
Update international-organization.csl

biejanpt:
Two issues with the style:

- Author name is repeated in bibliography, but should be replaced by "———" after first entry.
- Publications by same (group of) author(s) is not sorted by publication year.

### DIFF
--- a/international-organization.csl
+++ b/international-organization.csl
@@ -426,7 +426,7 @@
       </choose>
     </layout>
   </citation>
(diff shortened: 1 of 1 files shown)


adam3smith:
The sorting fix is definitely right, but I'm not seeing the author substitute in recent articles -- what are you basing that on?

biejanpt:
> The sorting fix is definitely right, but I'm not seeing the author substitute in recent articles -- what are you basing that on?

You're absolutely right! I seem to have mistakenly looked at older IO articles..

rmzelle:
@biejanpt, I've taken out the author substitute change, and accepted you sorting fix. Thanks!

PR #5104, opened 2020-11-13
Create harvard-harper-adams-university.csl

cpina:
Hi :-) long time no see! In the style I've added two documentation URLs: the University guide for authors and also a GitHub repository with some tooling, examples, etc. I'm happy to delete this one but I thought that it might be useful to have it here.

### DIFF
--- a/harper-adams-university-harvard.csl
+++ b/harper-adams-university-harvard.csl
@@ -188,7 +188,7 @@
             <text variable="publisher-place"/>
             <text variable="publisher"/>
           </group>
-       </group>
+        </group>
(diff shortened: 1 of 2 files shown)


cpina:
Thanks @adam3smith for the style renaming and indentation. Ironic for a Python programmer (me) to get the indentation wrong! (I thought that it would be picked up by `rake` locally...).

We think that the style is good to go, we just made some more changes before an assignment, thanks!

rmzelle:
@cpina, welcome back, and thanks for the style!

PR #5106, opened 2020-11-15
Fix file modes

cpina:
These CSL files were set as executable 755 (rwxr-xr-x).
Marks them as 644 (rw-r--r--) like the rest of files


cpina:
This didn't cause any problem more than looked strange when I was listing files in the terminal (as they appeared green in my setup)

rmzelle:
@cpina, I didn't know Git kept track of permission settings. Learning something new every day! https://stackoverflow.com/questions/40978921/how-to-add-chmod-permissions-to-file-in-git

Thanks!

Issue #4753, opened 2020-05-01
`bibtex.csl` breaks with space in author last name


rschuetzler:
The generation of the `citeKey` in `bibtex.csl` creates an invalid key if there are spaces in the authors last name. For example, the following, as you can see, creates the cite key of `van de Schoot_Yerkes_Mouw_Sonneveld_2013`, and spaces in the key break parsing.

```
@article{van de Schoot_Yerkes_Mouw_Sonneveld_2013, 
title={What Took Them So Long? Explaining PhD Delays among Doctoral Candidates}, 
volume={8}, 
ISSN={1932-6203}, 
url={https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3720867/}, 
DOI={10.1371/journal.pone.0068839}, 
number={7}, 
journal={PLoS ONE}, 
author={van de Schoot, Rens and Yerkes, Mara A. and Mouw, Jolien M. and Sonneveld, Hans}, 
year={2013}, 
month={Jul} 
}
```

adam3smith:
Nothing we can do about this -- CSL doesn't allow modifying strings. For this and other reasons we don't recommend relying on CSL to generate bibtex. Personally I'd prefer to just remove the style altogether, but since some people find it useful, we're leaving it on the repository. 
FWIW, we'll likely get a citekey variable in CSL in the future that'd solve this particular problem.

woutermont:
> Nothing we can do about this -- CSL doesn't allow modifying strings. 

I'd like to reopen this issue, since I believe that "CSL doesn't allow modifying strings" is not enough reason to conclude that there is "nothing we can do about this." There might be other possibilities of circumventing this small but annoying issue in a style that otherwise works quite fine.

> FWIW, we'll likely get a citekey variable in CSL in the future that'd solve this particular problem.

Might I propose that, at least pending a dedicated citekey variable, we use an already existing variable which might be suited to the purpose? One example that comes to mind is the `citation-label` variable: it is not yet used by the bibtex style, and could therefore be employed as a preferred way to set the bibtex citekey (only when it is available, for compatibility purposes).

Pending your consideration, I'll create a pull request.

PR #5117, opened 2020-11-21
citation-label as preferred bibtex citeKey (issue #4753)

woutermont:
Enables correct bibtex.csl output when author's last name includes spaces, by choosing the `citation-label` variable as citekey instead, if it's available. (edit for ref: issue #4753)

### DIFF
--- a/bibtex.csl
+++ b/bibtex.csl
@@ -47,10 +47,17 @@
     </choose>
   </macro>
   <macro name="citeKey">
-    <group delimiter="_">
-      <text macro="author-short" text-case="lowercase"/>
-      <text macro="issued-year"/>
-    </group>
+    <choose>
+      <if variable="citation-label">
+        <text variable="citation-label"/>
+      </if>
+      <else>
+        <group delimiter="_">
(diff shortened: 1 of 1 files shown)


adam3smith:
Sorry, not going to take this. citation-label is something different than a citekey and we don't want to conflate them in official styles. 
Also, 1.0.2, [has been released with a citation-key variable](https://github.com/citation-style-language/schema/releases/tag/v1.0.2), so we're not far away from being able to update styles with it on the repository -- no reason to implement a workaround now.

woutermont:
Oh, my bad, hadn't caught that latest update! 

For what it's worth, I do think using `citation-label` was a valid option, since its role ("label identifying the item in in-text citations of label styles") fits this use perfectly: once one get's over the fact that it is somewhat dubios to treat bibtex as a citation style, it is very natural to classify it as a label style, in which the cite key is used within the text (i.e. the tex document) to refer to bibliographic items, exactly like labels typically do.

Of course I'm perfectly fine with using `citation-key` as well, now that it is available.

PR #5120, opened 2020-11-23
Create IUFRO-HQ.csl

IUFRO:
Style designed to be used by authors publishing with the International Union of Forest Research Organizations, but open to be used by anyone looking for a simple and consistent style

### DIFF
--- a/IUFRO-HQ.csl
+++ b/IUFRO-HQ.csl
@@ -1,0 +1,128 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" et-al-min="3" initialize-with="." page-range-format="expanded" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+  <!-- This style was edited with the Visual CSL Editor (https://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>International Union of Forest Research Organizations - Headquarters</title>
+    <title-short>IUFRO - HQ</title-short>
+    <id>http://www.zotero.org/styles/international-union-of-forest-research-organizations-headquarters</id>
+    <link href="http://www.zotero.org/styles/international-union-of-forest-research-organizations-headquarters" rel="self"/>
+    <author>
+      <name>Nelson Grima</name>
+      <email>grima@iufro.org</email>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="biology"/>
+    <updated>2020-11-23T07:51:25+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <macro name="author-short">
+    <names variable="author">
+      <name form="short" and="text" et-al-min="3"/>
+    </names>
+  </macro>
+  <macro name="author-count">
+    <names variable="author">
+      <name form="count"/>
+    </names>
+  </macro>
+  <macro name="author">
+    <names variable="author">
+      <name and="text" initialize-with="." name-as-sort-order="all" sort-separator=" "/>
+    </names>
+  </macro>
+  <macro name="issued">
+    <date variable="issued">
+      <date-part name="year"/>
+    </date>
+  </macro>
+  <macro name="publisher">
+    <group delimiter=": ">
+      <text variable="publisher"/>
+      <text variable="publisher-place" suffix="."/>
+    </group>
+  </macro>
+  <macro name="editor">
+    <names variable="editor" suffix=".">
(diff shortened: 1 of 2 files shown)


rmzelle:
Thanks for your submission. Is there an online style guide available for this style? Per https://github.com/citation-style-language/styles/wiki/Style-Requirements#10---documentation-link, we typically require each CSL style to include a link to the format it describes.

IUFRO-HQ.csl
```diff
@@ -0,0 +1,128 @@
...
+      <key macro="author" names-min="3" names-use-first="1"/>
+      <key macro="issued"/>
+    </sort>
+    <layout suffix=".">
+      <group delimiter=" ">
+        <text macro="author"/>
+        <text macro="issued" prefix="(" suffix=")."/>
+        <choose>
+          <if type="article article-magazine article-newspaper article-journal review manuscript paper-conference" match="any">
+            <text variable="title" text-case="sentence" font-style="italic" suffix="."/>
```
adam3smith, line 78: We don't use `text-case="sentence"` for titles and it will be deprecated in the future (as it can't handle proper names). If you want titles in sentence case, just don't use a `text-case` attribute. You should assume titles are stored in sentence case in the metadata.

IUFRO-HQ.csl
```diff
@@ -0,0 +1,128 @@
...
+  <macro name="publisher">
+    <group delimiter=": ">
+      <text variable="publisher"/>
+      <text variable="publisher-place" suffix="."/>
+    </group>
+  </macro>
+  <macro name="editor">
+    <names variable="editor" suffix=".">
+      <label form="short" text-case="capitalize-first" prefix="(" suffix=") "/>
+      <name and="text" et-al-min="5" initialize-with="."/>
```
adam3smith, line 47: please always use `et-al-min` and `et-al-use-first` jointly to avoid confusion. Inheriting one from a higher level is super hard to follow in a style.

IUFRO-HQ.csl
```diff
@@ -0,0 +1,128 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" et-al-min="3" initialize-with="." page-range-format="expanded" demote-non-dropping-particle="sort-only" default-locale="en-GB">
```
adam3smith, line 2: I'd generally avoid setting `et-al-min` at the style level, especially since it looks like you're overriding it for every occurrence.

IUFRO-HQ.csl
```diff
@@ -0,0 +1,128 @@
...
+  <!-- This style was edited with the Visual CSL Editor (https://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>International Union of Forest Research Organizations - Headquarters</title>
+    <title-short>IUFRO - HQ</title-short>
+    <id>http://www.zotero.org/styles/international-union-of-forest-research-organizations-headquarters</id>
```
adam3smith, line 7: the filename should be the same as the id (plus the `.csl` extension of course)

adam3smith (changes requested):
Some comments on the style -- also see Rintze's question about the style guide

IUFRO-HQ.csl
IUFRO: Thank you, I was not aware of this issue. I removed the 'text-case' attribute in the new version.

IUFRO-HQ.csl
IUFRO: Following this comment, I added 'et-al-use-first' in all instances where I used 'et-al-min'

IUFRO-HQ.csl
IUFRO: I removed 'et-al-min' at the style level

IUFRO-HQ.csl
IUFRO: I changed the name of the file to match the id.

IUFRO:
Thank you both for your informative comments. Please find the new version with your comments addressed in the 'Commits' section.

rmzelle:
> Thank you both for your informative comments. Please find the new version with your comments addressed in the 'Commits' section.

Thanks, almost there.

IUFRO:
> 
> 
> > Thank you both for your informative comments. Please find the new version with your comments addressed in the 'Commits' section.
> 
> Thanks, almost there.
> 
> Did you base this on an existing CSL style? If so, we'd like to know which one, so we can add a "template" link (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link)

Thanks! No, I did not use an existing CSL style. I opened the Visual CSL Editor (https://editor.citationstyles.org/visualEditor/), removed the example that appeared by default, and started filling in the necessary information until I got the style in the format I wanted.

I hope that helps!

rmzelle:
Okay, thanks!

PR #5124, opened 2020-11-23

e42mercury:
Adds missing sort element for inline citations. Fixes non-page number locator labels (chapters, etc.) to be long form and capitalize-first.


adam3smith:
substantive changes are correct. @rmzelle -- good to merge from my end.

rmzelle:
@e42mercury, thanks!

Issue #4474, opened 2019-12-23
Replace "styles-distribution" repo with "1.0.1" branch


rmzelle:
I was just contemplating whether it wouldn't be much clearer to get rid of https://github.com/citation-style-language/styles-distribution, and just have Travis CI automatically push any successful commit build in "master" to a "1.0.1" branch in this repo.

Benefits:

- Two fewer repositories (https://github.com/citation-style-language/styles-distribution and https://github.com/citation-style-language/distribution-updater can be retired).
- The "1.0.1" branch would be a stable location from where downstream projects can source v1.0.1 CSL styles, making it more straightforward for us to migrate "master" to a new CSL version whenever we want. Basically each CSL release would immediately get its own dedicated branch ("1.0", "1.0.1", etc.) from which downstream projects can source their styles. It would receive updates while the release is current and go dormant whenever we release a new CSL version.
- Probably would make it easier to address some packaging requests like https://github.com/citation-style-language/styles/pull/2706 as well.

Main question I think is whether we can easily reproduce some of the logic currently embedded in https://github.com/citation-style-language/distribution-updater/blob/master/scripts/styles_distribution.py in a Travis CI-run script (modified README.md, updating `<updated/>` timestamps with UTC commit time, deleting non ".csl" files except for renamed-styles.json).

See e.g. https://stackoverflow.com/a/52960498 and https://stackoverflow.com/a/48339945 for implementation options.

koppor:
If I get your description right, isn't that script just following:

```BASIC
10 git checkout master
20 git checkout -b 1.0.1-distribution
30 remove obsolete files from directory
40 git commit && git push
50 ... updates to master coming in ...
60 git merge master
70 git push
80 goto 50
```

For pushing, there is a GitHub action: https://github.com/ad-m/github-push-action

For executing scripts on the command line, one can use:

```yaml
    steps:
      - run: |
        git checkout 1.0.1-distribution
        git merge master
        git commit
```

The checkout has to be complete though: (see https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches)

```yaml
    steps:
      - uses: actions/checkout@v2
      - run: |
          git fetch --prune --unshallow
```

koppor:
BTW: I suffixed `-distribution` as branch name. It is very strange that version numbers appearing like SemVer are "floating". Although I agree that this is some special thing here.

It is also not common to have branch names similar to tags. Therefore that prefix to "irritate" also on the name.

[...]

PR #5126, opened 2020-11-25
Delete non-release files

rmzelle:
Follow-up of https://github.com/citation-style-language/styles/pull/5111 and https://github.com/citation-style-language/styles/issues/4474.

@adam3smith, @bwiernik, I assume it makes sense to just delete the Gem/Rake/Sheldon files in "v1.0.1" but keep:
* CONTRIBUTING.md
* README.md
* renamed-styles.json

(plus LICENSE.md) when we add one

Let me know if you agree. We'll have to make sure that these files are pushed from "master" to "v1.0.1" whenever they're updated.


rmzelle:
Further cleanup in https://github.com/citation-style-language/styles/pull/5131

bwiernik:
What are they support files we will need on this branch for GitHub Actions? @retorquere

retorquere:
There are no requirements for the release branches from the pov of the actions; the actions expect to be ran on pull requests (`sheldon.yaml`) or `master` (`merge.yaml`) only, and for that, they need to exist only on `master` (assuming merges always happen from, and pull requests always target, `master`). The only thing to keep in mind is that changes to `.csl` files on `styles`, and `.xml` files on `locales` will be overwritten by sheldon when `merge` is ran.

`merge.yaml` can be dropped into `locales` as-is BTW, unless the webhooks should not be ran on that repo. That's easy enough to exclude should it be needed. I am not sure about the travis setup for locales -- I don't see any tests being ran for it, nor are the distribution-updaters triggered, but shelbot is pinged on success and fail.

We should probably turn off shelbot BTW.

rmzelle:
> `merge.yaml` can be dropped into `locales` as-is BTW, unless the webhooks should not be ran on that repo.

Zotero only updates locale files for each release, IIRC, so they currently don't use a webhook for that to get more frequent updates than that.

> I am not sure about the travis setup for locales -- I don't see any tests being ran for it, nor are the distribution-updaters triggered, but shelbot is pinged on success and fail.

It's a rather low-traffic repo. You can see some builds back at travis-ci.org (e.g. https://travis-ci.org/github/citation-style-language/locales/builds/629659918).

retorquere:
Another apropos -- I briefly considered "inlining" sheldon since we don't need shelbot anymore, but having it in a separate repo makes it easy to update for both the locales and the styles repo -- the actions are set up to always fetch the latest master from the sheldon repo, so fixes to sheldon can just be made there and new test (re)runs will pick it up.

retorquere:
> Zotero only updates locale files for each release, IIRC, so they currently don't use a webhook for that to get more frequent updates than that.

I can exclude it for locales so that a single copy can just work for both.

> It's a rather low-traffic repo. You can see some builds back at travis-ci.org (e.g. https://travis-ci.org/github/citation-style-language/locales/builds/629659918).

right -- I had forgotten that Travis runs `rake` automatically if it thinks you're testing a ruby project. A replacement action for that will be trivial if it is desired -- seems reasonable to not ditch tests we have.

retorquere:
Sheldon doesn't report back to the issue on the test results for locales right? There's just the test status as GH display it.

rmzelle:
> seems reasonable to not ditch tests we have.

Yeah, we'd like to keep the tests for "locales".

> Sheldon doesn't report back to the issue on the test results for locales right? There's just the test status as GH display it.

Correct. For "styles", there is value in having comments that show how the style renders, but we don't need that for "locales".

retorquere:
I've submitted a PR. `merge.yaml` is just a copy of the one in this repo, `sheldon.yaml` doesn't actually call sheldon, and doesn't run on pull requests (I didn't see any in that repo), just on push.

retorquere:
No wait, halt. That was dumb. Fix incoming.

bwiernik:
I don't see us updating 1.0.1 styles much once we move to 1.0.2, but what about 1.1? If we wanted to update a 1.0.2 style once `master` refers to 1.1, what would the process be?

rmzelle:
> I don't see us updating 1.0.1 styles much once we move to 1.0.2, but what about 1.1? If we wanted to update a 1.0.2 style once `master` refers to 1.1, what would the process be?

We could set up a second dev branch for v1.0.2 styles (which would need its own customized GitHub Action so it runs the right tests, uses the right CSL schema, and deploys to the right branch). E.g.:

Before:
"master" > "v1.0.2"

After:
"master" > "v1.1"
"master-v1.0.2" > "v1.0.2"

This way the "v1.x.x" branches are always stable (e.g. "v1.0.1" will always only have CSL 1.0.1 styles in it), will always only have validating styles that pass all tests (since we should only push to them via the GitHub Actions after a successful build in the master branches), and will be clean (with no files irrelevant to the end users).

You'd branch master-v1.0.2" off "master" just before you switch "master" to CSL 1.1.

bwiernik:
I like that.

rmzelle:
> "master-v1.0.2"

Thinking about this a little bit more, we could also name this branch "master-v1.0.x", although I don't think anybody is anticipating a v1.0.3 release where you'd benefit from a more general name.

PR #5130, opened 2020-11-26
Fix Harvard Praxisforschung Editors

adam3smith:
See 
https://forums.zotero.org/discussion/86362/herausgeber-aendern#latest

### DIFF
--- a/harvard-institut-fur-praxisforschung-de.csl
+++ b/harvard-institut-fur-praxisforschung-de.csl
@@ -46,7 +46,8 @@
       <name name-as-sort-order="all" sort-separator=", " delimiter="; " delimiter-precedes-last="always" form="long"/>
       <label form="short" prefix=" (" suffix=")"/>
       <substitute>
-        <text macro="editor"/>
+        <names variable="editor"/>
+        <names variable="translator"/>
         <text macro="anon"/>
       </substitute>
     </names>


adam3smith:
@retorquere -- first "real" test: that worked very nicely indeed!

retorquere:
Superb. We do still have to disable Travis.

rmzelle:
> We do still have to disable Travis.

I just disabled Travis at https://travis-ci.com/github/citation-style-language/styles/settings:

image

@retorquere, I think we now just need to:

* Add "CONTRIBUTING.md", "README.md", and "renamed-styles.json" to https://github.com/citation-style-language/styles/blob/dcf473a3707998b3a4d030eaae9fe631be6d3cb1/.github/workflows/merge.yaml#L40 so updates to these files in "master" are pushed to "v1.0.1"
* Rename "1.0" to "v1.0" for consistency
* Set up the same actions for the "locales" repo. @retorquere, do you mind setting a PR there as well?

rmzelle:
Oh, and @retorquere, do the current GitHub Actions still ping:

 *   https://shelbot.herokuapp.com/build
 *   https://styles-update.zotero.org:8826/
 *   https://styles-update.zotero.org:8827/

? It looks like Zotero picked up the change in the PR here (https://www.zotero.org/styles?q=id%3Aharvard-institut-fur-praxisforschung-de), although I'm not sure if that was because of left-over Travis CI activity. https://github.com/citation-style-language/styles-distribution/ no longer updates (not a big issue, as we can tell people to use the "v1.0.1" branch instead).

(see also https://styles-update.zotero.org:8826/ (https://github.com/citation-style-language/Sheldon/issues/32#issuecomment-727693180)

retorquere:
They currently do not -- that is from the Travis runs. What doe the pingees expect to receive?

rmzelle:
> They currently do not -- that is from the Travis runs. What doe the pingees expect to receive?

See https://github.com/citation-style-language/Sheldon/issues/32#issuecomment-727693180. It would be great if you could get https://styles-update.zotero.org:8826/ to work. We should be able to deprecate the other two.

rmzelle:
>     * Rename "1.0" to "v1.0" for consistency

Just did this (https://github.com/citation-style-language/styles/tree/v1.0) via https://stackoverflow.com/a/26440865/1712389.

retorquere:
I've updated the workflows, both repos now have a PR. The workflows are identical between the repos, so in the future, one can be edited if needed and then dropped into the other.

PR #5131, opened 2020-11-26
Further cleanup of v1.0.1

rmzelle:


bwiernik:
.gitignore probably needs to be kept or it could lead to messiness when checking out different branches

rmzelle:
> .gitignore probably needs to be kept or it could lead to messiness when checking out different branches

Does it? Wouldn't switching branches also switch .gitignore?

I'm also not sure the original contents of .gitignore is even relevant anymore.

PR #5132, opened 2020-11-26
ping distribution-updater

retorquere:
Assuming the distribution updater is going to bail on notifications of failed runs, this should do it (after setting the appropriate secret). If the ping for a failed build is still desired, that's just two lines and an `if` test extra.

### DIFF
--- a/.github/workflows/merge.yaml
+++ b/.github/workflows/merge.yaml
@@ -15,6 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+      DISTRIBUTION_UPDATER_TOKEN: "${{ secrets.DISTRIBUTION_UPDATER_TOKEN }}"
     steps:
     - uses: actions/checkout@v2
       if: github.event_name == 'push'
@@ -89,3 +90,8 @@ jobs:
         repository: 'release'
         commit_message: ${{ github.event.inputs.commit_message }}
       if: github.event_name == 'workflow_dispatch'
+
+    - name: ping distribution-updater
+      run: |
+        curl -H "Authorization: $DISTRIBUTION_UPDATER_TOKEN" -F 'payload={"type":"pull_request","branch":"master","status":0,"commit":"'$GITHUB_SHA'"}' https://styles-update.zotero.org:8826/
+        curl -H "Authorization: $DISTRIBUTION_UPDATER_TOKEN" -F 'payload={"type":"pull_request","branch":"master","status":0,"commit":"'$GITHUB_SHA'"}' https://styles-update.zotero.org:8827/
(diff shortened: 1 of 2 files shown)


retorquere:
How are we still getting these welcome messages? I can see [here](https://github.com/citation-style-language/styles/pull/5132/checks?check_run_id=1461029513) that all steps are skipped because no csl or xml files are changed.

rmzelle:
> How are we still getting these welcome messages? I can see [here](https://github.com/citation-style-language/styles/pull/5132/checks?check_run_id=1461029513) that all steps are skipped because no csl or xml files are changed.

Could Sheldon be monitoring the GitHub API as well (maybe via https://github.com/citation-style-language/Sheldon/blob/05f8ca911d8a08d646af4b69b2e22543628261f1/spec/sheldon/bot_spec.rb#L6)?

retorquere:
The only ways that I know a bit can monitor GH repos is by webhooks, or by by installing a GH app.

rmzelle:
@retorquere, do we still rely on the Heroku app, or can we solve this issue of @csl-bot welcome messages by disabling the app?

rmzelle:
(alternatively we can just pull @csl-bot's access)

retorquere:
We don't rely on the Heroku app, it can be decommissioned.

rmzelle:
> We don't rely on the Heroku app, it can be decommissioned.

Cool. I just disabled the Heroku app by turning off the dyno per https://stackoverflow.com/a/53198598/1712389.

rmzelle:
@retorquere, looks like this PR doesn't work quite right: https://github.com/citation-style-language/styles/runs/1463923292

rmzelle:
I also noticed that no builds are run for new commits in some old PRs. See e.g. my commit in https://github.com/citation-style-language/styles/pull/5124 from just now. It should at least run the build for merging the PR into "master", I think. Not sure why that doesn't work.

retorquere:
> @retorquere, looks like this PR doesn't work quite right: https://github.com/citation-style-language/styles/runs/1463923292

A PR has been submitted to fix that.

retorquere:
> I also noticed that no builds are run for new commits in some old PRs. See e.g. my commit in #5124 from just now. It should at least run the build for merging the PR into "master", I think. Not sure why that doesn't work. (maybe I'm wrong here; we should be able to deal with any outstanding pull requests without too much trouble)

I think that those commits run on a ref that doesn't have the workflow yet. Bummer, hadn't considered that. AFAICT it'd require merging master into the PR. There are not too many open PRs; I *think* I can do that? AT least locally that works. Where do PRs technically live in github? In the target repo (where we could merge in master) or in the fork (out of our reach).

adam3smith:
> Where do PRs technically live in github? In the target repo (where we could merge in master) or in the fork (out of our reach).

They live in the fork but by default the target repo editors get commit access to the PR branch, so it'd still be possible to merge in master. We're talking 6 relevant PRs, so manually running rake locally would also be very much feasible.

PR #5135, opened 2020-11-27
Create opto-electronic-advances.csl

lancefxh:
This is a CSL citation style for "Opto-Electronic Advances" which can't  be found in the Zotero Style Repository before. The style was created by myself.

### DIFF
--- a/opto-electronic-advances.csl
+++ b/opto-electronic-advances.csl
@@ -6,8 +6,8 @@
     <title-short>Opto-Electron Adv</title-short>
     <id>http://www.zotero.org/styles/opto-electronic-advances</id>
     <link href="http://www.zotero.org/styles/opto-electronic-advances" rel="self"/>
-    <link href="http://www.oejournal.org/J/OEA/Channel/15" rel="documentation"/>
     <link href="http://www.zotero.org/styles/nature" rel="template"/>
+    <link href="http://www.oejournal.org/J/OEA/Channel/15" rel="documentation"/>
     <author>
       <email>xuhaofan@hust.edu.cn</email>
       <name>Xuhao Fan</name>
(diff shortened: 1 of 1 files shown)


opto-electronic-advances.csl
```diff
@@ -1,10 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" default-locale="en-GB">
   <!-- This style was edited with the Visual CSL Editor (https://editor.citationstyles.org/visualEditor/) -->
   <info>
     <title>Opto-Electronic Advances</title>
     <title-short>Opto-Electron Adv</title-short>
-    <id>http://www.zotero.org/styles/opto-electronic-advances</id>
```
adam3smith: no, you need the id

opto-electronic-advances.csl
```diff
@@ -0,0 +1,150 @@
...
+        <text variable="title" font-style="italic"/>
+      </if>
+      <else>
+        <text variable="title"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="author">
+    <names variable="author">
+      <name and="symbol" delimiter-precedes-last="never" et-al-min="5" et-al-use-first="5" initialize-with=". " name-as-sort-order="all"/>
```
adam3smith, line 39: `et-al-min="5" et-al-use-first="5"` is the problem -- `et-al-use-first` needs to be at least one lower then `et-al-min`

adam3smith: (and you may want to delete it entirely -- the 6/1 set in the bibliography is more plausible and there's no reason to have it in two places)

lancefxh: okay, I will try again

lancefxh: A little difference between paper-conference and others in the suffix ";"and "."

rmzelle:

rmzelle:
Thanks!

PR #5137, opened 2020-11-28
Ping on push, not PR, document role of dist-updater

retorquere:


.github/workflows/merge.yaml
```diff
@@ -91,8 +91,12 @@ jobs:
         commit_message: ${{ github.event.inputs.commit_message }}
       if: github.event_name == 'workflow_dispatch'
 
+    # this is pinging Zotero more generally (styles page, Zotero
+    # client updates, etc.), not just the distribution-updater. Zotero
+    # is not the only client that needs to be notified when things
+    # like this update, the entire Zotero ecosystem relies on this.
     - name: ping distribution-updater
```
rmzelle, line 97: @retorquere, can we freely rename "ping distribution-updater" here? It's probably better to use something like "ping Zotero servers", as only https://styles-update.zotero.org:8827/ is specific to https://github.com/citation-style-language/distribution-updater, and we're looking to deprecate this (while keeping the https://styles-update.zotero.org:8826/ ping).

"$DISTRIBUTION_UPDATER_TOKEN" should probably also be "$ZOTERO_UPDATE_TOKEN" or something like that.

retorquere: > @retorquere, can we freely rename "ping distribution-updater" here?

Yes, no problem; `name` attributes are only friendly names shown while the workflow is running, they can be anything, and they're always optional

> "$DISTRIBUTION_UPDATER_TOKEN" should probably also be "$ZOTERO_UPDATE_TOKEN" or something like that.

That's also no problem.

rmzelle: @retorquere, by the way, if it's not too much to ask, it would be great if we could completely migrate from Travis CI to GitHub Actions. Apart from "styles" and "locales", we also used Travis CI for "journals" and "Sheldon":

https://travis-ci.com/github/citation-style-language/Sheldon
https://travis-ci.com/github/citation-style-language/journals

Looks like we might not need to migrate Sheldon, as it looks like the only thing Travis CI does there is deploying the new version of Sheldon to Heroku, which is no longer relevant (https://github.com/citation-style-language/Sheldon/blob/master/.travis.yml). I guess we can just delete ".travis.yml" (not sure if there are other files we can delete).

For the "journals" repo, we have some very limited checks to make sure the file "publishers.json" is proper JSON and describes the various directories in that repo (see https://github.com/citation-style-language/journals/blob/master/spec/repository_spec.rb).

rmzelle: (and I realize I keep adding requests, so don't hesitate to ask if you'd like a higher honorarium; not sure if you already agreed on a number with @adam3smith)

retorquere: > @retorquere, by the way, if it's not too much to ask, it would be great if we could completely migrate from Travis CI to GitHub Actions. Apart from "styles" and "locales", we also used Travis CI for "journals" and "Sheldon":
> 
> https://travis-ci.com/github/citation-style-language/Sheldon

Sheldon is just a Ruby package that has some convenience functions for the CI environment (now GHA). Travis used Sheldon rather than the other way around. Sheldon is pretty bare-bones compared to what it was because the CI setup is much less complex now, but we still do need Sheldon in that repo. The other CI workflows install Sheldon as part of their run.

But I had forgotten that Sheldon still had a `.travis.yml` file. I've removed that.

> https://travis-ci.com/github/citation-style-language/journals

This looks like a similar setup to what we have for the locales repo. Does it have a similar PR/release workflow, just centered on `.tab` files? That would be easy enough to accomodate.

> Looks like we might not need to migrate Sheldon, as it looks like the only thing Travis CI does there is deploying the new version of Sheldon to Heroku, which is no longer relevant (https://github.com/citation-style-language/Sheldon/blob/master/.travis.yml). I guess we can just delete ".travis.yml" (not sure if there are other files we can delete).

I've just deleted that, but deployment of shelbot to heroku was through heroku's auto-release pipeline. Whatever was committed on `master` would be deployed by a webhook heroku installed.

> For the "journals" repo, we have some very limited checks to make sure the file "publishers.json" is proper JSON and describes the various directories in that repo (see https://github.com/citation-style-language/journals/blob/master/spec/repository_spec.rb).

Which is ran with just calling `rake` right? That would make the pipeline the same as the locales repo. But for this repo the changes of interest are `publishers.json` and `*.tab`?

> (and I realize I keep adding requests, so don't hesitate to ask if you'd like a higher honorarium; not sure if you already agreed on a number with @adam3smith)

The only thing we had so far agreed on was that at least part of the work will be a donation to Zotero 😄 and that I would keep a rough estimate of time spent.

retorquere: The PR has been updated for support for the `journals` repo.

rmzelle: > This looks like a similar setup to what we have for the locales repo. Does it have a similar PR/release workflow, just centered on `.tab` files? That would be easy enough to accomodate.

No, "journals" is for internal use only, so no need for a release branch. We use that repo to store metadata for journals, and have a script to turn the metadata into dependent CSL styles.

> Which is ran with just calling `rake` right? That would make the pipeline the same as the locales repo. But for this repo the changes of interest are `publishers.json` and `*.tab`?

Right now the checks only cover publishers.json.

retorquere: Would this repo typically get PRs that you would want to have tested? Or just run `rake` on any PR/push?

rmzelle: > Would this repo typically get PRs that you would want to have tested? Or just run `rake` on any PR/push?

We'd want to test the PRs as well.

retorquere: PR submitted on the journals repo.

rmzelle: > > "$DISTRIBUTION_UPDATER_TOKEN" should probably also be "$ZOTERO_UPDATE_TOKEN" or something like that.
> 
> That's also no problem.

Can you confirm that https://github.com/citation-style-language/styles/pull/5137/commits/91dcad87f11843dfe6b7ea84585e2a03acb5c822 looks okay? If so it looks like I can merge this PR.

.github/workflows/merge.yaml
```diff
@@ -24,45 +24,48 @@ jobs:
...
       if: github.event_name == 'push'
       uses: dorny/paths-filter@v2
-      id: release
+      id: update
       with:
         list-files: shell
         filters: |
           updated:
-            - added|modified: [ '*.csl', '*.xml' ]
+            - added|modified: [ '*.csl', '*.xml', ]
```
rmzelle, line 44: @retorquere, looks like a superfluous comma here.

retorquere: fixed

rmzelle:
Thanks!

PR #5139, opened 2020-11-29
Update china-national-standard-gb-t-7714-2015-numeric.csl

buffalo-d:
Add dual language support for "et-al-min term"  (English: et al and Chinese: 等)in bibliography.


bwiernik:
Thanks! Your style doesn't validate correctly. Can you please run it though the CSL Validator (https://validator.citationstyles.org/) and fix any errors it reports?

adam3smith:
This is a more fundamental problem: multiple layouts with different locales aren't valid CSL. They're used in csl-m and happen to work in Zotero and other apps using citeproc-js (provided they allow the language to be specified in the metadata), but will fail in unknown ways for other processors. Since that's the only substantive change here, closing this.

PR #5143, opened 2020-12-01
Create ypfs-kpd.csl

runkelcorey:
Second iteration of YPFS Key Program Document variant (PR for first iteration closed). Validated by CSL Validator. This style allows contributors to the [Journal of Financial Crises](https://elischolar.library.yale.edu/journal-of-financial-crises/) to quickly format references for publication. It takes stock Chicago 17 in-text citations but creates new bibliographic entries that:
* emphasize the title
* link in-text citations with reference entries
* provide day and month when available for certain item types (the subjects of our journal often operate at timescales of days, weeks, or months)
* allow contributor to write short description
* provide stable URL
This style leverages two fields---`title-short` and `call-number`---not used in Chicago nor any other style included with Zotero to allow authors to use YPFS KPD simultaneously with Chicago (or to switch quickly between them), an oft-needed quirk of the *Journal*'s style.

### DIFF
--- a/ypfs-kpd.csl
+++ b/ypfs-kpd.csl
@@ -1,0 +1,657 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" page-range-format="chicago" demote-non-dropping-particle="display-and-sort">
+  <info>
+    <title>Yale Program on Financial Stability (KPD)</title>
+    <title-short>YPFS (KPD)</title-short>
+    <id>http://www.zotero.org/styles/ypfs-kpd</id>
+    <link href="http://www.zotero.org/styles/ypfs-kpd" rel="self"/>
+    <link href="https://www.zotero.org/styles/chicago-author-date" rel="template"/>
+    <link href="http://www.chicagomanualofstyle.org/tools_citationguide.html" rel="documentation"/>
+    <author>
+      <name>Sean Fulmer</name>
+      <email>sean.fulmer@yale.edu</email>
+    </author>
+    <author>
+      <name>Corey Runkel</name>
+      <email>corey.runkel@yale.edu</email>
+      <uri>https://orcid.org/0000-0001-8992-4434</uri>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="history"/>
+    <summary>The Key Program Document variant of YPFS style</summary>
+    <updated>2020-12-01T16:17:00-05:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="en">
+    <terms>
+      <term name="editor" form="verb-short">ed.</term>
+      <term name="container-author" form="verb">by</term>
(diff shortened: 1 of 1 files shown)


adam3smith:
Don't worry about the CI tests for now.

Could you clarify who exactly would be using this? I'm not quite clear I understand the relationship between the journal and the KPD database.

I don't understand this part -- that's not something you can do in a citation style:
> link in-text citations with reference entries

> This style leverages two fields---title-short and call-number
we're not particularly keen on this. CSL styles shouldn't be targeted just as Zotero. Moreover, `title-short` is (in the form of `title form="short"`) used by many styles, including several included by default in Zotero and other reference managers). call-number isn't used in styles often, but is frequently imported automatically, so that its inclusion in a citation style would create odd output. What's the specific target output here? Why wouldn't URL work?

runkelcorey:
Most cases are written by YPFS Research Associates; there are 12 of us now, though some cases could conceivably come from outside the Program. In some cases we use two separate, partially overlapping bibliography sections (see IV and V [here](https://elischolar.library.yale.edu/cgi/viewcontent.cgi?article=1104&context=journal-of-financial-crises)): one almost exactly Chicago (author-date); the other is the style proposed here.

Since we also maintain an archival project, we would like to track original URLs in addition to those provided by [our archive](https://ypfs.som.yale.edu/). This is as much for our own archival purposes (not targeted explicitly at Zotero, but I am assuming anyone using a CSL style is also using a citation manager) as for our RAs to cite original URLs when writing for outside publications that use Chicago or Harvard while citing ypfs.som.yale.edu URLs in our journal. Some styles often used for business history use `title-short` sparingly---in those cases, we recommend authors make ad-hoc adjustments---but at least 90% of the authors in our journal are us RAs, so we're able to keep the `title-short` field clear.

The use of `call-number` is a little more subtle. In the linked case above, you will notice that Key Program Documents (V) are segregated by type of document---Implementation Documents, or Reports/Assessments (though most cases have two or three more)---and that the number of documents can run well into the double digits. We instruct our RAs to replace `call-number` with `implementation` or `report` so that, when output, the bibliography is already sorted first into these categories, and then by author, date, and title. `call-number` is not used in Chicago, MLA, APA, APSA, Harvard, or Bluebook, which are about the only styles in which I can imagine our contributors also writing.

> link in-text citations

is simply an ambiguous use of the word 'link'. In cases without Chicago-style References sections (ie. with only KPD sections), we need to include the Chicago-style in-text citation so that readers can match citation with reference.

adam3smith:
This sounds like a super cool application of Zotero/CSL, but I'm somewhat skeptical that it needs to be on the style repository, especially now that Zotero recognizes CSL styles from github Raw links, so you can simply host it anywhere on github and  it'll install automatically from the link. 
Given that this does sound like it's almost entirely for internal use, is there a particular need to have it hosted here (and thus made available to 100,000s of users)?

runkelcorey:
There is a possibility that it helps other users. But mostly we did the work of creating a style and thought we might as well do the few extra steps to submit it.

adam3smith:
Thanks -- I think this makes more sense for you to just keep hosting yourself. Given the number of styles we have, we're keeping [relatively stringent criteria](https://github.com/citation-style-language/styles/wiki/Criteria-for-Accepting-Styles) for what makes sense to host here. 
Again, though, I think it's a really great application and I'm glad we're aware of it.

PR #5144, opened 2020-12-01
Add QHR (APA dependent)

adam3smith:


adam3smith:
@retorquere if I'm reading the test correctly, it completely ignores dependent styles & their changes&updates (presumably because it doesn't look in the subdirectory?)-- could they be included?

retorquere:
That's correct -- added.

PR #5149, opened 2020-12-02
Update canadian-biosystems-engineering.csl

mspeare:
Updated the style to include all authors in the reference list.


bwiernik:
Thanks! Per the style documentation, it looks like they use et al in the in text citations, so can you put that back into the author-short macro? https://www.csbe-scgab.ca/images/journal/prepareweb.pdf

mspeare:
I added back in the et al in the author short macro but I don't know if I submitted the changes correctly.

rmzelle:
Thanks!

PR #5154, opened 2020-12-03
Create geriatrics-gerontology-international.csl

POBrien333:
via https://forums.zotero.org/discussion/85492/style-request-geriatrics-gerontology-international/p1


rmzelle:
Thanks! (I set the default-locale to "en-US")

PR #5155, opened 2020-12-03
Create human-ecology.csl

POBrien333:
via https://forums.zotero.org/discussion/85672/style-request-journal-of-human-ecology/p1


rmzelle:
Thanks!

PR #5159, opened 2020-12-05
Update guide-des-citations-references-et-abreviations-juridiques.csl

rafjaf:
Added support for legislative bills


rmzelle:
Thanks!

PR #5160, opened 2020-12-06
Amsterdam university press: remove a stray space

slippycheeze:
This removes a stray space in one of the `<link>` elements that breaks validation.


adam3smith:
Thanks. I'm curious -- where did this break validation? It doesn't look like our validator complained.

PR #5163, opened 2020-12-09
Update rheumatology.csl to vancouver-brackets-only-year-no-issue parent

POBrien333:
via https://forums.zotero.org/discussion/86583/format-for-rheumatology#latest


adam3smith:
Thanks!

PR #5164, opened 2020-12-09
fix issue with artwork in modern-humanities-research-association.csl

POBrien333:
via https://forums.zotero.org/discussion/86599/mhra3-citation-for-artwork-has-a-blank-field

User confirmed that fix works.


rmzelle:
Thanks!

PR #5165, opened 2020-12-11
Push README.md to release branch

rmzelle:


rmzelle:
@retorquere, could you take a look at the run failure at https://github.com/citation-style-language/styles/actions/runs/414681544 ?

retorquere:
Looks like the token doesn't have the requisite permissions: https://github.com/citation-style-language/styles/runs/1535289254?check_suite_focus=true#step:18:28:

> remote: Permission to citation-style-language/locales.git denied to csl-bot.

adam3smith:
Is the problem the token or csl-bot? I suspect the latter -- I don't see csl-bot as having access to the locales repo? Did this work on styles? Because the problem should be the same.

retorquere:
It's csl-bot, and if csl-bot doesn't have push access to `locales`, that explains it. It doesn't affect `styles` in the same way -- the action is being ran as `github-action` which for the duration of the workflow run has all access to the repo the workflow runs on. But the workflow in `styles` pushes a change to `locales` when the workflows update in `styles`, and `github-action` does not have cross-repo permissions, so we use csl-bot's access specifically for that part.

But no workflows were updated in that commit. That bit should not have ran at all. I'll check that out.

adam3smith:
Any concerns about giving csl-bot write access to locales? Happy to do that

retorquere:
I can't think of any. Qua attack surface it wouldn't be better or worse than having any other committer's account being compromised.

adam3smith:
Done -- the Style distributor team, which includes csl-bot now has write access to locales.

Currently @dstillman is part of that team -- any reason you need to be? Currently the team has access to 
https://github.com/citation-style-language/styles-distribution and https://github.com/citation-style-language/distribution-updater
both of which we're retiring if I understand correctly.

Also, just FYI, I've removed you as an admin from /styles

rmzelle:
> Currently the team has access to
> https://github.com/citation-style-language/styles-distribution and https://github.com/citation-style-language/distribution-updater
> both of which we're retiring if I understand correctly.

Yes, I'd like to retire these in favor of (protected) version branches in the main "styles" and "locales" repos.

And I see that https://github.com/citation-style-language/styles-distribution hasn't updated since 11/23. Since I just created https://github.com/citation-style-language/locales/tree/v1.0.1 to go along with https://github.com/citation-style-language/styles/tree/v1.0.1, we might want to retire "styles-distribution" sooner rather than later (we can just [archive](https://docs.github.com/en/free-pro-team@latest/github/creating-cloning-and-archiving-repositories/archiving-repositories) it and https://github.com/citation-style-language/distribution-updater), announce the shift on Twitter, and update our website (https://citationstyles.org/developers/#/code-repositories).

retorquere:
> And I see that https://github.com/citation-style-language/styles-distribution hasn't updated since 11/23.

That's partially because Travis was broken for a while, and partially because I stripped Sheldon in the move to GH actions later. The stripped sheldon means the tests on Travis must have started failing structurally, and thus the updater wasn't pinged for a while.

rmzelle:
> The stripped sheldon means the tests on Travis must have started failing structurally, and thus the updater wasn't pinged for a while.

Ah, that part still depends on Travis CI? I think I might have shut off Travis CI for the "styles" repo when we were migrating to GitHub Actions:

image
https://travis-ci.com/github/citation-style-language/styles/settings

Anyway, I don't think we need a grace period for https://github.com/citation-style-language/styles-distribution (worst case scenario is that downstream users don't get style updates). Better to just document and announce people should switch to https://github.com/citation-style-language/styles/tree/v1.0.1.

dstillman:
> Currently @dstillman is part of that team -- any reason you need to be?

Nope.

I can remove the distribution-updater from Zotero servers now, correct?

To recap:

  - styles-update.zotero.org:8827 will be removed
  - styles-update.zotero.org:8826 will remain and should still be pinged on commits

retorquere:
@rmzelle it doesn't rely on it anymore, but it took a while before I got to it in the new situation, and in the mean while, Travis couldn't ping, and then we decided to not update that repo for release anymore.

retorquere:
For the grace period, it wouldn't be hard to submit to both the branch and the other repo if so desired.

rmzelle:
@dstillman:
> I can remove the distribution-updater from Zotero servers now, correct?

@retorquere:
> For the grace period, it wouldn't be hard to submit to both the branch and the other repo if so desired.

@retorquere, if you don't mind implementing it, yes, it would be cleaner if we would first announce the deprecation of styles-distribution followed by a grace period of e.g. 2 months so any downstream users can migrate while still receiving style updates.

@dstillman, yes, (eventually) we'd only drop the ping to styles-update.zotero.org:8827.

retorquere:
Working on it (actually I think it's ready, but I want to give it a once-over), but the Netherlands has gone into a sudden full lockdown and I need to rearrange everything for my work, so I have my hands a but full right now.

PR #5168, opened 2020-12-11
Create journal-de-la-societe-des-oceanistes.csl

POBrien333:
via https://forums.zotero.org/discussion/85453/style-request-jso-journal-de-la-societe-des-oceanistes

### DIFF
--- a/journal-de-la-societe-des-oceanistes.csl
+++ b/journal-de-la-societe-des-oceanistes.csl
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
-  <!-- Polyglot; journal publishes in English and French -->
   <info>
     <title>Journal de la Société des Océanistes</title>
     <title-short>JSO</title-short>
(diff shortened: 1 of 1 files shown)


journal-de-la-societe-des-oceanistes.csl
```diff
@@ -0,0 +1,352 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+  <!-- Polyglot; journal publishes in English and French. -->
```
rmzelle, line 3: @POBrien333, can you confirm the default-locale should be British English, and that this is not either a French or polyglot journal?

adam3smith:
> can you confirm the default-locale should be British English, and that this is not either a French or polyglot journal?

the journal is polyglot English/French, but they do indeed use all English citations, even for French works in French articles, so the default locale is correct.

journal-de-la-societe-des-oceanistes.csl
POBrien333: as usual I forgot to take the locale out. thanks for spotting.

POBrien333:
thanks for checking on this, Sebastian.

rmzelle:
> the journal is polyglot English/French, but they do indeed use all English citations, even for French works in French articles, so the default locale is correct.

Great, thanks both. (I started adding the "Polyglot" XML comment as a reminder that a style should not carry a default-locale value, so in this case the comment should indeed be removed)

rmzelle:
Thanks!

POBrien333:
Looks like there is an issue with the styles being pushed over to zotero.org/styles.
Style is not appearing:
https://www.zotero.org/styles?q=journal-de-la-societe-des-oceanistes

adam3smith:
It's there now. We're still fine-tuning the updated bot, so that'll occasionally happen

POBrien333:
Thank you. :)

PR #5169, opened 2020-12-12
fix sorting in universitat-bern-institut-fur-musikwissenschaft-note.csl

denismaier:
For some reason sorting didn't work with the previous sort key. Should work now.


rmzelle:
> For some reason sorting didn't work with the previous sort key. Should work now.

With Zotero? Do you think this is a bug?

Note that the two versions sort differently, even though `macro="author_title-bibliography"` calls the other two macros. With `macro="author_title-bibliography"` as sort key, there is only a single sort key (a concatenation of the output of macro="primary-contributor" and macro="title_volumes"). With the new code, there are two separate sort keys that are evaluated sequentially. (not sure this difference can lead to differences in practice)

denismaier:
> With Zotero? Do you think this is a bug?

Yes. In Zotero. Don't know if that's a bug. I haven't tested with other citeprocs.
> 
> Note that the two versions sort differently, even though `macro="author_title-bibliography"` calls the other two macros. With `macro="author_title-bibliography"` as sort key, there is only a single sort key (a concatenation of the output of macro="primary-contributor" and macro="title_volumes"). With the new code, there are two separate sort keys that are evaluated sequentially. (not sure this difference can lead to differences in practice)

I actually thought the differences wouldn't be too severe. But the earlier version didn't sort the bibliography at all. the first cited item appeared first in the bibliography. So that could speak for a bug, wouldn't it? 

At least, the current version is more aligned with what seems to be common practice among most styles, right?

rmzelle:
> At least, the current version is more aligned with what seems to be common practice among most styles, right?

Yes, it's generally better to cleanly delineate the various sort keys if they should be evaluated in order.

Can you generate a minimum reproducible example that shows the different sorting results with the two style variants? You can export the relevant Zotero items to "Zotero RDF", zip it (so you have a GitHub-accepted extension: https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/file-attachments-on-issues-and-pull-requests), and attach it here. Please also provide the Zotero version, and paste the rendered output here.

fbennett:
I've traced the sort failure to a ``citeproc-js`` bug. It didn't like having the keys set to a macro that attempts to render an empty date variable (``original-date``). That was an artifact of a time when such macros discarded everything but the date, an off-spec choice that was rightly reversed in light of complaints from the field. [1]

I've made some additional changes that should improve the accuracy of sorts done with macros that call dates in addition to other variables, so the original macro may work when the revised processor eventually trickles through to a Zotero release. Nonetheless, it's probably a good policy to cast separate keys in ``cs:sort`` as @rmzelle suggests, even if only for the sake of clarity.

[1] https://github.com/Juris-M/citeproc-js/commit/13bd7d13b8230f86ff72e47c93b888c286ae8fd8

rmzelle:
> Nonetheless, it's probably a good policy to cast separate keys in `cs:sort` as @rmzelle suggests, even if only for the sake of clarity.

Perfect, thanks for finding the bug. I'll merge this then.

PR #5174, opened 2020-12-17
Update citation disambiguation for mis-quarterly

geritwagner:


adam3smith:
What is this based on? (e.g., comments from editor, style guide, published examples)?
Also, if you remove `disambiguate-add-givenname` you should also remove the `givenname-disambiguation-rule` attribute (which becomes meaningless)

geritwagner:
Thank you for catching the `givenname-disambiguation-rule`. I removed it in the last commit.
The changes are based on published examples (e.g., [1](https://misq.org/it-enabled-self-monitoring-for-chronic-disease-self-management-an-interdisciplinary-review.html), [2](https://misq.org/optimizing-and-satisficing-the-interplay-between-platform-architecture-and-producers-design-strategies-for-platform-performance.html), [3](https://misq.org/can-outsouring-of-information-technology-foster-innovations-in-client-organizations-an-empirical-analysis.html)) since the [manuscript guidelines](https://misq.org/manuscript-guidelines) do not cover citation disambiguation.

adam3smith:
So in https://misq.org/optimizing-and-satisficing-the-interplay-between-platform-architecture-and-producers-design-strategies-for-platform-performance.html 
I'm actually seeing add-name for Yoo, Lyytinen et al. 2010 vs. Yoo, Henfridsson, and Lyytinen 2010 (rather than Yoo et al 2010a/b)

I'm seeing et al. with a/b for a couple of papers in https://misq.org/it-enabled-self-monitoring-for-chronic-disease-self-management-an-interdisciplinary-review.html (Jospe et al. and Aharonovich et al.) but there is no a/b in the bibliography and that suggests to me the referencing wasn't checked very carefully in the first palce

I couldn't quickly find a relevant examples in https://misq.org/can-outsouring-of-information-technology-foster-innovations-in-client-organizations-an-empirical-analysis.html

nor could I find any examples for add-givenname (going either direction)

geritwagner:
There are obviously a few inconsistencies in the citation style of papers published in this journal. I have searched for examples of references to first authors who differ only in given names (the other authors and the year being equal) to verify whether `add-givenname` is used or not. I could not find any.
Not sure whether there is a corresponding regex for checking more papers or whether we will get a response from the journal. How should we proceed with this?

adam3smith:
If you don't mind emailing the journal, that's certainly worth a try -- they're often helpful (it is, after all, in their interest to get manuscripts the way they want them)

geritwagner:
Of course, I will reach out to them and check what the preferred rule is.

geritwagner:
I revised the csl file according to the specifications provided by the managing editor of the journal. However, the current version applies `disambiguate-add-givenname` before `disambiguate-add-names`(despite a different ordering in the csl file). Not sure whether I misspecified the csl disambiguation rules or whether this is a local problem (using pandoc 2.11.2 and citeproc 0.2).

adam3smith:
The disambiguation rules have a fixed order when they're all present (add-givenname, add-names, year-suffix). It's incredibly rare for both add-names and add-givenname to apply to the same citation, so it doesn't really matter, but to the extent it does, I'd be surprised if anyone actually wanted et al. references with different first authors disambiguated by listing second authors rather than by clarifying that there are actually different first authors. 

Since the order of attributes is the only change here and that doesn't have any effect, I'm closing this -- thanks so much for communicating with the MIS editors!

PR #5176, opened 2020-12-18
Updates in research-institute-for-nature-and-forest.csl (development to version 0.1.2)

florisvdh:
A few small changes; essentially a `text` element has been dropped from the `title` macro if `type` is `webpage`.


adam3smith:
Thanks!

PR #5180, opened 2020-12-21
Create world-journal-of-hepatology.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/371034/#Comment_371034

I guess we can make the rest of the 42 journals for Bashideng.


rmzelle:
> I guess we can make the rest of the 42 journals for Bashideng.

Are you up for that? I found a list of title and ISSNs at https://www.wjgnet.com/bpg/GerInfo/259. Would help us get close to 10,000 styles! (https://pinux.info/csls_counter/)

At that point it would also be better to create a general BPG independent style, and turn this one into a dependent.

rmzelle:
Thanks!

PR #5182, opened 2020-12-22
Create Mammalogy-Notes.csl

dlizcano:
An Open Access Journal 
https://mammalogynotes.org/

### DIFF
--- a/mammalogy-notes.csl
+++ b/mammalogy-notes.csl
@@ -4,6 +4,7 @@
     <title>Mammalogy Notes</title>
     <title-short>MaNo</title-short>
     <id>http://www.zotero.org/styles/mammalogy-notes</id>
+    <link href="http://www.zotero.org/styles/mammalogy-notes" rel="self"/>
     <link href="http://www.zotero.org/styles/council-of-science-editors-author-date" rel="template"/>
     <link href="https://mammalogynotes.org/ojs/index.php/mn/about/submissions" rel="documentation"/>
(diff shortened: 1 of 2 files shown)


mammalogy-notes.csl
```diff
@@ -0,0 +1,255 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
```
rmzelle, line 2: @dlizcano, thanks. The `default-locale="en-US"` here fixes the style to US English. Is that correct? Or is the journal bilingual (English/Spanish)? If the latter, we would typically remove `default-locale` so that users can specify the desired locale to use with the style.

adam3smith: I looked through a couple of issues -- while the articles are mostly Spanish, the citations are mostly English, so the default-locale makes sense as is (some Spanish-language references do use Spanish terms, but we can't handle that distinction in CSL)

adam3smith:
@rmzelle good to merge from my end.

rmzelle:
Thanks!

mammalogy-notes.csl
dlizcano: The the journal is bilingual (English/Spanish). 
Thanks for accepting the citation-style.

PR #5183, opened 2020-12-22
Update ophthalmic-genetics.csl

gundestrup:
updates with correct style in text.
Ophtalmic Genetics uses (1) and not superscript as reference.
Please see documentation
https://www.tandfonline.com/action/authorSubmission?show=instructions&journalCode=iopg20#style


gundestrup:
will update path

PR #5184, opened 2020-12-22
correct et. at.

gundestrup:


gundestrup:
I don't understand the error, it looks like it is because I have moved the file from dependent?

adam3smith:
@retorquere -- could you check out the test run? I can't figure out why it's failing.

gundestrup:
Looks like the test is trying to test on the file that is not existing to
my eyes.
Since the files are created / deleted after the failed run...


ons. 23. dec. 2020 19.51 skrev Sebastian Karcher <notifications@github.com>:

retorquere:
It's because the patch fails to apply -- sheldon did not report that earlier but will from now on. I've cloned the styles repo and I've tried to apply https://patch-diff.githubusercontent.com/raw/citation-style-language/styles/pull/5184.patch, and that does indeed fail. I don't know what to make of that. diff thinks it's a reversed patch and if I have it assume `-R` it applies without error.

I've reverted Sheldon to manually apply the patch, but I really don't understand how the PR can say "Merging can be performed automatically" while the patch fails to apply.

PR #5185, opened 2020-12-23
fix issues in g-giappichelli-editore.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/363064/#Comment_363064

Can't find the guidelines, nor a public example, so just fixing the hardcoded french issues by using predefined terms.


rmzelle:
Thanks!

PR #5194, opened 2020-12-30
Move content from wiki pages to markdown files

bwiernik:
Moves content on style requests, style editing, and repository quality control to markdown files for easier access and version control. 

In doing so, I updated the Style Development guidelines to refer to two policy changes we've discussed for some time--using UUIDs instead of URIs for style ids and using localized locale labels instead of English labels. Are we good with updating those?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,11 +4,11 @@ The CSL style repository is the central location on the web for collecting and m
 Software products like Zotero, Mendeley, and Papers all draw their styles from our repository.
 
 We welcome style submissions (and corrections), and are particularly interested in styles for journals and published style guides.
-If you wish to submit a different type of style, please first check our [Criteria for Accepting Styles](https://github.com/citation-style-language/styles/wiki/Criteria-for-Accepting-Styles).
+If you wish to submit a different type of style, please first check our [Criteria for Accepting Styles](https://github.com/citation-style-language/styles/blob/master/README.md#criteria-for-inclusion).
 
 To submit a style, please follow the following steps (for locale files, follow the same steps in the  [locales](https://github.com/citation-style-language/locales) repository):
 
-#### 1. Check that your style meets all our [style requirements](https://github.com/citation-style-language/styles/wiki/Style-Requirements)
+#### 1. Check that your style meets all our [style requirements](https://github.com/citation-style-language/styles/blob/master/STYLE_REQUIREMENTS.md)
 
 #### 2. [Validate](https://validator.citationstyles.org/) your style against the CSL schema, and correct any validation errors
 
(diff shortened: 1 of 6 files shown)


rmzelle:
> using UUIDs instead of URIs for style ids and using localized locale labels instead of English labels. Are we good with updating those?

I don't think our CI currently allows for UUIDs as style IDs.

Happy with starting using localized locale labels in style titles, although we should probably fix up all the existing styles as well (and it would be nice if we could set up some CI to check `default-locale` values against our locale labels to catch cases where we forgot to add a locale label or made a formatting variation/typo, e.g. by using https://github.com/citation-style-language/locales/blob/master/locales.json, but that shouldn't be a showstopper).

bwiernik:
Should we update our CI to permit either an existing style Id or a UUID? Are those currently on travis or GitHub actions?

rmzelle:
> Should we update our CI to permit either an existing style Id or a UUID? Are those currently on travis or GitHub actions?

I think updating our CI was the only thing standing in the way of accepting UUIDs, so yes, once that's in order we can instruct submitters to use UIDS for novel styles. All tests are currently run on GitHub Actions.

bwiernik:
Does the basic logic of matching against a static list of the current ids or a UUID sound good to you?

adam3smith:
> a static list of the current ids

not sure that's a great idea -- we'd keep carrying around a list with 10,000 ids unless I'm misunderstanding. Could we just add uuid to the permitted ID set or is your concern that we want to enforce UUIDs? 
Also, we need a trivial mechanism in place for generating them for our often inexperience contributors, _especially_ if we're requiring them strictly.

bwiernik:
I think we should require UUIDs strictly. 

My thinking with the list was to permit existing styles to keep their URI IDs, but to force UUIDs for new styles. I suppose it's not too much work for us to manually enforce that.

Can GitHub actions detect if a new file is created versus an existing file modified? That would work for the distinction. 

For creating a UUID, how about we make the UUID in the CI script and instruct the submitter to change the ID to that value?

rmzelle:
> to markdown files for easier access and version control.

By the way, we don't have to make the conversion right away, but I think we should ideally start every sentence on a new line for markdown documentation.

rmzelle:
> In doing so, I updated the Style Development guidelines to refer to two policy changes we've discussed for some time--using UUIDs instead of URIs for style ids and using localized locale labels instead of English labels. Are we good with updating those?

@bwiernik, could you revert these edits, at least for this pull request? (in general, I always strongly prefer format conversions (like wiki > markdown file) to be separated from any content changes, so we get clear diffs of any edits; the current commit seems to contain both)

rmzelle:
And for the format conversion, it looks like the README.md and CONTRIBUTING.md have some links to the wiki: https://github.com/citation-style-language/styles/search?l=Markdown&q=wiki&type=code.

Other than that, I just need to update my own saved replies (https://github.com/settings/replies):

> Thanks for your submission. Is there an online style guide available for this style? Per https://github.com/citation-style-language/styles/wiki/Style-Requirements#10---documentation-link, we typically require each CSL style to include a link to the format it describes

> Thanks! Did you base this on an existing CSL style? If so, we'd like to know which one, so we can add a "template" link (see https://github.com/citation-style-language/styles/wiki/Style-Requirements#8---template-link).

rmzelle:
Thanks! For now we'll probably want to keep the wiki page active and add links to each of the now-migrated pages pointing to the new location.

rmzelle:
> add links to each of the now-migrated pages pointing to the new location.

Done.

PR #5195, opened 2020-12-31
New version of ABNT style

bcdavasconcelos:
This version of the ABNT style is an adaptation for use in the field of classical studies and ancient philosophy. It should be useful for all those who are working in these areas in Brazilian universities.


adam3smith:
Thanks and sorry for the wait -- is this variant documented somewhere, e.g. on the page of a classics department, or is it used in a specific journal? Given the multitude of options for formatting, we're reluctant to accept styles based on a general sense of them being generally a good fit.

bcdavasconcelos:
Hi, and likewise, apologies for the late reply.   
No, not really. The problem is that the field of classics is full of _ad hoc_ rules when it comes to citing sources, because each author (and sometimes each work) has an edition that is taken as the main reference. Because of this, one is often forced to refer to the edition of a classical work (and not the author or work itself) or the translator of the work, especially whilst doing philological work. This style is an effort to address that within the framework of the ABNT rules.   
Be it as it may, I already reviewed the CSL file, so the one I submited is already outdate. Feel free to throw it away ;)

PR #5196, opened 2021-01-04
Create pedosphere.csl

POBrien333:
via https://forums.zotero.org/discussion/78159/style-request-pedosphere#latest

Request is from 2019 and I was never able to find good info. At some stage it must've moved to Elsevier, but with it's unique style.

Also, sounds like a web forum for pedos. haha


rmzelle:
Thanks!

PR #5197, opened 2021-01-04
Create the-quarterly-journal-of-economics.csl

POBrien333:
via https://forums.zotero.org/discussion/83576/style-for-quarterly-journal-of-economics#latest


adam3smith:
(Half of the examples had straight quotes half regular ones -- pretty sure the latter is right).

PR #5198, opened 2021-01-04
Create open-gender-journal.csl

POBrien333:
via https://forums.zotero.org/discussion/86318/style-request-open-gender-journal#latest


adam3smith:
Thanks!

PR #5199, opened 2021-01-04
Update quaternaire.csl

flosigime:
line 234, suppress ' entry-spacing="1" line-spacing="1" '

### DIFF
--- a/quaternaire.csl
+++ b/quaternaire.csl
@@ -53,7 +53,7 @@
   </macro>
   <macro name="editor-chapter">
     <names variable="editor">
-      <name and="text" delimiter-precedes-last="never" initialize-with=". " sort-separator=" "/>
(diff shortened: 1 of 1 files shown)


quaternaire.csl
```diff
@@ -231,7 +231,7 @@
     </layout>
   </citation>
   <!-- ***************** Bibliographie ***************** -->
-  <bibliography entry-spacing="1" line-spacing="1">
+  <bibliography>
```
adam3smith, line 234: This is fine, but any reason for the change? These are the default values, so this change wouldn't actually change the citation style output.

flosigime: yes, in ms word and LO writer, it creates two blank lines, and if i suppress it, i have no blank lines any more. It's strange, if it is the default values...

image

rmzelle:
@flosigime, thanks!

PR #5200, opened 2021-01-05
Update quaternaire.csl

flosigime:
line 56, change initialize-with=". " in initialize-with="."

### DIFF
--- a/quaternaire.csl
+++ b/quaternaire.csl
@@ -53,7 +53,7 @@
   </macro>
   <macro name="editor-chapter">
     <names variable="editor">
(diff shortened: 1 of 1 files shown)


adam3smith:
Note that you want to make multiple changes to the same style in the same pull request. If we merge #5199 , merging this would show a merge conflict. (This change looks good. you could just change it to the previous pull request)

flosigime:
arf, sorry for that, github is still a bit weird to me. So, i make this change in #5199 ?

adam3smith:
> So, i make this change in #5199 ?

Yes, as I see you've already done, thanks. Closing here.

PR #5201, opened 2021-01-05
Create haute-ecole de gestion-de geneve-iso-690-fr.csl

grolimur:
French version of the already existing citation style (`haute-ecole-de-gestion-de-geneve-iso-690.csl` soon to be renamed `haute-ecole-de-gestion-de-geneve-iso-690-en.csl` for consistency)


grolimur:
The checks failed but I don't understand why. Can someone give me any tips on that?

adam3smith:
Something went rather wrong here: The reason the checks fail is that the style is in the dependent folder but shouldn't be (it's an independent style). But you also appear to have deleted the branch from which you made the pull request?

Also, as a question: we do need separate citation styles per language? This couldn't just be handled using the locales?

grolimur:
Now that you raise this @adam3smith, I realise that I can indeed handle that using locales.
I update the file and create a new pull request.

PR #5207, opened 2021-01-10
Remake mammallia.csl for Oct/2020 guidelines.

POBrien333:
via https://forums.zotero.org/discussion/87015/request-for-citation-style-urgent#latest


mammalia.csl
```diff
@@ -52,14 +53,43 @@
     </names>
   </macro>
   <macro name="title">
-    <text variable="title"/>
+    <text variable="title" font-style="italic"/>
```
adam3smith, line 56: That's not right, is it? Looks like should be italics just for books & similar, not for journals & chapters?

adam3smith:
Thanks!

PR #5215, opened 2021-01-15
Create comparativ.csl

POBrien333:
via https://forums.zotero.org/discussion/86240/style-request-journal-comparativ#latest


adam3smith:
It looks like the date for journal articles is missing, which requires some more re-writing since you use the volref macro across reference types

adam3smith:
(I checked on the ibid issue noted by requester -- style looks good & ibid. works for me as expected)

POBrien333:
You're a star! Thanks.

PR #5217, opened 2021-01-16
fix issue with issued for journal-of-international-business-studies.csl

POBrien333:
https://forums.zotero.org/discussion/87101/issue-with-jibs-reference-style#latest


adam3smith:
Thanks!

PR #5218, opened 2021-01-16
Removed space in year only citation

rknx:
When author name is suppressed in in-line citation, a space would be prefixed before year. This commit removes that space.


adam3smith:
Great, thanks!

PR #5221, opened 2021-01-18
Create vcs.csl

florianjansen:
CSL style for vcs.pensoft.net, validated with https://validator.citationstyles.org/


adam3smith:
It looks to me like the generic Pensoft style should work just fine here (https://www.zotero.org/styles?q=id%3Apensoft-journals ) is that not the case? How does it differ?

PR #5223, opened 2021-01-19
Create Leidraad voor juridische auteurs 2019 (Dutch)

pjrwees:
This style extends leidraad-voor-juridische-auteurs-dutch.csl but it reflects the 2019 version of the Leidraad style manual. Please let me know if I should edit the existing style.


adam3smith:
Thanks -- what's the adoption situation of the style in the Netherlands? Is everyone immediately going to use the new version, or will there be people on the old Leidraad edition and others on the new one? For the latter case, we typically rename the _old_ style and make the new one the standard Leidraad one.

pjrwees:
Journals and most authors use the 2019 version by now. I do not see why anyone would use the older versions. Also, the old style was based on the 2013 Leidraad, skipping the 2016 version.

rmzelle:
@pjrwees, in that case it's better to just update the existing style. (I already made that change and will accept the changes now)

Hartelijk bedankt!

PR #5225, opened 2021-01-20
Create Quoc-Thai-Ngo-ModificationCiteThemRight-10th.csl

wanincoco:
This is the style that some of the University in Vietnam use (seperate between Vnese and other language references)
All you need to do is add 'V" to the note of the Vnese citation


adam3smith:
I don't love using the note variable for sorting, I'm afraid. I'm guessing you're doing this to sort by language, which is understandable, but we don't really want to encourage that sort of data entry. FWIW, CSL will shortly update to a new version which includes a dedicated language variable.

Otherwise, there's a fair amount of clean-up required in this style before we can take it -- some of it is just redundant, other things are, as per the validation, just not going to work reliably-- how did you create it? Sticking closer to an existing style would be a better idea.

adam3smith:
@wanincoco see my questions above, please?

adam3smith:
I'm closing this for inactivity for now, but we're always happy to re-open if you come back to working on this -- just leave a comment here.

PR #5228, opened 2021-01-25
Add new style for U of Mannheim, Germanistische Linguistik

zuphilip:
This is a new style for the University of Mannheim, Germanistische Linguistik used for Bachelor/Master Thesis. The documentation gives sometimes some choices, but we fixed here for the CSL one of them. Moreover, some points seem to be inconsistencies/errors in the documentation. I spoke this through with someone from the chair.

### DIFF
--- a/universitat-mannheim-germanistische-linguistik.csl
+++ b/universitat-mannheim-germanistische-linguistik.csl
@@ -1,0 +1,298 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="de-DE">
+  <!-- This style was edited with the Visual CSL Editor (https://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>Universität Mannheim - Germanistische Linguistik (German)</title>
+    <id>http://www.zotero.org/styles/universitat-mannheim-germanistische-linguistik</id>
+    <link href="http://www.zotero.org/styles/universitat-mannheim-germanistische-linguistik" rel="self"/>
+    <link href="http://www.zotero.org/styles/universitat-bremen-institut-fur-politikwissenschaft" rel="template"/>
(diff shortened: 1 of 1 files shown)


universitat-mannheim-germanistische-linguistik.csl
```diff
@@ -0,0 +1,293 @@
...
+        </else>
+      </choose>
+    </group>
+  </macro>
+  <macro name="locator">
+    <choose>
+      <if type="article-journal">
+        <choose>
+          <if variable="volume" match="none">
+            <text variable="DOI" prefix="DOI: "/>
```
adam3smith, line 155: This seems odd? [Above](https://github.com/citation-style-language/styles/pull/5228/files#diff-32f81e0fcfd685671f7aaca703b7ea32a6bc49f691674810465684aec1d12094R78) you use the "correctly" displayed DOI, here with the much less useful DOI: prefix -- is that on purpose? If so, could you convince them to change this in the guide?

universitat-mannheim-germanistische-linguistik.csl
```diff
@@ -0,0 +1,293 @@
...
+  </macro>
+  <macro name="container-prefix">
+    <choose>
+      <if type="chapter article-journal article-newspaper article-magazine" match="any">
+        <text term="in" text-case="capitalize-first" suffix=":"/>
+      </if>
+    </choose>
+    <choose>
+      <if type="paper-conference" match="any">
+        <text value="präsentiert auf" text-case="capitalize-first" suffix=": "/>
```
adam3smith, line 194: use term instead

adam3smith (changes requested):
Two quick things -- looks good otherwise!

universitat-mannheim-germanistische-linguistik.csl
zuphilip: Oh yes, that is more like a left-over from the template style used here. For journal articles the volume should always been shown and if the article is an online article, then additionally the URL. That is what the documentation says. I extended this to use the DOI instead of the URL whenever possible. -- This is fixed in the new version.

zuphilip:
Thank you for the report @adam3smith. Please have a look at the new version, where I fixed both issues.

rmzelle:
@zuphilip, great, thanks!

PR #5229, opened 2021-01-26
fix small formatting issues for mclc.csl

POBrien333:
https://forums.zotero.org/discussion/comment/373377/#Comment_373377


modern-chinese-literature-and-culture.csl
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="expanded">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
```
adam3smith, line 2: This change seems wrong: expanded is the preferred way to handle full page ranges (as it will also expand them if they're in the metadata in short form). Were you able to replicate the reported issue with page ranges?

POBrien333: I was wondering about that.
When I looked the page range was minimalised, but since it was on "expanded" I wondered what was going on. Now I can't reproduce it.
Thanks for double checking.

rmzelle:
Thanks!

PR #5230, opened 2021-01-26
Update zeitschrift-fur-theologie-und-philosophie.csl

knorb:
Updated according to modified style sheet of the journal (as of Dec 01, 2020): 
https://ztp.jesuiten.org/ojs2/index.php/ZTP/about/submissions; 
https://drive.google.com/file/d/1GFU2oCwkN9J8a1Fy1Zf4527NKe5a0zYX/view?usp=sharing


adam3smith:
Great, thanks!

PR #5231, opened 2021-01-27
Update geochimica-et-cosmochimica-acta.csl

martinvoigt:
Added collapsed citations

Documentation: https://www.elsevier.com/journals/geochimica-et-cosmochimica-acta/0016-7037/guide-for-authors

> Multiple citations having the same authorship information are sorted by the earliest year, with all years groups together, e.g., (Li et al., 1996; Brown et al. 2001, 2008a,b, 2013; Stone and Smith, 2004).


adam3smith:
Thanks!

PR #5236, opened 2021-01-31
Update technische-universitat-dresden-medizin.csl

POBrien333:
via https://forums.zotero.org/discussion/87396/changing-the-stile-to-first-author-year-a-or-b-by-more-than-one-identical-author#latest


rmzelle:
Thanks!

PR #5238, opened 2021-01-31
Update author-year disambiguation

rknx:
APS uses year-suffix (like Jones et. al 2019b) to disambiguate rather than first name or second author. Since all 3 disambiguation methods were kept true, it was causing multiple authors to show up in in-line citations. This commit removes first name disambiguation and second author disambiguation and only keeps the year-suffix disambiguation.


rmzelle:
Thanks!

PR #5240, opened 2021-02-02
Fix variable used for the label after indication of number of pages

nchachereau:
A small fix to the variable referenced by the label for displaying the number of pages. 

(Interestingly, the original coding did not cause problems with Zotero, even though the CSL specification [states](https://citation-style-language.readthedocs.io/en/stable/specification.html#label) that the label should not be rendered if the selected variable is empty. citeproc-js seems to somehow match `page` and `number-of-pages`. pandoc-citeproc, however, is not so lenient, [as a user of this style discovered](https://github.com/LausanneCitationStyle/Lausanne/issues/19)).


rmzelle:
Thanks!

PR #5241, opened 2021-02-03
Create IETEJR

matheen03:


rmzelle:
@matheen03, thanks. Is this style for the "IETE Journal of Research" (https://www.tandfonline.com/toc/tijr20/current)?

If so, do you know if this journal follows standard IEEE style, or did you have to make format changes to the CSL style?

adam3smith:
At a minimum, in-text citations here are different -- the journal uses [1,2] instead of IEEE's [1],[2]. That'd also still need to get fixed in the style by moving the square brackets to the layout.

PR #5242, opened 2021-02-03
Change "Dutch" to "Nederlands" in titles

rmzelle:
Follow-up of https://github.com/citation-style-language/styles/pull/5194.

@bwiernik, I'm wondering if we should have a dedicated PR for each language. Might be slightly easier on Zotero's update infrastructure, as we'll be changing a lot of styles. Could also use this opportunity to do some manual checks to make sure every style with a default-locale value has the appropriate locale descriptor in the title, although eventually we should have a CI check for this.


bwiernik:
That sounds good to me.

PR #5243, opened 2021-02-03
Make world-journal-of-hepatology.csl to bpg.csl parent

POBrien333:
Goes with independent journals https://github.com/citation-style-language/journals/pull/115


POBrien333:
Not sure why it's not validating.

rmzelle:
> Not sure why it's not validating.

We moved some documentation from the wiki to the main repository, and forgot to exempt these files from our check against non-CSL files. Unrelated :).

rmzelle:
@POBrien333, thanks! (I converted world-journal-of-hepatology.csl to a dependent style so it doesn't disappear while I prep the other dependents; tiny bit cleaner)

PR #5245, opened 2021-02-03
Create journal-of-open-research-software.csl

POBrien333:
via https://forums.zotero.org/discussion/87460/style-request-journal-of-open-research-software#latest


rmzelle:
Thanks!

POBrien333:
Thanks for merging.
Looks like there is an issue with the repository, as the style doesn't show up on zoteroo.org/styles
https://www.zotero.org/styles?q=journal%20of%20open%20research

PR #5249, opened 2021-02-05
Create crispr-journal.csl

hjmalloy:
Updated citation format to conform to The Crispr Journal reference guidelines.


rmzelle:
Thanks!

PR #5259, opened 2021-02-11
Create journal-of-surgical-oncology.csl

POBrien333:
via https://forums.zotero.org/discussion/84001/style-request-journal-of-surgical-oncology#latest


adam3smith:
This doesn't look right as AMA -- I'm seeing citations in square brackets and colons after author list in the styleguide -- didn't check further for differences

POBrien333:
Made it an independent style, but with AMA base.
Free paper here: https://onlinelibrary.wiley.com/doi/epdf/10.1002/jso.26350

rmzelle:
@POBrien333, thanks!

PR #5260, opened 2021-02-11
Create the-sociological-review.csl

POBrien333:
via https://forums.zotero.org/discussion/86762/style-request-the-sociological-review#latest

### DIFF
--- a/dependent/the-sociological-review.csl
+++ b/dependent/the-sociological-review.csl
@@ -7,6 +7,7 @@
     <link href="http://www.zotero.org/styles/apa" rel="independent-parent"/>
     <link href="https://www.thesociologicalreview.com/submit-a-paper/technical-notes-for-authors/" rel="documentation"/>
     <category citation-format="author-date"/>
+    <category field="social_science"/>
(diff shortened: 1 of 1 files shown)


rmzelle:
@POBrien333, if this is APA style, are you sure the localization should be "en-GB"?

rmzelle:
@bwiernik, do you know if the latest APA CSL style behaves well with "en-GB" as locale? (we have a UK journal here that says it's using APA style, so I'm wondering if we should keep it with "en-US" or could have the dependent style override the locale to "en-GB")

cc @adam3smith

adam3smith:
The terms are all defined using lang="en" and the style doesn't use quotes, so the only thing that will change is the date format:
It's "February 13, 2021" in en-US and "13 February 2021" in en-GB. If the styleguide doesn't specify the date format and only says APA, I'd assume the US format is correct and use en-US

bwiernik:
Some item types like magazines give the bibliography date as (2020, September 4)--not sure if it's worth bothering with a new style just to modify that.

rmzelle:
Thanks all!

PR #5261, opened 2021-02-11
Create qeios.csl

QeiosLab:
Added file containing the citation style for Qeios (qeios.com) ISSN 2632-3834

### DIFF
--- a/qeios.csl
+++ b/qeios.csl
@@ -37,9 +37,9 @@
       <term name="retrieved">available from</term>
       <term name="section" form="short">sect.</term>
       <term name="et-al">et al.</term>
-	  <term name="page" form="short">
-      	<single>p.</single>
-      	<multiple>pp.</multiple>
+      <term name="page" form="short">
+        <single>p.</single>
+        <multiple>pp.</multiple>
       </term>
     </terms>
   </locale>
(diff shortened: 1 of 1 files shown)


rmzelle:
@QeiosLab, thanks. Could you please make sure that "Allow edits from maintainers" in the right-hand column of this pull request is checked? (see https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests)

I'd like to make a few minor XML format changes.

QeiosLab:
@rmzelle, thank you very much. We've just checked the box.

adam3smith:
Could you explain what up with the curly brackets? `{m/` etc. -- they don't appear in the styleguide and I don't quite understand why you'd want them in a reference?

QeiosLab:
@adam3smith, thanks for asking. Basically, the reason is that we’d like the references to be readable also by machines.
For example, that characters sequence would allow us to parse the references when you submit the article’s full text. After that, we’d automatically remove the special characters and publish the article with well-polished references.

rmzelle:
> Basically, the reason is that we’d like the references to be readable also by machines.

Are authors asked to use this format as well in their manuscripts? Or is this machine-readable format only for internal use?

QeiosLab:
@rmzelle, on Qeios, references can be listed in any standard referencing style as long as the style is clear, intelligible, and remains consistent within a given work (Article, Definition, or Peer Review). Even though we remove the curly brackets on copy-paste from Word, authors can keep them or they can compose new works directly on the platform using that style. In both cases, we will accept it as it doesn’t break any rules. We’ve updated our styleguides accordingly.

rmzelle:
@QeiosLab, thanks, we just accepted the style into the repository.

We just wanted to make sure this style wasn't solely used internally, in which case wide distribution through the style repository wouldn't be of much use (as you would be the only user). But if you share this style with authors it definitely qualifies for inclusion.

PR #5265, opened 2021-02-12
Update european-journal-of-international-law.csl

cathbrend:
Added a semi-colon to separate 2 subsequent citations in a note


adam3smith:
Thanks -- not sure what's up with the preview for citations.

PR #5268, opened 2021-02-14
move filters to inert file to pacify Sheldon

### LINKED ISSUE #5263: Github Actions fails on edit to non-CSL files
@retorquere, Github Actions seems to always fail when we make edits to non-CSL files. See e.g. the edit to file "spec/spec_helper.rb" in https://github.com/citation-style-language/styles/pull/5256/ which causes a failure in https://github.com/citation-style-language/styles/runs/1884910562.

retorquere:
Closes #5263


rmzelle:
@retorquere, thanks!

PR #5269, opened 2021-02-14
Major update to Gallia.csl

NMonteix:


rmzelle:
@NMonteix, did you mean to make bigger changes? Right now you're only updating the documentation link, but "Major update" suggests there would be more.

NMonteix:
Indeed it was meant to be bigger... Wrong file (in fact a strange thing from Zotero: I installed the new style this summer, used it, it disappeared and came back to the official repo version on the 12th of February, and this version remained instead of my new installation of the changed file).
Sorry for that

rmzelle:
> (in fact a strange thing from Zotero: I installed the new style this summer, used it, it disappeared and came back to the official repo version on the 12th of February, and this version remained instead of my new installation of the changed file).

I made a change to the style a few days ago (instead of indicating the language of non-English styles in the title in English ("French"), we switched to using that language itself ("Français"). If you used a custom version with the same style ID, that update would have triggered Zotero to update your version to the latest repository version.

rmzelle:
Merci!

NMonteix:
Merci à toi


Le 15/02/2021 à 14:30, Rintze M. Zelle a écrit :

PR #5280, opened 2021-02-18
Adding publishers' names

carljohanberglund:
The editor of this journal has started requiring the names of publishers in addition to their location. This patch reflects this change.


rmzelle:
@carljohanberglund, thanks!

PR #5281, opened 2021-02-19
Update journal-of-orthopaedic-trauma.csl

POBrien333:
revamp of style: https://forums.zotero.org/discussion/87771/style-error-journal-of-orthopaedic-trauma#latest

Current free article can be found here: https://journals.lww.com/jorthotrauma/toc/2020/11003


rmzelle:
@POBrien333, thanks!

PR #5282, opened 2021-02-19
Add Haaga-Helia University of Applied Sciences Harvard style

mrautio:
* This PR adds the referencing style used in [Haaga-Helia University of Applied Sciences](https://www.haaga-helia.fi/en).
* Style includes en-GB and fi-FI locales.
* [Referencing style's "conformance tests"](https://github.com/mrautio/hh-templates/tree/main/test/cases/03_references) were done using pandoc.

### DIFF
--- a/haaga-helia-university-of-applied-sciences-harvard.csl
+++ b/haaga-helia-university-of-applied-sciences-harvard.csl
@@ -1,0 +1,246 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="fi-FI">
+  <info>
+    <title>Haaga-Helia University of Applied Sciences - Harvard</title>
+    <id>http://www.zotero.org/styles/haaga-helia-university-of-applied-sciences-harvard</id>
+    <link href="http://www.zotero.org/styles/haaga-helia-university-of-applied-sciences-harvard" rel="self"/>
+    <link href="https://libguides.haaga-helia.fi/referencing" xml:lang="en-GB" rel="documentation"/>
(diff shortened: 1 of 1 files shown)


rmzelle:
@mrautio, thanks! Two things:

* Some reference managers (like Zotero) don't allow a style's default-locale to be overridden, so if this style is meant to be usable in both Finnish and British English, we would typically not include a default-locale.
* As this is a Finnish university, we would typically give the style a Finnish name (i.e. "Haaga-Helia ammattikorkeakoulu")

If that sounds reasonable we can make these changes before accepting the style.

mrautio:
@rmzelle Both comments sounds reasonable to me! I've removed default-locale now.

I'm keeping the title as-is just in case: The other FI counterpart, [Metropolia](https://github.com/citation-style-language/styles/blob/master/metropolia-university-of-applied-sciences-harvard.csl), has an English title. I.e. probably Haaga-Helia and Metropolia both should follow the same title convention.

rmzelle:
> I'm keeping the title as-is just in case: The other FI counterpart, [Metropolia](https://github.com/citation-style-language/styles/blob/master/metropolia-university-of-applied-sciences-harvard.csl), has an English title. I.e. probably Haaga-Helia and Metropolia both should follow the same title convention.

We can rename that style to "Metropolia Ammattikorkeakoulu" for consistency, which might make sense as it does have a default-locale set to Finnish.

mrautio:
Sounds good :)

adam3smith:
@rmzelle -- I left the ID and filename for the other style alone, but not sure what to do in this case. You want to rename? Also not sure about lowercase or uppercase for Ammattikorkeakoulu

rmzelle:
> I left the ID and filename for the other style alone, but not sure what to do in this case. You want to rename? Also not sure about lowercase or uppercase for Ammattikorkeakoulu

I think we're good as-is.

@mrautio, thanks!

PR #5285, opened 2021-02-21
Fix Haaga-Helia style's locales

mrautio:
Quickly checked [Zotero style repository](https://www.zotero.org/styles/?q=haaga-helia) and realized that the locale setup is probably too pedantic in the configuration.

Incorrect render (no locale match?) in zotero.org/styles example output:
```
Hisakata, R., Nishida, S. and Johnston, A. 2016. An adaptable metric shapes perceptual space. Current Biology, 26, 14, page 1911–1915. available at: https://doi.org/10.1016/j.cub.2016.05.047. Accessed: October 3, 2016.
```

Expected output:
```
Hisakata, R., Nishida, S. & Johnston, A. 2016. An adaptable metric shapes perceptual space. Current Biology, 26, 14, pp. 1911–1915. URL: https://doi.org/10.1016/j.cub.2016.05.047. Accessed: October 3, 2016.
```

### DIFF
--- a/haaga-helia-university-of-applied-sciences-harvard.csl
+++ b/haaga-helia-university-of-applied-sciences-harvard.csl
@@ -13,7 +13,7 @@
     <category citation-format="author-date"/>
     <category field="generic-base"/>
     <summary>Haaga-Helia University of Applied Sciences referencing style (Finnish and English)</summary>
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks! I'm pretty sure the Zotero Style Repository renders styles with "en-US" if no default-locale is specified.

PR #5286, opened 2021-02-21
Update universite-du-quebec-a-montreal-departement-dhistoire.csl

McKaulick:
- Removed uppercase in op. cit. author last name in citation
- Changed "in" to "dans" for book chapter (should be in French)
- Removed the italic for the container-title
- Removed parenthesis prefix and suffix for a URL
- Removed suffix to coll.
- Removed short title for op. cit.
- Moved collection before year.


rmzelle:
Thanks!

McKaulick:
@rmzelle Thanks for you help !

PR #5287, opened 2021-02-21
Update journal-of-orthopaedic-trauma.csl

POBrien333:
small error I had introduced with the last changes: https://forums.zotero.org/discussion/comment/375713/#Comment_375713


rmzelle:
Thanks!

PR #5289, opened 2021-02-22
Update early-christianity.csl

carljohanberglund:
Correcting author name initialization, volume–issue separation, and commas.


rmzelle:
@carljohanberglund, thanks!

PR #5290, opened 2021-02-24
Create cns-spectrums.csl

POBrien333:
via https://forums.zotero.org/discussion/87885/cns-spectrums-style-missing-in-zotero#latest


rmzelle:
Thanks!

PR #5299, opened 2021-02-25

alex-ball:
Brings the style in line with new reference specifications introduced in 2020:

- Book known by its title
- Entry from a print book known by its title
- Entry from an online book known by its title
- Online map

### DIFF
--- a/harvard-university-of-bath.csl
+++ b/harvard-university-of-bath.csl
@@ -141,7 +141,7 @@
       <if type="bill broadcast book dataset graphic legal_case map motion_picture musical_score patent report song speech webpage thesis" match="any">
         <text variable="title" font-style="italic"/>
       </if>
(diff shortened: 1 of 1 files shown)


harvard-university-of-bath.csl
```diff
@@ -449,6 +490,19 @@
...
+            <text macro="edition"/>
+            <group delimiter=" ">
+              <text variable="title" font-style="italic"/>
+              <text macro="online"/>
+            </group>
+            <text macro="series"/>
+            <text macro="status"/>
+            <text macro="publisher"/>
+          </group>
+        </else-if>
```
rmzelle, line 505: @alex-ball, I cleaned up the use of groups here. Output should be the same, but I edited this by hand without testing, so let me know if I made a mistake here.

rmzelle:
Thanks!

Issue #5224, opened 2021-01-20
Copernicus update - preprints and software details


copernicus-publications:
Hi all,

@POBrien333 was so kind to update the Copernicus style recently, thank you for that!
Because that got rolling, I decided to implement a thorough update, the old style was hopelessly out of date (our fault).
I picked up where @POBrien333 left and think I have it all up to speed now.
Before I create the pull request, I was wondering if anyone could offer advice:

Software:
because we feature dataset and software citations as well, I have implemented both.
However, Zotero being on 1.0.1 and not master, it complains upon import that `type=software` is not valid.
To not scare away current Zotero users, should I just leave software out and rather commit a quick update later?

Preprints:
we feature a lot of preprints, so I decided to use `type=article ` as recommended and add the suffix we require ( [preprint] ) to all of those types. (see under References [here](https://publications.copernicus.org/for_authors/manuscript_preparation.html))
However, I was wondering if there is a smarter way? If I could define a "sub-type" preprint somewhere so that only those receive the [preprint] suffix, that would be the cleanest way, but I couldn't come up with a smart solution.
Any ideas?

If no one objects, I will commit in a few days.
Thank you all for the hard work!

Edit: relating to #4367 : I have used the `archive` field for the platform as well (arxiv.org, Copernicus preprint servers, biomed etc) but I am happy to do otherwise! 
It was [discussed ](https://discourse.citationstyles.org/t/preprint-dataset-repository-what-csl-variable/1570)in this crosspost.
Using `publisher` is a bad choice, as many people will be adamant that preprints are not _published_.
The same goes for datasets btw, a single publisher could have many repositories that hold the data (NASA has many..), but what you want to identify by using `archive` is the repository.

Issue #4367, opened 2019-10-17
Preprint/dataset repository: What CSL variable?


bwiernik:
Encountering a wrinkle with preprints and datasets. From most preprint repositories, the name of the server is getting stored by Zotero's translators as `container-title`. This had seemed like a good idea (ala Journal titles). However, I am now seeing that some preprints and many datasets import with `container-title` used to indicate the collection or larger body of works the item belongs to (e.g., doi.org/10.3886/ICPSR35531.v3). 

Conceptually, I think the preprint/data repository makes good conceptual sense as either `archive` or `publisher` rather than `container-title` (this is, for example, how the OSF repositories export in their metadata [as `publisher`], and CrossRef's posted content type uses "group-id", which would fit well with `publisher`; ICPSR stores its name in `publisher`). Most citation styles print `publisher`, so it seems like this would work well without many changes to existing styles.

Could we come to a consensus as to where these best belong?
(Cross-posted to https://discourse.citationstyles.org/t/preprint-dataset-repository-what-csl-variable/1570)

adam3smith:
(I think we should just not apply stale bot to regular tickets)

rmzelle:
> (I think we should just not apply stale bot to regular tickets)

See https://github.com/citation-style-language/styles/pull/4466. Feel free to merge. (I think we might also want to disable autoclosing of PRs)

bwiernik:
Let's go with `publisher` here

PR #5303, opened 2021-03-04

copernicus-publications:
I tried to get some input related to this style in #5224, but no one responded. 
I deleted type="software" which I had originally included, to be compliant with 1.0.1 (the validator did not check out otherwise).
Also, for type="dataset" I chose to refer to the repository in the field "archive" for lack of guidance what a better solution would be.
There has been a discussion about this here #4367 
but to no end. 

Otherwise, this now reproduces the style for all journals by Copernicus Publications.

### DIFF
--- a/copernicus-publications.csl
+++ b/copernicus-publications.csl
@@ -4,7 +4,7 @@
     <title>Copernicus Publications</title>
     <id>http://www.zotero.org/styles/copernicus-publications</id>
     <link href="http://www.zotero.org/styles/copernicus-publications" rel="self"/>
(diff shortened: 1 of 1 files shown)


rmzelle:
@copernicus-publications, thanks, and sorry for the lack of response to your other issue.

The CSL style repository is still on CSL 1.0.1. For the other points, @bwiernik, @adam3smith, do you have any recommendations?

copernicus-publications:
@rmzelle no problem at all. I just wanted to be as compliant with current standards as possible. I thought maybe there is a "best practice" on how to deal with data publications and some of the other issues touched.

PR #5305, opened 2021-03-04
Sort American Physiological Society citation style according to updated author instructions

mz2:
Sort the bibliography in order of citation number, instead of primarily by author name and secondarily by issued date

### DIFF
--- a/american-physiological-society.csl
+++ b/american-physiological-society.csl
@@ -160,8 +160,7 @@
   </citation>
   <bibliography second-field-align="flush">
     <sort>
-      <key macro="author"/>
-      <key variable="issued"/>
+      <key variable="citation-number"/>
     </sort>
     <layout suffix=".">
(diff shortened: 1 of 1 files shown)


rmzelle:
@mz2, https://journals.physiology.org/author-info.manuscript-composition#EXAMPLE%20REFERENCES says "Reference lists should be arranged alphabetically by author".

adam3smith:
Matias emailed me to flag this; it's requested and confirmed by the managing editor so presumably correct but since a style that goes against published guidelines is going to create confusion for users. @mz2 could you check with Karie when they're planning to update the public guidance?

kariek17:
Our Reference guidelines are published here: https://journals.physiology.org/author-info.references

kariek17:
This style also applies to the Journal of Neurophysiology, which it had not previously. Just an FYI.

PR #5307, opened 2021-03-05
Allow privileged testing of PRs

retorquere:
With this PR, the "sheldon" test for PRs can run in one of two modes:

## standard mode

* The workflow checks out master on citation-style-language/styles to get the test framework
* The workflow checks out the PR into `pull-request`.
* The tests run the code from master, but spec_helper sees it is testing a PR, so loads the styles from there instead

## privileged mode

This only kicks in when the PR is labeled (edit: with a specific label that will trigger this behavior). Only repo owners can label, so this should be safe. Where with manual runs, each run would have to be started manually anew if you want to re-test the PR after syncing new commits, the label remains so would only have to be set once, and you also can clearly see which PRs run in privileged mode.

* The workflow checks out the PR *and runs the test from there*

----

I'll submit a buddy PR for the locales repo to update the spec_helper if this PR gets accepted. This PR removes Sheldon's diff-vetting; we could still apply that, but this setup should have the same safeguards against PR-exfiltration. The only extra of vetting is that PR tests will fail for mixed PRs as an extra alert, but I'm not sure whether that's worth having now.

### DIFF
--- a/.github/workflows/sheldon.yaml
+++ b/.github/workflows/sheldon.yaml
@@ -20,10 +20,10 @@ jobs:
 
     # otherwise, checkout the current master, and the pr to the subdirectory 'pr'
     - uses: actions/checkout@v2
-      if: ! contains(github.event.pull_request.labels.*.name, 'owner-test')
+      if: "! contains(github.event.pull_request.labels.*.name, 'owner-test')"
     - uses: actions/checkout@v2
       path: pull-request
-      if: ! contains(github.event.pull_request.labels.*.name, 'owner-test')
+      if: "! contains(github.event.pull_request.labels.*.name, 'owner-test')"
       with:
         ref: ${{ github.event.pull_request.head.sha }}
 
--- a/.github/workflows/sheldon.yaml
+++ b/.github/workflows/sheldon.yaml
@@ -14,15 +14,18 @@ jobs:
     # this can only be triggered by people with repo write access -- such as people that can add
     # labels to a PR
     # https://securitylab.github.com/research/github-actions-preventing-pwn-requests
-    - uses: actions/checkout@v2
+    - name: Checkout repo for OWNER TEST
+      uses: actions/checkout@v2
       if: contains(github.event.pull_request.labels.*.name, 'owner-test')
       with:
         ref: ${{ github.event.pull_request.head.sha }}
 
     # otherwise, checkout the current master, and the pr to the subdirectory 'pr'
-    - uses: actions/checkout@v2
+    - name: Checkout base repo for pull-request test
+      uses: actions/checkout@v2
       if: "! contains(github.event.pull_request.labels.*.name, 'owner-test')"
-    - uses: actions/checkout@v2
+    - name: Checkout pull-request
+      uses: actions/checkout@v2
       if: "! contains(github.event.pull_request.labels.*.name, 'owner-test')"
       with:
         path: pull-request
--- a/.github/workflows/sheldon.yaml
+++ b/.github/workflows/sheldon.yaml
@@ -2,7 +2,7 @@ name: Pull request feedback
 
 on:
   pull_request_target:
-    types: [ opened, synchronize ]
+    types: [ opened, synchronize, labeled ]
 
 jobs:
   test:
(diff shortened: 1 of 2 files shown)


rmzelle:
Does privileged mode get activated with any label, or a specific one?

retorquere:
Specific one -- the workflow currently uses `owner-test`, but it can be anything you choose.

retorquere:
Oh I forgot -- Sheldon does still do asset vetting as it renders the styles. That could also be simplified after this -- Sheldon currently fetches the changed styles via HTTP, that is no longer required because the assets are already on disk. But Sheldon will work as it currently is.

rmzelle:
@retorquere, thanks!

rmzelle:
Did the build for https://github.com/citation-style-language/styles/pull/5309 (https://github.com/citation-style-language/styles/runs/2053700809) behave like expected? It seems like it ran the tests without the need for a label.

rmzelle:
I also renamed the label to "safe to test" as used in the example at https://securitylab.github.com/research/github-actions-preventing-pwn-requests before merging your PR.

retorquere:
> Did the build for #5309 (https://github.com/citation-style-language/styles/runs/2053700809) behave like expected? It seems like it ran the tests without the need for a label.

It's going to run without a label, but it will do the two-checkout (as it did) to use the code from `master` to test the content from the pull request, cloned into `pull-request`. If you use the special label, it will use the code from the pull request to test the content of the pull request.

> I also renamed the label to "safe to test" as used in the example at https://securitylab.github.com/research/github-actions-preventing-pwn-requests before merging your PR.

It doesn't really matter what it's called, but "safe to test" doesn't seem to me to disclose what the behavior is behind the special label. Every PR is supposed to be safe to test.

Ah wait, it is not the case that all PRs are held until they're labeled to get a go-ahead to run the tests. Tests on PRs are going to run as soon as they are submitted, it's just that you can enable the special mode for subsequent tests by applying the label. The act of labelling the PR will invoke the first of these privileged tests. Unlabeled tests will run as always, but in two-checkout mode to split the environment that provides the tests, and the environment that provides the thing-to-test.

rmzelle:
Thanks for the explanations! Maybe we should change the label to something like "vetted_run-with-PR-code" for clarity, then.

retorquere:
Maybe `vetted` suffices?

rmzelle:
I just added the "safe to test" label to https://github.com/citation-style-language/styles/pull/5309 and restarted the GitHub Actions jobs, but it still installed Ruby 2.7.1.

retorquere:
Right -- the test code will be loaded from the PR, but the Ruby is set in the workflow, and since it is the workflow definition itself that lays out how the test environment should be set up, you can't test workflow files themselves in a PR. Conundrum.

The only options I see are:

* Test the workflow in a fork. In the fork you can push the workflow to main/master or a branch (so not a PR on your fork, but merged into your fork), once it works, you submit the PR, or
* We change the workflow (temporarily) on styles `master` to add a build matrix so that both Ruby 2 and 3 are tested, but disable the `3` tests in the workflow by default, so that it only runs with a specific label (can be the same as the privileged label). When it works, we remove the build matrix and leave only ruby 3.

adam3smith:
I'm afraid this has broken some things in the test suite:
1. Build for new styles fail, see e.g. https://github.com/citation-style-language/styles/pull/5314 I tested adding the `save to test` label on https://github.com/citation-style-language/styles/pull/5312 and rerunning and that fixed the build, but obviously that shouldn't be necessary.
2. Updates in styles don't show in the preview. See e.g. https://github.com/citation-style-language/styles/pull/5310 -- which adds DOIs but they don't show.

retorquere:
> I'm afraid this has broken some things in the test suite:
> 
>     1. Build for new styles fail, see e.g. #5314 I tested adding the `save to test` label on #5312 and rerunning and that fixed the build, but obviously that shouldn't be necessary.

Absolutely. I had forgotten to tell Sheldon about the `pull-request` directory -- existing styles would appear to pass because Sheldon would test the style in the base directory, not from the pull-request.

>     2. Updates in styles don't show in the preview. See e.g. #5310 -- which adds DOIs but they don't show.

Same problem. I've fixed Sheldon, and a re-run of those tests will pull those fixes in, and they should now pass.

rmzelle:
> Right -- the test code will be loaded from the PR, but the Ruby is set in the workflow, and since it is the workflow definition itself that lays out how the test environment should be set up, you can't test workflow files themselves in a PR. Conundrum.

Ah, right. I'll just merge the Ruby 3.0.0 PR directly in master, then. Pretty sure it should work (https://github.com/ruby/setup-ruby#supported-versions specifies support for 3.0.0, and the tests ran fine locally with Ruby 3.0.0 with `bundle exec rake`).

So if we ever want to test changes to our GitHub Actions YAML files, it's probably easiest/best to just first test those changes in a fork in "master"?

retorquere:
That's what I did; generating PRs on your own repo is simple and I used that to test. I think it doesn't actually need to be on main/master, I think any branch will do, as long as the workflow is checked in before the tests start on the action.

PR #5308, opened 2021-03-07
Update masarykova-univerzita-pravnicka-fakulta.csl, pravnik.csl, iso690-full-note-cs.csl

OTristanF:
fix hardcoded locator prefix


rmzelle:
@OTristanF, thanks!

PR #4130, opened 2019-06-06
Update harvard-anglia-ruskin-university.csl

POBrien333:
https://forums.zotero.org/discussion/77691/output-style-harvard-anglia-ruskin#latest

PR #5310, opened 2021-03-08

POBrien333:
add DOI

via https://github.com/citation-style-language/styles/pull/4130#issuecomment-791460222

### DIFF
--- a/harvard-anglia-ruskin-university.csl
+++ b/harvard-anglia-ruskin-university.csl
@@ -2,6 +2,7 @@
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
   <info>
     <title>Anglia Ruskin University - Harvard</title>
+    <title-short>ARU</title-short>
     <id>http://www.zotero.org/styles/harvard-anglia-ruskin-university</id>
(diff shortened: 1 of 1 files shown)


rmzelle:
@POBrien333, did you confirm this in the style guide? The quick guide and full guide at https://library.aru.ac.uk/referencing/harvard.htm both seem to be from 2019.

POBrien333:
Yeah, check "Articles with DOI" just show an example of an article with volume, page AND DOI.
https://library.aru.ac.uk/referencing/harvard.htm

rmzelle:
@POBrien333, thanks!

Harbender:
> Yeah, check "Articles with DOI" just show an example of an article with volume, page AND DOI.
> 
> https://library.aru.ac.uk/referencing/harvard.htm

Problem is with mendeley under the type(s) there is only limited selection 
Eg there's no articles with doi choice.. only journal article
Any suggestions how to overcome this issue?
And there is an error -- articles with doi 
They have the [e-journal] after the journal title name 

If possible mayb allow the old csl file available too incase someone needs the old version. Sorry im still new to this github and the codes image

POBrien333:
In Mendeley you can add the DOI in its specific field. Under "Catalog IDs". Then it'll be able to pull it into the output.

PR #5312, opened 2021-03-09
Create journal-for-the-study-of-the-new-testament.csl

carljohanberglund:
CSL Style for Journal for the Study of the New Testament


adam3smith:
(ignore the test failure; your style is fine)

rmzelle:
@carljohanberglund, thanks!

PR #5313, opened 2021-03-09
british-journal-of-anaesthesia.csl: add comma delimiter between non-sequential citations eg. 1 4 7-9 -> 1, 4, 7-9 

JohannesNE:
Add comma between (non-sequential) numbers


adam3smith:
Thanks!

PR #5316, opened 2021-03-10
Create university-of-gothenburg-apa7-swedish-legislations

amanda-pettersson:
APA 7 edited to support references for Swedish legislations.


rmzelle:
@amanda-pettersson, thanks! Are you aware of the existing CSL style https://github.com/citation-style-language/styles/blob/master/university-of-gothenburg-apa-swedish-legislations.csl? It looks like that one is still based on APA 6th edition, so I assume this version can replace that style?

amanda-pettersson:
Yes, I'm aware of the old one (based on APA6). If it's possible to keep both (since some still use the old version) that would be great, if not, this one can replace the old style. I can see how it could be confusing to keep both since they are so alike.

rmzelle:
We can keep both.

One thing we sometimes do is in such cases is transfer the style ID from the APA 6th edition style to this new 7th edition style, so users that have the old style installed are automatically upgraded (if they use software like Zotero that automatically updates styles based on updates made here). People who'd want to keep using the 6th edition would then have to reinstall the style. I don't know if that makes sense here?

We can also just give a new style ID to the 7th edition style, in which case all users would have to manually update to this new edition.

amanda-pettersson:
Okay, perfect! I think the best option would be your second suggestion - to give a new style ID to the 7th edition style.

rmzelle:
@amanda-pettersson, thanks!

PR #5319, opened 2021-03-11
Update yozgat-bozok-universitesi-fen-bilimleri-enstitusu.csl

bizgi:
citation style is updated.


rmzelle:
@bizgi, thanks. Did you base this new version off an existing CSL style? (the style still lists the numeric http://www.zotero.org/styles/elsevier-with-titles" style as the template, but if you redid the style using a different style, we'd like to update that information)

bizgi:
@rmzelle hi. I updated template info.

rmzelle:
@bizgi, thanks!

PR #5331, opened 2021-03-17
update page-range-format for Anglia

icornelius:
set to "expanded"

also update locators-chapter: add space character after volume number


rmzelle:
@icornelius, thanks!

PR #5333, opened 2021-03-19
Update journal-of-the-american-society-of-nephrology.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/378353/#Comment_378353

Guidelines: "List all authors when the number of authors is six or fewer. When the number of authors is greater than six, use et al. for all authors over six."
https://www.asn-online.org/g/blast/files/JASN%20Instructions%20for%20Authors.pdf


adam3smith:
Thanks!

PR #5337, opened 2021-03-21
Create journal-of-business-finance-and-accounting.csl

Nicola-Maria-Fiore:
Journal of Business Finance & Accounting
https://onlinelibrary.wiley.com/journal/14685957
https://onlinelibrary.wiley.com/page/journal/14685957/homepage/forauthors.html

### DIFF
--- a/dependent/journal-of-business-finance-and-accounting.csl
+++ b/dependent/journal-of-business-finance-and-accounting.csl
@@ -1,0 +1,1926 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" page-range-format="expanded" default-locale="en-US">
+  <info>
+    <title>Journal of Business Finance &amp; Accounting</title>
+    <title-short>JBFA</title-short>
+    <id>http://www.zotero.org/styles/journal-of-business-finance-and-accounting</id>
+    <link href="http://www.zotero.org/styles/journal-of-business-finance-and-accounting" rel="self"/>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+    <link href="http://www.zotero.org/styles/apa" rel="template"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
@Nicola-Maria-Fiore  -- could you please allow edits on your three pull requests as described here: https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

That'll make it much easier for us to fix the small issue that break the automated tests.

Nicola-Maria-Fiore:
The first time I pulled without forking the repository. That is why I am not able to check "allow edit for mantainers". Do I need to fork the project first, and then pull the request, right?

adam3smith:
@Nicola-Maria-Fiore github forks automatically when you create a pull request -- are you sure you aren't able to check that box on this page as described above?

Nicola-Maria-Fiore:
@adam3smith I checked several times, I am really sorry. I used that box before, but I cannot see it now 
image

adam3smith:
Maybe double-check in a different browser -- it should really be there in every case. Might be that your theming is interfering with visibility?
What'd need to happen here is
1. rename the file to remove the `dependent/` part
2. Remove all non English locales from the style (which now has a default locale)
3. Run the code through https://formatter.citationstyles.org/ when you're done.

There might be more, but that'd be the initial steps

adam3smith:
But before we proceed -- the journal just says to follow APA 6th edition -- any reason we shouldn't just use APA style here? What did you modify?

Nicola-Maria-Fiore:
@adam3smith - I did not modify anything. It is just like APA style. I wanted to add it to the CSL repository for general availability. As in the following case:
image

Nicola-Maria-Fiore:
@adam3smith As I said in the other commit ("european accounting review") I realized I made a mistake. I think we could delete these two commits ("journal of business finance" and "european accounting review") and create new (and shorter) ones, as long as they are both "dependent" styles. I now realized how to write, looking at here: https://editor.citationstyles.org/styleInfo/?styleId=http%3A%2F%2Fwww.zotero.org%2Fstyles%2Faccounting-forum
Do you agree?

PR #5344, opened 2021-03-23
Please rename to Tijdschrift-voor-geneeskunde-en-gezondheidszorg

sanderdumont:


rmzelle:
@sanderdumont, thanks. We try to be conservative in changing style IDs (and file names, which are currently linked), as the Zotero devs have expressed concerns about breakage.

PR #5348, opened 2021-03-24
Update clinical-and-translational-immunology.csl

POBrien333:
via https://forums.zotero.org/discussion/88597/wrong-style-for-clinical-translational-immunology#latest

Goes with https://github.com/citation-style-language/journals/pull/118


rmzelle:
@POBrien333, thanks!

PR #5350, opened 2021-03-26
Create iran-manual-style

sinasabbagh:
Modified version of Chicago date-author style, edition 16

* Modified quotations and delimiters to match Persian writing style.
* Also changed months names, although I don't think this matters in Chicago style.
* The title of this csl is set to Iran Manual Style, to match the name that is used by Irandoc. Please note that they have an official permission from Chicago Universiry Press to do this.

### DIFF
--- a/iran-manual-of-style.csl
+++ b/iran-manual-of-style.csl
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="display-and-sort" page-range-format="chicago" default-locale="fa-IR">
   <info>
-    <title>Iran Manual of Style - شیوه‌نامه ایران</title>
+    <title>Iran Manual of Style - شیوه‌نامه ایران (فارسی)</title>
     <id>http://www.zotero.org/styles/iran-manual-of-style</id>
     <link href="http://www.zotero.org/styles/iran-manual-of-style" rel="self"/>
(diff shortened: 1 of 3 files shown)


adam3smith:
Thanks -- A couple of questions:
- The name "Iran Manual Style" sounds weird in English. The original in Farsi is  شیوه‌نامه ایران  ?  
How would "Iran Manual of Style - شیوه‌نامه ایران  (فارسی)" work as a title?
And is there just the author-date version of this, or are there footnotes versions as well?

adam3smith:
@ sinasabbagh Please see my questions above -- would be great to get this merged!

sinasabbagh:
I'm terribly sorry for the delay. This was lost in my inbox.
To answer your question:

-  yes the original title is شیوه‌نامه ایران
- Actually you're right about the title "Iran Manual of Style" is correct and is used in English description of the manual.

sinasabbagh:
I've updated the title. I don't see why it needs فارسی in the title. We can add it though, if saying explicitly that it is in Persian is necessary.

adam3smith:
@rmzelle -- this is gtg from my end -- see what you want to do about adding (or not) the language to the style.

rmzelle:
> I've updated the title. I don't see why it needs فارسی in the title. We can add it though, if saying explicitly that it is in Persian is necessary.

We always add the language to the titles of CSL styles that are locked to non-English languages. Two big advantages are that people can then quickly identify styles created for their specific language, and that users are informed what locale a citation style will render in. For consistency we'd like to add it here as well.

rmzelle:
@sinasabbagh, thanks!

PR #5358, opened 2021-03-29
Update journal-of-consumer-research.csl

POBrien333:
via https://forums.zotero.org/discussion/88701/jcr-style#latest

### DIFF
--- a/journal-of-consumer-research.csl
+++ b/journal-of-consumer-research.csl
@@ -135,6 +135,7 @@
   </macro>
   <citation et-al-min="4" et-al-use-first="1" et-al-subsequent-min="3" et-al-subsequent-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year">
     <sort>
+      <key macro="author-short"/>
       <key macro="year-date"/>
     </sort>
     <layout prefix="(" suffix=")" delimiter="; ">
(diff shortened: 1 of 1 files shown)


journal-of-consumer-research.csl
```diff
@@ -132,7 +134,6 @@
   </macro>
   <citation et-al-min="4" et-al-use-first="1" et-al-subsequent-min="3" et-al-subsequent-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year">
     <sort>
-      <key macro="author-short"/>
```
adam3smith, line 135: I don't thinks this is right -- p. 3 of the styleguide 
> Multiple citations are listed alphabetically and separated by semicolons: (Gatignon and Robertson 1985; Green and Gold 1981).•Multiple citations by the same author(s) are separated by a comma: (Moschis and Moore 1978, 1979).

POBrien333: current output: 
(Borges 1999; Hancké, Rhodes and Thatcher 2007; Isaacson 2011, 244–52)
(Accadia et al. 2012b, 2012a)

I´m not sure I'm seeing what you mean.

adam3smith: That's because that example has alphabetical & chronological order co-inciding. Look at the 2nd automated citation below:
(Mares 2001; Fenner et al. 2019) -- that should be the other way around.

POBrien333: Ach. You're in the sort section... Now I get it. :)

adam3smith:
Thanks!

PR #5359, opened 2021-03-29
Update and rename journal-of-neurosurgery.csl to dependent/journal-of… (make AMA dependent)

POBrien333:
…-neurosurgery.csl

via https://forums.zotero.org/discussion/88712/journal-of-neurosurgery-style#latest


POBrien333:
Not sure what the error is. Help welcome. :)

adam3smith:
You're removing a style (or rather, turning it into a dependent style) on which a bunch of other styles either depend or point as template

POBrien333:
Alright. I´ll check those then and see what I can do.

POBrien333:
@adam3smith 
Thank you. This should be fine now. Those 3 journals are just JNS sister journals.

rmzelle:
@POBrien333, thanks!

PR #5360, opened 2021-03-29
Update catholic-biblical-association.csl

dstark:
This update

1. Adds current CBA documentation,
2. Adds support for custom citations specified by CBA and stored in annote,
3. Uses Chicago-style page range truncation,
4. Enforces the use of title case,
5. Allows CBA style to make use of a collection-title-short variable stored in Extra,
6. Corrects an improper output of a work cited with only editors as responsible parties (i.e., from ", ed. [name(s)]" to "[name], ed." or "[names], eds.",
7. Eliminates the spurious uppercasing of "ibid." if it appears not at the start of a sentence,
8. Removes a spurious initial capital from text-based editions (e.g., "rev. ed."),
9. Adds a sub verbo note where it had previously been left out of an initial citation,
10. Corrects improper delimitation and spacing with volume-page citations,
11. Changes the page range scheme to Chicago, and overrides the default e*n dash range delimiter with a hyphen, and
12. Adds a period at the end of a citation.

This update also offers partial support for section-based citations. But I'm having trouble

1. Getting §§ instead of § when there's more than one section being cited and
2. Getting section ranges to truncate properly according to Chicago-style rules.

If there are suggestions particularly about this last item, I'd be most grateful to be pointed to what I'm missing. And this is my first try at proposing style updates directly. So, please pardon and feel free to correct me if there's a better way of doing what I'm trying to achieve here.

Thank you so much!

### DIFF
--- a/catholic-biblical-association.csl
+++ b/catholic-biblical-association.csl
@@ -405,13 +405,14 @@
       <if type="article-journal">
         <text variable="page" prefix=" "/>
       </if>
+	  <!-- Corrects the dictionary entry total page range output when a volume is cited. -->
       <else-if type="chapter entry-dictionary entry-encyclopedia" match="any">
         <choose>
           <if variable="volume">
-            <number variable="volume" form="numeric" suffix="."/>
+            <number variable="volume" form="numeric" suffix=":"/>
           </if>
         </choose>
-        <text variable="page" prefix=" "/>
+        <text variable="page"/>
       </else-if>
     </choose>
   </macro>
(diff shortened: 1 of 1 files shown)


adam3smith:
@dstark -- just to confirm that this is ready for review?

dstark:
Yes, @adam3smith. I think so. Thanks so much for having a look at it.

rmzelle:
> I'm having trouble
> 
>     1. Getting §§ instead of § when there's more than one section being cited and
> 
>     2. Getting section ranges to truncate properly according to Chicago-style rules.
> 
> 
> If there are suggestions particularly about this last item, I'd be most grateful to be pointed to what I'm missing.

@dstark, thanks! For these two issues, could you give some examples of what the desired output is, and what the current output is with your style? Please also mention the software you're using (Zotero?) and its version.

dstark:
@rmzelle, thanks. The output should be § when there's one section cited (e.g., §51). The output should be §§ when there's more than one (e.g., §§51–52). But I seem unable to get the multiple section sign output.

Yes, I'm using Zotero (5.0.97-beta.15+205fe7f03). The doubling of the section sign works fine in the SBL style (https://www.zotero.org/styles/society-of-biblical-literature-fullnote-bibliography, although in that style there's an erroneous space added between the section sign and the section number or range). I've tried to mimic what I've seen in the SBL style in the updates I submitted for CBA to get the single / multiple section signs working, but apparently I'm missing something. :-|

Any thoughts on what that might be would be most appreciated, and please feel free to ask if I can provide anything further.

Thank you so much!

PR #5361, opened 2021-03-29
update doi prefix

Nicola-Maria-Fiore:
<text variable="DOI" prefix="Available at https://doi.org/"/>. 
I am the original creator of the style. I realized that this detail has to be changed


adam3smith:
Thanks!

PR #5365, opened 2021-04-03
Add "H. Online Video (e.g., YouTube)"

ocouch:
https://forums.zotero.org/discussion/58614/ieee-youtube-biblio-output-is-wrong
And a couple of minor formatting corrections (eg. removing full-stops "." from the end of URLs).

### DIFF
--- a/ieee.csl
+++ b/ieee.csl
@@ -36,11 +36,15 @@
     <contributor>
       <name>Brenton M. Wiernik</name>
     </contributor>
+    <contributor>
+      <name>Oliver Couch</name>
+      <email>oliver.couch@gmail.com</email>
+    </contributor>
     <category citation-format="numeric"/>
     <category field="engineering"/>
     <category field="generic-base"/>
     <summary>IEEE style as per the 2018 guidelines, V 11.12.2018.</summary>
-    <updated>2019-12-20T09:14:19+00:00</updated>
+    <updated>2020-06-15T03:21:46+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
@@ -83,7 +87,7 @@
           <date-part name="year" form="long"/>
         </date>
       </if>
-      <else-if type="bill book chapter graphic legal_case legislation motion_picture song thesis" match="any">
+      <else-if type="bill book chapter graphic legal_case legislation song thesis" match="any">
         <date variable="issued">
           <date-part name="year" form="long"/>
         </date>
@@ -92,6 +96,13 @@
         <date variable="issued">
           <date-part name="month" form="short"/>
           <date-part name="year" prefix=" "/>
+        </date>
+      </else-if>
+      <else-if type="motion_picture" match="any">
+        <date variable="issued" prefix="(" suffix=")">
+          <date-part name="month" form="short" suffix=". "/>
+          <date-part name="day" form="numeric-leading-zeros" suffix=", "/>
+          <date-part name="year"/>
         </date>
       </else-if>
       <else>
@@ -201,7 +212,7 @@
               <group delimiter=" " prefix="(" suffix=")">
                 <text term="accessed"/>
                 <date variable="accessed">
-                  <date-part name="month" form="short" strip-periods="false"/>
+                  <date-part name="month" form="short" strip-periods="false" suffix="."/>
                   <date-part name="day" form="numeric-leading-zeros" prefix=" " suffix=", "/>
                   <date-part name="year" form="long"/>
                 </date>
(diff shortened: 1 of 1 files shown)


ieee.csl
```diff
@@ -276,7 +306,7 @@
   </citation>
   <!-- Bibliography -->
   <bibliography entry-spacing="0" second-field-align="flush">
-    <layout suffix=".">
```
ocouch, line 279: This needs to be removed and reapplied case by case as URLs should not end in a full-stop.

ieee.csl
```diff
@@ -353,6 +392,11 @@
             <text macro="publisher"/>
             <text macro="issued"/>
             <text macro="page"/>
+            <choose>
```
ocouch, line 395: Include the URL in the bibliography for items sourced online.

ieee.csl
```diff
@@ -262,6 +280,18 @@
       <text variable="locator"/>
     </group>
   </macro>
+  <macro name="location">
```
ocouch, line 283: Location is included in the macros for conference presentations, but the representation is different for online videos. It's probably worth moving the location information from those other macros, and making this a dedicated location macro that handles all cases.

ieee.csl
```diff
@@ -214,20 +225,27 @@
...
             </date>
           </group>
-          <text term="online" prefix="[" suffix="]" text-case="capitalize-first"/>
-          <group delimiter=": ">
-            <text term="available at" text-case="capitalize-first"/>
-            <text variable="URL"/>
-          </group>
+          <choose>
+            <if type="motion_picture">
+              <text value="Online Video" prefix="[" suffix="]"/>
```
ocouch, line 239: This is as per the IEEE guide, but it's not going to be translatable. I'm not familiar with setting up terms for localization yet, so someone clue-ier than I could fix this up to make it localizable.

ocouch:
This PR is ready to go if someone wants to review it? @adam3smith @POBrien333 @rmzelle

ieee.csl
```diff
@@ -201,7 +212,7 @@
               <group delimiter=" " prefix="(" suffix=")">
                 <text term="accessed"/>
                 <date variable="accessed">
-                  <date-part name="month" form="short" strip-periods="false"/>
+                  <date-part name="month" form="short" strip-periods="false" suffix="."/>
```
adam3smith, line 215: why are you adding the period suffix here? The month abbreviation contains a period.

ieee.csl
```diff
@@ -214,20 +225,27 @@
         <text variable="DOI" prefix="doi: "/>
       </else-if>
       <else>
-        <group delimiter=". ">
+        <group delimiter=". " suffix=". ">
           <group delimiter=": ">
             <text term="accessed" text-case="capitalize-first"/>
             <date variable="accessed">
-              <date-part name="month" form="short" suffix=" "/>
+              <date-part name="month" form="short" suffix=". "/>
```
adam3smith, line 232: ditto here on the suffix.

ieee.csl
```diff
@@ -214,20 +225,27 @@
         <text variable="DOI" prefix="doi: "/>
       </else-if>
       <else>
```
adam3smith, line 227: given how we're using this now, you'll want to turn this into `<else-if variable="URL">`

ieee.csl
```diff
@@ -353,6 +392,11 @@
             <text macro="publisher"/>
             <text macro="issued"/>
             <text macro="page"/>
+            <choose>
+              <if variable="URL">
```
adam3smith, line 396: No need to test for URL here (and if testing should be for URL or DOI) since we do this in the macro.

ieee.csl
```diff
@@ -342,6 +372,15 @@
             <text variable="number"/>
             <text macro="issued"/>
           </group>
+        </else-if>
+		<!-- Online Video -->
+        <else-if type="motion_picture">
+          <text macro="location" suffix=". "/>
```
adam3smith, line 378: I don't quite understand this -- what do you envision "location" is in the example citations? I'm reading the sample ctitations as author (date) "title" [medium], Available...

adam3smith (changes requested):
Thanks! Some comments, some questions.

ieee.csl
ocouch: I'll do some more testing/investigation and get back to you. It wasn't included (hence why I added it). May be a localization thing?

ieee.csl
ocouch: It's a physical location:

Video Owner/Creator, Location (if available). Title of Video: In Initial Caps. (Release date). Accessed: 
Month Day, Year. [Online Video]. Available: http://URL.onlinevideo.org

ocouch:
Thanks for taking a look :)

I've got some testing to do and will get back to you.

ieee.csl
ocouch: Ok, I think this may have been because I was testing with citation that was released in May...

It does seem to work fine (although I did update Zotero recently, so it's possible there was an issue but it's been resolved. Either way I'll delete this.

ieee.csl
ocouch: Done.

ieee.csl
ocouch: Changed to geographic-location

ocouch:
@adam3smith Changes made :)

adam3smith:
Sorry for the delay here -- unfortunately this isn't quite right with respect to punctuation yet. Standard items should all end with a period, and that's missing. Most items, except journal articles, should have a period before the access macro, not a comma. I can try to work on this because it's a lot of fiddling, but also very happy for you to continue.

ocouch:
No worries, and thanks for being thorough! I'm not quite sure how the suffix prioritisation works, so I'll do some experimentation and update it - might be able to have `suffix="."` set for the whole bibliography, but have URL's override it?

I also noticed the IEEE guide uses comma inside the quotations (punctuation-in-quotes) - but I'll leave it as is for now. It can go in a separate pull request after figuring out how best to handle that one (I'm not too clear how localization works atm, I'll have to do some reading).

ocouch:
Ok, I think I've got it working, basically:
```
<group delimiter=", " suffix=". " >
...Most of the reference
</group>
<text macro="access">
```
This way they _always_ have a full stop, and then the appropriate suffix for Doi or URL is selected in the access macro and it gets appended at the end, and as the access macro is setup at the moment it'll only show the URL if there's no DOI available.

Journal Articles (Periodicals in IEEE terms) have `<group delimiter=", " suffix=", " >` instead.

ocouch:
I just realised if there's no DOI or URL, that leaves the Journal with a comma at the end instead of a full stop, but CSL can't handle nested \<if> statements, to account for this so I'm not sure how else to handle trying to reproduce the guide exactly.

Tbh, it's probably better to override the guide for the basic format and be consistent with the other types (and other Periodical examples in the guide)

image
image
image
image

ocouch:
@adam3smith Ready when you are :)

PR #5367, opened 2021-04-05
Adicionar ponto e espaço antes de "Disponível em:"

samuel-rosa:
Verifiquei a necessidade dessa correção no uso de trabalho publicado em evento.


rmzelle:
@samuel-rosa, thanks. Could you try the new version (you can download it from https://raw.githubusercontent.com/citation-style-language/styles/7943bfa1d2009dfb987fd869ee0f9de401eb5d1a/associacao-brasileira-de-normas-tecnicas.csl)? Your original edit seems to make create some side-effects elsewhere in the style.

adam3smith:
I'll take this as is -- there may be more to do, but this is definitely an improvement.

PR #5368, opened 2021-04-08
add medium field for eBooks to MLA 8th

POBrien333:
via https://forums.zotero.org/discussion/70629/entry-type-for-ebook

Not sure if we want this, but this is a suggestion.


POBrien333:
@adam3smith 
Could we merge this one?
As far as the OP goes they said it worked fine.

bwiernik:
Thanks!

PR #5369, opened 2021-04-08
Update triangle.csl

TriangleUMR5206:
Publication date added for type : chapter (note section).


rmzelle:
@TriangleUMR5206, thanks!

PR #5374, opened 2021-04-10
Create istanbul-medical-journal.csl

karamelikli:
Please add this file to the list.


rmzelle:
Thanks for your submission. Is there an online style guide available for this style? Per https://github.com/citation-style-language/styles/blob/master/STYLE_REQUIREMENTS.md#10---documentation-link, we typically require each CSL style to include a link to the format it describes.

And did you base this on an existing CSL style? If so, we'd like to know which one, so we can add a "template" link (see https://github.com/citation-style-language/styles/blob/master/STYLE_REQUIREMENTS.md#8---template-link)

rmzelle:
@karamelikli, thanks!

karamelikli:
Thanks Rintze

PR #5375, opened 2021-04-10
fix et-al seetings for ageing-and-society.csl

POBrien333:
https://forums.zotero.org/discussion/88981/style-error-ageing-and-society


adam3smith:
Thanks!

PR #5378, opened 2021-04-13
[WIP] Update style-manual-australian-government.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/380515/#Comment_380515

Bring style in line with new guidelines.


adam3smith:
I think we might want to keep the old version rather than replace it. I'd suspect given how prominent it was, people will be attached to it for some time e.g. the Tasmania style seemed to describe the old version of the style manual.

POBrien333:
Makes sense. How should I call these two versions?

PR #5381, opened 2021-04-13

dstark:
Two changes:
1. Adds support for custom citations via the annote variable.
2. Corrects a spurious space after the §(§) sign(s) when citing by section.


rmzelle:
@dstark, thanks!

PR #5384, opened 2021-04-14
Update university-of-york-harvard-environment.csl

AVWoman:
Removing erroneous delimiter from author-short


rmzelle:
@AVWoman, thanks!

PR #5387, opened 2021-04-15
Update revue-forestiere-francaise.csl

timotheefouqueray:
Changed the quote instruction from true to false in the bibliography section, so that the name of the report/speech/chapter/etc. is not framed by quote marks


rmzelle:
@timotheefouqueray, thanks!

PR #5391, opened 2021-04-15
Updating infoclio.ch styles

nchachereau:
This is a small update to a custom style for history students in Switzerland. See #234 for the original pull request, and #3898 for the last update.

Changes:

- `publisher` field is only displayed for specific types, rather than almost any type (this change was necessary because many translators have started including the "publisher" variable in the Extra field for journal articles).
- New CSL file for a common style for German-speaking historians, where footnotes only include a short version of the reference. Except for this change, the file is the same as `infoclio-de.csl`.

I know I tend to use the same text in my pull request, but that does not mean I'm not sincere when I write: Thank you again for taking the time to maintain the styles!


rmzelle:
> I know I tend to use the same text in my pull request, but that does not mean I'm not sincere when I write: Thank you again for taking the time to maintain the styles!

You're welcome, and we love the detailed pull request descriptions!

rmzelle:
@nchachereau, thanks for contributing to the repository!

PR #5392, opened 2021-04-15
Update ocean-and-coastal-research.csl

azguth:
Fixed some issues with container title suffix and added isssue after the volume. Also fixed the format of Editor names from short to long.


rmzelle:
@azguth, thanks!

PR #5393, opened 2021-04-16
Update south-african-medical-journal.csl

POBrien333:
via https://forums.zotero.org/discussion/89106/updated-citation-style#latest


rmzelle:
@POBrien333, thanks!

PR #5397, opened 2021-04-19
Update journal-of-nutrition.csl

POBrien333:
via https://forums.zotero.org/discussion/89158/style-error-am-j-clin-nutr#latest


rmzelle:
@POBrien333, thanks!

PR #5400, opened 2021-04-20
Create the-bovine-practitioner.csl

POBrien333:
via https://forums.zotero.org/discussion/89179/style-request-bovine-practitioner#latest


rmzelle:
@POBrien333, thanks! 🐮

PR #5404, opened 2021-04-21
Update mammalia.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/381247/#Comment_381247


rmzelle:
@POBrien333, I rather not link to Google Docs, but I see that the link to author instructions at https://www.degruyter.com/journal/key/MAMM/html#callForPapersHeader appears to be broken. I sent the journal an email asking for this to be fixed.

adam3smith:
Thanks! (Link to docs was working when I tested, might have been added or fixed)

PR #5407, opened 2021-04-22
Adds missing space after title and moves space from title prefix to author suffix.

JohannesNE:
First commit (2a88753) fixes an error.
The second (9e7784c) should not change the output.


rmzelle:
@JohannesNE, thanks. I made a small edit to use a group with a delimiter, which should generate the same output but is a bit cleaner and more robust.

PR #5419, opened 2021-04-27
Create wirtschaftsuniversitat-wien-institut-fur-transportwirtschaft-u…

georgd:
…nd-logistik.csl


adam3smith:
@georgd -- I removed the duplicate punctuation. That might not have shown in Zotero, but better to not rely on the citeproc to suppress things like that. Let me know if you have any concerns about this, otherwise this looks good to merge.

georgd:
@adam3smith thanks for spotting this. The citeproc-test-runner doesn’t output the double colons so this is what made me overlook it. But please, don’t merge yet, I’ve found some places to simplify which I’ll do and there’s the sort in a single macro issue as well.

georgd:
@adam3smith from my side it’s good to merge now.

adam3smith:
Thanks!

PR #5421, opened 2021-04-27
Create society-of-biblical-literature-1st-edition-fullnote-bibliography

dstark:
Some journals in biblical studies require that the 1st edition of the SBL Handbook of Style be used, or they require a style very similar to this. This file is just a download from https://raw.githubusercontent.com/citation-style-language/styles/929692740c06108efba284eab38e565bee87754f/society-of-biblical-literature-fullnote-bibliography.csl with then some minor modifications that should, I think, disambiguate it from the styles for the 2nd edition of the SBL Handbook of Style. See also this related Zotero forum thread: https://forums.zotero.org/discussion/49879/sbl-1st-ed#latest


adam3smith:
Sorry for the long wait -- this makes complete sense to include, probably should have kept it all along.

dstark:
No worries. Thanks so much!

PR #5422, opened 2021-04-28
small fixes for jopr.csl

POBrien333:
https://forums.zotero.org/discussion/comment/381770/#Comment_381770


rmzelle:
@POBrien333, thanks!

PR #5425, opened 2021-04-29
Create studies-in-the-history-of-gardens-and-designed-landscapes.csl

POBrien333:
via https://forums.zotero.org/discussion/88971/citation-style-required-by-the-journal-studies-in-the-history-of-gardens-designed-landscapes#latest


rmzelle:
@POBrien333, thanks!

PR #5426, opened 2021-04-29
Create topoi-orient-occident-auteur-date.csl

NMonteix:
First of 2 styles for the same journal

### DIFF
--- a/topoi-orient-occident-auteur-date.csl
+++ b/topoi-orient-occident-auteur-date.csl
@@ -75,16 +75,6 @@
       </substitute>
     </names>
   </macro>
-  <macro name="access">
-    <group>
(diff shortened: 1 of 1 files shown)


NMonteix:
The two problems pointed by Travis (a name problem in this line ` <link href="http://www.zotero.org/styles/topoi-orient-occident-auteur-date" rel="self"/>` and an uncalled macro have not been seen by the validator

adam3smith:
> The two problems pointed by Travis (a name problem in this line <link href="http://www.zotero.org/styles/topoi-orient-occident-auteur-date" rel="self"/> and an uncalled macro) have not been seen by the validator

That's on purpose -- these aren't technically validation problems (the CSL styles will work fine) but repository rules.

PR #5427, opened 2021-04-29
Create topoi-orient-occident-classique.csl

NMonteix:
This is a second style for the same journal (ISSN: 1161-9473), see PR #5426. I know there is a possibility to introduce an exception in ISSN check, but I don't remember where is the file to change (...)

### DIFF
--- a/topoi-orient-occident-classique.csl
+++ b/topoi-orient-occident-classique.csl
@@ -1,10 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" class="note" demote-non-dropping-particle="sort-only" page-range-format="expanded" default-locale="fr-FR">
   <info>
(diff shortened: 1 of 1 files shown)


NMonteix:
This is a second style for the same journal (ISSN: 1161-9473), see PR #5426. I know there is a possibility to introduce an exception in ISSN check, but I don't remember where is the file to change (...)

rmzelle:
> This is a second style for the same journal (ISSN: 1161-9473), see PR #5426. I know there is a possibility to introduce an exception in ISSN check, but I don't remember where is the file to change (...)

It's https://github.com/citation-style-language/styles/blob/master/spec/filters.yaml, but it looks like our test for duplicated ISSNs is broken at the moment. Don't worry about it.

NMonteix:
Ok

Great (not that the test is broken...)

Thanks

Le 10/05/2021 à 04:16, Rintze M. Zelle a écrit :

PR #4444, opened 2019-12-10
Update john-benjamins-publishing-company-collective-volumes.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/344832/#Comment_344832
Haven't fixed the "van" issue.

PR #5430, opened 2021-05-02
Create apa6-for-international-journal-of-learner-corpus-research.csl

elenlefoll:
The first version of a CSL stylesheet for the International Journal of Learner Corpus Research (John Benjamins) based on APA6.

### DIFF
--- a/international-journal-of-learner-corpus-research.csl
+++ b/international-journal-of-learner-corpus-research.csl
@@ -1,20 +1,20 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" page-range-format="expanded" default-locale="en-US">
   <info>
-    <title>John Benjamins Publishing Company - International Journal of Learner Corpus Research</title>
+    <title>International Journal of Learner Corpus Research</title>
     <title-short>IJLCR</title-short>
-    <id>http://www.zotero.org/styles/john-benjamins-publishing-company-international-journal-of-learner-corpus-research</id>
-    <link href="http://www.zotero.org/styles/john-benjamins-publishing-company-international-journal-of-learner-corpus-research" rel="self"/>
+    <id>http://www.zotero.org/styles/international-journal-of-learner-corpus-research</id>
+    <link href="http://www.zotero.org/styles/international-journal-of-learner-corpus-research" rel="self"/>
     <link href="http://www.zotero.org/styles/apa-6th-edition" rel="template"/>
     <link href="https://benjamins.com/series/ijlcr/stylesheet" rel="documentation"/>
-    <issn>2215-1478</issn>
-    <eissn>2215-1486</eissn>
     <author>
       <name>Elen Le Foll</name>
       <email>elefoll@uos.de</email>
     </author>
     <category citation-format="author-date"/>
     <category field="linguistics"/>
+    <issn>2215-1478</issn>
+    <eissn>2215-1486</eissn>
     <updated>2021-05-02T02:45:06+00:00</updated>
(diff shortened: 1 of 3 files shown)


elenlefoll:
Ready for inclusion in stylesheet repository.

adam3smith:
@rmzelle thoughts on the naming here? I've followed the existing 2 John Benjamins styles we have (https://www.zotero.org/styles?q=id%3Ajohn-benjamins-publishing-company-iconicity-in-language-and-literature and  https://www.zotero.org/styles?q=id%3Ajohn-benjamins-publishing-company-linguistik-aktuell-linguistics-today), but in retrospect, I'm not quite sure why we're putting the publisher in the title?

rmzelle:
> I'm not quite sure why we're putting the publisher in the title?

Searching back, I think the publisher had multiple style variants that were each used across multiple journals. See
https://github.com/citation-style-language/styles/pull/3772#issuecomment-437745660, https://forums.zotero.org/discussion/comment/344885/#Comment_344885, and https://github.com/citation-style-language/styles/pull/4444#issuecomment-568226149.

In this case, if we have no indication that this style is used beyond this specific journal, I wouldn't include the publisher name.

international-journal-of-learner-corpus-research.csl
```diff
@@ -0,0 +1,1514 @@
...
+    <group>
+      <choose>
+        <if locator="page" match="none">
+          <label variable="locator" form="short" suffix=" "/>
+        </if>
+      </choose>
+      <text variable="locator"/>
+    </group>
+  </macro>
+  <!-- The IJLCR stylesheet actually states: "When citing or quoting a publication with three or more authors for the first time in the text, please list all authors’ names. For every consecutive mention of this reference, please give only the first author plus “et al." As far as I know, this cannot be achieved in Zotero. Et al. is therefore always printed when there are 3 or more authors -->
```
rmzelle, line 1469: > 

@elenlefoll, this can be achieved in CSL through use of a `<choose/>` element that test for "position" of the in-text citation, e.g. with `position=”first”` or `position=”subsequent”`. See https://docs.citationstyles.org/en/1.0.1/specification.html#choose. Do you feel comfortable enough making this edit yourself? I didn't check the rest of the XML comments to see if you documented any other issues you ran into. Please let us know if there are.

adam3smith: No choose necessary. These are standard et -al-subsequent settings, already used in APA 6, no?

elenlefoll: Great to hear that this is possible @rmzelle! 

@adam3smith, I've just tested it again and the stylesheet currently does not do this automatically even though I did not make any changes to this section of the original APA 6 template.

How do you recommend I proceed?

international-journal-of-learner-corpus-research.csl
```diff
@@ -0,0 +1,1514 @@
...
+  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true">
```
adam3smith, line 1470:
```suggestion
  <citation et-al-min="6" et-al-use-first="1" et-al-subsequent-min="3" et-al-subsequent-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true" collapse="year" givenname-disambiguation-rule="primary-name">

```

adam3smith: This is taken straight from APA 6th -- not sure why that's not in your version of the style? You definitely want the et-al parts, see if you also want the additional attributes at the end.

elenlefoll: I've just tested for suggestion and it looks great, thank you!

elenlefoll:
As far as I can tell, all issues have now been resolved.

adam3smith:
Great, thanks!

PR #5431, opened 2021-05-03
fix punctuation issue in springer-basic-author-date-no-et-al.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/382021/#Comment_382021


adam3smith:
Thanks!

PR #5436, opened 2021-05-06

dstark:
When a custom citation is triggered by the annote variable being used, this change inserts a comma and space (rather than space only) as a delimiter before the locator. This helps prevent the need to store the comma in annote and improves interoperability with similar styles (e.g., Catholic Biblical Association) that require an identical abbreviated reference but a different delimiter (space only).

### DIFF
--- a/society-of-biblical-literature-fullnote-bibliography.csl
+++ b/society-of-biblical-literature-fullnote-bibliography.csl
@@ -32,7 +32,7 @@
     <category citation-format="note"/>
     <category field="theology"/>
     <summary>Society of Biblical Literature format with full notes and bibliography</summary>
-    <updated>2021-04-13T15:55:26+00:00</updated>
+    <updated>2021-05-06T18:37:26+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en-US">
@@ -684,8 +684,8 @@
                   </if>
                   <else>
                     <group delimiter=" ">
-                      <label variable="locator" form="short"/>
-                      <text variable="locator"/>
+                        <label variable="locator" form="short"/>
+                        <text variable="locator"/>
                     </group>
                   </else>
                 </choose>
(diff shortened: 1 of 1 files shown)


society-of-biblical-literature-fullnote-bibliography.csl
```diff
@@ -938,10 +938,20 @@
       <choose>
         <!-- Support custom citations via the annote variable. For discussion, see https://forums.zotero.org/discussion/comment/317370 -->
         <if variable="annote">
-          <group delimiter=" ">
-            <text variable="annote"/>
-            <text macro="point-locators-subsequent"/>
-          </group>
+          <choose>
+            <if locator="sub-verbo">
```
adam3smith, line 942: This looks good in principle, but from your description I don't understand why this is limited to "sub-verbo"? Shouldn't it be for all locators?

dstark: Fair point. [SBL style requires no comma](https://sblhs2.com/2017/04/20/citing-reference-works-7-greek-language-tools/) with section-based citations (thus, e.g., "BDF §123"). Sub verbo is probably the most common contrary case where the comma _should_ appear.

But it's probably simpler test for whether the locator is a section and then omit the comma, including it everywhere else. There might be some other places where it should be omitted, but it looks at least like [page-based citations with custom abbreviations should have the comma too](https://sblhs2.com/2017/03/30/citing-reference-works-2-lexica/) (e.g., "BDB, 562a."). So, having the comma is probably the better catch-all. I'll update the request presently.

dstark: Though, there's a wrinkle in this approach: Out of the sources that SBL style specifies abbreviation-based citations for, some also use page number as a locator. But among these, some include a comma and space delimiter (as in the example above). Others include only a space (e.g., [PL](https://sblhs2.com/2017/01/31/mignes-patrologia-latina/) and [PG](https://sblhs2.com/2017/05/02/pg-citations/)).

The only way I'm thinking of for the software to distinguish when to use one or the other delimiter is by testing the value of the `annote` variable. But am I right that that's [neither currently possible nor likely to happen in the foreseeable future](https://github.com/citation-style-language/csl-evolution/issues/26)?

If so, would you see any other options for disambiguating between when to give the comma+space delimeter and when the space only version?

adam3smith:
> The only way I'm thinking of for the software to distinguish when to use one or the other delimiter is by testing the value of the annote variable. But am I right that that's neither currently possible nor likely to happen in the foreseeable future?

That's right. I think if the citation style prescribes inconsistent rules for items of the same type & structure, we're out of luck. The only option then would be to introduce a complex system along the lines of what juris-m does for different jurisdictions, and that really feels like overkill here.

Issue #5438, opened 2021-05-10
Duplicate ISSNs aren't being reported


rmzelle:
@retorquere, it looks like our test for duplicate ISSNs isn't working.

E.g. 1161-9473 is currently duplicated (https://github.com/citation-style-language/styles/search?q=1161-9473), but the commit that introduced the duplicated merged without complaint (https://github.com/citation-style-language/styles/actions/runs/826493406 from https://github.com/citation-style-language/styles/pull/5427). Could you take a look?

See also:
https://github.com/citation-style-language/styles/blob/931c057e5d1554ad10c7ab7acaa6bf55a9098d82/spec/repository_spec.rb#L17
https://github.com/citation-style-language/styles/blob/931c057e5d1554ad10c7ab7acaa6bf55a9098d82/spec/filters.yaml

(probably related to https://github.com/citation-style-language/styles/pull/5268)

retorquere:
I don't yet know what causes it, but it seems unrelated to splitting off `filters.yaml`; if I recreate the PR environment:

* clone master into `~/github/styles`
* clone master again in `~/github/styles/pull-request`
* remove `topoi-orient-occident-classique.csl` from `~/github/styles`
* run `rake` in `~/github/styles`

I do get the "The CSL Style Repository may not contain any (non-excepted) duplicate ISSNs" error.

retorquere:
I've also removed `topoi-orient-occident-classique.csl` on my fork, and [added it back in in a PR](https://github.com/retorquere/styles/runs/2543109435?check_suite_focus=true) and the test correctly fails there. Still no idea how the CSL repo could be different than that.

retorquere:
Ah OK. So, what do we do here?

rmzelle:
Would it be possible to run things like the ISSN duplicate check in PRs in a post-merge setting, so that outdated PR branches don't hide errors like this one?

rmzelle:
> it can be ran as part of the merge test

Ideally it would be tested automatically on every PR commit.

retorquere:
The `sheldon` workflow (which also runs the spec tests) runs when a PR is opened, synchronized (= more commits pushed), or when its labels change.

rmzelle:
[...]

PR #5440, opened 2021-05-10
Create bibliothecae-it.csl

POBrien333:
via https://forums.zotero.org/discussion/87588/style-request-bibliothecae-it#latest

### DIFF
--- a/bibliothecae-it.csl
+++ b/bibliothecae-it.csl
@@ -10,24 +10,11 @@
       <name>Patrick O'Brien</name>
     </author>
     <category citation-format="author-date"/>
-    <category field="generic-base"/>
+    <category field="humanities"/>
     <issn>2283-9364</issn>
     <updated>2021-05-10T14:30:04+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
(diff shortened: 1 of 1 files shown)


retorquere:
@rmzelle we should fix the problem that initiated #5438 to curb false negatives while we figure out how to address #5438.

rmzelle:
> @rmzelle we should fix the problem that initiated #5438 to curb false negatives while we figure out how to address #5438.

Done via https://github.com/citation-style-language/styles/pull/5444. Sorry about that.

bibliothecae-it.csl
```diff
@@ -0,0 +1,246 @@
...
+        <single>p.</single>
+        <multiple>pp.</multiple>
+      </term>
+      <term name="and others"/>
+      <term name="letter" form="short">lettera</term>
+    </terms>
+  </locale>
+  <macro name="responsibility">
+    <names variable="author">
+      <name delimiter-precedes-last="never" et-al-min="4" et-al-use-first="4"/>
```
adam3smith, line 30: et-al-min needs to be > et-al-use-first for logical reasons. Do you want 4/3 here or 4/1 as below?

POBrien333: thanks for spotting that error!

rmzelle:
@POBrien333, thanks!

PR #5441, opened 2021-05-10
Create grant_proposal_style.csl 

MatthiasSiewert:
Grant proposal style (numeric brackets; lead author et al., no title, journal abbr, volume, pages (year))
This was created to generate compact grant proposal styles with numeric e.g. [1], [1-3] in-text references and short, typically one-lined bibliography.
I created this style for myself, as I found no corresponding style.


adam3smith:
Thanks! I don't think we'll take this for the repository -- preferences and requirements for styles in grant proposals vary quite a bit (e.g., no et al. in NSF proposals in the US) so I think this would be confusing unless it follows a particular funder's requirements.

PR #5443, opened 2021-05-11

OTristanF:
1. `ANON.` changed to `Anon.`
2. Added publisher info to webpage and blog post


adam3smith:
Ignore the automated test failure, that's not about this pull request.
@libora -- I think you've been heavily involved with these styles, is that right? Is the above change correct?

adam3smith:
Thanks!

PR #5458, opened 2021-05-17
Create apa6-for-register-studies.csl

elenlefoll:
First version of a citation stylesheet for Register Studies (John Benjamins)

### DIFF
--- a/register-studies.csl
+++ b/register-studies.csl
@@ -1,20 +1,20 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" page-range-format="expanded">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" page-range-format="expanded" default-locale="en-GB">
   <info>
-    <title>Register Studies (Benjamins) based on APA6</title>
-    <title-short>APA6 for RS</title-short>
-    <id>http://www.zotero.org/styles/apa6-for-register-studies</id>
-    <link href="http://www.zotero.org/styles/apa6-for-register-studies" rel="self"/>
+    <title>Register Studies</title>
+    <title-short>RS</title-short>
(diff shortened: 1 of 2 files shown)


rmzelle:
@elenlefoll, thanks. I see that https://benjamins.com/catalog/rs says "The full reference list should follow guidelines provided by the American Psychological Association (6th edition). A few examples follow; please consult the APA manual for full details."

But it looks like you modified the APA 6th edition style format. Are you sure the journal uses its own custom format, and doesn't accept plain APA 6th edition style references?

elenlefoll:
Hi @rmzelle, yes, that's right, the journal follows APA6 for the bibliography **but** the in-text citations are in a different style. All this modified stylesheet does different to the APA6 template I used is the formatting of the in-text citations.

register-studies.csl
```diff
@@ -0,0 +1,1512 @@
...
+    <group>
+      <choose>
+        <if locator="page" match="none">
+          <label variable="locator" form="short" suffix=" "/>
+        </if>
+      </choose>
+      <text variable="locator"/>
+    </group>
+  </macro>
+  <citation et-al-min="3" et-al-use-first="2" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true">
```
adam3smith, line 1469: The `et-al-use-first="2"` here seems wrong to me. The RS styleguide has
> (Johnson et al. 2014: 43)

Your version produces

>(Johnson, Meyer, et al. 2014: 43)

adam3smith:
@elenlefoll -- I have one remaining question above, everything else looks good to me. Thanks!

register-studies.csl
elenlefoll: Thanks for picking this up. You're absolutely right, it should be: et-al-use-first="1".

rmzelle: I removed the comma in front of "et al." as well.

rmzelle:
@elenlefoll, thanks!

register-studies.csl
elenlefoll: Thank you!

PR #5463, opened 2021-05-20
Update gallia-prehistoire.csl

tomax31:
I forgot to change a quotes to "false"… l.118

### DIFF
--- a/gallia-prehistoire.csl
+++ b/gallia-prehistoire.csl
@@ -1,17 +1,16 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" delimiter-precedes-et-al="never" initialize-with="." page-range-format="expanded" demote-non-dropping-particle="sort-only" default-locale="fr-FR">
-  <!-- This style was edited with the Visual CSL Editor (https://editor.citationstyles.org/visualEditor/) -->
   <info>
-    <title>Gallia Préhistoire (Français)</title>
+    <title>Gallia Préhistoire (French)</title>
     <title-short>GP</title-short>
     <id>http://www.zotero.org/styles/gallia-prehistoire</id>
     <link href="http://www.zotero.org/styles/gallia-prehistoire" rel="self"/>
     <link href="http://www.persee.fr/collection/galip" rel="documentation"/>
-    <link href="https://journals.openedition.org/galliap/430" rel="documentation"/>
-    <link href="http://www.mae.parisnanterre.fr/gallia-prehistoire/" rel="documentation"/>
+    <link href="https://journals.openedition.org/galliap" rel="documentation"/>
+    <link href="https://www.mshmondes.cnrs.fr/gallia-prehistoire" rel="documentation"/>
     <author>
       <name>Thomas Perrin</name>
-      <email>tperrin@univ-tlse2.fr</email>
+      <email>thomas.perrin@cnrs.fr</email>
     </author>
     <author>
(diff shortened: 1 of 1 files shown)


gallia-prehistoire.csl
```diff
@@ -1,17 +1,16 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" delimiter-precedes-et-al="never" initialize-with="." page-range-format="expanded" demote-non-dropping-particle="sort-only" default-locale="fr-FR">
-  <!-- This style was edited with the Visual CSL Editor (https://editor.citationstyles.org/visualEditor/) -->
   <info>
-    <title>Gallia Préhistoire (Français)</title>
+    <title>Gallia Préhistoire (French)</title>
```
adam3smith, line 4: please revert this -- we're now using all language labels in the actual language

gallia-prehistoire.csl
```diff
@@ -62,8 +61,8 @@
   </macro>
   <macro name="editor">
     <names variable="editor">
-      <name initialize-with="." name-as-sort-order="all"/>
-      <label form="short"/>
+      <name et-al-min="3" initialize-with="." name-as-sort-order="all" sort-separator=" "/>
```
adam3smith, line 64: The `et-al-min` isn't necessary here; it's inherited from the citation/bibliography. Please remove.

gallia-prehistoire.csl
```diff
@@ -210,7 +200,7 @@
             </group>
           </if>
           <else>
-            <text variable="URL" text-decoration="underline"/>
+            <text variable="DOI"/>
```
adam3smith, line 203: no prefix like "https://doi.org/" or so?

gallia-prehistoire.csl
```diff
@@ -336,7 +321,7 @@
       <key variable="title"/>
     </sort>
     <layout suffix=".">
-      <group delimiter=" &#8211; ">
+      <group delimiter=" – ">
```
adam3smith, line 324: please revert - we use escaped characters where it's not immediately visible what character is used.

gallia-prehistoire.csl
```diff
@@ -241,14 +231,9 @@
...
       </if>
-      <else-if type="chapter entry-dictionary entry-encyclopedia" match="any">
-        <group delimiter=", ">
-          <text variable="publisher-place" font-style="italic"/>
-        </group>
-      </else-if>
       <else-if type="manuscript thesis report" match="any">
         <group delimiter=", ">
-          <text variable="publisher-place"/>
+          <text variable="locator"/>
```
adam3smith, line 236: this seems odd - why put the locator here? The locator is the pinpoint assigned to a given citation -- included here, it will never actually be used (this is also the case for the locator in the `vol-pages` macro. Could you say what the goal here is?

adam3smith (changes requested):
Thanks -- some change requests, some questions above. Please update your pull request by modifying the style here (see instructions above). Do _not_ start a new pull request.

gallia-prehistoire.csl
tomax31: yes you're right, this is a mistake and a confusion with the pblisher place. I've deleted this condition for thesis/reports and in the vol-pages macro too.

tomax31:
Sorry for the new pull request, but the way of updating a style remains totally confused for me without creating a new pull-request. If I believe the help pages, I guess I have to merge my changes with a previous version, but I don't see this merge button. I will try again, but not sure to be able to do it.

PR #5468, opened 2021-05-25
Create worlds-veterinary-journal.csl

POBrien333:
via https://forums.zotero.org/discussion/89379/style-request-world-s-veterinary-journal#latest

Adapted for article-journal. Rest of the guidelines is not clear.


rmzelle:
@POBrien333, thanks!

PR #5469, opened 2021-05-25
Update experimental-biology-and-medicine.csl

EKP-LHSC:
Just adding bolding to the volume number in journal citations.  Received feedback from the EBM editorial office that journal volume numbers need to be bold in their bibliographies.


rmzelle:
@EKP-LHSC, thanks!

PR #5470, opened 2021-05-26
Update wirtschaftsuniversitat-wien-institut-fur-bwl-des-aussenhandels…

dkufner:
….csl

just added editor in names


rmzelle:
@dkufner, thanks!

PR #5472, opened 2021-05-26
Create ethnographiques-org.csl

POBrien333:
via https://forums.zotero.org/discussion/89410/style-request-ethnographiques-org#latest

### DIFF
--- a/ethnographiques-org.csl
+++ b/ethnographiques-org.csl
@@ -5,14 +5,14 @@
     <id>http://www.zotero.org/styles/ethnographiques-org</id>
     <link href="http://www.zotero.org/styles/ethnographiques-org" rel="self"/>
     <link href="http://www.zotero.org/styles/metropolitiques" rel="template"/>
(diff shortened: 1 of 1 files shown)


ethnographiques-org.csl
```diff
@@ -0,0 +1,245 @@
...
+      <term name="ordinal-01">ère</term>
+      <term name="ordinal-02">e</term>
+      <term name="ordinal-03">e</term>
+      <term name="ordinal-04">e</term>
+      <term name="cited">op. cit.</term>
+      <term name="page" form="short">p.</term>
+      <term name="issue" form="short">n°</term>
+    </terms>
+  </locale>
+  <macro name="author">
```
adam3smith, line 29: Any reason not to add proper substitute here instead of the choose (and also include the title -- see the webpage citation in the auto-preview)

POBrien333: Reason is I didn't look at that part as the guidelines are a bit messy. ha.
Will fix. Thanks for checking. :)

rmzelle:
@POBrien333, merci!

PR #5476, opened 2021-05-28
Update harvard-bournemouth-university.csl

4mot:
- Old documentation hyperlink lead to 404
- When more than one source is cited, the delimiter should be ;
- Oxford commas are not used for authors/editors/translators in BU-Harvard


rmzelle:
@4mot, thanks!

PR #5478, opened 2021-06-01
Create ulua-revista-de-historia-sociedad-y-cultura.csl

POBrien333:
via https://forums.zotero.org/discussion/89581/style-request-ulua-revista-de-historia-sociedad-y-cultura#latest

### DIFF
--- a/ulua-revista-de-historia-sociedad-y-cultura.csl
+++ b/ulua-revista-de-historia-sociedad-y-cultura.csl
@@ -63,7 +63,7 @@
   </macro>
   <macro name="author-short">
     <names variable="author">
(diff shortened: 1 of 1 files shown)


adam3smith:
@POBrien333 -- I didn't find anything on the et-al rules: but I did see an item with for authors listed in the bibliography, so I bumped the numbers higher. Did you see any rule or pattern about this?

I also turned et al italic in text as I saw that (Ibarra et al. 2014 here: https://ulua.uv.mx/index.php/ulua/article/view/2685/4547 )

rmzelle:
(looks good to me otherwise)

adam3smith:
Thanks!

POBrien333:
Thank you for making those amendments.

PR #5479, opened 2021-06-01
Update hochschule-bonn-rhein-sieg.csl

daryoush86:


adam3smith:
There's also a note style which might get added later in the same style guide.

PR #5485, opened 2021-06-02
remove et-al from restoration-ecology.csl

POBrien333:
https://forums.zotero.org/discussion/90076/style-request-restoration-ecology-update#latest


adam3smith:
Thanks!

PR #5487, opened 2021-06-02
Multiple references of the same author : sorted by date first, ascending

pmartinolli:
Previously, the multiple references of the same author were sorted by title first, following the CMS, I am suggesting to sort them by ascending date of publication first.

Auparavant, les références multiples d'un même auteur étaient classées par titre, suivant le CMS je propose de les classer par date de publication ascendant en priorité. 

Sources :
- https://www.chicagomanualofstyle.org/book/ed17/part3/ch14/figures/fig009.html 
- https://www.chicagomanualofstyle.org/book/ed17/part3/ch15/psec018.html 
- https://www.chicagomanualofstyle.org/book/ed17/part3/ch14/psec071.html

### DIFF
--- a/chicago-fullnote-bibliography-fr.csl
+++ b/chicago-fullnote-bibliography-fr.csl
@@ -1541,9 +1541,9 @@
   <bibliography hanging-indent="true" et-al-min="11" et-al-use-first="7" subsequent-author-substitute="&#8212;&#8212;&#8212;" entry-spacing="0">
     <sort>
       <key macro="contributors-sort"/>
+      <key variable="issued"/>
       <key variable="title"/>
       <key variable="genre"/>
-      <key variable="issued"/>
     </sort>
     <layout suffix=".">
       <choose>


adam3smith:
The relevant section is
> In a bibliography (as opposed to a reference list; see 15.18), titles by the same author are normally listed alphabetically. 

from https://www.chicagomanualofstyle.org/book/ed17/part3/ch14/psec071.html

Not sure why you think this should be changed?

pmartinolli:
I think the vast majority of works made with this style (in articles, book sections, thesis) use a list of sources of the type "list of references" not of the type "bibliography". If you also think it is then the most relevant section is this one : https://www.chicagomanualofstyle.org/book/ed17/part3/ch15/psec018.html
 > Chronological order for repeated names in a reference list

Thank you for your work!

(IMHO the confusion between "list of references" and "bibliography" is very common in academia and I find it complicated to explain to students.)

adam3smith:
The Chicago Manual makes the distinction between reference list and bibliography based on the reference type. 
Author-date systems use a reference list -- that's discussed in chapter 15, and in line with the Manual, we sort by date for Chicago (author-date). 

Note systems use a bibliography, that's chapter 14 in the Manual, and follows alphabetic sorting as per the passage I (and you) link to above. I think it's very hard to argue that the current sorting of the styles isn't a correct implementation of Chicago Manual.

As always, you're absolutely welcome to distribute custom versions of the style, but we'll keep the official one as is.

pmartinolli:
I agree. Thank you for this clear explanation.

PR #5491, opened 2021-06-06
Update din-1505-2.csl

POBrien333:
via https://forums.zotero.org/discussion/89833/suppress-author-issue-with-a-comma#latest


adam3smith:
Thanks!

PR #5495, opened 2021-06-08
Update zeitschrift-fur-fantastikforschung.csl

simifilm:
Added `translated-title` for films.

### DIFF
--- a/zeitschrift-fur-fantastikforschung.csl
+++ b/zeitschrift-fur-fantastikforschung.csl
@@ -15,7 +15,7 @@
     <category field="generic-base"/>
     <issn>2192-0885</issn>
     <summary>Stil für die Zeitschrift für Fantastikforschung</summary>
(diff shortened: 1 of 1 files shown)


simifilm:
Do I understand this correctly that `translated-title` is not part of spec? It works in Zotero though and is listed [here](https://github.com/dhimmel/csl-schema/blob/f216964aa152e78c8d90090e903cc63b7a47d6dc/schemas/styles/csl-variables.rnc)

adam3smith:
Yes -- you're looking at (a fork of) the development version of the spec. The official specs are here:
https://docs.citationstyles.org/en/stable/specification.html#appendix-iv-variables
The current, 1.0.1 schema is tagged as 1.0.1 https://github.com/citation-style-language/schema/releases/tag/v1.0.1

`translated-title` specifically was also removed from the 1.0.2 release as per https://github.com/citation-style-language/schema/pull/341

PR #5498, opened 2021-06-10
Update APA disambiguation rules to only initials

bwiernik:
I had thought we couldn't support the changed rule in APA 7th edition, but I realize that what they want is just "primary-name-with-initials" described in a convoluted way


rmzelle:
@adam3smith, looks fine to me. Will you merge this after giving it a look over as well?

adam3smith:
Yeah, I tested this this morning, works exactly as expected.

PR #5499, opened 2021-06-11
move citation number outside first field

JohannesNE:
In the bibliography layout, citation-number was inside the first group (field). This made second-field-align="flush" mess up the layout.


adam3smith:
Thanks, good catch

PR #5500, opened 2021-06-11
revise Manchester Univ. Press style (note-bib)

icornelius:
1. set page-range-format to "minimal" (as directed by press style)
2. fix doi prefix (as directed by press style)
3. use ibid. (permitted by press style)


rmzelle:
@icornelius, thanks! (also for being a continuing sponsor of CSL!!)

PR #5501, opened 2021-06-12

cookie365:
A student recently pointed out to us that the sorting of works by the same author does not meet the requirements of our style. It should be sorted by author, then year, then title.


adam3smith:
Thanks!

PR #5505, opened 2021-06-14
Update hawaii-international-conference-on-system-sciences-proceedings…

### LINKED ISSUE #5503: Could you update HICSS citation style?
Hi!

Thank you all for preparing the HICSS reference style!

Now the style on github orders the references by authors alphabetical order.
https://github.com/citation-style-language/styles/blob/master/hawaii-international-conference-on-system-sciences-proceedings.csl

For the conference HICSS this year https://hicss.hawaii.edu/
HICSS says:
“References should be organized according to the order in which they are referred to in the paper.”

Could you kindly update the style?

I think having the citation number ordered according to their appearance in the paper is similar to IEE citation style, but the bibliography format still needs to be HICSS style. So it would be great if you can update the HICSS style on github.

Sorry that I am not familiar with github. It would be great if you can help. Because the HICSS submission deadline is June 15 this year. Thank you so much!

Best,
Sarah

POBrien333:
closes #5503

### DIFF
--- a/hawaii-international-conference-on-system-sciences-proceedings.csl
+++ b/hawaii-international-conference-on-system-sciences-proceedings.csl
@@ -94,10 +94,6 @@
     </layout>
   </citation>
   <bibliography et-al-min="7" et-al-use-first="3" entry-spacing="1">
-    <sort>
-      <key macro="author"/>
-      <key variable="issued"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks!

PR #5506, opened 2021-06-14
Fix ArchéoSciences

nfrerebeau:
Uses prefixes instead of suffixes to delimit volume and issue of journal articles.


POBrien333:
A better, more robust way of doing this is actually using group delimiters here. :)

adam3smith:
Also, given that this is a French style, shouldn't there be a space (or nbsp) before the colon?

nfrerebeau:
Thanks for suggesting the group delimiters and the missing space before the colon! I just made the changes.

adam3smith:
Thanks!

PR #5509, opened 2021-06-15
Update proceedings-of-the-royal-society-b.csl

benjaminmoon:
Change the bibliography format to use hanging indent. Updated contributors and and update date and time.

### DIFF
--- a/proceedings-of-the-royal-society-b.csl
+++ b/proceedings-of-the-royal-society-b.csl
@@ -24,11 +24,15 @@
       <name>Michael Doube</name>
       <email>mdoube@rvc.ac.uk</email>
     </contributor>
+    <contributor>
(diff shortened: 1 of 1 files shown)


adam3smith:
What is this trying to fix? `second-field-align` and `hanging-indent-true` are rarely used together and looking at the journal, second-field-align would seem to be the correct option here.

benjaminmoon:
I was using this style through pandoc and wasn’t getting the alignment of the bibliography: entries 2 onwards were indented in the PDF produced. I realise this may be more an issue of the pandoc conversion as I’d agree with you that second-field-align is better suited.

adam3smith:
If you can reproduce that, absolutely do report to citeproc-pandoc. Closing here as the style itself seems correct as is.

PR #5510, opened 2021-06-17
Patch 4

Brenkacz:


adam3smith:
@rmzelle do you recall the history behind the title of the style? I could go either way on this. The substantive changes look good to me.

rmzelle:
@adam3smith, we have used the original title format for other styles (like the-university-of-tokyo-law-review.csl), but since they call themselves "Eksploatacja i Niezawodność – Maintenance and Reliability" at http://www.ein.org.pl/ let's go with that. Since it's an English-only journal we can drop the "English" as well here.

rmzelle:
@Brenkacz, thanks!

(@adam3smith, just swapped the dash in the title for a hyphen)

PR #5514, opened 2021-06-17
Create fundamental-and-applied-limnology.csl

POBrien333:
via https://forums.zotero.org/discussion/86932/style-request-fundamental-and-applied-limnology#latest


rmzelle:
@POBrien333, thanks! (the instructions to authors say all manuscripts must be in English)

PR #5518, opened 2021-06-20
Create European-Heart-Journal-2021.csl

mattdepuydt:
Hello,

GitHub first time user here. I created this style for the European Heart Journal since the two available (european-heart-Journal and European-society-of-cardiology) were outdated.

Looking forward to hearing your comments.

Matthieu

### DIFF
--- a/European-Heart-Journal-2021.csl
+++ b/European-Heart-Journal-2021.csl
@@ -1,0 +1,169 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" default-locale="en-GB" demote-non-dropping-particle="never">
+  <info>
+    <title>European-Heart-Journal-2021</title>
+    <id>http://www.zotero.org/styles/European-Heart-Journal-2021</id>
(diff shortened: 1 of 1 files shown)


adam3smith:
If the ESC style has changed, we'd want this as an update to `european-society-of-cardiology`, not as a separate style. The current style looks pretty good to me, though, so maybe could you first say what's wrong with it?

mattdepuydt:
@adam3smith > If the ESC style has changed, we'd want this as an update to `european-society-of-cardiology`, not as a separate style. The current style looks pretty good to me, though, so maybe could you first say what's wrong with it?

Thank you for the quick reply. I found three issues with current style using Mendeley.
The first issue is for a long list of authors where the current style uses 'et al.' cannot be used according to the journal instructions. The second issue is that the publisher should not appear either. See both issues in the snapshot for the citation of a journal article.
image

The last issue I found right now is that book section in the bibliography doesn't appear either e.g. _In: Arnold E, ed. McDonald's Blood Flow in Arteries: Theoretical, Experimental and Clinical Principles_ (see journal instructions to authors)

Thank you in advance for your help

adam3smith:
Continued here: https://github.com/citation-style-language/styles/pull/5518

PR #5519, opened 2021-06-21
Update european-society-of-cardiology.csl

adam3smith:
Continued from https://github.com/citation-style-language/styles/pull/5518


adam3smith:
[don't merge; I don't think all of these changes are quite right]

PR #5523, opened 2021-06-22
Create stellenbosch-university-harvard-2021.csl

Kirchnerv:
The previous style created by my colleague Yusuf Ras (yr@sun.ac.za) has a few issues that needed correcting, but the more we tried to fix it, the more we broke it! So we would like to generate a new style based on University of Cape Town Harvard style.

### DIFF
--- a/stellenbosch-university-harvard-2021.csl
+++ b/stellenbosch-university-harvard-2021.csl
@@ -1,0 +1,318 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+  <info>
+    <title>University of Cape Town - Harvard</title>
+    <title-short>SU-Harvard</title-short>
+    <id>http://www.zotero.org/styles/stellenbosch-university-harvard-2021</id>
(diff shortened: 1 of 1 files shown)


adam3smith:
Duplicate of #5524

stellenbosch-university-harvard-2021.csl
```diff
@@ -0,0 +1,318 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
+  <info>
+    <title>University of Cape Town - Harvard</title>
+    <title-short>SU-Harvard</title-short>
+    <id>http://www.zotero.org/styles/stellenbosch-university-harvard-2021</id>
+    <link href="http://www.zotero.org/styles/stellenbosch-university-harvard-2021" rel="self"/>
+    <link href="http://www.zotero.org/styles/harvard-university-of-cape-town" rel="template"/>
+    <link href="https://libguides.sun.ac.za/c.php?g=742962&p=5316902" rel="documentation"/>
```
rmzelle, line 9: @Kirchnerv, the main problem here was the ampersand ("&") in the URL, which is a special character in XML files and must be 'escaped' as `&amp;` instead. Unescaped ampersands invalidate the entire XML style and make them unparsable by computers.

PR #5527, opened 2021-06-23
Create alternatif-politika.csl

alternatifpolitika:
Style for the journal Alternatif Politika

### DIFF
--- a/alternatif-politika.csl
+++ b/alternatif-politika.csl
@@ -15,7 +15,6 @@
     <category field="social_science"/>
     <category field="political_science"/>
     <category field="humanities"/>
-    <issn>1309-0593</issn>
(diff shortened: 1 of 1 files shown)


rmzelle:

alternatifpolitika:
> Thanks! Did you base this on an existing CSL style? If so, we'd like to know which one, so we can add a "template" link (see https://github.com/citation-style-language/styles/blob/master/STYLE_REQUIREMENTS.md#8---template-link)

Thanks @rmzelle! I used an existing CSL style, but completely changed it. I'm not sure but I think it was: http://www.zotero.org/styles/journal-of-peace-research

rmzelle:
@alternatifpolitika, great, thanks!

PR #5529, opened 2021-06-25
Fix wrong article title prefix (period to comma)

StianOby:
Erroneous period after author name corrected to comma.


adam3smith:
Thanks!

PR #5531, opened 2021-06-27
Update civilta-italiana.csl

Pepppino:
Fixed the lowercase characters on the ibids: now if they are at the beginning of the citation they are correctly capitalised. Before the fix, they were always lowercase.


adam3smith:
Thanks! (we don't use the "published" element, so deleted that.)

PR #5532, opened 2021-06-27
Fix missing § locators in juristische-schulung.csl

Quintus:
The locator for book citations in this journal is only ommitted if the work is referenced by page (German "S." abbreviation). Any other locators should not be ommitted. Most notably, many German law science books are cited by § plus margin number (and some only by margin number). The previous version would swallow the § in that case.

  -quintus


adam3smith:
Thanks!

Issue #4816, opened 2020-05-20
Journal update to citation style-Estuaries and coasts


wryanjames:
There has been an update to the bibliography style in Estuaries and Coasts. They have switched (Jan 2020 is the earliest I could detect) from full first names to initials follow by a period. I wasn't able to find an updated style and thought this might be of interest given the style has not been updated since 2014. It seems to be a recent journal update. I have attached a pdf of author guidlines that includes the update citation style. I apologize because I would submit and edit myself, but I am not familiar with csl file formats 

[Estuaries_and_Coasts_IFA_Jan2020.pdf](https://github.com/citation-style-language/styles/files/4657641/Estuaries_and_Coasts_IFA_Jan2020.pdf)

rmzelle:
@POBrien333, would you mind taking a look? See https://github.com/citation-style-language/styles/blob/master/dependent/estuaries-and-coasts.csl.

POBrien333:
Hi @wryanjames 

Could you help me with checking some of the dependent styles of the parent style springer-humanities-author-date?
Could you check 3 or 4 of any of these journals and see if they also changed the style? Based on that we either then change the parent style or make an independent style for _Estuaries_.

thanks.

[...]

PR #5534, opened 2021-06-29
Update and rename dependent/estuaries-and-coasts.csl to estuaries-and…

POBrien333:
…-coasts.csl

via https://github.com/citation-style-language/styles/issues/4816


rmzelle:
@POBrien333, thanks!

PR #5535, opened 2021-06-30
Create bakhtiniana-journal-of-discourse-studies.csl

POBrien333:
https://forums.zotero.org/discussion/90249/style-request-bakhtiniana-journal-of-discourse-studies#latest


rmzelle:
@POBrien333, thanks! Looks like it's a bilingual journal (Portuguese/English), but we can take it as is.

PR #5536, opened 2021-07-01
Add support for blog posts (CSL: post-weblog).

StianOby:
Blogs are cited in the same way as newspaper articles, so I added support for them by just tacking on "post-weblog" in the applicable if-loops of relevant macros.


StianOby:
Fixed another minor error with name particle display.

rmzelle:
@StianOby, thanks!

PR #5538, opened 2021-07-01
Update tud bwl marketing style

christinawenzel:
Minor changes to match the current citation rules at the chair for marketing at TUD:
  - number of explicitly named authors
  - `/` for separating authors


rmzelle:
@christinawenzel, thanks!

PR #5539, opened 2021-07-02
Further improvements to norsk-henvisningsstandad-for-rettsvitenskapelige-tekster.csl

StianOby:
- Refactor so that the "page" term is used rather than hardcoded text prefixes.
- Refactor DOI macro and added DOI support to additional item types (book, book chapter, thesis, paper-conference)


rmzelle:
@StianOby, thanks!

PR #5540, opened 2021-07-02
Add hanging-indents to infoclio.ch styles

nchachereau:
By popular demand, infoclio.ch has decided to use hanging-indents for their custom styles for history students in Switzerland (see #234 for the original pull request and backlinks to subsequent updates).

Thank you for taking the time to maintain the styles, this is such important work for so many scholars! :pray: 

(cc @maehr)

### DIFF
--- a/infoclio-de-kurzbelege.csl
+++ b/infoclio-de-kurzbelege.csl
@@ -57,7 +57,7 @@
       </choose>
     </layout>
   </citation>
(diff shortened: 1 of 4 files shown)


maehr (approved):

rmzelle:
@nchachereau, @maehr, thanks, and thanks for the thanks!

PR #5541, opened 2021-07-07
OSCOLA.csl: Add publiser info to generic secondary sources (CSL type "article").

StianOby:
An attempt to fix this issue: https://forums.zotero.org/discussion/comment/359163/#Comment_359163

PS: I am not intimately familiar with the code for the OSCOLA CSL style. Thus my technique for fixing the issue may not be optimal. Please double-check my work before committing to master.

### DIFF
--- a/oscola.csl
+++ b/oscola.csl
@@ -255,7 +255,7 @@
   <!--publication info -->
   <macro name="publisher">
     <choose>
-      <if type="book chapter broadcast personal_communication manuscript paper-conference article-newspaper report legislation motion_picture speech interview thesis entry-encyclopedia webpage post-weblog article document" match="any">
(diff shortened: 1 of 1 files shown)


StianOby:
(I suspect that this has slipped through the cracks, therefore tagging @adam3smith – sorry if I'm mistaken.)

adam3smith:
Sorry for the delay here. We actually want `document` for generic going forward. `article` will be used for preprints as of CSL 1.2

StianOby:
Thanks for pointing out this CSL 1.2 issue! AFAIK there are no real differences between generic documents and preprinst (working papers) in OSCOLA. Thus I attempted to fix the CSL 1.2 issue by just adding the "document" item type as well. I hope this is OK.

StianOby:
PS: why is this suddenly not validating? Validator is complaining about a "premature EOF [End Of File] at line 168". But I haven't touched that line...

adam3smith:
Sorry, I jumped the gun there -- document isn't available in 1.0.1

PR #5542, opened 2021-07-07
Add a new journal (magnetic resonance in medical sciences)

akchan:
Hi, 

I have added a journal style of magnetic resonance in medical sciences (MRMS) and want to merge this to official CSL repository for public benefit.

The journal is an international and peer reviewed medical journal for magnetic resonance imaging.

Journal home  [https://www.jsmrm.jp/modules/en/index.php?content_id=6](https://www.jsmrm.jp/modules/en/index.php?content_id=6)

This new style have passed to validation using [CSL Validator](https://validator.citationstyles.org).

### DIFF
--- a/magnetic-resonance-in-medical-sciences.csl
+++ b/magnetic-resonance-in-medical-sciences.csl
@@ -3,7 +3,8 @@
   <info>
     <title>Magnetic Resonance in Medical Sciences</title>
     <title-short>Magn Reson Med Sci</title-short>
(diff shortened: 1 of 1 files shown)


rmzelle:

akchan:
>rmzelle

Thank you for your reply.
This style has been built from scratch. So, I didn't add template tag. :)

Then, what should be filled to template tag? Is it ok to be empty?

rmzelle:
@akchan, okay, thanks! Yes, in that case we wouldn't add a template link.

PR #5543, opened 2021-07-07

POBrien333:
via https://forums.zotero.org/discussion/90651/update-incorrect-style-request-veterinary-pathology


rmzelle:
@POBrien333, thanks!

Issue #5544, opened 2021-07-08
Update brain.csl


rmzelle:
https://academic.oup.com/brain/pages/General_Instructions

> References
> 
> In-text citations
> 
> Brain is transitioning to numerical citations in-text, based on the AMA Reference Style (11th Edition). ...

@POBrien333, do you have time to check if their instructions are for vanilla AMA (american-medical-association.csl)? Request came in this week via https://citationstyles.org/contact/#/contact-form.

PR #5546, opened 2021-07-08
Update brain.csl to ama.csl dependent

POBrien333:
closes https://github.com/citation-style-language/styles/issues/5544

Make dependent of AMA11.


rmzelle:
@POBrien333, thanks!

PR #5547, opened 2021-07-08
Create szociologiai-szemle.csl

adamstefkovics:


rmzelle:

adamstefkovics:
Yes, it's been based on https://www.zotero.org/styles/apa
thank you!!

rmzelle:
@adamstefkovics, thanks!

PR #5548, opened 2021-07-09
Create zeitschrift-fur-papyrologie-und-epigraphik.csl

Jmuccigr:
New style for this international journal.


Jmuccigr:
I'm a little confused by this error. I suspect it means that dependent styles shouldn't have their own macros and I should make this a non-dependent style.

POBrien333:
A dependent style CANNOT have macros <citation or <bibliography.

What you are doing is just making a new independent style, based on a template. So, take the style out of the dependent folder and the test error will disappear.

Also, no need to open an issue. There is only limited capacity, but the two main curators of the repository will get to this when they can.

Jmuccigr:
> A dependent style CANNOT have macros <citation or <bibliography.
> 
> What you are doing is just making a new independent style, based on a template. So, take the style out of the dependent folder and the test error will disappear.
> 
> Also, no need to open an issue. There is only limited capacity, but the two main curators of the repository will get to this when they can.

Thanks for both comments. Will make the change.

rmzelle:
@Jmuccigr, you're right that the citation requirements/guidance at https://ifa.phil-fak.uni-koeln.de/zeitschriften-reihen/zeitschrift-fuer-papyrologie-und-epigraphik-zpe/wie-sollen-die-manuskripte-eingereicht-werden are very minimum. Did you base your style on the format used by the journal in published articles?

rmzelle:
> I'm a little confused by this error. I suspect it means that dependent styles shouldn't have their own macros and I should make this a non-dependent style.

In CSL, we have "independent" styles which are complete styles and must be placed in the root of this repository, and "dependent" styles which are more like aliases and must be placed in the "dependent" subfolder. This was an independent style in the wrong spot, hence the error.

Jmuccigr:
> @Jmuccigr, you're right that the citation requirements/guidance at https://ifa.phil-fak.uni-koeln.de/zeitschriften-reihen/zeitschrift-fuer-papyrologie-und-epigraphik-zpe/wie-sollen-die-manuskripte-eingereicht-werden are very minimum. Did you base your style on the format used by the journal in published articles?

Yes, based on what that and also on changes made to a recent submission I made to the journal (which is why I came up with the style in the first place).

rmzelle:
@Jmuccigr, okay, great. Thanks!

PR #5553, opened 2021-07-11
Update institut-français-darcheologie-orientale-en.csl

NicolasSouchon:
Update of institut-français-darcheologie-orientale-en.csl according to the new version of the standards (available here: https://www.ifao.egnet.net/uploads/publications/normes/IFAO_publications_normes_bibliographiques_pub_egypto_2021_angl.pdf)


rmzelle:
@NicolasSouchon, thanks!

PR #5556, opened 2021-07-13
Create modern-language-association-9th-edition.csl

lpuwa:
The MLA Handbook 9th edition, section  has changed the formatting for DOIs to use the prefix "https://doi.org/" in place of "doi:". This is the only siginificant change from the 8th edition that effects the current 8th CSL file. (I am new to github - I hope I'm following the correct procedure to request this.)  This new version is a modified copy of styles/modern-language-association.csl


adam3smith:
Thanks! Added here https://github.com/citation-style-language/styles/pull/5561

PR #5557, opened 2021-07-13
Create international-journal-of-automotive-technology.csl

POBrien333:
via https://forums.zotero.org/discussion/85782/style-request-international-journal-of-automotive-technology#latest


rmzelle:
@POBrien333, thanks! 🚗

PR #5559, opened 2021-07-14
Update mohr-siebeck-recht.csl

POBrien333:
via https://forums.zotero.org/discussion/90740/mohr-siebeck-style-trennung-der-zitate#latest


rmzelle:
This change seems rather poorly supported. The old mohr.de domain seems to be dead, but I found some possible author guidelines at https://www.mohrsiebeck.com/service/fuer-autoren. Could you take a look if this is documented anywhere?

POBrien333:
I had looked through those documents already, but another user added their voice to this on the forum thread now.

rmzelle:
> I had looked through those documents already

Did you identify one of that could be a good replacement for "http://www.mohr.de/fileadmin/user_upload/Hinweise_Autoren_PDF/Merkblaetter/merkjura2006.pdf"?

POBrien333:
Let me get in touch with them.

adam3smith:
I say we just take this based on the comments in the tread.

PR #5561, opened 2021-07-17
Switch MLA to 9th edition

adam3smith:
Supersedes #5556 
Currently only changes DOI format.
Looking at changes, the other thing might be about adding format/medium, 
but that can wait

### DIFF
--- a/modern-language-association.csl
+++ b/modern-language-association.csl
@@ -11,7 +11,7 @@
     </author>
     <category citation-format="author"/>
     <category field="generic-base"/>
-    <summary>This style adheres to the MLA 0th edition handbook. Follows the structure of references as outlined in the MLA Manual closely</summary>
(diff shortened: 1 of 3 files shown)


modern-language-association.csl
```diff
@@ -11,8 +11,8 @@
     </author>
     <category citation-format="author"/>
     <category field="generic-base"/>
-    <summary>This style adheres to the MLA 8th edition handbook. Follows the structure of references as outlined in the MLA Manual closely</summary>
-    <updated>2018-12-13T20:05:10+00:00</updated>
+    <summary>This style adheres to the MLA 0th edition handbook. Follows the structure of references as outlined in the MLA Manual closely</summary>
```
StianOby, line 14: Seems to be a typo here ("0th" edition should be "9th" edition).

PR #5563, opened 2021-07-19
Update technische-universitat-dresden-linguistik.csl

fussballlinguist:
Minor changes:
1) DOI will be reported if available
2) Access date of online references will be reported
3) Collection title and number will be reported right behind the book title
4) If no author or editor is specified, "o. A." (ohne Autor) will re reported.

### DIFF
--- a/technische-universitat-dresden-linguistik.csl
+++ b/technische-universitat-dresden-linguistik.csl
@@ -16,7 +16,7 @@
     <category field="linguistics"/>
     <category field="literature"/>
     <summary>Zitierstil entsprechend den Vorgaben der linguistischen Professuren am Institut für Germanistik der Technischen Universität Dresden. Der Stil orientiert sich an den Richtlinien von 'Deutsche Sprache: Zeitschrift für Theorie Praxis Dokumentation', hg. vom IDS Mannheim</summary>
-    <updated>2020-04-24T11:56:46+00:00</updated>
+    <updated>2021-07-19T08:34:39+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="de">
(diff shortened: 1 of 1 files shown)


technische-universitat-dresden-linguistik.csl
```diff
@@ -27,6 +27,11 @@
     </terms>
   </locale>
   <macro name="creator-year-long">
+    <choose>
+      <if match="none" variable="author editor">
+        <text value="o. A." suffix=" "/>
+      </if>
+    </choose>
```
rmzelle, line 34: @fussballlinguist, this is fine, but it's slightly nicer to use `<substitute/>` to print "o. A.". There is also a dedicated "anonymous" term that can be used to print "o. A.".

Example of a style using `<substitute/>` as discussed above: https://github.com/citation-style-language/styles/blob/6152ccea8b7d7a472910d36524d1bf3557a83bfc/im-gesprach.csl#L43-L55
The German CSL locale file, which already uses "o. A." (with a non-breaking space) for the abbreviated form of the "anonymous" term: https://github.com/citation-style-language/locales/blob/ec6d62a9e7c049001a55086a495022520989441c/locales-de-DE.xml#L38

rmzelle: And I'll just merge this as-is, but if you are willing to make this change please feel free to open a new pull request for this.

rmzelle:
@fussballlinguist, thanks!

fussballlinguist:
Hi, I would like to change the DOI prefix to "https://doi.org". It works as it should, but when using the export function or drag'n'drop function in Zotero, only the DOI itself is hyperlinked and not the complete url. In Markdown syntax, looks like this: 

Mustermann, Max (2021): Kapiteltitel. In: Gates, Bill/Jobs, Steve (Hg.): Buchtitel (= Reihentitel 5). Musterstadt: Musterverlag. S. 1–20. https://www.doi.org/[10.15/00001](https://doi.org/10.15/00001).

How can I fix this?

adam3smith:
Post to Zotero forums. CSL styles don't handle hyperlinking at all

PR #5564, opened 2021-07-19
Create the-american-journal-of-dermatopathology.csl

POBrien333:
via https://forums.zotero.org/discussion/90808/request-for-assistance-in-adding-vancouver-ref-style-with-3-authors-and-then-et-al#latest


rmzelle:
@POBrien333, thanks!

PR #5568, opened 2021-07-25
Update sist02.csl

kdmsnr:
Add translator

ref. https://jipsti.jst.go.jp/sist/handbook/sist02_2007/sist02.htm#4-1-4


rmzelle:
@kdmsnr, thanks!

PR #5570, opened 2021-07-27
Create etri-journal.csl

POBrien333:
Minimal style only really adapted for article-journal.

via https://forums.zotero.org/discussion/87891/style-request-etri-j#latest


rmzelle:
@POBrien333, thanks!

PR #5573, opened 2021-07-28
Update for corrections

PriscilaSJUFRGS:
Correction of musical score type, disambiguation rule, ebook expression and others corrections.


rmzelle:
@PriscilaSJUFRGS, thanks!

PR #5580, opened 2021-08-01
[Frontiers in ...] Remove formatting for volume

andjar:
It seems that the style for Frontiers ... has changed and that volume should no longer be in bold and journal names should not be in italics: https://www.frontiersin.org/about/author-guidelines


andjar:
Great, thanks! Will update

adam3smith:
Thanks!

PR #5584, opened 2021-08-06
Create international-journal-of-clinical-research.csl

POBrien333:
via https://forums.zotero.org/discussion/91074/ama-style-but-with-a-square-bracket-and-no-superscript#latest


rmzelle:
@POBrien333, thanks!

PR #5586, opened 2021-08-07
New Style for "Neue Kriminalpolitik (Deutsch)"

AlexVonB:
Style for "Neue Kriminal Politik", https://www.nk.nomos.de/

This style implements the style guide provided by the publisher (pdf in German): https://www.nk.nomos.de/fileadmin/nk/doc/NK_Autorenhinweise_2015.pdf

### DIFF
--- a/neue-kriminalpolitik.csl
+++ b/neue-kriminalpolitik.csl
@@ -14,17 +14,17 @@
     <category citation-format="note"/>
     <category field="law"/>
     <issn>0934-9200</issn>
-    <updated>2021-08-07T13:57:34+00:00</updated>
+    <updated>2021-08-07T14:04:37+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="authors-inline">
     <names variable="author" font-style="italic">
-      <name form="short" font-style="normal" delimiter="/" name-as-sort-order="all"/>
+      <name form="short" font-style="normal" delimiter="/" et-al-min="3" et-al-use-first="1" name-as-sort-order="all"/>
     </names>
   </macro>
   <macro name="authors-bib">
     <names variable="author" font-style="italic">
-      <name form="short" font-style="normal" delimiter=" / " name-as-sort-order="all"/>
+      <name form="short" font-style="normal" delimiter=" / " et-al-min="19" et-al-use-first="18" name-as-sort-order="all"/>
     </names>
   </macro>
   <macro name="number-and-year">
@@ -38,7 +38,7 @@
       </else>
     </choose>
   </macro>
-  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-givenname="true">
+  <citation disambiguate-add-givenname="true">
     <layout delimiter="; " suffix=".">
       <choose>
         <if type="article-journal" match="any">
@@ -95,7 +95,7 @@
       </choose>
     </layout>
   </citation>
-  <bibliography et-al-min="19" et-al-use-first="18">
+  <bibliography>
     <sort>
       <key macro="authors-bib"/>
       <key variable="issued"/>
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks! Did you create this CSL style from scratch, or did you base it on an existing CSL style? If the latter, we'd like to know which one, so we can add a "template" link (see https://github.com/citation-style-language/styles/blob/master/STYLE_REQUIREMENTS.md#8---template-link)

neue-kriminalpolitik.csl
```diff
@@ -0,0 +1,164 @@
...
+    </author>
+    <category citation-format="note"/>
+    <category field="law"/>
+    <issn>0934-9200</issn>
+    <updated>2021-08-07T14:15:12+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <macro name="authors-inline">
+    <names variable="author" font-style="italic">
+      <name form="short" font-style="normal" delimiter="/" name-as-sort-order="all"/>
```
adam3smith, line 22: Would it make sense to have substitute elements here for editor and title so that things like edited volumes and webpages look better? (Check the automatically generated sample citations to see what I mean.) Same for author-bib

adam3smith (changes requested):
Could you check on my comments about authors? The sample citations look a bit odd still

neue-kriminalpolitik.csl
AlexVonB: You are right, I'll have to test some things with the bibliography, I'll let you know when this is ready. Thanks!

AlexVonB: Now it is resolved, thank you for your feedback; I did not think of books without authors or websites :)

neue-kriminalpolitik.csl
```diff
@@ -0,0 +1,176 @@
...
+      <substitute>
+        <names variable="editor">
+          <name form="short" delimiter="/" name-as-sort-order="all"/>
+        </names>
```
adam3smith, line 26:
```suggestion
        <names variable="editor"/>
 ```
it will inherit `name` formatting from the author.

File: neue-kriminalpolitik.csl [hunk 3]
```diff
@@ -0,0 +1,176 @@
...
+        <text variable="title" form="short"/>
+      </substitute>
+    </names>
+  </macro>
+  <macro name="authors-bib">
+    <names variable="author" font-style="italic">
+      <name form="short" font-style="normal" delimiter=" / " name-as-sort-order="all"/>
```
adam3smith, line 33:
```suggestion
      <name form="short" font-style="normal" delimiter=" / " name-as-sort-order="all"/>
      <label form="short" prefix=" (" suffix=")"/>
```

Presumably editors should be labelled as such

neue-kriminalpolitik.csl
```diff
@@ -0,0 +1,176 @@
...
+      <substitute>
+        <names variable="editor">
```
adam3smith, line 35: same as above

neue-kriminalpolitik.csl
```diff
@@ -0,0 +1,176 @@
...
+          <choose>
+            <if match="all" variable="container-title">
+              <group delimiter=" " prefix=". ">
+                <text term="in" suffix=":"/>
+                <choose>
+                  <if match="any" variable="editor">
+                    <names variable="editor">
+                      <name form="short" delimiter=" / " et-al-min="2" et-al-use-first="1"/>
+                    </names>
+                    <text term="editor" form="short" prefix="(" suffix="),"/>
```
adam3smith, line 164: you really want to do this with label, not with term -- see example above.

adam3smith (commented):
Some further questions/suggestsions.

neue-kriminalpolitik.csl
AlexVonB: Label does not have "editor", at least not in the online editor. The validator is unhappy, as well. I reverted the change, please let me know if i misunderstood something. Thanks!

adam3smith: See the code I gave you for l. 33 above. Label doesn't take a creator type argument because it inherits that from names

AlexVonB:
@adam3smith I really appreciate your continuous help! Sorry for being late to react, and thank you for not giving up on this PR <3

PR #5587, opened 2021-08-08
revise Manchester UP monograph style (note-bib)

icornelius:
print variable="translator" for type="chapter"


rmzelle:
@icornelius, thanks!

PR #5589, opened 2021-08-09
Create clinical-pulmonary-medicine.csl

POBrien333:
via https://forums.zotero.org/discussion/88996/clinical-pulmonary-medicine#latest

100% match. The parent style is from the same publisher, so I just made a dependent style.


rmzelle:
@POBrien333, thanks!

PR #5591, opened 2021-08-10
Create american-journal-of-sociology.csl

POBrien333:
via https://forums.zotero.org/discussion/91067/style-request-american-journal-of-sociology-ajs


rmzelle:
@POBrien333, thanks!

PR #5414, opened 2021-04-24
Create AFTE CSE.csl

PavelGiv:

PR #5593, opened 2021-08-12
Create council-of-science-editors-brackets.csl

POBrien333:
continues #5414

1. create independent CSE brackets style
2. create dependent style for AFTE Journal


adam3smith:
Thanks?

PR #5594, opened 2021-08-12
Update journal-of-paleontology.csl

POBrien333:
via https://forums.zotero.org/discussion/91165/style-error-journal-of-paleontology#latest


adam3smith:
Thanks!

Issue #5401, opened 2021-04-20
Renaming an existing style


jjrondeau:
Hi,
We are planning to rename our "universite-du-quebec-a-montreal" style "universite-du-quebec-a-montreal-APA". I am wondering if I should proceed as for any other change within the code.

And once this will be done, what will be the impact for users that have already added this style to their Zotero? Will they have to reload the renamed style? Or will the UQAM style be automatically replaced by the renamed UQAM-APA style?

Thank you

rmzelle:
We try to keep changes to style IDs (and file names) to a minimum. Would it work to just change the style title?

ocouch:
@rmzelle There's no ideal way out of this, but what about duplicating the file and renaming, and set up the old file as a dependent style? (and insert comments in the file directing style editors to modifythe new style, and recommending they update their links).

That way you maintain legacy compatibility while keeping maintenance fairly straightforward.

rmzelle:
> @rmzelle There's no ideal way out of this, but what about duplicating the file and renaming, and set up the old file as a dependent style? (and insert comments in the file directing style editors to modifythe new style, and recommending they update their links).
> 
> That way you maintain legacy compatibility while keeping maintenance fairly straightforward.

We'd either change the style ID and include the renaming in https://github.com/citation-style-language/styles/blob/master/renamed-styles.json, or leave the style ID as is. I'd really like to avoid having duplicates in the repository.

Zotero does use https://github.com/citation-style-language/styles/blob/master/renamed-styles.json, but IIRC only updates it with new releases of Zotero. This would automatically update the installed style in a Zotero installation, but I think it would still result in permanent breakage of word processor documents generated with the old style (Zotero will likely complain to the user that it can't find the old style and ask the user to select a different one).

jjrondeau:
My apologies for the late reply. I'm back to work on this file. 
So, if I understand correctly, I could change the style title for Université du Québec à Montréal - APA (Français - Canada)  and style ID for http://www.zotero.org/styles/universite-du-quebec-a-montreal-apa? To avoid confusion, I would prefer to avoid to have 2 styles in the repository, named Université du Québec à Montréal - APA (Français - Canada) and Université du Québec à Montréal (Français - Canada)

PR #5596, opened 2021-08-13

jjrondeau:
The style has been modified to comply with the APA 7th ed. with a few adaptations for legal documents. At this point, we would like to rename it universite-du-quebec-a-montreal-apa.

### DIFF
--- a/universite-du-quebec-a-montreal.csl
+++ b/universite-du-quebec-a-montreal.csl
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" et-al-min="3" et-al-use-first="1" demote-non-dropping-particle="never" default-locale="fr-CA">
   <info>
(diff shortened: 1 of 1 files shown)


universite-du-quebec-a-montreal.csl
```diff
@@ -5,7 +5,7 @@
     <title-short>UQAM</title-short>
```
rmzelle, line 5: @jjrondeau, I assume you'd like to change the title to `<title>Université du Québec à Montréal - APA (Français - Canada)</title>` as well?

To continue our discussion from https://github.com/citation-style-language/styles/issues/5401, in general it's better to just leave the style ID as-is, as most users will just look at the style title and probably never notice the style ID, and changing the ID has some significant downsides.

jjrondeau: Yes, exactly. Changing only the title would be fine. Style ID (and self link?) can stay unchanged if this is possible, although the Visual CSL Editor warn me it doesn't respect style repository convention.
Thank you.

rmzelle:
@jjrondeau, thanks!

PR #5598, opened 2021-08-13
Remove extra accent in ArcheoSciences

nfrerebeau:
There is no accent in the name of the ArcheoSciences journal (see https://journals.openedition.org/archeosciences/).


rmzelle:
@nfrerebeau, merci!

PR #5602, opened 2021-08-19
Update independent style for journal-of-cataract-and-refractive-surgery.csl

POBrien333:
https://forums.zotero.org/discussion/91276/citation-style#latest


rmzelle:
@POBrien333, thanks!

PR #5603, opened 2021-08-19
Update style-manual-australian-government.csl

POBrien333:
Reworking of style based on the new guidelines: https://www.stylemanual.gov.au/style-rules-and-conventions/referencing-and-attribution/author-date

They really completely changed the style. Can you confirm what I am seeing? I'm just amazed that they so drastically changed it.

We have the following dependencies of this style: 

- Southern Cross University - Harvard
- The University of Sydney - Harvard
- University of Tasmania - Harvard

Via https://forums.zotero.org/discussion/89878/agps-vs-elsevier-harvard#latest


adam3smith:
Yes, they did completely change the style. The background is that this emerged out of a commercial printing manual which was actually pretty expensive and then, quite reasonably, the government at some point just decided to just create a public version of their own style. 
I'd actually keep (just like with cite-them-right) both styles. Let's call it "Style manual for authors, editors and printers 6th edition (Snooks & Co.)"
For the dependent styles, I checked all three and they all use the old style:
https://utas.libguides.com/referencing/Harvard
https://libguides.scu.edu.au/harvard
https://libguides.library.usyd.edu.au/c.php?g=508212&p=3476130

POBrien333:
Thanks for checking the dependent styles.

In this case, I'll rename the old style to the name you suggested and add a new file with the old name?

adam3smith:
and then switch all the dependents over, yes, thanks! I updated the style name above. The manual actually has a Vancouver and a note style, too, but no one has ever asked for them.

POBrien333:
Great. I think that should be it. :)

Some heavier lifting on the repository this week.

PR #5604, opened 2021-08-21
Update american-journal-of-sociology.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/388721/#Comment_388721


adam3smith:
Thanks!

PR #5605, opened 2021-08-22

jotaemi:
Edit to show interviewer in the reference list as discussed here: https://forums.zotero.org/discussion/comment/369403/#Comment_369403


adam3smith:
Thanks! This might take a week or two to review because we're testing changes to Chicago and similar high use styles more, but looks good in general

adam3smith:
Thanks!

PR #5610, opened 2021-08-27
fix erros in universitat-zu-koln-seminar-fur-abwl-und-finanzierungslehre.csl

POBrien333:
Found errors while using this style as a template.


adam3smith:
Thanks!

PR #5612, opened 2021-08-31
Create harvard-cite-them-right-11.csl

POBrien333:
Guidelines a user provided from the official Cite them Right webpage:
Part 1: [Cite them right Harvard 11th Edition Guidance.pdf](https://github.com/citation-style-language/styles/files/7082155/Cite.them.right.Harvard.11th.Edition.Guidance.pdf)
Part 2: [Cite them right Harvard 11th edition Guidance (other half).pdf](https://github.com/citation-style-language/styles/files/7089124/Cite.them.right.Harvard.11th.edition.Guidance.other.half.pdf)

via https://forums.zotero.org/discussion/89182/cite-them-right-11th-edition-harvard#latest

Changes made:
* Names are now shortened without the space (Campbell, J.L. and Pedersen, O.K. instead of Campbell, J. L. and Pedersen, O. K.)
* remove space after "doi:" prefix for DOI variable
* add preprint conditional for article-journal in "container-title" macro
* Add medium for artwork


adam3smith:
Could you
1. Rename the current cite-them-right style to `harvard-citem-them-right-10th-edition`
2. Rename this style to harvard-cite-them-right

(our practice is to keep old editions around for a bit but have the main style point at the most recent edition).

adam3smith:
Thanks!

POBrien333:
Awesome. Thanks for merging this so quickly.

Should we rename this style in any way?
https://www.zotero.org/styles?q=id%3Aharvard-cite-them-right-no-et-al

adam3smith:
ah good call -- could you just make a no-et-al version of the 11th edition style, add 11th edition to the title (but not the id) and replace is (it's a judgment call, but I don't think we need both 10th and 11th edition no et al. version; if people complain we can put the 10th edition version back in)

PR #5613, opened 2021-08-31
Rough implementation of book reviews

johnhpercival:
Built on code from @haraldaa https://forums.zotero.org/discussion/comment/359785/#Comment_359785
Added functionality for subsequent citations
Not checked against any edge cases

### DIFF
--- a/society-of-biblical-literature-fullnote-bibliography.csl
+++ b/society-of-biblical-literature-fullnote-bibliography.csl
@@ -296,7 +296,17 @@
         <text variable="title" font-style="italic" text-case="title"/>
       </else-if>
       <else-if variable="reviewed-author">
-        <text variable="title" font-style="italic" prefix="review of "/>
+        <choose>
+          <if variable="reviewed-title" match="any">
+            <group delimiter=", review of ">
+              <text variable="title" quotes="true"/>
+              <text variable="reviewed-title" font-style="italic"/>
+            </group>
+          </if>
+          <else>
+            <text variable="title" font-style="italic" prefix="review of "/>
+          </else>
+        </choose>
       </else-if>
       <else>
         <text variable="title" quotes="true" text-case="title"/>
@@ -317,7 +327,17 @@
       </else-if>
       <else-if variable="reviewed-author">
         <group delimiter=", ">
-          <text variable="title" font-style="italic" prefix="Review of "/>
+          <choose>
+            <if variable="reviewed-title" match="any">
+              <group delimiter=", review of ">
+                <text variable="title" quotes="true"/>
+                <text variable="reviewed-title" font-style="italic"/>
+              </group>
+            </if>
+            <else>
+              <text variable="title" font-style="italic" prefix="Review of "/>
+            </else>
+          </choose>
           <names variable="reviewed-author">
             <label form="verb-short" text-case="lowercase" suffix=" "/>
             <name and="text" delimiter=", "/>
@@ -729,8 +749,8 @@
                   </if>
                   <else>
                     <group delimiter=" ">
-                        <label variable="locator" form="short"/>
-                        <text variable="locator"/>
+                      <label variable="locator" form="short"/>
+                      <text variable="locator"/>
                     </group>
                   </else>
                 </choose>
--- a/society-of-biblical-literature-fullnote-bibliography.csl
+++ b/society-of-biblical-literature-fullnote-bibliography.csl
@@ -299,12 +299,12 @@
         <choose>
           <if variable="reviewed-title" match="any">
             <group delimiter=", review of ">
-              <text variable="title" quotes="true"/>
-              <text variable="reviewed-title" font-style="italic"/>
+              <text variable="title" quotes="true" text-case="title"/>
+              <text variable="reviewed-title" font-style="italic" text-case="title"/>
             </group>
           </if>
           <else>
-            <text variable="title" font-style="italic" prefix="review of "/>
+            <text variable="title" font-style="italic" prefix="review of " text-case="title"/>
           </else>
         </choose>
       </else-if>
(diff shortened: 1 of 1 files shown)


dstark:
@johnhpercival, thanks for putting this together. I hadn't gotten back to it yet. What I'm getting with a sample looks correct for **footnotes** for an "untitled book review" in SBLHS with one exception.

This style is still outputting quotation marks, but SBLHS2 §6.3.4 asks for no quotation marks. So, for instance, this style outputs

1. A. Andrew Das, “review of _Paul, Judaism, and the Gentiles: Beyond the New Perspective_,” by Francis Watson, Interpretation 63 (2009): 200, 202.
2. Das, “review of _Paul, Judaism, and the Gentiles: Beyond the New Perspective_ (by Watson),” 200.

But SBLHS2 wants 

1. A. Andrew Das, review of _Paul, Judaism, and the Gentiles: Beyond the New Perspective_, by Francis Watson, Interpretation 63 (2009): 200, 202.
2. Das, review of _Paul, Judaism, and the Gentiles: Beyond the New Perspective_ (by Watson), 200.

For the **bibliography**, SBLHS2 also wants no quotation marks. _And_ "Review" should be capitalized. I haven't been sure how to get "review" lowercased in the footnote except to put something in the language field, but then that breaks the capitalization in the bibliography.

I wonder—instead of putting "review of" in the title field, could we use the presence of a reviewed author to automatically add that text in the proper casing for the footnote and then in the bibliography? So, the title field would then just be the book title but the item type would be a journal article? (Failing something like this, we might just have to leave the lowercased "review" in the bibliography to be addressed in final manual edits.)

Thanks again for taking a look at this!

johnhpercival:
@dstark , thanks, I'm not getting the quote marks that you are, but getting the desired SBLHS2 output (in the Zotero Style Editor box). I am also seeing "Review" capitalised for bibliography. Could it be a locale problem or something like that creating this discrepancy?

I'm aware of two problems that this rough implementation creates:
1. Sort order in bibliography will likely be based on the book title rather than the text "review of"
2. The "suppress author" option results in citations that begin with lower case "review of", which is probably not desired behaviour.

dstark:
@johnhpercival, I guess I must not have pulled in the updated style file properly earlier. When I pulled it down just now, I'm seeing exactly what you're describing and what SBLHS2 seems to be expecting in this scenario.

On #1, you're right that the bibliography sort order doesn't take "review" into account. But this is a definite step forward.

Thanks so much!

johnhpercival:
@dstark, thanks for that confirmation. I think we're both agreed that this meets the basic requirements for SBLHS2 book reviews, but with the limitations noted above. 

Apologies for the multiple commits. I tried to make a further update to put the "review of" text as a <term> rather than hard-coded deep in the CSL, but realised that I could not define my own term for that purpose, so I have reverted to the original commit.

dstark:
> Apologies for the multiple commits. I tried to make a further update to put the "review of" text as a rather than hard-coded deep in the CSL, but realised that I could not define my own term for that purpose, so I have reverted to the original commit.

No worries on my account. I agree custom variables would be a nice addition at some point. :-) Thanks again for your work on this update!

society-of-biblical-literature-fullnote-bibliography.csl
```diff
@@ -315,6 +347,9 @@
       <else-if type="bill book graphic legal_case legislation motion_picture report song" match="any">
         <text variable="title" form="short" font-style="italic" text-case="title"/>
       </else-if>
+      <else-if variable="reviewed-author">
+        <text variable="title" font-style="italic" prefix="review of "/>
```
adam3smith, line 351: How about this: it uses the `reviewed-title` variable where available and otherwise falls back to your solution. `reviewed-title` isn't currently in Zotero but is available in CSL, will likely be in Zotero in the future and can currently be used in Zotero using Extra (and in other tools using CSL). 

```suggestion
        <choose>
          <if variable="reviewed-title" match="any">
            <group delimiter=", ">
              <text variable="title" quotes="true"/>
              <text variable="reviewed-title" prefix="review of "/>
            </group>
          </if>
          <else>
            <text variable="title" font-style="italic" prefix="review of "/>
          </else>
        </choose>
```

johnhpercival: Thanks, that looks like a reasonable suggestion. Two quick thoughts on the implementation:
1) You probably need line 355 "reviewed-title" variable to be in italics
2) You probably need changes near lines 299 & 320 for complete coverage

adam3smith:
@dstark @johnhpercival Looks good to me -- do you want to give it one last test/look with my updates?

johnhpercival:
Getting there. I think we may need to add text-case="title" to various lines (302, 303, 307, 333, 334, 338, 374, 375, 379, etc.?). Sorry in a rush so can't dig into this just now.

johnhpercival:
Also, I'm wondering about the example given in SBLHS2:

9. Jaroslav Pelikan, “The Things That You’re Liable to Read in the Bible,” review of _The Anchor Bible Dictionary_, ed. David Noel Freedman. _New York Times Review of Books_, 20 December 1992, 3.

Although we can't (currently?) generate this via Zotero as it's an edited volume rather than an authored volume, the period after the editor's name and before the journal name is different to what this proposed change achieves. However, this particular example may be a mistake in SBLHS, I'm not sure.

I also cannot see where in SBLHS2 that it defines the style for a newspaper article. 6.3.9 has a magazine article, but that layout is very similar to a journal article, different to the current CSL output for a magazine or newspaper article. But that may require a separate thread!

dstark:
@adam3smith, thanks. This looks like a good step forward to me.—Just to double check: You weren't intending your edits to solve the bibliography sorting issue, correct? Somehow, I thought I remembered seeing that in a comment from you here, but I may well have imagined it as I'm not seeing it now and reviews still sort in the bibliography according to the first letter of the title rather than "R" for "Review of."

@johnhpercival, I haven't interacted with SBL Press about that section specifically. But I have scratched my head over it as well. I'm not sure I understand the difference between a "titled book review" and a "review article" in SBLHS2 §6.3.4. From what I see, usually the only titled book reviews are review articles. So, the example that you're citing seems very strange (though it is indeed what SBLHS2 has), especially given the guidance that "titled book reviews should be cited as normal journal articles"—which is apparently different from the following "review article" example? What they're trying to communicate with the Pelikan example isn't clear to me, and I'd guess also isn't entirely distinct from the "review article" category.

So, for what it's worth, I'd certainly suggest going ahead and pushing these changes and leaving that example aside for now.

johnhpercival:
@adam3smith @dstark , I've now had a chance to address the sort issue, and the title-case issue. Are you happy with this version now?

adam3smith:
Looks good to me. I'll merge once dstark has had a look. Thanks!

dstark:
Looks good to me too, @adam3smith.

Thanks so much for putting this together, @johnhpercival!

adam3smith:
Thanks!

PR #5619, opened 2021-09-02
Update urad-rs-za-makroekonomske-analize-in-razvoj.csl

majazaloznik:
Minor fixes:
* italics only for book and legislation titles
* URLs in legislation
* 3 or more authors in citation -> et al.


rmzelle:
@majazaloznik, thanks!

majazaloznik:
you're a star @rmzelle! :)

PR #5624, opened 2021-09-07
Create netherlands-journal-of-geosciences.csl

POBrien333:
via https://forums.zotero.org/discussion/91633/style-request-netherlands-journal-of-geosciences-geologie-en-mijnbouw#latest

### DIFF
--- a/netherlands-journal-of-geosciences-geologie-en-mijnbouw.csl
+++ b/netherlands-journal-of-geosciences-geologie-en-mijnbouw.csl
@@ -1435,7 +1435,7 @@
       </group>
     </layout>
   </citation>
-  <bibliography hanging-indent="true" et-al-min="8" et-al-use-first="6" et-al-use-last="true" entry-spacing="0" line-spacing="2">
+  <bibliography hanging-indent="true" entry-spacing="0" line-spacing="2">
     <sort>
       <key macro="author"/>
       <key macro="issued-sort" sort="ascending"/>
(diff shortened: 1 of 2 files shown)


adam3smith:
Some quick notes based on examples:
1. Citation should be et-al-min=3
2. No et al. in bibliography
3. No pp. for pages in chapter
4. Lowercase eds. for edited books
5. We're sure about the DOI? I didn't see that mentioned

POBrien333:
Thank you for reviewing.

1. done
2. I have this example from [here ](https://www.cambridge.org/core/services/aop-cambridge-core/content/view/FAF7C70E8F2DAF8DDC18B84B18BF146B/S0016774619000064a.pdf/div-class-title-review-of-induced-seismicity-in-geothermal-systems-worldwide-and-implications-for-geothermal-systems-in-the-netherlands-div.pdf)
**Yeck, W.L., Weingarten, M., Benz, H.M., McNamara, D.E., Bergman, E.A., Herrmann, R.B.,  & Earle, P.S.**, 2016. Far‐field ressurization likely caused one of the largest injection induced earthquakes by reactivating a large preexisting basement fault structure. Geophysical Research Letters, **43**(19), 198–207.
3. done
4. done
5. See the Schoenball citation in above linked sample paper

adam3smith:
Thanks! For 2. I don't understand what the example shows. You currently have
`et-al-min="8" et-al-use-first="6" et-al-use-last="true"` -- I think that should be removed. The paper you link to has at least one reference (Benato et al.) with 8 authors. Your example has 7 authors so wouldn't be affected by this either way.

POBrien333:
Ah, sorry. Copy/paste from .pdf is always crappy.
Yeck, W.L., Weingarten, M., Benz, H.M., McNamara, D.E., Bergman, E.A., Herrmann, R.B., ... & Earle, P.S., 2016. (note the "...")

adam3smith:
ah, got it -- I'd say this is an error in the reference given that Zhang et al. two below it has two more authors (10 vs. Yeck et al's 8) and doesn't use the elipsis. 
See also Sherburn et al. 2015b with 9 authors, Sewell et al. with 8 and Van Wees et al. with 8.

adam3smith:
Thanks!

PR #5627, opened 2021-09-13
Update harvard-swinburne-university-of-technology.csl to meet published spec

kcastellino:
This PR makes several changes so that the CSL format matches the Swinburne Harvard spec correctly:

- Use ampersands in references
- Update documentation link to current URL
- Remove references to Monash University
- Remove `anon` macro, as Swinburne Harvard falls back to the title
- Italicise webpage title
- Remove suffix from months in dates
- Allow plural locators
- Disable hanging indents in bibliography
- Add logic for placement of title based on whether author is known

### DIFF
--- a/harvard-swinburne-university-of-technology.csl
+++ b/harvard-swinburne-university-of-technology.csl
@@ -83,9 +83,12 @@
   <macro name="year-date">
     <choose>
       <if variable="issued">
-        <date variable="issued" form="text">
-          <date-part name="year"/>
-        </date>
+        <choose>
+          <if is-uncertain-date="issued">
+            <text term="circa" form="short" suffix=" "/>
+          </if>
+        </choose>
+        <date variable="issued" form="text" date-parts="year"/>
       </if>
       <else>
         <text term="no date" form="short"/>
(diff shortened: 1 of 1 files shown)


kcastellino:
Hold on merging, I'd like to double-check some of the formatting

kcastellino:
Okay, I think the style is now pretty much correct. The PR test doesn't seem to be generating previews correctly - the references are missing a trailing period after the (bracketed) URL. Zotero adds a period at the end of each entry correctly

harvard-swinburne-university-of-technology.csl
```diff
@@ -167,7 +186,11 @@
           <text macro="author"/>
           <text macro="year-date"/>
         </group>
-        <text macro="title"/>
+        <choose>
+          <if match="any" variable="author editor translator">
```
adam3smith, line 190: Not sure I understand the need for this if here? If the title is used instead of author via substitute, it gets suppressed here anyway

adam3smith (commented):
Thanks. Overall looks great. I do have one question inline

harvard-swinburne-university-of-technology.csl
kcastellino: Oh does it? Sorry, I didn't realise it behaved like that. I'll fix it up

adam3smith:
Great, thanks!

PR #5629, opened 2021-09-13
Change representation of locators

Braintelligence:
The GCEE decided on a different representation of locators


adam3smith:
Thanks!

PR #5632, opened 2021-09-15
Create anglia-ruskin-erhvervsakademi-aarhus.csl

saphyron:
edited Anglia Ruskin style for use on Erhvervsakademi Aarhus


adam3smith:
You just need to move this out of the dependent folder, which you can do by editing the file name

saphyron:
> You just need to move this out of the dependent folder, which you can do by editing the file name

Thanks for the message, found out I made that mistake so closed to pull request and made a new one.

PR #5641, opened 2021-09-28
Update vancouver-imperial-college-london.csl

simomac:
Update to the Vancouver citation style from Imperial College London Library reflecting our latest guidelines.

### DIFF
--- a/vancouver-imperial-college-london.csl
+++ b/vancouver-imperial-college-london.csl
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="UTF-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="expanded" default-locale="en-GB">
   <info>
     <title>Imperial College London - Vancouver</title>
@@ -9,10 +9,14 @@
     <author>
       <name>Sebastian Karcher</name>
     </author>
+    <contributor>
+      <name>S. Mackenzie</name>
+      <uri/>
+    </contributor>
     <category citation-format="numeric"/>
     <category field="medicine"/>
-    <summary>Vancouver style according to the Imperial College London Guidelines</summary>
-    <updated>2014-09-06T16:03:01+00:00</updated>
+    <summary>Vancouver style from Imperial College London Library - 2021</summary>
+    <updated>2021-09-21T11:50:36+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
@@ -30,7 +34,7 @@
   <macro name="author">
     <names variable="author">
       <name sort-separator=" " initialize-with="" name-as-sort-order="all" delimiter=", " delimiter-precedes-last="always"/>
-      <label form="long" prefix=", "/>
+      <label form="short" prefix=", [" suffix="]"/>
       <substitute>
         <names variable="editor"/>
       </substitute>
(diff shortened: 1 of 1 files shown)


vancouver-imperial-college-london.csl
```diff
@@ -256,8 +254,16 @@
   <macro name="journal-location">
     <choose>
       <if type="article-journal article-magazine review review-book" match="any">
-        <text variable="volume"/>
-        <text variable="issue" prefix="(" suffix=")"/>
+        <choose>
```
adam3smith, line 257: this seems unnecessarily complex -- what is it designed to fix/address?

vancouver-imperial-college-london.csl
```diff
@@ -40,24 +44,29 @@
...
     <choose>
-      <if type="report" match="none">
+      <if type="report article article-journal" match="none">
         <group delimiter=": " suffix=";">
           <choose>
             <if type="thesis">
               <text variable="publisher-place" prefix="[" suffix="]"/>
             </if>
-            <else-if type="speech"/>
+            <else-if type="speech article-journal report" match="any"/>
```
adam3smith, line 69: so just to confirm: reports would get neither publisher nor publisher-place? That seems unusual?

vancouver-imperial-college-london.csl
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="UTF-8"?>
```
adam3smith, line 1: revert please

vancouver-imperial-college-london.csl
```diff
@@ -77,10 +85,9 @@
     <choose>
       <if variable="URL DOI" match="any">
         <group delimiter=" ">
-          <text term="available at" text-case="capitalize-first" suffix=":"/>
           <choose>
             <if variable="DOI">
-              <text variable="DOI" prefix="doi:"/>
+              <text variable="DOI" prefix="doi:" suffix="."/>
```
adam3smith, line 90: shouldn't the prefix be "https://doi.org" in line with the guidelines?

adam3smith (changes requested):
Some small change requests and some questions. Thanks!

adam3smith:
@simomac will you be able to make the requested changes? Thanks in advance!

simomac:
Hi Sebastian, working on the changes, amends will be with you by end of today.

Best
Simon

vancouver-imperial-college-london.csl
simomac: Yes, that's been confirmed by our Liason librarians.: '...I think they have a point here, the publisher and place should be included for reports. However the print guide and web guide doesn’t include this information so for consistency I think we should stick with the status quo. When we look at referencing again next year we can decide whether to change this across the board....'

vancouver-imperial-college-london.csl
simomac: No, this should stay just as ‘doi’, we don’t include the https: part

vancouver-imperial-college-london.csl
simomac: This is designed to distinguish the issue number , where it exists, from the volume by placing it in parenthesis

vancouver-imperial-college-london.csl
simomac: done

vancouver-imperial-college-london.csl
adam3smith: But that's not what it's doing: the issue is already in parentheses with the previous code. For some reason you're checking if volume and issue are numbers? That seems particularly problematic for issue, which you also see as "Fall", e.g.

vancouver-imperial-college-london.csl
adam3smith: FWIW, your guidance is contradictory here. You have some examples with doi: but also 


And doi: is discouraged by CrossRef and the DOI foundation. Any chance you could change/unify that in the guidance?

adam3smith:
Thanks! Still have questions about two of these

vancouver-imperial-college-london.csl
simomac: Hi, this has been simplified.

vancouver-imperial-college-london.csl
simomac: changed o https/..... as requested

adam3smith:
Thanks!

PR #5651, opened 2021-10-05
harvard-university-of-bath.csl: correct et-al-min

alex-ball:
- Citations should abbreviate to “et al.” starting with four authors rather than three
- Tidy groupings for graphics/maps


adam3smith:
Thanks!

PR #5656, opened 2021-10-07
update style file

USITC-Zotero:
We needed to fix a bug where magazine author names appeared in bibliographies incorrectly.


adam3smith:
Thanks!

PR #5661, opened 2021-10-14
Corrections for Conservation Biology style

SR--:
1. book editors
2. DOI for online articles

Forum discussion is [here](https://forums.zotero.org/discussion/11440/style-request-conservation-biology#latest)


adam3smith:
Thanks!

Issue #5608, opened 2021-08-25
Style Request: [wikipedia markup templates]


sallybaker:
Wikipedia has its own citation standards available here: https://en.wikipedia.org/wiki/Wikipedia:Citing_sources that are currently not available via CSL Styles.

fgnievinski:
code could be borrowed from [Citer](https://github.com/5j9/citer), a citation generator tool for Wikipedia.

fgnievinski:
code could also be borrowed from [Zotero](https://www.zotero.org/support/kb/zotero_and_wikipedia)

adam3smith:
(we can't copy code in other programming languages. CSL has a very distinct structure - and is much more limited than full programming languages like  javascript or python - which is also cause for some of our concerns in implementing this)

adam3smith:
Closed via the PR linked above

fgnievinski:
in citations of book sections, chapter title and book title seem swapped.

PR #5662, opened 2021-10-14
wikipedia generic citation

sallybaker:
Initial brief version to address #5608.

### DIFF
--- a/wikipedia-templates.csl
+++ b/wikipedia-templates.csl
@@ -19,7 +19,7 @@
   <!-- Top level macro which defines what bibliographical format will be returned -->
   <macro name="citations">
     <choose>
-      <if type="book chapter">
+      <if type="book chapter" match="any">
         <text macro="book-citation"/>
       </if>
       <else-if type="article-journal">
@@ -86,7 +86,7 @@
   </macro>
   <macro name="date">
     <choose>
-      <if type="book article-journal chapter">
+      <if type="book article-journal chapter" match="any">
         <date variable="issued" prefix=" | date=">
           <date-part name="day" form="numeric"/>
           <date-part name="month" form="long" prefix=" "/>
--- a/wikipedia-templates.csl
+++ b/wikipedia-templates.csl
@@ -173,7 +173,7 @@
   <!-- Bibliography -->
   <bibliography>
     <sort>
-      <key macro="author-statement"/>
+      <key macro="contributor-statements"/>
       <key macro="date"/>
     </sort>
     <layout>
--- a/wikipedia-templates.csl
+++ b/wikipedia-templates.csl
@@ -10,15 +10,68 @@
     <link href="http://www.zotero.org/styles/wikipedia-templates" rel="self"/>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
     <link href="https://en.wikipedia.org/wiki/Template:Citation" rel="documentation"/>
+    <link href="https://en.wikipedia.org/wiki/Template:Cite_book" rel="documentation"/>
+    <link href="https://en.wikipedia.org/wiki/Template:Cite_journal" rel="documentation"/>
     <category field="generic-base"/>
     <category citation-format="numeric"/>
     <updated>2021-10-14T13:30:00+10:00</updated>
-   </info>
+  </info>
+  <!-- Top level macro which defines what bibliographical format will be returned -->
+  <macro name="citations">
+    <choose>
+      <if type="book chapter">
+        <text macro="book-citation"/>
+      </if>
+      <else-if type="article-journal">
+        <text macro="journal-citation"/>
+      </else-if>
+      <else>
+        <text macro="generic-citation"/>
+      </else>
+    </choose>
+  </macro>
+  <!-- Inline citations are bibliographical entries wrapped in 'ref' tags -->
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks! So my biggest concern here are the authors. Looking at https://en.m.wikipedia.org/wiki/Template:Cite_book just using "author" isn't valid and it makes sense to me to split into last/first, but especially with the numbering, I don't think CSL can do this. Which does bring me back to my original concern: is it a good idea for is to provide a low-quality output format? Wouldn't it make more sense to steer users towards tools able to produce complete and correct output?

adam3smith:
> The state library services all of Queensland, so our audience is also school kids and amateur researchers who simply wouldn't have use for more professional tools - we're trying to fill the gap :)

To start with this: I don't mean "more professional" -- I see why people may not want to install and use Zotero, say -- but this would perform  worse than the cite function in the Visual editor (see https://en.wikipedia.org/wiki/Wikipedia:VisualEditor ) which just requires pasting a URL, DOI, or ISBN and automatically generates and inserts the template, so would seem ideal for amateur researchers and school kids. (I was actually hoping the it'd work with your permalinks, but that's unfortunately not the case, likely because those pages are built dynamically).

For the numbers, I wasn't referring to page numbers, I think we can handle those (CSL has `page` which refers to a page range and `number-of-pages` which refers to the total number of pages), but to the numbers that should be displayed with authors. If you take e.g. the Fenner et al. reference, correct template would be:

```
{{Cite journal|
 doi = 10.1038/s41597-019-0031-8|
 issn = 2052-4463|
 volume = 6|
 issue = 1|
 pages = 28|
 last1 = Fenner|
 first1 = Martin|
 last2 = Crosas|
 first2 = Mercè|
 last3 = Grethe|
 first3 = Jeffrey S.|
 last4 = Kennedy|
 first4 = David|
 last5 = Hermjakob|
 first5 = Henning|
 last6 = Rocca-Serra|
 first6 = Phillippe|
 last7 = Durand|
 first7 = Gustavo|
 last8 = Berjon|
 first8 = Robin|
 last9 = Karcher|
 first9 = Sebastian|
 last10 = Martone|
 first10 = Maryann|
 last11 = Clark|
 first11 = Tim|
 title = A data citation roadmap for scholarly data repositories|
 journal = Scientific Data|
 accessdate = 2019-04-27|
 date = 2019-04-10|
 url = http://www.nature.com/articles/s41597-019-0031-8}}
```
 there's no way CSL can handle the consecutive numbering. Without it, multiple authors are ambiguous.

To be clear, I'm not dead set against this and certainly open to be convinced -- the goal of making it easier to insert references into WP is certainly worthwhile

sallybaker:
Yeah, I get what you mean about 'bad tools' :) 

So two options for the authors: 
1. 'authors' is still a valid, although discouraged, parameter for journal articles, so leave as is  (https://en.wikipedia.org/wiki/Template:Cite_journal#Authors) 
2. implement family and given names with the prefixes lastx & firstx plus an embedded note to add numbering. So it would return: 
`| lastx= | firstx= | lastx= | firstx=` 
`Note: please make sure you update author numbering before including on wikipedia. i.e. last1 | first1 | last2 | first2 | last3 ...`

adam3smith:
I don't think the x-option is acceptable user experience. If you're intending this in part for school kids, it has to just work; they can't be expected to meticulously edit within the template.
`authors` is an option and it looks like the citations at least come out fine.

I'd be interested to hear from experienced Wikipedians like @pigsonthewing (feel free to tag/pull in others). If they think this is OK/worthwhile, I'm happy to facilitate it. If they're concerned this is going to cause more problems than it solves within WP, I think I'd be inclined to follow their advice.

pigsonthewing:
> 
> One last question: is `author` or `authors` the better tag?

The plural: "authors: Free-form list of author names" (vs. "author: this parameter is used to hold the complete name of a single author (first and last) or to hold the name of a corporate author. This parameter should never hold the names of more than one author.")

adam3smith:
Great, thanks! @sallybaker -- so we can move along here, but I'd like a minimal viable first iteration of this that at least handles journal articles and books in addition to generic material. Lmk if you run into any issues.

fgnievinski:
consider adopting the "vauthors" parameter (for comma-separated list of author names in Vancouver style) instead of free-form "authors", because its parsing is already supported on the Wikipedia side (see [here](https://en.wikipedia.org/wiki/Module:Citation/CS1)); "vauthors" was adopted in the [Wikipedia template filling tool](https://citation-template-filling.toolforge.org/cgi-bin/index.cgi) (see [example](https://citation-template-filling.toolforge.org/cgi-bin/index.cgi?ddb=&type=pubmed_id&id=30971690)).

adam3smith:
Thanks for this. Looking good to me now. Have you tested the output in the Wikipedia sandbox? That's the last thing I'd want to do before merging, but I'll also take someone else's testing

sallybaker:
@adam3smith, plugged the CSL code editor output into the wikipedia sandbox - preview looks pretty good to me:
Screen Shot 2021-11-01 at 9 18 08 am

adam3smith:
Yeah, that looks very good indeed, thanks for your work on this.

adam3smith:
OK -- thanks also to @pigsonthewing  and @fgnievinski  for your input on this.
This is available for distribution immediately (and will e.g. show up on zotero.org/styles within 1h), but you'll have to check with ExLibris support on when/how often they update styles; we have no influence on that.

PR #5668, opened 2021-10-19
Add composer.json

hxdef:
Hello everyone!

Since the deployment process has been updated (the styles-distribution repo is [deprecated](https://github.com/citation-style-language/styles/blob/v1.0.1/README.md#versioning-and-style-distribution)), I decided to open a new pull request. Here is a link to the [old one](https://github.com/citation-style-language/styles/pull/2706).
The main purpose of these changes is adding composer.json file to publish on the [packagist service](https://packagist.org/). To make this the actions workflow scripts have been updated. Action steps include:
- synchronizing of composer.json file with the release branch
- bumping version of csl files and push tags which will be used to release package on packagist

We added the [composer.json from the old pull request](https://github.com/citation-style-language/styles/blob/f231c6bc514752ce0bd933aab94592f886f62cd4/composer.json).
Also, I think it would be great to have a release branch name in the separate environment variable.

Thank you


hxdef:
Hello @adam3smith and @retorquere 

Thanks for the information and leading questions! I optimized some my updates.
Let me answer on your questions:

> Do you understand why that test is failing? I would have thought your last commit takes care of that?

It happens because of the composer.json file was added to the repository, but test allows only csl-files. Therefore, composer.json should be excluded from checking. This can be done by adding the corresponding entry to EXTRA_FILES section in filters.yaml.
The main question is how to add an exception in a proper way? Should this be added to the master branch before the rest of the changes?

> Also, we need composer in the master branch? It couldn't just sit in the distribution (i.e. 1.0.1) one?

To create a package on packagist.org it's important to have composer.json in the master branch. This doesn't work with other branches.

> 1. Why does it checkout out the repo again? The top of the workflow already does a checkout.

That was an unnecessary logic. I got rid of this in the current version.

> 2. A variable for the branch name would be an option, but it's already set centrally as steps.release.outputs.branch

Thanks for info. I wanted to use this in a previous version but didn't find the variable I needed.

> 3. as the CI workflows are propagated to the locale repo, they should check for if: github.repository == 'citation-style-language/styles' so they're not also ran on the locales repo.

Actually composer.json file is needed in the locales repository too. So no extra condition is required. I nothiced that there is the related request in this PR - https://github.com/citation-style-language/locales/pull/161

> It looks like the sheldon workflow also does something with composer?

That was unnecessary changes. I removed this logic in the current version.

adam3smith:
@retorquere this now looks good to me. Any concerns from your end? Otherwise I'd merge

retorquere:
I'm still not sure why a separate `update_composer` step is needed alongside the existing `update` step. I'd prefer these things remain centralized unless there is reason not to.

`actions/create-release` is unmaintained and the project suggest some others to use that are maintained. Do we have assets that we could attach to a release? Otherwise I don't see much difference between tagging a release and creating a GH release, but there could well be other benefits I'm missing.

hxdef:
Hello @adam3smith and @retorquere 

> I'm still not sure why a separate update_composer step is needed alongside the existing update step. I'd prefer these things remain centralized unless there is reason not to.

I just thought to separate this logic to make it more obvious for understanding. But you're right, we can combine it with already existed step ("Check for relevant changes"). See my last updates.

> actions/create-release is unmaintained and the project suggest some others to use that are maintained. Do we have assets that we could attach to a release? Otherwise I don't see much difference between tagging a release and creating a GH release, but there could well be other benefits I'm missing.

Yes. There are assets that can be attached to releases. I replaced actions/create-release with [softprops/action-gh-release](https://github.com/softprops/action-gh-release). It seems to be one of the most popular at the moment.

adam3smith:
@retorquere this looks good to me & I'd love to see this merged after years of waiting (counting the superseded repo) -- but my ability to read CI/github-action code is limited -- could you check over this one more time & see if you have remaining major concerns?

retorquere:
Looks good to me too

PR #5673, opened 2021-10-20
Create tyndale-bulletin-fullnote-bibliography

dstark:
Create style for the Tyndale Bulletin.


dstark:
@rmzelle The file I've submitted to create this new style passes CSL validation, but I'm obviously doing something wrong in the submission process as the checks don't all pass. But I'm not sure what I'm being told is wrong on the styles/locales front. Could I trouble you for some pointing in the right direction?

Thanks so much!

dstark:
> They just have that one style, right? In that case we'd go without nthe fullnote, bibliography in title and id Also, do they have an issn?

Yes. Just the one style, and I've submitted a revised file that includes the (e)issn information.

tyndale-bulletin-fullnote-bibliography.csl
```diff
@@ -0,0 +1,1054 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="chicago" default-locale="en-GB">
+  <info>
+    <title>Tyndale Bulletin (full note, bibliography)</title>
```
adam3smith, line 4: Sorry, wasn't clear here: if there's just one style, this should just be `Tyndale Bulletin`

tyndale-bulletin-fullnote-bibliography.csl
```diff
@@ -0,0 +1,1054 @@
...
+    <id>http://www.zotero.org/styles/tyndale-bulletin-fullnote-bibliography</id>
```
adam3smith, line 5: This should be 
```suggestion
    <id>http://www.zotero.org/styles/tyndale-bulletin</id>
```
etc. for filename, and self link

adam3smith (changes requested):
Could you still rename as indicated? Rest looks great, thanks!

adam3smith:
Thanks!

dstark:
@adam3smith thanks for approving the style. My apologies I misread the earlier post as saying _if_ it was decided to drop the fullnote-bibliography terminology from the style name and such rather than that _was_ what was to be done. ;-)

Appreciate your help.

PR #5678, opened 2021-10-25
Update bluebook-inline.csl

michallahav:
There is an error using all 3 bluebook csl files.
A user has suggested the following addition:
"The bluebook csl file included the format for citation but not bibliography; I added bibliography and the error is gone."
https://forum.pkp.sfu.ca/t/bluebook-citation-style/55204/6
Worth a try?
Thanks,
Michal

### DIFF
--- a/bluebook-inline.csl
+++ b/bluebook-inline.csl
@@ -227,4 +227,37 @@
       </choose>
     </layout>
   </citation>
+  <bibliography et-al-min="3" et-al-use-first="1">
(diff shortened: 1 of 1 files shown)


adam3smith:
Styles without bibliography are perfectly valid and Bluebook doesn't define one.
I'd assume this is an issue with citeproc-php (that's what OJS uses, right?) and should be reported and fixed there

Issue #5680, opened 2021-10-26
Style Request: [AMA 9th edition]


meeshazee:
Aoa i want AMA 9th edition citation for this journal
https://meridian.allenpress.com/angle-orthodontist/pages/how-to-submit
eISSN 1945-7103 ISSN 0003-3219

examples:
Journal article: Campbell JL, Pedersen OK.The varieties of capitalism and hybrid success.Comp. Polit. Stud. 2007;40:307-332
Book: Mares I. Firms and the welfare state: When, why, and how does social policy matter to employers? New York. Oxford University Press; 2001.

POBrien333:
I wouldn't call that a high-impact journal, but I had a go at the style. When it closes (it will say "merged") you'll be able to find it on the style repository.

PR #5685, opened 2021-10-26
Create the-angle-orthodontist.csl

POBrien333:
Closes https://github.com/citation-style-language/styles/issues/5680

### DIFF
--- a/the-angle-orthodontist.csl
+++ b/the-angle-orthodontist.csl
@@ -13,7 +13,7 @@
     <category field="medicine"/>
     <issn>0003-3219</issn>
     <eissn>1945-7103</eissn>
-    <updated>2021-10-26T10:26:58+00:00</updated>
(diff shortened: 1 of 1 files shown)


the-angle-orthodontist.csl
```diff
@@ -0,0 +1,233 @@
...
+          </group>
+          <text variable="page" prefix=", " suffix=" "/>
+          <group prefix="(" suffix=")." delimiter=" ">
+            <text variable="authority"/>
+            <date variable="issued">
+              <date-part name="year"/>
+            </date>
+          </group>
+        </else-if>
+        <else-if type="webpage post post-weblog" match="any">
```
adam3smith, line 178: If they have their own condition, these should probably get a URL? I'd use current bAMA as a template

adam3smith (commented):
One quick Q

adam3smith:
Thanks!

PR #5686, opened 2021-10-27
journal-of-palm-oil-research.csl fix several errors

POBrien333:
via https://forums.zotero.org/discussion/comment/392419/#Comment_392419


adam3smith:
Changes look great, but while we're at it, could you also update the author macros to include a substitute element with editor and title?

adam3smith:
Thanks!

PR #5697, opened 2021-11-01
Matter

pooyanmakvandi:
Please add the style of Matter (cell press, Elsevier). It is similar to the style of Nature


adam3smith:
Same or similar? And if you are asking for that to be created, a pull request is the wrong way. See https://github.com/citation-style-language/styles/blob/master/REQUESTING.md#requesting-csl-styles

adam3smith:
Closing for inactivity. Happy to reopen if my above queries are answered

PR #5701, opened 2021-11-02
Update biomed-central.csl

### LINKED ISSUE #5700: Question about biomed central URL/DOI display
The biomed-central style displays a link to the DOI, but only if the URL variable is set. This is confusing, because entries that have a DOI but no URL will never have their DOI linked.  It's not clear to me why the url variable must be set in order for DOIs to render. 

The code in question is here:

https://github.com/citation-style-language/styles/blob/1efce81f8a8bd18d97a46f1beda41d5ca8d9e362/biomed-central.csl#L90-L95

POBrien333:
Closes #5700 

Guidelines here: https://bmcbioinformatics.biomedcentral.com/submission-guidelines/preparing-your-manuscript/research-article

### DIFF
--- a/biomed-central.csl
+++ b/biomed-central.csl
@@ -12,10 +12,13 @@
     <contributor>
       <name>Sebastian Karcher</name>
     </contributor>
+    <contributor>
+      <name>Patrick O'Brien</name>
+    </contributor>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks!

PR #5706, opened 2021-11-05
Update iso690-author-date-fr-no-abstract.csl

pagiraud:
Hello,
Here I propose an update to take into account all the improvements from the iso690-author-date-fr style, made by @grolimur  and alii since 2013.

As a reminder, iso690-author-date-fr-no-abstract is only a fork/copy of iso690-author-date-fr truncated of lines concerning abstracts and notes (yes, this is a very trivial fork).

Thanks to all for maintaining this repo over the years.


adam3smith:
Makes sense, thanks!

PR #5708, opened 2021-11-08

jvandermolen:
Added speech format, fixed text-case for chapter titles (book-container macro), and added page range format


adam3smith:
Great, thanks!

PR #5710, opened 2021-11-08
Update thieme-german.csl

POBrien333:
Bring in line with most recent guidelines from 2019: 
https://www.thieme.de/statics/dokumente/thieme/final/de/dokumente/sw_%20autorenlounge/ARL_Wiss_Publikation.pdf


adam3smith:
Thanks!

PR #5712, opened 2021-11-09
Bch corr

NMonteix:
Small correction on the place to indicate the series


bulletin-de-correspondance-hellenique.csl
```diff
@@ -79,7 +79,7 @@
       </group>
     </macro>
     <macro name="page-num">
-      <group delimiter=" ">
+      <group delimiter=" ">
```
adam3smith, line 82: did you mean to make changes to the spaces here? I can't figure out how they're different -- they both seem like regular spaces, not nbsp

NMonteix: It was not my intent, anyway I have inserted back the nbsp if they had disappeared and remove a two spaces sequence
It should be fine now

PR #5715, opened 2021-11-10
Update journal-of-computer-applications-in-archaeology.csl

RonaldVisser:
The style was not completely in concordance with the description on https://journal.caa-international.org/about/submissions/, amongst others space between the initials and the use of full stops.


adam3smith:
Thanks!

PR #5718, opened 2021-11-10
Create asian-myrmecology.csl

POBrien333:
via https://forums.zotero.org/discussion/92564/asian-myrmecology-citation-style-request#latest

### DIFF
--- a/asian-myrmecology.csl
+++ b/asian-myrmecology.csl
@@ -1,0 +1,215 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" demote-non-dropping-particle="sort-only" version="1.0" class="in-text" default-locale="en-GB">
+  <info>
+    <title>Asian Myrmecology</title>
+    <id>http://www.zotero.org/styles/asian-myrmecology</id>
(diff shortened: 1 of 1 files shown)


asian-myrmecology.csl
```diff
@@ -0,0 +1,215 @@
...
+    </names>
+  </macro>
+  <macro name="author">
+    <names variable="author">
+      <name initialize-with="" sort-separator=" " name-as-sort-order="all" delimiter-precedes-last="always" et-al-min="4"/>
+      <label prefix=", " form="short"/>
+      <substitute>
+        <names variable="editor"/>
+        <names variable="translator"/>
+      </substitute>
```
adam3smith, line 46: Could you add title to substitute here and in author-short? See webpage on preview, which doesn't look right. (Otherwise, the style guide is a hot mess, so this looks like the most reasonable implementation...)

adam3smith (changes requested):

asian-myrmecology.csl
POBrien333: Done. Thanks for reviewing. :)

adam3smith:
Thanks!

PR #5719, opened 2021-11-10
Modify american-journal-of-archaeology.csl

Jmuccigr:
Correcting to follow cited style guide.

### DIFF
--- a/american-journal-of-archaeology.csl
+++ b/american-journal-of-archaeology.csl
@@ -15,7 +15,7 @@
     <issn>0002-9114</issn>
     <eissn>1939-828X</eissn>
     <summary>The American Journal of Archaeology style</summary>
-    <updated>2017-04-13T16:17:59+00:00</updated>
+    <updated>2021-11-12T14:22:00+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale>
@@ -32,7 +32,7 @@
             <if variable="author">
               <names variable="editor">
                 <label form="verb-short" prefix=" " text-case="capitalize-first" suffix=" "/>
-                <name and="text" delimiter=", "/>
+                <name and="text" delimiter=", " suffix=". "/>
               </names>
             </if>
           </choose>
@@ -266,8 +266,15 @@
   </macro>
   <macro name="publisher">
     <group delimiter=": ">
-      <text variable="publisher-place"/>
-      <text variable="publisher"/>
+      <choose>
+        <if type="thesis" match="none">
+          <text variable="publisher-place" prefix=". "/>
+          <text variable="publisher"/>
+        </if>
+        <else>
+          <text variable="publisher"/>
+        </else>
+      </choose>
     </group>
   </macro>
   <macro name="date">
@@ -287,7 +294,14 @@
   </macro>
   <macro name="event">
     <group>
-      <text term="presented at" suffix=" "/>
+      <choose>
+        <if variable="genre" match="none">
+          <text term="presented at" suffix=" " text-case="capitalize-first"/>
+        </if>
+        <else>
+          <text term="presented at" suffix=" " text-case="lowercase"/>
+        </else>
+      </choose>
       <text variable="event"/>
     </group>
   </macro>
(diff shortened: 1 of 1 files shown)


Jmuccigr:
Should have said "remove day-month at start of biblio entry" above. The year stays of course.

adam3smith:
Is this ready for review or are you planning to make further edits?

Jmuccigr:
> Is this ready for review or are you planning to make further edits?

Yeah, I think it's good now, after making one more tweak to the journal article biblio format. And thanks for checking. I realize I keep editing it as I see mistakes in my outputted biblio. I'm sure there are other minor things, but let's go ahead with this.

american-journal-of-archaeology.csl
```diff
@@ -320,14 +337,13 @@
         <text macro="day-month" prefix=", "/>
       </else-if>
       <else>
-        <group prefix=". " delimiter=", ">
+        <group prefix="" delimiter=", ">
```
adam3smith, line 340: I don't understand the problem with the period in the prefix here? I'm seeing a period before Ph.D. diss in the style guide and the solution with `text value` above is really awkward.

american-journal-of-archaeology.csl
```diff
@@ -198,7 +198,7 @@
     <choose>
       <if type="article-journal">
         <text variable="volume" prefix=" "/>
-        <text variable="issue" prefix=" "/>
+        <text variable="issue" prefix=" (" suffix=")"/>
```
adam3smith, line 201: where is that from? I'm actually not seeing issue numbers at all in the style guide

american-journal-of-archaeology.csl
```diff
@@ -300,7 +317,7 @@
       <if variable="title" match="none"/>
       <else-if type="thesis"/>
```
adam3smith, line 318: wouldn't you get duplicate genre here for presentations and other items with variable `event` if you don't exclude it?

american-journal-of-archaeology.csl
```diff
@@ -300,7 +317,7 @@
...
       <else>
-        <text variable="genre" text-case="capitalize-first" prefix=". "/>
```
adam3smith, line 303: If not from here, where are you getting the period from?

american-journal-of-archaeology.csl
```diff
@@ -356,14 +371,13 @@
...
-      <text macro="day-month" text-case="capitalize-first" suffix=". "/>
       <text macro="title"/>
       <text macro="description"/>
-      <text macro="secondary-contributors" prefix=". "/>
+      <text macro="secondary-contributors" prefix=" "/>
       <text macro="container-title"/>
       <text macro="container-contributors" prefix=", "/>
       <text macro="locators-chapter"/>
-      <text macro="edition" prefix=". "/>
+      <text macro="edition" prefix=". " strip-periods="true"/>
```
adam3smith, line 380: The strip-periods here is a bit odd, especially when you think of non-numeric entries like "2nd rev. and ext. edition" -- what periods are you trying to remove?

adam3smith (commented):
A couple of questions/comments inline

american-journal-of-archaeology.csl
Jmuccigr: No, that's correct, it is missing from the style guide. I'm looking at bibliography in articles from the journal for that. For example:

Screen Shot 2021-11-15 at 10 11 14

american-journal-of-archaeology.csl
Jmuccigr: Because the locators macro that follows it uses a period as a prefix to books, so you can get two periods in a row if the edition ends with one, as in your example. 

IF there's a better way to handle that, great.

american-journal-of-archaeology.csl
Jmuccigr: I believe this is because it is preceded by the title of the dissertation which already ends with a period, so if you include it here too, you get things like `"Title of my dissertation." . Ph.D., Alma Mater University.` .

If there's a better way to handle it, great.

american-journal-of-archaeology.csl
Jmuccigr: I believe it's at the end of the title, which precedes the event, which is itself part of the issue.

Jmuccigr:
Is there more to be done here? @adam3smith?

adam3smith:
Nothing to do, but I do need some time to test out some of the issues I flagged & you responded to

Jmuccigr:
> Nothing to do, but I do need some time to test out some of the issues I flagged & you responded to

Sounds good. Thanks.

adam3smith:
Sorry for the delay. The style is a bit of a mess (prior to your edits) with prefixes and suffixes all over and I do worry that'll cause some weit behavior in edge cases, but your edits work fine and it would be a ton of work to clean up, so taking this as is.
Thanks!

PR #5720, opened 2021-11-11
 Create steel-research-international.csl

c-schubert:
First version for a Citation Style Language (csl) file for the journal "Steel Research International":

The sadly very poorly defined specifications are taken from here (reference section):
https://onlinelibrary.wiley.com/page/journal/1869344x/homepage/author-guidelines

The style file should work fine for the most common reference types. There are some parts which may need manual fine tuning of a paper's reference section in the end, before committing the paper. Nevertheless, this style should still be a work saver.

In particular, there are some difficulties i.e. for patents or preprint specifications and the variables that the CSL offers (nor preprint type or company variable available), which probably cannot be fixed right now (perfectly) and need manual correction.
Also, I guess there is no way to automatically use or define journal or conference abbreviations in the csl file at the moment?

### DIFF
--- a/steel-research-international.csl
+++ b/steel-research-international.csl
@@ -1,0 +1,222 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>Steel Research International</title>
+    <title-short>Steel Res. Int.</title-short>
+    <id>http://www.zotero.org/styles/steel-research-international</id>
+    <link href="http://www.zotero.org/styles/steel-research-international" rel="self"/>
+    <link href="https://onlinelibrary.wiley.com/page/journal/1869344x/homepage/author-guidelines" rel="documentation"/>
+    <author>
+      <name>Christian Schubert</name>
+    </author>
+    <contributor>
+      <name>Christian Schubert</name>
+      <email>5238962+c-schubert@users.noreply.github.com</email>
+    </contributor>
+    <category citation-format="author-date"/>
+    <category field="engineering"/>
+    <eissn>1869-344X</eissn>
+    <issnl>1611-3683</issnl>
+    <summary>
+    Citation style for the Steel research international
+    
+    There are some difficulties i.e. with patents, preprint  specifications and the variables that the CSL offers (nor preprint type or company variable availably), which probably cannot be fixed right now (perfectly) and need manual correction.
+    Also there is no way to automatically use or define journal or conference abbrevitation in the csl file right now?
+    </summary>
+    <updated>2021-11-10T16:22:30+01:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+<locale xml:lang="en-US">
+    <date form="text" delimiter=" ">
+      <date-part name="month" form="short" strip-periods="true"/>
+      <date-part name="year" font-weight="bold"/>
+    </date>
+    <terms>
+      <term name="chapter" form="short">Ch.</term>
+      <term name="editor" form="short">eds</term>
+    </terms>
+  </locale>
+  <macro name="author">
+    <names variable="author">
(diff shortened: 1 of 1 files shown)


steel-research-international.csl
```diff
@@ -0,0 +1,220 @@
...
+    </names>
+  </macro>
+  <macro name="title">
+    <choose>
+      <if type="book">
+        <text variable="title" font-style="italic"/>
+      </if>
+      <else-if type="chapter">
+        <text term="in" suffix=" "/>
+        <text variable="title" font-style="italic"/>
```
adam3smith, line 53: This can't be right. Should this be container-title?

steel-research-international.csl
```diff
@@ -0,0 +1,220 @@
...
+      <date-part name="year" font-weight="bold"/>
+    </date>
+    <terms>
+      <term name="chapter" form="short">Ch.</term>
+      <term name="editor" form="short">eds</term>
+    </terms>
+  </locale>
+  <macro name="author">
+    <names variable="author">
+      <name sort-separator=" " initialize-with="." delimiter=", " delimiter-precedes-last="always"/>
```
adam3smith, line 40: You'll want to add a label here (which would appear for editors)

steel-research-international.csl
```diff
@@ -0,0 +1,220 @@
...
+        <text term="in" suffix=" "/>
+        <text variable="title" font-style="italic"/>
+      </else-if>
+      <else-if type="paper-conference">
+        <text term="in" suffix=" "/>
+        <text variable="container-title" font-style="italic"/>
+      </else-if>
+      <else-if type="speech">
+        <text term="presented at" suffix=" "/>
+        <text variable="title" font-style="italic"/>
```
adam3smith, line 61: Same here: do you want event? Title would be the title of the presentation

steel-research-international.csl
```diff
@@ -0,0 +1,220 @@
...
+            <group delimiter=" " prefix=", ">
+              <text variable="URL"/>
+            </group>
+            <text macro="accessed-date"/>
+          </if>
+        </choose>
+      </if>
+      <else-if variable="DOI">
+        <group font-style="normal" text-decoration="none">
+          <text variable="DOI" prefix=", "/>
```
adam3smith, line 118: No prefix like doi: or https://doi.org?

adam3smith (changes requested):
Thanks! A couple of questions/requests.
Preprint is CSL `article`
By company you mean the patent assignee? 
And no custom abbreviations, no.

steel-research-international.csl
c-schubert: Yes I probably do. There is no real default for this case. So for now I use what seems to make sense to me and that would be to use `Eds:` as a prefix, as it is also given when author and editor are specified.

steel-research-international.csl
c-schubert: This works as aspected for my zotero test entry:

image
image

so I guess `title` is fine?

steel-research-international.csl
c-schubert: Indeed, here `event` should be used.

steel-research-international.csl
c-schubert: From the template I think `DOI:` should be used.

steel-research-international.csl
adam3smith: As expected how? The code.produces "in: (chapter) title" which makes no sense for a book section, since "in" indicates the book/container an item was published in.

steel-research-international.csl
adam3smith: No, don't prefix since this starts the bibliography entry and would look like it's out of alphabetical order. Just using label after `name`, either in parentheses or preceded by a comma, works

steel-research-international.csl
c-schubert: Ah OK, sorry my bad `container-title` is totally right. I oversaw that there is an extra booktitle option in zotero ...

steel-research-international.csl
c-schubert: There should be no alphabetical order in the bibliography, instead it should be in citation order, so this should be no problem. However, I will change it as you proposed (I hope I got it right..), because it looks little more consistent.

adam3smith:
Great, thanks!

PR #5725, opened 2021-11-13
Add original-date extra to proper place as per SBL Handbook 6.2.17-18

dhacker29:


adam3smith:
Could you please revert the medium changes? They break thesis and interview citations (at a minimum, maybe more). We can talk about how to best achieve the ebook citations as per 6.2.25, but let's keep this to the original date so it gets merged quickly.

adam3smith:
Just as an FYI - the reason nIm not merging this is the changed delimiter between publisher and year -- this may actually be a regression introduced by this change, or the preview is off, but I need some time to take a closer look

society-of-biblical-literature-fullnote-bibliography.csl
```diff
@@ -877,6 +917,12 @@
         <group prefix=" (" suffix=")">
           <group delimiter=", ">
             <group delimiter=" ">
+          <text macro="event-note"/>
+          <group delimiter="; ">
+            <text macro="originally-published"/>
+            <group delimiter=", ">
+              <text macro="reprint-note"/>
+              <text macro="publisher"/>
```
adam3smith, line 925: Yeah, so by moving the publisher down here into the group, you get the year preceded by a semicolon rather than a comma for items without original date, which is incorrect. That's just for notes, it's fine for the bibliography

dhacker29: I have reworked the commit to fix the issue.

adam3smith:
@retorquere the test here shouldn't fail
1. I don't think we even want this to run for styles without default locale or with an en default locale
2. Styles can have other info in parentheses

retorquere:
That would mean the only test that remains is that if a default-locale is specified, it must be in the locales repo. Correct?

adam3smith:
No, the second test should be:
If default-locale AND !default-locale starts-with en- 
--> title should include natural language version of default-locale in parentheses

There are a couple of cases where we just give the language, not the country, so if we could also allow that?

retorquere:
> No, the second test should be: If default-locale AND !default-locale starts-with en- --> title should include natural language version of default-locale in parentheses
> 
> There are a couple of cases where we just give the language, not the country, so if we could also allow that?

Both done

adam3smith:
Great, thanks!

PR #5726, opened 2021-11-14
haaga-helia-university-of-applied-sciences-harvard.csl: no date cites need to be unique

mrautio-p:
Make cites unique when citing references from the same author without known issuing date.
Example of unique citing of four different references: (Musk s.a.a; Musk s.a.b; Github s.a.; Musk 2006; Musk 2007a; Musk 2007b)

### DIFF
--- a/haaga-helia-university-of-applied-sciences-harvard.csl
+++ b/haaga-helia-university-of-applied-sciences-harvard.csl
@@ -13,7 +13,7 @@
     <category citation-format="author-date"/>
     <category field="generic-base"/>
     <summary>Haaga-Helia University of Applied Sciences referencing style (Finnish and English)</summary>
-    <updated>2021-02-21T03:51:30Z</updated>
+    <updated>2021-11-14T07:02:42+00:00</updated>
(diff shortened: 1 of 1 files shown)


haaga-helia-university-of-applied-sciences-harvard.csl
```diff
@@ -167,6 +167,8 @@
       </else-if>
       <else>
         <text term="no date"/>
+        <!-- no date cites need to be unique -->
+        <text variable="year-suffix" prefix=" "/>
```
adam3smith, line 171: If you add year suffix explicitly in one place, you also have to explicitly add it to the actual date.

adam3smith (changes requested):

haaga-helia-university-of-applied-sciences-harvard.csl
mrautio-p: removed the attribute. the prefix attribute was incorrect and shouldn't be there in the first place. thanks for noticing it!

mrautio-p:
@adam3smith made the requested change. the prefix attribute was incorrect

adam3smith:
Sorry you misunderstood my comment: once you set year-suffix explicitly on the no date term, you also need it on the year. Otherwise you won't get 2001a, 2001b anymore

PR #5728, opened 2021-11-15
remove et-al from bibtex.csl

pstjohn:
Is there a need to use et-al truncation in the bibtex csl? Typically the bibtex formatter would specify whether and how the author list should be truncated, and the current implementation leaves bibtex assuming that the final author is "et al", resulting in citations like

Firstauthor, F., Secondauthor, S., & al., et. (2021). Title. *Journal.* *1*(1)

Specifically, the crossref api seems to use this style sheet to return bibtex data, and it would be great not to truncate the data coming from their service:
https://www.crossref.org/labs/citation-formatting-service/

### DIFF
--- a/bibtex.csl
+++ b/bibtex.csl
@@ -54,7 +54,7 @@
   </macro>
   <macro name="author-short">
     <names variable="author">
-      <name form="short" delimiter="_" delimiter-precedes-last="always"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
This is fine (putting the et al cut off back into the citekeys so that those stay unchanged and note overly long). 
I'm not 100% sure, but I don't think CrossRef uses CSL to format bibtex -- there are several things in their bibtex that don't look like our style -- so that problem is very likely due to something else.

PR #5730, opened 2021-11-17
fix minor erros in polar-research.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/393746/#Comment_393746


polar-research.csl
```diff
@@ -199,8 +200,8 @@
       <key macro="author"/>
       <key variable="title"/>
     </sort>
-    <layout>
-      <group suffix=".">
+    <layout suffix=".">
+      <group delimiter=". ">
```
adam3smith, line 204: This causes issues with the DOI display (see preview)

adam3smith (changes requested):

adam3smith:
Thanks!

PR #5731, opened 2021-11-17
Remove group around no-date cluster

bwiernik:
Unnecessary and removing ensures good performance across citeprocs

https://github.com/andras-simonyi/citeproc-el/issues/70#issuecomment-971911773


adam3smith:
Thanks!

PR #5744, opened 2021-11-25
Update china-national-standard-gb-t-7714-2015-author-date.csl

redleafnew:
Add only to show the number for bill broadcast legal_case legislation patent report song, otherwise, if there is 'number:' in Extra field, the number will show for item type of journal.


redleafnew:
please remove the space before bill.

PR #5749, opened 2021-11-27
Create clinical-kidney-journal.csl

lfmorrone:
The style of Clinical Kidney Journal is superimposable to the style of Nephrology Dialysis Transplantation. So, I started from this style and changed the only name and ISSN parameters. Next, I successfully tested for validation and formatted the cls file and am trying to validate it.


adam3smith:
Thanks! To explain -- dependent styles actually don't contain any CSL code: they just link to the parent style, so only one of the styles needs to be maintained.

PR #5752, opened 2021-11-30
Create journal-of-health-care-for-the-poor-and-underserved.csl

POBrien333:
via https://forums.zotero.org/discussion/93207/request-for-jhcpu-citation-style-to-be-added#latest


adam3smith:
Thanks!

PR #5755, opened 2021-12-01
Update article format for STM journal

fdchevalier:
Science Translational Medicine requires now a period after a journal article title instead of a comma.


adam3smith:
Thanks!

PR #5758, opened 2021-12-03
Add publisher information to dictionaries and encyclopedias for SBL 2nd edition

Addisonbean:
SBL 2nd edition is currently missing the publisher information when citing a dictionary or encyclopedia entry.

### DIFF
--- a/society-of-biblical-literature-author-date.csl
+++ b/society-of-biblical-literature-author-date.csl
@@ -483,6 +483,7 @@
         </group>
       </group>
       <text macro="point-locators"/>
+      <text macro="issue"/>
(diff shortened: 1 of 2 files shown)


adam3smith:
I don't think this is correct: In 6.3.6 I'm seeing:

> 33. Krister Stendahl, "Biblical Theology, Contemporary," _IDB_ 1: 418-32.

That's also consistent with the author-date format in 6.5.11

adam3smith:
Closing this because I'm fairly sure I'm right here. Happy to re-open if I've misread the manual, just post here

PR #5775, opened 2021-12-10
Create journal-of-global-health.csl

POBrien333:
via https://forums.zotero.org/discussion/90749/style-request-journal-of-global-health#latest

Based on .pdf. https://jogh.org/wp-content/uploads/2021/11/jogh-11-08009.pdf


adam3smith:
Thanks!

PR #5778, opened 2021-12-13
Add encyclopedia type to wikipedia-templates.csl

sallybaker:


wikipedia-templates.csl
```diff
@@ -146,7 +171,7 @@
   </macro>
   <macro name="pagination">
     <choose>
-      <if type="article-journal chapter">
+      <if type="article-journal chapter entry-encyclopedia">
```
adam3smith, line 174: This needs match="any" (should have been in previous version too)

adam3smith (changes requested):
Looks good. One quick fix

wikipedia-templates.csl
sallybaker: Indeed. Thanks for picking that up. Fixed now.

PR #5779, opened 2021-12-13
Update and rename dependent/organization-studies.csl to organization-…

POBrien333:
…studies.csl

via https://forums.zotero.org/discussion/93420/recommendation-needed-reference-list-i-need-apa-7-but-with-full-first-names-instead-of-initials#latest


adam3smith:
Thanks!

PR #5782, opened 2021-12-13
Update gallia.csl

NMonteix:


adam3smith:
It looks like you have lost some nbsp? On purpose?

NMonteix:
I don't understand what happened with the nbsp. Anyway, they're back (even if Github transformed the `&#160;` html code in spaces that don't appear to be nbsp... They should be though

NMonteix:
By the way, strangely enough, the sample citations given above are as such:
(Hancké et al. dir.  **;** 2007 ; Collectif  **;** 2012)
(Mares  **;** 2001 ; Fenner et al.  **;** 2019)
[with a semicolon between author and date]
While the code is 
`<layout prefix="(" suffix=")" delimiter="&#160;; ">`
`<text macro="author-short" suffix=" "/>`
`<text macro="year-date"/>`
`<group prefix=", ">`
And hence should be rendered by (Author1 date ; Author2 date). With the Zotero style editor, the rendering is fine.

adam3smith:
The citeproc generating the previews has a bug with group delimiters. The Zotero output is correct

adam3smith:
Thanks!

PR #5789, opened 2021-12-15
Create medicinski-razgledi.csl

SterArcher:


medicinski-razgledi.csl
```diff
@@ -0,0 +1,166 @@
...
+    <category citation-format="numeric"/>
+    <category field="medicine"/>
+    <issn>0025-8121</issn>
+    <eissn>1581-3797</eissn>
+    <updated>2021-12-06T19:54:16+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <macro name="author">
+    <names variable="author" font-weight="normal" delimiter=",">
+      <name delimiter-precedes-et-al="always" delimiter-precedes-last="always" et-al-min="3" initialize-with="" name-as-sort-order="all" sort-separator="  "/>
```
adam3smith, line 23: You have an et-al-min here without an et-al-use-first, which isn't technically invalid, but is confusing in styles (and may have poorly determined output across different processors) so we're requiring both elements for the repository

SterArcher: @adam3smith  Thank you for your prompt reply. How should i go about adressing this issue to resolve it?

adam3smith: Just adding et-al-use-first="1" (which I assume is what you want?) will fix the test

SterArcher: image

I'm sorry @adam3smith but the problem persists... Anything else I could do?

adam3smith: Same issue in line 85

SterArcher: Works now, thank you for your time and knowledge! @adam3smith

medicinski-razgledi.csl
```diff
@@ -0,0 +1,166 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" delimiter-precedes-et-al="always" delimiter-precedes-last="always" initialize-with="" demote-non-dropping-particle="sort-only" default-locale="sl-SI">
```
adam3smith, line 2: Are you sure you want to set the style to sl? All terms I'm seeing uses in references bloke et al and eds. are English

medicinski-razgledi.csl
```diff
@@ -0,0 +1,166 @@
...
+  </macro>
+  <citation collapse="citation-number">
+    <sort>
+      <key variable="citation-number"/>
+    </sort>
+    <layout font-style="italic" delimiter=", " prefix="(" suffix=")">
+      <text variable="citation-number"/>
+    </layout>
+  </citation>
+<bibliography et-al-min="3" et-al-use-first="1" second-field-align="flush">
```
adam3smith, line 85:
```suggestion
<bibliography et-al-min="4" et-al-use-first="3" second-field-align="flush">
```
Sorry, I had misunderstood the requirements

medicinski-razgledi.csl
SterArcher: Yes, it should be et al. Will change back to English.

medicinski-razgledi.csl
```diff
@@ -0,0 +1,166 @@
...
+    <sort>
+      <key variable="citation-number"/>
+    </sort>
+    <layout font-style="italic" delimiter=", " prefix="(" suffix=")">
+      <text variable="citation-number"/>
+    </layout>
+  </citation>
+<bibliography et-al-min="4" et-al-use-first="3" second-field-align="flush">
+    <sort>
+      <key variable="call-number"/>
```
adam3smith, line 87: Sorry I hadn't seen this: is this a typo? I'd think you want citation-number ?

SterArcher: Sorry I am not really versed in this (I made this style because the journal I was submitting to didn't have it), to what specifically is this aimed at? Could you help me? Thanks!

adam3smith: It determines how the bibliography gets sorted. I assume it should just be by order of appearance? Currently it starts by sorting by the call number, i.e. how books are cataloged in a library

SterArcher: image
Ok, i have deleted the sort part in accordance with the csl spec sheet, I think it should be ok now (if i understand the spec correctly).

adam3smith:
Thanks!

PR #5790, opened 2021-12-15
Create art-libraries-society-of-north-america-arlisna-reviews.csl

m406m947:
Initial submission of the Art Libraries Society of North America (ARLIS/NA) Reviews (ISSN 2374-4723) citation style based on Chicago.


m406m947:
Moving style out of dependent. It's actually a unique style which uses Chicago as a starting point but citation display differs substantially.

PR #5792, opened 2021-12-16
Removed "page" (s. ) term for articles that do not have page numbers.

StianOby:
Currently, in the bibliography articles without a page number (e.g. preprints with only a DOI) will still display the page term: "s. " - which is equivalent to "p. " in English.

This commit adds a check for whether a page number exists, and omits the page term if there is no page number.

### DIFF
--- a/norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
+++ b/norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
@@ -286,12 +286,10 @@
             <text prefix=", " macro="title"/>
             <text prefix=" " font-style="italic" variable="container-title"/>
             <text prefix=" " macro="issued"/>
-            <choose>
-              <if variable="page" match="any">
-                <text prefix=" " suffix=" " term="page" form="short"/>
-                <text variable="page"/>
-              </if>
-            </choose>
+            <group delimiter=" " prefix=" ">
+              <text term="page" form="short"/>
+              <text variable="page"/>
+            </group>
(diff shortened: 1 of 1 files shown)


norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
```diff
@@ -286,8 +286,12 @@
...
             <text prefix=" " font-style="italic" variable="container-title"/>
             <text prefix=" " macro="issued"/>
-            <text prefix=" " suffix=" " term="page" form="short"/>
-            <text variable="page"/>
+            <choose>
+              <if variable="page" match="any">
+                <text prefix=" " suffix=" " term="page" form="short"/>
+                <text variable="page"/>
+              </if>
+            </choose>
```
bwiernik, line 294:
```suggestion
            <group delimiter=" " prefix=" ">
              <text term="page" form="short"/>
              <text variable="page"/>
            </group>
```

bwiernik (changes requested):
Thanks!

The appropriate way to do this is to put the term and variable inside of a `group`. Terms are suppressed in a group when all of the variables are empty.

norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
StianOby: I had a slight hunch that there was a simpler solution. Thanks for spelling it out, so that I can learn!

PR #5793, opened 2021-12-16
Update monographs-of-the-palaeontographical-society.csl

benjaminmoon:
These changes are made in consultation with the journal editors – a new version of the Notes for Authors is to be published shortly (with a new website).

- add documentation link to publishing society website (<palaeosoc.org>)
- changed to sorting citations by author surname then by issue year
- change bibliography sorting to author surname then issue year
- remove macro for counting number of authors as this is not used
- change citation year-suffix collapsing to "year-suffix", not ranged
- include new field for DOI at end of references


bwiernik:
Thanks! 

If there is still some time to consult on the changes, I might suggest using the prefix "https://doi.org/" instead of "doi: ", as that is the preferred prefix of CrossRef and the DOI Foundation.

PR #5796, opened 2021-12-17
Create turabian-fullnote-bibliography-no-ibid

dhacker29:
Removes the ibid function from Turabian style as it is not preferred by many institutions and it also has the issue of using ibid for the first reference on a page which should not happen.


bwiernik:
Any particular reason not to just use Chicago (full note)?

dhacker29:
Professor requires us to use Turabian style.

bwiernik:
Turabian style is just slightly simplified Chicago.

bwiernik:
@adam3smith what do you think?

adam3smith:
I think the experience of Turabian is dumb, but since that's inevitable, the style makes sense

turabian-fullnote-bibliography-no-ibid.csl
```diff
@@ -0,0 +1,811 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>Turabian 8th edition (full note) no ibid</title>
```
bwiernik, line 4:
```suggestion
    <title>Turabian 8th edition (full note, no ibid)</title>
```

bwiernik:
Thanks!

PR #5797, opened 2021-12-20
Update universidade-do-porto-faculdade-de-engenharia-chicago.csl

serbernardo:
Change in title an author in order to be in line with our Chicago Manual of Style Portuguese Version: universidade-do-porto-faculdade-de-engenharia-chicago-pt.csl


adam3smith:
Wait, I had completely missed that there already is a version. Is having a separate version for Portuguese and English really necessary? We very much try to avoid that.  Couldn't this be handled through locale definitions?

serbernardo:
Well it has some changes: For instance with 3 authors in english it has a comma before the and (Khoramishad, H., O. Alizadeh, aand L. F. M. da Silva)
In Portuguese that is very strange and so we have to have (Khoramishad, H., O. Alizadeh e L. F. M. da Silva)

PR #5799, opened 2021-12-23
Replace "sub verbo" with "sub-verbo"

bwiernik:


bwiernik:
@adam3smith When do you want to merge this?

adam3smith:
hmmm -- right before switching schema?

PR #5800, opened 2021-12-23
Non-breaking 1.0.2 validation fixes (et-al affixes, 'container' date)

bwiernik:


adam3smith:
What's up with the Iberoromanistik style? Those aren't et al changes.

bwiernik:
Good to merge @adam3smith ?

universitat-basel-iberoromanistik.csl
```diff
@@ -28,7 +28,7 @@
         <names variable="translator" delimiter="/" prefix="[" suffix="]">
           <label form="short" suffix=" "/>
           <name delimiter=" / " delimiter-precedes-last="always"/>
-          <et-al prefix=" (" suffix=") " font-style="italic"/>
+          <et-al font-style="italic"/>
```
adam3smith, line 31: Do we need to redefine the term here as for the [] case  above?

wiley-vch-books.csl
```diff
@@ -65,7 +65,7 @@
       <text variable="citation-number" suffix="."/>
       <text macro="authors"/>
       <choose>
-        <if match="all" is-uncertain-date="issued accessed original-date container">
+        <if match="all" is-uncertain-date="issued accessed original-date">
```
adam3smith, line 68: (no need to fix this here, but this also looks wrong...)

adam3smith (commented):
One question

universitat-basel-iberoromanistik.csl
bwiernik: No, I checked the documentation and they don't actually put et al in parentheses

wiley-vch-books.csl
bwiernik: As in, the rendering logic is incorrect?

adam3smith: Yeah. What does in print have to do with uncertain date in general and uncertain access date in particular

adam3smith:
Ok to merge then

PR #5801, opened 2021-12-23
Fix term hacks for transition to CSL 1.0.2

bwiernik:


bwiernik:
This will be merged once we switch to validating against CSL 1.0.2

adam3smith:
(I can't believe this was valid before 🤣)

bwiernik:
Abusing the zoology term was pretty clever hahaha 🤣

PR #5807, opened 2021-12-27
Update Society of Biblical Literature Full Note for 6.1.6

dhacker29:
SBL Handbook 6.1.6 recommends using DOI instead of URL if possible. This change checks for DOI variable and prefers it over URL if present.


bwiernik:
Thanks!

PR #5808, opened 2021-12-28
Update SBL 2nd ed. (Full Note) for 6.2.25 Electronic Book

dhacker29:
eBook designation should be at the end of the citation rather than after the title according to SBL Handbook 6.2.25 Electronic Book


bwiernik:
How does this affect the placement of information for interviews and other items types?

dhacker29:
> 

From my testing they all still work as they should.

denismaier:
@bwiernik can we merge this? It looks like the errors have to do with other styles.

PR #5809, opened 2021-12-28
Update Turabian (full note) for 17.1.3.2 Reprint Editions

dhacker29:
Add the originally-published and original-date extras


adam3smith:
Thanks!

PR #5821, opened 2022-01-03
Create mycobiology.csl

POBrien333:
via https://forums.zotero.org/discussion/92820/style-request-mycobiology#latest

Sample paper 1 https://www.tandfonline.com/doi/pdf/10.1080/12298093.2021.1965706?needAccess=true
Sample paper 2 https://www.tandfonline.com/doi/pdf/10.1080/12298093.2021.1975919?needAccess=true


adam3smith:
Thanks!

PR #5822, opened 2022-01-03
Create the-korean-journal-of-mycology.csl

POBrien333:
via https://forums.zotero.org/discussion/92819/style-request-the-korean-journal-of-mycology#latest

Sample paper: http://www.kjmycology.or.kr/view/N0320490302.pdf

### DIFF
--- a/the-korean-journal-of-mycology.csl
+++ b/the-korean-journal-of-mycology.csl
@@ -302,7 +302,7 @@
       <text variable="citation-number"/>
     </layout>
   </citation>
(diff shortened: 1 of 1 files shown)


adam3smith:
ah, too bad, that didn't work, I'll probably just put this into a new PR (or you can)

retorquere:
For security reasons, sheldon refuses to run on PRs that attempt to change both the testing infra (the Gemfile in this case) and styles.

retorquere:
Although we could make it so that if it's labeled safe to test Sheldon will allow mixed PRs.

adam3smith:
Thanks!

PR #5826, opened 2022-01-04
Update tyndale-bulletin.csl

dstark:
Corrects output for footnotes and bibliography per §§11.3.8–11.3.9 of the Tyndale Bulletin style guide for titled and untitled book reviews.

### DIFF
--- a/tyndale-bulletin.csl
+++ b/tyndale-bulletin.csl
@@ -16,14 +16,12 @@
     <issn>0082-7118</issn>
     <eissn>2752-7042</eissn>
     <summary>Tyndale Bulletin format with full notes and bibliography</summary>
-    <updated>2021-10-20T19:25:50+00:00</updated>
+    <updated>2022-01-02T01:47:37+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
     <!-- Tyndale Bulletin style differs from SBLHS2 only rarely (§4.1). The departures are that Tyndale Bulletin uses 
       1) hyphens, rather than en dashes, in page ranges (§§6, 6.3);
       2) insists that DOIs be included where available (§§11.1, 11.3.7); and
-      3) uses British-style quotation mark conventions (§§8.1, 11.1, 11.6.6–11.3.8). N.B.: §11.3.8 shows punctuation coming outside the quotation mark, rather than inside. But it needs to be confirmed that this isn't a mistake in the Tyndale Bulletin style guide (as is, e.g., the italicized comma after the book title in that section's examples). 
-      
-      N.B.: Currently, the citation for an untitled book review puts the book title in quotation marks rather than italics (contra §11.3.9), as does the same code in the SBL base style. Could this be due to the difference between en-US and en-GB? -->
+      3) uses British-style quotation mark conventions (§§8.1, 11.1, 11.6.6–11.3.8). N.B.: §11.3.8 shows punctuation coming outside the quotation mark, rather than inside. But it needs to be confirmed that this isn't a mistake in the Tyndale Bulletin style guide (as is, e.g., the italicized comma after the book title in that section's examples). -->
   </info>
   <locale xml:lang="en-GB">
     <style-options punctuation-in-quote="true"/>
@@ -267,6 +265,19 @@
       <else-if type="bill book graphic legal_case legislation motion_picture report song" match="any">
         <text variable="title" font-style="italic" text-case="title"/>
       </else-if>
+      <!-- Include "review of " prefix, italicize the book titles in untitled book reviews per §§11.3.8–11.3.9. -->
+      <else-if variable="reviewed-author">
+        <choose>
+          <!-- Include the review title, if avaialable, in quotation marks per §11.3.8. -->
+          <if variable="reviewed-title">
+            <text variable="title" quotes="true" suffix=", " text-case="title"/>
+            <text variable="reviewed-title" font-style="italic" prefix="review of " text-case="title"/>
+          </if>
+          <else>
+            <text variable="title" font-style="italic" prefix="review of " text-case="title"/>
+          </else>
+        </choose>
+      </else-if>
       <else>
         <text variable="title" quotes="true" text-case="title"/>
       </else>
@@ -283,6 +294,17 @@
       </if>
       <else-if type="bill book graphic legal_case legislation motion_picture report song" match="any">
         <text variable="title" font-style="italic" text-case="title"/>
+      </else-if>
+      <!-- Include the reviewed author for book reviews per §§11.3.8–11.3.9. -->
+      <else-if variable="reviewed-author">
+        <choose>
+          <if variable="reviewed-title">
+            <text variable="title" quotes="true" suffix=", " text-case="title"/>
+            <text variable="reviewed-title" font-style="italic" prefix="review of " text-case="title"/>
+          </if>
+        </choose>
+        <text variable="title" font-style="italic" prefix="Review of " suffix=", by " text-case="title"/>
+        <names variable="reviewed-author"/>
       </else-if>
       <else>
         <text variable="title" quotes="true" text-case="title"/>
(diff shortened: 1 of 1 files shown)


dstark (commented):
Updated the last updated time stamp.

dstark:
@adam3smith or others, I seem to have done something wrong with submitting this update. But I'm unsure what "Error: The process '/usr/bin/git' failed with exit code 1" is telling me I need to do.

If you wouldn't mind clarifying, I'll be happy to resubmit my edits accordingly. Thank you so much!

adam3smith:
@retorquere -- if I interpret the error correctly, the checkout only works for PRs from branches on our repo, which unfortunately both you and me used during testing. David's fork/branch is up to date, so should really work.

@dstark not your fault, sorry for the problem; we'll figure this out and let you know if we need you to do anything.

retorquere:
I know what the fix is, I'll be back at my computer in 30 minutes.

retorquere:
The problem with the re-indent are fixed.

dstark (commented):
Ok great. Thanks so much, @retorquere and @adam3smith!

Looks like I might need to resubmit to get the checks to run all the way through again. So doing that now.

tyndale-bulletin.csl
```diff
@@ -284,6 +295,17 @@
...
       </else-if>
+      <!-- Include the reviewed author for book reviews per §§11.3.8&#8211;11.3.9. -->
+      <else-if variable="reviewed-author">
+        <choose>
+          <if variable="reviewed-title">
+            <text variable="title" quotes="true" suffix=", " text-case="title"/>
+            <text variable="reviewed-title" font-style="italic" prefix="review of " text-case="title"/>
+          </if>
+        </choose>
+        <text variable="title" font-style="italic" prefix="Review of " suffix=", by " text-case="title"/>
```
adam3smith, line 306: This should be an `else` condition, no? Otherwise you get the title twice when there is a reviewed title

tyndale-bulletin.csl
```diff
@@ -267,6 +265,19 @@
...
       </else-if>
+      <!-- Include "review of " prefix, italicize the book titles in untitled book reviews per §§11.3.8&#8211;11.3.9. -->
+      <else-if variable="reviewed-author">
+        <choose>
+          <!-- Include the review title, if avaialable, in quotation marks per §11.3.8. -->
+          <if variable="reviewed-title">
+            <text variable="title" quotes="true" suffix=", " text-case="title"/>
+            <text variable="reviewed-title" font-style="italic" prefix="review of " text-case="title"/>
+          </if>
+          <else>
```
adam3smith, line 276: Like you did here

adam3smith (changes requested):
I think there's a small mistake still, noted above.

tyndale-bulletin.csl
dstark: It does. Thanks. Sorry I missed that. I'll resubmit presently.

tyndale-bulletin.csl
```diff
@@ -267,6 +265,19 @@
       <else-if type="bill book graphic legal_case legislation motion_picture report song" match="any">
         <text variable="title" font-style="italic" text-case="title"/>
       </else-if>
+      <!-- Include "review of " prefix, italicize the book titles in untitled book reviews per §§11.3.8-11.3.9. -->
+      <else-if variable="reviewed-author">
+        <choose>
+          <!-- Include the review title, if avaialable, in quotation marks per §11.3.8. -->
+          <if variable="reviewed-title">
+            <text variable="title" quotes="true" suffix=", " text-case="title"/>
+            <text variable="reviewed-title" font-style="italic" prefix="review of " text-case="title"/>
```
bwiernik, line 274:
```suggestion
            <group delimiter=", ">
              <text variable="title" quotes="true" suffix=", " text-case="title"/>
              <group delimiter=" ">
                <text term="review-of" font-style="italic"/>
                <text variable="reviewed-title" font-style="italic" text-case="title"/>
              </group>
            </group>
```

bwiernik: Please do this structure here and elsewhere in the style. Use the review-of term is better than hard coding English text

bwiernik (changes requested):

tyndale-bulletin.csl
dstark: Noted. Thanks for the pointer to review-of. I didn't have that term in the copy of the CSL specification I was working from.

dstark:
@adam3smith and @bwiernik, I think I've addressed the changes you requested. But please let me know if I've missed something and am just not seeing it since I'm still learning my way around github. :-) Thanks so much for your thoughts on these updates.

tyndale-bulletin.csl
```diff
@@ -284,6 +300,20 @@
       <else-if type="bill book graphic legal_case legislation motion_picture report song" match="any">
         <text variable="title" font-style="italic" text-case="title"/>
       </else-if>
+      <!-- Include the reviewed author for book reviews per §§11.3.8-11.3.9. -->
+      <else-if variable="reviewed-author">
+        <choose>
+          <if variable="reviewed-title">
+            <text variable="title" quotes="true" suffix=", " text-case="title"/>
+            <text variable="reviewed-title" font-style="italic" prefix="review of " suffix=", by " text-case="title"/>
```
bwiernik, line 308: Please do terms rather than hard coded prefixes here as well

tyndale-bulletin.csl
```diff
@@ -284,6 +300,20 @@
...
+            <names variable="reviewed-author"/>
```
bwiernik, line 309: This need a "verb" label I would think

tyndale-bulletin.csl
```diff
@@ -307,6 +337,17 @@
...
         <text variable="title" form="short" font-style="italic" text-case="title"/>
       </else-if>
+      <!-- For short citations, use the short review title if available. Otherwise, use a "review of " prefix, italicize the short book title per §11.3.8. -->
+      <else-if variable="reviewed-author">
+        <choose>
+          <if variable="reviewed-title">
+            <text variable="title" quotes="true" text-case="title"/>
+          </if>
+          <else>
+            <text variable="title" font-style="italic" form="short" prefix="review of " text-case="title"/>
```
bwiernik, line 347: Term here too

bwiernik (changes requested):

tyndale-bulletin.csl
dstark: I think I have the other comments handled. But my apologies that I'm not following this one. Would you mind clarifying?

In case they're a helpful common point of reference, below are a couple of example bibliography entries that I've been using to test the titled and untitled citation forms.

Akiyama, Kengo. ‘Diverging Teachers of Righteousness’, review of _Sacred Texts and Paradigmatic Revolutions: The Hermeneutical Worlds of the Qumran Sectarian Manuscripts and the Letter to the Romans_, by J. David Stark. _ExpTim_ 126.7 (2015): 358–59.
Das, A. Andrew. Review of _Paul, Judaism, and the Gentiles: Beyond the New Perspective_, by Francis Watson. _Int_ 63 (2009): 200, 202. https://doi.org/10.1177/002096430906300220.

bwiernik: Please use the `label` element in `names` to produce "by" rather than a suffix https://docs.citationstyles.org/en/stable/specification.html#label-in-cs-names

dstark: Thank you. It does make total sense not to hard-code the "by" either.

To misquote the patriarch Abraham, though—let not him who actually knows what's going on be angry with me. But let me ask just once more …

Instead of `label`, would `<text term="by"/>` also be acceptable here with the addition of `<term name="by" form="verb">by</term>` above under `<terms>`? E.g.,

`<text variable="reviewed-title" font-style="italic" suffix="," text-case="title"/>`
`<text term="by"/>`
`<names variable="reviewed-author"/>`

I can get this to work and be valid. But I must be missing something with `names` and `label` since, even when I try just to copy how I'm seeing `label` used inside `names` elsewhere, I end up with a not valid XML error. :-/

adam3smith:
```suggestion
            <names variable="reviewed-author">
              <label variable="verb" suffix=" "/>
              <name and="text"/>
            </names>  
```

Though I honestly think just using the term is fine in this case (I am wondering, though, if you really don't need a `name` element, e.g. to set and="text". If a book by two authors is reviewed, I assume that's going to be "by Adam Smith and Karl Marx" and not "by Adam Smith, Karl Marx"?

dstark:
Thanks so much for the guidance, folks. I think the style is now ready for review with the changes you requested. If there's anything else, please just let me know.

bwiernik:
Thanks!

PR #5828, opened 2022-01-06
switch modern-pathology to dependent style to nature.csl

POBrien333:
via https://forums.zotero.org/discussion/93770/style-error-modern-pathology#latest


adam3smith:
Unfortunately this won't work. They use different et al (see Zotero thread)

POBrien333:
thanks for spotting that.
This is a nature.csl copy with just the et-al settings changed

adam3smith:
Weirdly, the journal actually publishes using _Nature_ style with et-al-use-first="1", but given how insistent they are in their guidelines, let's do 6 as they have.

PR #5831, opened 2022-01-06
Update university-of-roehampton-harvard.csl

longskatedeath:
Apologies for the double-posting; I couldn't work out how to modify my existing change proposal!

When generating citations, it was omitting the 'et al' on resources with 4+ authors. I've added it to line 449 this time (which I hope is the correct one), but can't be sure as I was ultimately using the Visual CSL editor to put it together in the first instance.


adam3smith:
Thanks!

PR #5832, opened 2022-01-07
Update administrative-science-quarterly.csl

POBrien333:
via https://forums.zotero.org/discussion/93799/request-fix-style-for-administrative-science-quarterly#latest


bwiernik:
Thanks!

PR #5843, opened 2022-01-14
in-text >> superscript for T&F ACS style

POBrien333:
via https://forums.zotero.org/discussion/93915/no-supercribt-for-taylor-francis-american-chemical-society-acs-style#latest


adam3smith:
@bwiernik could you please use GitHub's squash and merge to keep the commit history a bit more manageable?

adam3smith:
That's a repo setting? I thought it was individual. But yes, if there's a setting for that, let's use it

PR #5846, opened 2022-01-15
add medium to IEEE

adam3smith:
See 
https://forums.zotero.org/discussion/93801/how-to-cite-lectures-in-the-ieee-citation-style-using-zotero#latest
(not doing type/genre because it interferes with online presentations)

### DIFF
--- a/ieee.csl
+++ b/ieee.csl
@@ -238,14 +238,14 @@
                 <text variable="medium" text-case="capitalize-first"/>
               </else-if>
               <else>
-            <text term="online" text-case="capitalize-first"/>
-            <choose>
-              <if type="motion_picture">
-                <text term="video" text-case="capitalize-first"/>
-              </if>
+                <text term="online" text-case="capitalize-first"/>
+                <choose>
+                  <if type="motion_picture">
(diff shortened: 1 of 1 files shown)


ieee.csl
```diff
@@ -234,11 +230,21 @@
...
           <group prefix="[" suffix="]" delimiter=" ">
-            <text term="online" text-case="capitalize-first"/>
             <choose>
-              <if type="motion_picture">
-                <text value="video" text-case="capitalize-first"/>
+              <if variable="medium" match="any">
+                <text variable="medium" text-case="capitalize-first"/>
               </if>
+              <else-if type="speech" variable="genre" match="all">
+                <text variable="medium" text-case="capitalize-first"/>
```
bwiernik, line 238: This will never render anything because if an item has medium it will hit the previous condition

bwiernik (changes requested):

ieee.csl
adam3smith: sorry, that was a typo. I meant to use genre, but ended up removing this again because "type" is used for things that definitely shouldn't be in square brackets for IEEE.

PR #5856, opened 2022-01-17
Create jurisprudence.csl

OTristanF:
Suitable for Jurisprudence – a journal published by Charles University, Faculty of Law.

### DIFF
--- a/jurisprudence.csl
+++ b/jurisprudence.csl
@@ -278,7 +278,7 @@
       </if>
       <else>
         <text term="available at" text-case="capitalize-first" suffix="&#160;"/>
-        <text variable="URL" prefix="&lt;" suffix="&gt;."/>
+        <text variable="URL" prefix="&lt;" suffix=">."/>
       </else>
     </choose>
   </macro>
--- a/jurisprudence.csl
+++ b/jurisprudence.csl
@@ -211,9 +211,6 @@
   <macro name="publisher-place">
     <text variable="publisher-place"/>
   </macro>
-  <macro name="printers">
-    <text variable="publisher"/>
-  </macro>
   <macro name="publisher">
     <text macro="publisher-place"/>
   </macro>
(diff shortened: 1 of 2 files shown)


OTristanF:
Isn't the automatic "re-indent" commit (92e3331) wrong? It replaced `&gt;` with `>`.

bwiernik:
@retorquere

retorquere:
The re-indenter does more than re-indent, it also normalizes entities, it has always done this. It isn't necessary to write `&gt;` inside attribute values.

bwiernik:
Thanks, I had thought that >< needed to be escaped

retorquere:
Outside attribute values these do need to be escaped.

OTristanF:
Should I change `&lt;` to `<` then?

retorquere:
If you don't, the re-indenter will (but only in attribute values)

OTristanF:
@retorquere, the re-indenter has changed just `&gt;` to `>`, without changing the corresponding `&lt;` to `<`. It is quite strange.

retorquere:
@rmzelle, any idea? I haven't changed the re-indenter, so this will be long-standing behavior.

bwiernik:
I don't see anywhere where `&gt;` is mentioned in the formatter library. So is that coming from the imported XML Parser?

retorquere:
That's my guess, but that'd be pretty strange for a widely used domparser.

rmzelle:
I just ran the original style submission through https://formatter.citationstyles.org/ (in Firefox), and it doesn't touch either `&gt;` or `&lt;`.

retorquere:
Damn, that means it's an issue in the `xmldom` package. I'll see if I can create an isolated testcase and submit it there.

bwiernik:
Personally, I think it's less confusing to leave the entities in both attributes and values, rather than changing them in the one place where they aren't needed

retorquere:
It's not something I actively do. I have an isolated testcase for xmldom, it's in their serializer.

rmzelle:
Maybe related to https://github.com/xmldom/xmldom/issues/339 ?

retorquere:
Yeah, they're deeming it desirable behavior (https://github.com/xmldom/xmldom/issues/198). I'll see if this can be either worked around or made configurable.

retorquere:
Blegh, it's not in a place where I can monkey-patch the problem away. But according to the reasoning in https://www.w3.org/TR/xml/#CleanAttrVals, they are right:

> Well-formedness constraint: No < in Attribute Values

bwiernik:
Ah, so let's just let it be then? Or post hoc edit ">" back to "&gt;"?

retorquere:
Not the latter, I'd say. Without a proper xml serializer we'd have to resort to regex hacks. I'd sooner patch the xml serializer before applying it.

bwiernik:
@adam3smith thoughts?

adam3smith:
I don't see any reason we shouldn't just leave > alone, yes.

PR #5857, opened 2022-01-18
Update journal-of-clinical-neurophysiology.csl

POBrien333:
via https://forums.zotero.org/discussion/93975/style-error-journal-of-clinical-neurophysiology#latest

Sample article available here: https://journals.lww.com/clinicalneurophys/Pages/currenttoc.aspx


bwiernik:
Thanks!

Issue #5352, opened 2021-03-26
MLA Style (8th edition) citations for journal articles incorrectly including publisher name


happywanderer23:
My library is using "modern-language-association.csl" in our Ex Libris Primo discovery system. 

Recently I noticed that journal article citations are incorrectly including the publisher name as part of the citation. Here are 2 examples:

Citation #1 (Routledge should not be included):

Billings, Linda. “Should Humans Colonize Mars? No.” Theology and Science, vol. 17, no. 3, Routledge, 2019, pp. 341–46, doi:10.1080/14746700.2019.1632524.

Permalink: https://alliance-primo.hosted.exlibrisgroup.com/permalink/f/o4maap/TN_cdi_crossref_primary_10_1080_14746700_2019_1632524

Citation #2 (Kluwer Academic Publishers should not be included):

Schenker, Paul S, et al. “Planetary Rover Developments Supporting Mars Exploration, Sample Return and Future Human-Robotic Colonization.” Autonomous Robots, vol. 14, no. 2, Kluwer Academic Publishers, 2003, pp. 103–26, doi:10.1023/A:1022271301244.

Permalink: https://alliance-primo.hosted.exlibrisgroup.com/permalink/f/o4maap/TN_cdi_gale_infotracacademiconefile_A155106390

I believe that this issue started occurring sometime after February 2021.  Is there a fix for this issue?

Stephanie

JaniceChristopher:
This has been an ongoing problem in the MLA 8th, and I see it's also a problem in the MLA 9th .csl. Can someone please fix this so my reference librarians will quit bugging me? Thank you!

POBrien333:
@happywanderer23  and @JaniceChristopher 

As you can see in the linked PR this has been addressed and you can update either of the styles in your documents to get the update.

PR #5859, opened 2022-01-19
Update mla8 and mla9

POBrien333:
closes https://github.com/citation-style-language/styles/issues/5352

Let's see if you like this :)

### DIFF
--- a/modern-language-association-8th-edition.csl
+++ b/modern-language-association-8th-edition.csl
@@ -194,7 +194,7 @@
   </macro>
   <macro name="publisher">
     <choose>
-      <if type="article-journal" match="none">
+      <if type="article-magazine article-newspaper article-journal" match="none">
         <text variable="publisher"/>
       </if>
     </choose>
--- a/modern-language-association-8th-edition.csl
+++ b/modern-language-association-8th-edition.csl
@@ -193,7 +193,11 @@
     </group>
   </macro>
   <macro name="publisher">
-    <text variable="publisher"/>
+    <choose>
+      <if type="article-journal" match="none">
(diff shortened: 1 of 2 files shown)


bwiernik:
What does MLA say about newspapers and magazines for publisher?

adam3smith:
At least in the 8th edition, all periodicas are treated the same: no publisher, no place of publication (for local Newspapers, it can be added in square brackets after the name of the paper, but that should likely be in the data, not automated)

modern-language-association-8th-edition.csl
```diff
@@ -193,7 +193,11 @@
     </group>
   </macro>
   <macro name="publisher">
-    <text variable="publisher"/>
+    <choose>
+      <if type="article-journal" match="none">
```
bwiernik, line 197:
```suggestion
      <if type="article-magazine article-newspaper article-journal" match="none">
```

modern-language-association.csl
```diff
@@ -194,7 +194,11 @@
     </group>
   </macro>
   <macro name="publisher">
-    <text variable="publisher"/>
+    <choose>
+      <if type="article-journal" match="none">
```
bwiernik, line 198:
```suggestion
      <if type="article-magazine article-newspaper article-journal" match="none">
```

bwiernik (changes requested):

POBrien333:
Thanks :)

adam3smith:
Thanks! We should think about how to implement this more widely - with Zotero importing publishers via embedded metadata, e.g., we'll see more of this for less prominent styles

PR #5860, opened 2022-01-20
Update China national standard GB/T 7714 styles

zepinglee:
This is a rework of the China national standard GB/T 7714 styles (both 2005 and 2015 versions). The outputs are now more closely to the standards.

I've collected all the example entries from the standard in [zepinglee/china-national-standard-gb-t-7714-2015/gbt7714-numeric.txt](https://github.com/zepinglee/china-national-standard-gb-t-7714-2015/blob/standard/gbt7714-numeric.txt) as well as structured JSON data in that repo. This is a comparison between the standard and the output of the original `.csl` styles: <https://github.com/zepinglee/china-national-standard-gb-t-7714-2015/compare/standard...original>. After my edit, less differences are produced ([zepinglee/china-national-standard-gb-t-7714-2015@standard...main](https://github.com/zepinglee/china-national-standard-gb-t-7714-2015/compare/standard...main#diff-cfbf4589528ea9beaeebb77942283da69f584518d16eefcb63677b17ea328615)). Most of the differences are related to multi-language issues. They cannot be solved with current CSL v1.0.2 as far as I know.

The notable changes are as follows.

1. Entry types from CSL v1.0.2 are added.
2. The document links are updated.
1. The role of `editor` is allowed to be the primary contributor of a book instead the anonymous term.
2. The anonymous term is omitted in numeric and note styles as specified in the standard.
3. Sentence case (actually no case conversion) is used for titles and container titles instead of title case. This is not explicitly specified in the style but is used in the all the examples.
4. The names of proceeding editor are moved after the "//" mark (which is used as "In:").
5. The edition number is now checked if it is numeric.
6. The `page-range-delimiter` is changed to hyphen as specified in the standard, which is similar to Vancouver style.
7. The hyphens in ISO dates (yyyy-mm-dd) are corrected (rather than en dashes).
8. The missing issued date in `report` type is fixed.
9. Cite items are not sorted. 
10. A space is inserted in the author-date citation ahead of `et-al` term in Chinese, which is required by the standard.

### DIFF
--- a/china-national-standard-gb-t-7714-2015-numeric.csl
+++ b/china-national-standard-gb-t-7714-2015-numeric.csl
@@ -258,7 +258,7 @@
           <choose>
             <if variable="archive">
               <text value="A"/>
-            </if >
+            </if>
             <else>
               <text value="M"/>
             </else>
@@ -426,10 +426,7 @@
   </citation>
   <bibliography entry-spacing="0" et-al-min="4" et-al-use-first="3" second-field-align="flush">
     <!-- 取消这部分注释可以使用 CSL-M 的功能支持双语 -->
-    <!-- <layout locale="en">
-      <text variable="citation-number" prefix="[" suffix="]"/>
-      <text macro="entry-layout"/>
-    </layout> -->
+    <!-- <layout locale="en"><text variable="citation-number" prefix="[" suffix="]"/><text macro="entry-layout"/></layout> -->
     <layout>
       <text variable="citation-number" prefix="[" suffix="]"/>
       <text macro="entry-layout"/>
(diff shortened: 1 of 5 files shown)


china-national-standard-gb-t-7714-2015-numeric.csl
```diff
@@ -1,242 +1,435 @@
...
+      <term name="open-inner-quote">‘</term>
+      <term name="close-inner-quote">’</term>
     </terms>
   </locale>
-  <macro name="accessed-date">
-    <date variable="accessed" delimiter="&#8211;" prefix="[" suffix="]">
-      <date-part name="year"/>
-      <date-part name="month" form="numeric-leading-zeros"/>
-      <date-part name="day" form="numeric-leading-zeros"/>
+  <locale>
```
adam3smith, line 36: Is there are reason you have two `locale` sections in the style? Since it has a default locale, I don't see a reason not to write everything into the `zh-CN` locale.

zepinglee: This `locale` section without `xml:lang` is convenient for creating variants with the CSL-M locale extension (multiple `layout`s with `locale` attributes) enabled for practical uses. The standard is actually a bilingual style and the formats of numeric `date` and `page-range-delimiter` are independent of different languages.

adam3smith: Thanks, that makes sense

adam3smith:
Thanks, also for the detailed summary of changes

PR #5861, opened 2022-01-20
Update deutsche-gesellschaft-fur-psychologie.csl

tobistocker:
I needed this citation style to work with translated works, which was not implemented in the current version. I added this functionality to the style for Deutsche Gesellschaft für Psychologie.

I also took the liberty to update some references, since the link to the 4th edition was not working. I cannot tell if the citation rules for translated books or classic works are the same in the 4th edition. The rest of the citation style is compatible and compliant with the 5th edition, so I assume this was safe to update.

I have impelemented two cases:
1) classic works
2) translated works

They are being cited differently and the bibliography entry looks slightly different for these two cases. I decided to use the "translator" variable to distignuisch the two cases, I assumed this would be safe.

### DIFF
--- a/deutsche-gesellschaft-fur-psychologie.csl
+++ b/deutsche-gesellschaft-fur-psychologie.csl
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="de-DE">
   <info>
-    <title>Deutsche Gesellschaft für Psychologie (Deutsch)</title>
+    <title>Deutsche Gesellschaft für Psychologie (Deutsch) 5. Auflage</title>
+    <title-short>DGP 5. Auflage</title-short>
     <id>http://www.zotero.org/styles/deutsche-gesellschaft-fur-psychologie</id>
     <link href="http://www.zotero.org/styles/deutsche-gesellschaft-fur-psychologie" rel="self"/>
-    <link href="https://www.psychologie.uni-bonn.de/de/studium/richtlinien-zur-manuskriptgestaltung/view" rel="documentation"/>
     <link href="https://github.com/citation-style-language/styles/pull/432" rel="documentation"/>
-    <link href="https://www.hogrefe.de/shop/richtlinien-zur-manuskriptgestaltung-75884.html" rel="documentation"/>
+    <link href="https://elibrary.hogrefe.com/book/10.1026/02954-000" rel="documentation"/>
     <author>
       <name>Daniel Hirsbrunner</name>
       <email>dhirsbrunner@gmx.ch</email>
@@ -18,10 +18,14 @@
       <name>Patrick O'Brien</name>
       <email>obrienpat86@gmail.com</email>
     </contributor>
+    <contributor>
+      <name>Tobias Stocker</name>
+      <email>tobistocker@gmail.com</email>
+    </contributor>
     <category citation-format="author-date"/>
     <category field="psychology"/>
-    <summary>Style for the German society for psychology, based on the 4th style manual (Feb-2016).</summary>
-    <updated>2017-03-22T14:15:50+00:00</updated>
+    <summary>Style for the German society for psychology, based on the 5th style manual (2019).</summary>
+    <updated>2022-01-20T12:24:00+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="de">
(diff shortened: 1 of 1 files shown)


adam3smith:
(The test shouldn't fail -- I think it just expects the parentheses at the end of the title)

POBrien333:
via https://forums.zotero.org/discussion/93988/style-error-deutsche-gesellschaft-fuer-psychologie/p1

deutsche-gesellschaft-fur-psychologie.csl
```diff
@@ -1,12 +1,12 @@
...
   <info>
-    <title>Deutsche Gesellschaft für Psychologie (Deutsch)</title>
+    <title>Deutsche Gesellschaft für Psychologie 5. Auflage (Deutsch)</title>
+    <title-short>DGP 5. Auflage</title-short>
     <id>http://www.zotero.org/styles/deutsche-gesellschaft-fur-psychologie</id>
     <link href="http://www.zotero.org/styles/deutsche-gesellschaft-fur-psychologie" rel="self"/>
-    <link href="https://www.psychologie.uni-bonn.de/de/studium/richtlinien-zur-manuskriptgestaltung/view" rel="documentation"/>
     <link href="https://github.com/citation-style-language/styles/pull/432" rel="documentation"/>
-    <link href="https://www.hogrefe.de/shop/richtlinien-zur-manuskriptgestaltung-75884.html" rel="documentation"/>
+    <link href="https://elibrary.hogrefe.com/book/10.1026/02954-000" rel="documentation"/>
```
bwiernik, line 9:
```suggestion
    <link href="https://doi.org/10.1026/02954-000" rel="documentation"/>
```

deutsche-gesellschaft-fur-psychologie.csl
```diff
@@ -279,6 +283,65 @@
...
+      <if variable="issued">
+        <group>
+          <choose>
+            <if variable="original-date">
+              <date variable="original-date">
+                <date-part name="year"/>
+              </date>
+              <choose>
+                <if variable="translator">
+                  <text value="; übers. "/>
```
bwiernik, line 297: What does `übers.` mean in this context?

deutsche-gesellschaft-fur-psychologie.csl
```diff
@@ -279,6 +283,65 @@
...
+                </if>
+                <else>
+                  <text value="/"/>
```
bwiernik, line 300: These 2 text values should be delimiters on a group, otherwise they might be printed when data are missing. The choose test should be wrapped around the whole group

deutsche-gesellschaft-fur-psychologie.csl
```diff
@@ -279,6 +283,65 @@
...
+    <group prefix="(" suffix=")">
+      <choose>
+        <if variable="references">
+          <text variable="references"/>
+        </if>
+        <else>
+          <choose>
+            <if variable="original-date">
+              <group>
+                <text value="Original erschienen"/>
```
bwiernik, line 331: Use the "originally-published" term instead

deutsche-gesellschaft-fur-psychologie.csl
```diff
@@ -279,6 +283,65 @@
...
+                <text macro="original-date" prefix=" "/>
+                <choose>
+                  <if variable="translator">
+                    <text variable="original-title" prefix=": "/>
```
bwiernik, line 335: I'm a bit confused by the prefix here. What's an example citation with a translator look like?

bwiernik (changes requested):

bwiernik:
That last commit changes the output behavior. Can you give an example of what a reference with a translator is supposed to look like so I can understand?

deutsche-gesellschaft-fur-psychologie.csl
tobistocker: From the manual (p. 123):
Lindsay, P. H. & Norman, D. A. (1981). _Einführung in die Psychologie_ (H.-D. Dumpert, F. Schmidt, M. Schuster & M. Steeger, Übers.). Berlin: Springer. (Original erschienen 1977: Human information processing).

tobistocker: A classical work would not have the translator part neither an original title. Therefore the last bit would only state (Original erschienen 1930) for example.

tobistocker:
Sorry for the confusion @bwiernik.

The citation and bibliography entry for a translated book would look like this (From the manual, p. 123):

bibliography:
Lindsay, P. H. & Norman, D. A. (1981). Einführung in die Psychologie (H.-D. Dumpert, F. Schmidt, M. Schuster & M. Steeger, Übers.). Berlin: Springer. (Original erschienen 1977: Human information processing).

Citation in the text:
(Lindsay & Norman, 1977/1981)

deutsche-gesellschaft-fur-psychologie.csl
tobistocker: From what I just learned, it is the term "orignal-work-published". Will add this.

tobistocker: Apparently, if I use the term "originally-pubished", the validator will not accept the CSL. When I use "original-work-published", the Validator is happy. However, importing it to Zotero, it gives me an error, sayin the CSL is not valid. What I got from the debug log is this:
`failed to validate: stdin:35:6: error: attribute ^name with invalid value "original-work-published" stdin:324:16: error: attribute ^term with invalid value "original-work-published" error: invalid input `

I am riddled. It does actually work and does what it's supposed to do, but gives me an error when installing the style.
What am I missing?

tobistocker: Alright, so I checked and I can't find a single .csl in the entire repository that uses "originally-published" nor "original-work-published" (the latter would be correct, corresponding to the CSL 1.0.2 specs). However, Zotero does not like it, so I assume this is a bug within Zotero's CSL validator.

I created a report with Zotero (https://forums.zotero.org/discussion/94053/report-id-1661039569#latest).

My question is: should I implement the CSL as it should be (using the term) or would I leave it with the hard coded text element so the CSL could be used, assuming the changes might be merged at some point?

tobistocker: Zotero has not implemented CSL 1.0.2 yet and the term was introduces with this version. @bwiernik, would it be acceptable to avoid using the term and go with the hard coded text and update this later, when Zotero has implemented CSL 1.0.2?

tobistocker (commented):
The output (in-text reference as well as bibliography entry) are now correct. If the variable original-date is given, both dates will be displayed. If a translator is specified AND the variable original-title is given, the original title will also be shown in the bibliography (this is not mandatory, but both ways are ok with DGP). Hogrefe Verlag confirmed my interpretation of the manual so I would leave the final choice to the end-user.

adam3smith:
@bwiernik leaving this one for you to review and merge

bwiernik:
Thanks!

PR #5865, opened 2022-01-24
Update sodertorns-hogskola-oxford.csl

SagaNy91:
Improved macro publisher (with group), removed unnecessary macros. Fixed errors in editor (macro contributor). Included "." in layout for both citation and bibliography, which improves citing multiple publications in the same note.


adam3smith:
Looks good, thanks!

PR #5868, opened 2022-01-28
Remove publishers for articles

adam3smith:
see 
https://forums.zotero.org/discussion/94153/style-issue-between-copy-and-cite-while-write#latest

### DIFF
--- a/american-fisheries-society.csl
+++ b/american-fisheries-society.csl
@@ -91,13 +91,13 @@
           <text variable="publisher-place"/>
         </group>
       </else-if>
-      <else>
+      <else-if type="article-journal article-magazine article-newspaper" match="none">
         <group delimiter=", ">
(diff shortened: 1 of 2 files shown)


POBrien333:
We've had a few of these recently, no?

adam3smith:
Zotero is importing fields that are in the metadata but not in the Zotero item type into Extra, where they're picked up by citation styles. This is going to become a growing issue and we need to fix it systematically, I'm afraid. (While publishers are basically never cited for articles in English, this is a common requirement in Brazil and Russia, so adding them to the metadata makes sense.)

bwiernik:
There are a few other systematic changes needed (eg, changes so that Zotero can get rid of its "include URLs for paper articles" option). Do we want to discuss how we should go about doing those?

PR #5873, opened 2022-01-30
Create chicago-manual-of-style-16th-edition-latvian-RSU.csl

ksedleni:
This is a style created according to the requirements of Riga Stradins University, based on Chicago manual of style 16 and for Latvian language

### DIFF
--- a/chicago-latvian-RSU.csl
+++ b/chicago-latvian-RSU.csl
@@ -3,7 +3,7 @@
   <info>
     <title>Chicago Manual of Style 16th edition (author-date, Latvian, RSU)</title>
     <title-short>Chicago 16 LV RSU</title-short>
-    <id>http://www.zotero.org/styles/chicago-manual-of-style-16th-edition</id>
+    <id>http://www.zotero.org/styles/chicago-latvian-RSU</id>
     <link href="http://www.zotero.org/styles/chicago-manual-of-style-16th-edition" rel="self"/>
     <link href="http://www.zotero.org/styles/chicago-author-date" rel="template"/>
     <link href="https://forums.zotero.org/discussion/21077/modificated-style-kukutza-for-basque/" rel="documentation"/>
@@ -440,7 +440,7 @@
   </macro>
   <macro name="contributors">
     <names variable="author">
-      <name and="text" delimiter-precedes-last="never" et-al-min="10" et-al-use-first="10" name-as-sort-order="first"/>
+      <name and="text" delimiter-precedes-last="never" et-al-min="10" et-al-use-first="9" name-as-sort-order="first"/>
       <label form="verb-short" text-case="lowercase" prefix=", "/>
       <substitute>
         <text macro="editor"/>
--- a/chicago-latvian-rsu.csl
+++ b/chicago-latvian-rsu.csl
@@ -79,11 +79,11 @@
       <term name="retrieved">iegūts</term>
       <term name="scale">mērogs</term>
       <term name="version">versija</term>
-      <term name="open-quote">&quot;</term>
-      <term name="close-quote">&quot;</term>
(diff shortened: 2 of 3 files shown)


ksedleni:
I have no idea if this is the right place to ask this, but: what is happening next to the code? is the style now publicly available? How can it be found? If not - how to make it available?

adam3smith:
We'll review the code, likely have some question/requests for changes. Then it's (likely) merged, at which point it becomes widely available, e.g. on the Zotero style repository. (It's technically "public" already in the sense that you can point people at the file here, but I assume that's not what you had in mind)

ksedleni:
> We'll review the code, likely have some question/requests for changes. Then it's (likely) merged, at which point it becomes widely available, e.g. on the Zotero style repository. (It's technically "public" already in the sense that you can point people at the file here, but I assume that's not what you had in mind)

Thanks for reply. What I whant to do is to use this style with Mendeley Cite tool (and that other people could ot that) which requires a path to the place where the style is to be found. But I cannot figure out which path that could be.

adam3smith:
If you need a link to the current style, that's here: https://raw.githubusercontent.com/citation-style-language/styles/e681bee7adca6e407a83ef4ea7484e7427660d03/chicago-latvian-rsu.csl

- Does this have any changes other than the locale to the main 16th edition Chicago author-date? (if so which)
- This seems to copy the entire Latvian locale into the style, which isn't necessary. The locale is here: https://github.com/citation-style-language/locales/blob/master/locales-lv-LV.xml -- if something is wrong there, it should be fixed there. The locale section should only include translations particular to the Chicago author-date translation
- I assume RSU is Riga Stradins University? Is there a link to those requirements online?

bwiernik:
@ksedleni Can you answer the questions above?

ksedleni:
> If you need a link to the current style, that's here: https://raw.githubusercontent.com/citation-style-language/styles/e681bee7adca6e407a83ef4ea7484e7427660d03/chicago-latvian-rsu.csl
> 
> * Does this have any changes other than the locale to the main 16th edition Chicago author-date? (if so which)
> * This seems to copy the entire Latvian locale into the style, which isn't necessary. The locale is here: https://github.com/citation-style-language/locales/blob/master/locales-lv-LV.xml -- if something is wrong there, it should be fixed there. The locale section should only include translations particular to the Chicago author-date translation
> * I assume RSU is Riga Stradins University? Is there a link to those requirements online?

Thank you for info and questions! This version indeed has few changes. I ran side-to side comparison and found that mostly the changes are in a number of Latvian terms which are changed for other versions. It is not a question of anything being wrong, simply there are different preferences in different styles on how to do or call things. There is also a section of code that differs, but I am not such an expert in code to figure out what that chunk does. At the same time - a lot does overlap. RSU is indeed Riga Stradins University. Those requrements unfortunately are not posted online and they are only available in Latvian. Unfortunately I was not able to install in my Mendeley Cite the locale that you posted above and thus I could not test what are the differences in behaviour of those two styles.

adam3smith:
Sorry for the long wait; as a general rule, we don't accept styles for which we can't check the citation manual in some form -- it's fine if it's in Latvian. Would you be able to upload a copy here?

adam3smith:
Closing for inactivity -- we're happy to re-open this if you find time to come back to it: simply post a short message here.

ksedleni:
Hi, sorry for lack of attention to this. Attached is the manual for courseworks. See pages 10-14 for the reference style. There are some minor mistakes in the manual itself though.
[ND_metodiskie noradijumi 2021-2022 (2).docx](https://github.com/citation-style-language/styles/files/9652326/ND_metodiskie.noradijumi.2021-2022.2.docx)

PR #5879, opened 2022-02-02
Update and rename acta-psychiatrica-scandinavica.csl to dependent/act…

POBrien333:
…a-psychiatrica-scandinavica.csl

via https://forums.zotero.org/discussion/94257/the-style-repository-for-acta-psychiatrica-scandinavica-is-not-correct#latest


adam3smith:
thanks

PR #5881, opened 2022-02-02
Create hemijska-industrija.csl

dekanski:
New Style


adam3smith:
Thanks! I did remove the use of et al from the style since the author instructions explicitly say not to use it.

PR #5882, opened 2022-02-02
ISO formatted dates in Swedish Oxford citation style

Ambrosiani:
Changing date formatting to year-month-day with leading zeros.

### DIFF
--- a/sodertorns-hogskola-harvard-ibid.csl
+++ b/sodertorns-hogskola-harvard-ibid.csl
@@ -118,7 +118,7 @@
               <date variable="accessed" delimiter="-">
                 <date-part name="year"/>
                 <date-part name="month" form="numeric-leading-zeros" strip-periods="true"/>
-                <date-part name="day"/>
+                <date-part name="day" form="numeric-leading-zeros"/>
               </date>
             </group>
           </if>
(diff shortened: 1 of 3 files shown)


Ambrosiani:
ping @SagaNy91 – thoughts on this minor change?

SagaNy91 (commented):
The changes in "Quoted" work well, the form is 2022-01-03. I didn't know about the "numeric-leading-zeros" form. However, for magazine articles/newspaper articles and the other types listed besides posts, it is supposed to look like this: 1/2 2022 (it is February 1 2022).  In these cases it should not include "0" before the day or the month. <- I know that it will probably not look good when referencing a blog post (or similar) in another language than Swedish. But since I don't want to lock the style to only Swedish, I think it is okay. I can update my pull request

Ambrosiani:
ah ok, thx. Not an expert on Oxfort. Reverting "quoted" to the original. I added leading zeros to the Södertörn Harvard styles as well (although that should probably be another pull request)

SagaNy91:
> ah ok, thx. Not an expert on Oxford. Reverting "quoted" to the original. I added leading zeros to the Södertörn Harvard styles as well (although that should probably be another pull request)

Thank you. I am not the author of the Harvard styles though, so I can't speak for them.

adam3smith:
Thanks!

PR #5886, opened 2022-02-02
Create international-journal-of-cosmetic-science.csl

POBrien333:
via https://forums.zotero.org/discussion/94051/requesting-new-style-citation-international-journal-of-science-cosmetic#latest


adam3smith:
So, weir wrinkle on this one: some time last year, the journal seems to have switched over to some version of Vancouver, see e.g. https://onlinelibrary.wiley.com/doi/epdf/10.1111/ics.12760 (I looked at a bunch of articles, both research & review). It's not reflected in their guidelines, but I'm not sure if it makes sense to put up a style that they don't actually follow anymore? Maybe ask the requester to contact them and ask?

POBrien333:
Great. I've adapted the style for article-journal and chapter to match with current papers.

adam3smith:
OK, let's go with this, see what they say, thanks!

PR #5887, opened 2022-02-03
remove default-locale journal-de-la-societe-des-oceanistes.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/398551/#Comment_398551

Was hoping the automatic locale also adds space before delimiters, but alas no. This needs a 2nd style in french?


adam3smith:
I'd be inclined to just add the (non-breaking) space before colons since the requester mentions most contributions are in French. Easy to remove using search & replace and I'd like to avoid two spaces. I believe we are planning more localization for 1.1, but not sure if at that level.

bwiernik:
I thought citeproc-js did add a non breaking space before French's "big" punctuation?

adam3smith:
> I thought citeproc-js did add a non breaking space before French's "big" punctuation?

Not that I'm aware, but I could be wrong. Can you find any pointers to that?

bwiernik:
It's not documented, just hard coded into the processor somewhere. 

In 1.1, the plan is for the new punctuation terms to be substituted in for their standard counterparts by the processors

adam3smith:
Thanks!

PR #5892, opened 2022-02-04
update norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl

StianOby:
Add/improve basic handling of preparatory work using the "report" item type (until we get more item types with CSL updates).

### DIFF
--- a/norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
+++ b/norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
@@ -260,8 +260,15 @@
             <text macro="nor-legislation"/>
           </else-if>
           <else-if type="report">
-            <text macro="author-short"/>
-            <text macro="issued"/>
+            <choose>
+              <if variable="author" match="none">
+                <text variable="title"/>
+              </if>
+              <else>
+                <text macro="author-short"/>
+                <text macro="issued"/>
+              </else>
+          </choose>
           </else-if>
           <else>
             <text variable="title"/>


bwiernik:
What do you mean by "preparatory work"? Unpublished manuscripts use the `manuscript` type. Posted preprints and working papers in repositories, use the `article` type.

StianOby:
This is preparatory work for legislation (travaux preparatoires). Such preparatory works mostly consists of reports from government-appointed commissions, parliament committes, etc.

Zotero/CSL law-related item types (case/legislation/bill/etc) and fields for those item types are extremely US/common law-centric. Some improvements are coming in CSL 1.2 if I remember correctly, but there are currently no plans to take the jurisdiction-specific features of JurisM into mainline Zotero. Which is a shame, since that essentially means making mainline Zotero only fully usable for US lawyers that work purely with domestic legal sources.

For now, therefore, we have to use workarounds – and the item type "report" seems most relevant of those available.

adam3smith:
FWIW, Zotero (and CSL) is just generally not suitable for serious legal citations -- not for the US either, given different citation rules by jurisdiction there -- and Zotero devs have been pretty clear that they're not intending to implement full legal support given the scope that'd entail and expertise required. We can tinker at the margins a bit, but as juris-m shows, it really requires a whole separate infrastructure and logic.

StianOby:
Fully aware of Zotero's limitations for legal citation. Still, its better than all its competitors (apart from Juris-M, obviously). Thanks for merging!

PR #5893, opened 2022-02-05
Create trinity-college-dublin-zoology-botany-environmental-sciences-h…

POBrien333:
…arvard.csl

Style I've developed for a requestor via emails.
Requestor is Asst. Professor at TCD and style has been reviewed and signed off
[FBU44000_Thesis Writing Guidelines 2021-22 References.docx](https://github.com/citation-style-language/styles/files/8008429/FBU44000_Thesis.Writing.Guidelines.2021-22.References.docx)
.


adam3smith:
did you get a sense if these are going to be fairly permanent? I don't mind adding a style for a single department (we have a bunch) but wouldn't want this if this is for a specific cohort (the guidelines say 2021-22 in the filename, but that doesn't have to mean they're just for this year)

POBrien333:
No, this will be permanent. They've only really started with this.

PR #5894, opened 2022-02-05
Fix missing prefix for URL field

nikstur:
It seems like an empty prefix is missing for the URL field.
The screenshot shows this missing space before "Available: "

image

The change seems to fix it. Please let me know if the solution is shortsighted.


adam3smith:
Looks good thanks. The prefix is there for all other cases in the macro

PR #5896, opened 2022-02-06
Create Turabian-8th-edition-STFT-Widya-Sasana-Malang (Bahasa Indonesia).csl

chrystianmalau20:
I changed the original "Style Turabian 8th" to the Indonesian edition. This is very meaningful because it is used at the Widya Sasana College of Philosophy-Theology (STFT), Malang. I hope this will be approved. Thanks.


chrystianmalau20 (commented):
< daftar pustaka  hanging-indent = " true "  et-al-min = " 11 "  et-al-use-first = " 7 "  next-author-substitute = " ——— "  entri -spasi = " 1 " >

PR #5900, opened 2022-02-07
make journal names title case

bmichanderson:
Trying to update australian-systematic-botany via its parent style international-journal-of-wildland-fire. I'm using the style in Paperpile and full journal titles are not capitalized.


bwiernik:
Thanks!

PR #5901, opened 2022-02-07

anfourny:
…al-et-europeen.csl

I have received feedback from a user and made some changes accordingly. Mainly, there were commas added manually after some sections which led to double commas and there were missing spaces between sections.


bwiernik:
Thanks!

PR #5909, opened 2022-02-14
Update isnad-dipnotlu.csl

osmndrmz:
Update book-with editor(s) and standart item types.


bwiernik:
CSL 1.0.2 now has a proper `standard` type, so please use that instead

bwiernik:
Hi @osmndrmz, as noted at https://github.com/citation-style-language/styles/pull/5910#issuecomment-1102371470, please do not include the workaround with `book`.

adam3smith:
@bwiernik -- I'd be OK leaving the workaround in here given how poorly supported standard is in existing implementations. I've moved the test for the proper item type above it so it takes precedent. OK to merge after resolving conflicts?

bwiernik:
Sure

adam3smith:
Thanks!

PR #5910, opened 2022-02-14
Update isnad-metinici.csl

osmndrmz:
Update book-with editor(s) and standart item types.


bwiernik:
CSL 1.0.2 now has a proper `standard` type, so please use that instead

bwiernik:
Hi @osmndrmz, it doesn't look like you actually incorporated the `standard` item type into the style, only in the comment at the top? Can you please use the proper type in the style itself?

adam3smith:
Thanks!

PR #5913, opened 2022-02-16

dstark:
SBL style routinely wants a comma after an abbreviation-based citation (facilitated via annote) when the locator is sub-verbo. Other situations (sections, pages) sometimes (or often) don't include a comma.


dstark (commented):
Omits a comma after a dictionary title or abbreviation in the first citation of that source per SBLHS2.

dstark (commented):
Simplifies the DOI, URL coordination sections per [this comment](https://github.com/citation-style-language/styles/commit/0e6c74f0676a8dfca357f03a5fe886e0cced4fac#r68836799).

adam3smith:
@dstark -- is this ready for review?

dstark:
@adam3smith, it sure is. Did I miss a step in marking it as ready for review?

Thanks so much for having a look at it!

dstark:
Ah, got it yes. Noticing different things on different days did spread out the number of commits rather more than I'd anticipated.

Thanks, again, for reviewing.

society-of-biblical-literature-fullnote-bibliography.csl
```diff
@@ -137,13 +137,28 @@
...
     <choose>
       <if type="chapter paper-conference" match="any">
-        <group delimiter=", " prefix=" ">
-          <text macro="editor-translator-verb-comma"/>
-          <names variable="container-author">
-            <label form="verb" text-case="capitalize-first" suffix=" "/>
-            <name and="text" delimiter=", "/>
-          </names>
-        </group>
+        <!-- Supports citation of introducetions, prefaces, and forewords per SBLHS2 §6.2.14. -->
```
denismaier, line 140: Fix typo: introducetions->introductions

society-of-biblical-literature-fullnote-bibliography.csl
```diff
@@ -309,7 +324,18 @@
         </choose>
       </else-if>
       <else>
-        <text variable="title" quotes="true" text-case="title"/>
+        <!-- Supports citations of introductions, prefaces, or forewords per SBLHS2 § 6.2.14. -->
+        <choose>
+          <if variable="part-title">
+            <group delimiter=" ">
+              <text variable="part-title" suffix=" to "/>
```
denismaier, line 331: I know there's no `to` term, but hardcoding "to" makes the style less portable across locales. I'm not sure what would be the best solution. Maybe leave it to the authors to add "to" to `part-title`? That would still be less than ideal, but at least you could use the appropriate terms. @adam3smith  WDYT?

society-of-biblical-literature-fullnote-bibliography.csl
```diff
@@ -345,7 +371,18 @@
         </group>
       </else-if>
       <else>
-        <text variable="title" quotes="true" text-case="title"/>
+        <!-- Supports citations of introductions, prefaces, or forewords per SBLHS2 § 6.2.14. -->
+        <choose>
+          <if variable="part-title">
+            <group delimiter=" ">
+              <text variable="part-title" suffix=" to " text-case="title"/>
```
denismaier, line 378: as per above

society-of-biblical-literature-fullnote-bibliography.csl
```diff
@@ -840,9 +900,9 @@
       </else-if>
       <else-if type="entry-dictionary entry-encyclopedia" match="any">
         <choose>
-          <if position="first" variable="volume" match="all">
+          <if position="first" match="all">
             <text macro="point-locators-subsequent" prefix=" "/>
-            <!--SBL 6.3.6 prefers space in first entry before vol info.-->
+            <!-- Per SBLHS2 §6.3.6, the first reference will, by definition have the dictoinary title or abbreviation preceding the locators, volume number included. No comma should appear after the title or abbreviation and before the locators. -->
```
denismaier, line 905: typo: dictoinary->dictionary

society-of-biblical-literature-fullnote-bibliography.csl
denismaier: Yes, I think adding a "to" term to the specs would be appropriate. I'll open an issue for that, but this will probably take some time.

bwiernik: What's the context here? "to" is the verb form of `recipient`

bwiernik: Hmm, that seems a little problematic. If the part title is an actual title and not just "Introduction", you get

"John Doe, This is the title of the part not just a descriptor to A very important book, by Jane Smith ..."

denismaier: Good call. Any way to circumvent that? Maybe this: if it has a real title you'd use item type chapter, title instead of part-title, with the book's title ending up in container-title, or so?

denismaier: Also, this encourages style specific data entry, but IDK if there's a better way...

bwiernik: I would enter an Introduction as a `chapter`. That's what's appropriate and what we've advised for years and what, eg, Zotero implements with its labeling of ` chapter` as "Book Section". `part-title` isn't really conceptually appropriate. That's intended for something like "Papers on statistical inference, Part 4: Demonstrations".

This is an example of frankly silly style inconsistencies that are difficult to automate. Why should a chapter be cited treating its title as a title, but an introduction treating its title as something like a descriptor?

I don't have a great solution other than to ignore the inconsistency and just cite an introduction like any other book section. I suppose one could do something like entering "Introduction" as `genre` and having no title. But that's more style-specific data entry and would lead, eg, to it appearing as `[Introduction]` in APA style

dstark: @bwiernik, I totally agree that, conceptually, this is a needless complication in SBL style. But in this respect, SBL style is simply echoing Chicago (e.g., 17th ed., 14.110). That doesn't make it any less of a needless complication. But it's not a _discipline-specific_ needless complication. :-) Using `part-title` for this kind of situation may some have merit in connection with, e.g., CMS, 17th ed., 1.48, in the sense that it groups a section that's on the same organizational level as others with proper part tiles, even though the "introduction" or "preface" (etc.) might not have a part title stated as such. 

In principle, a better option might be to have a variable like "front-matter" that could be called for this specific situation and, therefore, avoid the problems that you point out very fairly for the use of "part-title" here.

Using `recipient` also seems less than obvious but is doubtless better than hard-coding " to ". I'll be happy to make that swap if there seems to be a workable way to accommodate the front matter type information as such.

Thanks so much for your thoughts and, @denismaier, for your careful review of these possible updates.

denismaier: `recipient` will work in English, but not in German.

denismaier:
By the way, there's another update for this style that should maybe merged first.

society-of-biblical-literature-fullnote-bibliography.csl
bwiernik: Yeah, "recipient" isn't appropriate for this case. I'll think some more of how to handle this situation

dstark:
> By the way, there's another update for this style that should maybe merged first.

@denismaier, I'm certainly happy for the updates related to front matter citation to be held back while the others are merged. Unless I'm missing something, it looks like that's the only point needing further sorting out in the discussion with @bwiernik above.

If there's something I should do to move the partial update forward (or should have done to keep things simpler in the future), please just let me know.

Thanks, again, so much.

society-of-biblical-literature-fullnote-bibliography.csl
adam3smith: @bwiernik this one is waiting for your input

adam3smith:
Merge conflicts can be confusing (this one was pretty simple, though), so I took care of that. Will look at the whole patch asap

denismaier:
I'd opt for option 2:
> accept all of these updates to the repository except for those pertaining to handling introductions, prefaces, etc.?

Let's then discuss the other changes in another PR. Ok?

denismaier:
However, there are still conflicts that have to be resolved first.

dstark:
@denismaier, that sounds good to me. I believe I've resolved the conflict that had cropped up. Please feel free to ask if I can provide anything further.

adam3smith:
Thanks! But right now the introduction-related changes are still in, right?

adam3smith:
> Do I need to submit another file with those changes pulled out? - I wasn't sure if that was the way things went or if you all with approval privileges just declined those changes.

Yes, if you could just edit this pull request accordingly? It'd take me longer to make sure that I get exactly the right changes in and out. Thanks!

dstark (commented):
Removes support for frontmatter citations based on `part-title`. Adds quotes around s.v. locators per SBL style.

adam3smith:
Thanks -- no need to touch the update stamp, btw. Once merged the update field gets set automatically for the version of the style in the [current release branch of the repository](https://github.com/citation-style-language/styles/tree/v1.0.2) (i.e. v1.0.2)

dstark:
> no need to touch the update stamp, btw

That's nice to know. Thanks for the additional information, and the merge!

dstark:
@denismaier, could you point me to the issue you mentioned opening about adding a `to` term? Although, as you've noted, it might take some time, I'd just like to keep in the loop on that possible update if possible. Thanks so much!

denismaier:
sure, here it is: https://github.com/citation-style-language/schema/issues/414

dstark:
> sure, here it is: [citation-style-language/schema#414](https://github.com/citation-style-language/schema/issues/414)

Thanks so much!

PR #5914, opened 2022-02-17
Update organization-studies.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/399632/#Comment_399632

add publisher-place for book-like items

edit: the Style Editor still had the independent style, so this is wrong. Not sure what to do with this style now


bwiernik:
Yes. With these changes it needs to be independent

bwiernik:
Thanks!

PR #5919, opened 2022-02-22

POBrien333:
https://forums.zotero.org/discussion/94662/newspaper-citations-in-society-of-american-archaeology#latest

Might not be the right thing.


bwiernik:
Looks reasonable enough for now. Thanks!

PR #5921, opened 2022-02-22
Added fixes for steel-research-international.csl

c-schubert:
- correct use of super-case for in-text citations
- added month for paper-conference
- added page/page-first for journal or article types.

### DIFF
--- a/steel-research-international.csl
+++ b/steel-research-international.csl
@@ -22,11 +22,12 @@
     
     There are some difficulties i.e. with patents, preprint  specifications and the variables that the CSL offers (nor preprint type or company variable availably), which probably cannot be fixed right now (perfectly) and need manual correction.
     Also there is no way to automatically use or define journal or conference abbrevitation in the csl file right now?</summary>
-    <updated>2021-11-10T16:22:30+01:00</updated>
+    <updated>2022-02-22T10:00:30+01:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en-US">
     <date form="text" delimiter=" ">
+      <date-part name="day" suffix=" " range-delimiter="-" form="numeric-leading-zeros"/>
       <date-part name="month" form="short" strip-periods="true" suffix=" "/>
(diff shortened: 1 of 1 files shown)


bwiernik:
Thanks! Could you also take a look at the issues raised in the "summary" section of the style? CSL has a `patent` type and a type for preprints (`article`), so all of those issues can be addressed. Abbreviations for journals and conferences are called using `form="short"` with `container-title`. Automatic generation of abbreviations is handled by the citation program, not the CSL style.

bwiernik:
Look at line 23 of the style file

c-schubert:
Ok, I am sorry. I totally forgot about that. I added some explanation to the issue with patent and preprints. As, I think it cannot be solved perfectly atm. Many thanks :)

bwiernik:
For preprints, the item type is `article`, the repository is `publisher`, and the archive ID is `number`. 

For the patent company, does this typically refer to the Assignee of the patent, or is that another role?

c-schubert:
Preprint: Alright, that should be no problem then, what variable (if any) is usually used for the version(number) `v1,v2 ...` of a preprint? I guess `version`  variable could be used?

Patent: In my understanding, the `publisher` would more correspond to the association, which granted the patent. I think the company or person who owns a patent is called "patent owner". In my understanding for patents the patent inventors must be persons, the patent owner can be (a) person(s) or a company.

bwiernik:
So when you are referring to the company, you mean the owner of the patent?

Yes `version` refers to the version for all item types

adam3smith:
Owner is assignee in English, so we have that I think?

bwiernik:
There's a field in Zotero, but it's not mapped to anything. I think CSL `publisher` makes sense as the patent office would be `authority`. Or perhaps `contributor`?

adam3smith:
I don't like publisher - it's systematically off (the assignee doesn't publish anything)  and also in terms of citation practice: publisher is almost always cited, assignees quite rarely, so this would require constant exceptions for patents in style code. Contributor seems fine, if not ideal

bwiernik:
Between contributor and waiting until CSL 1.1 when we could make a new `assignee` variable, what do you think?

adam3smith:
Id almost be inclined to wait. The demand for this in citations isn't huge

PR #5923, opened 2022-02-23
Update antiquites-africaines.csl

samarkar:


bwiernik (changes requested):
Thanks! Please don't split an element other than summary across multiple lines

samarkar:
Thank you very much Brenton for your help!

PR #5927, opened 2022-02-24
MLA Publication Date should include day of month

LouieK22:
Apologies in advance if this is not the correct process for contributing. I'm normally a software engineer, not an English major.

In referencing the MLA 9th edition style guide, it appears as though the publication date should include the day of the month in the date format. See these examples on Purdue OWL: https://owl.purdue.edu/owl/research_and_citation/mla_style/mla_formatting_and_style_guide/mla_works_cited_periodicals.html

### DIFF
--- a/modern-language-association.csl
+++ b/modern-language-association.csl
@@ -205,8 +205,8 @@
       <if type="book chapter paper-conference motion_picture" match="any">
         <date variable="issued" form="numeric" date-parts="year"/>
       </if>
-      <else-if type="article-journal article-magazine" match="any">
-        <date variable="issued" form="text" date-parts="year-month-day"/>
+      <else-if type="article-journal" match="any">
+        <date variable="issued" form="text" date-parts="year-month"/>
       </else-if>
       <else-if type="speech" match="none">
         <date variable="issued" form="text"/>
--- a/modern-language-association.csl
+++ b/modern-language-association.csl
@@ -206,7 +206,7 @@
         <date variable="issued" form="numeric" date-parts="year"/>
       </if>
       <else-if type="article-journal article-magazine" match="any">
-        <date variable="issued" form="text" date-parts="year-month"/>
+        <date variable="issued" form="text" date-parts="year-month-day"/>
       </else-if>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thank you -- definitely the right way to contribute, yes.

I'm less sure this is correct for MLA. OWL says to use day-month-year for magazines and year for periodicals. In MLA8 I'm seeing "what's printed on the cover" as instructions, which is not super helpful. I think I'd be inclined to just remove article-magazine from that line so that it defaults to the full date. article-journal should definitely not be the full date, both manual and OWL agree on that. I'd keep the month there for now, though, given the examples I'm seeing in MLA 8, though if someone has the 9th edition at hand, happy to change that to just year if appropriate.

LouieK22:
I'm gonna run and grab the MLA 9th edition from the library and get back to this next week.

LouieK22:
Okay, so after a look-through of the MLA 9th edition handbook, here's what I was able to figure out. The publication date that is used in the Works Cited should be the date given on the source. Except for...
- e-books should only give the year even if the source gives a more specific date, this maintains consistency with normal books (are these mediums even separate in CSL?)
- government documents should include the date only and follow the "principle for publication dates of e-books and other nonperiodical works", except for government webpages which should include the full date according to the example given

To me it doesn't seem like even MLA knows how the dates are supposed to be formatted judging by some of the contradictory wording used in their own handbook.

I think if we do year only for ``book chapter paper-conference motion_picture`` and ``year-month-day`` for everything else, while keeping the special case for ``speech``, we *should* be fine. I'm not sure how pedantic most people will be, but this might just be a point where we have to outsource some responsibility to the user to use the "correct" date.

bwiernik:
It doesn’t seem like day should apply to article-journal though. What do you think @adam3smith ?

adam3smith:
Yeah, I think this makes most sense: print the whole date for magazine articles but stick with year-month for journal articles (I agree the rules here are a bit weird/impenetrable, but I'd also expect them to not be too strict about this as long as the citations are reasonable).

PR #5928, opened 2022-02-25
Update archaeonautica.csl

samarkar:


archaeonautica.csl
```diff
@@ -1,24 +1,23 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
   <info>
-    <title>Archaeonautica</title>
-    <title-short>Archaeonautica</title-short>
-    <id>http://www.zotero.org/styles/archaeonautica</id>
+    <title>Archaeonautica_encours</title>
+    <title-short>Archaeonautica_encours</title-short>
+    <id>http://www.zotero.org/styles/archaeonautica_encours</id>
```
bwiernik:
```suggestion
    <title>Archaeonautica</title>
    <title-short>Archaeonautica</title-short>
    <id>http://www.zotero.org/styles/archaeonautica</id>
```

bwiernik (changes requested):
Thanks! It looks like there may be some duplicated punctuation added in a few places because of the includes of the . in both prefixes and suffixes. 

For this reason, it's best practice to avoid things like `". "` as a prefix or a suffix. Instead, use a `<group delimiter=". ">`. 

For example, it would be best in this style if the bibliography looked like this:

```
 <group suffix="." delimiter=". ">
        <text macro="year-date"/>
        <text macro="title"/>
        <text macro="translator"/>
        <text macro="edition"/>
        <text macro="event"/>
        <text macro="place"/>
        <text macro="publisher"/>
        <text macro="collection"/>
        <text macro="page-bib"/>
        <text macro="archive"/>
      </group>
```

And the individual macros did not include any prefixes or suffixes holding the spaces or dots. 

Do you think you could go through the style and clean that up?

bwiernik:
In line 219, can you please delete `    <text term="translator" form="verb" prefix=", " suffix=" "/>` and instead use a `<label>` inside the `<names>` element?

samarkar:
Dear Brenton, thank you very much for your patience and your careful proofreading. I am correcting the style according to your indications but, concerning your last remark, what is your suggestion? I would like to keep the possibility of automatic translation of the  name variable Translator.

bwiernik:
@samarkar Sorry on the delay. You have misunderstood. A `<label>` is still automatically translated. `<label>` is appropriate way to add role terms to a name.

bwiernik:
@samarkar The `<label>` is now fine as-is.

However, looking at the example citations rendered above, there are several formatting issues (e.g., in the first example, there is no space between "Editor." and the URL). I would like you to please go through the style carefully and consistently use groups and delimiters to control punctuation and spaces, rather than prefixes/suffixes. Please carefully test your style to ensure that the output is correct.

adam3smith:
I fixed the missing space before the URL. Would indeed be good to take a closer look at this, but overall this seems like an improvement, so taking it.

PR #5935, opened 2022-03-02
Create CRCAO Light

PhilPons:


adam3smith:
We already have a style for CRCAO -- could you clarify how this differs and what the relationship to the other style is?

PhilPons:
The CRCAO style is a style with references in notes, while this new style will be used for in-text references. This complementary style is also intended to serve as a model for the publications in the "Recherches sur les civilisations de l'Asie orientale" series ([https://books.openedition.org/cdf/9748](https://books.openedition.org/cdf/9748)).

PhilPons:
This style is validated by the CSL Validator. So I'm sorry, but I don't understand why the style is not accepted?

bwiernik:
Please rename the file to crco-

bwiernik:
Please re-read my comment above carefully. The filename needs to be `crcao-light-intext.csl`

PhilPons:
Yes, I'm sorry. I read your full comment just after. It must be better now?

bwiernik:
Okay, that is resolved. 

But following up on @adam3smith's comment above. There is already http://www.zotero.org/styles/centre-de-recherche-sur-les-civilisations-de-l-asie-orientale on the repository. This CRCAO style is also an author-date style based on APA.

So, is this new style an update for that existing style, or what exactly is the relationship? The author-date CRCAO style has existed for some time already.

PhilPons:
The CRCAO style has indeed been around for some time. But it is a "note" style for the use of CRCAO researchers. It is not an author-date style.
On the other hand, this new style is an author-date style which is specifically intended for use in the publications of the "Recherches sur les civilisations de l'Asie orientale" series (https://books.openedition.org/cdf/9748).
Both are really important to CRCAO.

bwiernik:
My apologies, I had misread the other CRCAO style (too many open tabs) and thought it was also in-text.

bwiernik:
Thanks!

PR #5936, opened 2022-03-02
Create ascomycete.org

franvala:


bwiernik:
Can you please rename the style file to `ascomyocete_org.csl`?

It's currently missing the `.csl` part at the end, and please replace `.org` in the filename and Style ID with `_org`

PR #5937, opened 2022-03-02
Create societe-mycologique-de-france

franvala:


bwiernik:
Can you please rename the style file to `societe-mycologique-de-france.csl`?

It's currently missing the `.CSL` part at the end.

PR #5943, opened 2022-03-03
Create taylor-and-francis-national-library-of-medicine-XUND BMA HF.csl

hum84:
Hello! would it be possible to share this style in zotero?


hum84:
does somebody can help me to change a csl style to my format and help to load it into zotero?

bwiernik:
@hum84 Thanks!

Per @adam3smith's comment here: https://github.com/citation-style-language/styles/issues/5944#issuecomment-1058725628, I'm not sure if this style is really needed? There are several existing Vancouver styles with brackets that can be set to German https://www.zotero.org/styles?q=vancouver%20(brackets

If this isn't sufficient, can you please change the documentation link in this style to the actual style guidelines for your style?

adam3smith:
Closing this for inactivity -- happy to re-open if you get back to us on the question above.

PR #5950, opened 2022-03-04

alex-ball:
- Updates formatting to align with summer 2021 changes to the specification.
- Introduces support for some CSL v1.0.2 entry types.


adam3smith:
Thanks! Note that for Zotero, the style will currently throw a warning when installed, bc they haven't updated their built-in validator. It will still work, though. I don't think other tools check validation.

PR #5951, opened 2022-03-05
Update harvard-limerick.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/400562/#Comment_400562

is the style name adequate?


adam3smith:
Yeah, I think style name works

adam3smith:
Thanks!

PR #5954, opened 2022-03-08

joelnitta:
Fixes disambiguation of et al names (should be "false").

An editor just changed this in my MS. For example, "Nitta, Ebihara, et al. (2020)" should be "Nitta et al. (2020a)".


bwiernik:
Thanks!

PR #5957, opened 2022-03-09
Create mary-ann-liebert-harvard.csl

POBrien333:
via citationstyles.org form >> forwarded by Sebastian to me.

I will also have dependent styles to go with these styles.


adam3smith:
Thanks! As we generate the dependents, will we also need to delete independent ones?

POBrien333:
Everything is up in this PR: https://github.com/citation-style-language/journals/pull/127

How do you know which styles we already have and need deleting?

PR #5960, opened 2022-03-10
Create journal-of-macromarketing.csl

POBrien333:
via https://forums.zotero.org/discussion/91364/journal-of-macromarketing/p1


adam3smith:
Thanks!

PR #5961, opened 2022-03-11
Create expert-reviews-in-molecular-medicine.csl

POBrien333:
via https://forums.zotero.org/discussion/92687/expert-reviews-in-molecular-medicine#latest

The in-text "Ref." or "Refs" cannot be done with CSL, correct?

They want "....blal (Ref. 1) said that was confirmed by everyone else (Refs 2-5)."

### DIFF
--- a/expert-reviews-in-molecular-medicine.csl
+++ b/expert-reviews-in-molecular-medicine.csl
@@ -1,0 +1,201 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>Expert Reviews in Molecular Medicine</title>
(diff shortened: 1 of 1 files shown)


adam3smith:
> The in-text "Ref." or "Refs" cannot be done with CSL, correct?

We could add "Ref." to the layout prefix, but we can't make the singular/plural distinction, so I think better to leave that for the user to add as prefix or so. Not ideal, but...

adam3smith:
Looking at the author guidelines, I do wonder whether we shouldn't use square brackets as explicitly requested there? Always hard to say when those are wrong & when journals change this in typeset. I'd be somewhat inclined to go with the latter assumption here?

POBrien333:
It's a 50/50 chance. I mean, if editors request it out hopefully a user reports it and we can just change it.

bwiernik:
@POBrien333 Let’s go with the brackets as per the style guidelines

bwiernik:
Thanks!

PR #5964, opened 2022-03-16
Update the-journal-of-egyptian-archaeology.csl

POBrien333:
fix error with subsequent locators showing label.
Came in via emails.

> I found your name as author in this file and wonder if you are supporting/maintaining it and if you can assist?  I think I found a bug in it:
> I am using MS Word 365 and downloaded latest version of Paperpile (beta) yesterday which includes your CSL file by default.
>There are a few little difficulties, and I'm not sure whether some of them are Paperpile or the CSL but the following seems to be the CSL:
>The first reference shows the page number without the P. correctly.  e.g.
>[1] J. Quibell, Excavations at Saqqara 1911 – 1912: The Tomb of Hesy (Cairo, 1913), 19.

>Subsequent references include "P." - which they should not.
>[1] Quibell, Excavations at Saqqara 1911 – 1912: The Tomb of Hesy, p. 19.

### DIFF
--- a/the-journal-of-egyptian-archaeology.csl
+++ b/the-journal-of-egyptian-archaeology.csl
@@ -206,7 +206,7 @@
         <choose>
           <if variable="URL" match="any">
             <group delimiter=" ">
-              <text variable="URL" prefix="&lt;" suffix="&gt;"/>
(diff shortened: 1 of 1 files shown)


bwiernik:
Thanks!

PR #5965, opened 2022-03-16
update USITC.csl

USITC-Zotero:
Hello, I would like to request an update to the USITC.csl file. We resolved the following bugs

1) News article sections needed to be removed from footnotes
2) Abstracts sometimes blocked short-titles due a buggy logical condition
3) Added legal citation information

Thank you,
USITC-Zotero
[USITC_v1.2.zip](https://github.com/citation-style-language/styles/files/8262384/USITC_v1.2.zip)


USITC-Zotero:
Perhaps, I put this in the wrong location? Let me try a second time elsewhere.

PR #5968, opened 2022-03-16
Changed quotation marks to Spanish ones

ruben2s:
As stated by the Royal Spanish Academy in "Ortografía de la lengua española" (2010 edition, page 386), spanish quotation marks (« ») should be used instead of the english ones (“ ”).
This can also can be checked online in https://www.rae.es/dpd/comillas.


POBrien333:
You can actually set `quotes="true"` on those variables instead of using prefix/suffix. As the locale is set to Spanish, it will use the correct quotation marks. This makes for a more robust style and avoid some other little issues that could pop up.

PR #5970, opened 2022-03-17
Update american-medical-association.csl

POBrien333:
via https://forums.zotero.org/discussion/95093/reference-output-for-thesis-item-is-completely-wrong-using-ama-style#latest

Didn't check this in guidelines and just based it on what OP posted.

Let me know if this looks good and then I get the other AMA versions updated as well. :)

### DIFF
--- a/american-medical-association-alphabetical.csl
+++ b/american-medical-association-alphabetical.csl
@@ -86,7 +86,7 @@
   </macro>
   <macro name="title">
     <choose>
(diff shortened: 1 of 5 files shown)


adam3smith:
Thanks, yes, looks exactly right

POBrien333:
Great, that's all done. Hope I didn't mess anything up :)

PR #5977, opened 2022-03-22

YoungZoe:
change demote non dropping particle to never as we realised that names such as van Gogh are appearing under G rather than V. We want them appearing under V.


bwiernik:
Thanks!

PR #5984, opened 2022-03-26
improve journal-of-avian-biology.csl

POBrien333:
via https://forums.zotero.org/discussion/95691/update-citation-style-journal-of-avian-biology#latest


adam3smith:
Thanks!

PR #5986, opened 2022-03-27
Create ethnobiology-letters.csl

yoblo99:
First Zotero style for Ethnobiology Letters, a journal published by the Society of Ethnobiology. https://ojs.ethnobiology.org/index.php/ebl


bwiernik:
Thanks!

PR #5997, opened 2022-03-30
Fix annales-medico-psychologiques

symac:
Up to this PR, annales-medico-psychologiques was a dependent style of elsevier-vancouver. But the [instructions to authors](https://www.elsevier.com/journals/annales-medico-psychologiques-revue-psychiatrique/0003-4487/guide-for-authors) indicate that references should be alphabetically sorted, which can be confirmed by looking at articles from this journal where references are numbered based on the author alphabetically sorted list rather than the order of appearance in the article.

This PR is removing the dependent style and creating a new one that is mainly a copy of the elsevier-vancouver one, adding a sort option in the bibliography.


annales-medico-psychologiques.csl
```diff
@@ -0,0 +1,153 @@
...
+    <sort>
+      <key variable="citation-number"/>
+    </sort>
+    <layout prefix="[" suffix="]" delimiter=",">
+      <text variable="citation-number"/>
+    </layout>
+  </citation>
+  <bibliography entry-spacing="0" second-field-align="flush" et-al-min="7" et-al-use-first="6">
+    <sort>
+      <key variable="author" sort="ascending"/>
```
denismaier, line 80: I think you should add a second sort key if the author key is identical. (Probably using `title`?)

denismaier: Also, check how references without authors are sorted.

denismaier (changes requested):

bwiernik:
@symac Can you please address the questions above? Thanks!

adam3smith:
@symac -- I'm closing here to keep our PR queue manageable. Happy to re-open, just let us know by leaving a comment here.

PR #5999, opened 2022-04-01
cell.csl: Add a macro for DOI and webpage accessed date.

nvall:
Hi,
- I recently found that Cell journals require the use of DOI at the end of the reference list see: https://www.cell.com/cell/authors#format `We encourage the inclusion of DOIs in all journal article references`
- Some references are sometimes from webpages and it is better to add the accessed date.

This pull request adds a macro to display DOI for articles and Accessed date to webpages.

Hope this will be usefull,
Best,
Nicolas

### DIFF
--- a/cell.csl
+++ b/cell.csl
@@ -58,13 +58,9 @@
     </names>
   </macro>
   <macro name="doi">
-      <if type="article-newspaper" match="any">
-        <choose>
-          <if variable="DOI">
-            <text variable="DOI" prefix="https://doi.org/"/>
-          </if>
-        </choose>
-      </if>
+        <if variable="DOI">
+          <text variable="DOI" prefix="https://doi.org/"/>
+        </if>
   </macro>
   <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true" collapse="year">
     <sort>
--- a/cell.csl
+++ b/cell.csl
@@ -57,34 +57,14 @@
       <label form="short" prefix=", "/>
     </names>
   </macro>
-  <macro name="accessed-doi">
-    <choose>
+  <macro name="doi">
       <if type="article-newspaper" match="none">
         <choose>
           <if variable="DOI">
             <text variable="DOI" prefix="https://doi.org/"/>
           </if>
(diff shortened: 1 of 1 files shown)


bwiernik:
Thanks! Do you have an actual example of a website citation in the journal having access dates? Most citation styles do _not_ include access dates for website citations. So unless there is information otherwise for Cell, please remove that.

cell.csl
```diff
@@ -57,6 +57,36 @@
...
     </names>
   </macro>
+  <macro name="accessed-doi">
+    <choose>
+      <if type="article-newspaper" match="none">
+        <choose>
+          <if variable="DOI">
+            <text value=" doi:"/>
+            <text variable="DOI"/>
+          </if>
```
bwiernik, line 67:
```suggestion
          <if variable="DOI">
            <text variable="DOI" prefix="https://doi.org/"/>
          </if>
```

This is the preferred format from CrossRef

bwiernik (changes requested):

nvall:
Thanks for your suggestions. I have remove URL part of the changes and added a bit of formating so it fits with Cell guidelines.

nvall (commented):
Thanks, I pushed a new version with DOI corrected and URL removed.

cell.csl
```diff
@@ -57,6 +57,13 @@
       <label form="short" prefix=", "/>
     </names>
   </macro>
+  <macro name="doi">
+    <choose>
+      <if variable="DOI">
+        <text variable="DOI" prefix="https://doi.org/"/>
+      </if>
+    </choose>
```
bwiernik, line 65:
```suggestion
    <choose>
      <if variable="DOI">
        <text variable="DOI" prefix="https://doi.org/"/>
      </if>
      <else-if type="webpage post post-weblog" match="any">
        <text variable="URL"/>
      </else-if>
    </choose>
```

bwiernik (changes requested):
I think it would be good to include the URL for webpages if they don't have a DOI. Otherwise there is basically no retrieval information at all

bwiernik:
Thanks!

PR #6005, opened 2022-04-05
Render status field (for forthcoming, in press, etc.) to date groups

mikebader:
Based on [this Zotero forum thread](https://forums.zotero.org/discussion/68198/apa-style-in-press-articles), I added an option to print the status field if issued does not exist and status does exist (and still defaults to 'n.d.' if neither exists) using the code in the [APA style](https://github.com/citation-style-language/styles/blob/51ff3dc6caa528d0658e2b36c35ab1b298511197/apa.csl#L477-L481).


bwiernik:
Thanks!

PR #6011, opened 2022-04-07

alex-ball:
Due to differences in how `<cs:substitute>` and repeated variables are handled by different processors, this update removes a reliance on `<cs:substitute>` and updates certain tests accordingly.

Relates to alex-ball/bathbib#16. Relates to jgm/citeproc#101.

### DIFF
--- a/harvard-university-of-bath.csl
+++ b/harvard-university-of-bath.csl
@@ -322,7 +322,7 @@
           <date variable="issued">
             <date-part name="day" suffix=" "/>
             <date-part name="month"/>
-          </date>  
+          </date>
         </if>
       </choose>
       <group>
@@ -540,14 +540,14 @@
               <choose>
                 <if variable="URL DOI accessed" match="any">
                   <text term="online" text-case="capitalize-first" prefix="[" suffix="]"/>
-                </if>      
+                </if>
               </choose>
             </else-if>
             <else-if variable="archive" match="none">
               <choose>
                 <if variable="URL DOI accessed" match="any">
                   <text term="online" text-case="capitalize-first" prefix="[" suffix="]"/>
-                </if>      
+                </if>
               </choose>
             </else-if>
           </choose>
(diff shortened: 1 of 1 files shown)


adam3smith:
@bwiernik -- you had been suggesting a possible alternative to this over on pandoc. I don't love getting rid of substitute because of processor differences, do you want to take a look here?

bwiernik:
Yes as far as I can see, there isn't a problem. The `<substitute>` just needs to call a macro, rather than the `title` variable alone. 

From https://github.com/alex-ball/bathbib/issues/16, something like:

```
<macro name="title-plus-extra"/>
  <group delimiter=" ">
    <text variable="title" font-face="italic"/>
    <choose>
      <if variable="medium">
        <text variable="medium" prefix="[" suffix="]"/>
      </if>
      <else-if variable="URL">
        <text term="online" text-case="capitalize-first" prefix="[" suffix="]"/>
      </else-if>
    </choose>
  </group>
</macro>
<names variable="author">
  <substitute>
    <names variable="editor"/>
    <text macro="title-plus-extra"/>
  </substitute>
</names>
```

alex-ball:
In the most recent automated test, I think the delimiter gets lost in a `<group><choose><if><text/><text/></if></choose></group>` construct. This doesn't happen in my local testing and I think goes against the CSL spec. Is this a known (isolated) problem with the test or something I should guard against in the style?

denismaier:
Just for reference, there's been a commit that changes this behaviour in pandoc's citeproc: https://github.com/jgm/citeproc/commit/2010df5a760aaac751e0575552382ee2e27ef0c4

The latest release happened after that commit, but it's not mentioned in the [changelog](https://github.com/jgm/citeproc/blob/master/CHANGELOG.md).

alex-ball:
Thanks, I spotted that too after my last amendment to this PR.

I have managed to get it down (as far as I can tell) to one minor edge case where this behaviour makes a difference, but it favours the current pandoc implementation; I will try and flip it so favours citeproc-js instead, in preparation for that commit being merged.

denismaier:
Update: the branch has been merged yet so `substitute` should work as expected (with the next version of pandoc).

adam3smith:
Thanks -- sorry for the delay. Feel free to post reminders after ~2 weeks, this shouldn't have waited a whole months for final review & merge.

PR #6013, opened 2022-04-10
harvard-university-of-the-west-of-england.csl

Bunniz:
et al has been italicised and the order of grouped in-line citation has also been changed to descending order based on published year of the reference, as per UWE referencing style requirements.


bwiernik:
Thanks!

Issue #5916, opened 2022-02-21
bibtex.csl missing fields


sebhofmann:
I discovered that bibtex offers a keywords field and csl offers a keywords field, but its not used in the bibtext style.
The language field is also missing.

https://github.com/citation-style-language/styles/blob/master/bibtex.csl

bwiernik:
These could potentially be added, but you really should avoid using this CSL style if at all possible and instead use a proper BibTeX exporter

PR #6021, opened 2022-04-14
Update bibtex

bwiernik:
Closes https://github.com/citation-style-language/styles/issues/5916

@retorquere 'place' isn't a BibTeX field is it? Do some BibTeX applications use it?


retorquere:
It's sometimes used as an alias for location, but I don't think many styles support it.

bwiernik:
Thanks!

PR #6026, opened 2022-04-17
Liebert update

adam3smith:
See https://github.com/citation-style-language/journals/pull/127


POBrien333:
thank you for merging this! Will get back to Heather now all her styles are up. :)

PR #6027, opened 2022-04-19

POBrien333:
via https://forums.zotero.org/discussion/91633/style-request-netherlands-journal-of-geosciences-geologie-en-mijnbouw#latest

### DIFF
--- a/netherlands-journal-of-geosciences-geologie-en-mijnbouw.csl
+++ b/netherlands-journal-of-geosciences-geologie-en-mijnbouw.csl
@@ -1092,7 +1092,7 @@
         <else>
           <group>
             <text term="volume" form="short" plural="true" text-case="capitalize-first" suffix=" "/>
-            <number variable="number-of-volumes" form="numeric" prefix="1–"/>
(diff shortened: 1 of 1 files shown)


POBrien333:
@bwiernik 
I fixed the style as per the above thread, but get the number-of-pages duplicated like such:
`Isaacson, W., 2011. Steve Jobs (630 pp.). Simon & Schuster (New York, NY): 630 pp.`

Do you have an idea how to tackle this issue?

bwiernik:
You have `<text macro="locators"/>` on line 812, but `locators` is also called as part of the condition at line 527 in the `description` macro:

```
        <else-if type="book interview manuscript motion_picture musical_score pamphlet post-weblog speech webpage" match="any">
          <group delimiter="; ">
            <text macro="locators"/>
          </group>
        </else-if>
```

Either remove `locators` from there or add a new condition without it for the offending types.

POBrien333:
yup :)

bwiernik:
Thanks!

PR #6031, opened 2022-04-19
Adds space after title writing in ABNT style file

JhonathanLepidus:
The targeted ABNT style file has a problem in its title writing. It doesn't include a space after it, leading to `title` and `container-title` being shown gathered in the citation. Here is a made-example:

```
CABRAL, J. F.; Factors associated with functional disability in older adultsSciELO Preprints
```

After our correction, it is printed like:

```
CABRAL, J. F.; Factors associated with functional disability in older adults SciELO Preprints
```

### DIFF
--- a/universidade-de-sao-paulo-escola-de-comunicacoes-e-artes-abnt.csl
+++ b/universidade-de-sao-paulo-escola-de-comunicacoes-e-artes-abnt.csl
@@ -145,7 +145,7 @@
         <text variable="title" font-weight="bold"/>
       </else>
     </choose>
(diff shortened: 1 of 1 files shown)


bwiernik:
Thanks! Please use a group with a delimiter rather than this approach. See the bibliography section of APA.CSL for an example. 

Should the separator not be something like `. ` or `; ` rather than a space alone?

bwiernik:
Thanks

JhonathanLepidus:
Thank you!

PR #6035, opened 2022-04-22
style for Revista Ladinia

schtailmuel:
Submission for the style of the "Ladinia" journal [1] published annually by the Ladin Institute “Micurá de Rü”.
It is based on APA [2] and includes adaptations according to the guidelines for authors [3].

[1] [https://www.micura.it/en/activities/ladinia](https://www.micura.it/en/activities/ladinia)
[2] [http://www.zotero.org/styles/apa](http://www.zotero.org/styles/apa)
[3] [https://www.micura.it/en/activities/ladinia/style-sheet](https://www.micura.it/en/activities/ladinia/style-sheet)

### DIFF
--- a/revista-ladinia.csl
+++ b/revista-ladinia.csl
@@ -19,7 +19,7 @@
 		<category citation-format="author-date"/>
 		<category field="linguistics"/>
 		<issn>1124-1004</issn>
-		<updated>2022-04-22T12:16:36+01:00</updated>
+		<updated>2022-04-25T13:40:00+01:00</updated>
 		<rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
 	</info>
 	<locale xml:lang="it">
@@ -27,7 +27,8 @@
 			<term name="and others"></term>
 			<term name="no date">s.a.</term>
 			<term name="no-place">s.l.</term>
-			<term name="no-publisher">N.N.</term>
+			<term name="anonymous">N.N.</term>
+			<term name="volume">vol.</term>
 			<term name="editor">
 				<single>ed.</single>
 				<multiple>eds.</multiple>
@@ -39,7 +40,8 @@
 			<term name="and others"></term>
 			<term name="no date">s.a.</term>
 			<term name="no-place">s.l.</term>
-			<term name="no-publisher">N.N.</term>
+			<term name="anonymous">N.N.</term>
+			<term name="volume">vol.</term>
 			<term name="editor">
 				<single>ed.</single>
 				<multiple>eds.</multiple>
@@ -51,7 +53,8 @@
 			<term name="and others"></term>
 			<term name="no date">s.a.</term>
 			<term name="no-place">s.l.</term>
-			<term name="no-publisher">N.N.</term>
+			<term name="anonymous">N.N.</term>
+			<term name="volume">vol.</term>
 			<term name="editor">
 				<single>ed.</single>
 				<multiple>eds.</multiple>
@@ -62,7 +65,7 @@
 		<choose>
 			<if variable="author">
 				<names variable="author">
-					<name form="short" et-al-use-first="1">
+					<name name-as-sort-order="all" delimiter="/" form="short">
 						<name-part name="family" font-variant="small-caps"/>
 					</name>
 					<et-al term="and others" />
@@ -70,15 +73,14 @@
 			</if>
 			<else-if variable="editor">
 				<names variable="editor">
-					<name name-as-sort-order="all" delimiter="/">
+					<name name-as-sort-order="all" delimiter="/" form="short">
 						<name-part name="family" font-variant="small-caps"/>
 					</name>
-					<label prefix=" (" suffix=")"/>
 					<et-al term="and others" />
 				</names>
 			</else-if>
 			<else>
-				<text term="no-publisher"/>
+				<text term="anonymous"/>
 			</else>
 		</choose>
 	</macro>
@@ -102,7 +104,7 @@
 				</names>
 			</else-if>
 			<else>
-				<text term="no-publisher"/>
+				<text term="anonymous"/>
 			</else>
 		</choose>
 	</macro>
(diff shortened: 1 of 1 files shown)


revista-ladinia.csl
```diff
@@ -0,0 +1,163 @@
...
+        <single>ed.</single>
+        <multiple>eds.</multiple>
+      </term>
+    </terms>
+  </locale>
+  <macro name="author">
+    <choose>
+      <if variable="author">
+        <names variable="author">
+          <name form="short" et-al-use-first="1">
```
adam3smith, line 64: No need for the et-al bit here

revista-ladinia.csl
```diff
@@ -0,0 +1,163 @@
...
+          <date-part name="year"/>
+        </date>
+      </if>
+      <else>
+        <text term="no date"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="publisher-place">
+    <choose>
```
adam3smith, line 121: I think you'll want to restrict this macro to some item types -- e.g. you don't want s.l. for journal articles.

revista-ladinia.csl
```diff
@@ -0,0 +1,163 @@
...
+    <sort>
+      <key macro="author-bibliography"/>
+      <key macro="issued-year"/>
+      <key variable="title"/>
+    </sort>
+    <layout suffix=".">
+      <text macro="author-bibliography" suffix=": "/>
+      <group delimiter=", ">
+        <text variable="title" font-style="italic"/>
+        <text variable="container-title" prefix="in: "/>
```
adam3smith, line 152: are these not localized? And even if not, could we make this

```suggestion
       <group delimiter=": ">
          <text term="in"/>
          <text variable="container-title" />
        </group>
```

revista-ladinia.csl
```diff
@@ -0,0 +1,163 @@
...
+        <text variable="volume" prefix="vol. "/>
```
adam3smith, line 153: ditto on using a term rather than a prefix here, too.

adam3smith (changes requested):
Thanks! A number of questions & requests inline

revista-ladinia.csl
```diff
@@ -0,0 +1,163 @@
...
+  </macro>
+  <citation et-al-min="2" et-al-use-first="1" initialize-with="." disambiguate-add-givenname="true" disambiguate-add-year-suffix="true" collapse="year">
+    <sort>
+      <key macro="author"/>
+      <key macro="issued-year"/>
+    </sort>
+    <layout prefix="" suffix="" delimiter=", ">
+      <group delimiter=" ">
+        <text macro="author"/>
+        <text macro="issued-year" prefix="(" suffix=")"/>
```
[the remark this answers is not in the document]
schtailmuel: you are right, I will change that

revista-ladinia.csl
```diff
@@ -0,0 +1,163 @@
...
+          </name>
+          <et-al term="and others"/>
+        </names>
+      </if>
+      <else-if variable="editor">
+        <names variable="editor">
+          <name name-as-sort-order="all" delimiter="/">
+            <name-part name="family" font-variant="small-caps"/>
+          </name>
+          <label prefix=" (" suffix=")"/>
```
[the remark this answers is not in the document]
schtailmuel: I did not use substitute because it did not work as expected in the case where no authors and no editors are specified.
For that, I would need a nested `substitute`, right? It does not work if I define it as follows:

```
<names variable="author">
  <name name-as-sort-order="all" form="short">
  </name>
  <et-al term="and others"/>
  <substitute>
    <names variable="editor">
      <name name-as-sort-order="all" delimiter="/">
      </name>
      <et-al term="and others"/>
      <substitute>
        <text term="anonymous"/>
      </substitute>
    </names>
  </substitute>
</names>
```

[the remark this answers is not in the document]
bwiernik: No, you can have multiple entries in substitute. The first matching one is used. 

You can just use
```
<substitute>
  <names variable="editor"/>
  <text term="anonymous"/>
</substitute>
```

The options on `<names>` are inherited from the parent `<names>` so you don't have to specify them again.

revista-ladinia.csl
```diff
@@ -0,0 +1,163 @@
...
+    <choose>
+      <if variable="publisher-place">
+        <text variable="publisher-place" suffix=" "/>
+      </if>
+      <else>
+        <text term="no-place" suffix=" "/>
+      </else>
+    </choose>
+  </macro>
+  <citation et-al-min="2" et-al-use-first="1" initialize-with="." disambiguate-add-givenname="true" disambiguate-add-year-suffix="true" collapse="year">
```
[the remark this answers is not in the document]
schtailmuel: indeed, thanks!

if there are more than three entries, I only provide the first one without `et al.` (that's why I use the and others term - which is empty - for et-al)

revista-ladinia.csl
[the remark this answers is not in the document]
schtailmuel: Screenshot from 2022-04-25 13-51-50

Here it should display `(N.N. 1980)` and `N.N.: Book without author, s.l. 1980.` but it doesn't...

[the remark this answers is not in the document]
schtailmuel: I guess this issue only occurs inside the Zotero Style Editor, I just saw that the generated sample translations below are as expected.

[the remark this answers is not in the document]
adam3smith: No, this is a bit of a complex issue related to groups. You'll need to put the anonymous term in a macro and call that in substitute for the style to work properly in Zotero, Mendeley, and other tools using the same citation processor (the one we use for previews here is slightly different).

revista-ladinia.csl
```diff
@@ -0,0 +1,154 @@
...
+      <name name-as-sort-order="all" delimiter="/" form="short">
+        <name-part name="family" font-variant="small-caps"/>
+      </name>
+      <et-al term="and others"/>
+      <substitute>
+        <names variable="editor"/>
+        <text term="anonymous"/>
+      </substitute>
+    </names>
+  </macro>
```
adam3smith, line 74:
```suggestion
        <text macro="anonymous"/>
      </substitute>
    </names>
  </macro>
  <macro name="anonymous">
    <text term="anonymous"/>
  </macro>
```

revista-ladinia.csl
```diff
@@ -0,0 +1,154 @@
...
+  <macro name="author-bibliography">
+    <names variable="author">
+      <name name-as-sort-order="all" delimiter="/">
+        <name-part name="family" font-variant="small-caps"/>
+      </name>
+      <et-al/>
+      <substitute>
+        <names variable="editor translator">
```
adam3smith, line 82: you can handle this simpler by adding the label to author above (where it doesn't appear bc author doesn't have a label) and then just use `<names variable="editor translator"/>` in substitute

revista-ladinia.csl
```diff
@@ -0,0 +1,154 @@
...
+    <layout suffix=".">
+      <text macro="author-bibliography" suffix=": "/>
+      <group delimiter=", ">
+        <text variable="title" font-style="italic"/>
+        <group delimiter=": ">
+          <text term="in"/>
+          <text variable="container-title"/>
+        </group>
+        <group delimiter=" ">
+          <text term="volume"/>
```
adam3smith, line 142: I think you'll want form="short" here, right?

adam3smith (changes requested):
Thanks for going through the issues so thoroughly. I think we're pretty much there. A couple more comments inline

revista-ladinia.csl
schtailmuel: much more elegant, thanks!

revista-ladinia.csl
schtailmuel: right, thanks!

adam3smith:
Thanks!

PR #6036, opened 2022-04-24
Create cstr-2022-author-date.csl

cheneyfeng3:
via: https://forums.zotero.org/discussion/96609/style-request-cstr-2022-author-date-csl#latest

### DIFF
--- a/cstr-2022-author-date.csl
+++ b/cstr-2022-author-date.csl
@@ -1,0 +1,680 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" class="in-text" names-delimiter=". " name-as-sort-order="all" sort-separator=" " demote-non-dropping-particle="never" initialize-with=" " initialize-with-hyphen="false" page-range-format="chicago">
+  <!-- 
+    首字母小写 
+    initialize-with 
+   -->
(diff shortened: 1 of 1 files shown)


bwiernik:
Thanks! A few comments to start before we review the style in detail. 

Please paste your style to http://validator.citationstyles.org/ and fix the validation issues it lists.

Please also remove the unused macros from the style. 

Please spell out "Common Science and Technology Resource" in full in the filename, style ID, and style title. Please remove "2022" from those. 

Please remove the comments for things like explaining the different types of styles. 

Please link to the actual documentation to your style, rather than to the generic Chicago manual documentation

bwiernik:
Looking at this, it appears the major thing you are trying to accomplish is adding the “CSTR” variable to a generic citation style. 

Two comments on that:

1. This doesn’t really appear to be a case where a single specific style is needed. The CSTR might be cited in a variety of styles. 
2. CSTR is not a valid CSL variable. For now, the appropriate way to cite CSTR in CSL styles is to enter the item’s URL as `https://cstr.cn/<CSTR>`. This is how, for example, ArXiV IDs are cited. CSL may get formal support for a wider range of ID variables in the future. But for now, `URL` is the best approach. 

I’m going to close this for now. For now, your best course of action is to direct your users to save their URLs for their items appropriately.

PR #6038, opened 2022-04-25
Create bibliothek-forschung-und-praxis.csl

jmiba:
New style according to the style guide of the OA journal Bibliothek - Forschung und Praxis.


jmiba:
> How do you want to call the journal? I'm mostly seeing "Bibliothek: Forschung und Praxis" but also "Bibliothek - Forschung und Praxis" any preferences?

Actually, it is "BIBLIOTHEK – Forschung und Praxis" but we are also content with "Bibliothek – Forschung und Praxis".

adam3smith:
Dankeschön!

PR #6040, opened 2022-04-27
Create health-physics.csl

POBrien333:
via https://forums.zotero.org/discussion/96705/citation-style-for-health-physics-journal#latest


adam3smith:
Sorry for the late review. The chapter example doesn't look quite right -- that should end with

> New York: Oxford University Press; 2001:184–213.

POBrien333:
Thanks for checking. Fixed now. :)

adam3smith:
Thanks!

PR #6045, opened 2022-04-30
Create elsevier-with-titles-italic-no-doi.csl

haziq1301:
Updated Elsevier format but with Italic article title.


bwiernik:
Is this for a specific journal? Elsevier generally requires DOIs for all of their journals

PR #6046, opened 2022-05-01
Bibtex: surround citations with \cite{...}

hu-a:


adam3smith:
@bwiernik -- thoughts on this? I'm generally quite skeptical about this style, but to the extent that it does exist, adding `\cite` might make sense to allow for document conversion? otoh, it does make it impossible to use any other command, which is, I assume, why this wasn't added to the original style.

(Also, should we add the new citation-key variable here?

adam3smith:
Yeah, thinking more about this, hard-coding the `\cite` command is probably a bad idea overall, also keeping backwards compatibility in mind, so I'm going to close this, sorry about that.

PR #6050, opened 2022-05-04

tehH1035:
Removal of space in between year, month, and date for website access by adding range-delimiter="" (prior to this update, the format would be shown as year- month- day).

Looking at it, it doesn't seem like there was a change, so can someone help investigate this?

### DIFF
--- a/american-chemical-society.csl
+++ b/american-chemical-society.csl
@@ -248,6 +248,7 @@
               <text macro="pages"/>
             </group>
           </group>
+        </else-if>
         <else-if type="webpage">
           <group delimiter=". ">
             <text variable="title" font-style="italic"/>
--- a/american-chemical-society.csl
+++ b/american-chemical-society.csl
@@ -251,12 +251,12 @@
         </else-if>
         <else-if type="webpage">
           <group delimiter=" ">
-            <text variable="title"/>
+            <text variable="title" font-style="italic" suffix="."/>
             <text variable="URL"/>
-            <date variable="accessed" prefix="(accessed " suffix=")" delimiter=" ">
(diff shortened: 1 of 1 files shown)


american-chemical-society.csl
```diff
@@ -255,8 +255,8 @@
             <text variable="URL"/>
             <date variable="accessed" prefix="(accessed " suffix=")" delimiter=" ">
```
POBrien333, line 256: The delimiter attribute here sets those spaces.
What do the guidelines say these dates should look like? (I don't have access to the guidelines)

tehH1035: Thanks for fixing this!

According to the ACS manual (https://pubs.acs.org/doi/full/10.1021/acsguide.40305), the recommended format for general websites is:

> Author or Organization (if any). _Title of Site_. URL (accessed YYYY-MM-DD).

Actually, on a related note, based on that new format, the title of site should be updated to include the suffix as "." and the title of the site should be in italics.

POBrien333:
`        <else-if type="webpage">
          <group delimiter=". ">
            <group delimiter=" ">
              <text variable="URL"/>
              <date variable="accessed" prefix="(accessed " suffix=")">
                <date-part name="year"/>
                <date-part name="month" prefix="-" form="numeric-leading-zeros"/>
                <date-part name="day" prefix="-" form="numeric-leading-zeros"/>
              </date>
            </group>
          </group>
        </else-if>`

bwiernik:
@POBrien333  When you want to show multi-line code blocks, put 3 ticks \`\`\` on the line before the code and 3 on the line after

POBrien333:
excellent. cheers for that.

adam3smith:
Thanks!

PR #6051, opened 2022-05-05
Create journal-of-midwifery-science.csl

POBrien333:
via https://forums.zotero.org/discussion/96908/vancouver-style-with-brackets-and-references-in-alphabetical-order#latest


adam3smith:
Great, thanks!

adam3smith:
(I realize there was still an open question on the forum, but I agree that it looks like the style does this correctly)

POBrien333:
Great, thanks for confirming that's coded correctly. :)

PR #6053, opened 2022-05-09
Create journal-of-occupational-and-environmental-medicine.csl

POBrien333:
via https://forums.zotero.org/discussion/96380/style-request-journal-of-occupational-and-environmental-medicine#latest

Based on the actual publication. Guidelines don't seem to match .pdfs.


adam3smith:
Thanks!

PR #6062, opened 2022-05-16

copernicus-publications:
Hi,

it seems container-title-short is not used very often by default (e.g. in Mendeley) and this creates problems for our users.
So I adapted:
If "container-title-short" exists, use it, else, go with "container-title form=short"

Thanks!

### DIFF
--- a/copernicus-publications.csl
+++ b/copernicus-publications.csl
@@ -267,6 +267,7 @@
                 <text variable="page"/>
                 <text macro="doi"/>
               </group>
+            </else-if>
             <else>
               <group delimiter=", " suffix="," prefix=" ">
                 <text macro="title"/>
--- a/copernicus-publications.csl
+++ b/copernicus-publications.csl
@@ -255,7 +255,7 @@
             <else-if type="article-journal" match="any">
               <group delimiter=", " prefix=" ">
                 <text variable="title"/>
-                    <text variable="container-title" form="short"/>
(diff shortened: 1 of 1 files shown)


bwiernik:
Thanks. It would be better to just always use `variable="container-title" form="short"`. `container-title-short` is really only used to test if a short form time is specifically present in the data

copernicus-publications:
Hi,
I only used container-title-short in that one instance (for journal-article), anywhere else it is container-title.
Or do you mean we should drop the if clause checking for container-title-short altogether? I thought that would be ok, if people explicitly defined their own journal shorts?

Please advise,  thanks!

copernicus-publications.csl
```diff
@@ -255,7 +255,14 @@
             <else-if type="article-journal" match="any">
               <group delimiter=", " prefix=" ">
                 <text variable="title"/>
-                <text variable="container-title-short"/>
+                <choose>
```
adam3smith, line 258: You can just go with `<text variable="container-title" form="short"/>` which is equivalent to the entire conditional

bwiernik:
> Or do you mean we should drop the if clause checking for container-title-short altogether? I thought that would be ok, if people explicitly defined their own journal shorts?

Yes, they are functionally the same. Just use `form="short"`

adam3smith:
Thanks -- I've replaced `container-title-short` throughout as discussed above

PR #6063, opened 2022-05-17
Create australasian-journal-of-philosophy.csl

antonyeagle:
Added a new author-date style for the *Australasian Journal of Philosophy*.


adam3smith:
Thanks & sorry for the delay -- this looks great overall; I have two quick questions/requests

1. Did you start out with an existing style? If so, could you add that as a template?
2. The custom terms duplicate a lot of what's already in the en-GB locale: could you remove anything covered there already (e.g., everything related to quotes) or is there a particular reasons it's in the style?

antonyeagle:
Thanks for the feedback, have made those changes as detailed in my comment.

adam3smith:
Thanks!

PR #6069, opened 2022-05-24
Update american-institute-of-physics.csl

JorgeG94:
I am about to submit a paper to [Molecular Physics ](https://www.tandfonline.com/journals/tmph20) and when looking at the reference guidelines I noted that it uses American Institute of Physics which Zotero has. However, based on the [document](
http://www.tandf.co.uk/journals/authors/style/reference/tf_O.pdf) provided by them. The link here downloads a PDF, you can navigate to the site [here](https://www.tandfonline.com/action/authorSubmission?show=instructions&journalCode=tmph20#references) and do it yourself. 

The citation style needs citations to look like this -> [1] and [1-14] while the one in the Zotero style repository it uses superscripts. This PR aims to fix this and update it to what Mol. Phys. requires.

### DIFF
--- a/american-institute-of-physics.csl
+++ b/american-institute-of-physics.csl
@@ -83,9 +83,9 @@
   </macro>
   <citation collapse="citation-number">
     <sort>
-      <key variable="citation-number">
+      <key variable="citation-number"/>
     </sort>
-    <layout delimiter="," vertical-align="sup"/>
+    <layout delimiter="," vertical-align="sup">
       <text variable="citation-number"/>
     </layout>
   </citation>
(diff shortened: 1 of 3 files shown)


adam3smith:
Thanks -- while this is right for the T&F AIP style, the [actual AIP publication manual,](https://publishing.aip.org/wp-content/uploads/2021/03/AIP_Style_4thed.pdf) which the AIP style follows, does prescribe superscript numbers, so the style is OK. We'd take a new style for the T&F version with metadata along these lines

````
<title>Taylor &amp; Francis - American Institute of Physics</title>
<id>http://www.zotero.org/styles/taylor-and-francis-aip</id>
```

JorgeG94:
Oh this makes entire sense, so I should basically redraft this as a new style? right?

adam3smith:
> Oh this makes entire sense, so I should basically redraft this as a new style? right?

Correct, yes -- you can do that in the same PR if you know how to or start a new one if that's easier. Thanks!

PR #6070, opened 2022-05-24
Update glossa.csl

lisalevinson:
changed all authors to be Lastname, Firstname and separated by ampersand, according to updated stylesheet (as requested by Glossa journal editor)


adam3smith:
Thanks -- that ampersand as delimiter is odd, but it's clearly what they want.

PR #6071, opened 2022-05-25
removes abstract from issued macro

denismaier:
I don't see why the the `issued` macro should render the abstract. @manuelmeister Is there any reason for that? I couldn't find anything in the Zitierrichtlinien.


manuelmeister:
Hey @denismaier 
I'm not sure. I've sadly never studied social anthropology ;) I've just created the citation style for my girlfriend.
I asked her and so far she always deleted the abstract. So probably a reasonable suggestion :)
Thanks

denismaier:
Ok. Thanks!

PR #6075, opened 2022-05-27
add title-short to all SBL variants

POBrien333:
Noticed this a few times already as I always search for SBL and can only find 1 style on the repository.

### DIFF
--- a/society-of-biblical-literature-1st-edition-fullnote-bibliography.csl
+++ b/society-of-biblical-literature-1st-edition-fullnote-bibliography.csl
@@ -25,7 +25,7 @@
     <category citation-format="note"/>
     <category field="theology"/>
     <summary>Society of Biblical Literature, 1st edition, format with full notes and bibliography</summary>
-    <updated>2012-10-26T01:15:26+00:00</updated>
+    <updated>2022-05-27T01:15:26+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en-US">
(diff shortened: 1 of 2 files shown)


society-of-biblical-literature-1st-edition-fullnote-bibliography.csl
```diff
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US" page-range-format="chicago">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="chicago" default-locale="en-US">
   <info>
     <title>Society of Biblical Literature 1st edition (full note)</title>
+    <title-short>SBL</title-short>
```
denismaier, line 5: Wouldn't it make sense to add the edition? Maybe change it to "SBL (1st edition)"

society-of-biblical-literature-fullnote-bibliography.csl
```diff
@@ -2,6 +2,7 @@
 <style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="chicago" default-locale="en-US">
   <info>
     <title>Society of Biblical Literature 2nd edition (full note)</title>
+    <title-short>SBL</title-short>
```
denismaier, line 5: Wouldn't it make sense to add the edition? Maybe change it to "SBL (2nd edition)"

denismaier (changes requested):

society-of-biblical-literature-1st-edition-fullnote-bibliography.csl
adam3smith: Agree - while the main function here is findability (which this PR provides) it seems weird to have the exact same short title for multiple styles. @denismaier 's suggestion follows what we do for APA

POBrien333:
Thanks for the comments. I've addressed this in the PR.

denismaier (approved):

PR #6076, opened 2022-05-27
Update avian-research.csl

POBrien333:
goes with https://github.com/citation-style-language/styles/edit/master/dependent/avian-research.csl


adam3smith:
Thanks!

PR #6079, opened 2022-05-31
Add DOI and fix printing author names in Population and Économie et statistique

nalimilan:
This reflects the new style of these journals.

Should I also change the updated date of the dependent style Population (English)?


adam3smith:
Thanks! No need to change the date on the dependent style. It'll still pick up the changes.

PR #6084, opened 2022-06-02
fix et al. settings for journal-of-neuroimaging.csl

### LINKED ISSUE #6083: journal-of-neuroimaging issues
Hi,

This style was created recently: https://github.com/citation-style-language/styles/pull/6048

I wanted to point out that it does not have et al as follows: "List all authors if six or fewer (for papers with more than six authors, list only the first three followed by et al.)."
https://onlinelibrary.wiley.com/page/journal/15526569/homepage/forauthors.html#:~:text=JON%20will%20consider%20both%20invited,before%20preparing%20such%20a%20submission.

POBrien333:
closes #6083

Sample paper: https://onlinelibrary.wiley.com/doi/epdf/10.1111/jon.12983

### DIFF
--- a/journal-of-neuroimaging.csl
+++ b/journal-of-neuroimaging.csl
@@ -14,7 +14,7 @@
     <category field="medicine"/>
     <issn>1051-2284</issn>
     <eissn>1552-6569</eissn>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks!

PR #6085, opened 2022-06-06
Update and rename dependent/vox-sanguinis.csl to vox-sanguinis.csl

POBrien333:
via https://forums.zotero.org/discussion/97607/style-error-vox-sanguinis#latest


POBrien333:
@bwiernik 
Any chance we could get this released so I can write to the publishers? I'd like to have them remove their statement regarding Zotero before it can do more damage to its reputation. :)

POBrien333:
Thank you! I've just emailed the editors. Let's see if we get anything back. :)

PR #6097, opened 2022-06-21
Update technische-universitat-dresden-linguistik.csl

fussballlinguist:
I rewrote some variables (title, container-title, and doi) as macros so that 
1) collection titles and numbers are part of the book title and not separated by periods
2) the DOIs are clickable URLs as in APA 7th ed.


fussballlinguist:
HI, I don't understand the error message. I checked the style in the Style Validator and it passed.

adam3smith:
Did you remove the locale on purpose? If so, please also remove the (Deutsch) in title/short-title.
Everything else looks good.

fussballlinguist:
Not on purpose, that was my helpless try to get through the check which always complained about the locale. I restored it.

adam3smith:
Thanks!

PR #6098, opened 2022-06-21
Update unified-style-sheet-for-linguistics.csl

fussballlinguist:
I added a IF NOT statement to the access macro so that the URL with access date is only printed if there is no DOI (which does not need an access date). This change was agreed with the author Mark Dingemanse beforehand.


fussballlinguist:
I do not understand why the check fails since I did not make any changes to the locales part…

adam3smith:
Ignore that (tests check existing styles and something went wrong with a previous merge)

fussballlinguist:
Okay, what can I do now? The check still fails…

adam3smith:
Just ignore the failing check -- we can still see that your style passes & will review and merge. Same for your other PR. Sorry for the confusion!

adam3smith:
Thanks!

PR #6100, opened 2022-06-22
Create journal-fur-medienlinguistik

fussballlinguist:


journal-fur-medienlinguistik.csl
```diff
@@ -0,0 +1,204 @@
...
+        <text macro="container-title"/>
+      </group>
+    </group>
+  </macro>
+  <macro name="publication-place-publisher">
+    <group delimiter=" ">
+      <text variable="publisher-place" suffix=":"/>
+      <text variable="publisher" quotes="false" font-variant="normal" font-weight="normal" vertical-align="baseline"/>
+      <text macro="collection-title-number"/>
+    </group>
```
adam3smith, line 64:
```suggestion
    <group delimiter=" ">
      <group delimiter=": ">
        <text variable="publisher-place" suffix=":"/>
        <text variable="publisher"/>
      </group>
      <text macro="collection-title-number"/>
    </group>
```

Always more robust to use punctuation in delimiters (here, in case there's no publisher for some reason)

adam3smith (changes requested):

adam3smith:
Thanks!

PR #6102, opened 2022-06-22
Create unified-style-sheet-for-linguistics-cv.csl

fussballlinguist:


adam3smith:
Thanks -- I'm not super inclined to take this for the repository: there's virtually no limit of styles we can create once we start adding CV and other variants to all of them. We added one option for APA so it's available via Zotero & other sites drawing style directly from the API.
Obviously you're more than welcome to share this otherwise and Zotero will install styles directly from github (e.g. both bwiernik and I have a repository of custom styles that don't make sense for the repository)

fussballlinguist:
Thanks for your reply. I actually submitted the style in order to use it via the API. The German Applied Linguistics Association plans to display the publication lists of the sections dynamically via the API and since there is no "sort by date" option, my idea was to submit a derivate of "Unified Style Sheet for Linguistics".
But I understand your concerns. Maybe we will just use APA-CV then (even if we have some concerns with that one, first of all the low visibility of women because of the initials instead of full first names).

adam3smith:
You can actually use custom hosted styles with the Zotero API by specifying the URL of the style -- you just can't use them for Zotero's own CV feature. So yes, I'd keep this off the repository and run the site either using a self-hosted version of your style or using the APA CV style

PR #6103, opened 2022-06-22
Update advanced-materials.csl

POBrien333:
fix some minor issues
found while making dependent style as per https://forums.zotero.org/discussion/97893/style-request-advanced-theory-and-simulations#latest


adam3smith:
Thanks!

PR #6104, opened 2022-06-22

ibadger:
Hi. I work at the University for the Creative Arts. We are in the process of updating our Harvard referencing guide. This has meant I need to update a couple of aspects of our CSL, if there is a better way of amending the code do let me know. The guide now states that we only use et al in the in-text citations whereas the CSL currently applies et al in the bibliography as well, so I have changed the number of authors, directors, editors and translators displayed in the bibliography so that there is no limit and they are all listed. We have also amended the way that personal communication displays a reference in the bibliography, we previously used this for citing emails in our CSL but have amended the guide so that personal communication can also reference interviews and other forms of personal communication which led to changes to to CSL. Ian


adam3smith:
Sorry for the delay here -- this all looks great.

PR #6115, opened 2022-06-28
Update harvard-cite-them-right-no-et-al.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/411079/#Comment_411079

addendum to #[6113](https://github.com/citation-style-language/styles/pull/6113)


adam3smith:
Oh, it most certainly is odd, but it's a citation style foisted upon poor UK undergrads, so we might as well give them what they want.

adam3smith:
Thanks!

PR #6117, opened 2022-07-01
Update nature.csl

POBrien333:
via https://forums.zotero.org/discussion/98158/referencing-preprint-in-nature-format

### DIFF
--- a/nature-no-et-al.csl
+++ b/nature-no-et-al.csl
@@ -12,7 +12,7 @@
     <category citation-format="numeric"/>
     <category field="science"/>
     <category field="generic-base"/>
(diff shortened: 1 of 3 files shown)


adam3smith:
Thanks! Since this is one of the first ones where we explicitly code preprint, I wanted to get this just right so we can use the code as a template -- see updated version.

POBrien333:
Great. I'll add this adjustment for the other nature style(s) when I get a chance.

POBrien333:
Pleaser check if I did everything 100% correct. Thanks :)

adam3smith:
Yes, looks great, thanks! (FWIW, this will look weird if preprints have neither URL nor DOI, but I think we can ignore that possibility)

PR #6129, opened 2022-07-06
Add csl for SDMI

SterArcher:


SterArcher:
Could I please get help on this error, verifier worked fine.

SterArcher:
Could I please get help on this error, verifier worked fine.

adam3smith:
@SterArcher -- is this going to be a relevant style for more than this year's congress? If this will remain in use for subsequent years, we'd be happy to take it, but not sure that having a style for a single conference up on the repository makes sense.

adam3smith:
@SterArcher see my question above

SterArcher:
Sry for late reply @adam3smith , this congress happens every year. So the style will be relevant in the future also.

PR #6135, opened 2022-07-10
suffix '.' after first group; changed e-mail

SuperDomek:
Moved the suffix dot from after the whole citation to after the first group. Updated the creator e-mail domain.


bwiernik:
Thanks!

PR #6140, opened 2022-07-11
Include chapter title in Science style

lcnittl:
According to the official [Instructions for preparing an initial manuscript](https://www.science.org/content/page/instructions-preparing-initial-manuscript#science-citation-style) of the Science Magazine, chapter titles should be included before the book title.


bwiernik:
Thanks!

PR #6143, opened 2022-07-11

POBrien333:
via https://forums.zotero.org/discussion/98352/small-update-needed-for-style-acta-universitatis-agriculturae-sueciae#latest

### DIFF
--- a/acta-universitatis-agriculturae-sueciae.csl
+++ b/acta-universitatis-agriculturae-sueciae.csl
@@ -21,6 +21,10 @@
   <locale xml:lang="en">
     <terms>
       <term name="edition" form="short">ed</term>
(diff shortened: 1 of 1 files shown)


POBrien333:
Not sure this is actually working. hmm...

bwiernik:
What do you mean?

POBrien333:
It's not working as imagined. Can you test it out just to double check?

bwiernik:
What is looking wrong?

POBrien333:
I found the source of the problem.
I've implemented the type "software" which is a new type as of CSL1.0.2, but obviously, Zotero is still on 1.0.1 where "software" is mapped to just "book".

PR #6145, opened 2022-07-12
Add style for the Geographical Analysis journal

HPDell:
Checklist:

- [x] Title Abbreviations
- [x] Title Diacritics
- [x] Style Locale
- [x] File Name
- [x] Style ID
- [x] "self" Link
- [x] License
- [x] "template" Link
- [x] ISSN
- [x] "documentation" Link
- [x] XML Indentation
- [x] Validation

### DIFF
--- a/geographical-analysis.csl
+++ b/geographical-analysis.csl
@@ -143,22 +143,22 @@
           <choose>
             <if type="bill book graphic legal_case legislation motion_picture report song thesis" match="any">
               <group delimiter=", ">
-              <group delimiter=" ">
-                <text macro="edition"/>
-                <text macro="editor"/>
+                <group delimiter=" ">
+                  <text macro="edition"/>
+                  <text macro="editor"/>
+                </group>
+                <text macro="publisher" suffix="."/>
               </group>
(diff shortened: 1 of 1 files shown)


geographical-analysis.csl
```diff
@@ -0,0 +1,178 @@
...
+        </group>
+      </if>
+      <else-if type="paper-conference" match="any">
+        <group delimiter=": " prefix="">
+          <text variable="publisher-place"/>
+          <text variable="publisher"/>
+        </group>
+      </else-if>
+      <else>
+        <group delimiter=", " prefix=", ">
```
bwiernik, line 83: This prefix is going to produce some weird results with the prefix that is included when the publisher macro is called in the bibliography. I suggest removing the prefixes here and then converting them to groups and delimiters in the bibliography. 

Please do this throughout the style when you are using prefixes or suffixes as delimiters between elements. 

Also, do you mean for journal, newspaper, and magazine articles to have their publishers included? If not, please specify what types you mean, rather than using `else`?

adam3smith:
@HPDell please see the comments above & let us know if you need any assistance

PR #6150, opened 2022-07-22
Create MEDLINE PubMed Journal Article Citation Format

yfpeng:
[MEDLINE PubMed Journal Article Citation Format](https://www.nlm.nih.gov/bsd/policy/cit_format.html)

1. Based on AMIA style
2. Add doi, PMID, and PMCID
3. Remove URL
4. Remove et al


adam3smith:
Thanks! How is this different from https://www.zotero.org/styles?q=id%3Anational-library-of-medicine-grant-proposals ?

adam3smith:
I'll close this -- happy to re-open when you get a chance to come back to this, but I do think we have a style that effectively follows these guidelines.

PR #6155, opened 2022-07-25
Remove issue numbers from BJOC style

dcambie:
According to BJOC instruction, issue numbers should only be included for "journal with non-continuous (i.e. issue-based) pagination". And the vast majority of chemistry journals have continuous pagination within volumes so this field should not be included.

This PR is motivated by the fact that I just got a format check report from bjoc instructing me to "please delete all issue numbers (numbers in parentheses) from journal references in the reference list ;)

### DIFF
--- a/beilstein-journal-of-organic-chemistry.csl
+++ b/beilstein-journal-of-organic-chemistry.csl
@@ -154,7 +154,6 @@
                 <if variable="volume">
                   <group delimiter=" ">
                     <text variable="volume" font-style="italic"/>
-                    <text variable="issue" prefix="(" suffix=")"/>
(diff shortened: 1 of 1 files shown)


POBrien333:
I can't make inline comments somehow, but can you change the short title from Beilstein to BJOC?

dcambie:
> I can't make inline comments somehow, but can you change the short title in line 6 from Beilstein to BJOC?

Done

bwiernik:
Thanks!

This is such an annoying rule, but it makes sense to assume what the majority of journals in the field do rather than being inclusive if the journal is going to insist on the formatting

PR #6156, opened 2022-07-25
Update vancouver.csl

POBrien333:
via https://forums.zotero.org/discussion/98637/vancouver-csl-and-formatting-of-reference-when-citing-part-of-a-homepage-web-site#latest

Is this the best way of solving this? I also included this in the journal-access macro. Should I take it out from there as it's not related to journals?

### DIFF
--- a/vancouver.csl
+++ b/vancouver.csl
@@ -79,8 +79,8 @@
     </choose>
   </macro>
   <macro name="publisher">
-    <!--discard publisher info for articles-->
     <choose>
+    <!--discard publisher info for articles-->
       <if type="article-journal article-magazine article-newspaper" match="none">
         <group delimiter=": " suffix=";">
           <choose>
--- a/vancouver.csl
+++ b/vancouver.csl
@@ -22,7 +22,7 @@
     <category citation-format="numeric"/>
     <category field="medicine"/>
     <summary>Vancouver style as outlined by International Committee of Medical Journal Editors Uniform Requirements for Manuscripts Submitted to Biomedical Journals: Sample References</summary>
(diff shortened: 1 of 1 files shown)


adam3smith:
They way I understand the instructions (you can look at them in excruciating detail in [chapter 25 of _Citing Medicine_](https://www.ncbi.nlm.nih.gov/books/NBK7274/?report=reader), for "parts of a webpage" we'd actually want the website title first, and the title in the end. On the other hand, for the "homepage," we'd want the title in the "regular" position.

The way I'd solve this is to text for container-title for webpages: where it exists, we assume we're looking at part of a website and it should come first. Where it doesn't exist, we assume title is the title of the homepage and we use the current format. WIll be a bit messy, but shouldn't be too bad. Does that sound right?

vancouver.csl
```diff
@@ -158,13 +169,28 @@
...
+        </choose>
       </else>
     </choose>
   </macro>
   <macro name="title">
-    <text variable="title"/>
+    <choose>
+      <if type="webpage" match="all">
+        <choose>
+          <if match="none" variable="title">
```
adam3smith, line 185: is this supposed to be `container-title` here? as written, it doesn't make sense. (Sorry for the long delay; I had missed that you worked on this again)

POBrien333: I had a look at it 2 weeks ago, but couldn't get my head around all of this. I've just checked now and made the change, but please double check this PR actually makes sense. :)

adam3smith:
OK, thanks for all the work on this -- I think I sorted this out reasonably cleanly

PR #6159, opened 2022-07-26
Create the-depositional-record.csl

POBrien333:
via https://forums.zotero.org/discussion/98516/style-request-the-depositional-record#latest

OA papers:
https://onlinelibrary.wiley.com/doi/pdf/10.1002/dep2.173
https://onlinelibrary.wiley.com/doi/epdf/10.1002/dep2.166

Style adapted based on papers.


adam3smith:
Thanks!

PR #6164, opened 2022-08-02
Change "event" to "event-title"

swifterslb:
Recently, the "event“ variable was replaced by "event-title". Item type conferencePaper and presentation were affected.


bwiernik:
Thanks!

PR #6166, opened 2022-08-02
journal-of-hearing-science-2022.csl

JHSifps:
Minor style changes

### DIFF
--- a/journal-of-hearing-science.csl
+++ b/journal-of-hearing-science.csl
@@ -89,22 +89,22 @@
   </macro>
   <macro name="container-title">
     <choose>
-      <if type="article-journal article-magazine chapter paper-conference article-newspaper" match="any">
-        <group suffix=" " delimiter=" ">
-          <text variable="container-title" form="short"/>
-          <choose>
-            <if variable="URL">
-              <text term="internet" prefix="[" suffix="]" text-case="capitalize-first"/>
-            </if>
-          </choose>
-        </group>
-        <text macro="edition" prefix=" "/>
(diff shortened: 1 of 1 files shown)


bwiernik:
Thanks. Are you sure that's correct? Now there isn't any punctuation between a book title and the publisher? What is the problem you're trying to fix?

JHSifps:
Dear Brenton,

I try to update a JHS Zotero style to our instruction for authors on the JHS website. I followed the Editing CSL Styles Guide. However, I am not IT worker and have no experience with such an issue, as you can see.  I wanted to withdraw full-stops from journals abbreviated name: Curr Biol

There are many other things to change, and perhaps it would be better to create a new style for JHS. What is your opinion?

With regards
Aleksandra

Journal of Hearing Science
Managing Editor

adam3smith:
Thanks -- I reworked this to actually reflect the style in the author guidelines

JHSifps:
Thank you very much! I wonder if I could ask you to do some changes in the future? We are now in the process of changing our website and the guidelines for authors will change too.
With regards
Aleksandra

adam3smith:
> Thank you very much! I wonder if I could ask you to do some changes in the future? We are now in the process of changing our website and the guidelines for authors will change too.

You can request changes via the Zotero forums, though no guarantees how fast that'll be (anywhere from 2 weeks to 2 months is typical). 
If you need something quicker, either paying for it or doing it yourself (ideally with clean code) typically gets this in faster.

PR #6170, opened 2022-08-03
Create sociologia-ruralis.csl

BjoernSchwedn:
For the journal 'Sociologia Ruralis' no official citation style was provided by the publisher. 

The now uploaded style is based on an already existing style ([http://www.zotero.org/styles/journal-of-rural-studies](url) respectively [http://www.zotero.org/styles/elsevier-harvard](url)) and has been adapted to the journals' author guidelines (see [https://onlinelibrary.wiley.com/page/journal/14679523/homepage/forauthors.html](url)). 

**Note:** However, the guidelines on the journals' website differ from its actual bibliography-format in the articles. I have therefore adapted the bibliography-format to a recent article (year 2022) [https://doi.org/10.1111/soru.12382](url) and crossed-checked it with some others.

### DIFF
--- a/sociologia-ruralis.csl
+++ b/sociologia-ruralis.csl
@@ -5,7 +5,7 @@
     <title-short>SocRur</title-short>
     <id>http://www.zotero.org/styles/sociologia-ruralis</id>
     <link href="http://www.zotero.org/styles/sociologia-ruralis" rel="self"/>
-    <link href="http://www.zotero.org/styles/journal-of-rural-studies" rel="template"/>
+    <link href="http://www.zotero.org/styles/elsevier-harvard" rel="template"/>
     <author>
       <name>Bernhard Grüner</name>
       <email>bernhard.gruener@uibk.ac.at</email>
--- a/sociologia-ruralis.csl
+++ b/sociologia-ruralis.csl
@@ -1,0 +1,226 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="in-text" version="1.0" demote-non-dropping-particle="never" default-locale="en-GB" xmlns="http://purl.org/net/xbiblio/csl">
+  <info>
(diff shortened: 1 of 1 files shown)


sociologia-ruralis.csl
```diff
@@ -0,0 +1,226 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" default-locale="en-GB">
+  <info>
+    <title>Sociologia Ruralis</title>
+    <title-short>SocRur</title-short>
+    <id>http://www.zotero.org/styles/sociologia-ruralis</id>
+    <link href="http://www.zotero.org/styles/sociologia-ruralis" rel="self"/>
+    <link href="http://www.zotero.org/styles/elsevier-harvard" rel="template"/>
+    <author>
```
POBrien333, line 9: Add a link to the style's documentation (even if it's outdated) using link with rel=documentation

sociologia-ruralis.csl
```diff
@@ -0,0 +1,226 @@
...
+      <name>Bernhard Grüner</name>
+      <email>bernhard.gruener@uibk.ac.at</email>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="sociology"/>
+    <category field="generic-base"/>
+    <summary>Based on 'Journal of Rural Studies'-Style</summary>
```
POBrien333, line 16: Please add the ISSNs of the style. See: https://www.ncbi.nlm.nih.gov/nlmcatalog

adam3smith:
Thanks, looks great -- seems like a fairly recent change in the journal style that's not reflected in the guidelines. Older (~2019) articles still look more like the guidelines.

PR #6172, opened 2022-08-09

ssha4900:
Reiterating on our APA version to focus on referencing legal items in such a way that it aligns with the practice of this institution.

### DIFF
--- a/ucl-university-college-apa.csl
+++ b/ucl-university-college-apa.csl
@@ -1368,9 +1368,9 @@
              For uncodified regulations, assume future code section is in status. -->
         <group delimiter=". ">
           <group delimiter=", ">
-          <!--Rettelse: linje 1387 og 1388 angiver at der skal være populærtitel OG fuld titel i reference i lit. listen. Bemærk der bruges title-short og ikke form="short"-->
+            <!--Rettelse: linje 1387 og 1388 angiver at der skal være populærtitel OG fuld titel i reference i lit. listen. Bemærk der bruges title-short og ikke form="short"-->
             <text variable="title-short" font-style="italic"/>
-             <text variable="title" form="long" font-style="italic"/>
+            <text variable="title" form="long" font-style="italic"/>
             <group delimiter=" ">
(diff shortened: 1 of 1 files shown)


ucl-university-college-apa.csl
```diff
@@ -214,7 +214,22 @@
     </choose>
   </macro>
   <macro name="date-sort-date">
```
adam3smith, line 216: I don't understand what you're doing here? Essentially you're using the same sort date for everything except a conference paper without publication info -- but what is that getting sorted by?

adam3smith (changes requested):
Sorry for the delay -- the legislation change looks fine, but I don't understand what you're doing with the date-sort macro, could you clarify and/or fix?

adam3smith:
Please see review above

ucl-university-college-apa.csl
ssha4900: Hi Adam

I'm confused as to your comment, but mostly because I don't remember touching that part at all. Our reference style is a (slightly) modified version of APA 7 and when I check that source I see the same on line 503-505.
image

So I would love to clarify if I could, but I don't think I have made that change myself. But I am also a total novice, so please steer me in the right direction if you can!

adam3smith: The issue is the content of the macro, not its existence -- it's completely different (and likely needlessly complicated for your purposes) in APA 7. I'd suggest just reverting this to your previous version, which was simply

```
  <macro name="date-sort-date">
    <date variable="issued" form="numeric"/>
  </macro>
```

If you do need something changed about the sort behavior, could you say what?

ssha4900: I see what you mean now. And no, it must have been an mistake. I saw lines that appeared twice for no reason, so I think there must have been some dodgy copy+paste on my part

adam3smith:
Thanks!

ssha4900:
> 

Thanks for your patience! :D

PR #6176, opened 2022-08-11
remove DOI for printed articles organizational-studies.csl

POBrien333:
via https://forums.zotero.org/discussion/98994/request-fix-for-organization-studies-style#latest


bwiernik:
What a silly deviation from APA style. 

Thanks for the update!

PR #6180, opened 2022-08-15
Delete harvard-cardiff-university-old.csl

YoungZoe:
This style is no longer in use by Cardiff University


bwiernik:
@adam3smith should we delete or make a dependent?

adam3smith:
No reason for a dependent style -- [there's still a Cardiff Harvard style](https://www.zotero.org/styles?q=id%3Acardiff-university-harvard), this was just the legacy version the library had asked us to keep up for a bit longer and apparently that's no longer needed now.

PR #6181, opened 2022-08-16
Create harvard-xi-an-jiaotong-liverpool-univeisity.csl

ous50:


ous50:
But that's how my university spell.... If I remove that apostrophe, the pronunciation is completely different :(

adam3smith:
Thanks & sorry for the delay.
This style doesn't seem to match the linked style guide very closely yet. Just had a cursory look but:
- Et al shouldn't be used for 3 authors in text
- Et al should be in italics
- Et al should be used in the bibliography
- Year in bibliography should be in parentheses
- Quotation marks instead of italics for the title of a book chapter
- (eds.) should be in parentheses

There's likely more -- could you fix this up to match the style guide a bit more closely, please?

adam3smith:
@ous50 -- please see my question above

ous50:
> @ous50 -- please see my question above

Sorry for the late reply. I was so busy these days and did not notice your opinion. I am looking documentations to modify it.

adam3smith:
Thanks!

ous50:
thanks for your correction :)

PR #6186, opened 2022-08-19
Update rassegna-degli-archivi-di-stato.csl

Pepppino:
The link to the documentation was updated, capitalizing of first names was added, a missing comma in the printout of conference papers was added.


adam3smith:
Thanks!

PR #6188, opened 2022-08-22
Update cell-numeric-superscript.csl

POBrien333:
via https://forums.zotero.org/discussion/99081/add-doi#latest

It looks like Cell generally has added DOIs now as it appears in the guidelines of OP's style, but also in the guidelines we based our style on:
https://www.cell.com/current-biology/authors
https://www.cell.com/chem/authors


adam3smith:
I went with DOI throughout -- I'd assume they want it for datasets & preprints, too, e.g.

PR #6189, opened 2022-08-22

POBrien333:
via https://forums.zotero.org/discussion/99017/space-before-pp-number-in-chapters-of-edited-books-harvard-anglia-ruskin-output-style#latest


adam3smith:
Thanks!

PR #6190, opened 2022-08-22

POBrien333:
via https://forums.zotero.org/discussion/99136/cardiff-harvard-zotero-citations-no-longer-work#latest


adam3smith:
Thanks!

PR #6197, opened 2022-08-29
Delete moore-theological-college.csl as per university

POBrien333:
https://forums.zotero.org/discussion/99368/style-request-remove-moore-theological-college#latest


adam3smith:
Thanks

PR #6198, opened 2022-08-29
Update mcgill-fr.csl

POBrien333:
via https://forums.zotero.org/discussion/99373/citation-edit-how-to-add-smallcaps-to-name-only-not-surname#latest


adam3smith:
Thanks!

PR #6199, opened 2022-08-30
Update politix.csl

POBrien333:
via https://forums.zotero.org/discussion/99399/zotero-pages-de-certaines-citations-ne-saffichent-pas-style-politix#latest


adam3smith:
Thanks! That style was a mess....

PR #6201, opened 2022-08-31
Create isara-iso-690.csl

odebiez:
New style based on the ISO-690 standard and the work of Raphaël Grolimund and Pierre-Amiel Giraud in 2021/2022, adapted to the requirements of the **Isara** engineering school (Lyon, France) for use with Zotero:
- "Et al." (in italic) for 3 or more authors in the citation
- A comma between the author and the date in the citation
- Mention [Auteur ?] before the title if no author, editor or translator is indicated
- Mention [sans date] if no date is indicated


adam3smith:
Thanks!

PR #6205, opened 2022-09-02
Create mots.csl

Oxil-2:
CSL style for the French journal "Mots - Les langages du politique" (politics, linguistics, communication studies), cf. https://journals.openedition.org/mots/

### DIFF
--- a/mots.csl
+++ b/mots.csl
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" default-locale="fr-FR" >
-
-<!--
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" default-locale="fr-FR">
+  <!--
 
 Feuille de style CSL 
 à destination du logiciel Zotero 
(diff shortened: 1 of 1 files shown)


mots.csl
```diff
@@ -0,0 +1,298 @@
...
+comportant les instructions de formatage des références bibliographiques et de la bibliographie de littérature secondaire définies par la revue Mots [https://journals.openedition.org/mots/76]
+réalisée par Louis Autin (louis.autin@sorbonne-universite.fr / SU)
+
+ -->
+  <!-- INFORMATIONS RELATIVES A LA FEUILLE DE STYLE -->
+  <info>
+    <title>Mots (French)</title>
+    <title-short>Mots</title-short>
+    <id>http://www.zotero.org/styles/mots</id>
+    <link href="http://www.zotero.org/styles/mots" rel="self"/>
```
POBrien333, line 16: please add the style you used as a template as indicated here: https://github.com/citation-style-language/styles/blob/master/STYLE_REQUIREMENTS.md#csl-style-repository-requirements

mots.csl
```diff
@@ -0,0 +1,299 @@
...
+    </names>
+  </macro>
+  <macro name="auteurs">
+    <names variable="author">
+      <name name-as-sort-order="all" and="text" delimiter=" " delimiter-precedes-last="never" sort-separator=" ">
+        <name-part name="family" font-variant="small-caps"/>
+      </name>
+      <label form="short" prefix=" (" suffix=")"/>
+      <substitute>
+        <names variable="editor">
```
adam3smith, line 93: why are you using these elaborate substitutes instead of just `<names variable="editor"/>` (the other elements are inherited from author)

mots.csl
```diff
@@ -0,0 +1,299 @@
...
+            <text macro="indication-localisation"/>
+          </group>
+        </else>
+      </choose>
+    </layout>
+  </citation>
+  <!-- INSTRUCTIONS DE FORMATAGE DE LA BIBLIOGRAPHIE -->
+  <bibliography entry-spacing="0">
+    <sort>
+      <key macro="contributeurs-tri"/>
```
adam3smith, line 279: I don't understand the difference between the auteurs macro and this one -- they seem redundant. Why not just sort by `auteurs`?

mots.csl
```diff
@@ -0,0 +1,299 @@
...
+      </if>
+    </choose>
+  </macro>
+  <macro name="auteurs-ouvrage">
+    <choose>
+      <if type="chapter paper-conference" match="any">
+        <text term="in" suffix=" "/>
+        <group delimiter=", ">
+          <choose>
+            <if variable="author">
```
adam3smith, line 158: No need for these tests -- they are implicitly in `substitute` above

mots.csl
```diff
@@ -0,0 +1,299 @@
...
+    <choose>
+      <if type="article-journal">
+        <text variable="volume" prefix=", n° " suffix=","/>
+      </if>
+    </choose>
+  </macro>
+  <macro name="lieu-edition">
+    <choose>
+      <if type="book chapter thesis report paper-conference" match="any">
+        <group delimiter=", ">
```
adam3smith, line 188: don't use groups with only one element inside

mots.csl
```diff
@@ -0,0 +1,299 @@
...
+          <text variable="publisher-place" form="short"/>
+        </group>
+      </if>
+    </choose>
+  </macro>
+  <macro name="editeur">
+    <choose>
+      <if type="book chapter thesis report paper-conference" match="any">
+        <group delimiter=", ">
```
adam3smith, line 197: same here with groups

mots.csl
```diff
@@ -0,0 +1,299 @@
...
+        <text variable="page" prefix="p. "/>
+      </if>
+    </choose>
+  </macro>
+  <macro name="type-acces">
+    <group delimiter=" ">
+      <text variable="URL" prefix=" "/>
+      <choose>
+        <if variable="URL">
+          <group prefix="(consulté le " suffix=")">
```
adam3smith, line 240: it's much preferred to avoid hardcoding strings and using (and if necessary re-defining) terms instead.

mots.csl
```diff
@@ -0,0 +1,299 @@
...
+    </group>
+  </macro>
+  <!-- INSTRUCTIONS DE FORMATAGE DES CITATIONS -->
+  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-names="true" disambiguate-add-year-suffix="true">
+    <layout delimiter=" ; ">
+      <choose>
+        <if position="ibid-with-locator">
+          <group delimiter=", ">
+            <text term="ibid" font-style="italic" suffix="."/>
+            <text variable="locator" prefix="p. "/>
```
adam3smith, line 258: You definitely want to use a label for the locator, since that can also be other things than page (like chapter, column, etc.)

mots.csl
```diff
@@ -0,0 +1,299 @@
...
+        <if type="book report thesis" match="any">
+          <group>
+            <text term="volume" form="short" suffix=" "/>
+            <number variable="volume" form="numeric"/>
+          </group>
+        </if>
+      </choose>
+      <group delimiter=" ">
+        <choose>
+          <if locator="page" match="none">
```
adam3smith, line 68: why do you need this test if you're then printing the p. for pages anyway?

mots.csl
```diff
@@ -0,0 +1,299 @@
...
+        <text variable="title" quotes="true"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="auteurs-secondaires">
+    <choose>
+      <if type="chapter paper-conference" match="none">
+        <group delimiter=". ">
+          <choose>
+            <if variable="author">
```
adam3smith, line 133: again, you don't need these tests because of substitute

adam3smith (changes requested):
Thanks & sorry for the slow review -- a number of comments inline

Oxil-2:
> Thanks & sorry for the slow review -- a number of comments inline

Thank you so much for the review. I've tried to correct the errors and just pulled an updated file.

adam3smith:
Thanks, this looks great now -- yes, please do adjust the other two along similar lines.

PR #6209, opened 2022-09-05
Create proceedings-of-the-estonian-academy-of-sciences-author-date.csl

POBrien333:
via csl.org form
The journal has both an author-date and numeric style.
Styles have been reviewed by editors and signed off.

### DIFF
--- a/proceedings-of-the-estonian-academy-of-sciences-author-date.csl
+++ b/proceedings-of-the-estonian-academy-of-sciences-author-date.csl
@@ -1,0 +1,222 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>Proceedings of the Estonian Academy of Sciences (author-date)</title>
+    <title-short>PEAS</title-short>
+    <id>http://www.zotero.org/styles/proceedings-of-the-estonian-academy-of-sciences-author-date</id>
+    <link href="http://www.zotero.org/styles/proceedings-of-the-estonian-academy-of-sciences-author-date" rel="self"/>
+    <link href="http://www.zotero.org/styles/biostatistics" rel="template"/>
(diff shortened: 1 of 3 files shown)


adam3smith:
@retorquere -- I thought my last commit here would fix the error (but clearly it hasn't). Do you know from where (which branch/repo) exactly sheldon pulls the filters.yaml file (which contains the list of ISSNs excepted from the duplicate check) ?

retorquere:
I don't think it does a pull as such, it just expects the file to be present, and it is present in master.

retorquere:
On what branch did you make the change? The change needs to present in the PR, so maybe it just needs to be pulled into the PR.

adam3smith:
Right, the changes are in this branch, which is why I expected this to go through: https://github.com/citation-style-language/styles/pull/6209/commits/3ee732916be56d5e02fc9e92817a51f1eb5b74fa

retorquere:
I'll have to dive into this -- I had expected the changes tobe applied suing a call to `sheldon --apply`, but I removed that in march for some reason.

adam3smith:
OK, I'll go ahead and try merging this -- if we keep getting a failure we can always remove the ISSNs temporarily from one of the styles

retorquere:
> @retorquere -- I thought my last commit here would fix the error (but clearly it hasn't). Do you know from where (which branch/repo) exactly sheldon pulls the filters.yaml file (which contains the list of ISSNs excepted from the duplicate check) ?

I finally found it -- as a precaution, only PRs labeled "safe to test" are tested as-is, all other PRs are tested in a split environment:

1. The PR base, which offers the test environment, and checked out to `~`
2. The changed files in the P, checked out in `pull-request`

The styles to test are taken from 2., but all test infra (among which the spec/filters.yaml) are taken from 1. This is done so a PR can only offer inert data files for automated test, not change the environment tested with (which would allow exfil of API keys)

I don't know how to commit to 1. in a way it would be picked up as part of the PR. The easiest way to fix this that I can think of would be to change 1. to always be "latest master". This does mean that the test envornment could be ahead of the state of the PR, but on the other hand, it would make sure it's mergeable to master.

PR #6215, opened 2022-09-08
Create migration-policy-institute.csl

POBrien333:
via csl.org contact form (Lauren)
This has gone through 4 rounds of reviews and is finalised now.

They don't have public guidelines and they mainly write the pieces internally.

@adam3smith would be grateful if we could get this published in the next days. Thank you :)

### DIFF
--- a/migration-policy-institute.csl
+++ b/migration-policy-institute.csl
@@ -324,13 +324,6 @@
       </if>
     </choose>
   </macro>
-  <macro name="archive-note">
-    <group delimiter=", ">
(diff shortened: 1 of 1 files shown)


adam3smith:
hmm -- we typically don't post internal styles for small-ish orgs to the repository, see https://github.com/citation-style-language/styles/blob/master/README.md#criteria-for-inclusion
is that important to them?

adam3smith:
We have some flexibility on this, so let me know if this is an issue.

PR #6217, opened 2022-09-08
Update journal-of-medical-internet-research.csl

andreibosco:
Updated JMIR reference style based on https://support.jmir.org/hc/en-us/articles/115001333067

Changes:
- Removed square brackets around doi field.
- Added space after URL.
- Changed URL term 'cited' to 'accessed'.
- Removed URL term 'internet'.
- Moved 'access-date' to be the last item of the reference.


adam3smith:
(note to self: still need to fix accessed date for articles with doi)

adam3smith:
Thanks!

PR #6225, opened 2022-09-13
ready: Update oil-shale.csl

POBrien333:
completely reworked style with client (EAP via citationstyles.org)

Waiting for review from client.


adam3smith:
Pretty sure this is just an issue with the template missing from the feature branch. Going to merge

PR #6235, opened 2022-09-20
Create JAMCA.csl

asinensis:
This style is for Journal of American Mosquito Control Association. It is a modified version from CSE N-Y style.


JAMCA.csl
```diff
@@ -0,0 +1,261 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <!-- This style was edited with the Visual CSL Editor (https://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>Journal of American Mosquito Control Association (author-date)</title>
+    <title-short>JAMCA N-Y</title-short>
+    <id>http://www.zotero.org/styles/journal-of-american-mosquito-control-association</id>
+    <link rel="self" href="http://www.zotero.org/styles/journal-of-american-mosquito-control-association"/>
```
POBrien333, line 8: add another link with the template style you used and using `rel="template"`

POBrien333:
You'll need to change the filename to `journal-of-american-mosquito-control-association.csl` as it won't validate otherwise ;)

asinensis:
It seems like this style is not accessible from the Zotero Style
Repository.
Do you know if this will happen automatically or should I do something to
make it happen?

Best,
Yoosook

adam3smith:
The style will show up automatically once the pull request here is fully reviewed and merged

POBrien333:
Up top you see a green button that says "Open". That's because the pull request hasn't been reviewed and merged to the repository.
You'll need to be patient until the maintainers had s chance to do so. That's probably a few weeks out or so.

journal-of-american-mosquito-control-association.csl
```diff
@@ -0,0 +1,261 @@
...
+          </else-if>
+          <else>
+            <text macro="editor" suffix="." prefix=" "/>
+            <group prefix=" ">
+              <text variable="container-title" form="short" strip-periods="true" font-style="italic" suffix="."/>
+              <text variable="volume" prefix=" "/>
+              <text variable="issue" prefix="(" suffix=")"/>
+              <text variable="page" prefix=":"/>
+            </group>
+            <text variable="DOI" prefix=". doi:"/>
```
bwiernik, line 254: Use `https://doi.org/` as the DOI prefix, rather than `doi: `. This is the preferred format, and it is used this way in several articles in the journal I checked.

journal-of-american-mosquito-control-association.csl
```diff
@@ -0,0 +1,261 @@
...
+          </else>
+        </choose>
+      </group>
+      <text prefix=" " macro="access"/>
```
bwiernik, line 258: Only include the `access` macro if the item has no DOI. This is consistent with several articles in the journal I checked.

journal-of-american-mosquito-control-association.csl
```diff
@@ -0,0 +1,261 @@
...
+              <group suffix="." delimiter=". ">
+                <text macro="publisher"/>
+                <text prefix=" " macro="collection"/>
+                <text macro="pages"/>
+              </group>
+            </group>
+          </else-if>
+          <else-if type="review review-book" match="any">
+            <text macro="editor" prefix=" " suffix="."/>
+            <group prefix=" ">
```
bwiernik, line 237: Throughout the style, you frequently use prefixes and suffixes to hold spaces and other delimiters between elements. This is a bad practice, as it can lead to extraneous spaces or punctuation when a variable is missing. For example, if a journal article didn't have a volume, issue, or page number, the style currently would add 2 `.` between the container title and the DOI. 

Please go through your style and consistently use the `delimiter` argument on `<group>` to hold delimiters between elements, rather than prefix/suffix. Only use prefix/suffix for something that should always appear on the element, regardless of what other variables are or aren't there (e.g., the `()` around issue).

bwiernik (changes requested):
A few comments. Notably, please use `delimiter` on `<group>` to separate elements throughout your style, rather than prefix/suffix.

adam3smith:
@asinensis -- please let us know if you have any questions about the review above. Thanks!

asinensis:
Sorry, I had many compounding deadlines and travel so I did not have time
to review this yet. I should be able to get back to this after
thanksgiving. Sorry for the delay.

adam3smith:
Closing for inactivity -- happy to re-open if you get back to us on the questions above, just post here

PR #6236, opened 2022-09-21
Create journal-of-music-archaeology.csl

StefanHagel:
CSL for newly founded Journal of Music Archaeology, published at the Austrian Academy of Sciences: https://jma.vlg.oeaw.ac.at/

### DIFF
--- a/journal-of-music-archaeology.csl
+++ b/journal-of-music-archaeology.csl
@@ -4,9 +4,8 @@
   <info>
     <title>Journal of Music Archaeology</title>
     <title-short>JMA</title-short>
-    <id>f1732a6e-39c0-11ed-a261-0242ac120002</id>
+    <id>http://www.zotero.org/styles/journal-of-music-archaeology</id>
     <link href="http://www.zotero.org/styles/journal-of-music-archaeology" rel="self"/>
-    <link href="http://www.zotero.org/styles/harvard-cardiff-university" rel="template"/>
     <author>
       <name>Stefan Hagel</name>
     </author>


adam3smith:
Thanks & sorry for the delay in reviewing. @StefanHagel generally this looks great, but I'm a bit puzzled by the in-text citation format: that's really not supposed to have parentheses? It's unclear from the author guidelines (though note the year in parenthesis in the narrative citation example `West (1992: 153–55) ` ) but it seems very odd to have in-text citations without any marker like a parenthesis.

StefanHagel:
In the humanities, and perhaps especially in the kind of studies 
where I am at home, we have a tradition of very flexible usage 
of footnotes, which are much more often used for traditional 
argument. 

Therefore, although it may sometimes make sense to have 
short citations in the running text in parentheses, we mostly 
need the possibility to insert citations in the footnotes as in-text 
citations, whether also in parentheses or in narrative style.

As you know, this comes with the downside of not being easily 
convertible into other formats, but there's no way around that. 

(Compare the well-established Journal of Hellenic Studies.)

Thanks for all your work!!!
Stefan

On 16 Nov 2022 at 20:13, Sebastian Karcher wrote:

> 
> Thanks & sorry for the delay in reviewing. @StefanHagel generally this looks great, but 
> I'm a bit puzzled by the in-text citation format: that's really not supposed to have 
> parentheses? It's unclear from the author guidelines (though note the year in 
> parenthesis in the narrative citation example West (1992: 153-55) ) but it seems very 
> odd to have in-text citations without any marker like a parenthesis.
> -
> Reply to this email directly, view it on GitHub, or unsubscribe .


-- 
PD Dr Stefan Hagel
PI Ancient Music 
Österreichisches Archäologisches Institut | Austrian 
Archaeological Institute
Österreichische Akademie der Wissenschaften | Austrian 
Academy of Sciences

Hollandstraße 11-13/6.20, 1020 Wien, Österreich | Vienna, 
Austria
T: +43 1 51581-2341 
***@***.*** | https://www.oeaw.ac.at/oeai/
https://homepage.univie.ac.at/stefan.hagel/

adam3smith:
> Compare the well-established Journal of Hellenic Studies.)

So I looked at JHS & its styleguide and I think our style there is just wrong. The instructions for authors are very clear that citations should be in notes, so that should be a note style and I suspect that's the way to go here, as well, unless you're saying that citations can actually be used outside notes, too, which seems unlikely (and would be a very confusing stylistic choice).

StefanHagel:
Yes, the citations in JHS should be in notes, but notes are very 
different from what STEM-informed practice might think they 
are. There is no way a typical argumentative footnote in JHS 
can be expressed as a citation field wholesale, and conversion 
between these and notes-in-text styles could never work.

Here's a typical footnote, taken from a publication of mine in 
JHS:

    8   For this mechanism, cf. Howard (1893) 5-8; Bodley 
    (1946) (whose scales are unfortunately based on the 
    unfounded theory proposed by Schlesinger (1939); cf. 
    Landels (1981) 300-1); Masaraki (1974); Litvinsky 
    (1999). On a different slider mechanism, cf. also Byrne 
    (2002). In most cases only scattered sections of these 
    elaborate pipes remain [...]  

As I said, this can only be achieved by employing in-text mode 
and using in-text citations within footnotes. With the additional 
benefit of using them in illustration captions as well.

It's not wrong, it's Classics... And it has worked for me for many 
years across various journals.

Best,
Stefan


On 17 Nov 2022 at 6:26, Sebastian Karcher wrote:

> 
> Compare the well-established Journal of Hellenic Studies.)
> So I looked at JHS & its styleguide and I think our style there is just wrong. The 
> instructions for authors are very clear that citations should be in notes, so that should be 
> a note style and I suspect that's the way to go here, as well, unless you're saying that 
> citations can actually be used outside notes, too, which seems unlikely (and would be a 
> very confusing stylistic choice).
> -
> Reply to this email directly, view it on GitHub, or unsubscribe.



PD Dr Stefan Hagel
PI Ancient Music 
Österreichisches Archäologisches Institut | Austrian 
Archaeological Institute
Österreichische Akademie der Wissenschaften | Austrian 
Academy of Sciences

Hollandstraße 11-13/6.20, 1020 Wien, Österreich | Vienna, 
Austria
T: +43 1 51581-2341 
***@***.*** | www.oeaw.ac.at/oeai 
homepage.univie.ac.at/stefan.hagel

adam3smith:
I know what discursive footnotes are -- I've been working on citation styles for over a decade ;). 

You're just slightly misunderstanding how CSL note styles work:

> As I said, this can only be achieved by employing in-text mode and using in-text citations within footnotes. 

That's not the case: if you insert a citation in a `note` style into a footnote/endnote with CSL, it behaves like an in text citation, so the above is absolutely possible without an `in-text` style. (Slightly less complex discursive notes can also easily be generated by using prefix and suffix elements in the citation [as explained e.g. for Zotero here](https://www.zotero.org/support/word_processor_plugin_usage#customizing_cites)). 
The advantage of this is that a) you save yourself some time when just inserting simple citation -- which, looking e.g. at your recent article in _GRMS_, are still the majority of citations and b) you maintain compatibility with other note-based styles commonly used in the humanities. I don't think a scenario where someone wants to move from, say, Chicago Manual (note) to this (or the JHL or the GRMS) style is implausible or even rare, and designating them as `in-text` breaks that.

StefanHagel:
Still, 

1) we do not discourage authors from using simple in-text 
citations, especially when they only add the year to a name 
specified in the text. Why not offering flexibility and a whiff of 
personal style, especially in a field that brings together experts 
from very diverse fields?

2) I find having automatic citations in captions highly practical.

It is true that people might want to move between note and 
JHS/GRMS/JMA style. Your stance here is technical 
compatibility as long as compatibility is maintained by the user, 
which requires a high level of understanding what is going on. I 
do see the advantages. Mine is rather not pretending it is 
possible, i.e. users should realise quickly it does not *really* 
work and therefore not risk losing data because it seems 
generally to be working. Based on my experiences when first 
exploring citation software. 
That's perhaps a matter of 'philosophy', and my idea may be 
mistaken. Even so, do the points above justify not switching to 
note style?

Having GRMS and JMA working similarly and keeping these 
compatible would be important because there's a substantial 
overlap between potential contributors to both (and JHS, btw).

Again, thank you for your incredible contribution to the life of so 
many of us!


On 17 Nov 2022 at 18:21, Sebastian Karcher wrote:

> 
> I know what discursive footnotes are -- I've been working on citation styles for over a 
> decade ;).
> You're just slightly misunderstanding how CSL note styles work:
> As I said, this can only be achieved by employing in-text mode and using in-text citations within 
> footnotes.
> That's not the case: if you insert a citation in a note style into a footnote/endnote with 
> CSL, it behaves like an in text citation, so the above is absolutely possible without an 
> in-text style. (Slightly less complex discursive notes can also easily be generated by 
> using prefix and suffix elements in the citation as explained e.g. for Zotero here).
> The advantage of this is that a) you save yourself some time when just inserting simple 
> citation -- which, looking e.g. at your recent article in GRMS, are still the majority of 
> citations and b) you maintain compatibility with other note-based styles commonly used 
> in the humanities. I don't think a scenario where someone wants to move from, say, 
> Chicago Manual (note) to this (or the JHL or the GRMS) style is implausible or even 
> rare, and designating them as in-text breaks that.
> -
> Reply to this email directly, view it on GitHub, or unsubscribe.


-- 
PD Dr Stefan Hagel
PI Ancient Music 
Österreichisches Archäologisches Institut | Austrian 
Archaeological Institute
Österreichische Akademie der Wissenschaften | Austrian 
Academy of Sciences

Hollandstraße 11-13/6.20, 1020 Wien, Österreich | Vienna, 
Austria
T: +43 1 51581-2341 
***@***.*** | https://www.oeaw.ac.at/oeai/
https://homepage.univie.ac.at/stefan.hagel/

adam3smith:
I've spent some time thinking about this and I'm sorry to say that I remain unconvinced. My biggest concern here is consistent behavior for users: we have a fairly large number of styles that use author-date in notes, and as far as I know, all of them except the three mentioned here insert notes. That includes most styles covering classics, such as Journal of Roman Archaeology, [Oxford Studies on the Roman Economy](https://www.zotero.org/styles/oxford-studies-on-the-roman-economy), [Religion in the Roman Empire](https://www.zotero.org/styles/religion-in-the-roman-empire), [The Journal of Roman Studies](https://www.zotero.org/styles/the-journal-of-roman-studies), [Bulletin de correspondance hellénique](https://www.zotero.org/styles/bulletin-de-correspondance-hellenique), [Archeologia Classica](https://www.zotero.org/styles/archeologia-classica) to name just a couple of the ones that I found in a couple of minutes.  That's in addition to author-date styles in notes used in other disciplines (we have a bunch of them in political science, e.g.) and by various universities. 
I think wanting all such styles to work the same way isn't technical -- it really comes down to not behaving in a way that's unintelligible for users, i.e. a good user experience. And even if you think compatibility with short note styles like Chicago isn't a big deal (of which I'm still not convinced), surely we'd want compatibility between styles using the same citation patterns in the same field.

So I'll insist here that both this and your other style really should be note styles and that we'll switch over the JHS style. 
Yes, this does come at the cost of citations in image captions, but image captions are in a weird separate category of citations anyway: many styles want completely different formats for citations there, at least [Zotero seems wary](https://www.zotero.org/support/kb/debugging_broken_documents) about people using automated citations in them, and most journal submission systems require you to copy & paste captions separately anyway), so while I won't claim they're useless -- you clearly like them -- I don't think they're a dealbreaker.

PR #6237, opened 2022-09-21
Create greek-and-roman-musical-studies.csl

StefanHagel:
Style for Brill's journal Greek And Roman Musical Studies, https://brill.com/GRMS

### DIFF
--- a/greek-and-roman-musical-studies.csl
+++ b/greek-and-roman-musical-studies.csl
@@ -6,7 +6,6 @@
     <title-short>GRMS</title-short>
     <id>http://www.zotero.org/styles/greek-and-roman-musical-studies</id>
     <link href="http://www.zotero.org/styles/greek-and-roman-musical-studies" rel="self"/>
-    <link href="http://www.zotero.org/styles/harvard-cardiff-university" rel="template"/>
     <link href="https://brill.com/fileasset/downloads_products/Author_Instructions/GRMS.pdf" rel="documentation"/>
     <author>
       <name>Stefan Hagel</name>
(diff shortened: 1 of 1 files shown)


greek-and-roman-musical-studies.csl
```diff
@@ -0,0 +1,291 @@
...
+        <if type="bill book graphic legal_case legislation motion_picture report song" match="any">
+          <group delimiter=". " prefix=" ">
+            <group prefix=" ">
+              <text macro="title" quotes="false"/>
+              <number variable="volume" form="numeric" prefix=" " suffix=" "/>
+              <text macro="editor"/>
+              <text macro="translator"/>
+              <text macro="additional-info"/>
+              <choose>
+                <if type="motion_picture" match="any" variable="URL">
```
adam3smith, line 215: Since the `access` macro already contains a test for URL, this test is unnecessary

greek-and-roman-musical-studies.csl
```diff
@@ -0,0 +1,291 @@
...
+            <text variable="container-title" font-style="italic"/>
+            <group prefix=" ">
+              <text variable="volume"/>
+              <text variable="issue" prefix="(" suffix=")"/>
+            </group>
+            <group prefix=", ">
+              <label variable="locator" form="short"/>
+              <text macro="pages"/>
+            </group>
+            <group prefix=". ">
```
adam3smith, line 283: remove groups with only one item in them (also, there are two prefixes here; don't rely on the processor to remove unnecessary double spacing)

adam3smith (changes requested):
Thanks! Some questions, some clean-up requests

PR #6244, opened 2022-09-24
Create gnosis-journal-of-gnostic-studies.csl

POBrien333:
via https://forums.zotero.org/discussion/99911/style-request-gnosis-journal-of-gnostic-studies#latest

### DIFF
--- a/gnosis-journal-of-gnostic-studies.csl
+++ b/gnosis-journal-of-gnostic-studies.csl
@@ -59,19 +59,10 @@
     </choose>
   </macro>
   <macro name="publisher">
-    <choose>
-      <if type="book chapter paper-conference" match="any">
         <group delimiter=": ">
           <text variable="publisher-place"/>
           <text variable="publisher-place"/>
         </group>
-      </if>
-      <else>
-        <group delimiter=", " prefix=", ">
-          <text variable="publisher-place"/>
-        </group>
-      </else>
-    </choose>
   </macro>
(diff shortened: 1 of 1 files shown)


denismaier:
Is that ready?

POBrien333:
Yes, usually I put "WIP" or the like in the title or comments.

gnosis-journal-of-gnostic-studies.csl
```diff
@@ -0,0 +1,153 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
```
denismaier, line 2: The style guide allows either British or American English. Would it make sense to drop the `default-locale`?

denismaier (changes requested):

gnosis-journal-of-gnostic-studies.csl
POBrien333: The only real parts where that shows is long dates where the day and month switches as well as how quotation marks are shown and handled.
This style has neither long dates (like e.g. for websites) nor quotation marks. So I don't think it matters too much in the case of this citation style and is more about the written text.
I'd leave as is and if someone comes requesting a change that's easy to make.

denismaier:
Thanks.

PR #6245, opened 2022-09-24
Update Cell to numeric-superscript style

adam3smith:
This comes from the editorial office, change will be on website shortly.
Stem Cell Reports, now independent, is the only style that remains with the old Cell format. I'm linking Negative Results to that style (it's an independent journal with poor instructions but looks to follow Cell)


bwiernik:
Thanks!

PR #6247, opened 2022-09-25
Update engineering-technology-and-applied-science-research.csl

dionojs:
change of DOI prefix and update of summary


dionojs (commented):
seems ok

dionojs:
what do I have to do now in order for the csl to be updated in the zotero repository?

adam3smith:
Thanks -- update will appear on the repository shortly

PR #6248, opened 2022-09-28
ACS, AMA, Vancouver: Remove hardcoded space after `citation-number`

dstillman:
https://github.com/zotero/zotero/issues/2633
https://discourse.citationstyles.org/t/space-after-first-field-in-plain-text-mode-when-using-second-field-align/1762

No longer necessary for citeproc-js after Juris-M/citeproc-js#207

### DIFF
--- a/american-chemical-society.csl
+++ b/american-chemical-society.csl
@@ -148,7 +148,7 @@
   </citation>
   <bibliography second-field-align="flush" entry-spacing="0">
     <layout suffix=".">
-      <text variable="citation-number" prefix="(" suffix=") "/>
+      <text variable="citation-number" prefix="(" suffix=")"/>
       <text macro="author" suffix=" "/>
       <choose>
         <if type="article-journal review" match="any">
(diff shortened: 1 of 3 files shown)


dstillman:
(Seems like the test runner is not running the updated citeproc-js.)

adam3smith:
> I guess the test runner uses citeproc-ruby, which doesn't have an equivalent fix?

Yes, we went with that for simplicity because the rest of the tests are in ruby

bwiernik:
@adam3smith do we have a doc anywhere listing known behavior difference between citeproc-ruby and citeproc-js?

denismaier:
Could someone make citeproc-ruby's behaviour consistent with citeproc-js? Or is this impossible due to whatever reason?

adam3smith:
@bwiernik -- no. The issue we're seeing the most is a slightly different interpretation of the relationship between groups and `choose` (IIRC, citeproc ruby treats `choose` like a group for delimiter purposes). Beyond that, we should be able to look at the test suite, no?

@denismaier almost certainly possible (with the possible exception of some pretty marginal stuff) -- I'm pretty sure Sylvester isn't interested/doesn't have time, though, so it'd have to be someone else.

PR #6250, opened 2022-09-29

ssha4900:
Replaced form="short" to "symbol" to always show § instead of par intext


adam3smith:
Thanks!

PR #6252, opened 2022-09-29
Create american-medical-association-no-url-alphabetical.csl

POBrien333:
via https://forums.zotero.org/discussion/99783/style-update-request-american-journal-of-sport-medicine#latest


adam3smith:
Hmm -- shouldn't the title for the AMA style be "no DOI", which is the more significant difference?

POBrien333:
I'll let you decide on what's best

adam3smith:
didn't realize we already have a no URL version, so let's stay with that label even if it's awkward.

adam3smith:
Thanks!

PR #6257, opened 2022-10-04
Create bibliotheque-de-la-societe-entomologique-de-france.csl

KwAntynn:
New csl, only a small adaptation of bulletin-de-la-societe-entomologique-de-france.csl

### DIFF
--- a/bibliotheque-de-la-societe-entomologique-de-france.csl
+++ b/bibliotheque-de-la-societe-entomologique-de-france.csl
@@ -128,14 +128,14 @@
       </group>
     </layout>
   </citation>
-  <bibliography delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-min="12" et-al-use-first="6" subsequent-author-substitute="–––">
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks! Could you clarify what exactly this is for? I couldn't find a publication by this name, so what would this be used for? Also, the documentation still links to the Bulletin -- is this style documented somewhere online?

KwAntynn:
Dear Sebastian,

Thank you for your email. In fact, this style is not for a journal but for the online consultation of our library (using zotpress). (The access to the library of our scientific society was restricted to our members, but before the end of the year it will be free of access.) We use the same style than the Bulletin, but adding the location of the book. 

After the submission, I saw it could not respect the criteria of inclusion, but I didn’t find, or able to make, another solution. Is it ok for you?

Best regards,
Quentin

adam3smith:
To use a style in ZotPress, you don't have to have it on the repository -- you can just add it using a direct link to the CSL, i.e. in your case https://raw.githubusercontent.com/KwAntynn/styles/patch-2/bibliotheque-de-la-societe-entomologique-de-france.csl

(There's a thread on this in the ZotPress forums as well I'm pretty sure)

KwAntynn:
Thank you very much for your help. It effectively works; I would have search more.

Best regards,
Quentin

PR #6259, opened 2022-10-04
Create eurosurveillance.csl

Polperobis:
This is a new style

### DIFF
--- a/eurosurveillance.csl
+++ b/eurosurveillance.csl
@@ -1,0 +1,270 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" et-al-min="7" page-range-format="minimal" demote-non-dropping-particle="sort-only" default-locale="en">
+  <info>
+    <title>Eurosurveillance</title>
+    <id>http://www.zotero.org/styles/eurosurveillance</id>
+    <link href="http://www.zotero.org/styles/eurosurveillance" rel="self"/>
+    <link href="http://www.zotero.org/styles/springer-vancouver-brackets" rel="template"/>
+    <link href="https://www.eurosurveillance.org/for-authors#References" rel="documentation"/>
+    <author>
+      <name>Vincent Auvigne</name>
+      <email>vincent.auvigne@ekipaj.com</email>
+    </author>
+    <contributor>
+      <name>Charles Parnot</name>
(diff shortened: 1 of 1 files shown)


eurosurveillance.csl
```diff
@@ -0,0 +1,270 @@
...
+    <title>Eurosurveillance</title>
+    <id>http://www.zotero.org/styles/eurosurveillance</id>
+    <link href="http://www.zotero.org/styles/eurosurveillance" rel="self"/>
+    <link href="http://www.zotero.org/styles/springer-vancouver-brackets" rel="template"/>
+    <link href="https://www.eurosurveillance.org/for-authors#References" rel="documentation"/>
+    <author>
+      <name>Vincent Auvigne</name>
+      <email>vincent.auvigne@ekipaj.com</email>
+    </author>
+    <contributor>
```
POBrien333, line 13: You can delete Charles from the contributors. 
Contributors are acknowledged through the template link in line 7.

Polperobis:
I do not understand why the tests failed. Can someone help me ?

POBrien333:
`expected et-al-min (7) and et-al-use-first () to be of same type`

Basically, if you have et-al-min, you also need et-al-use-first.
You have that correctly set on `<bibliography....`, but have a look at your line 2. I would assume you'd just want to delete that there, since it's handled further down anyway.

eurosurveillance.csl
```diff
@@ -0,0 +1,265 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" page-range-format="minimal" demote-non-dropping-particle="sort-only" default-locale="en">
```
POBrien333, line 2: I think it requires a specific English language.
en-GB or en-US are the most common ones.
Note that they can change your style, especially how dates are shown.

Polperobis: I do not understand your comment, as I had later on a message "All checks have passed" without changing the default-locale

adam3smith:
Thanks! Given that this explicitly points to the NLM guidelines that the Vancouver styles follow, is there really none of the existing Vancouver styles we could link here? Seems unnecessary to create yet another slightly modified version if the journal links to the exact same examples.

adam3smith:
@Polperobis see my question above

adam3smith:
Closing for inactivity; we're happy to re-open if you come back to this and reply to the questions above -- just post here.

PR #6269, opened 2022-10-11
Update ieee.csl

### LINKED ISSUE #6260: ieee and ieee-with-url have more differences than just the URL
I would expect that the difference between `ieee` and `ieee-with-url` is just the absence of the URL in the former.

However, there are many more differences. I noted because `ieee` recently (?) stopped rendering the title of proceedings, whereas `ieee-with-url` does this just fine.

POBrien333:
closes #6260 

See also https://forums.zotero.org/discussion/100349/not-all-fields-of-a-reference-is-put-into-the-reference-entry-in-word#latest

### DIFF
--- a/ieee.csl
+++ b/ieee.csl
@@ -177,7 +177,7 @@
       <if type="paper-conference speech" match="any">
         <choose>
           <!-- Published Conference Paper -->
-          <if variable="collection-editor editor editorial-director issue page volume" match="any">
+          <if variable="container-title" match="any">
             <group delimiter=", ">
               <group delimiter=" ">
                 <text term="in"/>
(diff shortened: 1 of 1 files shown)


ieee.csl
```diff
@@ -177,7 +177,7 @@
       <if type="paper-conference speech" match="any">
         <choose>
           <!-- Published Conference Paper -->
-          <if variable="collection-editor editor editorial-director issue page volume" match="any">
```
adam3smith, line 180: @bwiernik made this change explicitly in https://github.com/citation-style-language/styles/pull/4867 -- I just want to confirm that this is no longer necessary now that we have event-title separately from container-title, yes?

adam3smith:
OK, don't want to leave this broken any longer and I'm pretty sure it's right

PR #6273, opened 2022-10-14
Update and rename dependent/veterinary-record-open.csl to veterinary-…

POBrien333:
…record-open.csl

via https://forums.zotero.org/discussion/99804/style-error-veterinary-record-open#latest

Apparently this was already noticed a while ago:
https://github.com/citation-style-language/journals/blob/master/bmj/_skip.txt

Sample paper:
https://bvajournals.onlinelibrary.wiley.com/doi/epdf/10.1002/vro2.27


adam3smith:
Thanks!

PR #6275, opened 2022-10-14
Create rassegna-degli-archivi-di-stato-bibliografia-generale.csl

Pepppino:
Style for general and reasoned bibliographies on the Rassegna degli archivi di Stato


adam3smith:
Thanks (and sorry for the long wait) -- could you explain a bit more how this is different from the existing style of the same name and how it would be used?

Pepppino:
Regarding the notes, there is no difference. This style is meant for the creation of stand-alone bibliographies, intended to be a scientific work in themselves and the _Rassegna degli Archivi di Stato_, in that case, asks for the publications to be ordered chronologically, while the bibliographies at the end of articles (covered in my other style, which has been already published) are to be ordered alphabetically by author.

adam3smith:
(ignore the failing test)

adam3smith:
Thanks!

PR #6278, opened 2022-10-21
Hanging indent was missing in style

RonaldVisser:
Added hanging indent, according to the style on website of the journal (https://www.editorialmanager.com/iawa/default2.aspx).


adam3smith:
Thanks!

PR #6279, opened 2022-10-21
Update international-union-of-crystallography.csl

POBrien333:
via https://forums.zotero.org/discussion/100581/citing-webpage-in-iucr-journal#latest


adam3smith:
Thanks!

PR #6289, opened 2022-11-01

### LINKED ISSUE #6288: Genre displayed twice in title-less thesis in chicago-author-date
Given this CSL:

```json
[
  {
    "author": [
      {
        "given": "John",
        "family": "Doe"
      }
    ],
    "type": "thesis",
    "genre": "PhD thesis",
    "id": "doe2018",
    "issued": { "date-parts": [[2018]] }
  }
]
```

The `chicago-author-date` style gives the following bibliography:

> Doe, John. 2018. PhD thesis. PhD thesis.

Is this expected?

POBrien333:
fix #6288

I don't get why chicago-note or chicago-full-note don't have that problem though...

### DIFF
--- a/chicago-author-date.csl
+++ b/chicago-author-date.csl
@@ -220,7 +220,7 @@
     <choose>
       <if variable="title" match="none">
         <choose>
-          <if type="personal_communication thesis" match="none">
+          <if type="personal_communication speech thesis" match="none">
             <text variable="genre" text-case="capitalize-first"/>
           </if>
         </choose>
--- a/chicago-author-date.csl
+++ b/chicago-author-date.csl
@@ -220,7 +220,7 @@
     <choose>
       <if variable="title" match="none">
         <choose>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks -- I want to take a closer look at why the code is there in the first place. It seems weird to put genre in the place of the title as a general rule and there's another see of code further down in the description that prints genre (but already has thesis removed)

POBrien333:
Since we're working on Turabian/Chicago, would be great if we could also close this issue after you had time to review. :)

bwiernik:
Did that code originate in APA.CSL where I use something like it to print things like "[Photograph of a man with a hat]" for titleless (usually archival) items?

adam3smith:
Thanks! The code is originally for letters, btw., which often dont' have a title. It's going to rarely apply to presentations and barely ever for theses (I hope your thesis has a title!) But doesn't hurt to fix it

PR #6292, opened 2022-11-04
Update biochemical-society-transactions.csl

POBrien333:
via https://forums.zotero.org/discussion/100877/style-request-biochemical-society-transactions-new#latest

Only adapated for article-journal due to lack of documentation/examples


adam3smith:
(note to self: ok after fixing webpage)

adam3smith:
Thanks!

PR #6293, opened 2022-11-07
Create genetics-and-molecular-biology-without-doi

natbalb:
New version of Genetics and Molecular Biology style without DOI in the bibliography.


adam3smith:
Thanks! What is this for? If the journal actually doesn't want DOIs, we can remove them from the main style -- no need for a separate one. If the journal wants DOIs and this is a personal preference, we wouldn't take it for the repository.

natbalb:
Hi Sebastian, thanks for replying! The journal actually doesn't want DOIs
so it would be great if they were removed from the main style. Sorry for
the confusion, I assume I should have requested an edition instead of
creating a new style. Thank you, again!

PR #6294, opened 2022-11-07
Update smithsonian-institution-scholarly-press-author-date.csl

POBrien333:
via https://forums.zotero.org/discussion/comment/418459/#Comment_418459


adam3smith:
The author date style looks great, but shouldn't the note style also remove quotes for articles and chapters?

adam3smith:
Thanks!

PR #6296, opened 2022-11-08
Create the-journal-of-law-medicine-ethics.csl

POBrien333:
via https://forums.zotero.org/discussion/100132/style-request-journal-of-law-medicine-ethics#latest


adam3smith:
Thanks! The Journal is all over the place with and without "the" but in citations it always appears without, so I went with that.

PR #6298, opened 2022-11-09
Update ucl-university-college-harvard.csl

ssha4900:
Made a small change to disambiguate-add-givenname, to avoid Zotero from creating first name/initials to in-line citations (When Authors had same last name).


adam3smith:
Thanks!

PR #6301, opened 2022-11-10
Create european-review-of-international-studies.csl

POBrien333:
via https://forums.zotero.org/discussion/99360/style-request-european-review-of-international-studies#latest


adam3smith:
Thanks!

PR #6303, opened 2022-11-13
Create descending-publications-based-on-ieee-with-url.csl

AndsIO:
Many researchers prefer a descending publication list on their websites so that the total number and most recent papers published are visible at first glance. This improves the visibility of one's research for funding bodies.

### DIFF
--- a/descending-publications-based-on-ieee-with-url.csl
+++ b/descending-publications-based-on-ieee-with-url.csl
@@ -4,7 +4,7 @@
     <title>Descending Publications based on IEEE with URL</title>
     <id>http://www.zotero.org/styles/descending-publications-based-on-ieee-with-url</id>
     <link href="http://www.zotero.org/styles/descending-publications-based-on-ieee-with-url" rel="self"/>
-    <link rel="template" href="https://www.zotero.org/styles/ieee-with-url"/>
+    <link rel="template" href="http://www.zotero.org/styles/ieee-with-url"/>
     <link href="https://ieeeauthorcenter.ieee.org/wp-content/uploads/IEEE-Reference-Guide.pdf" rel="documentation"/>
     <link href="https://journals.ieeeauthorcenter.ieee.org/your-role-in-article-production/ieee-editorial-style-manual/" rel="documentation"/>
     <author>
@@ -190,9 +190,6 @@
   </macro>
   <macro name="doi">
     <text variable="DOI" prefix="doi: "/>
-  </macro>
-  <macro name="citation-number">
-    <text variable="citation-number"/>
   </macro>
   <citation collapse="citation-number">
     <sort>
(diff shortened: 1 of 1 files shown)


AndsIO:
Why is it not working in the bot's preview but in the Mendeley's online CSL Editor?

adam3smith:
Thanks!
So I appreciate the reason this is useful, but I don't think we'll take this for the official repository. We have one chronological style for CVs (APA) on the repository, but there's just no way we can keep adding one whenever someone prefers a different citation style for their CV and keep the repository manageable.
You're more than welcome to host & advertise a copy on your own repo, of course (several of us who maintain/regularly contribute here do the same for some custom styles).

As for the automated preview -- that's doing exactly what you tell it, i.e. sorting by ascending issue date (the citation number then becomes irrelevant). No idea why the Mendeley Editor behavior is different: it shouldn't be.

AndsIO:
> Thanks! So I appreciate the reason this is useful, but I don't think we'll take this for the official repository. We have one chronological style for CVs (APA) on the repository, but there's just no way we can keep adding one whenever someone prefers a different citation style for their CV and keep the repository manageable. You're more than welcome to host & advertise a copy on your own repo, of course (several of us who maintain/regularly contribute here do the same for some custom styles).
> 
> As for the automated preview -- that's doing exactly what you tell it, i.e. sorting by ascending issue date (the citation number then becomes irrelevant). No idea why the Mendeley Editor behavior is different: it shouldn't be. Edit -- a sorry, I see what you mean, you want the citation numbers reversed. I think that's actually a bug in the CSL processor for the editor. The above behavior seems right to me.

Sorry for the confusion. I tried to follow the solution from the zotero forum:
https://forums.zotero.org/discussion/71443/reverse-number-bibliography
With that workarorund the "Mendeley Code Editor", the "citationstyles.org Code Editor" and the github-Bot give me three different output examples. Not sure why. Thats what I try to achieve:
CSL edited
For an publication list the most recent paper should be always on top. With descending citation number, the total number of publications would always be directly visible. In addition, the individual numbers would not change if a new publication (with an incrementing citation number) were inserted at the top. Such a type of publication list is likely to be most common among researchers. Unfortunately, I could not find a style that shows this behavior. At least four universities in Germany are starting to use the Zotero Style Repository to automatically generate publication lists from the university bibliography for the working group's web pages. But in the classic style of a scientific article, the publication lists are unsuitable for us as scientists. The possibility of a descending sorting by citation number and date would be a huge plus from my point of view.

adam3smith:
You can ignore the automated preview here when it contradicts output in Zotero or the citation editor -- it runs on a different processor that's not widely used beyond these previews; it's identical in most cases but diverges in edge cases. Given the description of the behavior in the Zotero thread that does appear to be on purpose, so it's likely stable and you can rely on it.

As for displaying bibliographies on university websites-- the Zotero API can use custom-hosted styles, and other APIs shouldn't rely on styles being hosted on a particular repository either: having a particular custom style available on our repository shouldn't affect its usability for any given tool -- where it does, that's a design flaw of the tool that we're not the right place to work around.

PR #6304, opened 2022-11-14
Update zoological-journal-of-the-linnean-society.csl

aleferreiro:
Change the use of "et al." for in-text citations to a minimum of 3 authors. Turn to bold type, the authors and date in the bibliography.

### DIFF
--- a/zoological-journal-of-the-linnean-society.csl
+++ b/zoological-journal-of-the-linnean-society.csl
@@ -16,7 +16,7 @@
     <category field="biology"/>
     <issn>0024-4082</issn>
     <eissn>1096-3642</eissn>
-    <updated>2013-10-28T11:00:12+00:00</updated>
+    <updated>2016-04-26T04:54:18+00:00</updated>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks! What's the basis for the et al change? The author instructions say:
> When papers are by three authors, use all names on the first mention and thereafter abbreviate to the first name et al. For papers by four or more authors, use et al. throughout.

That's what the current style does. Your change would be unusual.

aleferreiro:
You´re welcome Adam! Author instructions detailed in the following [link](https://academic.oup.com/zoolinnean/pages/General_Instructions) state that:

> In the text, give references in the following forms: 'Stork (1988) said', 'Stork (1988: 331)' where it is desired to refer to a specific page, and '(Rapport, 1983)' where giving reference simply as authority for a statement. Note that names of joint authors are connected by '&' in the text. **For papers by three or more authors, use et al. throughout**.

That is the reason why I propose the et al change.

adam3smith:
Ah, thanks -- the journal moved publisher.

PR #6007, opened 2022-04-06
Create intellect-newgen-style.csl

amyrolla:
First CSL file for Intellect Books (sorry I placed this in dependent first by mistake)

PR #6308, opened 2022-11-17
Create intellect-newgen-books.csl

POBrien333:
continues #6007
Via https://forums.zotero.org/discussion/100932/style-request-intellect-books-publisher-style#latest

I've simplified this style a lot and got it more into our way of making these. It's adapted for the main item types, but more specialised item types might still be out of whack...


adam3smith:
Thanks!

PR #6311, opened 2022-11-21
Create journal-of-human-nutrition-and-dietetics.csl

POBrien333:
via https://forums.zotero.org/discussion/101288/journal-of-human-nutrition-and-dietetics#latest


adam3smith:
I don't think the BJN is suitable as a dependent: at least going by the style guide, citations should be in parentheses (though still superscript) and the year is in a different position. Did you see something else?
(The guidelines for JHND are a big mess with published articles all over the place, but I'm OK going with your current version you have for that).

adam3smith:
Thanks!

PR #6313, opened 2022-11-23
Create apa-HVL-7-nb.csl

mlje:
Høgskulen på Vestlandet APA7th Norsk-bokmål.

APA7th med norske tilpasninger.


bwiernik:
Same question as on your other style 

Before we fix the automated validation issue -- you do need a separate style for this? We already have https://www.zotero.org/styles?q=id%3Anorsk-apa-manual

PR #6314, opened 2022-11-23
Create apa-HVL-7-nn.csl

mlje:
Høgskulen på Vestlandet APA7th Norsk-nynorsk.

APA7th med norske tilpasninger.


adam3smith:
Before we fix the automated validation issue -- you do need a separate style for this? We already have https://www.zotero.org/styles?q=id%3Anorsk-apa-manual

PR #6317, opened 2022-11-26
Fix Brazilian quotes

igormcoelho:
According to locale, Brazilian quotes should be:
```
<term name="open-quote">“</term>
<term name="close-quote">”</term>
```
https://github.com/citation-style-language/locales/blob/master/locales-pt-BR.xml#L163-L164 

These are also causing error in LaTeX/Pandoc translation, as both are the same symbol... see: https://github.com/jgm/citeproc/issues/123

### DIFF
--- a/chicago-author-date.csl
+++ b/chicago-author-date.csl
@@ -65,8 +65,8 @@
       <term name="at">em</term>
       <term name="by">por</term>
       <!-- PUNCTUATION -->
-      <term name="open-quote">"</term>
-      <term name="close-quote">"</term>
+      <term name="open-quote">“</term>
+      <term name="close-quote">”</term>
     </terms>
   </locale>
   <macro name="secondary-contributors">


adam3smith:
cc @serbernardo  who introduced these originally.

igormcoelho:
Hi @adam3smith , thanks for the reply. I was wondering if this change was intentional, specially regarding LaTeX side... On plaintext, it's almost the same thing, comparing "this" and “this”, but all document writers do this automatically for us (like MS Word, LibreOffice Writer, or WPS Writer, ...), so people typically don't notice the difference. But for automated systems like LaTeX, it's fundamental to have a different symbol on open and close, since it does not do that automatically, like MS Word. 
Using the direct quote symbol `"` breaks LaTeX.

So, for us here in Brazil, we have the following pratical documents:
- ABNT NBR 6023 (2018) - https://www.normasabnt.org/abnt-nbr-6023/
- abntex latex package - https://www.ctan.org/pkg/abntex2
- specially, the abntex2cite package - https://linorg.usp.br/CTAN/macros/latex/contrib/abntex2/doc/abntex2cite.pdf
On academia, we strongly rely on abntex for local documents compatible with official rules, but for international journal it certainly varies. Regarding quotes, we typically won't use any quote on titles, but double quotes is accepted, per abntex2cite (page 18)

```
Por exemplo, para que todas as ocorrências de títulos sejam colocadas entre aspas, você pode usar o comando seguinte ainda no preâmbulo:
\newcommand{\bibtextitlecommand}[2]{‘‘#2’’}
```

See an example (from another Brazilian project of CSL ABNT implementation: https://github.com/virgilinojuca/csl-abnt)

> BRASIL. Ministério da Justiça. __Relatório de atividades__. Brasília, DF: Ministério da Justiça, 1993. 28 p.

In fact, the rules only require us to have the same typographic pattern over titles (bold, italic, etc), what certainly include quotes.

> 6-Regras Gerais de Apresentação
> ...
> 6.7 As referências, ordenadas em uma única lista, devem ser padronizadas quanto ao recurso tipográfco e à adoção dos elementos complementares. O recurso tipográfco (negrito, itálico ou sublinhado) utilizado para destacar o elemento título deve ser uniforme em todas as referências. ...

And finally, the citation-style-language/locales itself already defines quotes this way for pt-BR: https://github.com/citation-style-language/locales/blob/master/locales-pt-BR.xml#L163-L164

So, I don't know about pt (Portugal), but it's quite bad for us here in Brazil to have broken support in LaTeX, specially because `chicago-author-date.csl` is the most commonly used style in many softwares, including Pandoc.

adam3smith:
FWIW, it's valid CSL to use straight quotes as quotes (and otherwise), so pandoc/citeproc should try to get them to display (I think by substituting the Ascii character numbers?) without breaking the document in LaTeX, but that's a topic for over there.

[Wikipedia suggests curly](https://en.wikipedia.org/wiki/Quotation_mark#Portuguese) or angled quotes in Portugal, so I think we can change this.

adam3smith:
I think we'll want to remove the quotes from the `pt` locale entirely, though. I don't really see why we should/did treat them differently for one specific style. (For pt-BR, this is the same, for pt-PT, if we specify curly double-quotes, we should also specify single ones for the inner quotes)

igormcoelho:
Oh, that's really interesting wiki post @adam3smith !
Indeed, in Brazil, we don't even have this thing here « » (I need to copy and paste xD).

> The usage of curved quotation marks (ex. “quote” and ‘quote’) is growing in Portugal.[[79]](https://en.wikipedia.org/wiki/Quotation_mark#cite_note-altas-99)[better source needed] That is probably due to the omnipresence of the English language and to the corresponding difficulty or even inability of some machines (mobile phones, cash registers, calculators, etc.) to enter the angular quotation marks.
> In [Brazil](https://en.wikipedia.org/wiki/Brazil), angular quotation marks are rare, and curved quotation marks (“quote” and ‘quote’) are almost always used. This can be verified by the difference between a [Portuguese keyboard](https://en.wikipedia.org/wiki/Portuguese_keyboard_layout) (which possesses a specific key for « and for ») and a [Brazilian keyboard](https://en.wikipedia.org/wiki/Portuguese_keyboard_layout).

Regarding this...
> FWIW, it's valid CSL to use straight quotes as quotes (and otherwise), so pandoc/citeproc should try to get them to display (I think by substituting the Ascii character numbers?) without breaking the document in LaTeX, but that's a topic for over there.

Yep, maybe that's possible, but my solution was to manually replace these with a script... since both are the same symbol, it's tricky to know how to replace them, because we never know when it's opening something or ending.

I think curly quotes is really the standard, because only programmers use straight quotes ", as far as I know (does some other locale does this here?). Thanks a lot!

adam3smith:
Sorry for the delay here -- it makes more sense to fall back to the locale (which has curly quotes in Brazilian PT already as you know) then to randomly hardcode quotes into one single style (I just made that change). If the angle brackets are going out of fashion in Portugal, we can eventually adjust that locale.

PR #6318, opened 2022-11-27
Fix webpage in-text citation for council-of-science-editors-author-date.csl

POBrien333:
via https://forums.zotero.org/discussion/101393/csl-style-error-reference-with-no-printed-form#latest


adam3smith:
Thanks -- check out the simpler syntax just using title `form="short"` which falls back to `title` in the absence `title-short`

POBrien333:
excellent. wasn't aware of that :)

PR #6322, opened 2022-11-29
Update CSL for Mastozool Tropic

aleferreiro:
Changes in the names of authors, in the bibliography. XML was copied from the journal page.

### DIFF
--- a/mastozoologia-neotropical.csl
+++ b/mastozoologia-neotropical.csl
@@ -1,21 +1,28 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
   <info>
-    <title>Mastozoologia Neotropical</title>
-    <id>http://www.zotero.org/styles/mastozoologia-neotropical</id>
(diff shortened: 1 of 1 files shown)


adam3smith:
Sorry for the delay here, but I can't quite tell what's going on in this style -- it looks like you're trying to combine Mastozool. Neotropical with Journal of Mammalogy -- a bunch of things aren't quite right just starting with the metadata.

adam3smith:
@aleferreiro please see my question above

aleferreiro:
Sorry for the late response. The changes in the csl were proposed according to the [bibliography format](https://csl.mendeley.com/styles/510890671/Mastozoologia-Neotropical-2023) available at the [journal Matozoologia Neeotropical page](https://mn.sarem.org.ar/para-publicar-en-mn/). Original csl from zotero did not format int he proper font type the names of the author in the list of references. That is mainly the differences between the styles. However, the pull request was generated by copy-pasting the csl file available in the journal page I mentioned above. Sorry, if I caused some trouble with the code.

POBrien333:
@aleferreiro 
I've proposed a fix to the style on our repository. In general, it's always better to just have one place of truth and we usually ask the journals to link to the style rather than uploading it to their server and offering it from there. I understand they did this since the style was outdated, but we're more than open to fix issues. A simple report on the Zotero forums or here would suffice for that.
Are you in contact with the journal?

PR #6327, opened 2022-12-01
Update vox-sanguinis.csl

POBrien333:
via email form on csl.org (number 318, Miquel)

Very minor issues though, so I'm wondering i that's what he meant.


adam3smith:
I'm not even sure about removing the comma before editors, see e.g. doi.org/10.1111/vox.13118 which has the comma (and it looks weird without). Let's see what they say

adam3smith:
Thanks!

PR #6328, opened 2022-12-05

srk:
Add provision for the 'software' CSL type

### DIFF
--- a/royal-society-of-chemistry.csl
+++ b/royal-society-of-chemistry.csl
@@ -183,7 +183,8 @@
         <else-if type="software" match="any">
           <group delimiter=" ">
             <text variable="title" form="short"/>
-            <group delimiter=" " prefix="(version " suffix=")">
+            <group delimiter=" " prefix="(" suffix=")">
+              <label variable="version"/>
               <text variable="version"/>
             </group>
             <choose>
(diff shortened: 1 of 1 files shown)


royal-society-of-chemistry.csl
```diff
@@ -176,6 +180,30 @@
             </group>
           </group>
         </else-if>
+        <else-if type="software" match="any">
+          <group delimiter=" ">
+            <text variable="title" form="short"/>
+            <group delimiter=" " prefix="(version " suffix=")">
```
POBrien333, line 186: This group as it stands currently doesn't do anything. But you're better off using `<label variable="version"/>` anyway instead of the hardcoded version in the prefix.

srk (commented):
This seems to work for me. Replaced the hardcoded string 'version' with <label variable="version"/>

srk:
It generates the apppropriate title and version for either commercial software (with a publisher, place,etc.) or an open-source research code with a URL or DOI

royal-society-of-chemistry.csl
srk: Addressed the hardcoded 'version' issue.
For me, this actually works. Made sense to me to group together the 'title' and 'version' like this. But I would be OK with these being separate elements, so long as they are both present.

PR #6331, opened 2022-12-07
Create nys-nydanske-studier.csl

JacobThoegersen:
Used by https://www.nys.dk/, one of the leading Danish linguistics journals


JacobThoegersen (commented):
Looks good

JacobThoegersen:
Cool. Waiting for your guidance.

nys-nydanske-studier.csl
```diff
@@ -0,0 +1,431 @@
...
+    </choose>
+  </macro>
+  <macro name="chapter-prefix">
+    <choose>
+      <if type="chapter paper-conference" match="any">
+        <text term="in" text-case="capitalize-first"/>
+      </if>
+    </choose>
+  </macro>
+  <citation et-al-min="4" et-al-use-first="1" disambiguate-add-year-suffix="true" disambiguate-add-names="true" disambiguate-add-givenname="true">
```
adam3smith, line 386: This should be `et-al-min="3"` as per 5.9 p.6 of the guide, no?
More generally, I'd recommend removing all et al from the various places it's in `names` and at the top of the style and _only_ using it on `citation` and `bibliography` to improve readability and reliability of the style.

adam3smith:
Just one quick comment -- everything else looks good.

nys-nydanske-studier.csl
JacobThoegersen: Man, you are observant! I am impressed! 

But no, for some reason I can't explain, the style stipulates et-al-min=3 in in-line references (that's §5.9 as you referenced), but et-al-min=4 in the reference list. The latter is explained in §6.1 (bullet 3.3): "Ved fire eller flere forfattere/redaktører bruges ’m.fl.’ eller ’et al.’ efter første forfatter- eller redaktørnavn (fx Everaert, M.B.H. et al. 2015)" [in case of four or more authors/editors, use 'm.fl.' or 'et.al.' after the first author or editor name (e.g. Everaert, M.B.H. et al. 2015.)].

Can you help me with the most efficient and readable way to get this into the style? I would REALLY appreciate any help you can give. I know one or two things about editing a journal, but writing code is not my forte... I'll of course give it a shot if this is beyond the services you provide. But chances are, I'll only be making more inconsistencies...

In gratitude,
Jacob

adam3smith:
Thanks! You can see the cleaned up et al in this commit https://github.com/citation-style-language/styles/pull/6331/commits/c8bc1f03407b98276460b2ec35d53b77cc6bc1fe

If you could also confirm that I have the journal name right? I think it's Sprogstrudier, not just Studier, right?

JacobThoegersen (commented):
Spot on! Thanks for catching my really embarrasing error in the journal title... !!
It is "Nydanske Sprogstudier", yes; just like you had it in the code (but not quite as you have it in the message)

adam3smith:
Thanks!

PR #6332, opened 2022-12-07
Remove full stop

Hannahglad:
Small change to remove full stop after each entry in bibliography


adam3smith:
You're working with a very old version of the repository which causes problems with the automated check. This is probably just fine given how simple this change is, but could you try going to 
https://github.com/Hannahglad/styles/tree/patch-3 and see if you can sync fork/update branch?
image

PR #6334, opened 2022-12-09
Update journal-of-the-american-college-of-cardiology.csl

POBrien333:
via https://forums.zotero.org/discussion/101649/changing-the-citation-style#latest


adam3smith:
Thanks!

PR #6336, opened 2022-12-13
Create cambridge-a

EricMarcon:
Cambridge University Press (author-date) style now exists in [two flavors](https://www.cambridge.org/core/services/aop-file-manager/file/610be5e3a1acae0932de3cca/Cambridge-Reference-Styles.pdf): *CambridgeA* and *CambrigeB*.

*CambridgeB* seems to be identical to the original style (`cambridge-university-press-author-date.csl`) while *CambridgeA* uses the minimum punctuation necessary. *CambridgeA* is required to publish at least in the [Journal of Tropical Ecology](https://www.cambridge.org/core/journals/journal-of-tropical-ecology/information/author-instructions/preparing-your-materials).

This file is modified from the original `cambridge-university-press-author-date.csl` by Patrick O'Brien.


adam3smith:
@POBrien333 -- do you recall what documentation you used for [cambridge-university-press-author-date.csl](https://github.com/citation-style-language/styles/blob/master/cambridge-university-press-author-date.csl)? If this is indeed the Cambridge-B variant, I'm inclined to look at adding that to the style title (and using a similar format here)

POBrien333:
Hmm. Was looking at the history of these PRs, but cannot find it and this one link from the forum thread is useless: https://www.cambridge.org/authorhub/publish-with-cambridge/academic-books/find-an-editor

EricMarcon:
I only found [this document](https://www.cambridge.org/core/services/aop-file-manager/file/610be5e3a1acae0932de3cca/Cambridge-Reference-Styles.pdf) to describe Cambridge A and B styles. It is not dated and not complete (there's no guideline about the maximum number of authors in references for instance) but there is nothing else.

As for the genesis of cambridge-university-press-author-date.csl, there is [this thread](https://forums.zotero.org/discussion/13827/cambridge-university-press-style) on the Zotero forum (that extends from 2010 to 2017). 

I have not detected differences between cambridge-university-press-author-date.csl and the decription of CambridgeB, but I have not made a careful review of all reference types.

adam3smith:
Thanks!

PR #6337, opened 2022-12-14
Update norsk-apa-manual.csl

kbernt:
This will display f.ex.  (2021, 13. mars) and not (2021, mars 13). The first is correct according to the Norwegian APA-manual.


adam3smith:
Thanks!

PR #6338, opened 2022-12-14
Update norsk-apa-manual-note.csl

kbernt:
This should display f.ex. (2021, 13. mars) and not (2021, mars 13). The first one is correct according to the Norwegian APA7-manual (it is more in line with DD-MM commonly used in Norway).


adam3smith:
Thanks!

PR #6340, opened 2022-12-16
Update "Zeitschrift für deutsche Philologie"

denismaier:
Fixes double spaces before last editor, which appeared due to the combination of `delimiter=", "` on `cs:names` and `prefix=" "` on `cs:name`.

### DIFF
--- a/zeitschrift-fur-deutsche-philologie.csl
+++ b/zeitschrift-fur-deutsche-philologie.csl
@@ -49,8 +49,8 @@
       <if type="entry-encyclopedia" match="all" variable="author editor"/>
       <else>
         <names variable="editor translator" delimiter=", ">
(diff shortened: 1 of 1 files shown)


adam3smith:
The delimiter on cs:names shouldn't play a role in this, though (as it applies between different name types, i.e. translators and editors in this case), and if it did, the suffix on label would be worse, leading to " , " -- could you give an example of incorrect output here? And this is with Zotero?

denismaier:
Yes. With Zotero. I'll need to test with Pandoc. I can post an example, maybe on Monday. The spurious space is introduced before the second editor.

adam3smith:
Right, that makes sense. Nothing to do with the delimiter, but setting a prefix on cs:name prefixes _every_ name, hence the double space

PR #6344, opened 2022-12-21
Update australasian-journal-of-philosophy.csl

antonyeagle:
Added material to deal with series and series numbers for books and reports, including the `collection-number` variable. A couple of other minor tweaks.


adam3smith:
Thanks!

PR #6345, opened 2022-12-21
Update cambridge-university-press-author-date-cambridge-a.csl

EricMarcon:
The documentation writes that chapter pages are not preceded by "pp".
"Publisher, 10-12" is correct but "Publisher pp. 10-12" is not.

Sorry, I had missed that when I added the style.


adam3smith:
Thanks!

PR #6346, opened 2022-12-22
Update and rename world-politcs.csl to world-politics.csl

POBrien333:
https://forums.zotero.org/discussion/comment/424693/#Comment_424693


adam3smith:
(not quite sure what to do with this -- we try to avoid changing IDs to not mess with reference mangers ability to auto-update styles, but this is a mess)

bwiernik:
We should probably update the name to fix the mess in this case

adam3smith:
Thanks!

PR #6350, opened 2022-12-27
Update san-francisco-estuary-and-watershed-science.csl

POBrien333:
via https://forums.zotero.org/discussion/101925/san-francisco-estuary-watershed-science-reference-format#latest

Reworked style based on the guidelines from Feb/2022.


adam3smith:
See update in Zotero forums thread

adam3smith:
Thanks!

PR #6351, opened 2022-12-27
Create travail-et-emploi.csl

POBrien333:
via https://forums.zotero.org/discussion/101358/style-request-travail-et-emploi#latest


adam3smith:
Thanks!

PR #6352, opened 2022-12-28
Include title transliteration

robertfodor:
Guidance from APA: https://apastyle.apa.org/blog/transliterated-titles-references

The proposed solution is originally from @rrealrangel: https://gist.github.com/rrealrangel/1b11c587e6cadd1705a8d3b02ceb1717

Issues it fixes:
- https://forums.zotero.org/discussion/82405/csl-styles-that-handle-original-title
- https://forums.zotero.org/discussion/74386/how-to-reference-foreign-language-titles-according-to-apa-with-zotero

Original title appears according to APA guidelines if bib source contains "origtitle" tag  (Zotero Item Info / Extras contain "original-title")

### DIFF
--- a/apa.csl
+++ b/apa.csl
@@ -1012,7 +1012,7 @@
     </group>
     <choose>
       <if match="any" variable="original-title">
-         <text variable="original-title" prefix="[" suffix="]"/>
+        <text variable="original-title" prefix="[" suffix="]"/>
       </if>
     </choose>
   </macro>
--- a/apa.csl
+++ b/apa.csl
@@ -1010,6 +1010,11 @@
         </else>
       </choose>
     </group>
+    <choose>
+      <if match="any" variable="original-title">
+         <text variable="original-title" prefix="[" suffix="]"/>
+      </if>
+    </choose>
   </macro>
   <macro name="bracketed-intext">
     <group prefix="[" suffix="]">


adam3smith:
@bwiernik -- thoughts?

bwiernik:
Thanks, but `original-title` isn't appropriate here. The "original" title is what is stored in the `title` variable. The translation or transliteration is a user-provided translation that would be stored in something like a `translated-title` variable. That isn't currently available in CSL, so we expect such translations to be stored in `medium`, which stores the various sorts of additional explanatory information APA asks for.

robertfodor:
It'd be best if the title field had a language code mapping as well. Similarly how we have gender mapped for certain terms for French localisation. But that's a point we need to discuss in another thread.

But until a language code is available for title, this is an existing workaround for cases where you're writing a paper in English but the source is in another language. 

If you speak the source language you are supposed to reference the original title as well, but making sure the reader can discern what it is. E.g. English title in the title field and original title in the origtitle field. It's not a translation per se. However if you simply put the square brackets with the original title in the title field yourself and don't bother, the formatting will be off for books (not for articles), because the bracketed original title shouldn't be italicised per APA guidelines.

There are many more examples from the Zotero forums for using this. Also note that it uses the same logic as the reprints. Reprint date goes in date, but there's an original-date field. Not sure why this must be different. But it is your call.

Let me know your thoughts.

bwiernik:
I wrote that section of the CSL specs. User-supplied translations of non-English titles isn't what `original-title` is for (the opposite actually).

We had `translated-title` in an early draft of CSL 1.0.2, but removed it in favor of a planned "related items" system in CSL 1.1.0. We should perhaps revisit that decision and add `translated-title` to CSL 1.0.2. @adam3smith @bdarcus @denismaier 

More robust multilingual fields ala CSL-m is tentatively planned for a future version of CSL, but the timeline for that is not set.

denismaier:
> We had `translated-title` in an early draft of CSL 1.0.2, but removed it in favor of a planned "related items" system in CSL 1.1.0. We should perhaps revisit that decision and add `translated-title` to CSL 1.0.2. @adam3smith @bdarcus @denismaier

Is that possible without a version bump? I mean it should be simple to add a single variable, but is it really? If the answer to that is yes, it would probably a good move to add the variable given that the timeline for the next major version is unclear.

adam3smith:
> Is that possible without a version bump

I don't think so, no. We shouldn't have two different versions of the schema with the same version number. That's precisely what semantic versioning is for.

> I mean it should be simple to add a single variable, but is it really?

Depends on the definition of simple:
It requires updating the specs, the schema, notifying the citeprocs and making sure it's implemented in a critical mass, updating the validator, and updating the editor. None of that is hard individually, but it's a lot of moving parts.

PR #6353, opened 2022-12-28
Create life-science-alliance.csl

POBrien333:
via https://forums.zotero.org/discussion/101566/style-request-life-science-alliance#latest

Only adapted for article-journal due to lack of guidelines or examples.


adam3smith:
Thanks!

PR #6358, opened 2023-01-02
Update indian-journal-of-orthopaedics.csl

POBrien333:
via https://forums.zotero.org/discussion/94101/indian-journal-of-orthopedics-old-style#latest

Papers:
- https://link.springer.com/content/pdf/10.1007/s43465-022-00719-z.pdf
- https://link.springer.com/content/pdf/10.1007/s43465-022-00767-5.pdf


adam3smith:
initials should be with period (Hancké B.), right?

adam3smith:
Thanks!

PR #6360, opened 2023-01-03
Update frontiers.csl

POBrien333:
via https://forums.zotero.org/discussion/102071/word-plug-in-does-not-insert-n-d#latest


adam3smith:
Thanks!

PR #6364, opened 2023-01-08
Set citation givenname-disambiguation-rule=primary-name to ABNT styles

imsys:
Some ABNT styles used in Brazilian universities were missing a disambiguation-rule for the citation style.
This changes makes it fit with the rules in the [NBR 10520-2002](http://www2.uesb.br/biblioteca/wp-content/uploads/2016/05/NBR-10520-CITA%C3%87%C3%95ES.pdf).

### DIFF
--- a/universidade-de-sao-paulo-escola-de-comunicacoes-e-artes-abnt.csl
+++ b/universidade-de-sao-paulo-escola-de-comunicacoes-e-artes-abnt.csl
@@ -15,7 +15,7 @@
     <category citation-format="author-date"/>
     <category field="generic-base"/>
     <summary>De acordo com ABNT-NBR 10520.2002 and ABNT-NBR 6023.2018</summary>
-    <updated>2019-11-21T16:57:15+00:00</updated>
+    <updated>2023-01-08T04:00:00+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="pt-BR">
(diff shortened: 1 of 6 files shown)


adam3smith:
> I will soon push another commit to include ISBN and DOI to some ABNT styles that have this missing.

I think disambiguation is very likely to just be an oversight, but adding DOIs and ISBNs is a bit trickier: The local ABNT variants do exist for a reason, so you'd want to be quite sure that these are actually desired by the institutions which are specified in the respective style.

imsys:
> I think disambiguation is very likely to just be an oversight, but adding DOIs and ISBNs is a bit trickier: The local ABNT variants do exist for a reason, so you'd want to be quite sure that these are actually desired by the institutions which are specified in the respective style.

Good point. It is also better to discuss this changes first and open another pull request later if needed. 

The Alagoas version is the closest one to what I need. So, initially I was just going to fix the disambiguation for it, but I ended up getting in the flow and I did for more styles.

Then I saw some punctuation problems and I went with the flow of fixing it. The "Juiz de fora" version still has too many punctuation problems that is beyond my will to fix.

The Sergipe version has a ponctuation problem that the github bot captured well (double dots) that happens only when there is no author. It was calling a macro that was making the text bold and adding suffix=". ", so I changed to in that case to not call that macro, and I had to decide on letting the title bold or not. According to [NBR 6023:2018](https://www.ufpe.br/documents/40070/1837975/ABNT+NBR+6023+2018+%281%29.pdf/3021f721-5be8-4e6d-951b-fa354dc490ed) (page 40), it should not be bold in this case, so I left it without bold.

So those are the changes I did.

adam3smith:
Hi -- thanks again for this and sorry for the delay: is there any reason you're removing `text-case` on the given name completely for some of these and adding `capitalize-first` to others? I'd go with the former throughout, and in that case you can just completely remove that line (since it doesn't determine whether givenname gets printed, just how it is formatted/styled)

imsys:
I remember it was something related to compound names "SILVA, Carlos da", I will do some tests to check if there was a reason for some having "capitalize-first".

adam3smith:
@imsys are you still planning on working on this?

adam3smith:
@imsys I'd still be inerested in this, but the release of the ABNT 2023 norm since complicates this a bit -- any thoughts? Otherwise I'd be inclined to close this instead.

p3palazzo:
I think #6608 supersedes this.

PR #6367, opened 2023-01-11
Update open-window.csl

Griever158:
The previous style was created by an ex-employee on their own. This proposed version aligns correctly with the 2022 Open Window Referencing Style, as confirmed by myself, Karl van Heerden, the subject manager for Visual Culture and member of the referencing team for the Open Window. For any queries, please feel free to contact me at karl@openwindow.co.za

### DIFF
--- a/open-window.csl
+++ b/open-window.csl
@@ -1,4 +1,3 @@
-
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-GB">
   <!-- This style was edited with the Visual CSL Editor (https://editor.citationstyles.org/visualEditor/) -->
--- a/open-window.csl
+++ b/open-window.csl
@@ -231,8 +231,8 @@
   </macro>
   <citation disambiguate-add-year-suffix="true" year-suffix-delimiter="," disambiguate-add-names="true" disambiguate-add-givenname="false" collapse="year-suffix">
     <sort>
-      <key macro="date-issued-no-parentheses"/>
       <key macro="author"/>
+      <key macro="date-issued-no-parentheses"/>
     </sort>
     <layout prefix="(" suffix=")" delimiter="; ">
       <group delimiter=" ">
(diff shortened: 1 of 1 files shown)


open-window.csl
```diff
@@ -177,38 +146,104 @@
...
+    </choose>
+  </macro>
+  <macro name="author-count">
+    <names variable="author">
+      <name form="count"/>
+    </names>
+  </macro>
   <citation disambiguate-add-year-suffix="true" year-suffix-delimiter="," disambiguate-add-names="true" disambiguate-add-givenname="false" collapse="year-suffix">
     <sort>
       <key macro="date-issued-no-parentheses"/>
```
adam3smith, line 234: The sorting here is incorrect, isn't it? Style guide says to sort alphabetically (7.1)

open-window.csl
```diff
@@ -177,38 +146,104 @@
...
       <key macro="author"/>
     </sort>
     <layout prefix="(" suffix=")" delimiter="; ">
```
adam3smith, line 237: And the delimiter here should be a comma I think

open-window.csl
```diff
@@ -260,36 +293,44 @@
...
             <text macro="title"/>
             <text variable="genre"/>
-            <text variable="publisher"/>
+            <text macro="publisher"/>
           </else-if>
-          <else-if type="chapter paper-conference entry-encyclopedia entry-dictionary" match="any">
+          <else-if type="chapter entry-encyclopedia entry-dictionary" match="any">
             <text macro="title"/>
             <group delimiter=" ">
               <text term="in" text-case="capitalize-first"/>
```
adam3smith, line 325: Book chapters look wrong -- should be lowercase in, lowercase edited by and commas between elements, right?

adam3smith (changes requested):
Sorry for the delay in reviewing. A couple of things aren't quite right yet, best I can tell -- could you please check?

open-window.csl
Griever158: Hi Adam! Thank you so much for getting back to me! You are correct, the ordering of citations should be author then year. I have made this change, you should see it already?

open-window.csl
Griever158: The delimiter here should be a semicolon, this is correct.

open-window.csl
Griever158: So, funny story. They were wrong, but last week we revised our style guide, and we've actually changed it to be consistent with how it is formatted here: _Title_. Edited by J Smith. It's more consistent with translated works and interviews now. So this format is now correct.

adam3smith:
Thanks!

PR #6368, opened 2023-01-11
Update apa-no-ampersand.csl

ChrisAnderson-NOAA:
With disambiguate-add-names="true" et al was not functioning properly for >3. I believe default should be false to ensure it is first author + et al.

### DIFF
--- a/apa-no-ampersand.csl
+++ b/apa-no-ampersand.csl
@@ -1860,7 +1860,7 @@
       <text variable="locator"/>
     </group>
   </macro>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks, but no, the disambiguation rule for adding names in APA specifically extends beyond the number of authors displayed before et al. The style is correct as is. See e.g. https://apastyle.apa.org/style-grammar-guidelines/citations/basic-principles/same-year-first-author

PR #6370, opened 2023-01-12
Create dedicated Basque APA file

mirenbz:
Basque requires the order of some elements to be changed. This is not possible to localize and have in a single file, as I was explained [in the official forums](https://forums.zotero.org/discussion/102287/where-is-the-p-pp-prefix-located-in-apa-7th-edition). That's why I'm proposing this dedicated APA file for Basque that fixes the order of some elements that are currently wrong for Basque.

I have also made some corrections in the specific locale to match: https://github.com/citation-style-language/locales/pull/274/commits/cc60270ee41dc2456b76a904b31efb293d7cf384


bwiernik:
Could you add a comment before line 30 of the style detailing the changes that are made for Basque localization here?

(Like what is already there about item types and the basic APA structure)

That will help me in two ways:
1. I will be able to keep this style up to date with changes to the main APA.CSL
2. If features become available that make the separate style file unnecessary because the localization needs are built into the processor, we can retire the separate style

mirenbz:
Done!

adam3smith:
Thanks & sorry for the delay

PR #6382, opened 2023-01-21
Create gender-and-society.csl

celia-bouchet:
Citation guidelines for the Gender & Society journal

### DIFF
--- a/gender-and-society.csl
+++ b/gender-and-society.csl
@@ -60,18 +60,6 @@
       </if>
     </choose>
   </macro>
-  <macro name="editor">
-    <names variable="editor">
-      <name name-as-sort-order="first" and="text" sort-separator=", " delimiter=", " delimiter-precedes-last="always"/>
-      <label form="short" prefix=", "/>
-    </names>
-  </macro>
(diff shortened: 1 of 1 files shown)


celia-bouchet:
I tried to understand what the "may not have any unused macros" correspond to (the CSL style validor was fine)... From what I read [here](https://github.com/citation-style-language/styles/pull/3772), the problem might come from the fact that I rely on a Chicago style (16th edition) to construct this one, and Chicago styles have an exception to the "no unused macro" rule. Do I still need to edit it?

adam3smith:
Not quite, no. No CSL styles should define macros that aren't used. We make an example for Chicago variants so they are easier to update at the same time, but for your style you should delete the unused macros editor and translator

celia-bouchet:
> Not quite, no. No CSL styles should define macros that aren't used. We make an example for Chicago variants so they are easier to update at the same time, but for your style you should delete the unused macros editor and translator

Got it ! Thanks a lot.

adam3smith:
Thanks -- this looks great

PR #6385, opened 2023-01-25
Update geophysical-journal-international.csl

POBrien333:
https://forums.zotero.org/discussion/comment/426797/#Comment_426797


adam3smith:
Thanks!

PR #6388, opened 2023-01-27
Update apa.csl

goghard:
This is for localizing the “Original work published” text that appears at the end of a reference when the original date of a work is present. I created the term "original-work-published" within each locale section. Then, I used Google Translate to add a translation to each locale (but for Spanish, which is my first language). Next, in the macro “publication history", I replaced the “text value” with the correspondent “text term” I created.  I also corrected some mistakes in Latin (la) and Lithuanian (lt) locales.

### DIFF
--- a/apa-6th-edition-no-ampersand.csl
+++ b/apa-6th-edition-no-ampersand.csl
@@ -1096,14 +1096,8 @@
     <!--This should be localized -->
     <choose>
       <if type="bill legal_case legislation" match="any"/>
-      <else-if type="interview motion_picture song" match="any">
-        <text value="Original work recorded"/>
-      </else-if>
-      <else-if type="broadcast">
-        <text value="Original work broadcast"/>
-      </else-if>
       <else>
-        <text value="Original work published"/>
+        <text term="original-work-published" text-case="capitalize-first"/>
       </else>
     </choose>
   </macro>
--- a/apa-6th-edition.csl
+++ b/apa-6th-edition.csl
@@ -1095,14 +1095,8 @@
     <!--This should be localized -->
     <choose>
       <if type="bill legal_case legislation" match="any"/>
-      <else-if type="interview motion_picture song" match="any">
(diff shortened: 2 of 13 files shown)


adam3smith:
Thanks! Unfortunately, I don't think this is the way to go here: replacing the text value with the term makes sense, but there's no need to define the term in the style: it should be added in the locales themselves; no reason to think it'd be different for APA. I think we're a bit wary about trusting automated translation with this, but I'd be willing to discuss that for the locales.

For the LT/LA changes, do you speak those languages?

goghard:
It’s ok. This is just my first attempt to contribute somehow to the project, and I’m pleased to hear that replacing the text value with the term makes sense to you. I’m not sure whether other styles use a text string for the original date, but I’ll explore locales to see if I can learn something.  

Regarding LT/LA, I don’t speak these languages. When I was copying and pasting translations, I noticed that Latin words appeared in the Lithuanian section, so I did my best to address the problem. 

Thanks for answering!

goghard:
Hi Sebastian! I have been learning about Zotero locales, and now I can see why translating the term inside the CSL file is not the way to go. Thanks for explaining to me that. Now, I am limiting my proposal only to one line of code. Note that I added the parameter text-case="capitalize-first." I hope that you find this version more useful. 
My next step will be to propose a translation of the term in the Locales section. I’ll only contribute to Spanish, since that’s the other language I speak. Thanks again!

adam3smith:
@bwiernik this lgtm, but not sure where you are with your wider APA updates post CSL 1.0.2?

goghard:
Thank you again, @adam3smith, and nice to meet you @bwiernik. As I mentioned before, this is my first commit to the CSL community, so I’m happy to learn, and hopefully, to contribute with something valuable at some point. Is there a way to see the wider APA updates post CSL 1.0.2?

bwiernik:
Thanks! and sorry for the delay

PR #6390, opened 2023-01-31

ssha4900:
Changed the sort order of key macros.
He had difficulties with legislation being sorted under author in the bibliography, where it should not. In Denmark we sort legislation under title (or short title). This quick fix solved it, without apparently affecting the sorting of other material in a negative way.


adam3smith:
Thanks!

PR #6392, opened 2023-02-01
Update universite-de-sherbrooke-histoire.csl

amelancon:
Remove Uppercase to Author to update to Department new citation style.


adam3smith:
Thanks!

PR #6393, opened 2023-02-01
Create tr-clacso

goghard:
The Latin American Council of Social Sciences (Consejo Latinoamericano de Ciencias Sociales - CLACSO) is one of the main publishers in Latin America. This style is for Tramas y Redes, the main journal this publisher maintains. Other publications within CLACSO also follow this style. It is based on APA (7th) with significant modifications: First name of authors appears; places are restored; original work publication dates appear inside brackets both for in-text citations and for bibliographic references; newspaper and webpage dates are localized; en-dashes are replaced with dashes (at least for Spanish and Portuguese, the main languages in which CLACSO works); and there is no dot between authors and dates in the bibliography. This version, which is the first one, was tested using only the es-ES locale.


goghard:
It appears that the style has too many changes for being considered a dependent style. I’ll try a non-dependant one.

PR #6394, opened 2023-02-01
Create tr-clacso.csl

goghard:
The Latin American Council of Social Sciences (Consejo Latinoamericano de Ciencias Sociales - CLACSO) is one of the main publishers in Latin America. This style is for Tramas y Redes, the main journal this publisher maintains. Other publications within CLACSO also follow this style. It is based on APA (7th) with significant modifications: First name of authors appears; places are restored; original work publication dates appear inside brackets both for in-text citations and for bibliographic references; newspaper and webpage dates are localized; en-dashes are replaced with dashes (at least for Spanish and Portuguese, the main languages in which CLACSO works); and there is no dot between authors and dates in the bibliography. This version, which is the first one, was tested using only the es-ES locale. 
Previously, I tried to propose the file as an APA dependent style, but it failed all automated tests.

### DIFF
--- a/tr-clacso.csl
+++ b/tr-clacso.csl
@@ -1,0 +1,1774 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" page-range-format="expanded">
+  <info>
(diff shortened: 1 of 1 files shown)


POBrien333:
Continued here: #7697

POBrien333:
@goghard 
The style is now released and available from the repository.

PR #6398, opened 2023-02-07
Create polish-archives-of-internal-medicine.csl

PolArchInternMed:

PR #6399, opened 2023-02-07
Create polish-archives-of-internal-medicine.csl

PolArchInternMed:
Corrected. From the Zotero


adam3smith:
I'm confused by the two submissions here. I've fixed up the other one https://github.com/citation-style-language/styles/pull/6398 but not clear which one I should go with?

adam3smith:
Thanks!

PolArchInternMed:
Hello again!
Would it be possible to change en dash to a hyphen in the range of pages? Apparently, when creating the final version of articles in PDFs, our script is not recognizing en dashes and we need to manually change them to hyphens...
Thank you!
Kasia

PR #6410, opened 2023-02-14
Create APA7espanish.csl

vdecafe:
I have created a Spanish version for citing in APA version 7 format. This version is especially useful for thesis papers and scientific journals in Spanish.

### DIFF
--- a/APA 7 en español.csl
+++ b/APA 7 en español.csl
@@ -4,7 +4,7 @@
     <title xml:lang="apa-agr">APAespañol</title>
     <title-short>APA esp</title-short>
     <id>http://www.zotero.org/styles/apaespanol</id>
-    <link href="http://www.zotero.org/styles/apaespanol" rel="self"/>
+    <link href="ttp://www.zotero.org/styles/APA7espanish" rel="self"/>
     <link href="http://www.zotero.org/styles/apa" rel="template"/>
     <link href="https://normas-apa.org/introduccion/normas-apa-en-espanol/" rel="documentation"/>
     <author>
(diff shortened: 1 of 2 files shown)


bwiernik:
Thanks for this. Can you explain how the existing APA 7 style (specifically the APA 7 (no ampersand) version) with the locale set to Spanish is not adequate?

vdecafe:
> Thanks for this. Can you explain how the existing APA 7 style (specifically the APA 7 (no ampersand) version) with the locale set to Spanish is not adequate?

When using the APA 7 style (no ampersand) the cites have the number of the pages.

bwiernik:
What do you mean? Do you have an example? `number-of-pages` is not used in apa.csl

vdecafe:
> What do you mean? Do you have an example? `number-of-pages` is not used in apa.csl

Yep. That what happens when I use the APA no ampersand csl in Refworks

image

bwiernik:
That's not `number-of-pages`, those are specific page numbers locators on the in text citations (CSL `locator`) (e.g., if you are quoting text 

You appear to have added page numbers when you inserted each of the citations in RefWorks. Remove those using the RefWorks citation dialog. (I'm not familiar with the RefWorks citation interface--if you aren't sure how to work with the RefWorks citation interface, talk to your librarian for help.)

In any event, when a page number locator is provided, what you are showing is correct APA style, so the apa-no-ampersand.CSL is working correctly.

bwiernik:
If you are really certain that you have not entered page number locators when adding citations with RefWorks, it's possible this is some sort of bug with RefWorks adding them incorrectly. If that's the case, I suggest changing to a more reliable reference manager, such as Zotero

PR #6417, opened 2023-02-20
Create iso690-full-note-en

Elladan1337:
An English variant of the iso690-full-note which is used for example at Charles University, Faculty of Law, as well as other universities. Given an increased interest in writing texts in English, as opposed to Czech, I believe it is useful to have an English version. The style was creating by localising the current iso690-full-note-cs into English.

### DIFF
--- a/iso690-full-note-en
+++ b/iso690-full-note-en
@@ -1,0 +1,616 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" class="note" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>ISO-690 (full note, English)</title>
+    <id>http://www.zotero.org/styles/iso690-full-note-en</id>
+    <link href="http://www.zotero.org/styles/iso690-full-note-cs" rel="self"/>
+    <link href="http://www.zotero.org/styles/iso690-full-note-sk" rel="template"/>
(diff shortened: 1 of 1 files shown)


iso690-full-note-en.csl
```diff
@@ -0,0 +1,616 @@
...
+      <email>kosatka.martin@gmail.com</email>
+      <uri>http://github.com/elladan1337</uri>
+    </author>
+    <category citation-format="note"/>
+    <category field="law"/>
+    <summary>English ISO-690, full note.</summary>
+    <updated>2021-03-07T01:38:13+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="en">
```
adam3smith, line 25: I know these are in some other styles, but could you remove the ones that are identical to the local (in, online, from, accessed), and also remove no date, adding the square brackets in the style rather than the locale.

POBrien333:
@Elladan1337 
Could you please head here and click on Sync Fork?
https://github.com/Elladan1337/citation_styles/tree/patch-1

Elladan1337:
> @Elladan1337 Could you please head here and click on Sync Fork? https://github.com/Elladan1337/citation_styles/tree/patch-1

I went ahead and did it, is there anything else that needs to be done? Thanks!

PR #6427, opened 2023-02-24

nmueller18:
Remove the automatic sort order of in-text citations. It is more hindering than helpful.


adam3smith:
Thanks! I'm curious what about sorting in-text citations didn't work for you/your authors?

PR #6428, opened 2023-02-25
Update offa.csl

nmueller18:
1. The name of the journal has been officially shortend by the editors.
2. Remove the automatic sort order of in-text citations. It is more hindering than helpful.


adam3smith:
Thanks!

PR #6432, opened 2023-02-27
Create brgm_rapports.csl

brgmist:
Style to create the bibliographies of the BRGM reports: major modifications 
-> Derived from Elsevier Havard (with titles) style
->function "et al" in italic in the citation if more than one author
->authors in bold in the bibliography
-> for the type report, the mention N° is deleted


adam3smith:
You currently have this file twice. Please also check on https://github.com/citation-style-language/styles/blob/master/STYLE_REQUIREMENTS.md with respect to file naming and documentation

PR #6437, opened 2023-03-02
Update bibliothek-forschung-und-praxis.csl

jmiba:
Alter DOI presentation using URI style according to [DOI handbook section 2.6.2](https://www.doi.org/the-identifier/resources/handbook/2_numbering#2.6.2)


adam3smith:
Yay!

PR #6438, opened 2023-03-02

jjrondeau:
Several updates, mainly adding different authorship roles, code for prepublication, software, advance online publication and oral traditions of indigenous peoples.


adam3smith:
Thanks!

PR #6439, opened 2023-03-03
Update sodertorns-hogskola-oxford.csl

GretaL:
Added suffix for year when disambiguate citation.
Added format for newspaper article for subsequent citation.


adam3smith:
Thanks!

Issue #6405, opened 2023-02-08
KSSTA style should not show publisher


ttrepka:
Hello,
we are using CSL in our product and a user had issues submitting his manuscript to "Knee Surgery, Sports Traumatology, Arthroscopy" ([see the style](https://github.com/citation-style-language/styles/blob/master/knee-surgery-sports-traumatology-arthroscopy.csl)). He was told by a KSSTA editor to remove publishers from the bibliography before his manuscript can be accepted.

I could make a PR with the `publisher` macro removed but I'm not sure it's correct. The style guide doesn't specifically mention it and I see one book example where it seems to be used (Blackwell, London):
> South J, Blass B (2001) The future of modern genomics. Blackwell, London

So perhaps it should only be used for books but not e.g. for articles?

Style guide: https://www.kssta.org/authors-homepage/references/
Some examples: https://www.kssta.org/kssta-reference-style-examples/

icornelius:
`knee-surgery-sports-traumatology-arthroscopy.csl` is among the csl files flagged as "problematic" by @adam3smith @denismaier  in #5872.

PR #6441, opened 2023-03-06
Update knee-surgery-sports-traumatology-arthroscopy.csl

POBrien333:
closes https://github.com/citation-style-language/styles/issues/6405

- add (issue)
- add DOI for article-journal without page+volume
- conditional set for publisher information to only books, chapters, paper-conf, thesis (so preprint/article doesn't print out publisher information anymore)


adam3smith:
Thanks, also for detailed notes

PR #6442, opened 2023-03-06
correct name disambiguation rules

lnrdgrnt:
in the current version, the citations disambiguate across years rather than within years. i am an academic at the university of winchester and i asked the academic skills team whether this was correct. they confirmed that it was not correct. this correction now means that citations disambiguate only within years (by adding a letter a, b, c) etc.


bwiernik:
Thanks!

PR #6443, opened 2023-03-06
Create University of Derby CTR Harvard.csl

cbderbylib:
Created a custom Cite Them Right Harvard style for University of Derby students


POBrien333:
@cbderbylib 
Can you link to the guidelines for the style and adapt the info section with all the necessary information? Currently you've just submitted a copy of the CTR style which won't work. 
See also: https://github.com/citation-style-language/styles/blob/master/STYLE_REQUIREMENTS.md

PR #6448, opened 2023-03-09

ssha4900:
Changed disambiguate-add-givenname="true" to "false". It was creating an issue where initials where add to the in-text citations when the different authors had the same surname.

### DIFF
--- a/ucl-university-college-apa.csl
+++ b/ucl-university-college-apa.csl
@@ -17,7 +17,7 @@
     </contributor>
     <category citation-format="author-date"/>
     <category field="generic-base"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
> Changed disambiguate-add-givenname="true" to "false". It was creating an issue where initials where add to the in-text citations when the different authors had the same surname.

Are you sure that's not wanted, though? Adding initials in this case is correct APA style

ssha4900:
> Are you sure that's not wanted, though? Adding initials in this case is correct APA style

Thanks for the heads up. I just checked myself and you are indeed correct. i will not go further with this.

Thanks Adam!

PR #6449, opened 2023-03-11
XJTU thesis style

pencilheart:
Here is a Ph.D. thesis style for XJTU, China


xi-an-jiaotong-university.csl
```diff
@@ -0,0 +1,560 @@
...
+    <layout suffix="." locale="zh-CN">
+      <text variable="citation-number" prefix="[" suffix="]"/>
+      <text macro="entry-layout-zh"/>
+    </layout>
+    <!-- en格式 -->
+    <layout suffix=".">
+      <text variable="citation-number" prefix="[" suffix="]"/>
+      <text macro="entry-layout"/>
+    </layout>
+  </bibliography>
```
pencilheart, line 559: It's necessary to differentiate the chinese and english in this thesis, but this CSL-M (cs:layout) can't be validated.

PR #6450, opened 2023-03-11
Update guide-des-citations-references-et-abreviations-juridiques.csl

rafjaf:
Two corrections:
- disable 'ibid.' since its behaviour is erratic when non-Zotero references are also added in the document
- correct the use of "op. cit". when a journal article has a volume and no date (such as R.W.)

### DIFF
--- a/guide-des-citations-references-et-abreviations-juridiques.csl
+++ b/guide-des-citations-references-et-abreviations-juridiques.csl
@@ -23,7 +23,7 @@
       <term name="ordinal-02">e</term>
       <term name="ordinal-03">e</term>
       <term name="ordinal-04">e</term>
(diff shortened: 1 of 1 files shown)


adam3smith:
Hmm -- I don't like taking out ibid -- that seems like an unfortunate concession to non-standard usage making live more difficult for standard usage. You could work with near-note-distance if that's a concern and only show ibid in immediately following notes?

rafjaf:
Thank you for your quick reply. That seems a good suggestion, thank you, I will look up the documentation to understand how near-note-distance works and adapt accordingly

rafjaf (commented):
Adapted ibid with near-node as in the Guide n° 289

adam3smith:
Thanks!

PR #6453, opened 2023-03-12
Create czech-journal-of-international-relations.csl

POBrien333:
https://forums.zotero.org/discussion/102914/style-request-czech-journal-of-international-relations#latest

### DIFF
--- a/czech-journal-of-international-relations.csl
+++ b/czech-journal-of-international-relations.csl
@@ -14,7 +14,7 @@
     <category field="history"/>
     <issn>0323-1844</issn>
     <eissn>2570-9429</eissn>
-    <updated>2023-03-12T20:49:42+00:00</updated>
+    <updated>2023-03-25T13:10:04+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="es">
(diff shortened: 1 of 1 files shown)


adam3smith:
Year should be in parentheses -- notice the weird `(eds., 2009)` as per the author guidelines, which I think you can only do by splitting the parentheses between two elements in a way we usually don't like to.

Also, the publication the editor links to has `et-al-min="4" et-al-use-first="3"` in the bibliography.

POBrien333:
Thanks for reviewing.
Is that weird parenthesis structure even possible? Not even with testing for `editor` and making it conditional.

adam3smith:
So I was thinking along the lines of this
For the year:

```
<if variable="author">
<date variable ="issued" date-parts="year" form="numeric" prefix="(" suffix=")"/>
</if>
<else-if variable="editor">
<date variable ="issued" date-parts="year" form="numeric" suffix=")"/>
</else-if>
<else>
<date variable ="issued" date-parts="year" form="numeric" prefix="(" suffix=")"/>
</else>
```

and then for the author macro

```
    <names variable="author">
      <name delimiter=" &#8211; " delimiter-precedes-last="never" name-as-sort-order="all"/>
      <label form="short" prefix=" (" suffix=", "/>
```

(I'm OK if you don't want to deal with this, I'm more curious to see if we can actually get this right)

POBrien333:
Sorry for keeping you wait. First was avoiding it and then had a surgery this week.
I think this actually works. Have a look :)

adam3smith:
Excellent -- I hope things are well!

PR #6454, opened 2023-03-12
Update zeitschrift-fur-geschichtsdidaktik.csl

POBrien333:
https://forums.zotero.org/discussion/103613/fehlerhafter-zitierstil#latest

waiting if there is any other issues


POBrien333:
Not sure why it's not validating, but I just noticed it says "raptor-journal" here:
grafik

PR #6455, opened 2023-03-13
Fix malformed XML in raptor-journal.csl

michel-kraemer:
According the XML specification, the `<?xml ?>` declaration must be at the beginning of an XML file and nothing should precede it.

I encountered this error when I tried to read the file using citeproc-java. It seems the Java XML parser does not like byte-order marks. I hope, the BOM was not necessary in this case but it seems the file is just normal UTF-8.


adam3smith:
Thanks -- PR for fixing this during the re-indent/reformat step in the CI would be welcome.

PR #6458, opened 2023-03-15
Fix Chicago Cases & Newspaper sorting

adam3smith:


adam3smith:
https://forums.zotero.org/discussion/103674/chicago-not-showing-paragraph-number-in-first-citation-of-a-case#latest
https://forums.zotero.org/discussion/95772/chicago-style-bibliography-error-for-unsigned-articles#latest

PR #6459, opened 2023-03-16
Update the-accounting-review.csl

POBrien333:
https://forums.zotero.org/discussion/103701/the-accounting-review-style-update-request#latest

Fix minor things as by OP


adam3smith:
Thanks!

PR #6460, opened 2023-03-16
Update survey-of-ophthalmology.csl from ama.csl to its own independent style

POBrien333:
https://forums.zotero.org/discussion/103697/survey-of-ophthalmology-style-request#latest

edit:
@adam3smith & @bwiernik 
This actually needs an alphabetical Vancouver style. Let me know if you want me to create vancouver-alphabetical.csl and use that in the dependency call or just create a custom style for this journal.


bwiernik:
Can you please make an alphabetical Vancouver style and make this dependent?

POBrien333:
Great. Separate PE created so it's cleaner long term.

adam3smith:
sorry, bad news on this one but the et-al settings don't fit: this one is et-al-min="6" et-al-use-first="3" (also makes sense to include the DOI as we generally to do Elsevier and the journal does in print. Let's turn this into an independent after all.

POBrien333:
Thanks for checking. That's created now. I just took the vancouver-alphabetical and made the required changes as above.

adam3smith:
Thanks

PR #6461, opened 2023-03-16
Create vancouver-alphabetical.csl

POBrien333:
as per #6460
Creation of an alphabetical version of vancouver as it was required for a journal.


adam3smith:
Thanks

PR #6462, opened 2023-03-17
Create isara-iso-690-en.csl

odebiez:
English version of the french "Isara - ISO-690" style submitted in August 2022 and adapted to the requirements of the Isara engineering school (Lyon, France) for use with Zotero:
    "Et al." (in italic) for 3 or more authors in the citation
    A comma between the author and the date in the citation
    Mention [Author ?] before the title if no author, editor or translator is indicated
    Mention [no date] if no date is indicated

### DIFF
--- a/isara-iso-690-en.csl
+++ b/isara-iso-690-en.csl
@@ -1,0 +1,570 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" default-locale="en-US">
+  <info>
+    <title>Isara - ISO-690 (author-date, english)</title>
+    <id>http://www.zotero.org/styles/isara-iso-690-en</id>
+    <link href="http://www.zotero.org/styles/isara-iso-690-en" rel="self"/>
+    <link href="http://www.zotero.org/styles/iso690-author-date-fr-no-abstract" rel="template"/>
+    <link href="https://doc.isara.fr/en/course-material/bibliographic-referencing-guide/" rel="documentation"/>
+    <author>
+      <name>Oriane Debiez</name>
+      <email>lasource@isara.fr</email>
+    </author>
+    <author>
+      <name>Vincent Payet</name>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="generic-base"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
And this couldn't just be achieved by removing the locale from the existing ISARA style, making it multi-lingual?

odebiez:
> And this couldn't just be achieved by removing the locale from the existing ISARA style, making it multi-lingual?

I didn't know it could be that easy as I'm not trained in CSL, so I do the best I can with online tutorials... However, we want to make sure that the bibliography is in English, not that it adapts to the language of the user's device. What do you recommend?

odebiez:
Hi @adam3smith, sorry to ask again but I could really use your help... We created this new customized style because we want to make sure that the bibliography is in English for all our students, not that it adapts to the language of the user's device. What do you recommend I I do now?

odebiez:
Hi Sebastian @adam3smith, this is just a short message, as requested in your reply on the Zotero forum (https://forums.zotero.org/discussion/comment/451114#Comment_451114).
Happy holiday season

adam3smith:
Sorry for letting this linger for so long. 
So I looked at the two styles, and they are functionally completely equivalent. We very much try to avoid this sort of duplication, which is not just inefficient but also a maintenance nightmare, especially at scale.
The only difference between the styles is the [auteur?] vs. [author?] for items without an author, which you could handle by defining and calling an `anonymous` term in both English and French in the style. 
As for 

> We created this new customized style because we want to make sure that the bibliography is in English for all our students, not that it adapts to the language of the user's device.

With almost all products used with CSL -- like Zotero, Mendeley, etc. -- users can choose the language to be used by the style; you don't have to rely on the language of the used computer. If this is absolutely not viable, turning this into a dependent style is an option, but a single style would be much preferable -- your students will be smart enough to pick the right language.

POBrien333:
@odebiez 
Thank you for your patience.
I understand that you just copied the FR style and changed the locale. However, I see that a lot of delimiters use " : " which is typical for French, but absolutely not common for EN. Can you please confirm this is indeed not correct and if so fix it throughout your style? 
Also, I see you have a lot of hardcoded labels in affixes, such as "p." for pages. Please rework these and use the correct labels.
This ideally also gets fixed in the original French style. Also include the suggestion from adamsmith to use "anonymous".
In order to remove the unrelated validation errors, please go [here](https://github.com/odebiez/styles/tree/patch-2) and click on Sync Fork.

PR #6464, opened 2023-03-20
Update harvard-stellenbosch-university.csl

trespaul:
Various corrections, per [documentation](https://libguides.sun.ac.za/c.php?g=742962&p=5316902), including:

- change "Eds." to "(eds.)" (and likewise for translator)
- never truncate author list at first, subsequently start truncating at 3 authors, then use only first author + et al
- italicise "et al."
- re-use `editor-translator` macro instead of just substituting the names
- change spacing between bibliography entries for legibility
- add TODO comment to note that, in citations, `and` must be `"symbol"` when in parentheses but `"text"` in text.

### DIFF
--- a/harvard-stellenbosch-university.csl
+++ b/harvard-stellenbosch-university.csl
@@ -286,7 +286,8 @@
           <else-if type="chapter paper-conference entry-encyclopedia entry-dictionary" match="any">
             <text macro="title"/>
             <group delimiter=" ">
-              <text term="in" text-case="capitalize-first"/>
+              <!-- TODO: editors must be before title -->
+              <text term="in" prefix=", "/>
               <text variable="container-title" font-style="italic"/>
             </group>
             <text macro="book-details"/>
--- a/harvard-stellenbosch-university.csl
+++ b/harvard-stellenbosch-university.csl
@@ -289,17 +289,17 @@
             <text variable="publisher"/>
           </else-if>
           <else-if type="chapter paper-conference entry-encyclopedia entry-dictionary" match="any">
-                          <group delimiter=", ">
-  <group delimiter=" ">
-  <group>
-            <text macro="title"/>
-              <text term="in" prefix=", "/>
+            <group delimiter=", ">
+              <group delimiter=" ">
+                <group>
+                  <text macro="title"/>
+                  <text term="in" prefix=", "/>
+                </group>
+                <group delimiter=" ">
+                  <text macro="editor-translator" suffix="."/>
+                  <text variable="container-title" font-style="italic"/>
+                </group>
               </group>
-            <group delimiter=" ">
-              <text macro="editor-translator" suffix="."/>
-              <text variable="container-title" font-style="italic"/>
-            </group>
-            </group>
               <text macro="edition-volume"/>
               <text macro="collection-details"/>
             </group>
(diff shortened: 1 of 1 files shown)


harvard-stellenbosch-university.csl
```diff
@@ -31,26 +31,26 @@
...
   <macro name="author">
     <names variable="author">
-      <name name-as-sort-order="all" and="symbol" sort-separator=", " initialize-with="." delimiter=", " delimiter-precedes-last="never" et-al-min="9" et-al-use-first="8"/>
+      <name name-as-sort-order="all" and="symbol" sort-separator=", " initialize-with="." delimiter=", " delimiter-precedes-last="never"/>
+      <et-al font-style="italic"/>
       <label form="short" prefix=" " text-case="capitalize-first"/>
       <substitute>
-        <names variable="editor"/>
-        <names variable="translator"/>
+        <text macro="editor-translator"/>
```
adam3smith, line 53: what's the advantage of this? Using straight variables is generally preferred.

harvard-stellenbosch-university.csl
```diff
@@ -284,7 +286,8 @@
           <else-if type="chapter paper-conference entry-encyclopedia entry-dictionary" match="any">
             <text macro="title"/>
             <group delimiter=" ">
-              <text term="in" text-case="capitalize-first"/>
+              <!-- TODO: editors must be before title -->
```
adam3smith, line 289: Any reason you're not doing that? This wouldn't seem hard to do.

adam3smith (commented):
A couple of questions

harvard-stellenbosch-university.csl
adam3smith: `<names>` in substitute inherits the (already existing) label from the original `<names>` element, which you already have in l. 51, so you'd just have to adjust the label in l. 51. There is no author label, so the line is irrelevant for authors.

harvard-stellenbosch-university.csl
trespaul: The definition to render the editor is in the `book-details` macro, so I can't just rearrange the elements. One would have to create a different `book-details` (without the editor-translator macro in the position as it is) specifically for use in book sections. But I didn't want to simply duplicate the macro with a few changes since that would lead to redundant / duplication of code. What I had in mind was that the `book-details` macro would have to be refactored so that it could be reused by both books and book-sections. (I've looked at it again and now think `book-details` should be split into `edition-volume` and `collection-details` so that `editor-translator` could be placed either in-between them or before both.)

Would it be preferable to open an issue to track this task?

adam3smith: up to you -- having separate macros for editors for chapters/conference proceedings and elsewhere is very common in citation styles. It's not something I'd work on avoiding. If you want to wrap that into this PR, that's great, but if you want to wait for the next iteration, that's fine too. No need for an issue.

harvard-stellenbosch-university.csl
adam3smith: exactly, yes, and that keeps the editor-translator more flexible, e.g.

adam3smith:
Note that I reworked the chapter section of the bibliography a bit -- combining spaces in delimiters with prefixes doesn't work well.

trespaul:
Thanks! I noticed that that was a bit of a mess. I also see in the sample output that some spacing is still wrong (although I didn't see this in my testing setup using pandoc... 🤔) but I'll take a look --- I'm busy parsing the examples on the university library's website so I can use it as a basis for a testing setup.

trespaul:
Oh ok that's good to know, thank you very much!

PR #6465, opened 2023-03-20
Update offa.csl

nmueller18:
Fixed two small quirks concerning the position of the edition number and the volume of a series.

### DIFF
--- a/offa.csl
+++ b/offa.csl
@@ -208,15 +208,10 @@
     </date>
   </macro>
   <macro name="place-date">
+    <group delimiter="" prefix=" ">
     <choose>
       <if type="book thesis chapter paper-conference" match="any">
-        <group delimiter="" prefix=" ">
-          <choose>
-            <if match="none" variable="edition">
-              <text macro="edition"/>
-            </if>
-          </choose>
-        </group>
+        <text macro="edition"/>
         <group delimiter="" prefix="(" suffix=")">
           <group delimiter=" ">
             <choose>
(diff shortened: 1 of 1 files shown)


offa.csl
```diff
@@ -210,12 +210,14 @@
...
-        <choose>
-          <if match="none" variable="edition">
-            <text value=""/>
-          </if>
-        </choose>
-        <group delimiter="" prefix=" (" suffix=")">
+        <group delimiter="" prefix=" ">
+          <choose>
+            <if match="none" variable="edition">
+              <text macro="edition"/>
```
adam3smith, line 216: same here -- what's the goal of this?

adam3smith: I'm talking about l. 215-216, which are from your change (in both styles): you're first checking that edition _isn't_ present, then you're printing the edition macro, which does nothing but render the edition.

nmueller18: Many thanks for the clarification and sorry for the hassle! As I only deal with CSL every few months, I do not really know what I am doing, to be honest. Now I see that this declaration is non-sensical. I suppose
```		
<macro name="place-date">
  <group delimiter="" prefix=" ">
    <choose>
      <if type="book thesis chapter paper-conference" match="any">
         <text macro="edition"/>
           <group delimiter="" prefix="(" suffix=")">
              .
              .
              .
       </group>
      </if>
    </choose>
   </group>
  </macro>
```
would make more sense?

adam3smith: You might actually want a space for the delimiter, all the way at the top, so you get "1st edition (New York)" and not "1st editions(New York)" but you'd want to test. 
But in terms of logic, yes, this is what you'd want.

PR #6466, opened 2023-03-20
Update journal-of-neolithic-archaeology

nmueller18:
Corrected some small quirks with positioning of number of edition and volume of series.

### DIFF
--- a/journal-of-neolithic-archaeology.csl
+++ b/journal-of-neolithic-archaeology.csl
@@ -232,12 +232,12 @@
     <choose>
       <if type="book thesis chapter paper-conference" match="any">
         <group delimiter="" prefix=" ">
-        <choose>
-          <if match="none" variable="edition">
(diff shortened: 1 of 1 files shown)


journal-of-neolithic-archaeology.csl
```diff
@@ -231,12 +231,14 @@
...
-        <choose>
-          <if match="none" variable="edition">
-            <text value=" "/>
-          </if>
-        </choose>
-        <group delimiter=" " prefix=" (" suffix=")">
+        <group delimiter="" prefix=" ">
+          <choose>
+            <if match="none" variable="edition">
+              <text macro="edition"/>
```
adam3smith: This makes no sense -- if you don't have an edition variable, the edition macro is empty. What are you trying to do here?

adam3smith:
Thanks!

PR #6467, opened 2023-03-21
Create iso690-numeric-cat.csl

aniues:

PR #6468, opened 2023-03-21
Create iso690-author-date-cat.csl

aniues:


POBrien333:
relates to #6467

POBrien333:
@bwiernik 
question here. They obviously used a lot of locale overrides as probably back then the catalan locale wasn't in existence or well maintained. Is this ok as is or better should OP remove all unnessary duplications?

bwiernik:
@POBrien333 Can you check against the Catalan locale file and remove the ones that are the same? And could you do the same with the numeric version of the style?

PR #6472, opened 2023-03-22

franpra:
Added ‘<text macro="collection" suffix=", "/>’ in line 420 so that series and series number of conference proceedings can be displayed in the reference, if available.


adam3smith:
Thanks!

PR #6476, opened 2023-03-24

franpra:
Removed page from issue macro. Added page to article-journal and article-magazine display. Added issue to paper-conference display macro. Moved page so that it goes after publisher in  paper-conference display macro. Same thing with chapter display macro. These changes are aimed at displaying collections and volume correctly in book chapters and conference proceedings that are part of series.


adam3smith:
Thanks

PR #6477, opened 2023-03-25

franpra:
Changes made so that output references are aligned with the standard: 1) accessed term capitalization has been deleted in line 237 so that the first letter of the word 'consulta' is displayed lowercase; 2) macro volume has been added in line 209 so that it can be used in reference display of several item type; 3) elements in reference display macros have been rearranged and updated so that they better follow the standard.


adam3smith:
Thanks!

PR #6478, opened 2023-03-25
Update ieee.csl

franpra:
Several updates following the latest official IEEE guide (published August 2022). Collection macro has been added to support displaying series in references. Several updates in display macros have been made accordingly to accommodate series in books, chapters and conference papers.


adam3smith:
Thanks! FWIW, it's generally preferable to handle things like "vol." with a term (i.e. `<text term="volume" form="short"/>` but since the style uses affixes so heavily anyway, it doesn't greatly matter.

adam3smith:
Thanks!

PR #6479, opened 2023-03-27
Create trames.csl

POBrien333:
via private client.

The journal prints it in caps like "TRAMES". Not sure if we do that also.

Waiting for review of the style by the client.


adam3smith:
It's Trames in the ISSN catalog and the DOAJ, so let's go with that.

POBrien333:
This is now reviewed and ready for merging. I'd be grateful if you could merge this one ahead of the other PRs in the list. 😊

adam3smith:
Thanks!

PR #6480, opened 2023-03-28

JoshuaWeberFHNW:
Added article-newspaper to display correct in the bibliography

### DIFF
--- a/hochschule-fur-soziale-arbeit-fhnw.csl
+++ b/hochschule-fur-soziale-arbeit-fhnw.csl
@@ -177,7 +177,7 @@
   <!--Makro Darstellung Jahrgang, Ausgabe und Seitenzahl in Literaturverzeichnis-->
   <macro name="locators">
     <choose>
-      <if type="article article-journal article-magazine chapter article-newspaper">
+      <if type="article article-journal article-magazine chapter article-newspaper" match="any">
         <text variable="volume" prefix=" " suffix=". Jg."/>
         <text variable="issue" prefix=" (" suffix=")."/>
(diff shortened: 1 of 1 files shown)


hochschule-fur-soziale-arbeit-fhnw.csl
```diff
@@ -177,7 +177,7 @@
   <!--Makro Darstellung Jahrgang, Ausgabe und Seitenzahl in Literaturverzeichnis-->
   <macro name="locators">
     <choose>
-      <if type="article article-journal article-magazine chapter">
+      <if type="article article-journal article-magazine chapter article-newspaper">
```
denismaier, line 180: Adding article-newspaper makes sense, yes. But why is chapter in there? Doesn't that produce weird results?

denismaier (changes requested):

hochschule-fur-soziale-arbeit-fhnw.csl
JoshuaWeberFHNW: Thank you for your replies guys. I am not exactly sure what you mean, adam3smith. I don't want that this works with all types, this rule applies only for specific publication forms. Why though match="any"? (I am no expert in CSL, that in advance)

denismaier: `match="all"` -> all conditions have to match, which is impossible. Every item has one type so it can not be of multiple types at the same time. What you want is: match if it's one of those: `article`, `article-journal`, `article-magazine`, `chapter` or `article-newspaper`. (Not all of those at the same time.)

PR #6481, opened 2023-03-29
Update ieee.csl

POBrien333:
revert error with double edition introduced in #6478 and as reported at https://forums.zotero.org/discussion/104003/problem-bei-der-zitation-mit-ieee#latest
* edit "edition" macro
* remove edition macro call in bibliography as it's already called via the "locators" macro

I'm wondering about the line 82 `<text variable="edition" text-case="capitalize-first" suffix="."/>`. Do we want/need a suffix here?

### DIFF
--- a/ieee.csl
+++ b/ieee.csl
@@ -73,7 +73,7 @@
       <if type="bill book chapter graphic legal_case legislation motion_picture paper-conference report song" match="any">
         <choose>
           <if is-numeric="edition">
-            <group delimiter=" " suffix=", ">
+            <group delimiter=" ">
               <number variable="edition" form="ordinal"/>
               <text term="edition" form="short"/>
             </group>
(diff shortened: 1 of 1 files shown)


adam3smith:
> I'm wondering about the line 82 <text variable="edition" text-case="capitalize-first" suffix="."/>. Do we want/need a suffix here?

I don't think so. Where people put non-numeric values into edition, I'd expect those to be what they want verbatim (such as "completely revised edition with a new introduction by that author" or what have you)

adam3smith:
Thanks for the quick fix. That does suggest having an item with an edition in the automated test set would be good. Anything else you'd like?

POBrien333:
Yeah, I was thinking the same.
The Borges example from the CSL editor would be ideal.
Could we add the Campbell/Pedersen citation also? My eye is trained on that and it makes it a bit easier.

adam3smith:
I'll add those two and see if that causes any issues -- might take a little.

PR #6485, opened 2023-03-29
Update to UP Harvard Theology & Religion

schlittenhardtm:
Added support for speeches and interviews.
Error resolved: Deleted comma between container-title and issue in journal article bibliography.


adam3smith:
Thanks!

PR #6496, opened 2023-04-03
Locators in palaeontology journal styles

ms609:
## Input

`[@Smith2020 Fig. 1]`

## Previously:

(Smith 20201)

## Now:

(Smith 2020 figure 1)


ms609:
@adam3smith  Good catch, thanks; a number of the journals don't specify a preference for `page` vs `p.`, but those that do prefer the short form, so I've updated all styles to use this.

adam3smith:
Thanks, this is great

PR #6498, opened 2023-04-03
Update universite-du-quebec-a-montreal-prenoms.csl

jjrondeau:
Same updates as those recently loaded for its template: http://www.zotero.org/styles/universite-du-quebec-a-montreal (added authorship roles, new documents (advanced online publication, prepublication, software, first nation oral knowledge...), etc.

### DIFF
--- a/universite-du-quebec-a-montreal-prenoms.csl
+++ b/universite-du-quebec-a-montreal-prenoms.csl
@@ -1,14 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" et-al-min="3" et-al-use-first="1" initialize="false" demote-non-dropping-particle="never" default-locale="fr-CA">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" and="text" et-al-min="3" et-al-use-first="1" initialize="false" delimiter-precedes-last="never" demote-non-dropping-particle="never" default-locale="fr-CA">
   <info>
     <title>Université du Québec à Montréal - APA - prénoms (Français - Canada)</title>
     <title-short>UQAM prénoms</title-short>
     <id>http://www.zotero.org/styles/universite-du-quebec-a-montreal-prenoms</id>
     <link href="http://www.zotero.org/styles/universite-du-quebec-a-montreal-prenoms" rel="self"/>
     <link href="http://www.zotero.org/styles/universite-du-quebec-a-montreal" rel="template"/>
-    <link href="https://uqam-ca.libguides.com/etudes-feministes/citer-ses-sources" rel="documentation"/>
-    <link href="http://www.guidemt.uqam.ca/citer/style-uqam" rel="documentation"/>
-    <link href="https://uqam-ca.libguides.com/c.php?g=725279&amp;p=5195760" rel="documentation"/>
+	<link href="https://uqam-ca.libguides.com/etudes-feministes/citer-ses-sources" rel="documentation"/>
+	<link href="https://uqam-ca.libguides.com/lgb/zotero" rel="documentation"/>
+    <link href="https://style-apa.uqam.ca/" rel="documentation"/>
     <author>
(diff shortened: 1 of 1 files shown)


universite-du-quebec-a-montreal-prenoms.csl
```diff
@@ -67,91 +67,87 @@
     <choose>
       <if type="article-journal book chapter paper-conference entry-encyclopedia entry-dictionary" match="none">
         <names variable="editor" delimiter=", " prefix=" (" suffix=")">
-          <name and="symbol" initialize="false"/>
+          <name delimiter-precedes-last="never" initialize="false" name-as-sort-order="all"/>
```
adam3smith, line 70: I'm not a huge fan of adding things like `delimiter-precedes-last` into the `style` element, but since you're doing this, why are you adding it back here? Mixing and duplicating where `name` formatting is determined seems maximally confusing

universite-du-quebec-a-montreal-prenoms.csl
```diff
@@ -67,91 +67,87 @@
...
-                <text variable="title" font-style="italic" suffix="."/>
-              </if>
-            </choose>
-          </substitute>
+          <name delimiter-precedes-last="never" name-as-sort-order="all"/>
         </names>
-      </else>
+        <names variable="illustrator"/>
+        <names variable="director" delimiter=", ">
+          <name delimiter-precedes-last="never" name-as-sort-order="all"/>
```
adam3smith, line 85: you can just use 
`<names variable="director"/>` 
The delimiter isn't doing anything if there's on a single variable and the `name` attributes are inherited in substitute.

universite-du-quebec-a-montreal-prenoms.csl
```diff
@@ -67,91 +67,87 @@
...
+          <et-al font-style="italic"/>
+        </names>
+        <names variable="translator"/>
+        <choose>
+          <if type="report">
+            <text variable="publisher"/>
+            <text variable="title" font-style="italic"/>
+          </if>
+          <else-if type="entry-dictionary entry-encyclopedia" match="any">
+            <text variable="title" prefix="« " suffix=" »"/>
```
adam3smith, line 136: please handle quotes with `quotes="true"`

universite-du-quebec-a-montreal-prenoms.csl
```diff
@@ -452,13 +469,13 @@
         </choose>
       </else-if>
       <else-if type="manuscript" match="all" variable="issue">
-        <text variable="issue" prefix="[" suffix="]."/>
+        <text variable="issue" text-case="sentence" prefix="(" suffix=")."/>
```
adam3smith, line 472: what is the goal of sentence case here? That's rarely a good idea to use.

universite-du-quebec-a-montreal-prenoms.csl
```diff
@@ -849,55 +901,48 @@
...
-          </if>
-          <else>
-            <text variable="genre" prefix=" [" suffix="]"/>
-          </else>
-        </choose>
-      </else-if>
     </choose>
   </macro>
-  <citation et-al-min="3" et-al-use-first="1" et-al-subsequent-min="3" et-al-subsequent-use-first="1" disambiguate-add-year-suffix="true" collapse="year">
+  <citation name-form="short" delimiter-precedes-last="never" et-al-min="3" et-al-use-first="1" et-al-subsequent-min="3" et-al-subsequent-use-first="1" disambiguate-add-year-suffix="true" collapse="year">
```
adam3smith, line 940: no need to specify separate `et-al-subsequent` if it's the same as the general `et-al`

adam3smith (changes requested):
Some comments -- some refer to the old style, which is unnecessarily convoluted -- but you also made things even more so in some case and it'll help long-term maintenance if you clean this up a bit along the lines suggested.

POBrien333:
Continued in #https://github.com/citation-style-language/styles/pull/7719

POBrien333:
@jjrondeau 
I continued your work and my PR was just merged. The style should update within an hour.

PR #6505, opened 2023-04-13
Update universite-du-quebec-a-montreal-etudes-litteraires-et-semiologie.csl

jjrondeau:
Updated to be compliant with the December 2022 Presentation Guide update. The notes are now complete and several new document types have been added.


adam3smith:
Thanks!

PR #6507, opened 2023-04-17
Propose empty.csl

floe:
Completely empty CSL style (for hiding references, e.g. when rendering Quarto to HTML). Couldn't find anything suitable for that purpose, so I created an empty style myself.

### DIFF
--- a/empty.csl
+++ b/empty.csl
@@ -1,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>Completely empty CSL style (for hiding references, e.g. when rendering Quarto to HTML)</title>
+    <id>http://www.zotero.org/styles/empty</id>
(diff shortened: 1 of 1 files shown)


bwiernik:
Could you explain the use case for this?

adam3smith:
> Not sure the official styles repo is the place for such of hacks.

Yes, agree with that. Thanks for providing the Lua method for this

floe:
@denismaier, thanks for the suggestion, that would actually work as well. I submitted this in the first place because I was thinking there may be other usecases where citations should still be processed, but not rendered, but that is of course up to your discretion.

adam3smith:
Another reason not to take this is that e.g. citeproc-js (Zotero, Mendeley) doesn't like empty citations (because where they're not intentional, they can cause all sorts of issues) and replaces them with an all caps warning), so I think we'll keep this off the repository. The file will remain accessible as part of the PR, though, and you should feel free to link people to it.

adam3smith:
Oh, and, of course, thank you for contributing, even though this one doesn't end up in the repository.

adam3smith:
Oh, and, of course, thank you for contributing, even though this one doesn't end up in the repository.

PR #6510, opened 2023-04-18
Update universite-de-bordeaux-ecole-doctorale-de-droit.csl

fflamerie:
* author email updated
* CSL v1.0.2 : use of the new "op-cit" term instead of a text value
* CSL v1.0.2 : use of the "contributor" role for thesis directors ("thesis-director" macro) and preface writers ("preface-author" macro)


adam3smith:
Thanks!

PR #6511, opened 2023-04-18
Update ieee.csl

POBrien333:
https://forums.zotero.org/discussion/104443/conference-papers-presentations-in-ieee-style-doubling-place#latest


POBrien333:
I think we should also copy this over to the ieee-with-urls style. There were a few changes recently. Also, not 100% sure but I feel like remembering that the styles were not just copy/paste styles, but had a different "lineage" so to speak.

adam3smith:
Yes, I agree we want to carry these over to the other IEEE. I thought the two were aligned, but not sure

PR #6512, opened 2023-04-22
Update environmental-and-engineering-geoscience.csl

POBrien333:
https://forums.zotero.org/discussion/104502/style-request-environmental-and-engineering-geoscience#latest

Guidelines: https://aeg.memberclicks.net/assets/docs/Publications/EEG-Author-Instructions.pdf
example article: https://pubs.geoscienceworld.org/aeg/eeg/article/28/3/237/615981/Land-Subsidence-Due-to-Creep-of-the-Gulf-Coast


adam3smith:
I don't think these are small-caps -- I think the names are to `text-case="uppercase"` or am I missing something?

POBrien333:
The .pdf paper looks like small caps to me?

adam3smith:
Ah, hadn't looked at that, thanks (you can tell for sure if something is small caps by copying to a plain text editor).

PR #6514, opened 2023-04-26
Update ieee-with-url.csl

POBrien333:
As discussed in #6511 

- copy ieee.csl
- replace access macro with the one from ieee-with-urls.csl

Let me know if I did this right. :)

### DIFF
--- a/ieee-with-url.csl
+++ b/ieee-with-url.csl
@@ -198,7 +198,7 @@
   <macro name="access">
     <choose>
       <if variable="URL">
-        <group delimiter=". ">
+        <group delimiter=". " prefix=" ">
           <group delimiter=": ">
             <text term="available at" text-case="capitalize-first"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
The missing space before "Available" is an actual issue, not just in the preview (I think you can just add that to the macro)
A bit more tricky: journal articles without URL currently end in a comma.

adam3smith:
That doesn't fix this one, though, does it?
> A bit more tricky: journal articles without URL currently end in a comma.

POBrien333:
It came out fine. See the conditional in lines 300-307

adam3smith:
Ah -- we've lost the DOI somewhere along the way. That should definitely still be part of access, and then the code works. (without the DOI in access, items with a DOI but no URL -- what I was testing with -- still end in a comma)

adam3smith:
Thanks!

PR #6515, opened 2023-04-26
Update norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl

StianOby:
Fixed an error in the handling of the year for journal articles in the bibliography.


adam3smith:
Thanks!

PR #6518, opened 2023-04-28
HEG sytle update

grolimur:
updates based on the latest version of the ISO 690 standard, with improvements

### DIFF
--- a/environmental-and-engineering-geoscience.csl
+++ b/environmental-and-engineering-geoscience.csl
@@ -2,9 +2,10 @@
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
   <info>
     <title>Environmental &amp; Engineering Geoscience</title>
+    <title-short>EEG</title-short>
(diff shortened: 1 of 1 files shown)


iso690-author-date-en.csl
```diff
@@ -118,7 +118,7 @@
       <if type="book thesis map motion_picture song manuscript" match="any">
         <choose>
           <if variable="author editor translator" match="any">
-            <text variable="title" font-style="italic"/>
+            <text variable="title" font-style="italic" suffix=". "/>
```
adam3smith, line 121: Are you making the changes to all the other ISO-690 styles on purpose? (the PR title only mentions HEG). And if so, what's the idea of adding all the periods in the title suffix here? The title macro has period in the suffix every time it's called in the bibliography, so this shouldn't be necessary (and makes the style harder to maintain)

grolimur: Thanks @adam3smith. You are right. I removed these periods in the title macro.

adam3smith:
did you close this on purpose?

grolimur:
No, sorry. I closed it accidentslly when I answered your question.

grolimur:
@adam3smith Should I reopen it in order for this pull request be to taken in account in the next Zotero update?

PR #6522, opened 2023-05-03
Create gost-r-7-0-5-2008-numeric-by-appearance.csl

e-khurmamatov:
Based on [GOST R 7.0.5-2008 (numeric, alphabetical) style](https://github.com/citation-style-language/styles/blob/master/gost-r-7-0-5-2008-numeric-alphabetical.csl). 

This new style is necessary for universities in Russia that require citations to be ordered by appearance in the text.

### DIFF
--- a/gost-r-7-0-5-2008-numeric-by-appearance.csl
+++ b/gost-r-7-0-5-2008-numeric-by-appearance.csl
@@ -1,0 +1,173 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" class="in-text" default-locale="ru-RU">
+  <info>
+    <title>Russian GOST R 7.0.5-2008 (numeric, sorted by appearance, Ру́сский)</title>
+    <title-short>GOST R 7.0.5-2008 (numeric, sorted by appearance)</title-short>
+    <id>http://www.zotero.org/styles/gost-r-7-0-5-2008-numeric-by-appearance</id>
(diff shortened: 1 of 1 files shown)


e-khurmamatov:
Based on my recent findings, `gost-r-7-0-5-2008-numeric.csl` appears to perform even better and fulfill the same purpose as the previously suggested option. 

My initial hesitation in considering this style was due to the assumption that it was only available in English, given that the locale field in Zotero was grayed out and set to English (US) by default. However, upon further inspection, I discovered that the style actually includes a Russian translation within its code.

If you would like to change the language of `gost-r-7-0-5-2008-numeric.csl`, you can do so by modifying the `default-locale` field within the code itself. You can set it to either `ru-RU` to use the Russian translation or leave it empty to allow for the selection of language from the Zotero interface. I hope this additional information is helpful.

I will create an issue specifically for `gost-r-7-0-5-2008-numeric.csl` about language change support. As a result, I close this pull request.

e-khurmamatov:
Based on my recent findings, `gost-r-7-0-5-2008-numeric.csl` appears to perform even better and fulfill the same purpose as the previously suggested option. 

My initial hesitation in considering this style was due to the assumption that it was only available in English, given that the locale field in Zotero was grayed out and set to English (US) by default. However, upon further inspection, I discovered that the style actually includes a Russian translation within its code.

If you would like to change the language of `gost-r-7-0-5-2008-numeric.csl`, you can do so by modifying the `default-locale` field within the code itself. You can set it to either `ru-RU` to use the Russian translation or leave it empty to allow for the selection of language from the Zotero interface. I hope this additional information is helpful.

I will create an issue specifically for `gost-r-7-0-5-2008-numeric.csl` about language change support. As a result, I close this pull request.

PR #6527, opened 2023-05-05
Update bibliothek-forschung-und-praxis.csl

jmiba:
Disambiguate no-date publications by one author

### DIFF
--- a/bibliothek-forschung-und-praxis.csl
+++ b/bibliothek-forschung-und-praxis.csl
@@ -433,7 +433,7 @@
       </else-if>
       <else>
         <text term="no date" form="short"/>
-                <text variable="year-suffix" prefix="-"/>
+        <text variable="year-suffix" prefix="-"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
you really want the title there? I'd be more common to use a year suffix like "o. J.-a" in an author-date style (you'd have to set that explicitly, I believe)

jmiba:
> you really want the title there? I'd be more common to use a year suffix like "o. J.-a" in an author-date style (you'd have to set that explicitly, I believe)

Actually, a disambiguated without-date-term would be preferable but I believe this is not possible. If you put a term here it will not be disambiguated in case there are several documents by the same author without dates.

adam3smith:
Could you test this version? Should work

jmiba:
Thanks, @adam3smith, this is perfect!

adam3smith:
Thanks!

PR #6540, opened 2023-05-11
Update turabian-fullnote-bibliography.csl

### LINKED ISSUE #3496: Turabian 9th Edition
Has apparently just come out and is updated to conform to Chicago 17th. Would be good to update the style -- maybe just base it on CMoS and adjust as needed. 

If someone with a track record of submitting styles wants to take this on, we can buy you a copy of the manual.

POBrien333:
Closes #3496   

An initial draft of Turabian9, based on https://issuu.com/sbtswritingcenter/docs/torey_turabian9_blog.docx  

I have no access to the manual, so do not know if this correct or correctly implemented.

* remove ibid
* add accessed date for webpages without issued date


adam3smith:
Thanks -- looking at the manual, I think we can just make this a true dependent of Chicago: 

> A citation style used widely in the humanities and in some social sciences is the _notes bibliography_ or _notes style_ for short (also known as Turabian or Chicago notes-bibliography or notes style).

I've done a spot check and it looks exactly the same, just with a lot fewer details, so I think we should go with the dependent option. That'll also make our lives easier if people ask for an ibid. version, which is discouraged but still described.

POBrien333:
Great, thank you for the review. That's made into a dependent now.
Let's see how it does in the real world. :)

What's the history of Turabian anyway?! Never got why there was a nearly identical style. (not that all other styles are all nearly identical of each other... haha)

adam3smith:
I'll fix the errors

POBrien333:
Thanks :)

PR #6542, opened 2023-05-11
Create journal-of-the-american-philosophical-association.csl

POBrien333:
https://forums.zotero.org/discussion/104823/style-request-american-philosophical-association#latest


adam3smith:
Thanks (made the last requested change, see above)

POBrien333:
Thanks a lot!
(Just moves flat today, so appreciate you making the fix!)

PR #6543, opened 2023-05-12

POBrien333:
New style based on Springer.

via https://forums.zotero.org/discussion/104950/style-update-edit-help-request#latest


adam3smith:
Looks great, thanks! (Even without explicit guidance, I always try to make sure that no-author items have reasonable in-text citations in author-date styles. Just (2012) for a webpage without author is definitely not right.

PR #6546, opened 2023-05-14
Update springer-vancouver-brackets.csl

POBrien333:
See issue 2 on this forum post: https://forums.zotero.org/discussion/104977/academic-psychiatry-style-changes-needed#latest

**Add conditional so publisher information doesn't get printed for article-journal etc.**
I might am forgetting a book-like item type.
Or should we invert the conditional?

Guidelines: https://resource-cms.springernature.com/springer-cms/rest/v1/content/3322/data/Key+style+points

### DIFF
--- a/springer-vancouver-brackets.csl
+++ b/springer-vancouver-brackets.csl
@@ -70,7 +70,7 @@
   </macro>
   <macro name="publisher">
     <choose>
-      <if type="thesis book chapter paper-conference" match="any">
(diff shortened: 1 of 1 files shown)


adam3smith:
yeah, I think I'd want to just exclude from `article-journal/newspaper/magazine` -- thinks like report, article, dataset should probably have a publisher.

adam3smith:
Thanks! See above for what I had in mind

POBrien333:
ah, didn't ready properly. sorry and thanks for fixing. makes sense and I'll keep that in mind for future styles.

PR #6553, opened 2023-05-17
Fix erroneous comma in ACL in-text citation

Tai-Mai:
When doing an in-text citation, for example
"proposed by `Author` (`year`)",
there is an erroneous comma before the year, i.e.
"proposed by `Author` (, `year`)".
This is because the comma is set as a prefix to the year. Changing it to a suffix to the author fixes this problem.


adam3smith:
Thanks! (Generally better to use group delimiter rather than affixes whenever possible)

PR #6554, opened 2023-05-17
Fix et al and name settings in universitat bern institut fur sozialanthropologie

denismaier:
The [Style Guide](https://www.anthro.unibe.ch/unibe/portal/fak_historisch/dkk/anthro/content/e40416/e1010183/e1015537/files1015544/zitieren_ger_2020_ger.pdf) states:

Authors in reference list: 
> Alle Personen aufführen, und zwar in der Reihenfolge, in welcher sie im Buch/in der Zeitschrift stehen. Nur bei der ersten Person steht der Nachname vor dem Vornamen.

Authors in citations:
> Ab drei Personen: nur die Erste, die Übrigen mit et al. abkürzen: (Schweizer et al. 1999: 98)


POBrien333:
Looks good to me. In line 65 you also have an `and="text"` which is technically superfluous since you set it globally, but it doesn't really make a difference.

denismaier:
Oh yeah, that was in there already and I didn't bother to delete it.

denismaier:
Ok, I've now removed all instances of `and="text"` apart from the one on `cs:style`. Also, I don't see why we should leave `and="symbol"` in.

adam3smith:
Thanks!

PR #6556, opened 2023-05-17
Update mots.csl

Oxil-2:
Minor correction


adam3smith:
What's the goal here? -- the category doesn't change any style behavior and (as per the automated tests) all citation styles that use footnotes should use `citation-format="note"` (regardless of whether in the note they might be author-date)

Oxil-2:
Sorry, got confused between two branches, I'm not familiar with how Github works. So I just want to switch this style from the "note" format to the "author-date" / "in-text" format. I've made two changes to that end ("class" attribute + "citation-format" attribute).

adam3smith:
Thanks!

PR #6558, opened 2023-05-17
Create novum-testamentum.csl

dstark:
Add CSL style for the journal Novum Testamentum.


adam3smith:
Yes, this looks good -- review will take a little

PR #6559, opened 2023-05-18
FitopatologiaUFRPE.csl

bleomendes1:
Normas para confecção de dissertações e teses do Programa de Pós-graduação em Fitopatologia da UFRPE Brasil


adam3smith:
Please fix any issues here in this pull request as per the instructions above -- we're happy to help if you run into trouble, but you can't keep creating new pull requests -- that becomes a mess to deal with for us.

PR #6560, opened 2023-05-25
Update haute-ecole-de-gestion-de-geneve-iso-690.csl

grolimur:
removed "archive" field from bibliography

### DIFF
--- a/haute-ecole-de-gestion-de-geneve-iso-690.csl
+++ b/haute-ecole-de-gestion-de-geneve-iso-690.csl
@@ -415,8 +415,6 @@
               <text macro="title" suffix=". "/>
             </if>
           </choose>
-          <text variable="version" prefix="Version " suffix=". "/>
-          <!-- used for software until type="software" is implemented -->
           <text macro="edition" suffix=". "/>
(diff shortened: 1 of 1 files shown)


haute-ecole-de-gestion-de-geneve-iso-690.csl
```diff
@@ -356,15 +415,14 @@
               <text macro="title" suffix=". "/>
             </if>
           </choose>
-          <text variable="version" prefix="version " suffix=". "/>
+          <text variable="version" prefix="Version " suffix=". "/>
```
adam3smith, line 418: `version` now is a term you can use. Also, do you still need this now that `software` is available as an item type & mapped from Zotero?

adam3smith (commented):
One quick question above

haute-ecole-de-gestion-de-geneve-iso-690.csl
grolimur: With `software` available as a doctype in CSL, this line is not useful anymore. I will remove it.

adam3smith:
Thanks!

PR #6562, opened 2023-05-26

Braintelligence:
Change of representation of editors in books if only editors are provided


adam3smith:
Thanks!

PR #6563, opened 2023-05-28
Update geographie-et-cultures.csl

POBrien333:
rework style

https://forums.zotero.org/discussion/105153/style-error-geographie-et-cultures#latest


adam3smith:
Thanks! Looks like this is missing quotation marks/guillemets for articles & chapters -- also, I don't think we should do et al/author-use-last for the bibliography

adam3smith:
Thanks!

PR #6569, opened 2023-06-07

LPDagallier:
For inline citations with same author and same year, set the disambiguation suffixes (a, b, c, etc.) delimited with ", " (instead of "; " that was apparently the default when not specified). Example: (AuthorA et al. 2019; AuthorB et al., 2021a, b) Instead of previously: (AuthorA et al. 2019; AuthorB et al., 2021a; b)

### DIFF
--- a/american-journal-of-botany.csl
+++ b/american-journal-of-botany.csl
@@ -109,7 +109,7 @@
       </else>
     </choose>
   </macro>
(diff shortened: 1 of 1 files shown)


adam3smith:
Weird, this shouldn't be necessary, but I'm seeing the same behavior, so taking this.

PR #6570, opened 2023-06-09
Create Australian Archaeology (2023 Update).csl

CameronMcVicar:
Referencing Style based on the Australian Archaeological Association referencing system


adam3smith:
Thanks!
Could you say more about what you're trying to do? If the existing Australian Archaeology style is either incorrect and/or outdated, we'd want changes directly to that file/style. If this is something else, could you explain?

CameronMcVicar:
Hi,

The existing Australian Archaeology style available on the repository is incorrect in several ways. Mostly with the in-text references, for example, although it is author-date it requires a page number separated by a colon without a space and no page label (ie. p. Or pp.). So should appear as (Evans and Taylor 1999:56). I have updated these elements from the old style using the existing code. I then updated some of bibliography elements. Such as, no "editor" label after the authors in edited books; and, separating publishing-location and publisher by a colon not a comma. I think this is everything I've changed, the rest of the code remains the same.

Get Outlook for iOS<https://aka.ms/o0ukef>

POBrien333:
This looks like an old PR and the style has received updates in 2024 by another user in https://github.com/citation-style-language/styles/pull/7115
Closing this.

PR #6571, opened 2023-06-10
Update bmj.csl - citation layout

badgettrg:
vertical-align="sup" and suffix/prefix empty

### DIFF
--- a/bmj.csl
+++ b/bmj.csl
@@ -141,7 +141,7 @@
     <sort>
       <key variable="citation-number"/>
     </sort>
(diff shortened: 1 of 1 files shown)


adam3smith:
BMJ publishes with superscript, but the author guidelines very explicitly call for numbers in square brackets, not superscripted, so the current style is correct. See https://authors.bmj.com/writing-and-formatting/formatting-your-paper/#:~:text=BMJ%20formats%20references%20using%20Vancouver,'et%20al'%20if%20applicable.

PR #6572, opened 2023-06-12
Update the-quarterly-journal-of-economics.csl

Austin-Zheng:
- In the bibliography, for entries with multiple authors, have last name before first name for the first author only, and first name before last name for all other authors.

- In the bibliography, for types other than book, chapter, or article variants, do not enclose the year in parentheses.


adam3smith:
Thanks! (We'd also probably want to improve citations to preprints & working papers -- `article` in CSL -- for this)

PR #6573, opened 2023-06-13
chempublish-journal.csl

nurprat:


POBrien333:
I found in the journals' guidelines that they recommend a citation style to use, JMANS.
That has been added and now this is a dependent style on said journal's style.

POBrien333:
Closed with #7740

PR #6577, opened 2023-06-23
Update publicatiewijzer-voor-de-archeologie.csl

RonaldVisser:
Corrections of minor errors. 
- references to publications with two authors (no / but an 'en')
- et al. was shortened in Dutch to e.a.

### DIFF
--- a/publicatiewijzer-voor-de-archeologie.csl
+++ b/publicatiewijzer-voor-de-archeologie.csl
@@ -69,7 +69,7 @@
     </names>
   </macro>
   <macro name="author-short">
-    <names variable="author" delimiter="/">
+    <names variable="author">
       <name form="short" delimiter="/">
         <name-part name="family" text-case="capitalize-first"/>
       </name>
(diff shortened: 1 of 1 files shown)


publicatiewijzer-voor-de-archeologie.csl
```diff
@@ -64,8 +69,8 @@
     </names>
   </macro>
   <macro name="author-short">
-    <names variable="author">
-      <name form="short" and="text">
+    <names variable="author" delimiter="/">
```
adam3smith, line 72: I'll fix this here, but for future reference: the delimiter on `names` is the delimiter between different author types (e.g. if you have `<name variable="author editor">`). It doesn't do anything if only a single creator type is given, so no purpose for it here. The one you want is in `name`

adam3smith:
Thanks!

publicatiewijzer-voor-de-archeologie.csl
RonaldVisser: Thanks for correcting and explaining! And I'm sorry for the confusion

PR #6581, opened 2023-06-26
Update al-jamiah-journal-of-islamic-studies.csl

tonidjogja:
This style initially use double quote for journal article and the likes, so it needs to change the default language to en-US, instead of en-GB


adam3smith:
Thanks!

PR #6582, opened 2023-06-27
Create apa-numeric-no-superscript-brackets.csl

josmarcristello:
There was no APA with numbers in bracket that were not superscript (i.e. normal text). This rectifies that.

### DIFF
--- a/apa-numeric-no-superscript-brackets.csl
+++ b/apa-numeric-no-superscript-brackets.csl
@@ -236,76 +236,6 @@
       </group>
     </group>
   </macro>
-  <macro name="date-sort-group">
-    <choose>
-      <if variable="issued">
-        <text value="1"/>
-      </if>
-      <else-if variable="status">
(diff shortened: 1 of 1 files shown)


josmarcristello:
Hi, this is a variation of APA's 7th Edition that uses in-text references with brackets in regular type case, as opposed to uppercase as the only other variation that I could find in this repository. This .csl is what we're using in my laboratory for publications with the International Journal of Hydrogen Energy ([Example](https://www.sciencedirect.com/science/article/abs/pii/S0360319923002951)) - although publications don't necessary need to follow this strict format.

adam3smith:
Thanks. We're not going to take this for the repository in that case. We already have a style for that journal that is closer to their published format https://www.zotero.org/styles?q=International%20Journal%20of%20Hydrogen%20Energy and as I say above, try to limit the proliferation of APA variants unless they're needed to meet a specific requirement.

PR #6583, opened 2023-06-28
update bluebook-law-review.csl

POBrien333:
see https://forums.zotero.org/discussion/105567/bluebook-capitalisation-unreported-cases

### DIFF
--- a/bluebook-law-review.csl
+++ b/bluebook-law-review.csl
@@ -19,7 +19,7 @@
     <category citation-format="note"/>
     <category field="law"/>
     <summary>The Bluebook legal citation style for law reviews.</summary>
-    <updated>2022-03-09T09:33:23+00:00</updated>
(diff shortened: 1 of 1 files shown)


adam3smith:
I would check if you can add the institutions to reports here and if you want to work on unreported cases, do that in a different pull request. Might be worth checking in APA or Chicago Manual -- both of which follow Bluebook for case citations and have some case citation code nicely contained in separate macros, so might be possible to just copy over

adam3smith:
OK, so not changing reports, let's do cases separately (again, if you want to to into that at all)

adam3smith:
Thanks!

POBrien333:
Ok, will see if I have the time and will for the other issues. Stuff I don't fully understand and is hard to decipher doesn't necessarily motivate me. Ha

PR #6589, opened 2023-07-01

OTFlorian:
Use long title in subsequent citations since the style documentation does not allow using shortened titles.


adam3smith:
Thanks!

PR #6591, opened 2023-07-02
Update studii-teologice.csl

nturcan:
Added bibliography section.


adam3smith:
Thanks -- for future reference, if you port code from the citation to the bibliography, make sure it doesn't have locators in it -- those only make sense in the context of citations.

PR #6270, opened 2022-10-12
Partial update of APA styles for 1.0.2, including software, legal, most localizations

bwiernik:

PR #6596, opened 2023-07-11
Update apa csl to use software type 

rbran100:
https://docs.citationstyles.org/en/stable/specification.html#type-terms

The 'software' type is now available and it seems the apa style should use that, this fixes a bug where "[Computer Software]" is added for books when it should be for only software.

### DIFF
--- a/apa-annotated-bibliography.csl
+++ b/apa-annotated-bibliography.csl
@@ -1330,8 +1330,8 @@
       <else-if type="dataset">
         <text value="Data set"/>
       </else-if>
-      <else-if type="book" variable="version" match="all">
-        <!-- Replace with type="software" and term="software" as that becomes available -->
+      <else-if type="software" variable="version" match="all">
+        <!-- Replace term="software" as that becomes available -->
         <text value="Computer software"/>
       </else-if>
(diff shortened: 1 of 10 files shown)


adam3smith:
@bwiernik -- where are you with your APA updates? I don't want to take incremental PRs if they're going to mess with your larger update, but otoh also don't want to delay incremental improvements unnecessarily.

@rbran100 thanks for the PR. Let me clarify first if we're going to take this in general. If so, we'd want to fix a couple of things still.

bwiernik:
I've got this change implemented already--I'll merge a partial PR from the full update on Saturday

adam3smith:
Thanks for the quick reply -- closing here then, with again thanks to @rbran100: we'll add the equivalent of this PR as part of the larger set of APA updates here: https://github.com/citation-style-language/styles/pull/6270

(but your books really shouldn't have version numbers, and that's the only case in which you'd see "computer program" for a book)

bwiernik:
And thanks for the kick in the pants to get this update done @rbran100 !

bwiernik:
This (and a few other) portions of the 1.0.2 APA update are merged https://github.com/citation-style-language/styles/pull/6270

Issue #6598, opened 2023-07-14
ACM dropped prefix "DOI:"


fgnievinski:
ACM recently updated their Citation Style and Reference Formats webpage:
https://www.acm.org/publications/authors/reference-formatting

They've simplified the formatting of DOIs; previously it was, e.g.:
   DOI:https://doi.org/10.1145/1188913.1188915
Now it's just:
   https://doi.org/10.1145/1188913.1188915

I've confirmed it via email at production@hq.acm.org

I've updated only the two ACM SIG CSLs.
However, there are a lot of other "ACM" CSLs.

I've focused on the ones used by default in PKP/OJS, e.g.:
https://demo.publicknowledgeproject.org/ojs3/demo/index.php/demojournal/article/view/6
 
Previous related issue:
* #1996

POBrien333:
Closed with merge of [#6599](https://github.com/citation-style-language/styles/pull/6599) and [#6600](https://github.com/citation-style-language/styles/pull/6600)

PR #6600, opened 2023-07-14
Update acm-sig-proceedings.csl

fgnievinski:
as per issue #6598


adam3smith:
Thanks, but the period +space needs to go back into the prefix -- check the Fenner et al. sample citation (it also really should make a difference in the visual editor, not sure why it didn't for you). Could you update the two PRs, please?

POBrien333:
Continued in https://github.com/citation-style-language/styles/pull/7671

PR #6602, opened 2023-07-16
fix film/video authors for american-sociological-association.csl

POBrien333:
https://forums.zotero.org/discussion/106241/citing-video-recordings-in-asa#latest


bwiernik:
Thanks!

PR #6607, opened 2023-07-23
Update the-journal-of-clinical-investigation.csl

POBrien333:
- Fix various smaller issues in JCI
- make JCI Insight dependent

https://forums.zotero.org/discussion/106398/style-error-the-journal-of-clinical-investigation#latest


adam3smith:
Thanks! I don't think we want the [Internet] that' currently in the Fenner et al. citations example ever, do we?

adam3smith:
Thanks!

PR #6623, opened 2023-08-09
Create instituto-de-investigaciones-juridicas-unam.csl

nanounanue:


nanounanue:
This is the CSL required for some law journals in Mexico and LatAM

PR #6627, opened 2023-08-10
Create comptes-rendus-mecanique.csl

justinefabre:
This new style cancels and replaces the old dependent/comptes-rendus-mecanique.csl style, as our journal has left its previous publisher Elsevier and now has its own style. The old style can be removed from the repository.


justinefabre:
I corrected the "self-link" and "et-al-min" issues, but I am not sure about what to do with the two others issues.
This is not really an update of an existing dependent style, so I am not sure I must update the Elsevier style instead of posting this new one...

adam3smith:
superseded by https://github.com/citation-style-language/styles/pull/6685

PR #6632, opened 2023-08-10
Create comptes-rendus-physique.csl

justinefabre:
This new style cancels and replaces the old dependent/comptes-rendus-physique.csl style, as our journal has left its previous publisher Elsevier and now has its own style. The old style can be removed from the repository.


adam3smith:
superseded by https://github.com/citation-style-language/styles/pull/6685

PR #6635, opened 2023-08-13
Update berghahn-books-author-date-en-gb.csl

POBrien333:
https://forums.zotero.org/discussion/106865/berghahn-books-style#latest


adam3smith:
Thanks!

PR #6636, opened 2023-08-15
Update american-institute-of-aeronautics-and-astronautics.csl

POBrien333:
Major reworking of style. Probably missed a few things.

via https://forums.zotero.org/discussion/106919/zotero-aiaa-journal-style-fails-to-include-doi-numbers#latest

Guidelines: https://www.aiaa.org/publications/journals/reference-style-and-format


adam3smith:
Looks good -- for the editors in the chapter citations, where do you get the format from? [12] suggests it should be ", edited by P. A. Hall and D. Soskice"

POBrien333:
Good spot. Made the necessary edits.

adam3smith:
Thanks!

PR #6641, opened 2023-08-17
Closes #6576 Add support for using the forthcoming status.

### LINKED ISSUE #6576: Add "forthcoming" as a supported status to the blue book style.
I am not a citation expert.

When looking at the bluebook citation style documentation: https://www.legalbluebook.com/bluebook/v21/rules/17-unpublished-and-forthcoming-sources/17-3-forthcoming-publications

I see forthcoming as an option, but when searching this repository for "forthcoming" there is no result for bluebook.
https://github.com/search?q=repo%3Acitation-style-language%2Fstyles%20forthcoming&type=code

Also, I looked in the designated bluebook csl xml files and did not see forthcoming mentioned.
https://github.com/citation-style-language/styles/blob/master/bluebook-law-review.csl
https://github.com/citation-style-language/styles/blob/master/bluebook-inline.csl

Question: is this something that could be added or is it irrelevant?

trackleft:
Added additional types to the issuance  macro 

- book
- chapter

Then when crafting the issue date, I check the status variable, and place that as a prefix for the date.
```

                      <if variable="status">
                        <group prefix="" suffix=" ">
                          <text variable="status"/>
                        </group>
                      </if>

```

Other statuses might not make sense here, but I'm not sure how to check if the variable is forthcoming or retracted or other statuses.

### DIFF
--- a/bluebook-law-review.csl
+++ b/bluebook-law-review.csl
@@ -181,12 +181,21 @@
   </macro>
   <macro name="issuance">
     <choose>
-      <if type="article-journal article-magazine article-newspaper book broadcast chapter interview manuscript map patent personal_communication post-weblog song speech thesis webpage" match="any">
+      <!-- Check if the reference is marked as "forthcoming" -->
+      <if variable="status" match="forthcoming">
         <group>
-          <choose>
-            <if type="article-newspaper book chapter post-weblog thesis" match="any">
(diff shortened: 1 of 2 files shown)


adam3smith:
Thanks -- what is "redacted" in this context btw? A quick google doesn't show anything and I've never seen this in a article citation.

PR #6655, opened 2023-08-26
Create warta-pengabdian-andalas.csl

purnawanpp:


POBrien333:
@purnawanpp 
On your website you state that your journal follows APA.
If that is the case, we can make this style a dependent style instead of the custom style you currently have. Please confirm.

PR #6648, opened 2023-08-25
Update polish-archives-of-internal-medicine.csl

PolArchInternMed:
We need a hyphen instead of en dash in the range of pages. Please let me know, if I changed it correctly. Thank you!

PR #6657, opened 2023-08-28
Update polish-archives-of-internal-medicine.csl

PolArchInternMed:
Added: https://github.com/citation-style-language/styles/pull/6648#issuecomment-1693978626


PolArchInternMed:
Is there anything I need to do next? Or will it now be updated on Zotero etc?

adam3smith:
Thanks! (This will update automatically at Zotero & elsewhere)

PolArchInternMed:
Thank you

PR #6661, opened 2023-08-28
Update bibliotheque-universitaire-de-medecine-vancouver.csl

alerac:
* Style now available in both french and english language

### DIFF
--- a/bibliotheque-universitaire-de-medecine-vancouver.csl
+++ b/bibliotheque-universitaire-de-medecine-vancouver.csl
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="minimal" default-locale="fr-FR">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="minimal">
   <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
   <info>
-    <title>Bibliothèque universitaire de médecine - Vancouver (Français)</title>
+    <title>Bibliothèque universitaire de médecine - Vancouver</title>
     <title-short>BiUM Vancouver</title-short>
     <id>http://www.zotero.org/styles/bibliotheque-universitaire-de-medecine-vancouver</id>
     <link href="http://www.zotero.org/styles/bibliotheque-universitaire-de-medecine-vancouver" rel="self"/>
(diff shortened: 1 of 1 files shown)


alerac:
Hi,

Just a quick follow up to ask if there is maybe a problem with this pull request ? We started the new academic year on 19 september and some students are beginning to write their theses. It would be great for them to have the opportunity to write either in French or in English.

Please let me know if I have to correct something.

Thanks.

bibliotheque-universitaire-de-medecine-vancouver.csl
```diff
@@ -46,6 +46,31 @@
       </term>
     </terms>
   </locale>
+  <locale xml:lang="en">
```
adam3smith, line 49: Several of these (like `in`, `cited`, and `editor`) are already in the English locales, please remove them here.

bibliotheque-universitaire-de-medecine-vancouver.csl
```diff
@@ -46,6 +46,31 @@
...
+      <term name="internet">Internet</term>
+      <term name="collection-editor" form="long">
+        <single>editor</single>
+        <multiple>editors</multiple>
+      </term>
+      <term name="editor">
+        <single>editor</single>
+        <multiple>editors</multiple>
+      </term>
+      <term name="original-author">
```
adam3smith, line 68: This seems left over from the French? And/or should it be `inventor`?

adam3smith (changes requested):
Sorry for the delay. Just a couple of quick things, rest looks good

bibliotheque-universitaire-de-medecine-vancouver.csl
alerac: Hi,

No problems, I know there are a lot of styles to review ;-) Thanks for this review.

I adressed the requested changes. The ICMJE asks that _"References should follow the standards summarized in the NLM's Sample References (www.nlm.nih.gov/bsd/uniform_requirements.html) webpage and detailed in the NLM's Citing Medicine, 2nd edition (www.ncbi.nlm.nih.gov/books/NBK7256/)."_ https://www.icmje.org/icmje-recommendations.pdf (p.18 ; ii. Style and Format)

Both references stand that the label for `retrieve` term should be _"Available from"_. e.g. here https://www.nlm.nih.gov/bsd/uniform_requirements.html#item36 Is it ok if I left the label as _"Available from"_ in the style instead of using the default english locale label _"Retrieved"_ ?

I removed the others as requested.

Thanks again!

bibliotheque-universitaire-de-medecine-vancouver.csl
alerac: I removed it in both languages. It was unnecessary for this style and probably a left over from the template style.

bibliotheque-universitaire-de-medecine-vancouver.csl
adam3smith: Yes, I just don't want literal copies of the locale file in there. Re-defining terms is fine, that's what the section is for.

adam3smith:
Thanks!

alerac:
Thanks to you!

PR #6665, opened 2023-08-30
Create deutsche-zeitschrift-fur-sportmedizin.csl

POBrien333:
https://forums.zotero.org/discussion/72384/style-request-german-journal-of-sports-medicine#latest


adam3smith:
Thanks!

PR #6666, opened 2023-08-30
Create british-journal-of-nutrition.csl

POBrien333:
https://forums.zotero.org/discussion/99617/citation-for-british-journal-of-nutrition#latest


adam3smith:
Could you add the right documentation link, please?

PR #6670, opened 2023-08-31
Sage Harvard: fix DOI for article-journal

POBrien333:
https://forums.zotero.org/discussion/107360/style-request-environment-and-planning-e-nature-and-space


bwiernik:
Thanks!

PR #6672, opened 2023-09-04
Create journal-of-demographic-economics.csl

POBrien333:
https://forums.zotero.org/discussion/107491/author-full-name-in-bibliography#latest


adam3smith:
A couple of minor issues:
- spell out first names
- book titles (both chapters and monographs) should be title case
- in text delimiter is a comma.

More generally, I'm not sure whether including the square brackets is a good idea. The style is a hot mess, but seems to use narrative citations broadly, and omit author would do `Smith [(1776)]` Thoughts?

POBrien333:
Yeah, you're right. I've removed the square brackets.

PR #6676, opened 2023-09-05
Update the-university-of-winchester-harvard.csl

lnrdgrnt:
Any URL in references now needs to be italic. This edit adds font-style="italic" to the relevant lines.


adam3smith:
Thanks (auto-generated preview isn't right, but Zotero et al will do this correctly for DOIs)

PR #6677, opened 2023-09-06
Update norsk-apa-manual.csl

kbernt:
This makes it possible to use "entry-encyclopedia" for online encyclopedia articles with an update date - and to reproduce the full date in the bibliography, not just the year. Example:

In text: (Hoffmann, 2019).
In bibliography: 
Hoffmann, M. (2019, 29. november). Folkekunst. I Store norske leksikon. https://snl.no/.versionview/1031735


adam3smith:
Thanks!

PR #6682, opened 2023-09-13
Update and rename experimental-dermatology.csl to dependent/experimen…

POBrien333:
…tal-dermatology.csl

https://forums.zotero.org/discussion/107709/style-request-experimental-dermatology#latest


adam3smith:
Thanks!

PR #6690, opened 2023-09-14
Switch psychological-medicine.csl to APA6 dependent

POBrien333:
https://forums.zotero.org/discussion/comment/443431/#Comment_443431


adam3smith:
Thanks!

PR #6698, opened 2023-09-16
Create associacao-brasileira-de-normas-tecnicas-numerico.csl

thyagoleal:
Modified the citation from author-date to numeric format according to allowed In NBR 6023


POBrien333:
This style was created in a different PR: https://github.com/citation-style-language/styles/pull/6909

PR #6700, opened 2023-09-18
Update bibtex.csl to conform the standard for month

coelias:
the month should me in lowercase and not between brackets


adam3smith:
The current format isn't actually against standard -- BibTeX itself is silent on this and `{February}` is completely valid -- but I agree that the three letter lowercase version is typically used and more widely supported, so happy to switch this.

PR #6703, opened 2023-09-19
Update biochemical-journal.csl

POBrien333:
https://forums.zotero.org/discussion/107856/updating-the-biochemical-journal-format#latest

* change et-al rules to 7/6
* format container-title in italics
* add DOI, and URL fallback for webpage/post/post-weblog (it just looked very naked and don't think it does any harm)

### DIFF
--- a/biochemical-journal.csl
+++ b/biochemical-journal.csl
@@ -122,7 +122,7 @@
     </layout>
   </citation>
   <bibliography et-al-min="7" et-al-use-first="6" second-field-align="flush" entry-spacing="0">
-    <layout suffix=".">
+    <layout>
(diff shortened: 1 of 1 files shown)


bwiernik:
Maybe let's keep the `.` following the text before the DOI/URL as a delimiter and omit it after the DOI/URL? What do the style guidelines say there ?

POBrien333:
I think they actually don't have any dots at the end.

`41Raschi , A. , Miglietta , F. , Tognetti , R. and Van Gardingen , P.R. (2008 ) Plant Responses to Elevated CO2: Evidence From Natural Springs , Cambridge University Press`

adam3smith:
Thanks!

PR #6706, opened 2023-09-20

alex-ball:
This update allows the `article` type to be used (as intended in the spec) for preprints in a repository. Previously the `report` type had to be used, since the Harvard (Bath) style formats them in a very similar way to reports.


adam3smith:
Thanks!

PR #6709, opened 2023-09-20
Update APA for CSL 1.0.2

bwiernik:


bwiernik:
All of the major updates are in! Just some minor variables to add (eg, part-number/part-title) and checking!

bwiernik:
It's done. Now just need to propagate changes to other APA style variants.

bwiernik:
Will update the Basque and Turkish variants in separate PRs so that their original authors can review

PR #6710, opened 2023-09-21
Create universite-de-lausanne-institut-d-archeologie-et-des-sciences-…

leonienussbaum:
CSL Style for the Institut d'archéologie et des sciences de l'antiquité of the University of Lausanne


adam3smith:
Thanks & sorry for the wait (FWIW, the old style title wasn't the problem, there's just a backlog, but I don't think ASA was a particularly good label for the style, so I'm glad you changed it here)

PR #6714, opened 2023-09-24
Update haute-ecole-de-gestion-de-geneve-iso-690.csl

grolimur:
added information to distinguish patents' submission and publication dates +
locale cleaning

### DIFF
--- a/haute-ecole-de-gestion-de-geneve-iso-690.csl
+++ b/haute-ecole-de-gestion-de-geneve-iso-690.csl
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
   <info>
-    <title>Haute école de gestion de Genève - ISO 690</title>
-    <title-short>HEG Genève - ISO 690</title-short>
+    <title>Haute école de gestion de Genève (ISO 690)</title>
+    <title-short>HEG Genève (ISO 690)</title-short>
     <id>http://www.zotero.org/styles/haute-ecole-de-gestion-de-geneve-iso-690</id>
     <link href="http://www.zotero.org/styles/haute-ecole-de-gestion-de-geneve-iso-690" rel="self"/>
     <link href="http://www.zotero.org/styles/iso690-author-date-en" rel="template"/>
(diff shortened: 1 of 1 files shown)


haute-ecole-de-gestion-de-geneve-iso-690.csl
```diff
@@ -17,29 +17,37 @@
...
+      </term>
       <term name="page" form="short">
         <single>p.</single>
         <multiple>pp.</multiple>
       </term>
+      <term name="number-of-pages" form="short">
+        <single>p.</single>
+        <multiple>pp.</multiple>
+      </term>
+      <!-- the 2 following terms, never used, are hijacked to distinguish the diffrents dates (submitted, issued) of a patent -->
```
adam3smith, line 40: No, let's not do this  -- completely re-interpreting terms (this is for spring and summer, btw.) doesn't have any benefits over just using `text value` in the style and hardcoding the term, which I'd suggest doing here (since we indeed missed to include the date lables for patents in 1.0.2

grolimur: Hi @adam3smith,

I didn't want to that this way, but...

1. hardcoding the term is not suitable has the HEG citation style is multilingual
2. none of Zotero, Mendeley or Papers have a field corresponding "season-01", "season-02"... CSL variables
3. I really need to find a way to distinguish the submission date from the publication date in the reference

That said, I can wait a little bit. Is there any plan to include date labels for patents in CSL soon?

adam3smith: Right, I missed the multilingual part, but you're wrong about the lack of the CSL variables:
Season 1 is just Spring. It's currently not directly accessible via the Zotero date field, but CSL understands ISO extended date forms, so `Issued: 2022-21`  in Extra in Zotero, e.g., will be rendered as the localized season-01 (so Spring 2022 in English, printemps 2022 in French, etc.). A better date field in Zotero can happen at any time, which would allow native input of this. Pandoc, of course, already supports this. In other words, people may already have data in Zotero that would make this appear in the middle of the date in your style.
So if you must 'hijack' a variable, use another term, please.

grolimur: Thanks @adam3smith ! I didn't know that the date could be used as season variable from the `Extra` field.
I understand that updates in CSl cannot be quick.
As it seems that I have no other than hijacking a CSL variable, I'll look for another one.

adam3smith:
Thanks!

PR #6717, opened 2023-09-28
Create modern-language-association-annotated-bibliography.csl

calumapplepie:
Note: this style currently fails to render with hanging indents for an unknown reason.  I tested with Libreoffice, for the record

### DIFF
--- a/modern-language-association-annotated-bibliography.csl
+++ b/modern-language-association-annotated-bibliography.csl
@@ -5,12 +5,12 @@
     <title-short>MLA annotated</title-short>
     <id>http://www.zotero.org/styles/modern-language-association-annotated-bibliography</id>
     <link href="http://www.zotero.org/styles/modern-language-association-annotated-bibliography" rel="self"/>
+    <link href="http://www.zotero.org/styles/modern-language-association" rel="template"/>
     <link href="http://style.mla.org" rel="documentation"/>
     <author>
       <name>Calum McConnell</name>
       <email>calumlikesapplepie@gmail.com</email>
     </author>
-    <link href="http://www.zotero.org/styles/modern-language-association" rel="template"/>
     <category citation-format="author"/>
     <category field="generic-base"/>
     <summary>This style extends the MLA 9th Edition with an notes field, for use in an annotated bibliography. Currently breaks hanging indents.</summary>
--- a/modern-language-association-annotated-bibliography.csl
+++ b/modern-language-association-annotated-bibliography.csl
@@ -1,0 +1,341 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" page-range-format="minimal-two">
+  <info>
+    <title>Modern Language Association 9th edition - Annotated Bibliography</title>
+    <title-short>MLA annotated</title-short>
+    <id>http://www.zotero.org/styles/modern-language-association-annotated-bibliography</id>
+    <link href="http://www.zotero.org/styles/modern-language-association-annotated-bibliography" rel="self"/>
+    <link href="http://style.mla.org" rel="documentation"/>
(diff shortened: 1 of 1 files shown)


modern-language-association-annotated-bibliography.csl
```diff
@@ -0,0 +1,341 @@
...
+        </group>
+        <group delimiter=", ">
+          <!---This group corresponds to MLA's "Container 2"-->
+          <!--currently just using this one for archival info-->
+          <text macro="container2-title"/>
+          <text macro="container2-location"/>
+        </group>
+        <text macro="accessed"/>
+      </group>
+      <text variable="note" display="indent"/>
```
bwiernik, line 338:
```suggestion
      <text variable="note" display="block"/>
```

bwiernik (changes requested):
Thanks! I made a comment for how to fix the indenting

adam3smith:
I haven't tested, but I thought the display elements don't work well with formatting added through cs:bibliography (like hanging-indent="true") here?

I'll generally say that I'm still quite skeptical about including annotated bibliographies beyond the two we already have. It's a rabbit hole of possible additions that all vary in three lines of code & require maintenance.

calumapplepie:
> Thanks! I made a comment for how to fix the indenting 

The indentation problem isn't that the note has a first-line indent: that was an intentional choice on my part, to set the note apart from the citation.  The problem is that the _citation_ lacks a hanging indent as a result of the display problems

calumapplepie:
> I haven't tested, but I thought the display elements don't work well with formatting added through cs:bibliography (like hanging-indent="true") here?

It doesn't, no, and I mention the limitation in my style description

> I'll generally say that I'm still quite skeptical about including annotated bibliographies beyond the two we already have. It's a rabbit hole of possible additions that all vary in three lines of code & require maintenance.

That's true, but MLA is probably one of the most used styles in education, which is where a lot of people are tasked with creating annotated bibliographies in a PDF form.  I personally have been asked to make such a document thrice in three different courses, always in MLA format.

While the combinationatorial explosion is problematic, I don't think most styles will want, need, or get this additional form.  It's really only in an educational setting where you need to produce annotated bibliographies as standalone documents, as opposed to just reading the "notes" section in Zorero.

bwiernik:
I think MLA is a reasonable one to add. It's probably the most common format requested of students after APA (eg many US high school students).

calumapplepie:
I intentionally used display=indent, as opposed to the display=block added by @bwiernik .  The indent added isn't the problem I mention and face.  I want the note block to be indented; the problem is making it display as a block *at all* causes the citation to lose its hanging indent.

With display=indent:
image

With display=block:
image

Without either:
image

The ideal layout would probably be:
image

But I see no way to accomplish that.

bwiernik:
What program are you using to generate your bibliographies?

As @adam3smith mentions, the display behavior depends somewhat on the environment in which the citation is rendered. I don't see the behavior you describe with Zotero. 

With display="block" using Zotero's Word plugin or RTF export (hanging indent with annotation aligned with hanging):
image

With display="block" using Zotero's direct HTML export (hanging indent with annotation aligned with margin, with its own hanging indent): 
image

With display="indent" and html output, I see a hanging indent and then further indenting of the annotation:
image

With display="indent" and the Word plugin or RTF output, I see no hanging indent and the annotation indented:
image

bwiernik:
Based on this behavior and precedent with the existing annotated bibliography styles, display="block" is what should be used.

calumapplepie:
Appolgies, this fell off my to-do list.  I used Libreoffice; if the behavior can't be reproduced in other programs, I'll use display=block.

Maybe someone should look into making hanging-indent behavior more consistent, possibly by making it less of a special-case.

calumapplepie:
I believe this is in the state that @bwiernik desired, after the above force-push; if not, I am comfortable with the changes requested, now that I know it's a libreoffice-specific issue.  Thank you!

PR #6719, opened 2023-10-02
remove publisher-place for booklike acta-universitatis-agriculturae-s…

POBrien333:
…ueciae.csl

https://forums.zotero.org/discussion/108228/style-update-acta-universitatis-agriculturae-sueciae#latest


adam3smith:
Thanks!

PR #6724, opened 2023-10-06
Update univerza-na-primorskem-fakulteta-za-vede-o-zdravju-apa.csl

patrikpucerfvz:
Updated style to APA 7

### DIFF
--- a/univerza-na-primorskem-fakulteta-za-vede-o-zdravju-apa.csl
+++ b/univerza-na-primorskem-fakulteta-za-vede-o-zdravju-apa.csl
@@ -7,7 +7,7 @@
     <id>http://www.zotero.org/styles/univerza-na-primorskem-fakulteta-za-vede-o-zdravju-apa</id>
     <link href="http://www.zotero.org/styles/univerza-na-primorskem-fakulteta-za-vede-o-zdravju-apa" rel="self"/>
     <link href="http://www.zotero.org/styles/apa-6th-edition" rel="template"/>
-    <link href="https://apastyle.apa.org/style-grammar-guidelines/references/examples" rel="documentation"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
How is this style different from APA 7th set to Slovenian?

patrikpucerfvz:
> How is this style different from APA 7th set to Slovenian?

The APA 7th in Slovenian language uses "&" when citing instead of "in", which is the correct form in Slovenian.

bwiernik:
So it is identical to the APA 7 (no ampersand) style (which uses the localized word of "and" rather than &)?

patrikpucerfvz:
> So it is identical to the APA 7 (no ampersand) style (which uses the localized word of "and" rather than &)?

More or less, with some other minor changes like the editors' tagging, which is not capitalized (so ur. instead of Ur.) and stuff like that. These are all changes accepted at out faculty and we needed to change the original APA 7 style.

What are you suggesting? To change the original APA 7 in Slovenian instead?

adam3smith:
> What are you suggesting? To change the original APA 7 in Slovenian instead?

We weren't suggesting as much as wondering if it'd be possible to avoid some of the massive code duplication, but given the extent of the changes, I don't think that's the case here. I think if we were able to localize both the ampersand and the contextual comma, we could consider what to do about the lowercasing of abbreviations etc.

adam3smith:
That said -- there is no longer a link to documentation for this variant, which is a problem. If the style diverges from APA, just linking to the apastyle website isn't going to work.

patrikpucerfvz:
> That said -- there is no longer a link to documentation for this variant, which is a problem. If the style diverges from APA, just linking to the apastyle website isn't going to work.

I have not put a link to the documentation, because there's no official URL yet. The final draft of the documentation is under revision and I hope it will be published till the end of this week.

adam3smith:
OK, this is good to merge from my end. One final inquiry -- it lists APA 6th as a template. That's correct? This looks like you may have started from the 7th edition?

adam3smith:
Thanks!

PR #6725, opened 2023-10-07

POBrien333:
https://forums.zotero.org/discussion/108349/style-update-request-american-society-for-horticultural-science

Observation:
no "submitted" or "granted" CSL terms were available (for patents). Add in a future CSL update?


adam3smith:
Yes, we somehow missed those terms in 1.0.2 although they're pretty commonly needed

PR #6727, opened 2023-10-10
Create journal-of-rehabilitation-medicine.csl

POBrien333:
https://forums.zotero.org/discussion/99514/style-request-journal-of-rehabilitation-medicine#latest


adam3smith:
Thanks!

PR #6729, opened 2023-10-11
Update lund-university-school-of-economics-and-management.csl

Hannahglad:
The style has changed regarding how many authors you spell out and when you use et al. Therefore I have changed the number 4 to 3 to match the style.


adam3smith:
Thanks!

PR #6731, opened 2023-10-11
Update ecology-letters.csl

### LINKED ISSUE #6730: ecology-letters.csl: in-text citations are sorted chronologically instead of alphabetically
> In the text, references should always be listed alphabetically and in full unless there are three or more authors. Example: Able et al. 1997; Able & Charles 1986; Lowe et al. 1986.

Source: https://onlinelibrary.wiley.com/page/journal/14610248/homepage/ForAuthors.html

POBrien333:
closes #6730


adam3smith:
Thanks!

PR #6733, opened 2023-10-12
Update oikos.csl

POBrien333:
https://forums.zotero.org/discussion/108474/zotero-style-results-in-et-al-bibliography-although-it-should-not#latest


adam3smith:
Thanks!

PR #6740, opened 2023-10-15
Update iainutuban-tarbiyah.csl

LPPMiainutuban:
We want our style easier to find by our Student/Author on Mendeley/Zotero style database, so we change the style title, id, and summary.


adam3smith:
Why would this make the style easier to find? Also, is the change accurate -- i.e. is the style for all of IAINU or just for one faculty?

LPPMiainutuban:
Pardon me, our student just implented about this thing (software-based citation style) on their theses, So I think the updated style name will easier for them to find out. Also, When we propose this style early, we only have two faculty, but This year we have more faculty on IAINU. After several meeting about citation style, we decided to merge another faculty citation style into IAINU style only.

PR #6741, opened 2023-10-15
New Style for Ethnobiolgy and Conservation

iblima:
New style for [Ethnobiology and Conservation](https://ethnobioconservation.com/index.php/ebc) 2th edition.


iblima:
Update the commite.

PR #6742, opened 2023-10-15
Remove extra sprace

koppor:
Follow-up to https://github.com/citation-style-language/styles/pull/6649. @POBrien333 merged to `group`s in "chapter". I think, by accident, the space in following change was not removed:

```diff
-         <group delimiter=", " suffix=", ">
+         <group delimiter=", " suffix=". ">
```

This PR fixes it.

### DIFF
--- a/ieee.csl
+++ b/ieee.csl
@@ -363,7 +363,7 @@
           <text macro="access"/>
         </else-if>
         <else-if type="chapter">
-          <group delimiter=", " suffix=". ">
+          <group delimiter=", " suffix=".">
             <text macro="title"/>
             <group delimiter=" ">
               <text term="in" suffix=" "/>


koppor:
I am not 100% sure if the update fixes the issue.

Nevertheless, the probability is high ^^.

Here the effect of an update of the csl styles in JabRef. That lead me to check the updates in `ieee.csl`.

image

adam3smith:
FWIW, citation processors should strip ending and duplicate spaces -- I'm surprised that that's not happening in JabRef (are they using citeproc-java or citeproc BibLaTeX under the hood?) -- but always better to keep things tidy & happy to accept this. citeproc-ruby as used for the preview does also notice the extra space.

koppor:
> are they using citeproc-java or citeproc BibLaTeX under the hood?

JabRef is using citeproc-java `3.0.0-beta.2`. - Others are also reporting issues with spaces (e.g., citeproc-java:3.0.0-beta.2).

The failing test was https://github.com/JabRef/jabref/blob/260ea31cfc1986502e18227bcae57ef2624dd3fe/src/test/java/org/jabref/logic/citationstyle/CitationStyleGeneratorTest.java#L146C10-L146C10.

The issue was that before the final newline, there was a space:

```diff
-       String expectedCitation = "[1]B. Smith, “An article,” J. Jones, Ed., Somewhere: Great Publisher, 2021, pp. 1–10. \n";
+       String expectedCitation = "[1]B. Smith, “An article,” J. Jones, Ed., Somewhere: Great Publisher, 2021, pp. 1–10.\n";
```

Thus, a simple space strip would not help. In case of multiple references, one would need to replace `{space}\n`.

Thus, this is the more elegant solution. Thank you for merging!

PR #6744, opened 2023-10-16
add original-date to history-of-the-human-sciences.csl

POBrien333:
https://forums.zotero.org/discussion/comment/446101/#Comment_446101


adam3smith:
Thanks!

PR #6750, opened 2023-10-20
Update correction on associacao-brasileira-de-normas-tecnicas-ufrgs.csl

PriscilaSJUFRGS:
Including corrections after last update


adam3smith:
Thanks!

PR #6754, opened 2023-10-24
Update australasian-journal-of-philosophy.csl

antonyeagle:
change entry-spacing to 0 from default of 1; default is too loose (which I only noticed after pandoc updated to 3.1.8 and corrected its previously incorrect treatment of the defaults)


POBrien333:
combined into #6759

PR #6758, opened 2023-10-24
Create korea-baptist-theological-seminary-of-style-2021y-2nd-ed.csl

Choe-Uk:


POBrien333:
@Choe-Uk 
Can you confirm where this style is documented?

PR #6762, opened 2023-10-28
Create vancouver-brackets-doi-no-et-al and national-library-of-medicine-brackets-doi-no-et-al.csl 

pvzhelnov:
This is a tweak of the vancouver-brackets-no-et-al style:

https://github.com/citation-style-language/styles/commit/68ed5f7645a5238ee7e62e7540dcada514fd1c24

Also, a tweaked dependent NLM style is added.

The tweak is removing the "available at" URL and "accessed-date" elements for records containing a DOI field. Instead, the DOI is displayed in the "doi: 0000000/000000000000" notation as per documentation:

https://www.nlm.nih.gov/bsd/uniform_requirements.html#electronic

This tweak is especially important for the health and medical field. As the foreword to the last edition of the NLM Style Guide put it:

> For materials on the Internet the medium designator [Internet] is required following the title, and the URL is required. In 2015, almost everyone recognizes a string starting with http: or https: or doi: as a URL or address for something on the Internet. Do we still need to require Internet in brackets as a medium designator?

Source: https://www.ncbi.nlm.nih.gov/books/NBK310495/

It is noteworthy that PubMed does not use a space between the colon and the DOI - however, the NLM Style Guide recommends its use in the same notation as the general Vancouver guidance does:

> To use a DOI in a citation:
>
> - Begin with doi followed by a colon and a space
> - Enter the number supplied by the publisher
>
>     Example:
>
>     Puri S, O'Brian MR. The hmuQ and hmuD genes from Bradyrhizobium japonicum encode heme-degrading enzymes. J Bacteriol [Internet]. 2006 Sep [cited 2007 Jan 8];188(18):6476-82. Available from: http://jb.asm.org/cgi/content/full/188/18/6476?view=long&pmid=16952937 doi: 10.1128/JB.00737-06

Source: https://www.ncbi.nlm.nih.gov/books/NBK7281/box/A56888/

### DIFF
--- a/dependent/national-library-of-medicine-brackets-doi-no-et-al.csl
+++ b/dependent/national-library-of-medicine-brackets-doi-no-et-al.csl
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" default-locale="en-US">
+  <info>
+    <title>National Library of Medicine (brackets, DOI, no "et al.")</title>
+    <title-short>NLM</title-short>
+    <id>http://www.zotero.org/styles/national-library-of-medicine-brackets-doi-no-et-al</id>
+    <link href="http://www.zotero.org/styles/national-library-of-medicine-brackets-doi-no-et-al" rel="self"/>
+    <link href="http://www.zotero.org/styles/vancouver-brackets-doi-no-et-al" rel="independent-parent"/>
+    <link href="http://www.ncbi.nlm.nih.gov/books/NBK7256/" rel="documentation"/>
+    <category citation-format="numeric"/>
+    <category field="medicine"/>
+    <updated>2023-10-28T17:42:00-04:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+</style>
--- a/vancouver-brackets-doi-no-et-al.csl
+++ b/vancouver-brackets-doi-no-et-al.csl
@@ -1,0 +1,373 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" initialize-with-hyphen="false" page-range-format="minimal">
+  <info>
+    <title>Vancouver (brackets, DOI, no "et al.")</title>
+    <id>http://www.zotero.org/styles/vancouver-brackets-doi-no-et-al</id>
+    <link href="http://www.zotero.org/styles/vancouver-brackets-doi-no-et-al" rel="self"/>
+    <link href="http://www.zotero.org/styles/vancouver-brackets-doi-no-et-al" rel="template"/>
+    <link href="http://www.nlm.nih.gov/bsd/uniform_requirements.html" rel="documentation"/>
+    <author>
+      <name>Michael Berkowitz</name>
+      <email>mberkowi@gmu.edu</email>
+    </author>
+    <contributor>
+      <name>Sean Takats</name>
+      <email>stakats@gmu.edu</email>
+    </contributor>
+    <contributor>
+      <name>Sebastian Karcher</name>
+    </contributor>
+    <contributor>
+      <name>Paul Lin</name>
+    </contributor>
+    <contributor>
+      <name>Pavel Zhelnov</name>
+      <email>pavel@zheln.com</email>
+    </contributor>
+    <category citation-format="numeric"/>
+    <category field="medicine"/>
+    <summary>Vancouver style as outlined by International Committee of Medical Journal Editors Uniform Requirements for Manuscripts Submitted to Biomedical Journals: Sample References.  Modified to use brackets, DOI, and to list all authors in the bibliography.</summary>
+    <updated>2023-10-28T17:42:00-04:00</updated>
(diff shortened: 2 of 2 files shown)


adam3smith:
Thanks! Can you explain more specifically what this is for, though?

The foreword from Citing Medicine that you cite specifically says that the change to internet sources _should_ be considered in the future, but aren't in this edition. As for the DOI, that's provided in only in a couple of the examples, even though many, many of the sources presented in Citing Medicine did have DOIs in 2016. So while I'm all for removing the silly [Internet] and using DOIs wherever possible in 2023, I don't think that's an accurate reflection of Citing Medicine as published, so I'm a bit skeptical about adding these styles, certainly under that label.

FWIW, we have a good number of styles that do roughly this -- AMA style, e.g., as well as some of the publisher Vancouver variants, so it's not like authors have no options.

pvzhelnov:
Thanks for reviewing @adam3smith 

To your comments:

Yes, I believe you are right about the foreword – my mistake. They do indeed refer to the ANSI/NISO Z39.29 in [that passage](https://www.ncbi.nlm.nih.gov/books/NBK310495/) when I re-read it. And, to my knowledge, this standard has not been renewed since 2010, so the conclusion is valid in terms of that the [Internet] notation is still needed for any internet source as per [this standard](https://doi.org/10.3789/ansi.niso.z39.29-2005R2010) (see 3.2.4. Internet Citations).

However, what is also true is that Citing Medicine does include specific guidance for journal articles with DOI provided that goes:

> 71. Journal article with DOI provided
> Bhutta ZA, Darmstadt GL, Hasan BS, Haws RA. Community-based interventions for improving perinatal and neonatal health outcomes in developing countries: a review of the evidence. Pediatrics. 2005 Feb;115(2 Suppl):519-617. doi:10.1542/peds.2004-1441.

Source: https://www.ncbi.nlm.nih.gov/books/NBK7282/#A32751

As you can see in this arguably unambiguous example, the case exemplified is an _article published in print_ (note the page range indicated) but, for also having an electronic version, featuring a DOI. Clearly, the [Internet] provision is not strictly applicable here, but the DOI is preferably – and in accord with Citing Medicine – to be shown. For this reason, I am not sure if it is relevant how many sources presented in Citing Medicine did have DOIs in 2016, given the direct guidance on that matter from NLM.

At the same time, I cannot help but agree with you that this perceived clarity actually goes away when you look elsewhere in Citing Medicine:

> Puri S, O'Brian MR. The hmuQ and hmuD genes from Bradyrhizobium japonicum encode heme-degrading enzymes. J Bacteriol [Internet]. 2006 Sep [cited 2007 Jan 8];188(18):6476-82. Available from: http://jb.asm.org/cgi/content/full/188/18/6476?view=long&pmid=16952937 doi: 10.1128/JB.00737-06

Source: https://www.ncbi.nlm.nih.gov/books/NBK7281/box/A56888/

Here, we have another print article but with the dreadful [Internet]/cited notation this time.

Taking all this together, I have made a case for print articles that have a DOI, for which showing the DOI in the reference is not a matter of their nature (electronic v. print) or uniqueness (which has already been established at that point by the combination of journal, volume, and page range, as compared to URL or DOI for internet sources) but, rather, a pure matter of style and utility when looking it up online. I have also shown that Citing Medicine is not strict enough about such cases, but, at the same time, showing the DOI for a print reference is definitely permitted and encouraged by this manual.

As to my specific case, I am a long-standing user of the PubMed citation format. To be even more specific as to what finally motivated me to do this submission, I have recently started using [zbib.org](https://zbib.org/) routinely for downloading references, and their exported records tend to have a url field included, leading to all articles being shown as [Internet] sources when rendered with `national-library-of-medicine-brackets-no-et-al.csl`. So, I had two options: (1) to write a script parsing those bibs or (2) modify the CSL. Given that my understanding was (and proved correct) that showing the DOI is encouraged by Citing Medicine, I prepared this submission.

That said, I believe it is great that you raised that point. Because, strictly speaking, I need to modify the style now so that the [Internet] notation (including the date accessed) is hidden only for those records that contain both the url field and the pages field. Even more, numbers only should be allowed in the pages field to be more or less sure that this is a print publication (even though I recognize this is not a perfect strategy, but keen to read other suggestions).

I am happy with dropping the label if the alternative is to modify the parent Vancouver style. I could definitely go with my custom style too (or use one of those publisher Vancouver variations as you suggested), but I am really worried that the situation I am in might be really common in health sciences (i.e., DOIs not being shown for print articles when using the `national-library-of-medicine-brackets-no-et-al.csl` style), hence my pull request and invitation to this discussion.

You know, at the end of the day, maybe leaving the style at that (i.e., with the "-doi-" affix in the label) would actually be the perfect trade-off because the source manual both allows and does not enforce using the DOI for print articles. I stress that this use case might be of relevance to many.

pvzhelnov:
In other words, I am talking about modifying the CSL as follows:

* show the URL and date accessed for the article-journal document type if and only if the variable doi is present AND any of these conditions applies: any of the variables journal, volume, or page is not present, or page-first is not a number,
* show the DOI in the "doi: 0000000/000000000000" notation whenever the variable doi is present.

I recognize that the former modification will miss some electronic-only publications. However, this trades off against the doi being shown anyway. However yet, these would still be non-compliant with the manual, whereas the current edition, albeit generating excessive [Internet], is compliant.

Thus, it looks like this is gonna have to wait until the guidance explicitly allows for going DOI-only😂

pvzhelnov:
But, maybe we can at least add the doi modification? That is:

* show the DOI in the "doi: 0000000/000000000000" notation whenever the variable doi is present.

If in addition to URL, then be it. I don’t believe DOIs are shown in the basic Vancouver version. Specifically in my case, when I download refs from zbib.org, only the messy URLs are shown but not the DOIs, even though the DOI is included in the bib.

PR #6778, opened 2023-11-15
Update weed-science-society-of-america.csl

wyoweeds:
Changed line 179 in this style file. In-text citation groups should be separated by a semicolon instead of a colon, as noted in the journal instructions to authors: https://www.cambridge.org/core/journals/weed-science/information/author-instructions/preparing-your-materials


adam3smith:
Thanks!

PR #6784, opened 2023-11-22
Create leidraad-voor-juridische-auteurs-2022.csl

CaspervanWetten:
Support for the Dutch guidelines for judicial authors 2022 style

### DIFF
--- a/leidraad-voor-juridische-auteurs-2022.csl
+++ b/leidraad-voor-juridische-auteurs-2022.csl
@@ -1,0 +1,638 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" default-locale="nl-NL">
+<info>
+    <title>Leidraad voor juridische auteurs 2022</title>
+    <title-short>Leidraad 2022</title-short>
+    <id>http://www.zotero.org/styles/leidraad-voor-juridische-auteurs-2022</id>
+    <link href="http://www.zotero.org/styles/leidraad-voor-juridische-auteurs-2022" rel="self"/>
+    <link href="http://www.zotero.org/styles/leidraad-voor-juridische-auteurs" rel="template"/>
+    <link href="https://www.wolterskluwer.nl/auteur/voor-auteurs/auteursmiddelen-en-instructies" rel="documentation"/>
+    <link href="https://www.wolterskluwer.com/-/media/project/wolterskluwer/oneweb/www/lr/nl-media/auteurs%20voor-auteurs/pdf26920190802161350.pdf" rel="documentation"/>
+    <author>
+      <name>Casper van Wetten</name>
+      <email>c.o.vanwetten@gmail.com</email>
+    </author>
+    <category citation-format="note"/>
+    <category field="law"/>
+    <updated>2023-05-09T05:06:20+00:00</updated>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks -- are both the 2019 and the 2022 version in use or should we just replace the 2019 version with this one?

CaspervanWetten:
Good question, as far as I know (I'm just a student) everyone using the 2019 version is recommended to switch over to 2022 . Although I expect that very few people (will) use either of these anyways, I'd keep both just for archival reasons.

mastaal:
I was wondering if this PR could be merged, or that some extra work is possibly needed? @CaspervanWetten @adam3smith

CaspervanWetten:
As far as I'm concerned this PR can be merged at any moments. I've been using it personally in the meantime and haven't encountered any issues.

adam3smith:
The style looks fine, but we'd want to properly handle the naming (if you want to do this, we'll want this version to replace the existing leidraad style and keep the existing one with a different id/filename/title that clearly labels it -- we typically use edition number, but if edition name is all we have that's what we'll go by.

lrpvo:
This has been quiet for a while.

To clarify: the updated Leidraad (most recent, 2022) is the only one that is currently being used, and the only one acceptable for academic purposes. As a result, the 2019 edition should be removed, as it has been updated and replaced by the 2022 version. The changes are relatively small.

The code proposed here in #6784 was very clearly based on the work done previously for the 2019 edition, as almost all of the original code remains intact. I would suggest the original authors remain credited with their work, which was most recently updated in #5242. The proposed commit sees the names of those original authors removed.

Unfortunately, I don't think I am - at this time - competent enough to verify all the changes made to the "Leidraad" in 2022 are included in the code in this commit, as the full "Leidraad" is a 157 page document on the one hand, and I am unfamiliar with the code (and the language as a whole) on the other.

I.e., the most significant change in the Leidraad 2022 is the change to, for the shortened references in footnotes, no longer include the full reference to the academic journal but to instead follow the short-reference rules for books.

Example for shortened references, 2019:
- Book: Reid & Baldé 2020, p. 56.
- Journal: Sirbu, Miron & Fekkes, NJB 2021/146.
Example updated shortened references, 2022:
- Book: Reid & Baldé 2020, p. 56.
- Journal: Sirbu, Miron & Fekkes 2021.

As a result, I would expect to see a change in the macro reference-book-etc-short, and while a change is made, the proposed commit includes the number-journal-short, which seems not to be accurate with the Leidraad 2022.

Furthermore, the term "page" is adjusted from p. to r.o.; in Dutch legalese, r.o. is used as a subsection of a **legal decision only**, but this is not applicable to (recurring) journals, books, etc. I cannot determine if the code on line 35 of this commit actually changes the output in a relevant way, as some outputs do put "p." for page number

Would be happy to help someone with more coding experience to double-check and ensure this is a complete adaptation of the 2022 Leidraad.

CaspervanWetten:
Hallo lprvo,

Thank you for your additions, you are right that I removed the names of the authors that wrote the original xml schema. In hindsight, I shouldn't have done this and any improvement on this PR will restore their names. My apologies (this is my first time open source).

In further hindsight, this PR should not be merged. It is a sub-optimal implementation which (as my teachers pointed out)  does not always provide the expected result. 

Though, I would like to work on this further. I _think_ that if I spend the time I can get the hang of the xml scheme if I start small and validate my work along the way, however I am not familiar enough (nor do I understand it well enough) with the Leidraad 2022 to validate my own work (as proven by this commit). 
Do you think we could collaborate on this?

PR #6796, opened 2023-11-27
Update boletin-de-pediatria.csl

palantre:
Changes in journal abbreviations (adding a "." at the end) and adding a space between the year and the volume.


adam3smith:
Thanks! The extra space via suffix doesn't quite work -- see the test output for the Hancke et al edited volume. Try setting the space in a group instead?

palantre (commented):
Thanks for the suggestion. I've finally moved the extra space to the macro "date".

PR #6799, opened 2023-12-02
Create revue-dhistoire-des-sciences.csl

AntoineCDussault:
The CSL style repository did not contain any style for this journal (is has one for Revue d'histoire des sciences humaines, which is a different journal. I created one, tested it, and submitted a paper formatted through using it. I would like it to be useful to other authors submitting to this journal. This is my first attemps. Hope I'm doing things correctly.

### DIFF
--- a/revue-dhistoire-des-sciences.csl
+++ b/revue-dhistoire-des-sciences.csl
@@ -1,0 +1,295 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style class="note" version="1.0" et-al-min="3" et-al-use-first="1" page-range-format="chicago" demote-non-dropping-particle="sort-only" default-locale="fr" xmlns="http://purl.org/net/xbiblio/csl">
+  <!-- This style was edited with the Visual CSL Editor (https://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>Revue d'histoire des sciences (note)</title>
+    <title-short>CUP</title-short>
+    <id>http://www.zotero.org/styles/revue-dhistoire-des-sciences</id>
+    <link href="http://www.zotero.org/styles/revue-dhistoire-des-sciences" rel="self"/>
+    <link href="http://www.zotero.org/styles/art-history" rel="template"/>
+    <link href="https://www.revues.armand-colin.com/sites/default/files/rhs_consignesauxauteurs_maj2021.pdf" rel="documentation"/>
+    <author>
+      <name>Patrick O'Brien</name>
+      <email>obrienpat86@gmail.com</email>
+    </author>
+    <author>
+      <name>Antoine C. Dussault</name>
+    </author>
+    <category citation-format="note"/>
+    <category field="generic-base"/>
+    <summary>Ce style n'est pas parfait. En particulier, je n'ai pas réussi, dans le cas des citations &quot;op cit.&quot;, à faire apparaître la date seulement dans les cas où la publication citée apparaît avec d'autres publication du même auteur-e dans la première note où elle est citée. La date n'apparaît donc jamais il faut l'ajouter manuellement. Modifié à partir de cambridge-university-press-note de Patrick O'Brien  </summary>
+    <updated>2023-12-02T23:33:11+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <macro name="author">
+    <names variable="author">
+      <name and="text" delimiter-precedes-last="never" initialize="false" initialize-with=". "/>
+      <label form="short" strip-periods="false" prefix=" (" suffix=")"/>
+      <substitute>
+        <names variable="editor"/>
+        <names variable="translator"/>
+        <text macro="title"/>
+      </substitute>
+    </names>
+  </macro>
+  <macro name="author-short">
+    <names variable="author">
+      <name form="short" and="text"/>
+      <label form="short" strip-periods="true" prefix=", "/>
+      <substitute>
+        <names variable="editor"/>
+        <names variable="translator"/>
+        <text macro="title"/>
+      </substitute>
+    </names>
+  </macro>
+  <macro name="author-bibliography">
+    <names variable="author">
+      <name and="text" initialize-with=". " name-as-sort-order="first"/>
+      <label form="short" strip-periods="false" prefix=" (" suffix=")"/>
+      <substitute>
+        <names variable="editor"/>
+        <names variable="translator"/>
+        <text macro="title"/>
+      </substitute>
+    </names>
+  </macro>
+  <macro name="title">
+    <choose>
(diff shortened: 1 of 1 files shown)


POBrien333:
@AntoineCDussault 
I've reworked this style. Can you review it?

AntoineCDussault:
> @AntoineCDussault I've reworked this style. Can you review it?

Thanks for the improvements and my apologies for the slow reply. The result looks good. Could I download your improved version so I can further test it?

POBrien333:
Yes, of course you can download it to test it. That's what I was asking for, not just reviewing the code.

AntoineCDussault:
I reviewed the style and there were still touch-ups to be made. I made them in the attached file (pls change the extension to csl if needed). One thing I wasn't able to do is the following: when a reference is subsequent, the journal wants either "author + op. cit. n 4", if in note 4 there was only one citation from the referred-to author, or "author + date + op. cit. n 4", if in note 4 there were more than one citations from the referred-to author. I tried to do it with conditional and "disambiguated", but the year ends up always appearing. Thanks again for your help!
[revue-dhistoire-des-sciences 19.txt](https://github.com/user-attachments/files/24016748/revue-dhistoire-des-sciences.19.txt)

POBrien333:
You can make any changes here in this PR directly. No need to send files.

AntoineCDussault:
Ok, I just did so (I think). Hope I used to right procedure.

POBrien333:
I made some further improvements. Please test the style out to ensure I didn't introduce any mistakes.

AntoineCDussault:
Unfortunately, I'm noticing mistakes. Two of them are: 
1) Instead of "op. cit. in n 4" it writes op. cit. inref. 4"
2) weird comas appear for references with locator "volume", "chapter" etc. and the page locator does not appear.
Is there a way I can see the improvements you've made? That way I could try to implement them without causing these mistakes. Also, do you have any clue on how to solve the issue about op cit in with and without the date I was describing yesterday?
Thanks!

POBrien333:
Up top you can click on "commits" to see the difference in code for each commit you or I made.

I don't think what you asked yesterday is possible.
Disambiguation is generic, it's about the whole document. If there are two or more papers from Miller, 2004 for example. You cannot test for it on a per-footnote basis.

Also, your style currently doesn't output anything for the webpage example in our sample output. Please modify your style that websites/blogs/posts receive an output.

AntoineCDussault:
Since I'm unable to code (I'm working only through the Visual style editor), I wasn't able to see your changes. But so far I've been able to fix the coma issue, but not the "ref" instead of "n" issue. As far as I can analyse, the issue comes from the use of "first-reference-note-number (label)". So I'd need to be able to tell Zotero that the label should be "n" instead of "ref". I wasn't able to find how to do this, though. I could remove "first-reference-note-number (label)" and use a prefix "n", but I take it that that would be unelegant. Any suggestion/clue? Many thanks for your help!

POBrien333:
I've already redefined the term to be "n.". See the locale section at the start of the style.
Did you actually try the latest version I prepared?

AntoineCDussault:
My apologies for the delay. Now I think I have fixed all the issues (except for the year disambiguation, which you said cannot be done). Thanks again for your help!

revue-dhistoire-des-sciences.csl
```diff
@@ -0,0 +1,311 @@
...
+      <label form="short" prefix=" (" suffix=")"/>
+    </names>
+  </macro>
+  <macro name="volumes">
+    <group delimiter=" ">
+      <number variable="number-of-volumes"/>
+      <label variable="number-of-volumes" form="short"/>
+    </group>
+  </macro>
+  <macro name="edition">
```
POBrien333, line 157: replace with this:
`  <macro name="edition">
    <choose>
      <if is-numeric="edition">
        <group delimiter=" ">
          <number variable="edition" form="ordinal"/>
        </group>
      </if>
      <else>
        <text variable="edition"/>
      </else>
    </choose>
  </macro>`

revue-dhistoire-des-sciences.csl
```diff
@@ -0,0 +1,311 @@
...
+                <text macro="publisher"/>
+              </group>
+            </else-if>
+            <else>
+              <text macro="publisher"/>
+              <text macro="date" prefix="(" suffix=")"/>
+            </else>
+          </choose>
+        </group>
+        <text macro="pageref"/>
```
POBrien333: Locator does not exist in the bibliograpy. Please delete.

AntoineCDussault:
Not sure why it fails. The CSL style validator finds no issue. I removed the bibliography part because the journal does not require a bibliography. Maybe that was a mistake. Thanks.

revue-dhistoire-des-sciences.csl
AntoineCDussault: I'm a little stuck at this stage. Your suggestion made me realise that the journal style does not require a bibliography, so I deleted this whole part of the code. This generates issues not detected by the CSL style validator. Do you have any clue as to how I could get the style without code for bibliography to work? Many thanks!

POBrien333:
As it says "may not have any unused macros"
The macro that's named is not used. Delete the macro.

AntoineCDussault:
Thanks! I'll try this. A similar issue arose yesterday with another style I submitted (cahiers-francois-viete), and I was able to fix it.

revue-dhistoire-des-sciences.csl
```diff
@@ -0,0 +1,260 @@
...
+                      <group delimiter=", ">
+                        <text macro="volume"/>
+                      </group>
+                    </group>
+                  </else-if>
+                  <else-if type="thesis" match="any">
+                    <group delimiter=" ">
+                      <text variable="container-title"/>
+                      <text variable="genre"/>
+                      <group delimiter=" ">
```
POBrien333: A group with one element only doesn't make sense.

POBrien333:
Please see the changes.
Would you mind focusing on one style before creating new PRs? You're still making beginner mistakes that I'd rather you understand before opening new PRs that I then need to review and fix.

AntoineCDussault:
I get your point. I have a few other styles I recently created (for journal's I was submitting papers to). I'll do my best to apply the corrections you make to my already submitted styles before submitting these new ones.
In the style associated to this tread, am I correct that the change you've made is "&#160" to standard spaces? I'm trying to get the style to put unbreakable spaces. Why should I not use the code "&#160"? That's what I've found in zotero chats for unbreakable spaces.

AntoineCDussault:
Okay, I put back the nonbreaking spaces. The style seems to work fine now!

AntoineCDussault:
Well... I just put back the nonbreaking spaces in the style (as POBrien333 told me to do) and now I'm getting this message through email: "Automated tagging for PRs and issues / label Cancelled". Not sure what to do at this stage, I must say I'm becoming discouraged...

POBrien333:
This PR was merged a week ago and is public: https://www.zotero.org/styles?q=revue-dhistoire-des-sciences

That means you cannot make any changes to this, now closed, PR.
If you need changes in the style you'd need to make a new PR.

PR #6800, opened 2023-12-03
Create philosophia-scienciae

AntoineCDussault:
I found a style under this name in the zotero style repository, but it was not accurate at all. So I created one, tested it, and used it to format a manuscript I submitted. I would like it to be useful to other authors submitting to this journal.


adam3smith:
superseded by https://github.com/citation-style-language/styles/pull/6802

PR #6801, opened 2023-12-03
Update philosophia-scientiae.csl

AntoineCDussault:
I tried the existing style and it was not at all doing what the guidelines for authors require. I hence had to create a style derived from the closest one I found (apa no doi no issue). I would like my style to be useful to other authors submitting to this journal.


adam3smith:
superseded by https://github.com/citation-style-language/styles/pull/6802

PR #6802, opened 2023-12-03
Update philosophia-scientiae.csl

AntoineCDussault:
The style that was available for this journal was not doing at all what the guidelines for authors required. I thus created one not from the existing style but from the most closely similar I found.


adam3smith:
1. You should not set text-case="sentence" on titles -- it'll incorrectly lowercase proper nouns. CSL expects titles in sentence case, so for sentence case you shoudl just leave the text-case attribute out
2. The journal name is _[Philosophia Scientiæ](https://journals.openedition.org/philosophiascientiae/635)_ with a t, which is what the filename still is. You are using _scienciae_ (with a c) in the ID and self link (and title)

POBrien333:
Thanks. I've fixed up your style and actually overwrote the current one. (You were suggesting a new style altogether).
Style will be available for users now.

AntoineCDussault:
> Thanks. I've fixed up your style and actually overwrote the current one. (You were suggesting a new style altogether). Style will be available for users now.

Many thanks! Glad the style will now be useful to other people!

PR #6804, opened 2023-12-04
Update journal-of-pollination-ecology.csl

valentinitnelav:
I have updated the existing CSL file to support the inclusion of DOIs (Digital Object Identifiers) in citations as now suggested by the Journal of Pollination Ecology


valentinitnelav (commented):
In my initial pull request there was a missing whitespace just before the DOI in the final reference output. It is fixed now by just adding the needed whitespace

adam3smith:
Thanks!

PR #6809, opened 2023-12-08
Replace contributor email with identifier

adunning:
Replacing my inactive email address with a permanent identifier in a URI.


bwiernik:
Thanks!

PR #6814, opened 2023-12-13
Update institut-francais-darcheologie-orientale-etudes-arabes.csl

NicolasSouchon:
Update institut-francais-darcheologie-orientale-etudes-arabes.csl according to [new guidelines](https://www.ifao.egnet.net/uploads/publications/normes/IFAO_publications_normes_bibliographiques_pub_arabisantes_2023_fr.pdf)


adam3smith:
The creator rendering seems quite inconsistent here with et-al sometimes in italics sometimes not and for some reason translators -- but no other creator types -- are rendered in small caps? Could you check if that's actually all correct? It differs quite a bit in that respect from the English version of the style

NicolasSouchon:
@adam3smith you're right, they were errors, corrected now, thank you!

adam3smith:
Thanks!

PR #6815, opened 2023-12-15
Update ecole-de-technologie-superieure-apa.csl

Bibref:
Update to the style to include the Software document type now available in Zotero. Removal of the previous "hack" to give an output that looks like softwares to simply include the new type software.

### DIFF
--- a/ecole-de-technologie-superieure-apa.csl
+++ b/ecole-de-technologie-superieure-apa.csl
@@ -207,7 +207,7 @@
                 <text term="version"/>
                 <text variable="version"/>
               </group>
-              <text value="Logiciel" prefix="[" suffix="]"/>
+              <text term="software" text-case="capitalize-first" prefix="[" suffix="]"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
The way you're doing this doesn't work: you're actually removing italics from books and other item types that need them -- the hack is in the next line (testing for 'version' to determine that something is software.) You'll want to move that entire `if` section into a new `else-if type="software"`

Bibref:
> The way you're doing this doesn't work: you're actually removing italics from books and other item types that need them -- the hack is in the next line (testing for 'version' to determine that something is software.) You'll want to move that entire `if` section into a new `else-if type="software"`

Thank you for the comment. I adjusted the commit to reflect your comment and tested it out successfully.

Bibref (commented):
**This change should answer adam3smith comment on the pull request.

adam3smith:
Thanks!

PR #6816, opened 2023-12-19
update istanbul-medical-journal.csl

karamelikli:
Updated rules of citation for the journal of Istanbul Medicine


adam3smith:
Thanks!

PR #6823, opened 2023-12-22
Fix year disambiguation in chicago-author-date.csl when no issued date present

wiseth:
This PR fixes the following issues for references with no publication date:

## 1. Missing comma before "n.d." when accessed date is present

Only the publication date is displayed in in-text citations; when no publication date is present, the year displays as "n.d." even if a date of access is available. Per the [CMOS §15.44](https://www.chicagomanualofstyle.org/book/ed17/part3/ch15/psec044.html), "n.d." should be preceded by a comma:

Screenshot 2023-12-22 at 3 02 57 PM

Currently, we are not using a comma if _either_ the issued _or_ accessed date was present — even though the year will show as "n.d." in the latter case. Instead, we should omit a comma _only if_ the issued date is present.

`(Anon n.d.)` &rightarrow; `(Anon, n.d.)`

## 2. Incorrect location for year disambiguation in bibliography entry

If there are multiple references with the same author and the same publication year, a disambiguating letter should be added to the end of the year (as in "2023a"). If a publication year is absent for multiple references with the same author, the date placeholder (e.g., "n.d.") should be similarly disambiguated (as in "n.d.-a"). 

This is not currently working correctly. Instead, the disambiguating letter is being placed after other dates in the bibliography entry, such as the accessed date. If no such date is present, the disambiguating letter is missing altogether.

This PR fixes the issue. That is,

```
Anon. n.d. “Article A.” Wikipedia. Accessed March 14, 2023a. https://en.wikipedia.org/.
———. n.d. “Article B.” Wikipedia. Accessed March 14, 2023b. https://en.wikipedia.org/.
```

will now be corrected to

```
Anon. n.d.-a. “Article A.” Wikipedia. Accessed March 14, 2023. https://en.wikipedia.org/.
———. n.d.-b. “Article B.” Wikipedia. Accessed March 14, 2023. https://en.wikipedia.org/.
``` 

## 3. Missing year disambiguation in in-text citation

Similar to problem 2, the disambiguating letter is missing from in-text citations without publication dates. This PR corrects the issue.

`(Anon n.d.; n.d.)` &rightarrow; `(Anon, n.d.-a; n.d.-b)`

## Examples using Zotero

Before:
Screenshot 2023-12-22 at 3 09 45 PM

After:
Screenshot 2023-12-22 at 3 08 45 PM

### DIFF
--- a/chicago-author-date.csl
+++ b/chicago-author-date.csl
@@ -468,12 +468,15 @@
             <date-part name="year"/>
           </date>
         </group>
+        <text variable="year-suffix"/>
       </if>
       <else-if variable="status">
         <text variable="status" text-case="capitalize-first"/>
+        <text variable="year-suffix" prefix="-"/>
       </else-if>
       <else>
         <text term="no date" form="short"/>
+        <text variable="year-suffix" prefix="-"/>
       </else>
     </choose>
   </macro>
@@ -486,12 +489,15 @@
             <date-part name="year"/>
           </date>
         </group>
+        <text variable="year-suffix"/>
       </if>
       <else-if variable="status">
         <text variable="status"/>
+        <text variable="year-suffix" prefix="-"/>
       </else-if>
       <else>
         <text term="no date" form="short"/>
+        <text variable="year-suffix" prefix="-"/>
       </else>
     </choose>
   </macro>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks for the PR and detailed description. Not sure why/how the access date got into the conditional -- I wonder if we used access in citations without date previously. Anyway -- this is clearly correct CMoS, so again, thanks.

PR #6827, opened 2023-12-29

OTFlorian:
If volume is available, it needs to be mentioned (according to the documentation).

### DIFF
--- a/masarykova-univerzita-pravnicka-fakulta.csl
+++ b/masarykova-univerzita-pravnicka-fakulta.csl
@@ -207,13 +207,7 @@
     </choose>
   </macro>
   <macro name="publisher-place">
-    <group delimiter="; ">
-      <choose>
-        <if variable="publisher-place URL" match="any">
           <text variable="publisher-place"/>
-        </if>
-      </choose>
-    </group>
   </macro>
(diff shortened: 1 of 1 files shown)


masarykova-univerzita-pravnicka-fakulta.csl
```diff
@@ -238,27 +238,33 @@
...
     <choose>
       <if type="article-journal article-magazine article-newspaper" match="any">
-        <choose>
-          <if variable="issued" match="any">
-            <date variable="issued">
-              <date-part name="year" range-delimiter="&#8211;"/>
-            </date>
-          </if>
-          <else>
+        <group delimiter=", ">
```
adam3smith, line 241: You don't actually have to test for a variable before printing it: the test is implicit, i.e.:
```
          <group>
            <choose>
              <if variable="issued" match="any">
                <date variable="issued">
                  <date-part name="year" range-delimiter="&#8211;"/>
                </date>
              </if>
            </choose>
          </group>
```

is 100% identical to

```
                <date variable="issued">
                  <date-part name="year" range-delimiter="&#8211;"/>
                </date>
```

adam3smith:
Thanks!

PR #6829, opened 2024-01-02

simomac:
Added disambiguate-add-year-suffix code to deal with multiple sources with the same author and no date. ref.: https://forums.zotero.org/discussion/105914/disambiguate-add-year-suffix-not-working-when-data-is-n-d-no-date-term/p1

### DIFF
--- a/harvard-imperial-college-london.csl
+++ b/harvard-imperial-college-london.csl
@@ -107,6 +107,7 @@
       </if>
       <else>
         <text term="no date" form="short"/>
+        <text variable="year-suffix"/>
       </else>
     </choose>
   </macro>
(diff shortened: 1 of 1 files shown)


adam3smith:
If you set the term explicitly on no date you also have to explicitly set it for years.

simomac:
Thanks. There are 13 references to 'Year' in the code. Should I add '<text variable="year-suffix"/>' after each instance?
eg: After
105           <date-part name="year"/>
157            <text macro="year-date"/>
etc.?

adam3smith:
You only need to add it to the relevant year macro, not every time the word "date" appears.

PR #6833, opened 2024-01-04
Updating infoclio.ch styles

nchachereau:
This is an update to a custom style for history students in Switzerland. See #234 for the original pull request, and #5391 for the last update.

What this year's update brings:

- For some reason, Zotero now passes a "director" label (which it didn't do before), and the style would automatically display it. But our guidelines call for the director to be mentioned as the author, without any label. I added a special rule for this case.
- Uncertain dates are now supported (Zotero users can add uncertain dates by writing e.g. `date: 1932~` in the `Extra` field)
- If a book specifies the variable `event-title` (e.g. in the `Extra` field), it gets displayed between square brackets after the title. The use case is exhibition catalogues, for instance for art history.
- For many document types, missing dates were automatically reported using the CSL term `no date`. In some cases, however, the date should not appear (for instance, the guidelines provide for citing journals, using the book document type, and dates are not needed in this case). Unfortunately, there is apparently no way for a user to differenciate "There should be a date, but it's missing" from "No date needed".
- The publication year is no longer specified in subsequent citations.
- The link to the documentation has been updated.

I look forward to your feedback. Thank you for your invaluable work on this repository! 🙏

cc @janbaumanninfoclio

### DIFF
--- a/infoclio-de-kurzbelege.csl
+++ b/infoclio-de-kurzbelege.csl
@@ -5,7 +5,7 @@
     <id>http://www.zotero.org/styles/infoclio-de-kurzbelege</id>
     <link href="http://www.zotero.org/styles/infoclio-de-kurzbelege" rel="self"/>
     <link href="http://www.zotero.org/styles/infoclio-de" rel="template"/>
-    <link href="https://www.infoclio.ch/de/node/133932" rel="documentation"/>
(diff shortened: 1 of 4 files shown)


adam3smith:
Thanks, this all looks good (Zotero maps director to director because that now exists as a CSL variable, though I agree it's caused some issues with existing styles)

PR #6835, opened 2024-01-05
Create american-journal-of-roentgenology

drfrancisdeng:
Update of the dependent style to an independent style that correctly reflects the AJR style.


adam3smith:
You'll need to actually delete the dependent style in the same branch for this to pass CI. Also -- did you base this on any existing style? (If so, please include as `rel=template`)

POBrien333:
Continued in https://github.com/citation-style-language/styles/pull/7883

PR #6841, opened 2024-01-10
Create Bulletin-of-Applied-Glycoscience.csl

tkharborlight:


POBrien333:
@tkharborlight 
Thank you for your patience.
Could you point out where the style is described? The provided link for the documentation doesn't seem to show much of a style guide.
Alternatively, could you share a .pdf article to check the style? They say it's open access, but then have you register.

bwiernik:
@POBrien333 here is an open access article https://www.jstage.jst.go.jp/article/bag/1/2/1_KJ00009004734/_pdf/-char/en

POBrien333:
Thanks for digging that out!

PR #6849, opened 2024-01-16
Create american-meteorological-society_modified_national-science-foun…

senclimate:
…dation

Modified AMS styles for NSF proposal


adam3smith:
Thanks! My sense is that we probably don't want this for the repository. As you know, NSF doesn't care about the citation style as long as it doesn't use et al, so AMS-NSF isn't actually a specific style but rather a personal preference to use an AMS-type style without et-al in NSF proposals. That's obviously fine (and you're very much welcome to make this available through your github and/or website), but we limit the repository to official styles described by some form of authoritative guidance. Let me know if I misunderstood what this is for and, again, even if we're not going to end up accepting this PR, thank you for contributing.

bwiernik:
Closing for now. Please feel free to re-open if the style is called for by a specific agency or publisher

PR #6852, opened 2024-01-20
Update frontiers.csl

POBrien333:
https://forums.zotero.org/discussion/110927/3-errors-in-frontiers-citation-style#latest

I don't fully agree with OP on everything. Style based on guidelines.


adam3smith:
Thanks -- so this fixes 1 & 2 in that thread, the publisher in universities was always there, but we don't add the thesis type, is that right?

PR #6853, opened 2024-01-20
Update institut-francais-darcheologie-orientale.csl

NicolasSouchon:
Modification of macro "container-title" linked to recent modifications of the variable


adam3smith:
Thanks! (though to be clear -- I don't believe anything about this has changed in Zotero?)

NicolasSouchon:
@adam3smith my bad, seeing that the erroneous references continued, I found that it was a problem in my Zotero settings.

PR #6859, opened 2024-01-22
Rewrite China national standard GB/T 7714-1987

zepinglee:
This is a rewrite of the China national standard GB/T 7714-1987. The new version implements the format more faithfully and it's tested with all the sample entries in the standard.


bwiernik:
Thanks!

PR #6864, opened 2024-01-23

POBrien333:
https://forums.zotero.org/discussion/110972/style-error-acm-association-for-computer-machinery#latest


adam3smith:
Thanks!

PR #6867, opened 2024-01-24
Create wikipedia-vorlage-literatur.csl

192933488S:
Issues fixed and rename

### DIFF
--- a/wikipedia-vorlage-literatur.csl
+++ b/wikipedia-vorlage-literatur.csl
@@ -1,0 +1,304 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style
+	xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="de-DE">
+	<info>
+		<title>Wikipedia Vorlage Literatur (German Wikipedia Citation Style)</title>
+		<id>http://www.zotero.org/styles/wikipedia-vorlage-literatur</id>
+		<!-- Referenzen & Dokumentation -->
+		<link href="https://www.zotero.org/styles/wikipedia-vorlage-literatur" rel="self"/>
+		<link href="https://de.wikipedia.org/wiki/Vorlage:Literatur" rel="documentation"/>
+		<link href="https://de.wikipedia.org/wiki/Vorlage:Patent" rel="documentation"/>
+		<link href="https://de.wikipedia.org/wiki/Vorlage%3AInternetquelle" rel="documentation"/>
+		<link href="https://de.wikipedia.org/wiki/Benutzer:Cepheiden/Zotero" rel="documentation"/>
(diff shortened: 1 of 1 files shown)


192933488S (commented):
fixed

192933488S (commented):
ok

192933488S:
- self is set to https
- id is set to http

Please support.

alexriedlinger:
set both to http

192933488S (commented):
as adviced

adam3smith:
So the concern I have for this is that CSL is really not the right format for what's essentially a metadata export format. We accepted the existing Wikipedia style with some trepidation because librarians asked us to since it integrates into their library catalog. That doesn't seem to be the case here, and for Zotero, using a translator would just seem like a much better choice -- any reason you didn't go that route? 
And the German Wikipedia template really does need to be different from the English one? That's seems.... unfortunate.

192933488S:
> So the concern I have for this is that CSL is really not the right format for what's essentially a metadata export format. We accepted the existing Wikipedia style with some trepidation because librarians asked us to since it integrates into their library catalog. That doesn't seem to be the case here, and for Zotero, using a translator would just seem like a much better choice -- any reason you didn't go that route? And the German Wikipedia template really does need to be different from the English one? That's seems.... unfortunate.

Hi, and many thanks for your reply. There are some things about your reply that I can clarify, but others are out of scope for at least two reasons. Some parts of your message I do not understand. One reason is, I think, because the [Vorlage:Literatur ](https://de.wikipedia.org/wiki/Vorlage:Literatur) has a lengthy history beyond this CLS for use in Zotero, and two, this would be the wrong setting to discuss it, without collaborators in this area and outside of Wikipedia. If this format violates any of the requirements [and cannot be improved to fulfill those], it can (must) be rejected. No law exists that forces the German template vs. [cite-book](https://de.wikipedia.org/wiki/Vorlage:Cite_book), however German Wikipedia recommends it to users - as far as I am aware of. For some archeology into this, my suggestion is to search and read through [Vorlage_Diskussion:Literatur](https://de.wikipedia.org/wiki/Vorlage_Diskussion:Literatur) and also [Vorlage_Diskussion:Cite book](https://de.wikipedia.org/wiki/Vorlage_Diskussion:Cite_book). There, for example, an indication is that that nearly [~500k entries use/rely](https://templatecount.toolforge.org/index.php?lang=de&namespace=10&name=Literatur#bottom) on it vs. 12k using Cite book.

politikundbildung:
I'd like to add another vote for including the Style in the Style Repository. Even though I see how it is a somewhat odd hybrid between metadata export and citation style, I'd argue it is mostly used like the latter; You write a sentence in Wikipedia or add a quote, then quick copy the bibliographic information from Zotero using this style to provide a citation. Having this style in the style repository would make it available at zbib.org, which I think would be quite useful for editors who don't want to install Zotero or don't want to add any reference the quote in WP to their Zotero library.

192933488S:
"Closed with unmerged commits" ... it was merged or not?

PR #6865, opened 2024-01-23
Create sistema-de-bibliotecas-ufu-abnt-num.csl

victormpalma:
It creates the numeric version of Sistema de Bibliotecas UFU ABNT.

PR #6868, opened 2024-01-24
Create sistema-de-bibliotecas-ufu-abnt-num-sup.csl

victormpalma:
It creates numeric superscript version of Sistema de Bibliotecas UFU ABNT.


POBrien333:
See my comments from #6865

PR #6878, opened 2024-01-30
Update bourgogne-franche-comte-nature.csl

EstelleCamizuli:
Hello,
I have updated/created a new bourgogne-franche-comte-nature zotero style in order to match better the requested style for the paper. I hope this style would be cleared and will help the author to prepare better their article for publication in Bourgogne-Franche-Comté Nature review. 
Please let met know if something is missing or if I have not done follow the updated process wright.
Best Regards,
Estelle Camizuli


NicolasSouchon:
@EstelleCamizuli The error is due to the presence of unused macros (line 229-231: number; lines 364-366: isbn; lines 399-406: event). These macros should be deleted

adam3smith:
If you look at the example citations above, the webpage bibliography entry starts with a year. That's almost certainly not right -- the in text citation shows the title first, so presumably that should show in the bibliography. 
The easiest way to do this would be to replace the if/else-if logic of the `responsibility` macro with a substitute logic, but there are other ways to do this.

PR #6881, opened 2024-02-01
Update heiliger-dienst.csl

POBrien333:
https://forums.zotero.org/discussion/83084/style-request-heiliger-dienst-german#latest


adam3smith:
Merging this for now -- I think the different rules for page ranges with/without a series are possible using a `<if variable="collection">` test followed by completely separate logic, but let's get this off the queue first. Seems like a weird rule anyway

PR #6887, opened 2024-02-05
APA-CV Sorting

adam3smith:
Simplifying this fixes some weird behavior reported in https://forums.zotero.org/discussion/comment/454358/#Comment_454358

### DIFF
--- a/apa-annotated-bibliography.csl
+++ b/apa-annotated-bibliography.csl
@@ -202,6 +202,7 @@
     </choose>
   </macro>
   <macro name="date-bib">
+    <!-- TODO: Add new item types -->
     <group delimiter=" " prefix="(" suffix=")">
       <choose>
         <if is-uncertain-date="issued">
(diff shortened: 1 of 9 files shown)


adam3smith:
@bwiernik -- are you OK with this simple option? The sort using `date-bib` doesn't work correctly because citeproc gets confused by the date parts added separately for some item types and then sorts those item types first. Might be worth checking on for the general APA styles, too, though those would need a proper solution -- within a CV it probably doesn't matter if citations within a same year a sorted by date or by title.

bwiernik:
Hmm, we previously had a separate `date-sort` macro but removed it because of another issue and I couldn't remember why it was there. Let me compare against the manual and see.

bwiernik:
Hmm, so here was the issue that prompted changing from `variable='issued' form='numeric'` to the current behavior of the displayed bibliography date: https://forums.zotero.org/discussion/comment/434602#Comment_434602

The current PR changes from displayed bibliography date to `issued`, but not `form='numeric'`. Let me check if that avoids the previous odd behavior. Otherwise we might need to see about fixing this in citeproc-js.

bwiernik:
I checked, and the main APA has the same sorting problem. I don't remember this being the case previously -- I wonder if there was a change in citeproc-js? I'll check with how pandoc handles it.

image

bwiernik:
I will see if I can write a macro or set of macros that sidesteps the issue better.

bwiernik:
Hi @adam3smith I wrote some new sorting logic for the main APA style and then ported it over to APA CV.

One wrinkle with simple reverse chronological sorting for the CV style is that, within the same year, items with month/day information get sorted before items without it. This lead to web pages being sorted before articles, which seemed non-ideal for a CV.

Rather than revise the sorting logic extensively for the CV style, I opted to reverse sort on year, but then sort forward chronologically within years. For the purposes of the CV that seems fine and also honestly matches how many people build their CVs in my experience. 

What do you think? Could you test these out a bit and then I'll copy over to the other APA styles.

adam3smith:
Definitely looks good in the CV version, yes, haven't played with the vanilla APA yet, but I assume you tested that with a large enough set?

bwiernik:
Yep, sounds good

PR #6889, opened 2024-02-06
add DOI to american-political-science-association.csl

POBrien333:
https://forums.zotero.org/discussion/111573/need-updated-apsa-style-please#latest


bwiernik:
Thanks!

PR #6909, opened 2024-02-16
Create universidade-de-sao-paulo-ifsc-abnt-numerico.csl

sadisalvo:
I adpated an existing ABNT style to create one for USP-IFSC, using numeric citation instead of author-data. It would be useful for student to find the style by the name of institution, as most of the ABTN styles are author data and we use numeric.


bwiernik:
Thanks for the submission and your patience. 

There are some issues with the style, but I can help fix those. To clarify, is your intent to create a style that follows the standard ABNT formatting, but uses numeric in-text citations? Or are there other changes to standard ABNT as well (https://github.com/citation-style-language/styles/blob/master/associacao-brasileira-de-normas-tecnicas.csl)?

bwiernik:
@sadisalvo I'm going to go ahead and merge this now, please let me know if there are any issues

PR #6912, opened 2024-02-19
Update the-plant-cell.csl

POBrien333:
https://forums.zotero.org/discussion/110717/the-plant-cell-reference-format-update#latest

Public papers: https://academic.oup.com/plcell/issue/36/1

Nota bene: Guidelines don't explain their style, but style for article-journal is consistent in published papers. But I cannot find any other item types, minus webpages.


adam3smith:
Thanks!

PR #6920, opened 2024-02-25
Update food-and-agriculture-organization-of-the-united-nations.csl

swifterslb:
Correct some errors since the last revision.
Updated legal and database references according to the revised style guide.


adam3smith:
Thanks!

PR #6921, opened 2024-02-25
Update food-and-agriculture-organization-of-the-united-nations-numeri…

swifterslb:
Similar to #6920 

Correct some errors since the last revision.
Updated legal and database references according to the revised style guide.


adam3smith:
Thanks!

PR #6922, opened 2024-02-26
Update gait-and-posture.csl

Morgan435:
Updated maximum number of authors in bibliography (should be 6)


adam3smith:
This is more complicated for a dependent style. None of the existing ones
works?

Morgan435:
Well, if this is the style defined for the journal Gait & Posture, it does
not comply with the journal. I am an associate editor of the journal and
someone noticed the issue.
So, there might be another style that works but it is not the point. The
point is to make the style called the name of the journal compliant with
the requirements of the journal.

It seems the only problem is the limitation of listed authors.

Cheers,

*Morgan Sangeux*

*Profils Rg <https://www.researchgate.net/profile/Morgan_Sangeux>,
WebofScience <https://www.webofscience.com/wos/author/record/H-7100-2017>*

*tel: +33 6 65 73 25 42*

adam3smith:
superseded by #6923

PR #6923, opened 2024-02-26
Update gait-and-posture.csl

Morgan435:
Trying to update the style with limitation to the number of authors without reference to the dependent style to avoid issues.


POBrien333:
@Morgan435 
Could you please head here and click on "Sync fork"? Thanks
https://github.com/Morgan435/styles/tree/patch-3

PR #6926, opened 2024-02-27
Create BUAA Standard EN Format.csl

pepperedmutton:
citation format for Beihang University who claim to use GB 7714 but having their own interpretation of this standard


POBrien333:
@pepperedmutton 
Thank you for your patience.
Can you please adapt your style to fit our style requirements as outlined here? https://github.com/citation-style-language/styles/blob/master/STYLE_REQUIREMENTS.md
You removed a variety of lines from the <info> section that are necessary.

Can you also explain if the other style submission is just a localisation of this style? could they be merged and let the user select the language within Zotero?

PR #6927, opened 2024-02-27
Create BUAA format based on GB 7714.csl

pepperedmutton:
Modified GB 7714 for BUAA thesis


zepinglee:
We have already create a CSL style for the BUAA thesis format at <https://github.com/redleafnew/Chinese-STD-GB-T-7714-related-csl/blob/main/420beihang-university.csl> according to the specification at <http://graduate.buaa.edu.cn/info/1039/7831.htm>. It seems our implementation is more close to the format (see the differences at <https://github.com/redleafnew/Chinese-STD-GB-T-7714-related-csl/commit/6f6053b8c97d403fd32f4a37870a22b311897ba2>).

pepperedmutton:
Sorry I really need to apologize for that, I was under some very self contridicting instructions back then and that format was just enough for me to get by. I think it's very approperiate to remove my incomplete code. But I also need to remind you that the school does not alwasy stick to the GB it claims to follow, for example sometimes the type of media is omitted from the reference though it is required by the GB. You need to watch out for that when writting your thesis. Best wishes.

PR #6932, opened 2024-02-29
Update harvard-leeds-beckett-university.csl

Jenloo82:
Remove ampersand - 
I have replaced all 9 instances of and="symbol" with and="text". 

Remove hanging in Bibliography -
I have changed 247.<bibliography hanging-indent="true"> to 247.	<bibliography hanging-indent="false">


adam3smith:
Thanks!

PR #6935, opened 2024-03-05
Update utah-geological-survey.csl

inkenbrandt:
Change from "and others" to "et al." based on updated style guide C-137 (https://ugspub.nr.utah.gov/publications/circular/c-137.pdf).


inkenbrandt (commented):
Removal of et-al term

PR #6941, opened 2024-03-09
Update pontifical-biblical-institute.csl

POBrien333:
https://forums.zotero.org/discussion/112649/style-error-pontifical-biblical-institute#latest


adam3smith:
As the OP on the Zotero forums points out, the style is polyglot, and this will break for anything non-English. Could we redefine "at" to "here" for English and use the term (and then fix the other used locales if/as needed.
@bwiernik "here" would be a good term to add.

bwiernik:
How do we feel about adding terms and variables to CSL 1.0.2? I think it would be okay to do so.

bwiernik:
Sounds good. I'll put together a suggested list of terms and variables for 1.0.3

PR #6947, opened 2024-03-12
Create chicago-author-date-no-em-dash.csl

figgie10:
For authors and editors submitting manuscripts to a publisher who follows Chicago Manual 17th ed., 14.67: "Authors usually should not use the 3-em dash for repeated names in their manuscripts."

Based on
Chicago Manual of Style 17th edition (author-date)(2023-12-25 18:06:05)

Removed
subsequent-author-substitute="———"

Changed name of style
    <title>Chicago Manual of Style 17th edition (author-date, no em dash)</title>
    <id>http://www.zotero.org/styles/chicago-author-date-no-em-dash</id>
    <link href="http://www.zotero.org/styles/chicago-author-date-no-em-dash" rel="self"/>

### DIFF
--- a/dependent/chicago-author-date-no-em-dash.csl
+++ b/dependent/chicago-author-date-no-em-dash.csl
@@ -1,0 +1,684 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="display-and-sort" page-range-format="chicago">
+  <info>
+    <title>Chicago Manual of Style 17th edition (author-date, no em dash)</title>
+    <id>http://www.zotero.org/styles/chicago-author-date-no-em-dash</id>
+    <link href="http://www.zotero.org/styles/chicago-author-date-no-em-dash" rel="self"/>
+    <link href="http://www.chicagomanualofstyle.org/tools_citationguide.html" rel="documentation"/>
+    <author>
+      <name>Julian Onions</name>
+      <email>julian.onions@gmail.com</email>
+    </author>
+    <contributor>
+      <name>Sebastian Karcher</name>
+    </contributor>
+    <contributor>
+      <name>Richard Karnesky</name>
+      <email>karnesky+zotero@gmail.com</email>
+      <uri>http://arc.nucapt.northwestern.edu/Richard_Karnesky</uri>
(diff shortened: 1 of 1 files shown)


figgie10:
My apologies, but as a novice, I am not sure how to fix these errors. Can someone please advise?

adam3smith:
Let's figure out how we want to handle the larger issue -- we're not going to add a no-em-dash variant to every Chicago style. I'm happy to add one, _maybe_ two, but certainly not 10 (the number of CMoS 17th ed styles we have).
Open for suggestions here -- I think the benefits of adding the styles, allowing people to create print-ready manuscripts, outweigh the downsides, so I'd like to keep them in in general, but I'm aware of that section in the Manual, of course.

figgie10:
Thank you so very much for this! 

Yes, I agree that 10 is out of the question. Why not make just two for the two Chicago referencing styles? The formatting of lists in these systems differs, so in Chicago, we would need at least these two.

1. Notes-bibliography (Bibliography)
Chicago Manual of Style 17th ed. 14.67: "Authors usually should not use the 3-em dash for repeated names in their manuscripts." 
Zotero style: [Chicago Manual of Style 17th edition (note)](https://www.zotero.org/styles/chicago-note-bibliography)(2023-12-10 00:40:26)

2. Author-date (Reference List)
Chicago Manual of Style 17th ed. 15.67: "Authors usually should not use the 3-em dash for repeated names in their manuscripts." 
Zotero style: [Chicago Manual of Style 17th edition (author-date)](https://www.zotero.org/styles/chicago-author-date)(2023-12-25 18:06:05)

This way, authors can generate the bibliography or reference list without the 3-em dash for repeated names. 

Very grateful!

adam3smith:
yes, this makes sense to me, but let's use the full note one, which I think is the most widely used option, instead of the note.

figgie10:
Sounds like a plan. The criteria should surely be "most widely used styles" for notes-bibliography, author-date.

Thank you again. I know many wo wil be happy to see them here. :)

figgie10:
FYI, I used a different style for my (failed) version of notes-bibliography. 

[Chicago Manual of Style 17th edition (full note, short title subsequent)](https://www.zotero.org/styles/chicago-fullnote-bibliography-short-title-subsequent)(2023-12-10 00:40:26)

PR #6951, opened 2024-03-13
Update uppsala-university-library-harvard.csl

maria-b-a:
added name of director in source type film, added URL in some source types, made some changes in bill, legislation etc.


adam3smith:
Thanks!

PR #6956, opened 2024-03-14
apsa - remove subsequent author em dashes

aj-guerra:
"Avoid the use of ibid., idem, f., ff., op. cit. and loc. cit. (14.34–14.36). While these Latin abbreviations have been used for years in academia, they do not translate well, if at all, to online publication formats. Further, replacing the name in successive references of the same author(s) with three em-dashes (———) is no longer recommended by Chicago. It complicates the digitization process, does not align with online database sorting, and makes the source impractical outside of the context of the reference list (14.67)."

https://connect.apsanet.org/stylemanual/sample-page/references/

### DIFF
--- a/american-political-science-association.csl
+++ b/american-political-science-association.csl
@@ -16,6 +16,9 @@
     <contributor>
       <name>Patrick O'Brien</name>
     </contributor>
+    <contributor>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks!

PR #6958, opened 2024-03-15
Update norsk-apa-manual.csl

kbernt:
Updates the macro publication-history so that it uses the term origina-work-published, and provided translations into bokmål and nynorsk. Previously it used English ("Original work published").


adam3smith:
Great, thanks!

PR #6959, opened 2024-03-15
Update art-history.csl

POBrien333:
https://forums.zotero.org/discussion/112845/errors-in-art-history-style-sheet#latest


adam3smith:
Thanks!

PR #6961, opened 2024-03-17
Add no em-dash variants for two main Chicago styles

adam3smith:
See https://github.com/citation-style-language/styles/pull/6947


adam3smith (commented):
Closes #6946
closes #6947

PR #6963, opened 2024-03-18
Update zeitschrift-fur-theologie-und-kirche.csl

POBrien333:
https://forums.zotero.org/discussion/110268/style-error-zeitschrift-fuer-theologie-und-kirche-deutsch#latest


bwiernik:
@POBrien333 This one isn't producing any citation output for book or website items (see the example output above), and the bibliography formatting for books needs some delimiter between the title and editor. Should editor also be substituted for author when there is no author?

bwiernik:
Let's do these changes,

1. Substitute editor then title for author in citations and bibliography
2. Use the existing formats for newspaper and magazine in addition to journal articles and for entry, dictionary, and encyclopedia in addition to chapter items
3. For other item types, use this format

AUTHOR, Title. Publisher. [Series], Issued-Year. Locator.

PR #6968, opened 2024-03-20
Create elsevier-vancouver-et-al.csl

POBrien333:
This PR is to make a new version of elsevier-vancouver.csl with different et-al rules. I wasn't fully sure about the naming.

It's for this journal that would need this style: https://forums.zotero.org/discussion/112758/journal-pediatric-surgery-needs-updating#latest


bwiernik:
What are the et al rules for the existing Elsevier Vancouver? Any other journals have this style that you know of?

adam3smith:
The existing Vancouver is the standard NLM 7/6 et al -- 4/3 is generally not uncommon for Vancouver-type styles, but I'm not aware of any other Elsevier journal using it and a google search for that exact phrasing ("For references with more than 3 authors, only the first 3 should be listed, followed by 'et al.'") only yields _JPS_

POBrien333:
I've made this an independent style and also added JPS to the elsevier skip list (via another open PR for Elsevier).

PR #6969, opened 2024-03-21
Create openspeaks.csl

psubhashish:
See documentation at https://github.com/ofdn/OpenSpeaks-Before-AI/tree/main/Citation%20Style

### DIFF
--- a/openspeaks.csl
+++ b/openspeaks.csl
@@ -1,0 +1,342 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" initialize-with-hyphen="false" page-range-format="minimal">
+  <!-- This style was edited with the Visual CSL Editor (https://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>OpenSpeaks</title>
+    <id>http://www.zotero.org/styles/openspeaks</id>
+    <link href="http://www.zotero.org/styles/openspeaks" rel="self"></link>
+    <link href="https://github.com/ofdn/OpenSpeaks-Before-AIements.html" rel="documentation"/>
+    <author>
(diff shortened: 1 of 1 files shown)


openspeaks.csl
```diff
@@ -0,0 +1,342 @@
...
+  <locale xml:lang="fr">
+    <date form="text" delimiter=" ">
+      <date-part name="day"/>
+      <date-part name="month" form="short" strip-periods="true"/>
+      <date-part name="year"/>
+    </date>
+  </locale>
+  <macro name="author">
+    <names variable="author director" font-variant="normal" display="left-margin">
+      <name font-weight="normal" and="text" delimiter-precedes-last="always" et-al-min="3" initialize="false" initialize-with="" name-as-sort-order="all"/>
```
bwiernik:
```suggestion
      <name font-weight="normal" and="text" delimiter-precedes-last="always" et-al-min="3" et-al-use-first="1" initialize="false" initialize-with="" name-as-sort-order="all"/>
```

openspeaks.csl
```diff
@@ -0,0 +1,342 @@
...
+  </macro>
+  <citation collapse="citation-number">
+    <sort>
+      <key variable="citation-number"/>
+    </sort>
+    <layout prefix="(" suffix=")" delimiter=",">
+      <text variable="citation-number"/>
+    </layout>
+  </citation>
+  <bibliography et-al-min="3" et-al-use-first="3" second-field-align="flush">
```
bwiernik:
```suggestion
  <bibliography et-al-min="4" et-al-use-first="3" second-field-align="flush">
```

openspeaks.csl
```diff
@@ -0,0 +1,342 @@
...
+    <layout>
+      <text macro="title" quotes="false"/>
+      <group delimiter=". " prefix="— by " suffix=". ">
```
bwiernik, line 319: Throughout the style, please use delimiters on groups, rather than suffixes, to hold spaces and punctuation between elements.

PR #6972, opened 2024-03-21
Update pnas.csl for missing / in https address

pascaloettli:
A / was missing in the https address in line 70


bwiernik:
Thanks for catching that!

PR #6979, opened 2024-03-26
Update pnas.csl

POBrien333:
https://forums.zotero.org/discussion/comment/459248/#Comment_459248


bwiernik:
Looks like you inadvertently lost the delimiter for the book chapter  before the page numbers

POBrien333:
Hmm. In the editor it was duplicated liek ", , p.". Weird.

PR #6983, opened 2024-03-27
Update ieee.csl to incl. video directors | add standard type

POBrien333:
https://forums.zotero.org/discussion/113145/ieee-online-video-e-g-youtube#latest

I feel like we should include some other names as well. Any suggestions which?

-----
Also adds the "standard" type to style:
image

image
Requested here: https://forums.zotero.org/discussion/115455/zotero-not-compliant-with-the-ieee-style-guide/p1

### DIFF
--- a/ieee.csl
+++ b/ieee.csl
@@ -155,7 +155,7 @@
   </macro>
   <macro name="title">
     <choose>
-      <if type="bill book graphic legal_case legislation motion_picture song" match="any">
+      <if type="bill book graphic legal_case legislation motion_picture song standard" match="any">
         <text variable="title" font-style="italic"/>
       </if>
       <else>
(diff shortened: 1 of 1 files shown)


zepinglee:
I've collected all the sample entries of the IEEE manual into Zotero format (via <https://anystyle.io/>) and they may help testing CSL styles: <https://gist.github.com/zepinglee/f35d04cbbc240ad0de21de79818761cf>.

bwiernik:
@POBrien333 Should we go ahead and merge this now and can add more in a future PR?

POBrien333:
Yeah, I'd say so. Let's get that out to the users. :)

POBrien333:
Is the type "preprint" not a valid item type (yet)?

bwiernik:
Preprint is type "article"

PR #6994, opened 2024-04-01
Update universiteit-utrecht-onderzoeksgids-geschiedenis.csl

Sjorsschaap:
English version


POBrien333:
@Sjorsschaap 
Can you say what you want to achieve?
Currently, your PR makes the NL style into an ENG style. I'd assume you'd want both variants?

Sjorsschaap:
Ideally, yes. What this sets out to actieve is to allow for the use of this
citation style in English.

Op di 29 jul 2025, 16:52 schreef POBrien333 ***@***.***>:

POBrien333:
Right, but with your approach you'd remove the ability to have the style in NL.
I've removed the default-locale. That way the user has the choice. You can see how APA or Chicago let's a user set a language for a document, while this style was hardcoded into NL so far.

Can you go here and click on "Sync fork"?
https://github.com/Sjorsschaap/styles/tree/patch-2

PR #6995, opened 2024-04-01
Create eum-proposte-e-ricerche.csl

pcavicch:
This is the CSL style used by the magazine "Proposte e Ricerche" of Edizioni Universitarie Maceratesi (EUM). 
The Style is used by students of Humanities faculties to write their Theses.

### DIFF
--- a/eum-proposte-e-ricerche.csl
+++ b/eum-proposte-e-ricerche.csl
@@ -1,0 +1,438 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="never" default-locale="it-IT">
+  <info>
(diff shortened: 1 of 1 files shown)


POBrien333:
@pcavicch 
Thank you for your submission and patience.
Can you explain which style you based this style on as well as provide a link to the style's guidelines?

pcavicch:
Hi, thanks for your reply, but I figured it out. Don’t worry about it
anymore, it wasn’t that important.



*--*
*Paolo Cavicchini*


Il Lun 8 Set 2025, 11:35 POBrien333 ***@***.***> ha scritto:

POBrien333:
That's not the point.
We want to accept this style so other people can make use of the style that you generated. However, we want to give credit to whoever made the style that you based it on and therefore want to include that in the style's metadata.
We'd also want a link for the documentation of this style to verify it. Thank you for providing these.

Issue #6988, opened 2024-03-29
Requesting style: Journal of Microscopy


mdoube:
In-text citation:
Superscript,<sup>1</sup> numbered,<sup>2</sup> after punctuation.<sup>1,2</sup>

Bibliography: (looks a lot like APA v7)
1. Campbell, J. L., & Pedersen, O. K. (2007). The varieties of capitalism and hybrid success. _Comparative Political Studies_, _40_(3), 307-332. https://doi.org/0.1170010414006286542
2. Mares, I. (2001). Firms and the welfare state: When, why, and how does social policy matter to employers? In: P. A. Hall & D. Soskice (Eds.), _Varieties of capitalism. The institutional foundations of comparative advantage_ (pp. 184-213). New York, NY: Oxford University Press.

Author list is truncated after 19 names, followed by an ellipsis (...) and the last author's name. 

Recent open-access publication here: https://doi.org/10.1111/jmi.13288

Guidelines here: https://onlinelibrary.wiley.com/page/journal/13652818/homepage/forauthors.html

>References should be numbered in the text, with numbers placed at the end of the sentence in superscript e.g.: ...example of reference format1.
>
>The reference list should:
>
>• Be numbered and listed in order the reference appears in the paper.  
>• (where a journal is cited) - include authors' surnames and initials, date of publication, title of paper, name of journal in full, volume number (and issue number where volume pages do not run on), and first and last page numbers.  
>• (where a book is cited) - include authors' surnames and initials, title of book, year of publication, edition if relevant, followed by the publisher and town, county/state (and country if necessary) of publication.
>
>Examples of reference style:
>1. Ashford, A.E. (1998) Dynamic pleiomorphic vacuole systems: are they endosomes and transport compartments in fungal hyphae? Adv. Bot. Res. 28, 119-159.
>2. Causton, B. (1984) The choice of resins for electron immunocytochemistry. Immunolabelling for Electron Microscopy (ed. by J. M. Polak and I. M. Varndell), pp. 29-36. Elsevier, Amsterdam.
>3. Muller, C. (1966) Spherical Harmonics. Springer-Verlag, Berlin.
>4. Fischer-Parton, S. (1999) Role of pH, calcium and vesicle trafficking in regulating hyphal tip growth of Neurospora crassa. PhD Thesis, University of Edinburgh, Edinburgh.

PR #6996, opened 2024-04-02
Create journal-of-microscopy.csl

POBrien333:
closes https://github.com/citation-style-language/styles/issues/6988

Guidelines not in sync with actual papers.

### DIFF
--- a/journal-of-microscopy.csl
+++ b/journal-of-microscopy.csl
@@ -3,7 +3,7 @@
   <info>
     <title>Journal of Microscopy</title>
     <id>http://www.zotero.org/styles/journal-of-microscopy</id>
-    <link rel="self" href="http://www.zotero.org/styles/journal-of-microscopy"/>
+    <link href="http://www.zotero.org/styles/journal-of-microscopy" rel="self"/>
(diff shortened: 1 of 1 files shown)


bwiernik:
@POBrien333 Missing a delimiter between publisher and publisher-place

POBrien333:
I think that's an error of the generated examples in here.
Works fine and is set:
image

adam3smith:
(reminder that citeproc-ruby handles delimiters that involve groups and conditionals incorrectly.   We should try to get that fixed, though Sylvester is unfortunately not developing anymore

bwiernik:
@adam3smith @retorquere Sylvester pushed a fix to citeproc-ruby at RubyGems. Is there anything we need to do to update the repo to the latest version of the Gem?

retorquere:
Sheldon still pulls in 2.0. Easy to fix but I'd have to set up Ruby again -- if someone has a working ruby env, changing the gemspec file and running bundle install will get the lockfile updated, then check the lot in -- this repo pulls sheldon from github at every run so there's no separate publishing step.

bwiernik:
I think I have Ruby set up -- I'll check it out this weekend

PR #7004, opened 2024-04-04
Fork bibtex style to create more barebones biber style

nathdwek:
I have been using this more barebones biber style for a while, and find it easier to work with on the day to day, using a text-editor for bib files.

Summary of changes:
* Remove a few "typically verbose" attributes that are not used anyway by most common citation-styles: `abstractnote`, `annote`, `keywords`, `note`
* One attribute per line (LF line endings)
* `date` attribute as YYYY-MM-DD instead of `year` and `month` attribute 
* Add `@dataset` entry type
* Remove `language` attribute (maybe to discuss?)
* If there is a DOI, export only `DOI` attribute and not the `URL` attribute. (maybe to discuss?)

To sum up, make the resulting  .bib file easier to work with, mostly by removing attributes that are very rarely actually used by the citation styles, and by adding linebreaks between attributes. In addition, some biber specific updates: new dataset entry type, and date as a numeric rather than month and year as text.

Finally some cosmetic preferences that change the typical bibliography output: remove the language attribute, and print DOI and no URL is DOI is available, otherwise fall back on URL. Those can be discussed.

### DIFF
--- a/biber.csl
+++ b/biber.csl
@@ -1,0 +1,184 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>BibTeX generic citation style</title>
+    <id>http://www.zotero.org/styles/bibtex</id>
+    <link href="http://www.zotero.org/styles/bibtex" rel="self"/>
+    <link href="http://www.bibtex.org/" rel="documentation"/>
+    <author>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks! I really don't want to proliferate bibtex output from CSL, though -- CSL is not well suited to generate metadata output, and the bibtex.csl that we have already is a source of confusion. Reasonable reference managers should all have proper bibtex output. 
You're obviously welcome to use this yourself and publish it via your own github, just like many of us keep repositories of custom CSL styles that don't meet repository criteria, but we'll not take this for the repo.

PR #7007, opened 2024-04-08
Update mhra.csl to 4th edition

POBrien333:
…s-research-association-3rd-edition.csl

https://forums.zotero.org/discussion/113436/mhra-4th-edition#latest

Changes for version 4: https://www.mhra.org.uk/style/changes.html

- [ ] Adapt author-date style


bwiernik:
Thanks!

(And why oh why would an organization start to require DOIs, but mandate the 10-year-outdated format?)

PR #7010, opened 2024-04-09
CMoS: sort bibliography by only year part (discourse1843)

zepinglee:
Discussed in <https://discourse.citationstyles.org/t/multiple-works-by-the-same-author-published-in-the-same-year-ordered-incorrectly-per-chicago-style/1843>.

I only modified styles of CMoS 17th edition since I don't have access to earlier versions.


adam3smith:
Thanks!

zepinglee:
Great!

PR #7013, opened 2024-04-10
Create national-library-of-medicine-grant-proposals-all-names-doi-pmi…

masterrite:
There are people who prefer to list all of DOI, PMID, and PMCID at the end of each reference entry. NIH grants also forbids URL in bibliography, and some papers would show up URLs as source.

This is based on national-library-of-medicine-grant-proposals. I'm also happy to have this merged to that one, but I would appreciate helps (this is my first time doing GitHub PRs and creating Zotero styles)

### DIFF
--- a/national-library-of-medicine-grant-proposals-all-names-doi-pmid-pmcid
+++ b/national-library-of-medicine-grant-proposals-all-names-doi-pmid-pmcid
@@ -1,0 +1,259 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" initialize-with-hyphen="false" default-locale="en-US">
+  <info>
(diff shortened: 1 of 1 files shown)


POBrien333:
@adam3smith 
As we had a similar discussion on the Forums the other day, what's your take on this style?

adam3smith:
I'm skeptical. NIH specifically doesn't require a specific citation style.
I think we should have one that meets their requirements, which we do, but
beyond that, we should leave customization to individuals.
This isn't set in stone, so if we get lots of demand for a very specific
variant I'm happy to revisit, but there is an unlimited number of
possibilities

bwiernik:
We have heard a few comments now about NIH having a problem with URLs in the bibliography, so I think we should probably remove URL from the version we have in the repository (and format DOI not as a URL)

adam3smith:
Yes, agree with that, NIH is very strict on the URL

PR #7017, opened 2024-04-16
Create docteurimago

vetbenj:
We would like to add this CSL style for ours publications Docteur Imago and Tech Imago. BOM Presse is a publisher in french in the field of radiology.


bwiernik:
Thank you for your submission and patience! Could you say what differences there are between your style and the standard ISO 690-2 style?

PR #7027, opened 2024-04-25
Create royal-entomological-society.csl

POBrien333:
https://forums.zotero.org/discussion/113977/style-request-insect-conservation-and-diversity#latest


adam3smith:
I think we should just create this as royal society for entemology with dependents instead. https://resjournals.onlinelibrary.wiley.com/

POBrien333:
Good point. That's changed. I'll work on the depedents as well

adam3smith:
Thanks!

PR #7031, opened 2024-04-30
Update american-political-science-review.csl

POBrien333:
https://forums.zotero.org/discussion/comment/461922/#Comment_461922


adam3smith:
Thanks!

PR #7034, opened 2024-05-02
Update utah-geological-survey.csl

inkenbrandt:
Removed "and others" to reflect changes in the new UGS Style Guide C-137. https://doi.org/10.34191/C-137


adam3smith:
Thanks!

PR #7037, opened 2024-05-05
Update archaeonautica.csl

samarkar:
I made some changes to this style but only just realized that they were not taken into account in the Zotero repository. I'm not very comfortable with Github and I certainly made some mistakes, I apologize for that.

### DIFF
--- a/archaeonautica.csl
+++ b/archaeonautica.csl
@@ -17,7 +17,7 @@
     <issn>0154-1854</issn>
     <eissn>2117-6973</eissn>
     <summary>Style conçu pour la revue Archaeonautica. L’archéologie maritime et navale de la préhistoire à l’époque contemporaine. Centre Camille Jullian, CNRS</summary>
-    <updated>2022-03-28T00:32:09+00:00</updated>
+    <updated>2024-05-05T00:32:09+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="fr">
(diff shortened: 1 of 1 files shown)


archaeonautica.csl
```diff
@@ -216,11 +218,20 @@
     </choose>
   </macro>
   <macro name="translator">
+    <text term="translator" form="verb" prefix=", " suffix=" "/>
     <names variable="translator">
-      <name form="long" and="text" delimiter-precedes-last="never" sort-separator=" ">
+      <name form="long" and="text" delimiter-precedes-last="never" sort-separator=" " font-style="normal">
+        <name-part name="family" font-variant="small-caps"/>
+      </name>
```
bwiernik, line 225:
```suggestion
      <label form="verb" suffix=" "/>
      <name form="long" and="text" delimiter-precedes-last="never" sort-separator=" " font-style="normal">
        <name-part name="family" font-variant="small-caps"/>
      </name>
```

archaeonautica.csl
```diff
@@ -216,11 +218,20 @@
     </choose>
   </macro>
   <macro name="translator">
+    <text term="translator" form="verb" prefix=", " suffix=" "/>
```
bwiernik, line 221:
```suggestion
```

bwiernik (changes requested):
Thanks! For the creator role labels, please use `<label>` inside of `<names>`, rather than `<text term="..."/>` to avoid the terms being printed when those creators are not present (see the example website citation in the message above).

POBrien333:
@samarkar 
Could you please go [here](https://github.com/samarkar/styles/tree/patch-11) and click on "Sync fork"? Thanks

PR #7040, opened 2024-05-08
Create a dedicated Spanish APA file

mirenbz:
Spanish requires the order of some elements to be changed (date, specifically). This is not possible to localize and have in a single file, as I was explained [in the official forums](https://forums.zotero.org/discussion/102287/where-is-the-p-pp-prefix-located-in-apa-7th-edition). That's why I'm proposing this dedicated APA file for Spanish that fixes the order of the date elements that are currently wrong for Spanish.

### DIFF
--- a/apa-es.csl
+++ b/apa-es.csl
@@ -169,8 +169,8 @@
                 <choose>
                   <if variable="collection-editor editor editorial-director issue page volume" match="none">
                     <date variable="issued">
-                  <date-part prefix="" name="day" suffix=" de "/>
-                  <date-part prefix="" name="month" suffix=" de "/>
+                      <date-part prefix="" name="day" suffix=" de "/>
+                      <date-part prefix="" name="month" suffix=" de "/>
                     </date>
                   </if>
                 </choose>
--- a/apa-es.csl
+++ b/apa-es.csl
@@ -949,7 +949,7 @@
       </else>
     </choose>
   </macro>
-<macro name="volume-booklike">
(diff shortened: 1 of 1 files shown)


adam3smith:
I'm skeptical about creating all these different language versions. We already have the no ampersand version. The date in the APA bibliography is _not_ in regular form: (2024, May 9) also isn't how you'd write a date in English. Because the bibliography sorts by date, date-bib definitely needs to start with the year. There's an argument to be had whether it should be (2024, 9 de mayo) in Spanish, but there's also a case to be made for "mayo 9" which, while un-idiomatic/grammatical, preserves sort order (year, month, day).
We'll need to make month-day localizable at some point, but I really don't want to add a separate APA style for every language because of that -- debatable -- difference.

mirenbz:
I agree, actually, in that avoiding multiple files in desirable. I just proposed this because the examples I came across were using "natural" date orders (like [here](https://normas-apa.org/wp-content/uploads/Guia-Normas-APA-7ma-edicion.pdf) or [here](https://normas-apa.org/referencias/citar-periodicos/)), but your comment makes full sense and I have also found examples of that use in Spanish. 

Whatever you decide on this PR is fine by me; keep it to come back to the discussion in the future, or get rid of it completely (I can use my local version should anyone require "natural" language dates). 

Thanks for the insight!

adam3smith:
Closing this for now then -- it'll still remain searchable so the discussion isn't lost

PR #7045, opened 2024-05-10
Update journal-of-hazardous-materials.csl

RisenL:
update style for Journal of Hazardous Materials


POBrien333:
Continued here https://github.com/citation-style-language/styles/pull/7660

PR #7046, opened 2024-05-12
feat: Update jacc-cardiovascular-imaging.csl to current (2024) requirements.

laqua-stack:
Since the 2015 version, JACC: Cardiovascular Imaging changed its style from Vancouver to AMA.


POBrien333:
@laqua-stack 
Thank you for your patience.
Could you head [here](https://github.com/laqua-stack/styles/tree/master) and click on sync fork? That will remove the validation errors and then we're good to merge.

PR #7055, opened 2024-05-16
Update sciences-po-ecole-doctorale-note-french.csl

AnnaCouthuresIdrizi:


sciences-po-ecole-doctorale-note-french.csl
```diff
@@ -393,7 +393,7 @@
               </else>
             </choose>
             <text term="cited" font-style="italic" suffix="."/>
-            <text variable="locator" prefix="p.&#160;"/>
+            <text variable="locator" prefix="p.&#160;" suffix="."/>
```
adam3smith, line 396: This seems odd -- why do you want a period at the end only with a locator but not otherwise? This way, it'd also appear before the semicolon when multiple items are cited, is that right? It would be more common to set a `suffix="."` on `layout`

AnnaCouthuresIdrizi: Dear Sebastian,
Maybe this seems odd, but this is exactly how my institution want it..
Could you please validate this pull request ?
Best regards,
Anna

AnnaCouthuresIdrizi:
Hello,
Could you please validate this pull request before the end of July ?
Best regards,
Anna

AnnaCouthuresIdrizi:
Hello,
Please merge this request before students are coming back, please please please !
Best regards,
Anna

PR #7058, opened 2024-05-17
Update unified-style-sheet-for-linguistics-de-gruyter-literature.csl

POBrien333:
https://forums.zotero.org/discussion/107988/unified-style-sheet-for-linguistics-multiple-works-by-the-same-author


adam3smith:
Thanks!

PR #7069, opened 2024-05-31
Update council-of-science-editors.csl to Version 9

POBrien333:
https://forums.zotero.org/discussion/114791/style-request-council-of-science-editors-9th-edition-citation-sequence#latest


adam3smith:
We often keep at least one edition back, what do you think? (If we're doing this, we'd want to add 9th edition to this style's title -- and update the summary) and add the old style with 8th edition in both title and ID.

POBrien333:
Great, that's done.
Let me know if it's all correct and I'll propagate that to the other CSE variants we have.

PR #7071, opened 2024-06-05
Update journal-of-management.csl to 8/31/2017 guidelines.

POBrien333:
https://forums.zotero.org/discussion/114930/replacing-p-xx-with-xx-for-in-text-citations-in-journal-of-management-style#latest


adam3smith:
Something went wrong here -- you removed the page numbers for journal articles in the bibliography; not seeing immediately where, but double-checked in Zotero (and in the style guide that they should be there, which they should)

POBrien333:
Fixed now.

PR #7075, opened 2024-06-06

anfourny:
…al-et-europeen.csl

Changes made from feedback received from users. Most notably, fixed erroneous extra commas and updated link to the documentation.

### DIFF
--- a/uclouvain-centre-charles-de-visscher-pour-le-droit-international-et-europeen.csl
+++ b/uclouvain-centre-charles-de-visscher-pour-le-droit-international-et-europeen.csl
@@ -187,14 +187,7 @@
               <group delimiter=", ">
                 <text variable="authority"/>
                 <date form="text" variable="issued"/>
-                <choose>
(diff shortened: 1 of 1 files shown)


uclouvain-centre-charles-de-visscher-pour-le-droit-international-et-europeen.csl
```diff
@@ -187,7 +187,7 @@
               <group delimiter=", ">
                 <text variable="authority"/>
                 <date form="text" variable="issued"/>
-                <text variable="title" font-style="italic"/>
+                <text variable="title" form="short" font-style="italic"/>
```
adam3smith, line 190: @anfourny  note that this does exactly the same as the much more complicate choose/if/else conditional you added.

anfourny: Thanks :)

PR #7076, opened 2024-06-06
Create RHS-MHort-Harvard-Style.csl

anthortic:
The Royal Horticultural Society (RHS) in house Harvard style for the Master of Horticulture course.


POBrien333:
@anthortic 
Thank you for your patience.
Can you please go through your style and fix the info section according to our requirements? It's probably easiest to copy it from a different style and only adapt the parts you need to change. You've removed a variety of importants things.
See: https://github.com/citation-style-language/styles/blob/master/STYLE_REQUIREMENTS.md

PR #7090, opened 2024-06-23
Update journal-of-threatened-taxa.csl

xbais:
Subsequent author names (those after the first author name in any given reference)  in Bibliography for JoTT journal should have name followed by surname. Currently this is the reverse way. The new update fixes this.


xbais:
PR updated : https://github.com/citation-style-language/styles/pull/7093

PR #7091, opened 2024-06-23
Update journal-of-threatened-taxa.csl

xbais:
Subsequent author names in Bibliography (after the fist author name) should have name followed by surname. Currently this is the reverse way. Also, a comma followed by a space (", ") should precede the year in in-text citations. This CSL file is thus being updated.


xbais:
PR has been updated : https://github.com/citation-style-language/styles/pull/7093

PR #7102, opened 2024-06-28
Update glossa.csl author order and locale

lisalevinson:
add name-as-sort-order="all" to all author/editor categories as requested by Glossa editor to fix name ordering errors with edited volumes. change default locale to US.


bwiernik:
Thanks!

PR #7111, opened 2024-07-08
Numeric in-text citations for Social Science and Medicine and SSM-Population Health

kkellysci:
Updates dependent styles for the journals "Social Science and Medicine" and "Social Science and Medicine - Population Health" to use the Elsevier citation format with numeric in-text citations (eg. [1] [2,3]).

In their "Guide for authors" pages, both journals say "Indicate references by adding a number within square brackets in the text." but the current repository styles for both journals use (Author, year) instead.

Relevant sections in the "Guide for authors" pages:
https://www.sciencedirect.com/journal/social-science-and-medicine/publish/guide-for-authors#toc-49
https://www.sciencedirect.com/journal/ssm-population-health/publish/guide-for-authors#toc-53

### DIFF
--- a/dependent/social-science-and-medicine.csl
+++ b/dependent/social-science-and-medicine.csl
@@ -5,7 +5,7 @@
     <title>Social Science &amp; Medicine</title>
     <id>http://www.zotero.org/styles/social-science-and-medicine</id>
     <link href="http://www.zotero.org/styles/social-science-and-medicine" rel="self"/>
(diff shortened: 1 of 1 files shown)


POBrien333:
You're in the wrong place for that.
The dependent styles are generated over here and you can modify it there:
https://github.com/citation-style-language/journals/blob/master/elsevier/_journals.tab

PR #7113, opened 2024-07-09
Updates from ISO 690 2022 specification (only some) 

### LINKED ISSUE #7112: ČSN ISO 690 2022 czech citation style update
According to new ČSN ISO 690 norm specification from 2022 (https://www.citace.com/Vyklad-CSN-ISO-690-2022.pdf, https://www.iso690.zcu.cz/), some changes need to be made.

The current https://github.com/citation-style-language/styles/blob/b2be5aeeee7f00fd2032ac1daad995bbe95398cf/iso690-author-date-cs.csl CSL file is from 2011 ISO 690 specification, which is outdated.

corovcam:
Some fixes according to ISO 690 norm specification from 2022. Resolves #7112.

### DIFF
--- a/iso690-author-date-cs.csl
+++ b/iso690-author-date-cs.csl
@@ -16,8 +16,8 @@
     </author>
     <category citation-format="author-date"/>
     <category field="generic-base"/>
-    <summary>Style based on ČSN ISO 690:2011</summary>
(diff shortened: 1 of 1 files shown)


libora:
Deleting line 467 destroyed the bibliometric listing of some types of resources!!! For example, books, electronic resources.

PR #7132, opened 2024-07-18
Create juristhese-1.0

CelianGdfrd:
Creation of a new citation style for legal thesis in France

### DIFF
--- a/juristhese.csl
+++ b/juristhese.csl
@@ -64,7 +64,7 @@
         <text variable="title" font-style="italic" suffix=", "/>
       </else-if>
       <else-if type="article-magazine article-newspaper" match="any">
-        <text variable="title" prefix="«&#160;" suffix=" », "/>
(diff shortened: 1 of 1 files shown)


juristhese.csl
```diff
@@ -0,0 +1,174 @@
...
+          <text macro="auteur"/>
+          <text macro="titre"/>
+          <text macro="edition"/>
+        </else>
+      </choose>
+    </layout>
+  </citation>
+  <bibliography>
+    <layout>
+      <date form="text" date-parts="year-month-day" variable="issued"/>
```
POBrien333, line 171: this seems odd. 
If you don't have a bibliography I'd delete the whole block from lines 169-173, instead of just printing a date.

CelianGdfrd: My bad, it was supposed to be fixed ; it is now.

PR #7138, opened 2024-07-24
Temporarily fix missing delimiter when both writer(s) and director(s) are listed

rkopicki:
To comply with Section 10.12, the rule must support dual roles (Writer & Director) and separately list each writer and director with the appropriate role, i.e. (Writer), (Director). This may involve adding a new term "script-writer-director", similar to "editor-translator".

### DIFF
--- a/apa-annotated-bibliography.csl
+++ b/apa-annotated-bibliography.csl
@@ -185,8 +185,11 @@
               </choose>
             </if>
           </choose>
-          <!-- TODO: Replace `delimiter` with `collapse` to combine names variables when that becomes available. -->
-          <names variable="series-creator executive-producer" delimiter=", &amp; ">
+          <names variable="executive-producer">
+            <name name-as-sort-order="all" and="symbol" sort-separator=", " initialize-with=". " delimiter=", " delimiter-precedes-last="always"/>
+            <label form="long" prefix=" (" suffix=")" text-case="title"/>
+          </names>
+          <names variable="series-creator">
             <name name-as-sort-order="all" and="symbol" sort-separator=", " initialize-with=". " delimiter=", " delimiter-precedes-last="always"/>
             <label form="long" prefix=" (" suffix=")" text-case="title"/>
           </names>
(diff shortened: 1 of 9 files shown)


rkopicki (commented):
`delimiter=", &amp; "` creates too many (3) ampersands, so I believe `delimiter=", "` is a better choice for now.
Additionally, the guest and host roles are unlikely to combine (`collapse`) because the two roles appear to be mutually exclusive.
Returning to script-writer(s) and director(s), APA 7 Section 10.12 Audiovisual Works (p. 343) instructs to "Include the contributor roles in parentheses after each contributor's name," so individual writers and directors should not be grouped together.

bwiernik:
Current CSL cannot collapse multiple distinct name variables, so we can only get close to fully-correct APA style here. It is likely that a future version of CSL will include the ability to render multiple names variables as one element. That is, something like this "Glass, I. (Host), & Daisey, M. (Guest)". That is what the `collapse` comment refers to (see https://github.com/citation-style-language/schema/issues/442). If this is implemented, it will include the ability to also combine names that are identical (e.g., "Smith, J. (Writer & Director), & Jones, J. (Writer)")

Whenever there are multiple people in a role (e.g., several writers) and multiple roles (e.g., writer and director), then whatever delimiter choice is made will have the wrong number of ampersands (there will be one at the end of each role list). Specifying the delimiter to be `, & ` will at least give the correct number for somewhat common case that there is one person in each of the 2 roles (if there is one writer and director or if there is one host and guest). (The role labels are also combined at the end of the list for the role `(Writers)` rather than after each name.)

It's not a perfect solution, but it is the best that is currently possible with CSL.

A hack to get exactly correct APA citations for these types of items would be to enter all of the creator as `author` and put their role label in parentheses after the author's first name:

```
Barris || K. (Writer & Director)
```

```
Oakley || B. (Writer)
Weinstein || J. (Writer)
Lynch || J. (Director)
```

rkopicki:
Thank you very much for your explanation and additional information regarding the details of `collapse`. I am aware of the limits of the current version of CSL (1.0.2) and hope that it will expand to support more and more cases. I also understand that any changes to CSL also require appropriate updates to the processor, i.e. `citeproc-js`.

Thank you very much for your help.

bwiernik:
Closes https://github.com/citation-style-language/styles/issues/7144

PR #7139, opened 2024-07-24
Update journal-of-rheumatology.csl

POBrien333:
Via: https://forums.zotero.org/discussion/116265/style-request-the-journal-of-rheumatology#latest

Paper: https://www.jrheum.org/content/jrheum/51/7/663.full.pdf

I went over the common item types.


bwiernik:
It looks like the journal uses this format for accessed dates—could you update that?

[Internet. Accessed October 25, 2020.]

POBrien333:
Good spot. Thanks :)

PR #7141, opened 2024-07-29
Switch genetics.csl to CSE dependent

POBrien333:
https://forums.zotero.org/discussion/116332/errors-in-genetics-oxford-journal-csl-file#latest


bwiernik:
@POBrien333 The failing tests were because another style had `<link href="http://www.zotero.org/styles/genetics" rel="template"/>`

PR #7150, opened 2024-08-03

enyedi:
Added formatting for "standard" type (e.g. ISO 9001). The "MDPI Reference List and Citations Style Guide" (https://mdpi-res.com/data/mdpi_references_guide_v9.pdf) does not specify it, but sends to the ACS Style Guide (https://pubs.acs.org/doi/full/10.1021/acsguide.40303). The ACS guide does give an example for formatting a "standard":

institution or publisher. italic title; number; location, year. DOI: doinumber

I kept "year" in bold, as the rest of the MDPI reference styles, although ACS does not use bold there. I also kept "volume" and "page", maybe there are standards which use that.

### DIFF
--- a/multidisciplinary-digital-publishing-institute.csl
+++ b/multidisciplinary-digital-publishing-institute.csl
@@ -12,11 +12,11 @@
     </author>
     <contributor>
       <name>Patrick O'Brien</name>
-    </contributor>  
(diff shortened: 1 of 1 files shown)


bwiernik:
Thanks!

PR #7153, opened 2024-08-06

ibadger:
Add variable Director, previously had used only Author but this wasn't displaying the Director in references produced in Zotero. If there is a better way to do it do let me know

### DIFF
--- a/harvard-university-for-the-creative-arts.csl
+++ b/harvard-university-for-the-creative-arts.csl
@@ -14,7 +14,7 @@
     <category citation-format="author-date"/>
     <category field="generic-base"/>
     <summary>University for the Creative Arts Harvard style</summary>
-    <updated>2024-08-19T14:41:31+00:00</updated>
+    <updated>2024-08-19T14:57:37+00:00</updated>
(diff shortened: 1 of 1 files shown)


harvard-university-for-the-creative-arts.csl
```diff
@@ -278,7 +278,7 @@
             </group>
           </if>
         </choose>
-        <names variable="author" prefix="Directed by ">
+        <names variable="author director" delimiter=" " prefix="Directed by ">
```
bwiernik, line 281:
```suggestion
        <names variable="director">
          <label variable="director" text-case="capitalize-first" suffix=" "/>
```

bwiernik (changes requested):
Thanks, I updated to use a more robust approach.

ibadger:
I added a Macro for Director and resubmitted, it seemed to validate but am unsure if it is the best way to go about it. Sorry for the delay I have been away from work, and am a bit slow when it comes to coding

ibadger:
I made some changes a while ago, but forgot about it  - does it work now?

PR #7156, opened 2024-08-08
Update comptes-rendus-author-date.csl

justinefabre:
We'd like to suppress the label [Author, date] before the bibliography entries.


justinefabre:
We have eventually reworked our CSL file so that it can be of greater assistance to authors wishing to submit their manuscripts to our journal: mandatory metadata (publication date, DOI for certain types of document, etc.) that are not filled in are now marked with indications such as “MISSING RELEASE YEAR”.

PR #7163, opened 2024-08-14
Create chicago-fullnote-bibliography-manuscript.csl

annaassogba:
A variant of Chicago 17th Notes & Bibliography full note style with changes to manuscript and letter item types that better match Chicago guidance.

The specific edits I made are detailed below, with my rationale following each bullet point:

Titles:
- Removal of the quotation marks around manuscript and letter item types and manuscript collection titles. If you want an item to have quotation marks around the title, you can add curly quotation marks to the Zotero title field. This allows users to decide for themselves whether or not the item at hand requires quotation marks around the title.
- Removal of default title casing for manuscript and letter titles in footnotes. Manuscript and letter titles will appear as entered in the Zotero field. Chicago guidance is nuanced; users can decide for themselves whether title or sentence casing is most appropriate for the item at hand.

Dates:
- Removal of the parentheses around a date or “n.d.” in a manuscript item footnote. Chicago guidance does not contain parentheses around dates in manuscript citations.
- Removal of the “n.d.” from manuscript collection bibliography entries. If dates are associated with a manuscript collection, they typically appear as part of the title. The indication of “no date” is not needed for the collection as a whole.
- The date format for manuscript and letter citations has been changed from month-day-year to day-month-year. Chicago guidance accepts both formats, but their examples employ day-month-year.

Subsequent citations:
- If a date is present in the Zotero date field, it will appear in subsequent shortened citations for manuscript and letter item types. If no date is present in Zotero, the “n.d.” will not appear in subsequent shortened citations. Chicago guidance indicates that more information than just the title may be useful in subsequent citations. Dates can be useful to distinguish between multiple similar documents.
- If the user enters text into the Place field in manuscript item types, that text will appear at the end of the shortened citation. This can be used for names of manuscript collections or repositories, if needed (e.g., if you have documents with similar titles across multiple collections or repositories). It may be useful to distinguish multiple similar documents by the collection or holding repository.

A few notes about how to best use this edited citation style:
- Use separate manuscript item and collection records: Create separate Zotero records for each individual manuscript item that you cite from the archival collection, plus another Zotero record for the collection as a whole. Use the individual manuscript item records to create footnote citations. Use the collection record for the bibliography. Note that when you put together the bibliography, you can use the functionality of the Zotero plugin in Word to edit the bibliography: specifically, you can remove the individual manuscript items from the bibliography and add in the collection record.

- Additional editing needed for letters that are part of manuscript collections:

- Instead of using the “Author”, “Contributor”, and “Recipient” options in Zotero’s creator field, write [sender’s name] to [recipient’s name] in the title field, as you’d like it to appear in the citation.

- For shortened subsequent citations, enter the shortened version of the title into the “Short Title” field in Zotero (e.g., Smith to Williams).

### DIFF
--- a/chicago-fullnote-bibliography-manuscript.csl
+++ b/chicago-fullnote-bibliography-manuscript.csl
@@ -1,0 +1,1548 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="display-and-sort" page-range-format="chicago">
+  <info>
+    <title>Chicago Manual of Style 17th edition (full note) manuscript</title>
+    <id>http://www.zotero.org/styles/chicago-fullnote-bibliography-manuscript</id>
+    <link href="http://www.zotero.org/styles/chicago-fullnote-bibliography-manuscript" rel="self"/>
+    <link href="http://www.zotero.org/styles/chicago-fullnote-bibliography" rel="template"/>
+    <link href="http://www.chicagomanualofstyle.org/tools_citationguide.html" rel="documentation"/>
+    <category citation-format="note"/>
+    <category field="generic-base"/>
+    <summary>Chicago format with full notes and bibliography</summary>
+    <updated>2024-03-17T19:02:00+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="en">
+    <terms>
+      <term name="editor" form="verb-short">ed.</term>
+      <term name="translator" form="verb-short">trans.</term>
+      <term name="translator" form="short">trans.</term>
+      <term name="editortranslator" form="verb-short">ed. and trans.</term>
+      <term name="editortranslator" form="verb">Edited and translated by</term>
+      <term name="translator" form="short">trans.</term>
+    </terms>
+  </locale>
+  <macro name="editor-translator">
+    <group delimiter=", ">
+      <group delimiter=" ">
+        <choose>
+          <if variable="container-author reviewed-author" match="any">
+            <group>
+              <names variable="container-author reviewed-author">
+                <label form="verb-short" text-case="lowercase" suffix=" "/>
+                <name and="text" delimiter=", "/>
+              </names>
+            </group>
+          </if>
+        </choose>
+      </group>
+      <names variable="editor translator" delimiter=", ">
+        <label form="verb-short" text-case="lowercase" suffix=" "/>
+        <name and="text" delimiter=", "/>
+      </names>
+    </group>
+  </macro>
+  <macro name="secondary-contributors-note">
+    <choose>
+      <if type="chapter entry-dictionary entry-encyclopedia paper-conference" match="none">
+        <text macro="editor-translator"/>
+      </if>
+    </choose>
+  </macro>
+  <macro name="container-contributors-note">
+    <choose>
+      <if type="chapter entry-dictionary entry-encyclopedia paper-conference" match="any">
+        <text macro="editor-translator"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks! So, two separate issues here:
1. We're not going to create a separate variant for Chicago style. If we should fix something in the existing Chicago styles, that should happen in the styles. (Obviously you're free to use and distribute these styles yourself; we're just not going to publish them on the main CSL repository)
2.While I do agree that better handling of archival manuscripts is worth another look, I don't think I agree with many of the approaches taken here:
- Chicago Manual is consistent in using title case for titles in English
- I disagree with removing dates. 14.229 includes an example with n.d. -- everything else does include dates
- The parentheses around the date are tricky -- they are included for unpublished manuscripts in 14.215 but not for archival manuscripts -- might make sense to test for archival information in that case to determine the date format
- place (csl's `publisher-place` isn't intended for archival information. With CSL 1.0.2 we have a nice set of archival fields that should be used instead (and can be entered in Zotero using Extra)

annaassogba:
Thanks for looking at this and for your response. Sorry to take so long to respond! I agree with not creating a separate variant and would love if you were open to making some of the changes in the existing Chicago style (I also saw that Chicago just updated to 18th style, so I understand that any changes you would make would be better to go in the 18th style rather than invest time in changing 17th). However, I also understand if you don’t agree with the approach I took and I recognize that you may not want to adopt all of the changes I proposed. I did want to respond to the points you made, in case I can provide any clarification around why I made the choices I did that would help with deciding whether or not to adopt some of these changes. All Chicago Manual of Style guidance I refer to here is 18th edition. 

_Chicago Manual is consistent in using title case for titles in English_
I agree that Chicago Manual is very consistent with using title case, but I see small exceptions to this with manuscript items with descriptive titles, such as described in 14.121, and examples #7 and #10 in 14.127. My code change allows the user to decide, for manuscript items, whether or not title case is appropriate for their specific use case, according to Chicago guidance. However, I do agree that title case is probably the safest route to go if a user is unsure, so I understand not wanting to make this change. 

_I disagree with removing dates. 14.229 includes an example with n.d. -- everything else does include dates_
I haven’t removed dates anywhere, so I think this is a misunderstanding. I may just not have stated my changes very clearly. In the edits I made, if a date is entered into the date field, it will appear, in either footnotes or a bibliography entry. If no date appears in the date field, an “n.d.” will appear in a footnote citation, but not for a bibliography entry. I chose this approach because in both the examples I see in Chicago guidance (14.128) and in real-life examples I’ve examined, any dates pertaining to whole manuscript collections are part of the title, so it would be confusing to have a title with specified dates, followed by n.d. Since usually just archival collections as a whole are cited in bibliographies rather than individual items, I thought this would apply in most cases. For the small number of cases where an individual archival item should be cited in a bibliography (and contains no date), I think the user could figure out a workaround. 
I can see a separate solution might be to create a new item type for “archival collection”, which could have different rules than the item type “manuscript”. 

_place (csl's publisher-place isn't intended for archival information. With CSL 1.0.2 we have a nice set of archival fields that should be used instead (and can be entered in Zotero using Extra)_
I can understand that my workaround is not using the code as intended, but the current “Archive Place” in the Extra field is not sufficient for the problem I was trying to solve, though perhaps there is another edit to the code that would work instead of what I did. The problem I was trying to solve is that sometimes, when using multiple archival documents that have similar titles but are from different archival repositories, the user may want to include extra information in subsequent citations, to help distinguish the items from each other (see 14.127, #s13-15), and sometimes users may want to use a shortened name of the archival repository in this situation, as well. The Archive Place in Extra doesn’t currently solve this problem. 

_also disagree with requiring 'incorrectly' entering letters. CSL should be capable of doing this properly using author and recipient. No need for a title for letters._
I agree with you about this point; I think I was mistaken.

bwiernik:
@adunning Flagging these PRs in case there are fixes not implemented in your work.

Going to close this PR in favor of the major revision by adunning for Chicago 18th edition.

PR #7177, opened 2024-08-25
Added common polish style

Fariusz:
Added style compliant with polish standard PN-ISO:690:2002


POBrien333:
Can you link to the documentation where this style is lined out (in the style and here)?

PR #7178, opened 2024-08-27
Create elsevier-vancouver-no-et-al.csl

POBrien333:
https://forums.zotero.org/discussion/117328/how-can-i-change-reference-style-to-list-all-authors-and-not-et-al#latest

Style dependent fix: https://github.com/citation-style-language/styles/pull/7178

### DIFF
--- a/elsevier-vancouver-no-et-al.csl
+++ b/elsevier-vancouver-no-et-al.csl
@@ -88,7 +88,7 @@
       <text variable="citation-number"/>
     </layout>
   </citation>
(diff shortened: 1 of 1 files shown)


adam3smith:
Could you manually fix the dependent style in here -- the Elsevier automatic dependents are so large, there's bound to be an issue that I don't have time to troubleshoot (note I edited the independent style above)

POBrien333:
Thanks. I had issues with submitting the PR and had to redo it 3 times. Of course then forgot the actual fix.  ;)

PR #6787, opened 2023-11-23
New style for Informationswissenschaft: Theorie, Methode und Praxis

nchachereau:
This is a new style created for the journal [Informationswissenschaft: Theorie, Methode und Praxis / Sciences de l’information: théorie, méthode et pratique](https://bop.unibe.ch/iw/), published by the education program in information, library and archival science of the universities of Berne and Lausanne (Switzerland).

I was mandated by the editors of the journal, who will add author guidelines on the website. The style is based on the infoclio.ch style developed for history students in Switzerland (#5754). Changes were made in order for the style to be usable both in French and in German (the journal is bilingual).

Many thanks for all your work!

PR #7191, opened 2024-09-05
infoclio.ch: add mention of editor and translator

nchachereau:
A small update to the infoclio.ch styles, for history students (and scholars) - originally submitted in #234 and last updated in #6833.

Responding to a request of users, we add the mention of editors of a book after its title. The use case is for critical editions of an author's work (say, Adam Smith), where we want to mention both the author and the scientific editors. We also display the translator in the same place.

As usual, thanks so much for the volunteer work of maintaining the styles! :clap: 

By the way, may I draw attention to my other PR #6787 ? The journal editors, who commissioned the style, have asked me again about the status.

cc @janbaumanninfoclio


nchachereau:
That was fast, thanks a lot!

PR #7192, opened 2024-09-06
Fixing OSCOLA pinpoints for subsequent citations

jonas-sk:
In OSCOLA, citations to cases use square brackets for paragraph numbers (e.g. `[127]`), in contrast to the using "para" for other sources (e.g. `para 127`). This is described on [page 19 of the OSCOLA handbook](https://www.law.ox.ac.uk/sites/default/files/migrated/oscola_4th_edn_hart_2012.pdf#page=25&zoom=auto,-206,651).

This difference in citations also transfer to the short-form citations used for subsequent citations (see [page 5 of the handbook](https://www.law.ox.ac.uk/sites/default/files/migrated/oscola_4th_edn_hart_2012.pdf#page=11&zoom=auto,-206,480)). Here is the example from the handbook:

> ibid [34] (Lord Hope), [39] (Lord Scott), [43]–[47] (Lord Walker), [58]–[60] (Lord
Neuberger)

While the handbook and the OSCOLA style agree for ibid citations, for other subsequent citations, which use a cross-reference, the CSL style reverts back to using `para` (see bold highlight):

- How it should be:  
  > <sup>1</sup> _Isayeva v Russia_ [2005] ECtHR app no 57950/00 [189–190]. <sup>2</sup> ibid [188]. … <sup>30</sup>  _Isayeva v Russia_ (n 1) [179].
- How it is: 
  > <sup>1</sup> _Isayeva v Russia_ [2005] ECtHR app no 57950/00 [189–190]. <sup>2</sup> ibid [188]. … <sup>30</sup> _Isayeva v Russia_ (n 1)**, para 179**.

(Admittedly, the OSCOLA handbook doesn't have an explicit example for these cross-referenced citations for cases using brackets, there's IMO no reason to assume that the pinpoint style for case would change without explicit guidance to that end.)

This PR is an attempt at fixing it, but I'm not 100% sure if it is the best approach (paging @adam3smith). This is in response to forum threads like [these](https://forums.zotero.org/discussion/81480/oscola-square-brackets-for-paragraph-numbers) that also point to the same problem.


bwiernik (commented):
Thanks! One question

PR #7193, opened 2024-09-06
Hide hyphen in author first name

pencilheart:


bwiernik:
Do you have a link to any documentation indicating this is correct? Generally names with hyphens include the hyphens when initialized in English typography

bwiernik:
In several articles from Elsevier, names are initialized with hyphens, so it appears that the current styles are correct. 

Eg, https://www.sciencedirect.com/science/article/pii/S2666651024000111/pdfft?md5=5c01070780bb0f7ea417c3293322b19c&pid=1-s2.0-S2666651024000111-main.pdf

PR #7214, opened 2024-09-16
Update ethnomusicology.csl

POBrien333:
https://forums.zotero.org/discussion/comment/474201/#Comment_474201


bwiernik:
It looks like the short title should in quotes

PR #7217, opened 2024-09-18
Create socialvetenskaplig-tidskrift.csl

GretaL:
New style for the journal Socialvetenskaplig tidskrift

### DIFF
--- a/socialvetenskaplig-tidskrift.csl
+++ b/socialvetenskaplig-tidskrift.csl
@@ -1,0 +1,368 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only">
+  <info>
+    <title>Socialvetenskaplig tidskrift</title>
+    <title-short>Socialvetenskaplig tidskrift</title-short>
+    <id>http://www.zotero.org/styles/socialvetenskaplig-tidskrift</id>
+    <link href="http://www.zotero.org/styles/socialvetenskaplig-tidskrift" rel="self"/>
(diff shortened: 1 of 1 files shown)


bwiernik:
Thanks! One question I have is about "Anon" for items without authors. Looking at some issues of the journal, I see several examples where the item title is substituted when there is no author, and no examples of using "Anon".

There is some complex logic about when Anon is used in the style code. Can you describe when you are intending Anon to appear vs not?

GretaL:
In fact, I don't want Anon to appear at all. My reference examples haven't made me aware of this. I guess I need to work some more on the style.

socialvetenskaplig-tidskrift.csl
```diff
@@ -0,0 +1,368 @@
...
+    </choose>
+    <choose>
+      <if type="article-newspaper entry-dictionary entry-encyclopedia" match="any">
+        <choose>
+          <if variable="author editor" match="none">
+            <text variable="container-title"/>
+          </if>
+        </choose>
+      </if>
+    </choose>
```
bwiernik, line 71:
```suggestion
  <names variable="author">
    <name and="symbol" delimiter-precedes-last="never" initialize-with=". " name-as-sort-order="all"/>
  <label form="short" prefix=" "/>
  <et-al font-style="italic"/>
  <substitute>
     <names variable="editor"/>
     <names variable="translator"/>
     <names variable="director"
     <text variable="title"/>
   </substitute>
 </names>
```

bwiernik (changes requested):
I think you can remove most of the complex name logic and just do something like this and a corresponding approach for author-short

bwiernik:
I put a suggestion for simplifying the author macro

GretaL:
I uploaded a new version. The sugested changes did not turn out the way I wanted.

PR #7229, opened 2024-09-29
Create melanges-de-linstitut-dominicain-detudes-orientales-fr.csl

NicolasSouchon:
Create a new CSL style for the Mélanges de l'Institut dominicain d'études orientales collection (French Version, documentation here: https://journals.openedition.org/mideo/9227)


POBrien333:
@NicolasSouchon 
I've reworked your style to bring them in line with best practises.
They are both now published and will be able to find them on https://www.zotero.org/styles
You can use the links from there to point your writers to it.

Any further issues, please open a new PR (you can add changes for both styles into the same PR)

PR #7237, opened 2024-10-04
Create Harvard - The University of Newcastle Australia

kaibis:
My first pull request here. Hopefully followed all the instructions correctly. This is for the University of Newcastle, we need a recognisable display name in Lean Library (which unfortunately does not have an option to override the title metadata).


kaibis:
I think I made an error with the file name

PR #7238, opened 2024-10-04
Create harvard-the-university-of-newcastle-australia

kaibis:
My first (second?) pull request. Hopefully followed all the instructions correctly. This is for the University of Newcastle, we need a recognisable display name in Lean Library (which unfortunately does not have an option to override the title metadata).


kaibis:
file extention.. so close.

PR #7240, opened 2024-10-04
Create Eclética Química style

naoliv:
Include a new style for Eclética Química (https://revista.iq.unesp.br/index.php/ecletica)

### DIFF
--- a/ecletica-quimica.csl
+++ b/ecletica-quimica.csl
@@ -1,0 +1,309 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="expanded" default-locale="en-US">
+  <info>
+    <title>Eclética Química</title>
+    <title-short>Eclét. Quím.</title-short>
+    <id>http://www.zotero.org/styles/ecletica-quimica</id>
+    <link href="http://www.zotero.org/styles/ecletica-quimica" rel="self"/>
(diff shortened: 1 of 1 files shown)


naoliv:
FYI, the style is not generating the citations correctly; we are looking into a solution.

bwiernik:
Thanks! For books with editors and no authors (like the "Beyond varieties of capitalism" example above), should the editor appear in the author position, or the title?

naoliv:
Hi!

The title. "Beyond varieties of capitalism, 2007" is correct here.

POBrien333:
@naoliv 
I've made some major improvements to the style. Please revise and then we can merge.
Can you confirm if you work for that journal?
I see in your guidelines that you link to a .csl style hosted via Mendeley. These user-provided styles can't guarantee a correct format and also update with following versions of CSL. (indeed there is an error for Preprints). Mendeley copies the CSL repository, but very often is behind by a good bit. Further to that, I would not recommend Mendeley anymore to your users for a variety of reasons. We'd appreciate to instead link to official styles repository which can be found here: https://www.zotero.org/styles
That way only ONE version of this style exists, which the possibility for the community and the CSL maintainers to update the style. Let me know if you have any questions about this.

naoliv:
Yes, I work in the technical (IT) department of the journal.

I will check with our librarian, who is responsible for the style, to make sure everything is correct. After I return from vacation, I also plan to make a small update to include her as an author of the style.

Regarding the style link, I will notify both the librarian and the editor-in-chief so we can update it accordingly.

I’ll get back to you as soon as possible.
Thank you very much.

PR #7242, opened 2024-10-07
Update and rename hochschule-bonn-rhein-sieg.csl to hochschule-bonn-r…

POBrien333:
…hein-sieg-harvard.csl

1. I had a look at the old guidelines via archive.org and they were actually for the economics faculty. Hence I added that to the style name. 
2. As the style also describes a footnote variant I have also added "harvard" to the file name, so we don't need to change that in the future.

Guidelines (sent to me via email):
[Leitfaden fuer die Gestaltung schriftlicher Arbeiten .pdf](https://github.com/user-attachments/files/17274842/Leitfaden.fuer.die.Gestaltung.schriftlicher.Arbeiten.pdf)


POBrien333:
Not sure where it's getting the "freie-hochschule-stuttgart" from?!

POBrien333:
Thanks. I've added the documentation link.
However, the test shows and error that actually is not true. The template is our DGP style which certainly exists. Somehow it's mentioning another style.

adam3smith:
The test fails because another style uses this one as a template and you're changing the ID. We generally prefer not changing style IDs if at all possible and I think should avoid that here.

adam3smith:
Yes exactly -- change title but not id. I think the CI won't complain, but if it does, there's as n override

adam3smith:
Thanks!

PR #7251, opened 2024-10-09
Update anesthesiology.csl

NoPenguinsLand:
Update Anesthesiology bibliography style to reflect the following rule: "If there are 7 or more authors in a reference, list the first 3 authors’ names, followed by et al.” When there are 6 or fewer authors in a reference, all names should be listed."

https://pubs.asahq.org/anesthesiology/pages/instructions-for-authors-general#references


adam3smith:
Thanks!

PR #7252, opened 2024-10-10
Create neues-jahrbuch-fur-geologie-und-palaontologie.csl

POBrien333:
https://forums.zotero.org/discussion/118403/style-request-neues-jahrbuch-fuer-geologie-und-palaeontologie-abhandlungen#latest

### DIFF
--- a/neues-jahrbuch-fur-geologie-und-palaontologie.csl
+++ b/neues-jahrbuch-fur-geologie-und-palaontologie.csl
@@ -14,12 +14,12 @@
     <category field="geography"/>
     <issn>0077-7749</issn>
     <eissn>2363-717X</eissn>
-    <updated>2024-10-21T08:27:01+00:00</updated>
+    <updated>2025-01-18T15:43:52+00:00</updated>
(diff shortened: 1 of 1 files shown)


bwiernik:
Thanks! Should “(eds.)” appear in text when substituting, and does the style call explicitly for “anonymous” for un-authored items? If not, let’s substitute title

POBrien333:
The "(eds.)" I don't know. It's not in the guidelines and not in the example paper I looked at.
I've taken the anonymous out. That's just from the template style.
I tend to leave things as is unless I have a reason to change it.

neues-jahrbuch-fur-geologie-und-palaontologie.csl
```diff
@@ -0,0 +1,255 @@
...
+        <names variable="translator"/>
+      </substitute>
+    </names>
+  </macro>
+  <macro name="author-short">
+    <names variable="author">
+      <name form="short" and="symbol" delimiter-precedes-last="never">
+        <name-part name="family" font-variant="small-caps"/>
+      </name>
+      <label form="short" prefix=" (" suffix=")"/>
```
adam3smith, line 97:
```suggestion
```

adam3smith (changes requested):

PR #7259, opened 2024-10-14
Create international-affairs.csl

POBrien333:
https://forums.zotero.org/discussion/105844/style-request-international-affairs#latest

### DIFF
--- a/international-affairs.csl
+++ b/international-affairs.csl
@@ -14,7 +14,7 @@
     <category field="humanities"/>
     <issn>0020-5850</issn>
     <eissn>1468-2346</eissn>
-    <updated>2024-10-14T15:42:14+00:00</updated>
+    <updated>2024-10-16T20:43:58+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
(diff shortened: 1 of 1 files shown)


bwiernik:
@adam3smith Do you know why the CI is failing? The style file is passing the Validator

adam3smith:
@bwiernik -- no, let me test out a couple of things, sorry for the noise this will cause (In case you're not aware, the validator does have, in the pull down, an option to switch the validation schema to a strict schema which is the one we're using for CI. This style passes that, too).

adam3smith:
The was a square bracket in the second documentation link. I'm not sure why that would break anything, but I'm guessing some broken XML parser or regex (the link works with or without it).

international-affairs.csl
```diff
@@ -0,0 +1,285 @@
...
+          <number variable="edition" form="ordinal"/>
+          <label variable="edition" form="short"/>
+        </group>
+      </if>
+      <else>
+        <text variable="edition"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="author">
```
adam3smith, line 43: @POBrien333 are you sure this shouldn't have a `<substitute>` element? See the Beyond Varieties... sample citation, which looks weird.

bwiernik: If we don't have an indication otherwise, let's include our standard editor -> title substitution

POBrien333: Thanks for spotting. Sure.

PR #7261, opened 2024-10-15
Create a CSL File named as Intellect Newgen Style

Rajamurugan-CEG:
Adding a new csl file to the repository


POBrien333:
Hi,
Can you explain what this style is for exactly? How does it differ from the [Intellect Newgen Books](https://www.zotero.org/styles/intellect-newgen-books) style we already have?

POBrien333:
@Rajamurugan-CEG 
It looks like you had missed my question above. Can you clarify?

PR #7273, opened 2024-10-20
create csl ucsi university harvard

zzlb0224:


POBrien333:
@zzlb0224 
The guidelines state "The reference list should be typed according to the guidelines outlined in the American Psychological Association (APA) 7th edition."

Can you confirm that this style is identical to APA? If so we'd make a dependent style that just links to APA instead of this custom style.

PR #7276, opened 2024-10-21
Update the-journal-of-wildlife-management.csl

POBrien333:
https://forums.zotero.org/discussion/118963/style-error-journal-of-wildlife-management#latest


adam3smith:
Thanks!

PR #7284, opened 2024-10-26
Create elsevier-vancouver-short-author-list.csl

POBrien333:
https://forums.zotero.org/discussion/119117/style-error-journal-of-hepatology#latest

Please suggest a different name. Wasn't sure about what to call this.

### DIFF
--- a/elsevier-vancouver-short-author-list.csl
+++ b/elsevier-vancouver-short-author-list.csl
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" page-range-format="minimal" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
   <info>
-    <title>Elsevier - Vancouver (7/3 et al.)</title>
-    <id>http://www.zotero.org/styles/elsevier-vancouver-3-et-al</id>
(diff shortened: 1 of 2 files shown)


adam3smith:
We have reason to think this is more common than the one journal? If so, let's add to title ("et al." for 3+ authors) and maybe just do -short-author-list for id & filename? Happy to use something else for the filename, but not looking current version

POBrien333:
Thanks. That's changed

elsevier-vancouver-short-author-list.csl
```diff
@@ -0,0 +1,162 @@
...
+  </macro>
+  <citation collapse="citation-number">
+    <sort>
+      <key variable="citation-number"/>
+    </sort>
+    <layout prefix="[" suffix="]" delimiter=",">
+      <text variable="citation-number"/>
+    </layout>
+  </citation>
+  <bibliography et-al-min="7" et-al-use-first="3" second-field-align="flush" entry-spacing="0">
```
adam3smith, line 90: Sorry, one more question -- shouldn't this be 4/3? That's what the linked thread says

POBrien333: classic me. Forgetting the main thing getting sidetracked with the naming.

POBrien333:
Thanks. PR for change of dependent style here: https://github.com/citation-style-language/journals/pull/152

PR #7286, opened 2024-10-26
Create apa-7th-prt-legislation.csl

rppm:
This is an adaptation of APA 7th edition to include citation and bilbiography of Portuguese law according to portuguese legal citation (https://www.parlamento.pt/ArquivoDocumentacao/Documents/AR_Regras_Legistica.pdf). The correct way to use this style is to use document type legislation ("statute" in Zotero), putting the location (Lisboa) in the references field ("History" in Zotero).


rppm:
Error in the filename, sorry.

PR #7287, opened 2024-10-26
Create apa-7th-prt-legislation.csl

rppm:
This is an adaptation of APA 7th edition to include citation and bilbiography of Portuguese law according to portuguese legal citation (https://www.parlamento.pt/ArquivoDocumentacao/Documents/AR_Regras_Legistica.pdf). The correct way to use this style is to use document type legislation ("statute" in Zotero), putting the location (Lisboa) in the references field ("History" in Zotero).

### DIFF
--- a/apa-7th-prt-legislation.csl
+++ b/apa-7th-prt-legislation.csl
@@ -1,0 +1,1770 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" page-range-format="expanded">
+  <info>
+    <title>American Psychological Association 7th edition incl. Portuguese legislation</title>
+    <title-short>APA 7th (PRT-legislation)</title-short>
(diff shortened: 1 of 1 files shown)


bwiernik:
Thanks— the correct variable to indicate location for legal items in CSL is the `jurisdiction` variable. Can you please adapt your style to use this instead of co-opting the `history` variable?

rppm:
> Thanks— the correct variable to indicate location for legal items in CSL is the `jurisdiction` variable. Can you please adapt your style to use this instead of co-opting the `history` variable?

@bwiernik, changed the item variable to jurisdiction. The only problem I foresee is that neither Zotero nor Mendeley include `jurisdiction` in bill or statute item type.

rppm:
@bwiernik, remade the code to be as close as possible to APA 7th, but to include portuguese legislation.

bwiernik:
@adam3smith @POBrien333 I'm thinking it's okay to accept this considering we don't have a way to do modular legal references in standard CSL. 

What do you think the filename and title formatting should be?

PR #7292, opened 2024-10-28
Update revue-francaise-de-gestion.csl

rpierronnet:
At the request of the journal, change in the number of authors from which ‘et al.’ is used. is used (from 4 to 3).


POBrien333:
Could you click on the following link and hit the "Sync fork" button? That'll remove the validation errors.
https://github.com/rpierronnet/styles/tree/patch-5

rpierronnet:
Hi @POBrien333 
Done!

POBrien333:
Thank you. The style will be available in 1-2h.

PR #7299, opened 2024-10-30
Update modern-humanities-research-association.csl

POBrien333:
https://forums.zotero.org/discussion/comment/477733/#Comment_477733
https://forums.zotero.org/discussion/113436/mhra-4th-edition#latest

### DIFF
--- a/modern-humanities-research-association.csl
+++ b/modern-humanities-research-association.csl
@@ -23,7 +23,7 @@
     <category citation-format="note"/>
     <category field="generic-base"/>
     <summary>MHRA format with full notes and bibliography</summary>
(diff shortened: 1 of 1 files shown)


modern-humanities-research-association.csl
```diff
@@ -102,16 +102,13 @@
       </else>
     </choose>
   </macro>
-  <macro name="disambiguate">
+  <macro name="title-subsequent">
     <choose>
       <if disambiguate="true">
         <choose>
-          <if variable="title" match="none">
```
adam3smith, line 109: I think you want to leave that test for rare cases (like a letter without a title) but you want to remove the disambiguate="true" test above)

adam3smith:
Thanks!

PR #7308, opened 2024-11-05
Create ucl-press-note.csl

maaxlong:
New style created according to the style guide for UCL Press


kjklosowski:
Hi @maaxlong you should definietly fix the errors bot gave you. You defined quite a bit of macros that were not used there.

As for the et-al errors: et-al-user-first sets the number of names to be displayed when et al. is used, while et-al-min is the amount of numbers that have to be reached for the et al to be used. [See the documentation](https://docs.citationstyles.org/en/stable/specification.html) and feel free to try again :)

PR #7309, opened 2024-11-05
Update bulletin-de-correspondance-hellenique.csl

AlZorba:
Titles shouldn't be displayed with "title case" but with "capitalize-first"


POBrien333:
Thank you.

PR #7312, opened 2024-11-12
Update nottingham-trent-university-library-harvard.csl

POBrien333:
https://forums.zotero.org/discussion/119574/nottingham-trent-university-library-harvard-referencing-style#latest


adam3smith:
Don't you need and="text" in author-short? That looks like it was the main issue reported

POBrien333:
Yeah, the "and" is set globally on line 2 of the style

PR #7326, opened 2024-11-17
Create nature-single-spaced-no-et-al.csl

avikdut:
Screenshot 2024-11-16 231157
To create single-spaced reference list. Default Nature CSL style file produces double-spaced reference list in Word


avikdut:
Realized that this has to be an independent style file and not a dependent one. Closing PR.

PR #7327, opened 2024-11-17
Created new style Nature single-spaced no "et al."

avikdut:


bwiernik:
Could you say more about what this style is for? What journal requires it?

avikdut:
This is for Nature and Nature-style journals when citations are inserted
using the Zotero+Google Docs plugin.

The style inserts the reference list double spaced but that takes up a lot
of space and it would be better to insert single spaced versions with all
author names spelled out. It is also helpful for grant applications. Thanks.

avikdut:
I mean the Nature - no et al. Style inserts the reference list double
spaced, while this one inserts it single spaced.

bwiernik:
The Springer-Nature journal guidelines call for the reference list to be double spaced, so it doesn't seem like there is a publisher calling for this style. 

You're welcome to adapt and use a style for your own preferences, but we aren't going to accept a style like this for the repository. 

Thanks for your submission!

PR #7335, opened 2024-11-21
Create international-journal-of-geographical-information-science.csl

djw-easy:
Create IJGIS journal style.


djw-easy:
Create IJGIS journal style.

PR #7343, opened 2024-12-03
Update beltz-padagogik.csl

POBrien333:
https://forums.zotero.org/discussion/42351/style-request-beltz#latest

[Uploading Hinweise Manuskript Beltz Juventa.pdf…]()

### DIFF
--- a/beltz-padagogik.csl
+++ b/beltz-padagogik.csl
@@ -11,12 +11,12 @@
       <uri>https://github.com/zuphilip</uri>
     </author>
     <contributor>
-      <name>Philipp Zumstein</name>
+      <name>Patrick O'Brien</name>
     </contributor>
     <category citation-format="author-date"/>
     <category field="humanities"/>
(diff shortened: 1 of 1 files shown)


beltz-padagogik.csl
```diff
@@ -111,7 +126,14 @@
   <macro name="year">
     <group delimiter="/">
       <date date-parts="year" form="text" variable="original-date"/>
-      <date date-parts="year" form="text" variable="issued"/>
+      <choose>
+        <if match="any" is-uncertain-date="issued">
```
bwiernik, line 130: Should 'no date" also be shown if there is no `issued`?

POBrien333: I made a change, but I'm not sure I understand (awkward smile)

bwiernik: I meant, I would think the style should also show "no date" if `variable="issued" match="none"`

PR #7344, opened 2024-12-03
Chicago: Use title case for volume title

adunning:
This ensures the correct capitalization of a volume title stored in sentence case, such as:

> Knowles, David. *The Religious Orders in England.* Vol. 3, *The Tudor Age.* 3 vols. Cambridge: Cambridge University Press, 1959. <https://doi.org/10.1017/CBO9780511560668>.

I observe when testing with Zotero that, if the language is not set to English, the volume title is not set to sentence case as I would expect. I cannot however seem to find a way to control this in the CSL file: is this a bug in citeproc-js?

### DIFF
--- a/chicago-annotated-bibliography.csl
+++ b/chicago-annotated-bibliography.csl
@@ -545,7 +545,7 @@
               <text term="volume" form="short"/>
               <number variable="volume" form="numeric"/>
             </group>
-            <text variable="volume-title" font-style="italic"/>
+            <text variable="volume-title" text-case="title" font-style="italic"/>
           </group>
           <choose>
             <if variable="locator" match="none">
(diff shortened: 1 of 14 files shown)


bwiernik:
Casing is only applied to English titles. This is the expected behavior, as other languages have their own distinct casing rules and do not observe sentence/title casing conventions

adunning:
I am not seeing the expected behaviour you have described: with my proposed change, the volume title is being converted to title case even if the language is set to a non-English language code. Take this example:

> Stotz, Peter. *Handbuch zur lateinischen Sprache des Mittelalters.* Vol. 1, *Einleitung, Lexikologische Praxis, Wörter und Sachen, Lehnwortgut.* 5 vols. Handbuch der Altertumswissenschaft, II.5. Munich: Beck, 2002.

Zotero is capitalizing 'und' with `text-case="title"` on `volume-title`, which must be a bug.

PR #7346, opened 2024-12-04

ibadger:
This change is to do with Directors in the CSL. Zotero is using CSL 1.0.2 and uses the Director variable and tag, this doesn't display at the moment in our current CSL when referencing with Zotero. I made a commit in the summer but withdrew it as the changes suggested by the reviewer had a knock on impact, as I've since discovered that other software that students use at UCA (Paperpile, BibGuru, myBib) don't currently use CSL 1.0.2. and won't display the director or tag if I update the code as was suggested.

Would it be possible to approve this commit, where the code will display either director or author and use a prefix instead of a label?

"The "Director" name variable was introduced last year with CSL 1.0.2. Therefore, a lot of citation styles don't include it yet."

From <https://forums.zotero.org/discussion/comment/476797/#Comment_476797>

### DIFF
--- a/harvard-university-for-the-creative-arts.csl
+++ b/harvard-university-for-the-creative-arts.csl
@@ -14,7 +14,7 @@
     <category citation-format="author-date"/>
     <category field="generic-base"/>
     <summary>University for the Creative Arts Harvard style</summary>
-    <updated>2022-06-22T13:41:19+00:00</updated>
+    <updated>2024-12-04T15:27:19+00:00</updated>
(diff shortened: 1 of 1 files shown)


ibadger:
Hi
It worked on the https://validator.citationstyles.org/ webpage, but I am getting a 'Process completed with exit code 1.' error

Any advice on what to do would be greatly appreciated

Best wishes

Ian

adam3smith:
CI is currently broken -- please ignore the failure

PR #7348, opened 2024-12-06
Anesthesiology style add "et al" formatting

levin-lab:
Add "et al" formatting for the journal Anesthesiology. As per journal guidelines, use et al if more than 7 authors, and list the first three.


bwiernik:
Thanks, I merged this using a replacement PR to avoid the extraneous commits

levin-lab:
Thank you! It's been ages since I submitted a pull request so I was a bit rusty.

PR #7350, opened 2024-12-09
Add volume-title support to MHRA

adunning:
See <https://www.mhra.org.uk/style/chapter7.html#section_7_3>, example xi.


adunning:
Closing in favour of #7418.

PR #7354, opened 2024-12-11
Use "et al" in Romanian

bwiernik:


bwiernik:
@adam3smith @retorquere Any idea why the CI is broken?

retorquere:
We amended something about how yaml files were loaded (regarding a potential attack vector in YAML loading) but I honestly have no idea where we did that.

POBrien333:
via https://forums.zotero.org/discussion/120338/romanian-apa-citation-zotero-help#latest

retorquere:
I've reverted that fix, but this pr will have to merge master to get it.

PR #7359, opened 2024-12-17
Update leviathan.csl

POBrien333:
https://forums.zotero.org/discussion/120401/leviathan-deutsch

@adam3smith 
Could you help me understand why it's not printing "o. J."?


adam3smith:
It's because of the `original-date` lines. I don't quite understand why, to be honest.

POBrien333:
Interesting. Thanks for finding why. 
Fixed it with a workaround.

PR #7361, opened 2024-12-17
Update american-physics-society.csl

### LINKED ISSUE #7360: Specification of cis:if match="all" for field type
I was using the `american-physics-society` style for my thesis with Hayagriva and I noticed that webpages wouldn't render properly. 

It has the following else-if clause:
```xml
<else-if type="post-weblog interview manuscript pamphlet personal_communication post post-weblog speech webpage">
```

Now the CSL 1.0.2 specification gives the following:

```
The cs:if and cs:else-if elements may carry the match attribute to control the testing logic, with allowed values:

    “all” - (default), element only tests “true” when all conditions test “true” for all given test values
    “any” - element tests “true” when any condition tests “true” for any given test value
    “none” - element only tests “true” when none of the conditions test “true” for any given test value
```

According to this, type would have to match all types. I can see how with the BibTeX specification this is not possible, however it's not specifically stated that match="all" still only matches one style. Is that the case? Using the online code editor seems to verify that theory.

Also, for all other conditionals, the `american-physics-society.csl` does have the match="any" attribute, which adds to the confusion.

POBrien333:
closes #7360

### DIFF
--- a/american-physics-society.csl
+++ b/american-physics-society.csl
@@ -21,7 +21,7 @@
     <category citation-format="numeric"/>
     <category field="physics"/>
     <summary>Common style use by APS publications.</summary>
-    <updated>2024-12-1711:58:54+00:00</updated>
+    <updated>2024-12-17T11:58:54+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="author">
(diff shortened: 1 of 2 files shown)


american-physics-society.csl
```diff
@@ -21,7 +21,7 @@
     <category citation-format="numeric"/>
     <category field="physics"/>
     <summary>Common style use by APS publications.</summary>
-    <updated>2022-03-02T11:58:54+00:00</updated>
+    <updated>2024-12-1711:58:54+00:00</updated>
```
adam3smith, line 24:
```suggestion
    <updated>2024-12-17T11:58:54+00:00</updated>
```

adam3smith (changes requested):

PR #7366, opened 2024-12-19
Create ou-nl-cw-vademecum.csl

jhmj-io:
Style for the Humanitas Faculty of the Open University The Netherlands. https://www.ou.nl/faculteit-cultuurwetenschappen. The Human Readable form is available in the Vademecum of the faculty.

### DIFF
--- a/ou-nl-cw-vademecum.csl
+++ b/ou-nl-cw-vademecum.csl
@@ -1,0 +1,401 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" default-locale="nl-NL" version="1.0" demote-non-dropping-particle="display-and-sort">
+
+  <info>
+    <title>OU NL CWS (Nederlands)</title>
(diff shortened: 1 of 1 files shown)


POBrien333:
Have a look at our style requirements: https://github.com/citation-style-language/styles/blob/master/STYLE_REQUIREMENTS.md#csl-style-repository-requirements
Please, at a minimum, add the style you used as a template.

Also, the style's name (and file name, link, ID) should follow this pattern including the language, e.g.: Universiteit Utrecht - Onderzoeksgids Geschiedenis (Nederlands)
(for example "Vademecum to appear in the style name does not make sense)

POBrien333:
@jhmj-io 
Thanks for your patience.
I just made some edits, but noticed that you added some locale translations for terms because they were probably giving you english terms. That's because the "locale" for nl-NL is incomplete. Would you be so kind and open a PR to translate as many English terms here as you can? https://github.com/citation-style-language/locales/blob/master/locales-nl-NL.xml
Then we can review the local overrides of this style and get it merged, ready for the students coming back no.w

PR #7370, opened 2024-12-30
Create aves.csl

LouisNatagora:
CSL file for Aves (aves.natagora.be)

### DIFF
--- a/aves.csl
+++ b/aves.csl
@@ -3,8 +3,8 @@
   <info>
     <title>Aves</title>
     <id>http://www.zotero.org/styles/aves</id>
-    <link href="https://www.zotero.org/styles/aves" rel="self"/>
-    <link href="https://www.zotero.org/styles/acta-zoologica-academiae-scientiarum-hungaricae" rel="template"/>
+    <link href="http://www.zotero.org/styles/aves" rel="self"/>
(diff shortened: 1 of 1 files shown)


aves.csl
```diff
@@ -0,0 +1,231 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" and="symbol" default-locale="fr-BE" demote-non-dropping-particle="never">
```
bwiernik, line 2:
```suggestion
<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" and="symbol" default-locale="fr-FR" demote-non-dropping-particle="never">
```

bwiernik (changes requested):

bwiernik:
Thanks! We don't currently have a separate fr-BE locale. Other Belgian styles in the repository have used the fr-FR locale, so I'm going to switch to that for the time being so the style can be merged. 

Could you take a look at the list of French terms here and tell me if Belgian French writing uses any different terms?
https://github.com/citation-style-language/locales/blob/master/locales-fr-FR.xml

LouisNatagora:
Thank you! The fr-FR locale can be used. The very little differences between Belgian and French French are not present here.

PR #7371, opened 2024-12-30
Create norma-cehr.csl

UCP-CEHR:
Creation of the citation standard in use at UCP-CEHR for the Zotero programme

### DIFF
--- a/norma-cehr.csl
+++ b/norma-cehr.csl
@@ -1,0 +1,565 @@
+Norma "CEHR"- Código fonte:
+
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" class="in-text" default-locale="en-US">
+  <info>
+    <title>Norma Portuguesa 405 (Português)</title>
+    <title-short>NP 405</title-short>
+    <id></id>
+    <author>
+      <name>Universidade Católica</name>
+      <uri></uri>
+      <email></email>
+    </author>
+    <author>
+      <name>Inês Tripeça Alves</name>
+      <email>inestripeca6gmail.com</email>
+    </author>
+    <author>
+      <name>Cristina Cortês</name>
+      <email>ccortes@ua.pt</email>
+      <uri>http://www.mendeley.com/profiles/cristina-cortes/</uri>
+    </author>
+    <author>
+      <name>Diana Silva</name>
+      <email>dianasilva@ua.pt</email>
+    </author>
+    <author>
+      <name>Ana Bela Martins</name>
+      <email>abela@ua.pt</email>
+    </author>
+    <contributor>
+      <name>Susana Dias</name>
+      <email>sdias@ua.pt</email>
+    </contributor>
+    <contributor>
+      <name>Bella Nolasco</name>
+      <email>bellanolasco@ua.pt</email>
+    </contributor>
+    <contributor>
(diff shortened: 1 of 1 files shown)


kjklosowski:
Hi @UCP-CEHR. If you want to have it published remove first two lines from the code, as that's what is responsible for these errors. The code should start with `<?xml version="1.0" encoding="utf-8"?>` on line 1, and nothing before it.

kjklosowski:
@UCP-CEHR I'm glad you fixed these mistakes but the style itself has some problems and is very "bulky". For example:
- your medium macro has if else statement that produces the same output. The output itself is empty text box, which will not work (you need to change that to url variable, which is fed by reference manager),
- Also you have multiple issues with double punctuaction (visible even in bot's sample citation). As a rule of thumb, I'd recommend using group delimiters (and I can see you have some groups already defined), because they help to avoid these mistakes and they will not produce orphan dots if any variable is left empty in the reference manager.
- Also I can see some brute force choices, for example: in else-if (chapter or entry...) you have secondary-contributions macro which will work only when the citation is neither chapter nor conference paper nor patent. There are better ways to make it run only when the citation is an entry...
These are couple I spotted by skimming your code.

I matched the examples below.

I recommend using the time you wait before any maintainer (I'm not one, I can't do nothing, I only help) to optimize the code and its logic :). I recommend [Visual CSL Editor](https://editor.citationstyles.org/visualEditor/) so you can see everything better. Good luck with your style!

norma-cehr.csl
```diff
@@ -0,0 +1,516 @@
...
+            <text variable="title" form="short" quotes="false" font-style="italic"/>
+          </else>
+        </choose>
+      </substitute>
+    </names>
+  </macro>
+  <macro name="medium">
+    <choose>
+      <if variable="URL" match="any">
+        <text value="" prefix=" "/>
```
kjklosowski, line 126: Example one: empty medium macro

norma-cehr.csl
```diff
@@ -0,0 +1,516 @@
...
+            <text variable="number-of-pages"/>
+            <text variable="collection-title" prefix=" (" suffix="; "/>
+            <text variable="collection-number" suffix=")."/>
+          </group>
+        </else-if>
+        <else-if type="chapter entry-dictionary entry-encyclopedia" match="any">
+          <group>
+            <text macro="author" suffix=" - "/>
+            <text variable="title" suffix=". "/>
+            <text macro="secondary-contributors" suffix=". "/>
```
kjklosowski, line 370: better ways to achieve this than macro

norma-cehr.csl
```diff
@@ -0,0 +1,516 @@
...
+            <text macro="archive" suffix=". "/>
+          </group>
+        </else-if>
+        <else-if type="webpage">
+          <group>
+            <text macro="author" suffix=" - "/>
+            <text macro="title"/>
+            <text variable="collection-title" suffix=". "/>
+            <text macro="medium"/>
+            <text macro="publisher" prefix=". "/>
```
kjklosowski, line 453: That's the example of double punctuation error. If, for some reason, the medum macro will produce no output, you will end up with collection title. . publisher - as publisher has '. ' as a prefix.

Prefixes are generally dangerous and should be used wisely. I usually use them only if the object has some kind of if statement beforehand, but even then it's very rare.

bwiernik:
Closing in favor of #7487

Issue #6474, opened 2023-03-23
Title in foreign language


philcoul:
Hi Does anyone know the field to use in Zotero in order to include a title translation as below. The article is in French and I need to add the translation in English between brackets. I am using Turabian 8th edition fullnote
Thanks

Niemczewski, Wojciech. “La culture comme Religion : L’interprétation Postmoderne de la Relation entre la Culture et la Religion.” [Culture as Religion: The Postmodern Interpretation of the Relationship between Culture and Religion.] Doctorate’s Thesis, Université de Strasbourg, France, 2013. HAL Open Science. https://tel.archives-ouvertes.fr/tel-00924192/document.

philcoul:
Also, when the source is in French and it is a chapter in a book, I would need a way to add the translation of both chapter and the book titles like below
Thanks
---
Peach, Wesley. “Le Réveil au Québec: Un autre regard.” [Rethinking Revival in Quebec: Another Perspective] in Histoire du Protestantisme au Québec depuis 1960: Une Analyse Anthropologique, Culturelle et Historique [History of Protestantism in Quebec since 1960: An Historical, Cultural and Anthropological Analysis], edited by Richard Lougheed, Wesley Peach, and Glenn Smith, 125-46. Québec, Québec : Éditions La Clairière, 1999.

adam3smith:
(please put Zotero questions on the Zotero forums -- we don't have enough people to do user support here). 

There's no CSL variable or Zotero field for this. It's possible to hack together something using another variable (like original-title or medium or so) and modifying the CSL, but unless you have a ton of these, I'd just recommend fixing it by hand if you really do need to add a translation (I'll say doing that for French or other Western European languages seems unusual and unnecessary)

adam3smith:
I'd be up for preparing a limited 1.0.3 release with maybe a target date for early 2024. 
I'm not seeing anyone with enough energy currently to push through an ambitious 1.1 in the next couple of years. Maybe move to the forums?

PR #7372, opened 2025-01-02
Update chicago-fullnote-bibliography.csl to include original-title

POBrien333:
https://forums.zotero.org/discussion/120837/quick-question-about-translated-articles-in-zotero-cmos#latest


bwiernik:
The linked thread is better suited to a `translated-title` variable that we have [discussed](https://github.com/citation-style-language/styles/issues/6474). Maybe let's try to push out the small 1.0.3 update and use that instead here.

adam3smith:
I'll close this thread given the pending CMoS updates

PR #7374, opened 2025-01-07
Update american-journal-of-sociology.csl

POBrien333:
via https://forums.zotero.org/discussion/120945/style-error-american-journal-of-sociology#latest


POBrien333:
I had it in the bibliography already, but now added it to the in-text citations.

PR #7390, opened 2025-01-16
Update ieee.csl to use non-breaking space in author initials

eltos:
Avoids unfortunate breaking, i.e. produces
> A. Author, B. Author and
> C. Author, "A paper", ...

instead of
> A. Author, B. Author and C.
> Author, "A paper", ...

----

PS: Currently, only about 1% of the styles use non-breaking space and as such will suffer from the issue above.
- [11 styles using non-breaking space](https://github.com/search?q=repo%3Acitation-style-language%2Fstyles+%22initialize-with%3D%5C%22.%26%23160%3B%5C%22%22&type=code)
- [964 styles using plain space](https://github.com/search?q=repo%3Acitation-style-language%2Fstyles+%22initialize-with%3D%5C%22.+%5C%22%22&type=code)

### DIFF
--- a/ieee.csl
+++ b/ieee.csl
@@ -115,7 +115,7 @@
   </macro>
   <macro name="author">
     <names variable="author">
-      <name and="text" et-al-min="7" et-al-use-first="1" initialize-with=". "/>
+      <name and="text" et-al-min="7" et-al-use-first="1" initialize-with=".&#160;"/>
       <label form="short" prefix=", " text-case="capitalize-first"/>
       <et-al font-style="italic"/>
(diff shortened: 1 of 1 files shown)


bwiernik:
Thanks for the submission! I can certainly understand the aesthetic preference for non-breaking spaces here. However, IEEE's style guide does not call for non-breaking spaces, and they do not use non-breaking spaces in IEEE journals (e.g., see reference 30 here https://ieeexplore.ieee.org/document/10535236), so we are going to keep the IEEE CSL style as-is with a standard space.

PR #7392, opened 2025-01-17
Update cardiff-university-vancouver.csl

YoungZoe:
I have altered the macro name = date so that if a journal article has a volume and pages then only the year of publication appears but if that information is missing then we assume it is a preprint and the month and day also appear with year.

### DIFF
--- a/cardiff-university-vancouver.csl
+++ b/cardiff-university-vancouver.csl
@@ -177,9 +177,16 @@
   <macro name="date">
     <choose>
       <if type="article-journal article-magazine article-newspaper review review-book" match="any">
+<choose>
+<if match="none" variable="volume page">
         <group suffix=";" delimiter=" ">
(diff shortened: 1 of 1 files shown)


YoungZoe:
Not sure if there is anything else i need to do. Is the reindent doen automatically?

YoungZoe (commented):
Please can i my request to make this live be aprroved

bwiernik:
Thanks. Preprints have their own item type (`article`), so the correct approach to handle preprint formatting would be to add new code to the style that formats the `article` type.

Could you update your style accordingly?

bwiernik:
I see, you mean "ahead of print", not "preprint". "Preprint" generally refers to a scholarly work that has been posted online in a repository like BioRxiv or ArXiv but not yet peer reviewed. This change makes sense now

PR #7394, opened 2025-01-18
Create jime.csl

xolotl:
Creating a first version of a CSL file to support formatting references following the requirements of the Journal of Interactive Media in Education (JIME).


jime.csl
```diff
@@ -0,0 +1,362 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>Journal of Interactive Media in Education - Harvard</title>
+    <title-short>JIME - Harvard</title-short>
+    <id>http://www.zotero.org/styles/jime</id>
+    <link href="http://www.zotero.org/styles/jime" rel="self"/>
```
POBrien333, line 7: Please change the ID and and self link in lines 6+7 to the following according to our [submission guidelines](**url**):
..../harvard-journal-of-interactive-media-in-education   (instead of the .../jime)

Please also change the actual file name to match that, with the .csl suffix

jime.csl
```diff
@@ -0,0 +1,362 @@
...
+    <group delimiter=". ">
+      <group delimiter=" ">
+        <number vertical-align="baseline" variable="edition" form="ordinal"/>
+        <label vertical-align="baseline" variable="edition" form="short"/>
+      </group>
+      <group delimiter=" ">
+        <label text-case="capitalize-first" variable="volume"/>
+        <text variable="volume"/>
+      </group>
+      <group font-variant="normal">
```
POBrien333, line 153: Please delete any ` font-variant="normal"` throughout the style unless it's on purpose.

jime.csl
```diff
@@ -0,0 +1,362 @@
...
+      <text variable="publisher-place" suffix=": "/>
+      <text variable="publisher"/>
+      <text variable="page" prefix=":"/>
+    </group>
+  </macro>
+  <macro name="access">
+    <choose>
+      <if type="book chapter thesis" match="none">
+        <choose>
+          <if variable="DOI">
```
POBrien333, line 195: This can easily be simplified to the following as none of the groups do anything here and the DOI URI prefix should be closer to the variable anyway:
          <if variable="DOI">
          </if>

jime.csl
xolotl: Addressed via 3ec264f0095d0a4fa746917d82b343738a23513f

jime.csl
xolotl: Any use of font-variant="normal" came from the template CSL used to start this file. I have removed references to it via 482b205, including reducing lines 153-155 by removing what seems like an unnecessary group to just have one line to call the editor-translator macro.

jime.csl
xolotl: This logic came from the template CSL used to start this file. I didn’t see any provided example, but simplified the macro based on my understanding of what might be unnecessary via 926ef59.

xolotl:
Thank you for the careful review @POBrien333! I believe I have addressed all your suggestions, but the automated tests now seem to be failing with what may be an issue in the test code. I see error: "Run bundle exec sheldon --token=$GITHUB_TOKEN --failure --verbose 
bundler: failed to load command: sheldon (/home/runner/work/styles/styles/vendor/bundle/ruby/3.0.0/bin/sheldon).

PR #7395, opened 2025-01-18
Update american-journal-of-agricultural-economics.csl

nvilloria:
Chenged style so citations with more than for authors use et al. Eg.  Michler, Tjernström, Verkaart, and Mausch (2019) should be cited as Michler et al. (2019). As per May 2020 AJAE checklist.


POBrien333:
@nvilloria 
Could you please click [here](https://github.com/nvilloria/styles/tree/patch-2) and sync this PR using "Sync fork" you'll find up top? Then we can merge. thank you.

PR #7397, opened 2025-01-21
Create elsevier-vancouver-author-date-alphabetical.csl

### LINKED ISSUE #7393: In-text citation ordering is wrong for Journal of rock mechanics and geotechnical engineering (JRMGE) style
Hi,
I realized that the sytle for JRMGE is not right on zotero. For example citation below should appear as: 
(Raymond et al., 2008; Razdan et al., 2008; Kranz and Dillenardt, 2010)
instead of: 
(Kranz and Dillenardt, 2010; Raymond et al., 2008; Razdan et al., 2008)

POBrien333:
Closes #7393

@adam3smith I just created a new parent style that sorts alphabetically. Let me know if this is the correct approach and then I'll also make the change for the dependent style.

### DIFF
--- a/elsevier-vancouver-author-date-alphabetical.csl
+++ b/elsevier-vancouver-author-date-alphabetical.csl
@@ -12,7 +12,7 @@
     <category citation-format="author-date"/>
     <category field="generic-base"/>
     <summary>A style for some Elsevier journals, resembles Vancouver style, but using author-date format</summary>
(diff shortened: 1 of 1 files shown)


bwiernik:
I think this is fine

POBrien333:
Goes with https://github.com/citation-style-language/journals/pull/147#event-16047254976

PR #7399, opened 2025-01-22
Update animal-conservation.csl

POBrien333:
https://forums.zotero.org/discussion/comment/484124/#Comment_484124

Current paper: https://zslpublications.onlinelibrary.wiley.com/doi/epdf/10.1111/acv.12948


adam3smith:
Thanks!

PR #7400, opened 2025-01-22
Update ruby gems

adam3smith:


adam3smith:
I'm going to try merging this, I think the CI is running from master

PR #7403, opened 2025-01-24
Create the-british-journal-for-the-history-of-science.csl

POBrien333:
https://forums.zotero.org/discussion/115215/style-request-the-british-journal-for-the-history-of-science#latest


adam3smith:
Thanks!

PR #7407, opened 2025-01-27
Create etudes-chinoises.csl

PotatoSinology:
Create CSL for journal Études chinoises, published by les Press de l'INALCO.

NOTE: CSL 1.0.2 documentation instructs to italicize et-al by adding it to <names> (see https://docs.citationstyles.org/en/stable/specification.html#et-al), and that works, but the CSL style validator does not like it.


bwiernik:
Thanks. There is a bug in the validator where `et-al` isn't accepted if it follows `substitute` for some reason. Moving it before passes validation.

One last thing is that we need a link to style documentation. I couldn't see any author guidelines or style documentation on the INALCO website. Could you give a link to the documentation in `rel="documentation"` link? If the guidelines aren't published online, could you upload a copy of them to this thread?

PR #7411, opened 2025-01-29
Bluebook Law Review: add court to case citations

adrianparsons:
Adds the court before the date of the decision in parenthesis. For example, "D. Mass" in: 

> Ward v. Reddy, 727 F. Supp. 1407, 1412 (D. Mass. 1990).

This format is shown in the ["Cases" section on the Bluebook site](https://www.legalbluebook.com/bluebook/v21/quick-style-guide#:~:text=Cases) (publicly accessible).

More specifically, the Bluebook [states in B10.1.3](https://www.legalbluebook.com/bluebook/v21/bluepages/b10-cases/b10-1-full-citation) (subscription required):

> B10.1.3 – Court and Year of Decision
> Indicate parenthetically the deciding court followed by the year of decision (immediately following the page reference). When citing decisions of the United States Supreme Court, however, do not include the name of the deciding court. [Table T1](https://www.legalbluebook.com/bluebook/v21/tables/t1-united-states-jurisdictions) lists the correct abbreviations for courts in U.S. jurisdictions.
>  
> (i) The United States Supreme Court: Cite United States Reports (U.S.) if the opinion appears therein; otherwise, cite to Supreme Court Reporter (S. Ct.):
>  
> Meritor Sav. Bank v. Vinson, 477 U.S. 57, 60 (1986).
>  
> Weyerhaeuser Co. v. U.S. Fish & Wildlife Serv., 139 S. Ct. 361, 365 (2018).
>  
> (ii) Federal Courts of Appeals: Cite Federal Reporter (F., F.2d, F.3d) and indicate the name of the court parenthetically:
>  
> Env't Def. Fund v. EPA, 465 F.2d 528, 533 (D.C. Cir. 1972).
>  
> United States v. Jardine, 364 F.3d 1200, 1203 (10th Cir. 2004).
>  
> (iii) Federal District Courts: Cite Federal Supplement (F. Supp., F. Supp. 2d, F. Supp. 3d) and indicate the name of the court parenthetically:
>  
> W. St. Grp. LLC v. Epro, 563 F. Supp. 2d 84, 91 (D. Mass. 2008).
>  
> Harris v. Roderick, 933 F. Supp. 977, 985 (D. Idaho 1996).
> 
> 

This is similarly documented in [The Indigo Book](https://law.resource.org/pub/us/code/blue/IndigoBook.html#R12)

### DIFF
--- a/bluebook-law-review.csl
+++ b/bluebook-law-review.csl
@@ -212,6 +212,14 @@
               </date>
             </group>
           </if>
+          <else-if type="legal_case">
+            <group delimiter=" ">
+              <text variable="authority"/>
+              <date variable="issued">
+                <date-part name="year"/>
+              </date>
+            </group>
+          </else-if>
           <else>
             <group delimiter=", ">
               <text macro="editor-translator"/>


adrianparsons:
I'm not a legal scholar; I helped a friend correct this locally and I wanted to send the fix upstream. So, I'll ping some recent editors to this file to make sure I'm not breaking something (apologies in advance!): @POBrien333 @trackleft @adam3smith

adam3smith:
Thanks yes, this looks good (with the exception of SCOTUS cases, but oh well)

adrianparsons:
> Thanks yes, this looks good (with the exception of SCOTUS cases, but oh well)

Yeah... not sure the best way to deal with that. In any case, thanks!

Fruitbat:
The problem with "deciding court" is that it requires logic that csl doesn't offer.  The Court meta-data field should be the proper name of the Court, not the Bluebook abbreviation as listed in [Table T1](https://www.legalbluebook.com/bluebook/v21/tables/t1-united-states-jurisdictions).  Many sources of case information already provide that (e.g. CourtListener). Putting the Bluebook abbreviation in the Court field means that any other style no longer has access to the proper name.  The abbreviations (I.e. Table T1) should go in bluebook-law-review.csl and be determined by the style.

The logic required is not trivial.  When the "deciding court" is the United States Supreme Court, there can be a number of different ways to reference it (1 Wheat), (1 Cranch), an empty field, etc.  Generally, the field is omitted entirely for the Supreme Court.  So, putting the abbreviation in the Court field means deleting "United States Supreme Court" entirely from the meta-data of any legal_case you cite.  State Supreme courts also often have special rules about how to list "deciding court".    Again, putting the proper name of the court in Court would (theoretically) allow the style to translate it properly.

Unfortunately, CSL doesn't allow conditionals based on the content of fields, only whether the field *has* content.  I guess the thinking is that allowing styles to have conditionals would create too many issues.

There are a couple of alternatives, but these alternatives would require extensions to the csl processor, Zotero, or both:
1. Some sort of key/value pair function, where the proper name of a court can map to the Bluebook abbreviation could work.  The Supreme Court would need additional lookups based on other fields, but key/value pairs could support the logic.
2. A new meta-data field for the "legal_case" type, analogous to "Journal Abbr" field for the Journal Article type, a "Court Abbr" field could be added to the meta data.  This would allow the Court field to continue to contain proper meta-data, but allow a hack similar to this.

Personally, I think 1. is a much better solution.  It would be fast.  And the list of proper Court Names and their various abbreviations exist, so if the CSL key/value function existed, any style could implement a lookup.

2. would not require extending CSL, so it might be a more pragmatic solution.  the list of proper names and abbreviations could be used to programmatically overwrite the "Court Abbr" field as needed.   It's kludgy, but it could be made to work.

adrianparsons:
Wow, thanks for that background, @Fruitbat!

PR #7418, opened 2025-01-31
MHRA: Update further for 4th edition and refactor

adunning:
- Implement further changes from the MHRA Style Guide, 4th edition, notably the number of authors used for et-al abbreviation and the number range abbreviation format.
- Improve the naming of macros for better consistency and clarity: macros appear in the order of the bibliography, with dependent macros appearing immediately before their first usage.
- Enhance the handling of contributors, titles, and publication details.
- Convert all affixes where possible to delimiters.


bwiernik:
Looks good to me. Ready to merge?

adunning:
Thank you! There is more that could be done with it, but I think this is good enough for now.

PR #3053, opened 2017-10-16
Create uoy-modern-language-association-8th-edition-1.csl

AVWoman:
University of York MLA 8th style

PR #7426, opened 2025-02-04
Clean up University of York third-party styles

adunning:
Delete outdated University of York styles that do not materially modify other style guides. The updated pages at <https://subjectguides.york.ac.uk/referencing-style-guides> suggest that these are meant to make these styles easier to use, and not to create alternative definitions. For example, their MLA page currently points students to the 9th edition rather than the 8th edition defined here.


adam3smith:
These were originally created 8 years ago in https://github.com/citation-style-language/styles/pull/3053 and they explicitly requested separate styles. Wondering if @AVWoman is still around at York to speak to this (or have you talked to anyone at York by chance @adunning ?)

AVWoman:
As it happens we're just looking at these. At some point in the summer we do essentially want to 'get rid' of the duplicate styles and just retain the three Harvard ones. Our thinking was to change the dupes to independent-parent 'redirects', as we're conscious that people are still using the styles and we were keen to avoid things breaking anywhere. But if that's not the best approach then any advice would be gratefully received!

adunning:
Perfect – this will redirect everything.

bwiernik:
Closing because individual PRs were submitted converting these to dependent styles

PR #7430, opened 2025-02-06
Create australian-government-style-documentary-note.csl

CircularDog:
This style is consistent with Australian Government Documentary Note style 2024, including citation and bibliography styles in correct format. 

It is derived from the AIATSIS Documentary Note style by user @adam3smith, but serves a different purpose.


POBrien333:
Generally, we'd want a link to a styles documentation to include it in this repository as explained [here](https://github.com/citation-style-language/styles/blob/master/STYLE_REQUIREMENTS.md#csl-style-repository-requirements). Can you add that?

PR #7443, opened 2025-02-20
Create apa-single-spaced-no-url.csl

Sciurus365:
This is modified from `apa-single-spaced.csl`. No URLs are shown with this style (to save more space).

### DIFF
--- a/apa-single-spaced-no-url.csl
+++ b/apa-single-spaced-no-url.csl
@@ -1681,27 +1681,7 @@
     </group>
   </macro>
   <!--
-  <macro name="access">
-    <choose>
-      <if variable="DOI" match="any">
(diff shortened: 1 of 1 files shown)


adam3smith:
What is this for? We generally only accept styles that follow published style guides or journal requirements -- everything else you're welcome to host yourself (Zotero, e.g., will offer to install styles hosted on github)

bwiernik:
Thanks. The variations hosted in the repo are generally used by several journals or publishers (for example, many non-English journals use the "no ampersand" format). We try to avoid proliferation of APA variants (which create substantial maintenance burden), so I suggest you host this style on another repo or website (for example, I host several variants here: https://github.com/bwiernik/zotero-tools)

PR #7450, opened 2025-02-25
Update american-college-of-clinical-pharmacy.csl

POBrien333:
https://forums.zotero.org/discussion/122125/style-error-american-college-of-clinical-pharmacy#latest


adam3smith:
Well still need to remove the et al. for in text. I think setting and others to an empty string is the best way of doing that?

adam3smith:
No, this doesn't work because it's used in the bibliography -- that's why I suggested using a different style

PR #7452, opened 2025-02-27
Updated OSCOLA CSL style that uses journal abbreviations, if they are…

aarontimo:
… available. If no journal abbreviation is present, then the full title appears.

### DIFF
--- a/oscola-no-ibid.csl
+++ b/oscola-no-ibid.csl
@@ -10,10 +10,13 @@
     <author>
       <name>Sebastian Karcher</name>
     </author>
+    <contributor>
+      <name>Brenton M. Wiernik</name>
+    </contributor>
     <category citation-format="note"/>
     <category field="law"/>
     <summary>The OSCOLA Standards with no ibid. For a Zotero Group showing data-entry in Zotero see: https://www.zotero.org/groups/oscola_samples/items/order/itemType</summary>
-    <updated>2013-11-15T01:14:57+00:00</updated>
+    <updated>2025-08-26T01:14:57+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale>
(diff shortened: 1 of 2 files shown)


aarontimo:
Sorry - not sure why this is failiing. I used the CSL Validator 1.0.2, strict 1.0.2 and 1.0.1 and it passed. No issues. I used the CSL Visual Editor to make the changes. Why is it soooo hard to update/improve styles??

adam3smith:
It's not your style that's failing, something else is going on

adam3smith:
Your pull request is based on an ancient, unsynced fork, so you're picking up a bunch of issues with style in your fork that still follow CSL 1.0.1. 

See if you can sync the fork using the button on https://github.com/aarontimo/styles/tree/update-oscola-style
image

adam3smith:
But I'm actually not sure that we want this change (though thanks for submitting it!)-- the current version is deliberate:
OSCOLA allows the use of full journal titles. Automated abbreviations will, by default, come out wrong for users, e.g., of Zotero because the default is to use automated MEDLINE style journal abbreviations -- so we'll start by confusing users who aren't aware of that distinction. Then, many journals automatically import journal abbreviations into reference managers -- again, those would most likely not be conformant to OSCOLA abbreviations: again, we confuse users. Finally there may be cases when there are good reasons to keep journal abbreviations in the metadata but not use OSCOLA journal abbreviations: again, this version of the style would make this impossible. 
Given all this (and the fact that I dont' think we ever got a request for this change), I think we'll want to leave the official version as is. You're obviously more than welcome to use and even distribute yourself the version which defaults to abbreviations.

aarontimo:
Fair point. My experience with the old style was that none of the legal journals in my collection had journal abbreviations imported from Medline. Does Medline even cite legal journals? I don’t know. It didn’t appear to do so for me. I also had to resubmit to a journal using OSCOLA because I was using full title, not an abbreviation.

I’m not fussed though. Like you said, I can use it. Just thought it might save others some time.

Maybe it could be offered like another version of OSCOLA - I recall seeing one with ’no ibid’ (or something similar)...

adunning:
I am following up on the requests for OSCOLA, since I need to integrate its code into the New Hart's Rules styles.

Using the standard journal abbreviation field has been requested at least three times:

https://forums.zotero.org/discussion/104847/oscola-style-error/

https://forums.zotero.org/discussion/122923/oscola/

https://forums.zotero.org/discussion/64432/oscola-journal-abbreviation/

This change would allow the user to retain the full title of the journal and create a custom abbreviations file or fill in the correct abbreviation in the relevant field. Manual intervention is required whether the abbreviation field is used or not, so I would like to propose implementing this.

bwiernik:
@adam3smith @adunning There are some comments in the style about adding `jurisdiction` when it's available. It's now available in 1.0.2 — could one of you take a look and adjust as needed? I don't have enough context on this style to know what to edit.

PR #7455, opened 2025-02-27
Update modern-humanities-research-association-author-date.csl to 4th …

POBrien333:
…edition

https://forums.zotero.org/discussion/comment/486630/#Comment_486630

Following note style PR: https://github.com/citation-style-language/styles/pull/7007

Not checked yet. Just cobbled the old author-date with the new note-style together.


adunning:
Thanks for doing this. There are a lot of problems, however, in the note style, and I wouldn't recommend using the current version as a base. I have tried to address some of them in <https://github.com/citation-style-language/styles/pull/7418> but nobody has reviewed it, and in any case there is much more work to do.

In addition, MHRA are officially withdrawing the third edition, so I don't think it's necessary to retain the old version; the 3rd edition of the note style was deleted in <https://github.com/citation-style-language/styles/pull/7425>.

POBrien333:
Ah, that's good to know.
Will wait till it's merged and redo it.

bwiernik:
I merged the note PR. @adunning please feel free to keep updating in a new PR. @POBrien333 I think we can go ahead and update the author-date version based on note now and then make additional updates later as needed. Like @adunning says, we don't need to retain the 3rd edition style

PR #7461, opened 2025-03-04
Create journal-of-wound-management-EWMA.csl

BellavGFH:
Zotero reference style for Journal of Wound Management of the European Wound Management Association (https://ewma.org/jwm/).


kjklosowski:
Hi @BellavGFH. If you're still interested in fixing the issues to have the style published, try renaming the file to journal-of-wound-management.csl

[CSL Style Repository Requirements](https://github.com/citation-style-language/styles/blob/master/STYLE_REQUIREMENTS.md) section 4 states to drop text between parentheses in file name :)

POBrien333:
@BellavGFH
As the author guidelines state to follow "Use the style of the following examples, which are based with slight modification on the formats set forth in “Uniform Requirements for Manuscripts Submitted to Biomedical Journals,” also known as the “Vancouver” style for biomedical journals (JAMA 1993;269:22282-6)." 
I don't see that you made any specific edits to the style itself and it's just a copy of our vancouver.csl style.
Can you confirm if any changes are actually needed?
We can add this journal otherwise as a dependent style of vancouver.csl so authors can find the right style without having to maintain a separate style.

BellavGFH:
Hi @POBrien333 ,
TBH I'm not sure any more what I changed when I made the style when I wrote the manuscript. You are right that I based it on Vancouver, but AFAIK I did adapt something because the end result of Vancouver was off. Having said that though... I will be the Queen of Sheba if I remember what that was. I quickly ran both here, and I too can't see any difference (not that I don't believe you, just curious ;). So you are right:

> We can add this journal otherwise as a dependent style of vancouver.csl so **authors can find the right style** without having to maintain a separate style.

THX!

POBrien333:
Could I ask you to go [here](https://github.com/BellavGFH/styles/tree/patch-2) and click on the "Sync fork..." up top?
That will resolve the validation errors. :)

PR #7463, opened 2025-03-05
Create transactions-on-maritime-science.csl

POBrien333:
https://forums.zotero.org/discussion/122375/style-request-transactions-on-maritime-science#latest

paper: https://www.toms.com.hr/index.php/toms/article/view/698/582 (yeah, not in line with the guidelines, but have the feeling they're all just over the place, so I stuck with the guidelines)


adam3smith:
Another one -- these are all the ones I noticed

PR #7474, opened 2025-03-12
Update glossa.csl

jrw297:
This is an attempt to suppress the author middle initials in the bibliography citation in the style called glossa. How can I do it?


adam3smith:
CSL can't omit middle initials (also seems like a bit of a weird rule, I think)

PR #7477, opened 2025-03-13
Updates to USGS citation style to conform to "USGS Suggestions to Authors 8th edition"

baagaard-usgs:
Corrections based on "Suggestions to Authors of the Reports of the U.S. Geological Survey—The Official Style Guide of the U.S. Geological Survey, 8th Edition, Revised". This document is not yet publicly available, but I have verified that these style changes result in the correct formatting of references for datasets, software, journal articles, reports, and webpages.

- Add dataset and software types.
- Add DOIs to journal articles and reports using URL with accessed date.
- Report: Use genre when available.
- Webpage: Move institution to proper location.

### DIFF
--- a/us-geological-survey.csl
+++ b/us-geological-survey.csl
@@ -4,7 +4,7 @@
     <title>U.S. Geological Survey</title>
     <id>http://www.zotero.org/styles/us-geological-survey</id>
     <link href="http://www.zotero.org/styles/us-geological-survey" rel="self"/>
-    <link href="http://www.nwrc.usgs.gov/techrpt/sta28.pdf" rel="documentation"/>
+    <link href="https://atthecore.usgs.gov/media/files/suggestions-authors-reports-us-geological-survey-official-style-guide-us-geological" rel="documentation"/>
     <author>
       <name>Rintze Zelle</name>
       <uri>http://twitter.com/rintzezelle</uri>
@@ -15,9 +15,12 @@
     <contributor>
       <name>Jason Bellino</name>
     </contributor>
+    <contributor>
+      <name>Brad Aagaard</name>
+    </contributor>
     <category citation-format="author-date"/>
     <category field="geology"/>
-    <updated>2017-05-05T11:43:00+00:00</updated>
+    <updated>2025-03-03T10:06:00+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="editor-translator">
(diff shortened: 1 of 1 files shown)


adam3smith:
You really want that rule for DOIs? That seems quite unnecessary and at odds with pretty much all other style guides. When will the new guidelines be published? We've made bad experiences (in the form of lots of complaints) publishing style updates before they went public

adam3smith:
Also, the way you have coded webpage doesn't look robust for (commonly) missing data -- check the auto generated example. Can you take a look?

baagaard-usgs:
The 8th edition of Suggestions to Authors is dated 2020 with the revised version dated August 2024. I don't know why the 8th edition is not public, but the citation styles appear stable.

The DOIs for all publication types are URLS (https://doi.org/DOI) accompanied by the data accessed.

I will check the webpage layout.

adam3smith:
Oh, got it, misunderstood the 'not yet publicly available' part.

us-geological-survey.csl
```diff
@@ -195,6 +230,84 @@
...
+                <else>
+                  <text term="at"/>
+                  <text variable="URL" prefix=" "/>
+                </else>
+              </choose>
+            </group>
+          </group>
+        </else-if>
+        <else-if type="software">
+          <group suffix=":">
```
bwiernik, line 281: You should almost never include punctuation like this as a prefix or suffix, as it will show even if there is no text following. Use "delimiter" instead.

bwiernik (changes requested):
Left a comment on one line about using delimiter instead of prefix/suffix. Can you please make that change throughout the style?

us-geological-survey.csl
baagaard-usgs: The format is `[title] [version]:`. USGS style is to either include the version in the title or include it in the version field. I want the colon after version if both are given and after the title if just the title is given. Isn't it safe to require a title?

baagaard-usgs:
Here is the code I have for typesetting the `title` and `container-title` for a web page. Zotero is using the `container-title` if the `title` is empty, which is what I thought the problem was in typesetting the "CSL search by example" case in the test.

In the test case, is `title` empty? What field is providing "Citation Style Editor"?

```xml
	    <choose>
	      <if variable="title">
		<text variable="title"/>
		<choose>
		  <if variable="container-title">
		    <text variable="container-title" prefix=": "/>
		  </if>
		</choose>
	      </if>
	      <else-if variable="container-title">
	      </else-if>
	    </choose>
```

us-geological-survey.csl
bwiernik: You should never assume that any piece of data is present and always write the style to be robust to missing data. An untitled piece of software might have just a medium/genre and no title for example.

adam3smith:
This looks good. I think the issue was that title is used as a substitute element for author, so it gets omitted later on but still tests as true

PR #7486, opened 2025-03-19
Create zitierguide-leitfaden-zum-fachgerechten-zitieren-in-rechtswissenschaftlichen-arbeiten-5.csl

beahodel:
This style is an updated version of its 4th edition (http://www.zotero.org/styles/zitierguide-leitfaden-zum-fachgerechten-zitieren-in-rechtswissenschaftlichen-arbeiten)


bwiernik:
Can you please update the existing style rather than making a new one?

If needed, we can retain the previous edition as a new style (with 4th edition in the title and style ID), but we should only keep it if absolutely necessary.

beahodel:
Ok. Sorry for the inconvenience. I updated the existing style.

beahodel:
This is the update of the existing style: https://github.com/citation-style-language/styles/pull/7494

PR #7488, opened 2025-03-19
Adding PMID macro to AMA.csl

wmhorne:
AMA allows PMIDs to be included in references when desired.

This PR attempts to add a PMID to the end of the reference when a PMID is present.

In Zotero, this works as expected:

Screenshot 2025-03-19 at 3 47 51 PM

### DIFF
--- a/american-medical-association.csl
+++ b/american-medical-association.csl
@@ -84,6 +84,14 @@
       </if>
     </choose>
   </macro>
+<macro name="pmid">
+  <choose>
+    <if variable="PMID">
+      <text value="PMID: "/>
+      <text variable="PMID"/>
+    </if>
+  </choose>
+</macro>  
   <macro name="title">
     <choose>
       <if type="bill book graphic legal_case legislation motion_picture report song thesis" match="any">
@@ -284,6 +292,7 @@
         </else>
       </choose>
       <text prefix=" " macro="access"/>
+      <text prefix=". " macro="pmid"/>
     </layout>
   </bibliography>
 </style>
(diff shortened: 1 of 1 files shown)


bwiernik:
Can you quote the manual where it talks about including PMID? I don't currently have access to a copy of the AMA manual

wmhorne:
While not exactly a rule, it's mentioned under the Electronic References section:

Screenshot 2025-03-19 at 3 51 53 PM

Screenshot 2025-03-19 at 3 58 08 PM

Screenshot 2025-03-19 at 3 51 57 PM

Additionally, it's visible in this example citation when searching [the AMA site](https://academic.oup.com/amamanualofstyle/search-results?page=1&q=pmid&fl_SiteID=6376&SearchSourceType=1&allJournals=1): 


Screenshot 2025-03-19 at 4 00 53 PM

wmhorne:
PR: https://github.com/citation-style-language/styles/pull/7488

adam3smith:
Thanjs for this! We certainly shouldn't take this as part of the main AMA style -- that's used for lots of journals and I've never seen PMIDs in those style guides. 
I also think it's pretty clear from the AMA guide that you screenshot that PMIDs are not to be included -- they say they are usually used behind the scenes. If there's a clear general use case, I think we can consider a variant, but I'd want to see the need (as opposed to preferences) clearly established. FWIW, I don't think anyone has ever asked for or about this

wmhorne:
@adam3smith makes sense—I will make this change in a custom csl.

For what it's worth, does the approach I took seem like a good way to do it?

adam3smith:
Yes, that will work perfectly. FWIW, you can do this in a single line with exactly same output:

`<text variable="PMID" prefix=". PMID:"/>`

PR #7494, opened 2025-03-21
Update zitierguide-leitfaden-zum-fachgerechten-zitieren-in-rechtswissenschaftlichen-arbeiten.csl

beahodel:
Adaptations for 5th edition

### DIFF
--- a/zitierguide-leitfaden-zum-fachgerechten-zitieren-in-rechtswissenschaftlichen-arbeiten.csl
+++ b/zitierguide-leitfaden-zum-fachgerechten-zitieren-in-rechtswissenschaftlichen-arbeiten.csl
@@ -5,16 +5,19 @@
     <id>http://www.zotero.org/styles/zitierguide-leitfaden-zum-fachgerechten-zitieren-in-rechtswissenschaftlichen-arbeiten</id>
     <link href="http://www.zotero.org/styles/zitierguide-leitfaden-zum-fachgerechten-zitieren-in-rechtswissenschaftlichen-arbeiten" rel="self"/>
     <link href="http://www.zotero.org/styles/juristische-zitierweise-schweizer" rel="template"/>
-    <link href="http://www.worldcat.org/oclc/958980496" rel="documentation"/>
+    <link href="http://www.worldcat.org/oclc/1393619644" rel="documentation"/>
     <author>
       <name>Pascal Gutknecht</name>
       <email>literaturverwaltung@zhaw.ch</email>
     </author>
+    <contributor>
+      <name>Beatrice Hodel</name>
+      <email>literaturverwaltung@zhaw.ch</email>
+    </contributor>
     <category citation-format="note"/>
     <category field="law"/>
-    <summary>This style is based on Roger Müller, ZitierGuide Leitfaden zum fachgerechten Zitieren in rechtswissenschaftlichen Arbeiten, 4. Auflage, Zürich 2016 (ISBN 978-3-7255-7583-1). 
-	It is an adaption of the style "Juristische Zitierweise Schweizer" by Stephan Schlegel (https://www.zotero.org/styles/juristische-zitierweise-schweizer), which is based on the style "Juristische Zitierweise (Stüber)" by Oliver Wolf and Philipp Zumstein (https://www.zotero.org/styles/juristische-zitierweise).</summary>
-    <updated>2023-04-01T14:27:30+00:00</updated>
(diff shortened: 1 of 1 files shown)


bwiernik:
Thank you. If you need to make additional edits, please make them in this PR, rather than closing and opening a new one.

beahodel:
Sorry about that. I managed to add my edits here.

zitierguide-leitfaden-zum-fachgerechten-zitieren-in-rechtswissenschaftlichen-arbeiten.csl
```diff
@@ -121,28 +124,10 @@
...
-                      <else>
-                        <group delimiter="">
-                          <date date-parts="year" form="text" variable="issued" prefix="(" suffix=")"/>
-                          <text variable="page-first" prefix=" S. " suffix=" ff"/>
-                        </group>
-                      </else>
-                    </choose>
+                    <group delimiter="">
+                      <date date-parts="year" form="text" variable="issued"/>
+                      <text variable="page-first" prefix=" S. " suffix=" ff"/>
```
POBrien333: Please use the correct label here instead of hardcoding the label as a prefix.

zitierguide-leitfaden-zum-fachgerechten-zitieren-in-rechtswissenschaftlichen-arbeiten.csl
```diff
@@ -83,7 +86,7 @@
         <text macro="author-inline-citation"/>
         <choose>
           <if match="any" disambiguate="true">
-            <date date-parts="year" form="text" variable="issued"/>
+            <text variable="note"/>
```
POBrien333: What does this try to achieve?
We very very rarely use the "note" field in styles.

POBrien333:
I see a coule of instances where "(Hrsg.)" is hardcoded. Please use the correct label instead.
Also, please remove any instances for text-format="normal" if not necessary. (usually this is only ever used when wanting to for example unbold a part of something)

POBrien333:
@beahodel 
To remove the validation errors, please go here and click on "Sync fork": https://github.com/beahodel/styles/tree/patch-5

POBrien333:
@beahodel 
Please see the comments I had left recently. They would still need addressing.

beahodel:
Hi @POBrien333 , which comments exactly do you mean? I thought I removed the hardcoding of "(Hrsg.)" and the "normal" formating...

POBrien333:
Do you see this?
image

adam3smith:
If it says "pending" they don't. There's a 'finish review' button you need to click.

POBrien333:
Ahhhh! That explains that. Thanks :)

zitierguide-leitfaden-zum-fachgerechten-zitieren-in-rechtswissenschaftlichen-arbeiten.csl
beahodel: If several works by the same author are cited in the text, this citation style requires that a keyword from the title be given for these works instead of the year. I am aware that this is unusual and somewhat of a hack. Is there another way to do this?

adam3smith: Why not use `<text variable="title" form="short"/>` for the short title?

beahodel:
Hi. Is there anything else I should change, or is it ok like that?
Thank you

POBrien333:
@beahodel 
Webpages are currently not rendered properly. If you input your style on https://editor.citationstyles.org/codeEditor/ and choose the webpage "CSL search by example" item you'll see it doesn't render anything in-text and in the bibliography only the title is output.
Without access to the guidelines, which we try to avoid, it's very hard for us to actually help with styles.

beahodel:
Thank you for your quick reply. 
For websites this style requires not the exact URL but the URL of the relevant main website with a chain of links that was clicked on to get to the quoted text passage. The example in the style sheet says "www.admin.ch (Bundesrecht/Vernehmlassungen/Abgeschlossene Vernehmlassungen/2009/Verordnung zum Schutz vor Passivrauchen - Bericht), besucht am 11.08.2022" 
This was already solved in the previous version of this style in such a way that this text can be specified in the abstract. We provide additional information to our students on how to do this in Zotero.

POBrien333:
Would you mind copying that page from the guidelines here? I think we could find a better approach.

beahodel:
Hi. For copyright reasons, I am reluctant to upload the pages here. You can access them via this [link](https://drive.switch.ch/index.php/s/R9zRzQZfcYoBwbx).
I have made adjustments so that if there is no information in the abstract, the URL is displayed.

POBrien333:
Great. thanks a lot. Now I understand.
I changed the style to use the "annote" variable.
That can easily be defined in the extra field like so:
annote: https://www.helbing.ch

That way the abstract field which is used a lot can be used as normal without any issues.

beahodel:
Very cool! Thank you!
How long do you think it will take before the style is published?

POBrien333:
It should be available on the repository within the hour.
Thanks for your efforts with the style.

PR #7500, opened 2025-03-28
Update revista-ladinia.csl

schtailmuel:
format book edition after year


POBrien333:
I've simplified this approach as the conditional is not needed.
I also added the edition in-text since that's specified in the guidelines: `Indicate the number of editions by superscripting the year (e.g. BELARDI 20032`

Can you please head to https://github.com/schtailmuel/styles/tree/patch-3 and click on "Sync fork..." up top? That will remove the validation errors that are a bug.

schtailmuel:
> I've simplified this approach as the conditional is not needed. I also added the edition in-text since that's specified in the guidelines: `Indicate the number of editions by superscripting the year (e.g. BELARDI 20032`

thanks!

> Can you please head to https://github.com/schtailmuel/styles/tree/patch-3 and click on "Sync fork..." up top? That will remove the validation errors that are a bug.

done

POBrien333:
Thanks. Could you clarify in which languages the journal accepts articles in?
Ideally we add a polyglot line to the style.

schtailmuel:
> Thanks. Could you clarify in which languages the journal accepts articles in? Ideally we add a polyglot line to the style.

the journal publishes in Ladin, Italian, German and English.
You can consult the index here: https://www.micura.it/de/index-ladinia/ladinia

POBrien333:
Thanks.
The style is merged now and should be available now.

PR #7506, opened 2025-04-06
Update acta-ichthyologica-et-piscatoria.csl

teodorgeorgiev:
The journal changed the publisher and now uses a different citation style


POBrien333:
Thanks!

PR #7510, opened 2025-04-09
APA: Compartmentalize macros

adunning:
Facilitate the sharing of macros between APA and other styles by compartmentalizing macro functionality, separating the logic as much as possible. Provide additional headings to make the file arrangement more explicit.

Changes:

- Filter out personal correspondence without a means of access from the bibliography (based on #3005)
- Provide missing delimiter in author listing and simplify using default CSL settings (APA example 87)
- Do not combine script-writer and director in-text, providing slightly less broken results with APA example 87 (now matching at least one of the examples, though not both)
- Do not provide 'no.' label with a non-numeric `number` on a broadcast (APA example 87)
- Provide an alternative for the missing CSL `part-number` and `supplement-number` labels
- Support `chapter-number` in case of missing chapter page numbers
- Apply criteria for unpublished letters consistently; fixes example 402
- Provide `collection-title` for periodicals (e.g. new series, 2nd series)
- Provide `publisher-place` as a fallback in case of missing publisher (not specified in APA but necessary for pre-1900 books)

Tested against the APA 7th ed. collection in the Test Items Library at <https://www.zotero.org/groups/2205533/test_items_library/collections/MR2N872S> using Zotero, by generating HTML citations and bibliography using both the previous version and these files and comparing the results.

Variants generated using <https://github.com/adunning/csl-builder>.

### DIFF
--- a/apa-annotated-bibliography.csl
+++ b/apa-annotated-bibliography.csl
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" page-range-format="expanded">
+<style xmlns="http://purl.org/net/xbiblio/csl" and="symbol" class="in-text" demote-non-dropping-particle="never" initialize-with=". " names-delimiter=", " page-range-format="expanded" version="1.0">
   <info>
     <title>American Psychological Association 7th edition (annotated bibliography)</title>
-    <title-short>APA (annotated bibliography)</title-short>
+    <title-short>APA Style (annotated bibliography)</title-short>
     <id>http://www.zotero.org/styles/apa-annotated-bibliography</id>
     <link href="http://www.zotero.org/styles/apa-annotated-bibliography" rel="self"/>
     <link href="http://www.zotero.org/styles/apa" rel="template"/>
@@ -10,48 +10,53 @@
     <author>
       <name>Brenton M. Wiernik</name>
       <email>zotero@wiernik.org</email>
+      <uri>https://orcid.org/0000-0001-9560-6336</uri>
     </author>
+    <contributor>
+      <name>Andrew Dunning</name>
+      <uri>https://orcid.org/0000-0003-0464-5036</uri>
+    </contributor>
     <category citation-format="author-date"/>
     <category field="psychology"/>
     <category field="generic-base"/>
+    <summary>Author–date system of the Publication manual of the American Psychological Association: The official guide to APA style (2020)</summary>
     <updated>2024-07-09T20:08:41+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
     <terms>
-      <term name="editortranslator" form="short">
-        <single>ed. &amp; trans.</single>
-        <multiple>eds. &amp; trans.</multiple>
-      </term>
-      <term name="editor-translator" form="short">
-        <single>ed. &amp; trans.</single>
-        <multiple>eds. &amp; trans.</multiple>
-      </term>
-      <term name="translator" form="short">trans.</term>
-      <term name="interviewer" form="short">
-        <single>interviewer</single>
-        <multiple>interviewers</multiple>
-      </term>
-      <term name="collection-editor" form="short">
+      <term name="ad"> C.E.</term>
+      <term form="long" name="at">before the</term>
+      <term name="bc"> B.C.E.</term>
+      <term form="short" name="circa">ca.</term>
+      <term name="collection">archival collection</term>
+      <term form="short" name="collection-editor">
         <single>ed.</single>
         <multiple>eds.</multiple>
       </term>
-      <term name="performer" form="verb">recorded by</term>
-      <term name="circa" form="short">ca.</term>
-      <term name="bc"> B.C.E.</term>
-      <term name="ad"> C.E.</term>
-      <term name="issue" form="long">
+      <term form="short" name="editor-translator">
+        <single>ed. &amp; trans.</single>
+        <multiple>eds. &amp; trans.</multiple>
+      </term>
+      <term form="short" name="editortranslator">
+        <single>ed. &amp; trans.</single>
+        <multiple>eds. &amp; trans.</multiple>
+      </term>
+      <term form="verb" name="hearing">testimony of</term>
+      <term form="short" name="interviewer">
+        <single>interviewer</single>
+        <multiple>interviewers</multiple>
+      </term>
+      <term form="long" name="issue">
         <single>issue</single>
         <multiple>issues</multiple>
       </term>
-      <term name="software">computer software</term>
-      <term name="at" form="long">before the</term>
-      <term name="collection">archival collection</term>
+      <term form="verb" name="performer">recorded by</term>
       <term name="post">online post</term>
-      <term name="at" form="long">before the</term>
-      <term name="hearing" form="verb">testimony of</term>
-      <term name="review-of" form="long">review of the</term>
-      <term name="review-of" form="short">review of</term>
+      <term form="long" name="review-of">review of the</term>
+      <term form="short" name="review-of">review of</term>
+      <term name="software">computer software</term>
+      <term form="short" name="translator">trans.</term>
     </terms>
   </locale>
   <locale xml:lang="da">
@@ -71,7 +76,7 @@
   </locale>
   <locale xml:lang="fr">
     <terms>
-      <term name="editor" form="short">
+      <term form="short" name="editor">
         <single>éd.</single>
         <multiple>éds.</multiple>
       </term>
@@ -97,29 +102,26 @@
       <term name="et-al">et al.</term>
     </terms>
   </locale>
-  <!-- For Indigeneous Knowledge, assume the item is stored as `document` or `speech`
-       and that Nation/Community, treaty, where the Elder lives, 
-       and topic are all stored in `title`
-       cf. https://libguides.norquest.ca/c.php?g=314831&p=5188823.
-       If the item is stored as `interview`, assume that Nation/Community, treat, and topic
-       are stored in `title`, 'Oral teaching' or similar is stored in `archive`, and where
-       the Elder lives is stored in `archive-place`. 
-  -->
-  <!-- Reviews are detected if an item has type `review` or `review-book` or if it has any of the variables
-       `reviewed-title`, `reviewed-author`, or `reviewed-genre`. For the latter case, reviews are commonly 
-       stored as types `article-journal`, `article-magazine`, `article-newspaper`, `post-weblog`, or `webpage` 
-  -->
-  <!-- General categories of item types:
-       Periodical: article-journal article-magazine article-newspaper periodical post-weblog review review-book
-       Periodical or Booklike: paper-conference
-       Booklike: article book broadcast chapter classic collection dataset document
-                 entry entry-dictionary entry-encyclopedia event figure
-                 graphic interview manuscript map motion_picture musical_score
-                 pamphlet patent performance personal_communication post report
-                 software song speech standard thesis webpage
-       Legal: bill hearing legal_case legislation regulation treaty
+  <!-- General categories of CSL item types:
+
+       Periodical
+       : article-journal article-magazine article-newspaper periodical post-weblog review review-book
+
+       Periodical or Booklike
+       : paper-conference
+
+       Booklike
+       : article book broadcast chapter classic collection dataset document
+         entry entry-dictionary entry-encyclopedia event figure
+         graphic interview manuscript map motion_picture musical_score
+         pamphlet patent performance personal_communication post report
+         software song speech standard thesis webpage
+
+       Legal
+       : bill hearing legal_case legislation regulation treaty
   -->
   <!-- Equivalencies:
+
        classic == book
        document == report, but give full date
        standard == report
@@ -128,6 +130,7 @@
        collection == book, but give full date
   -->
   <!-- Role equivalencies:
+
        compiler == editor
        organizer, curator == chair
        script-writer == director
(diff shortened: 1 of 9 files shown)


bwiernik:
> Filter out personal correspondence without a means of access from the bibliography (based on Properly support personal communications in APA #3005)

For the numeric versions, did you check if numbering is correct for other items with this omitted?

> Provide missing delimiter in author listing and simplify using default CSL settings (APA example 87)

Could you elaborate on this one?

> Do not combine script-writer and director in-text, providing slightly less broken results with APA example 87 (now matching at least one of the examples, though not both)

This area is definitely a mess without being able to combine creators, but I remember the previous logic being the least-worst solution I could come up with, so let me check a few cases. 

> Provide an alternative for the missing CSL part-number and supplement-number labels

In any cases where you do this, be sure to leave a comment in the code to remove once these labels are fixed in CSL

> Apply criteria for unpublished letters consistently; fixes example 402

Could you elaborate on this?

> Provide collection-title for periodicals (e.g. new series, 2nd series)

This will not work. collection-title includes things like special issue and special section tittles, which should not be included. "New series" and similar should probably just be stored as part of the container title 

> Provide publisher-place as a fallback in case of missing publisher (not specified in APA but necessary for pre-1900 books)

APA calls for this sort of situation to provide the publisher or source where the item is accessed. This might be a modern publisher of a reproduction or archive information https://apastyle.apa.org/blog/citing-classical-religious-works

bwiernik:
Thanks for looking into these! I have a couple of comments and questions above. There are a few things that won't work -- could you revert those?

adunning:
1. Good thinking on numeric – I've removed the filtering from those versions.
2. I've added `and="symbol" demote-non-dropping-particle="never" initialize-with=". " names-delimiter=", "` to the global options on `<style>`, and removed the settings from the individual `<names>` elements where possible (adding overrides where this modified the output of the test library). For example 87a, this corrects the output from 'Oakley, B., & Weinstein, J. (Writers)Lynch, J. (Director)' to 'Oakley, B., & Weinstein, J. (Writers), Lynch, J. (Director)'. It's still not the correct 'Oakley, B. (Writer), Weinstein, J. (Writer), & Lynch, J.', but at least it's readable.
3. I haven't changed the `script-writer director` line for bibliography entries, but at least for the items in the test library, the inline version works better without these combined.
4. I've flagged these with TODO comments.
5. One of the conditionals for personal correspondence left out `archive-place`; this was providing a citation of 'G. P. Zacharius, personal communication, August 15, 1953' where it should simply read 'Zacharius, 1953'.
6. In the few CSL implementations that exist, a 'special issue' appears to go under `genre` (that is certainly how APA deals with it) and a special section title under `volume-title`. By contrast a `collection-title` distinguishes between completely different runs of the same journal; it's fairly rare, but absolutely essential to identifying the right publication when it occurs. This information needs to be stored separately in the Series field to achieve a correct citation with Chicago. I've never seen Zotero populate this field automatically with citation data.
7. This is to deal with old books that do not have a publisher, for which one would normally instead cite the place of publication (for example, de Cervantes Saavedra 1605–1615 in the Test Items Library). As far as I can tell, the APA editors didn't think of this situation, so this is an improvement rather than strictly following the manual; let me know if you think it should be removed.

bwiernik:
At a high level and with some examples, can you walk me through how the compartmentalization facilitates reuse in Chicago/MHRA? 

I think there is a trade-off to using macros because too many can mean you have to do a lot of jumping around between parts of the style, which can make understanding the logic and making updates more difficult. My general approach to using [macros has been to primarily use them to avoid duplicate code. I'm not opposed to moving more things into macros vs. direct rendering, but I'd like to understand the benefit.

adunning:
Thanks for your thoughtful comment. I agree that there are trade-offs with macros, especially when it comes to readability and maintainability. My goal here has been to make the structure of the style more transparent—both to simplify updates and to make cross-style reuse more feasible.

The need for this approach became especially clear while working with the Chicago styles, which are unusually complex. These require at least half of the macros to be duplicated for notes and bibliography, which should be with only minor differences in punctuation or capitalization. In practice, the 17th-edition Chicago CSL styles often handle these differently in ways that aren’t intentional, and even more variation appears when compared with author-date. The main reason seems to be that the macros have grown so long and intricate that differences go unnoticed. By contrast, I’ve found that when macros are broken down into smaller, purpose-specific components, such inconsistencies become much more obvious and easier to fix.

This compartmentalization has also made it possible to reuse logic from APA in rewriting Chicago—something that initially seemed infeasible because of stylistic differences. Once the APA macros were broken into discrete, well-named units, it became clear how much could be adapted directly. Your more structured approach in APA turned out to be highly applicable once the internal logic was made more visible.

A few concrete examples:

- Many issues in Chicago and MHRA styles come down to subtle inconsistencies—such as using `<text>` instead of `<label>`, or switching between verb-short and short—that are hard to spot in longer macros. By introducing dedicated ‘labelling’ macros and grouping them together, it became much easier to apply consistent patterns and detect these issues.
- Rendering of reviews is another good case. APA handles these much more cleanly, but it wasn’t immediately obvious how transferrable that logic was to other styles until the APA review code was isolated in its own macro. Once that was done, it became easy to see where improvements could be made to Chicago.
- The `patent-number` macro is an example of how pulling shared functionality out of buried macros (currently within `parenthetical` in APA and `issued` in Chicago) exposes structural similarities that weren’t obvious before. This makes it much easier to reuse or debug across styles, not just within a single file.

This restructuring also supports broader CSL goals around modularity and style comparison (cf. [Extracting, analyzing, macros](https://discourse.citationstyles.org/t/extracting-analyzing-macros/1819) on the CSL forum).

Finally, I’ve included comment headings in the file to clarify the groupings. These become especially useful when using an XML editor like oXygen, where the outline view helps to visualize the structure. Despite the additions and the macro breakdown, this reorganized version is only four lines longer than the current one.

Let me know if you’d find it helpful for me to walk through one or two specific cases in more detail.

bwiernik:
Awesome thanks.

bwiernik:
For special issues, I am meaning the title of the special issue holding an article, which should not be cited. (Versus citing the whole special issue, which is what the manual gives an example of.) Let me think about this one and look at some data. Before 1.0.2, I used `collection-title` for this information (and currently Zotero stores such information there), but I see that I specified that `volume-title` should be used for this information the 1.0.2 specification. So I think collection-title should be okay to show here. 

For the original-date in absence of a publisher, is that appearing in parentheses at the end of the reference?

adunning:
Yes, I've added `original-title`, `original-publisher`, etc. in the parentheses within `publication-history` alongside `original-date`.

adunning:
It appears that <https://github.com/citation-style-language/styles/pull/7510/commits/7ce7ec6bc37b5f30fe58f2e3df75d2d00e936718> has broken the delimiters in `citeproc-ruby`, but it's not a problem elsewhere.

bwiernik:
I'm going to try to get this reviewed and merged Sunday. Could you make a list of substantive changes (eg, cases not covered by the APA manual)?

adunning:
I've updated the original comment to include all substantive changes.

bwiernik:
@adunning There seems to be a bunch of extraneous delimiters showing up in the examples

adunning:
This is due to the bug in `citeproc-ruby` that the guidelines refer to: <https://github.com/citation-style-language/styles/blob/704ff9ffba533dd67bb40607ef27514c2869fa09/REVIEW_GUIDELINES.md?plain=1#L20>.

bwiernik:
> Support collection-title for periodicals (e.g. new series, 2nd series)

Let's not do this one, as the most common example of `collection-title` is things like special issue and special section titles, which should not be included. 

> Allow description of an anonymous review as an author substitute

Can you show me what this means?

> Provide 'advance online publication' term for a journal article with a DOI and date but no volume or issue (APA example 7)

Like with Chicago and MHRA I think let's not assume this is the case. Better to err on not including text that could be incorrect 

> Hide labels for a non-numeric chapter, line, or verse locator to fit the requirements for a composite canonical reference (APA 8.13)

Could you show me what this means? 

> Handle uncertain dates in more cases

Out of curiosity, what does this mean? Didn't it always show uncertain dates if they were certain?

> Provide correct term for guest (APA 10.12)

What does this mean?

bwiernik:
Thanks for doing this! Just a few more questions. 

Also, do you have the HTML pages you generated with old style and the new one that I could look at?

bwiernik:
And lastly could you double check that these thesis examples are correct? https://forums.zotero.org/discussion/126552/bug-apa-7th-ed-formatting-of-doctoral-theses

adunning:
Here are the files before and after these changes for the bibliography:

[APA bibliography old.html](https://github.com/user-attachments/files/22537685/APA.bibliography.old.html)
[APA bibliography new.html](https://github.com/user-attachments/files/22537682/APA.bibliography.new.html)

Citations:

[APA citations old.html](https://github.com/user-attachments/files/22537687/APA.citations.old.html)
[APA citations new.html](https://github.com/user-attachments/files/22537686/APA.citations.new.html)

On your specific queries:

> the most common example of `collection-title` is things like special issue and special section titles, which should not be included.

The title of a special issue goes under `volume-title` according to the spec; I assume that a special section would go under `section`. By contrast `collection-title` is for something like '3rd ser.' which is essential for identifying a journal with multiple series, where volume numbers are reused (e.g. you might have the _CSL Journal_, 2nd ser., 2 and _CSL Journal_, 3rd ser., 2).

For an example of an anonymous review, see <https://www.zotero.org/groups/2205533/items/XZSUW6QK>.

I have removed the `advance-online-publication` term, but it's not possible to render this example correctly without using it.

> > Hide labels for a non-numeric chapter, line, or verse locator to fit the requirements for a composite canonical reference (APA 8.13)
> 
> Could you show me what this means?

This is the approach to locators discussed in <https://github.com/citation-style-language/schema/issues/451>.

> 
> > Handle uncertain dates in more cases
> 
> Out of curiosity, what does this mean? Didn't it always show uncertain dates if they were certain?

I've forgotten where the `is-uncertain-date` handling was missing; this might have been in relation to the missing `event-date`.

> > Provide correct term for guest (APA 10.12)
> 
> What does this mean?

APA calls for 'guest expert' rather than 'guest'.

bwiernik:
> The title of a special issue goes under volume-title according to the spec; I assume that a special section would go under section. By contrast collection-title is for something like '3rd ser.' which is essential for identifying a journal with multiple series, where volume numbers are reused (e.g. you might have the CSL Journal, 2nd ser., 2 and CSL Journal, 3rd ser., 2).

We have for many years stored special issue and special section title information in `collection-title`, so I'm very hesitant to make a change to `volume-title`. 

In terms of how data is stored in the wild, I would generally expect that "CSL Journal, 2nd Series" would be stored together in `container-title`.

adunning:
The CSL spec is specific about this function of `volume-title`: 'Also use for titles of periodical special issues, special sections, and the like'.

I can see why someone might try filling in the name of a special issue in `collection-title`, but they will have run into problems very quickly: even the styles for the 16th edition of Chicago (which came out in 2010) use the variable for a journal series. The spec defines the variable as 'Title of the collection holding the item', which is exactly the function of '3rd ser.' or similar for serials. This information cannot be stored with `container-title` because it is not part of the title and thus unitalicized in styles other than APA. Zotero has mapped `collection-title` to a 'Series' field with journal articles for at least fifteen years.

bwiernik:
> Also use for titles of periodical special issues, special sections, and the like

lol, I apparently wrote that and forgot haha. Thanks for pushing back!

bwiernik:
> Zotero has mapped collection-title to a 'Series' field with journal articles for at least fifteen years.

Yes, but "2nd series" and the like are extremely rare outside of some fields (primarily in humanities).  The Zotero field is mostly used to store special issue/section information historically. We should probably discuss migration options to volume-title with Dan Stillman and Sebastian at some point

adunning:
@bwiernik Have you reached any conclusion on this? I can remove the use of `collection-title` with serials if that's the only thing blocking this.

adunning:
@bwiernik I have suggested changing the `<title>` on this to 'APA Style' to match the name of the [official website](https://apastyle.apa.org), the _APA Style Manual_, the _Concise Guide to APA Style_, etc. I take it that this is a marketing strategy to expand the style's use beyond psychology. 'American Psychological Association' would still make sense as the main element of the official title of the _Publication Manual_; just let me know if you would prefer to stick with that.

apa-annotated-bibliography.csl
```diff
@@ -1377,6 +1377,9 @@
       <else-if type="periodical" variable="supplement-number">
         <text term="supplement" text-case="capitalize-first"/>
       </else-if>
+      <else-if type="periodical" variable="editor title">
+        <text term="special-issue" text-case="capitalize-first"/>
```
bwiernik, line 1344: Hmm, not sure about this one. Someone could be referencing a range of issues of the periodical, such as the tenure of a specific editor in chief. But I would imagine the most common case by far would be citing a special issue, so probably okay to keep here. (And in my example, this could be overridden by providing `genre`)

bwiernik: @adunning Any way to distinguish special issues from special sections automatically?

apa.csl
```diff
@@ -1,57 +1,60 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" page-range-format="expanded">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" demote-non-dropping-particle="never" initialize-with=". " names-delimiter=", " page-range-format="expanded" version="1.0">
+  <!-- This file was generated by the Style Variant Builder <https://github.com/citation-style-language/style-variant-builder>. To contribute changes, modify the template and regenerate variants. -->
   <info>
-    <title>American Psychological Association 7th edition</title>
-    <title-short>APA</title-short>
+    <title>APA Style 7th edition</title>
+    <title-short>Publication Manual of the American Psychological Association, with Bluebook</title-short>
```
bwiernik, line 6:
```suggestion
    <title-short>Publication Manual of the American Psychological Association, with Bluebook legal references</title-short>
```

Let's say "legal references" for clarity

bwiernik:
Hi @adunning, thank you for your long patience. Merging this now. I have a few questions—could you take a look when you have a moment?

1. Prioritize container-author over editors in citing parts of monographic items

The reason for the previous priority is that APA style blog posts have stated that the chapter format should be used for edited books, not for "authored books" and said that parts of "authored books" should be cited as the whole mongraph, with the specific book part referenced inline. I think that guidance doesn't make much sense and leads to some odd situations, so happy to ignore that. I can't think of any examples off hand where it would be problematic to prioritize container-author over editor. 


2. Always interpret archive-place as referring to physical archives, per the CSL specification, rather than attempting to use this for both physical and digital archives, which does not work if a reference consists only of archive and archive-place. Assume that a database reference will be in number.

The original logic was referring to `archive_location` as potentially database ID, not `archive-place`, correct? I think I've come around to just storing this information in `number` always, but could you give an example of where distinguishing physical vs. digital archives based on `archive_collection` or `archive-place` didn't work?


Thank you again for your monumental efforts here!

PR #7522, opened 2025-04-14

dstark:
Adjust originally-published grouping in the bibliography so that items without original-date (i.e., co-publications where one publisher is just as original as another) get a semicolon rather than a period before the second set of publication facts.

### DIFF
--- a/society-of-biblical-literature-fullnote-bibliography.csl
+++ b/society-of-biblical-literature-fullnote-bibliography.csl
@@ -33,7 +33,7 @@
     <category citation-format="note"/>
     <category field="theology"/>
     <summary>Society of Biblical Literature format with full notes and bibliography</summary>
-    <updated>2024-05-14T14:28:04+00:00</updated>
+    <updated>2024-08-27T19:51:26+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en-US">
@@ -1078,13 +1078,27 @@
               <text variable="genre" text-case="capitalize-first"/>
             </if>
           </choose>
-          <group delimiter=". ">
-            <text macro="originally-published"/>
-            <group delimiter=", ">
-              <text macro="reprint"/>
-              <text macro="publisher"/>
-            </group>
-          </group>
+          <!-- If no original-date is present, use a semicolon in the bibliography, rather than a period. Doing so allows original-publisher and original-publisher-place to be used to handle co-publication situations where both publishers are equally original. See SBLHS2 §4.1.4.4 and CMOS17 §14.140. -->
+          <choose>
+            <if variable="original-date">
+              <group delimiter=". ">
+                <text macro="originally-published"/>
+                <group delimiter=", ">
+                  <text macro="reprint"/>
+                  <text macro="publisher"/>
+                </group>
+              </group>
+            </if>
+            <else>
+              <group delimiter="; ">
+                <text macro="originally-published"/>
+                <group delimiter=", ">
+                  <text macro="reprint"/>
+                  <text macro="publisher"/>
+                </group>
+              </group>
+            </else>
+          </choose>
           <text macro="issued"/>
         </group>
       </else-if>


bwiernik:
CSL currently assumes that multiple publishers are stored together in one `publisher` field.

bwiernik:
@adam3smith Thoughts here? Repurposing the `original-publisher` variable here feels very hacky and I don't want to necessarily encourage the level of style-specific data entry. 

Publisher is changed to a name variable in the current draft of the CSL 1.1 spec, but unclear when we might get that finalized and implemented.

bwiernik:
@dstark Zotero's citation processor has a capability of splitting the publisher and place fields on a semicolon and then joining them together with delimiters if arguments are set in the style. 

These style attributes aren't in the official CSL spec, so we wouldn't accept a style using them here, but you could host a variant elsewhere. I have not tested if this works. 

> The subgroup-delimiter attribute is a field-parsing hack coded into the citeproc-js processor, enabled when the processor is run in CSL-m mode. In a group containing only cs:text elements rendering the publisher and publisher-place variables, the processor will split the content of both fields on a semicolon. If the length of the resulting list objects is equal, each publisher/publisher-place pair will be joined with the delimiter string set on the enclosing cs:group element. The composed pairs are then joined using the subgroup-delimiter value.

https://citeproc-js.readthedocs.io/en/latest/csl-m/

denismaier:
> Thanks, @bwiernik. This fact possibly explains why I've sometimes seen different output from Pandoc's citeproc than with citeproc-js, though not on this issue. I'll bear this in mind as I watch for cases that I come across like this moving forward as it should help me sort out why, on certain rare occasions, I'm not seeing from citeproc what I anticipated based on citeproc-js.

No, that's probably not that. citeproc-js behaves differently than pandoc's citeproc with the same styles, but this is something different. You'll need to enable this explicitly in the style with `version="1.1mlz1"` to make the processor switch to csl-m mode.

bwiernik:
Thanks @denismaier wasn't sure what the next enabling step was

dstark:
> No, that's probably not that. citeproc-js behaves differently than pandoc's citeproc with the same styles, but this is something different.

Thanks @denismaier. Are you aware of any documentation that outlines the differences between the two?

(Sorry this is a slight a rabbit trail here. But I'm not seeing a way to do a DM here on GitHub.)

dstark:
Thanks for the feedback, all. Given the current CSL specification, it looks like the options are to

1. leave the style as-is and unable to produce exactly what SBL style requires for this scenario,
2. adapt the style to use the `original-publisher` and `original-publisher-place` fields [since co-publishers are both original](https://github.com/citation-style-language/styles/pull/7522#issuecomment-2813683528), or
3. use a non-CSL feature that works (only?) in citeproc-js.

If neither #2 nor #3 can be incorporated into the main style here,

- Would there be a better way to accommodate what SBL style requires for co-published works that could be merged into the main style?
- If I were to host elsewhere a forked style that handles this use case, could you help me understand whether there an upside of option #3 that makes up for its reliance on something that isn't in the current CSL specification and (it sounds like) isn't currently being contemplated for the one to come?

As always, thanks so much.

denismaier:
Regarding: "Would there be a better way to accommodate what SBL style requires for co-published works that could be merged into the main style?"
No, I don't think so. There's currently no other way to support this.

Regarding: "If I were to host elsewhere a forked style that handles this use case, could you help me understand whether there an upside of option 3 that makes up for its reliance on something that isn't in the current CSL specification and (it sounds like) isn't currently being contemplated for the one to come?"

Well, just have a look into the csl-m specs. csl-m features a couple of extensions that might be useful for a style like SBL (e.g., there's a nice consolidate-containers features). But this will then only work with Zotero. So, if you want to cover the multiple publishers requirement, I'd go with option 2.

denismaier:
> Thanks @denismaier. Are you aware of any documentation that outlines the differences between the two?
> 
> (Sorry this is a slight a rabbit trail here. But I'm not seeing a way to do a DM here on GitHub.)

Sorry, no. (As for the DM: there's also discourse.citationstyles.org)

dstark:
> Regarding: "Would there be a better way to accommodate what SBL style requires for co-published works that could be merged into the main style?" No, I don't think so. There's currently no other way to support this.
> 
> Regarding: "If I were to host elsewhere a forked style that handles this use case, could you help me understand whether there an upside of option 3 that makes up for its reliance on something that isn't in the current CSL specification and (it sounds like) isn't currently being contemplated for the one to come?"
> 
> Well, just have a look into the csl-m specs. csl-m features a couple of extensions that might be useful for a style like SBL (e.g., there's a nice consolidate-containers features). But this will then only work with Zotero. So, if you want to cover the multiple publishers requirement, I'd go with option 2. (Yes, I also think it's somewhat hacky, and it doesn't handle the case of even more publishers, but 2 publishers are already a rather rare exception, and more than 2 publishers are even less common.)

Got it. Appreciate your input!

bwiernik:
@adam3smith  thoughts here?

adam3smith:
Yeah, I appreciate the intention here, but don't like what this does to data entry. It also seems quite likely that SBL will follow Chicago in removing publisher place, at which point this can just be handled in the data. 
I'd say we just can't do this in CSL currently

bwiernik:
This will be handled if a future version of CSL where publisher is a name variable is released. Until then, closing for now. Thanks for the discussion folks.

PR #7537, opened 2025-04-22
Create lecciones-y-ensayos.csl

tomasdalessio:


bwiernik:
Thanks! Please make additional edits to the file to correct the failed tests in this pull request, rather than opening a new one.

tomasdalessio:
Hi! My style has passed the tests satisfactory so I wish to know which stepts should I follow for it to be added to the repository. It's the first time I've worked with github so I'm still learning how it functions, sorry.

tomasdalessio:
> Thanks! Please make additional edits to the file to correct the failed tests in this pull request, rather than opening a new one.

Hi! My style has passed the tests satisfactory so I wish to know which stepts should I follow for it to be added to the repository. It's the first time I've worked with github so I'm still learning how it functions, sorry.

revista-lecciones-y-ensayos.csl
```diff
@@ -0,0 +1,420 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" default-locale="es-ES">
+  <!-- This style was edited with the Visual CSL Editor (https://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>Revista Lecciones y Ensayos</title>
```
bwiernik, line 5:
```suggestion
    <title>Revista Lecciones y Ensayos (español)</title>
```

bwiernik:
No problem at all--I will let the tests and then review the output for some sample items to check for problems. I'll let you know if the style needs any updates or merge it if it looks good.

revista-lecciones-y-ensayos.csl
```diff
@@ -0,0 +1,420 @@
...
+  </macro>
+  <citation et-al-min="4" et-al-use-first="3" disambiguate-add-year-suffix="true" collapse="year-suffix">
+    <layout delimiter="; ">
+      <choose>
+        <if type="book">
+          <group delimiter=", ">
+            <text macro="author-short"/>
+            <text variable="title" form="short" font-style="italic" font-variant="normal"/>
+          </group>
+          <text macro="locator" prefix=", "/>
```
bwiernik, line 128: Move this line inside the group above for each of these formats.

revista-lecciones-y-ensayos.csl
```diff
@@ -0,0 +1,420 @@
...
+  </macro>
+  <macro name="year-issued">
+    <date variable="issued">
+      <date-part name="year"/>
+    </date>
+  </macro>
+  <citation et-al-min="4" et-al-use-first="3" disambiguate-add-year-suffix="true" collapse="year-suffix">
+    <layout delimiter="; ">
+      <choose>
+        <if type="book">
```
bwiernik, line 123: Please change subsequent item types to use "else-if" with the same "choose", rather than starting a new "choose" block for each. 

Please add an "else" at the end to capture other item types you haven't explicitly added, like article-magazine and webpage

revista-lecciones-y-ensayos.csl
```diff
@@ -0,0 +1,420 @@
...
+  </macro>
+  <macro name="author-short">
+    <names variable="author">
+      <name form="short" and="symbol" delimiter-precedes-et-al="never" delimiter-precedes-last="never" et-al-use-first="3" initialize-with=".">
+        <name-part name="family" font-variant="small-caps"/>
+      </name>
+      <substitute>
+        <names variable="editor"/>
+        <names variable="translator"/>
+      </substitute>
```
bwiernik, line 67: Add a `<text variable="title"/>` or similar to the "substitute" to handle items without an author/editor/translator

revista-lecciones-y-ensayos.csl
```diff
@@ -0,0 +1,420 @@
...
+          <text macro="jurisdiccion-ente"/>
+          <text variable="title-short" prefix=" (" suffix=")"/>
+          <text variable="title" prefix=", " font-style="italic" font-variant="normal"/>
+          <date date-parts="year" form="numeric" variable="issued" prefix=", ">
+            <date-part name="year"/>
+          </date>
+          <text variable="publisher-place" prefix=", "/>
+          <group>
+            <text variable="URL" prefix=", URL " suffix=" consultado el "/>
+            <date form="numeric" variable="accessed"/>
```
bwiernik, line 413: What is the format you are trying to achieve for the access date?

bwiernik (changes requested):
Thanks, a few comments that apply to several places to start. 

Several of the example items are showing odd formatting. Can you please check those too after addressing my other comments?

tomasdalessio:
All Checks passed!

bwiernik:
The title needs to be

`    <title>Revista Lecciones y Ensayos (español)</title>`

That is why the checks are failing.

tomasdalessio:
I've updated the title. Must it have else" at the end to capture other item types I haven't explicitly added? It's a style for a Law Review which doesn't include those types of entries.

bwiernik:
I would strongly recommend including an 'else' option with default formatting to help the style be robust to different levels of data quality (or if an author cites an unusual item type).

tomasdalessio:
Thank you very much! I've added default options for citing webpages, broadcasts, post-weblogs, posts, and song articles.
I hope it's ready to submit to the repository.

POBrien333:
I just started looking at this style, but all terms are hardcoded and the style is just a serial of "if" conditional instead of else-ifs. Dots are hardcoded. I'm not sure this is of good enough quality to accept on the repository as is and would say this should be reworked.
Also, the actual title is "Lecciones y ensayos", without the "Revista ".

revista-lecciones-y-ensayos.csl
```diff
@@ -0,0 +1,418 @@
...
+  <bibliography et-al-min="4" et-al-use-first="3" hanging-indent="true" subsequent-author-substitute="&#8212;&#8212;">
+    <sort>
+      <key macro="author"/>
+      <key macro="year-issued"/>
+    </sort>
+    <layout>
+      <choose>
+        <if type="book">
+          <text macro="author"/>
+          <text macro="title" quotes="false" prefix=", "/>
```
bwiernik: All of these prefixes should be groups and delimiters throughout the style

tomasdalessio (commented):
Hi @bwiernik,

I've pushed the requested rework. Summary of changes:

- Consolidated both the citation and bibliography layouts into a single `<choose>` with `<if>`/`<else-if>`/`<else>`, including a generic `<else>` fallback for unforeseen item types.
- Moved all hardcoded terms into the `<locale>` block (in, accessed, volume, issue, edition, editor label, quotes) and reference them via `<label>`/`term=`.
- Replaced the per-element `prefix` attributes with `<group delimiter=", ">`, and the hardcoded final periods with a layout `suffix="."`.
- Corrected the title to "Lecciones y Ensayos" and renamed the file to `lecciones-y-ensayos.csl` so it matches the style `id` and self link.

On the CI: the remaining `may not have any unused macros` failures are all in pre-existing styles (apa-*, chicago-*, taylor-and-francis, turabian) and are unrelated to this PR — my style isn't among them. I'll rebase on the latest master to clear them; please let me know if anything else is needed on my end.

One thing I'd like to flag for your review: a few legal item types intentionally map the journal's Zotero fields onto `references` (publication record / abbreviated court) and `abstract` (summary). This follows the journal's official citation manual and its Zotero data-entry workflow, so removing it would break how contributors enter legislation and case law. Happy to discuss if you'd prefer a different approach.

Thanks for the review!

lecciones-y-ensayos.csl
```diff
@@ -0,0 +1,417 @@
...
+            <text macro="pages"/>
+            <text macro="online"/>
+          </group>
+        </else-if>
+        <else-if type="thesis">
+          <group delimiter=", ">
+            <text macro="author"/>
+            <text variable="title" font-style="italic"/>
+            <group delimiter=" ">
+              <text value="tesis de"/>
```
POBrien333: This currently doesn't make sense to me.
The "genre" field for thesis includes a descriptor of the type of tesis "MS tesis", "PhD thesis" etc.
So a "tesis de PhD tesis" doesn't make sense here and I don't really want to accept this as is.
Can you explan what you want to achieve?

lecciones-y-ensayos.csl
```diff
@@ -0,0 +1,417 @@
...
+            <text macro="pages"/>
+          </group>
+        </else-if>
+        <else-if type="standard">
+          <group delimiter=", ">
+            <text variable="title"/>
+            <text macro="issued-full"/>
+            <text variable="publisher-place"/>
+            <group delimiter=" ">
+              <text value="e.v."/>
```
POBrien333: what does "e.v." mean in this context?

lecciones-y-ensayos.csl
tomasdalessio: Thanks for flagging this. You're right that the construction `"tesis de" + genre` would produce nonsensical output if the `genre` field already contained an English descriptor like "PhD thesis".

The root of this is a journal-specific convention: the official citation manual for Lecciones y Ensayos specifies that the 'genre'/Type field should be filled with the full Spanish type descriptor — e.g., `tesis de doctorado` or `tesis de maestría` — and the expected bibliographic output is:

> MAISLEY, Nahuel, *El derecho de la sociedad civil…*, tesis de doctorado defendida el 14/11/2018, Facultad de Derecho, Universidad de Buenos Aires.

I've updated the style so the 'genre' field is emitted directly (without prepending a hardcoded "tesis de"), and "defendida el" follows immediately after. This means the full phrase now comes from the genre field itself, exactly as the journal's contributors are instructed to enter it in Zotero. The output becomes: `[genre] defendida el [date], [institution]`.

I'll push the fix now.

lecciones-y-ensayos.csl
tomasdalessio: "e.v." stands for "entrada en vigor" (Spanish for "entry into force"), which is the date on which an international treaty becomes legally binding. It is a standard term in international public law and is required by the journal's citation manual for treaty references.

For example, the expected output for a treaty entry is:

> Convención de Viena sobre el Derecho de los Tratados, 23/05/1969, Viena, Austria, e.v. 27/01/1980, UNTS 1115:331.

There is no equivalent CSL term for this concept, so it is hardcoded as a fixed string — the same situation as `"defendida el"` for thesis defense dates.

lecciones-y-ensayos.csl
```diff
@@ -322,9 +322,9 @@
             <text macro="author"/>
             <text variable="title" font-style="italic"/>
             <group delimiter=" ">
-              <text value="tesis de"/>
               <text variable="genre"/>
-              <text macro="issued-full" prefix="defendida el "/>
+              <text value=", defendida el "/>
```
POBrien333: Please revert this change and add it as a prefix (obviously add the missing comma). But please keep in mind we make changes on purpose ;)

tomasdalessio: Done. Thanks

PR #7540, opened 2025-04-23
New style for journal Geographische Zeitschrift (Deutsch)

wiertz:
This is a newly written style for the German geography journal Geographische Zeitschrift. (The existing style "Geographische Zeitschrift" does not match the current citation style of the journal and has not been updated in a long time.)


bwiernik:
Thanks! Can you please update the current style file, rather than creating a new one? You can make that change in this PR, rather than opening a new one

wiertz:
OK. I replaced the old style, and also added a small fix to the locator, and the anonymous term as a fallback when neither author nor editor are specified. The latter took me a while to figure out: 

1. specifying `<text term="anonymous"/>` as a child of `<substitute>` would not render any text
2. specifying `<text macro="anonymous"/>` and a macro `<macro name="anonymous"><text term="anonymous"/></macro>` correctly renders the anonymous term. 

Not sure why 1. did not work as the [documentation](https://docs.citationstyles.org/en/stable/specification.html?highlight=substitute#substitute) suggests that any rendering element should do.

bwiernik:
That's odd and likely a bug, but it's fine to keep the macro approach for now. Thanks!

wiertz:
I just talked to the lead editor of the journal. It turns out the official styleguide is not up to date with current editing practice at the journal. I will add a few modifications over the next week and post here once its in a final state.

POBrien333:
@wiertz 
I see that made this into a german-only style. Looking at a recent issue I see they still publish a mix of german and english articles.
https://biblioscout.net/journal/gz/113/2

wiertz:
As far as I can see the linked issue does not contain any English articles, but I did indeed find one in 2023. Just checked with the editor, they still accept English articles. I'll adapt the style accordingly.

wiertz:
I made a few amendments and compared the style results with the bibliography of a recent article a colleague and I published. It seems to handle expected cases well. However, there certainly remain edge cases (less common types or missing fields) that are not yet handled well. If you have suggestions what I should take into account, please do let me know. 

Please also note that there currently are no definitive guidelines for the journal. The citation style is based on preferences and practices of acting editors. Once a new csl style is available, I'll propose to make these the new standard.

PS test fails do not concern this style or do they?

adam3smith:
To 'fix' the errors, could you please click on "Sync fork" on https://github.com/wiertz/styles/tree/patch-1 (they are indeed not caused by you, but it's helpful to have the full tests run on your submission)

PR #7545, opened 2025-05-01
Enhance DOI handling for biomed-central.csl

NourEdinDarwish:
Update DOI handling for improved citation practices:
- Always show DOIs for references using the doi: prefix format (The previous implementation only showed DOIs when page/volume weren't available, which wasn't explicitly required by BMC. Showing DOIs consistently for all references improves citation quality and scholarly discoverability, while still conforming to BMC's formatting requirements)
- Add special handling for dataset DOIs using http://dx.doi.org/ format as shown in examples

Based on BMC's guidelines at:
https://bmcbioinformatics.biomedcentral.com/submission-guidelines/preparing-your-manuscript/research-article

### DIFF
--- a/biomed-central.csl
+++ b/biomed-central.csl
@@ -15,10 +15,13 @@
     <contributor>
       <name>Patrick O'Brien</name>
     </contributor>
+    <contributor>
+      <name>Nour Edin Darwish</name>
+    </contributor>
     <category citation-format="numeric"/>
     <category field="medicine"/>
     <category field="biology"/>
-    <updated>2021-11-02T17:11:28+00:00</updated>
+    <updated>2025-05-01T18:48:57+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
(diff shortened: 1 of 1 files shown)


NourEdinDarwish:
@adam3smith 
Thank you for your feedback! From my experience with BMC, they do not have consistent standards for DOI inclusion - it varies widely across publications and appears to be person-specific. Some authors include DOIs with "doi:" prefixes across all references, others use "https://doi.org/" format, and many include DOIs randomly throughout their reference lists. This inconsistency stems from the lack of fixed guidelines regarding DOI reporting. In my experience, it's often reviewer-dependent - some reviewers specifically request DOIs, while others don't mention them at all. Looking at BMC's example citations on their website, they appear to expect at least one minimal identifier (either DOI or volume), but remain vague about whether including both is acceptable. They clearly specify using "http://dx.doi.org/" format for datasets, emphasizing dataset citation in multiple places, yet remain ambiguous about linking other DOIs. To summarize, I believe it would be better to include DOIs whenever possible in the format used in their examples, with dataset DOIs as links - this would satisfy reviewers who commonly request DOIs and provide better citation utility. While many published articles use linked DOIs for all references, I fear some reviewers might argue this format should be reserved for datasets only. There's no strong argument for including only DOI or only volume information, and what happens most of the time is that reviewers request DOI inclusion during manuscript revision. BMC editors typically don't systematically check whether DOIs are included, formatted as links, or formatted consistently across all references. Ultimately, BMC should provide clearer guidance on DOI formatting, but until then, this approach offers a practical solution that anticipates reviewer requests and enhances citation discoverability while remaining compliant with all observed BMC formatting patterns.

bwiernik:
I am okay with adding DOIs consistently, but let's consistently use the modern https://doi.org/ format throughout. This is the recommended format by CrossRef and seems to be an acceptable format by BMC. Can you please make those changes?

bwiernik:
@NourEdinDarwish following up here

NourEdinDarwish:
@bwiernik 
Could you please take a look?

PR #7546, opened 2025-05-03
Update chinese-journal-of-aeronautics.csl

HardLeft1004:
1. add <contributor> YZW and refresh <updated> timestamp
2. introduce zh locale with custom term lang-suffix → “[Chinese]” create new lang-note macro; append it to bibliography layout for Chinese items
3. switch author macro to use initials (initialize="true", empty initialize-with)
4. output citation number as “N. ” (number + dot + space) in bibliography
5. revise webpage layout: remove extra period after title, inject “ [Internet].” token for correct punctuation

### DIFF
--- a/chinese-journal-of-aeronautics.csl
+++ b/chinese-journal-of-aeronautics.csl
@@ -11,13 +11,24 @@
       <name>Ramanathan</name>
       <email>Ra.Ganapathi@elsevier.com</email>
     </author>
+    <contributor>
+      <name>YZW</name>
+      <email>yzw.19951004@outlook.com</email>
+    </contributor>
     <category citation-format="numeric"/>
     <category field="engineering"/>
     <issn>1000-9361</issn>
     <summary>A style for some Elsevier journals, resembles Vancouver style</summary>
-    <updated>2014-09-26T17:00:10+00:00</updated>
+    <updated>2025-05-03 10:42:58</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
+
+   <locale xml:lang="zh">
+    <terms>
+      <term name="lang-suffix">[Chinese]</term>
+    </terms>
+  </locale>
+ 
   <macro name="access">
     <choose>
       <if variable="DOI">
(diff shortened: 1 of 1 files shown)


zepinglee:
`<if locale="zh">` is not a valid condition in CSL (see the CSL [spec](https://docs.citationstyles.org/en/stable/specification.html) and the [validator](https://validator.citationstyles.org/)).

chinese-journal-of-aeronautics.csl
```diff
@@ -189,7 +210,10 @@
         </else-if>
         <else-if type="webpage">
           <group delimiter=" ">
-            <text variable="title" form="long" suffix="."/>
+            <!-- 题名后不加句点，避免出现双句点 -->
+            <text variable="title"/>
+            <!-- 插入“ [Internet].” -->
+            <text value="[Internet]" prefix=" " suffix="."/>
```
bwiernik, line 216: Use `term="internet"`

chinese-journal-of-aeronautics.csl
```diff
@@ -122,7 +143,7 @@
   </citation>
   <bibliography entry-spacing="0" second-field-align="flush" et-al-min="7" et-al-use-first="6">
     <layout suffix=".">
-      <text variable="citation-number"/>
+      <text variable="citation-number" suffix=". "/>
```
bwiernik, line 146: Remove these suffixes and add `delimited=". "` to the group

chinese-journal-of-aeronautics.csl
```diff
@@ -112,6 +123,16 @@
       </else>
     </choose>
   </macro>
+
+  <!-- 若条目处于 zh locale，则在条目末尾加空格+[Chinese] -->
+  <macro name="lang-note">
```
bwiernik, line 128: This isn't possible in CSL. Please remove.

chinese-journal-of-aeronautics.csl
```diff
@@ -11,13 +11,24 @@
...
     <category citation-format="numeric"/>
     <category field="engineering"/>
     <issn>1000-9361</issn>
     <summary>A style for some Elsevier journals, resembles Vancouver style</summary>
-    <updated>2014-09-26T17:00:10+00:00</updated>
+    <updated>2025-05-03 10:42:58</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
+
+   <locale xml:lang="zh">
```
bwiernik, line 26: This isn't valid in CSL. Please remove.

chinese-journal-of-aeronautics.csl
```diff
@@ -212,6 +236,7 @@
           <text macro="access" prefix=". "/>
         </else>
       </choose>
+      <text macro="lang-note"/>
```
bwiernik, line 239: Remove

bwiernik (changes requested):

PR #7549, opened 2025-05-05
Create polski-humanistyczny-kulturoznawczy-autor-rok.csl

kjklosowski:
This is a Polish author and date-type citation style with wide use range for cultural studies, digital humanities.

It adds proper citation of software/video games, movies, songs, webpages, social media posts for humanities. In this case, for it being an in-text citation style, the way the bibliography is rendered is more important.

It is based on University of Gdansk guidelines, but many Polish universities' humanities departments use a version of Polish International Standard Bibliographic Description, being very similar to each other: (https://www.bn.org.pl/dla-bibliotekarzy/normy,-formaty,-standardy/katalogowanie/isbd/#:~:text=Mi%C4%99dzynarodowy%20Znormalizowany%20Opis%20Bibliograficzny%20%E2%80%93%20International,formie%20akceptowanej%20przez%20spo%C5%82eczno%C5%9B%C4%87%20mi%C4%99dzynarodow%C4%85.)

Hence the very general and broad name (it translates to polish humanities cultural studies - author, date).

### DIFF
--- a/polski-humanistyczny-kulturoznawczy-autor-rok.csl
+++ b/polski-humanistyczny-kulturoznawczy-autor-rok.csl
@@ -1,0 +1,418 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" default-locale="pl-PL">
+  <!-- This style was edited with the Visual CSL Editor (https://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>Polski Humanistyczny Kulturoznawczy - autor, rok (Polski)</title>
+    <id>http://www.zotero.org/styles/polski-humanistyczny-kulturoznawczy-autor-rok</id>
+    <link href="http://www.zotero.org/styles/polski-humanistyczny-kulturoznawczy-autor-rok" rel="self"/>
+    <link href="http://www.zotero.org/styles/studia-bas" rel="template"/>
+    <link href="https://wydawnictwo.ug.edu.pl/dla-autorow/instrukcja-wydawnicza" rel="documentation"/>
+    <author>
+      <name>Kacper Kłosowski</name>
+      <email>kacper.klosowski00@gmail.com</email>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="humanities"/>
+    <category field="sociology"/>
+    <category field="anthropology"/>
+    <summary>A Polish in-text citation style with wide use range for cultural studies, digital humanities. Based on Uni. of Gdansk guildelines</summary>
+    <updated>2025-05-03T16:09:20+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="pl">
+    <terms>
+      <term name="et-al">et al.</term>
+    </terms>
+  </locale>
+  <macro name="title">
+    <text variable="title" text-case="capitalize-first" quotes="false" font-style="italic"/>
+  </macro>
+  <macro name="title-short">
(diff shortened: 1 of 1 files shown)


POBrien333:
@kjklosowski 
I am wondering if you want to really style a book with authors the way it currently is?
See the test output:
`Beyond varieties of capitalism: Conflict, contradiction, and complementarities in the European economy, red. Hancké Bob et al., Oxford University Press, Oxford and New York 2007.`

Generally, you could simplify this style a lot by using "substitute" instead of testing for "author", then a new conditional testing for "editor".

kjklosowski:
Hi @POBrien333! Thanks for looking at the code! Two things here:
#### 1. Using substitues
I have actually thought about it and decided to not use substitues at all for one specific reason, this reason being the rules for sorting bibliography entries. The guidelines ("standardy redakcyjne") mention the entries should be sorted by the author's name (which is expected) but they also mention that anthologies should be sorted by their title (in that order, first authors from A-Z then titles from A-Z).

Defining editors as a substitute of author variable would break that compliance and that's something I'd like to avoid. This would be very rare, as it would require the user to cite the anthology as a whole, instead of citing a piece of text (which has an author) as a part of the anthology (this being just a book section). While I have personally never done that, it is sometimes used in that department of cultural studies. So I'd prefer to keep these completly separate :) The same goes for motion-picture, where director and producer are not substitues for the same reason, as movies are generally also sorted by the title.


#### 2. The amount of if statements
This is actually something I'd like to ask you about. Now that I look at the code, indeed there are so many if statements, that are redundant actually, as I utilize groups wherever I can. I don't know, I guess my line of thinking was to include them as a 'double precaution' against empty variable causing some issues with double punctuation, double spaces etc. but to be honest the group-delimiter functionality handles that on its own... I mean pieces of code like if the variable URL is present, print URL macro etc. The only place where if statements really make sense is journals where, regarding if volume and issue are present, the citing looks different. Most of the remaining if statements could be deleted actually.

Do you think it's worth for me to spend some time to clean these if statements up?

I would ofc change #7547 in bulk. I did not know back then that it is possible to include many files in one PR. Sorry about that :P

uniwersytet-gdanski-kulturoznawstwo-autor-rok.csl
```diff
@@ -0,0 +1,406 @@
...
+  </macro>
+  <macro name="url-accessed">
+    <group delimiter=" ">
+      <text variable="URL"/>
+      <date form="numeric" variable="accessed" prefix="[" suffix="]"/>
+    </group>
+  </macro>
+  <macro name="locator">
+    <group>
+      <text variable="locator"/>
```
POBrien333, line 138: I usually add a "if locator type NOT page, add locator label" type of statement to render just the number for pages, but something like "chap. 5" if you cite from a chapter. You could add that if you see this applicable.
Also, that group is otherwise not necessary. :)

POBrien333:
1. Gotcha. That makes sense. Thank you for explaining.
2. they are unnecessary, but they also won't hurt. If performance was an issue with this it could make a difference, but it doesn't really.

kjklosowski:
@POBrien333 Then I think I will leave them there. Thanks for your help :)

POBrien333:
See the open comment I left also.

kjklosowski:
@POBrien333 What do you mean? Can't find it nowhere

uniwersytet-gdanski-kulturoznawstwo-autor-rok.csl
kjklosowski: @POBrien333 That's definietly an important addition! I don't know why I haven't noticed that. It's done :)

#7547 doesn't need a change, as all locator types have a label there :)

POBrien333:
Your two PRs are merged now and will appear on the repo within an hour or two. Thanks for your patience and collaboration

kjklosowski:
Thank you @POBrien333 for all the help you provide in the CSL project!

PR #7555, opened 2025-05-06

shelbaum:
Include (last visited) parenthetical only if issuance date not provided; put theses and unpublished manuscript titles in roman type and include type and institution for theses; treat reports like books, with small caps author and title; don't include comma after month if no day provided; include month and day for webpages if available; keep title formatting even if no author provided.


adam3smith:
Thanks - this all looks good

PR #7557, opened 2025-05-07
Update palaeontology.csl

pfalkingham:
updated to remove quotes/emdashes for repeated authors.

Just had paper accepted into palaeontology, and technical editor told me style has changed:

"Please do not use ditto marks (———) in your references. Include all author names in full."

After querying, technical editor (Sally Thomas) said:

"I am aware that there is a Palaeontology style sheet around but don’t know how we can either have this adjusted (as it has been out of date for many years) or taken down. "

So, I've changed it to not use the --- quotes for duplicate authors in the list.


adam3smith:
Thanks for the context!

PR #7567, opened 2025-05-14
Update angewandte-chemie.csl with preprint citation. 

codexyster:
In accord with DOI style preprint citation from [author guideline](https://onlinelibrary.wiley.com/page/journal/15213773/homepage/notice-to-authors)


angewandte-chemie.csl
```diff
@@ -136,6 +136,13 @@ This style has many limits due to csl constraints, most notably the inability to
             <text macro="year-date"/>
           </group>
         </else-if>
+        <else-if type="article" match="any">
+          <group delimiter=", ">
+            <text macro="year-date"/>
+            <text variable="publisher" suffix=" preprint"/>
```
bwiernik:
```suggestion
            <group delimiter=" ">
              <text variable="publisher"/>
              <text term="article"/>
            </group>
```

PR #7570, opened 2025-05-16
Remove `generic-base` category from derivatives

adunning:
According to the CSL spec, `generic-base` is for 'used for generic styles like Harvard and APA', but numerous local derivatives for use with specific journals, publishers, and institutions have retained it. This removes the category from such files, so that it is possible again to browse the official generic styles on the Style Repository.


bwiernik:
Great idea! Could you add this to the contributing and review guidelines file as well?

adunning:
Thank you – I've added this in #7576.

Issue #7571, opened 2025-05-16
ASA style has inconsistencies with ASA Style Guide 7th edition


Whodiduexpect:
## Example: A website retrieved from an institution with a known location and date
- The ASA style adds parentheses around URLs instead of treating them as a new item, which isn’t done in any of the examples in the 7th edition of the ASA style guide. I think this was a change made in the 6th edition.
- “Publisher” for a Web Page gets recognized by other CSL styles, but is omitted for ASA.
- The style always adds date of access for web pages, but the date of access is required "only if no date of publication or date of revision can be determined from the website." 

### ASA Style Guide 7th edition
> ASA (American Sociological Association). 2021. “Standing Committees.” Washington, DC: American Sociological Association. https://www.asanet.org/about/governance-and-leadership/standing-committees.

### CSL ASA
> ASA (American Sociological Association). 2021. “Standing Committees.” Retrieved June 1, 2022 (https://www.asanet.org/about/governance-and-leadership/standing-committees).

## Example: Research products from census and survey data
Dataset item type does not seem properly supported.

### ASA Style Guide 7th edition
> Ruggles, Steven, Katie Genadek, Ronald Goeken, Josiah Grover, and Matthew Sobek. 2017. _Integrated Public Use Microdata Series_: Version 7.0 [dataset]. Minneapolis: University of Minnesota. [doi:10.18128/D010.V7.0](http://doi.org/10.18128/D010.V7.0).

### CSL ASA
> Ruggles, Steven, Katie Genadek, Ronald Goeken, Josiah Grover, and Matthew Sobek. 2017. “Integrated Public Use Microdata Series.”

## Example: Journal Articles
- Style guide uses `doi:` DOI prefix, the CSL style uses `doi: ` (with a space).
- DOI prefix not included in hyperlink.
- Implementation issue: DOI does not get hyperlinked with Zotero Connector in Google Docs. Probably has to do with the fact that every entry is implemented as its own internal link.

### ASA Style Guide 7th edition
> Chase-Dunn, Christopher. 2013. “Five Linked Crises in the Contemporary World-System.” _Journal of World-Systems Research_ 19(2). [doi:10.5195/jwsr.2013.494](http://doi.org/10.5195/jwsr.2013.494).

### CSL ASA
[...]

PR #7577, opened 2025-05-20

POBrien333:
closes https://github.com/citation-style-language/styles/issues/7571


bwiernik:
Can you add replace the Anon in substitute with title macro? That's what is in a few examples in the manual

bwiernik:
Thanks!

PR #7578, opened 2025-05-22

POBrien333:
https://github.com/citation-style-language/styles/issues/7571

### DIFF
--- a/american-sociological-association.csl
+++ b/american-sociological-association.csl
@@ -16,7 +16,7 @@
     <category citation-format="author-date"/>
     <category field="sociology"/>
     <summary>The ASA style following the 6th/7th edition of the Style Guide (reference formatting is the same for both editions).</summary>
-    <updated>2025-05-20T11:36:01+00:00</updated>
+    <updated>2025-05-22T07:10:21+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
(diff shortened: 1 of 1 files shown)


american-sociological-association.csl
```diff
@@ -63,7 +63,13 @@
     <group>
       <choose>
         <if type="webpage post-weblog post" match="any">
-          <text variable="URL"/>
+          <group delimiter=". ">
+            <group delimiter=" ">
+              <text term="retrieved" text-case="capitalize-first"/>
+              <date form="text" variable="accessed"/>
+            </group>
```
Whodiduexpect, line 70: Make the access date only appear if there is no issued date.
```suggestion
            <choose>
              <if variable="issued" match="none">
                <group delimiter=" ">
                  <text term="retrieved" text-case="capitalize-first"/>
                  <date form="text" variable="accessed"/>
                </group>
              </if>
            </choose>
```

Whodiduexpect:
Although, I noticed another issue meanwhile. For reports, the CSL style seems to exclude the series title and the URL. This is another thing the Chicago CSL style has no issues with, but is not working properly in ASA (despite it being the same behavior). The style definition really shouldn’t look that different from Chicago, because ASA is more of a stylistic variant than its own entire style.

american-sociological-association.csl
```diff
@@ -63,7 +63,17 @@
...
+                  <text term="retrieved" text-case="capitalize-first"/>
+                  <date form="text" variable="accessed"/>
+                </group>
+              </if>
+            </choose>
+            <text variable="URL"/>
+          </group>
         </if>
         <else-if type="article-journal report dataset" match="any">
           <text variable="DOI" prefix="doi:"/>
```
Whodiduexpect: Use URL as a substitute if DOI is not available.

```suggestion
          <choose>
              <if variable="DOI">
                <text variable="DOI" prefix="doi:"/>
              </if>
            <else>
              <text variable="URL"/>
            </else>
          </choose>
```

american-sociological-association.csl
Whodiduexpect, line 1: GithHub won’t let me comment on it, but I also made this change to my local file to get the collection title to show up. Not sure if it’s the proper place to add it, though.
```diff
@@ -195,6 +202,7 @@
     <choose>
       <if type="report" match="any">
         <group delimiter=". ">
+          <text variable="collection-title" text-case="title"/>
           <text variable="genre" font-style="italic"/>
           <text variable="number"/>
         </group>
```

Whodiduexpect (changes requested):

bwiernik:
@POBrien333 could you check on DOI formatting and the position of series/collection-title?

PR #7582, opened 2025-05-25
Update modern-humanities-research-association-author-date.csl

POBrien333:
Remove date from book like bib entries.
source 8.4 here: https://www.mhra.org.uk/style/chapter8.html#section_8_3

via https://forums.zotero.org/discussion/124445/mhra-4thed-author-date-issue#latest


adunning:
I have just submitted a pull request in <https://github.com/citation-style-language/styles/pull/7583> that renders the bibliography in author–date style.

PR #7586, opened 2025-05-27
Create revista-espanola-de-enfermedades-digestivas

D01-MarcelVilasis:
Add style for the revista-espanola-de-enfermedades-digestivas (https://reed.es). Adapted with changes suggested in the last failed pull request


POBrien333:
@D01-MarcelVilasis 
I'm just looking into this PR. Does this journal just follow the vancouver style? Because I don't see that you made any changes to the style itself.
In that case we can create a dependent style.

PR #7597, opened 2025-06-12
Space with locators in MLA (sect.1 => sect. 1)

teoric:
Currently, locators in MLA are joined without intervening space, in my case "Absch.1" for "Absch. 1" ("sect.1" instead of "sect. 1"). This commit fixes this.


adunning:
Many thanks! This is a duplicate of #7569.

teoric:
> Many thanks! This is a duplicate of #7569.

Indeed.  Could not see that; sorry for the noise.

PR #7599, opened 2025-06-16
Update howard-hughes-medical-institute.csl

michavissar:
Added 3 lines instructing to print URL if PMID and DOI are both missing


michavissar:
Hoping someone can review this small change! Thank you!

PR #7604, opened 2025-06-23
Update university-of-york-apa.csl

AVWoman:
Redirecting to APA proper


AVWoman:
We're starting to tidy up the University of York styles to just have the Harvards be bespoke styles, and the rest to redirect to their proper versions. We're aware these therefore need to go into the "dependent" category but we're not sure how to do that-- any help gratefully received!

PR #7606, opened 2025-06-24
Update zeitschrift-fur-digitale-geisteswissenschaften.csl

Martin-de-la-Iglesia:


zeitschrift-fur-digitale-geisteswissenschaften.csl
```diff
@@ -47,6 +45,13 @@
           <name delimiter=" / " delimiter-precedes-last="always"/>
         </names>
       </if>
+      <else-if variable="editor">
+        <group suffix=" (Hg.)">
```
POBrien333: Please don't hardcode terms into the style like this. The style already had a locale with overrides. Please change those instead if you want to move from "Hg. von" to "Hg.".

zeitschrift-fur-digitale-geisteswissenschaften.csl
```diff
@@ -165,11 +156,11 @@
       </if>
     </choose>
     <choose>
-      <if type="chapter">
+      <if type="chapter paper-conference" match="any">
         <choose>
           <if variable="container-author editor" match="any">
-            <text variable="container-title" prefix="In: " suffix="."/>
-            <text macro="container-contributors"/>
+            <text macro="container-contributors" prefix=". In: " suffix=" (Hg.): "/>
```
POBrien333: Please also use `<text term="in"/>` instead of hardcoding it.

POBrien333:
@Martin-de-la-Iglesia 
Please take the various comments from bwiernik and myself into consideration.
I still see a variety of hardcoded terms that need addressing. Thank you.

POBrien333:
Note that the current output for websites produces a duplicated date: `
CSL search by example. In: Citation Style Editor. 2012. 2012. [http://editor.citationstyles.org/searchByExample/]`

Furthermore, I see that a link to an out of date style is given in the guidelines on the website. I assume you work for the journal. We would strongly recommend that a link to the style repository is given instead to avoid the circulation of various styles. That way the link also does not need to be manually updated in case of future updates to the style (from people like you or because CSL receives and update and a style is updated through that).
This is the link: https://www.zotero.org/styles?q=id%3Azeitschrift-fur-digitale-geisteswissenschaften

Martin-de-la-Iglesia:
> Note that the current output for websites produces a duplicated date: ` CSL search by example. In: Citation Style Editor. 2012. 2012. [http://editor.citationstyles.org/searchByExample/]`
> 
> Furthermore, I see that a link to an out of date style is given in the guidelines on the website. I assume you work for the journal. We would strongly recommend that a link to the style repository is given instead to avoid the circulation of various styles. That way the link also does not need to be manually updated in case of future updates to the style (from people like you or because CSL receives and update and a style is updated through that). This is the link: https://www.zotero.org/styles?q=id%3Azeitschrift-fur-digitale-geisteswissenschaften

Where do I find the Zotero entry for "CSL search by example"? I'd like to be able to reproduce the error.

POBrien333:
https://github.com/citation-style-language/csl-editor/blob/e31736c6e25268ee6d0f8f8279f97f5896afd6e6/src/exampleData.js#L170

Really any example that is a website would show the same problem. You call the date via "issued" and then later on in a conditional in "bibliography".

POBrien333:
Undoing some.of my changes is ok, but then you'd need to give some justification. I made those for a reason.

Martin-de-la-Iglesia:
> Undoing some.of my changes is ok, but then you'd need to give some justification. I made those for a reason.

Sorry about that; those edits were accidental. I'm going to try and revert them.

POBrien333:
Thanks. Congratulations.
The style should be available in an hour or 2 from the style repository and we would appreciate if that link is used for distributing your style via your website.

PR #7607, opened 2025-06-24
Update university-of-york-apa.csl

bwiernik:
Closes https://github.com/citation-style-language/styles/pull/7604


bwiernik:
@AVWoman Thanks! Cleaned up a few remaining changes needed (removing `class="in-text"` and `demote-non-dropping-particle="never"` since these are inherited from the parent style, moving the style to the `dependent` folder)

PR #7614, opened 2025-06-26
Remove default locale from Harvard Cite Them Right 

thisdell:
Modified from Cite Them Right to use "dkk." instead of "et al."

### DIFF
--- a/eu-interinstitutional-style-1st-edition.csl
+++ b/eu-interinstitutional-style-1st-edition.csl
@@ -1,4 +1,5 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+﻿
+<?xml version="1.0" encoding="utf-8"?>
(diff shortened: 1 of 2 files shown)


bwiernik:
What is the purpose of this style? What language is "dkk." used in?

bwiernik:
@thisdell Can you describe the purpose of this style? Is it intended for the style to be in English generally, but to have only "et al" be in Indonesian, or should the references be in Indonesian throughout?

bwiernik:
@adam3smith The standard Harvard Cite Them Right style has en-GB default locale. I'm guessing that this style is intended to be an Indonesian version of the style. Rather than making multiple different styles for different locales, should we just remove the default locale from the standard Harvard Cite Them Right style?

adam3smith:
I think so, yes. It makes sense for broadly used styles , and I'd assume most GB users will use the right locale so won't notice the change. @POBrien333 -- do you have thoughts on this?

bwiernik:
Going to merge for now. @POBrien333 let me know if you have any concerns.

PR #7617, opened 2025-06-30
Create NIH Grant Proposal.csl

venkmurthy:
Creates a better style for NIH grant proposal which omits URLs/Links and formats better than the existing derivative of NLM style

### DIFF
--- a/NIH Grant Proposal.csl
+++ b/NIH Grant Proposal.csl
@@ -1,0 +1,285 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" initialize-with-hyphen="false" default-locale="en-US">
+  <info>
+    <title>National Institute of Health Grant Proposal</title>
+    <title-short>NLM</title-short>
+    <id>http://www.zotero.org/styles/national-institute-of-health-grant-proposal</id>
+    <link href="http://www.zotero.org/styles/national-institute-of-health-grant-proposal" rel="self"/>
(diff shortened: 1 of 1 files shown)


venkmurthy:
I should say it may be appropriate to include the following as I modified an existing NLM style

    <contributor>
      <name>Michael Berkowitz</name>
    </contributor>
    <contributor>
      <name>Sebastian Karcher</name>
    </contributor>
    <contributor>
      <name>Matt Tracy</name>
    </contributor>

POBrien333:
Apart from the fact if we want to include this style in the repo (that's for the others to decide), please include a link to the template style you used. That gives enough credit to the makers of the original style. ;)

bwiernik:
The documentation page you linked to indicates that any citation format is fine, so long as necessary retrieval information and the PMCID is included, so I don't think that having two NIH/NLM styles is necessary. If you think there are improvements that can be made to the existing NLM/NIH grant proposal style, please edit that style rather than making a new style.

https://github.com/citation-style-language/styles/blob/master/national-library-of-medicine-grant-proposals.csl

I'm going to close this PR for now, but feel free to open a new one editing national-library-of-medicine-grant-proposals.csl

venkmurthy:
Thanks. The earlier one has some aspects which don't meet the requirements
(related to web links) and other formatting issues. Inclusion of web links can result in NIH grant applicantions being administratively rejected (after many months of work preparing an application!)

Would it be acceptable to propose an edit which changes it so markedly?

Thanks,

Venk

bwiernik:
Feel free to open a PR with edits to the existing style. But I'm not understanding how including URLs relevant to the citation would lead to a rejection.

venkmurthy:
Will do. The NIH views having web links as a way reviewers may be
compromised. Some have been manipulated to allow authors of grants to
geolocate reviewers.

Thanks!

PR #7622, opened 2025-07-07
Update STYLE_REQUIREMENTS.md

kjklosowski:
Added an important information about including Natural Language Version in styles' titles with default-locale different than English, which is - to my knowledge - completly undocumented explicitly.

See: Sheldon/bin/sheldon, "https://github.com/citation-style-language/Sheldon/blob/master/bin/sheldon#L281", lines 281-288.

Might be worth adding that to the requirements readme. Can save some unnecessary errors. I have learnt about this only when the bot gave me an error when I uploaded my own pull request. See: #7547.


adam3smith:
This is already included just a couple of lines below (l. 40 in the edited version). Feel free to improve, but let's not duplicate it

kjklosowski:
Thanks @adam3smith! I obviously have missed it and that's why I think it's a good idea to put more emphasis on it :)))

Also specifically mentioning the error code "Natural Language Version..." can be prolific as currently entering it to google produces nothing and I've already spotted couple of PRs with that issue unresolved for a long time.

The file should now be more readable and clear, and this info was combined into one section.

bwiernik:
Thank you. I rephrased the existing sentence to emphasize the "in the locale is not English" part and added the ABNT header as a second example in https://github.com/citation-style-language/styles/pull/7664

PR #7554, opened 2025-05-06
Adding style for L'Homme *for English-language manuscripts*

alexis-michaud:
This style is based on the style for _L'Homme_ already present in the repo. 
The reason for producing an English-language version is that the journal accepts manuscripts written in English, and the reference stylesheet needed to be adapted accordingly (using "and" instead of "et", and so on). 

I have checked it locally and it works (but I'm happy to chisel further if useful). Needless to say it's error-free against the [CSL Validator](https://validator.citationstyles.org/).

Credit is given to the author of the original file. I provide my own e-mail in the `email` field so the author of the original (French-language) version won't get unsolicited e-mail about someone else's adaptation of his work. 

```
    <author>
      <name>Mark Collins (English adaptation by Alexis Michaud, with help from Perplexity.ai and ChatGPT)</name>
      <email>alexis.michaud@cnrs.fr</email>
    </author>
```

Looking forward, 
With appreciation of Zotero & CSL

PR #7653, opened 2025-07-22
Update l-homme.csl

POBrien333:
via https://github.com/citation-style-language/styles/pull/7554

Reworked style to be in line with publications.
Example also here: https://journals.openedition.org/lhomme/51632#bibliography

By the looks of it this will need a separate style as they want " : " to delimit pages and I don't think we can set that, no?

### DIFF
--- a/l-homme.csl
+++ b/l-homme.csl
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style class="in-text" version="1.0" page-range-format="expanded" demote-non-dropping-particle="sort-only" default-locale="fr" xmlns="http://purl.org/net/xbiblio/csl">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" page-range-format="expanded" demote-non-dropping-particle="sort-only" default-locale="fr">
   <info>
-    <title>L'homme – Revue française d'anthropologie</title>
+    <title>L'homme &#8211; Revue française d'anthropologie (Français)</title>
     <id>http://www.zotero.org/styles/l-homme</id>
     <link href="http://www.zotero.org/styles/l-homme" rel="self"/>
     <link href="http://www.editions.ehess.fr/revues/lhomme/" rel="documentation"/>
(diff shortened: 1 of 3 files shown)


l-homme.csl
```diff
@@ -164,41 +182,47 @@
...
             </group>
           </else-if>
           <else-if type="article-journal article-magazine article-newspaper broadcast personal_communication thesis entry-dictionary entry-encyclopedia" match="any">
-            <text macro="title" quotes="true" suffix=". "/>
             <group delimiter=", ">
-              <text variable="container-title" font-style="italic"/>
-              <text variable="volume" prefix="vol.&#160;"/>
-              <text variable="issue" prefix="n°&#160;"/>
+              <text macro="title" quotes="true"/>
+              <group delimiter=": ">
```
adam3smith: Why not do something like this to get what they want, or am I misunderstanding the problem?
```suggestion
              <group delimiter="&#160;:&#160;">
```

POBrien333: Sure, but this would need to be set per language, so two styles would be needed, one for FR, one for ENG.

adam3smith: but the style is set to `fr` anyway?

POBrien333: Right, at the moment. Ideally I would've liked a polylang style for en and fr. Hence the question.

adam3smith: Ah, got it. No localized delimiters, no.

POBrien333: I've added an english style. Let me know if the naming is ok.

bwiernik: Can you check if citeproc-js produces the correct output for the style with with no space before the : when the locale is set to French? I have a vague memory that it automatically added the space before : for French output

adam3smith:
I think the naming is fine, thanks. You'd need to add the ISSNs as exceptions in the test suite (it's one of the .rb files, I'd have to search, too, but once you found them it's pretty obvious). I think the tests on this branch will still fail after that (because the files are pulled from the master branch) , but will then pass once we merge. Let me know if you can't find the right place to add this.

POBrien333:
Thanks. I actually couldn't find where to set this.

bwiernik:
The rules for the various CI checks are in th3 "spec" folder

PR #7658, opened 2025-07-24
Update zeitschrift-fur-digitale-geisteswissenschaften.csl

Martin-de-la-Iglesia:


POBrien333:
Dear @Martin-de-la-Iglesia 
I am sure GitHub seems a bit complicated and all, but please don't make new Pull Requests for the same thing.
image

Please continue in PR #7606

PR #7656, opened 2025-07-24
Add Istanus - Vancouver (no et al.) style

istanusacademy:
This CSL style is used in academic publications by Istanus Academy Biotechnology Co. Ltd.
It is a modified version of the standard Vancouver style, featuring:
Full author listing (no "et al.") in the bibliography
Round brackets in in-text citations, e.g., (1), (2,3)
Italicized journal and book titles
Compact page ranges (e.g., 123–5 instead of 123–125)
DOI or URL included at the end of each reference
We kindly request its inclusion in the official CSL style repository so it can be directly used via Zotero, Mendeley, and other CSL-compatible reference managers.
Maintainer: Istanus Academy Biotechnology Co. Ltd.Contact: istanusacademy@gmail.comWebsite: https://istanus.com

PR #7661, opened 2025-07-25
Create istanus-journal-on-applied-and-biological-sciences.csl

bwiernik:
Closes #7656


bwiernik:
@istanusacademy Thanks! I cleaned up a few things and fixed the automated checks. Since Istanus currently only publishes one journal, let's name the style after the journal at this time. We can adjust the name and add styles for other journals if additional publications are launched.

POBrien333:
Thanks for getting this over the finish line ;)

PR #7673, opened 2025-08-01
Update vodohospodarske-technicko-ekonomicke-informace.csl

TPOJ76:
We have updated the citation style according to the latest requirements of our Water Management Research Institute. We have made some changes to the order of initials and punctuation, and added or changed some terms.


POBrien333:
Together with #7675

PR #7677, opened 2025-08-01
Update nephrology-dialysis-transplantation.csl

bovender:
NDT (Nephrology, Dialysis, Transplantation) currently uses square brackets for citations, and the delimiter is a comma followed by a space. This also applies for dependent styles, e.g. Clinical Kidney Journal.

For reference, here is a current article in NDT:

https://academic.oup.com/ndt/article/40/7/1284/7929884


adam3smith:
Thanks! (confirmed on PDF, too)

PR #7678, opened 2025-08-01
Add ISO-690 (full note, English)

jpos1:
Added ISO-690 (full note, english)

### DIFF
--- a/iso690-full-note-en.csl
+++ b/iso690-full-note-en.csl
@@ -211,7 +211,7 @@
     </choose>
   </macro>
   <macro name="edition">
-    <choose>  
+    <choose>
       <if is-numeric="edition">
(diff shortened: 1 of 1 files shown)


jpos1 (commented):
Uldated file, lang set to en-US

adam3smith:
Thanks! Is this actually based on the Czech guidelines? That seems problematic since many countries tend to have their own implementations; if we're going to have ISO 690 English styles and call them that, I'd want them to be based on an English speaking country, most likely the UK or US.

bwiernik:
@jpos1 To @adam3smith's point, I would recommend starting from the iso690-author-date-en.csl style and editing the citation layout as needed, as the English and Czech versions of ISO 690 appear to be quite different.

PR #7612, opened 2025-06-25
Update and rename university-of-york-chicago.csl to dependent/univers…

AVWoman:
…ity-of-york-chicago.csl

Changing to a redirect to Chicago proper.

PR #7679, opened 2025-08-01
Chicago: Additional variants and corrections

adunning:
- Support parenthetical in-text citations based on notes system, defined in _CMOS_ 12.75–77
- Variants for ibid. and author-only subsequent citations with classic variants
- Additional variants with archival citations leading with the place first (_CMOS_ 14.205)
- Refer to additional variables for sorting; sort dates using the year only in author–date styles
- Render an untitled chapter numbered with `chapter-number`
- Initialize recipients in author–date
- Remove references to short forms of `container-title` to avoid user prompts to use journal abbreviations
- Use full name when disambiguating names with the same surname and initials
- Provide locators for legal types
- Add `personal_communication` to exceptions that assume archival references will be to physical rather than digital archives
- Use standard `sub-verbo` terms in CMOS17
- Fix repeated date in CMOS17 author–date
- Note inclusion of Bluebook in `title-short` for improved discovery and provide section references for variants
- Make University of York Chicago a dependent style (replaces #7612).
- Always handle uncertain dates
- Include `chapter-number` in short description
- Provide all shortened notes configurations for subsequent citations allowed in Turabian 16.4.1
- Restrict the use of the `forthcoming` term to journal articles with an `available-date` to cover CMOS 14.71, removing the automatic 'forthcoming' label with entirely undated journal articles

### DIFF
--- a/chicago-author-date-17th-edition.csl
+++ b/chicago-author-date-17th-edition.csl
@@ -2261,7 +2261,6 @@
         <if match="none" variable="container-title"/>
         <else-if match="none" type="periodical" variable="supplement-number volume-title"/>
         <!-- TODO: use `container-genre` here once available to allow a custom description of the journal volume -->
-        <else-if variable="collection-title supplement-number"/>
         <else-if variable="supplement-number volume-title">
           <text term="supplement" text-case="capitalize-first"/>
         </else-if>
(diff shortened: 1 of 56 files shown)


bwiernik:
@adam3smith In the MHRA PR, you said:

> I'm generally with you (or at least with what I think is your view): we should tread lightly on adding things like forthcoming to articles if it's only implicitly in the data. We have `status` to make that explicit.

Do you think we should also change the Chicago styles to only show the explicit `status` variable vs. implicating assuming that a no-date journal article is in press and showing the `forthcoming` term?

adunning:
I removed most uses of `forthcoming` in <https://github.com/citation-style-language/styles/pull/7679/commits/cb05071d31476e6d812d1da49c87a20d83f3265b>, meaning that it now only appears if an `available-date` is specified, covering CMOS 14.75 (where 'forthcoming' essentially becomes a decoration that the user shouldn't have to supply).

PR #7682, opened 2025-08-02
MHRA: Improve CSL coverage

adunning:
- Use additional variables in sorting
- Support `chapter-number` as a fallback for a missing `title` and `genre` or `page`
- Handle more uncertain dates
- Capitalize descriptions without a title in author–date
- Provide more consistent logic for labelling `part-number`
- Improve host/guest handling, analogous to interview format
- Remove University of York MHRA dependent style: the page at <https://subjectguides.york.ac.uk/reference-management/cite> links directly to the main MHRA style, and to the author–date rather than the notes system. Unlike most of the other UoY dependent styles, it is not modifying the style with a locale change, meaning that the extra style only creates clutter.


bwiernik:
@adam3smith What's your opinion on showing forthcoming/in press for journal articles without a date or status variable (that is, assuming they are in press vs undated)?

adam3smith:
I'm generally with you (or at least with what I think is your view): we should tread lightly on adding things like forthcoming to articles if it's only implicitly in the data. We have `status` to make that explicit.

adunning:
I'm very happy to make the logic for this more restrictive. MHRA gives no guidance on how to handle forthcoming works, so I have dropped it here. There is still a case in CMOS 14.75 where I think the `forthcoming` term is warranted (and I assume it is meant to have a function!), but that can be discussed in #7679.

PR #7685, opened 2025-08-06
Create greek-and-roman-musical-studie.csl

POBrien333:
Continues #6237 

Papers here:
https://brill.com/view/journals/grms/13/1/grms.13.issue-1.xml


bwiernik:
Can you add a title substitute for author and check the delimiter on the DOI for the Mares chapter?

PR #7686, opened 2025-08-06
Create associacao-brasileira-de-normas-tecnicas-anpd.csl

ecostadelle:
Brazilian ABNT rules as used by Autoridade Nacional de Proteção de Dados


bwiernik:
Could you describe what the differences are between this style and the standard ABNT style?

ecostadelle:
Sure! This style follows the ABNT NBR 6023:2025 standard, but introduces specific adjustments for legal documents, since the standard ABNT CSL styles do not fully comply with the official formatting for such references.

bwiernik:
Would it be better to apply those changes to the standard ABNT styles?

PR #7689, opened 2025-08-07
Lektoratbudrich patch 1 1

lektoratbudrich:
This pull request updates the barbara-budrich.csl citation style to match the 2025 author guidelines provided by the publisher.

### DIFF
--- a/barbara-budrich.csl
+++ b/barbara-budrich.csl
@@ -1,22 +1,17 @@
-<?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
+   <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never">
   <info>
-    <title>Barbara Budrich (deutsch)</title>
-    <title-short>Budrich</title-short>
-    <id>http://www.zotero.org/styles/barbara-budrich</id>
-    <link href="http://www.zotero.org/styles/barbara-budrich" rel="self"/>
-    <link href="http://www.zotero.org/styles/zeitschrift-fur-qualitative-forschung" rel="template"/>
-    <link href="https://budrich.de/Zeitschriften/Style_Sheet_2016.pdf" rel="documentation"/>
+    <title>Barbara Budrich (deutsch) 2025</title>
+    <title-short>Budrich 2025</title-short>
+    <id>http://www.zotero.org/styles/barbara-budrich-2025</id>
+    <link href="http://www.zotero.org/styles/barbara-budrich-2025" rel="self"/>
     <author>
-      <name>Jörg Schwarz</name>
-      <uri>https://github.com/jg-schwarz</uri>
+      <name>Katarina Willems</name>
     </author>
-    <category citation-format="author-date"/>
-    <category field="social_science"/>
-    <category field="humanities"/>
     <summary>This citation style for books and edited volumes of the publisher Barbar Budrich is based on the 'Style Sheet' for authors and editors from 2024 which is unfortunately not available from the publisher's website.</summary>
-    <updated>2024-06-13T17:28:01+00:00</updated>
+    <updated>2025-06-20T13:04:59+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+	<category citation-format="author-date"/>
+	<category field="social_science"/>
   </info>
   <locale xml:lang="de">
     <terms>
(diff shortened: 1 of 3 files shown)


POBrien333:
@lektoratbudrich 
You've changed a whole bunch of things that shouldn't have been changed. basically the whole info section should've been left as is, and you should add yourself as a `contributor`, not delete the original author of the style.
Then the style will also validate.

lektoratbudrich:
Yes, sorry. That didn't really work. I will try again.

POBrien333:
No problem. You can just edit the file of this PR.
Please just UPDATE the current .csl file. Don't create a new file with "-2025" appended. We only want one correct, updated style on the repository. Please keep the info section as is, only adding yourself as a contributor and maybe updating the documentation link if necessary.

adam3smith:
Those errors aren't yours. I'll send instructions to fix them in a second

barbara-budrich.csl
```diff
@@ -81,9 +84,9 @@
     </choose>
   </macro>
   <macro name="collection-title-number">
-    <group delimiter=" " prefix="(= " suffix=")">
-      <text variable="collection-title"/>
-      <text variable="collection-number"/>
+  <group delimiter=", " prefix="" suffix="">
+    <text variable="collection-title"/>
+    <text variable="volume" prefix="Band "/>
```
POBrien333: Instead of hardcoding the "Band" here as a prefix, could you use the adequate "label" here?

barbara-budrich.csl
```diff
@@ -146,6 +150,15 @@
           <text macro="creator-year-long"/>
           <text variable="title"/>
         </group>
+        <choose>
+          <if type="thesis">
+          <text value="Diss."/>
```
POBrien333: We tend to use the "genre" variable here, so a user can input whatever they require into the field in Zotero and it will be output. We tend to avoid hardcoding such text values.

barbara-budrich.csl
```diff
@@ -99,8 +102,9 @@
     </group>
   </macro>
   <macro name="doi">
-    <group delimiter=" ">
-      <text variable="DOI" prefix="https://doi.org/"/>
+    <group>
+    <text value="DOI: "/>
```
POBrien333: You could set the "DOI: " as a prefix instead of using "value" like the style was already doing. (generally, I try to avoid "value" at all costs

barbara-budrich.csl
lektoratbudrich: In this case it is required in the style sheet, that the type "thesis" always has this suffix. Could we keep it that way?

barbara-budrich.csl
lektoratbudrich: updated

barbara-budrich.csl
lektoratbudrich: updated

POBrien333:
Can you go here and click on "Sync fork"? https://github.com/lektoratbudrich/styles/tree/lektoratbudrich-patch-1-1
That will the validation errors that adamsmith mentioned above.

barbara-budrich.csl
POBrien333: @lektoratbudrich 
I don't see that you made any changes. Maybe something went wrong with your code upload?

barbara-budrich.csl
POBrien333: I guess you're not as experience with Zotero+CSL.
If you use the field "Type" in Zotero for a thesis which maps to the CSL variable "genre" you can use that variable in your citation style. That gives better control over the output of the style.

POBrien333:
@lektoratbudrich 
I've left a comment above and we'd still require you to only have one file in this PR, updating the original style and not adding a new -2025 file. Please remove the latter. If you're facing issues, please let us know.

barbara-budrich.csl
lektoratbudrich: @POBrien333 Hi, I just tried to commit the changes again. I got a successfull reply for the automated tests. Could you check if you can see the changes now?

POBrien333:
@lektoratbudrich 
Thanks. That's a lot better, although I made some minor changes now as one "value" was left.
Can you link to the current guidelines? (by updating the style)
The whole URL/DOI output looks odd to me and the guidelines currently linked in the style don't show this behaviour at all.

lektoratbudrich:
> @lektoratbudrich Thanks. That's a lot better, although I made some minor changes now as one "value" was left. Can you link to the current guidelines? (by updating the style) The whole URL/DOI output looks odd to me and the guidelines currently linked in the style don't show this behaviour at all.

@POBrien333 unfortunately the current guidelines can not be found online, so I can not update the link in the style. I have to check if we can upload it again. I could send you the document in the meantime.

POBrien333:
Could you add the file to the opening post of this PR? Then we can just link to the PR.

lektoratbudrich:
We will upload the file on our website. I'll get back to you as soon as its done.

POBrien333:
@lektoratbudrich 
I've just made some further improvements to the style. Do double check those and let us know once the guidelines are available and we can then merge this PR.

lektoratbudrich:
@POBrien333 the guidelines are online now: https://budrich.de/wiki/doku.php?id=formatierung:stylesheet

POBrien333:
Excellent. This is now merged.

PR #7691, opened 2025-08-07
Create environmental-chemistry-and-ecotoxicology.csl

Naveen-369:
This is the citation template for the "Environmetal Chemistry and Ecotoxicology" journal. This emphasize more on the journal articles and can also handles books and chapters.


POBrien333:
Just looking into the style in detail:
As this journal is an Elsevier journal they tend to follow standard citation styles and indeed the following style we have is a match: https://www.zotero.org/styles?q=id%3Aelsevier-with-titles&format=numeric

Published papers: https://www.sciencedirect.com/journal/environmental-chemistry-and-ecotoxicology

PR #7693, opened 2025-08-08
European Journal of International Law: Remove "last visited" for URLs

errotu:
"last visited" is not required for Internet sources according to the official EJIL stylesheet: https://ejil.org/about/Stylesheet_EJIL_2018.pdf#page=7

Also, the current version did not work flawlessly and created a closing square bracket.


POBrien333:
Can you update the documentation link in the info section as well as add "EJIL" as title-short underneath the title? (That way you can find it with acronym).

errotu:
Done! Thanks for the recommendations.

PR #7701, opened 2025-08-12
Update australian-archaeology.csl

mickmorrison:
Added 'url' field to item-type 'software' and amended timestamp


adam3smith:
Thanks!

PR #7705, opened 2025-08-14
Create japan-journal-of-stuttering-and-other-fluency-disorders.csl

KoichiM0:
Creation of the CSL for the "Japan Journal of Stuttering and Other Fluency Disorders," a journal in Japanese and English. Validated with 1.0.2 strict, with an error with the variable "author-yomi" (appearing on three lines), which is required for sorting bibliography in Japanese, and should be overridden when registering in Zotero.

### DIFF
--- a/japan-journal-of-stuttering-and-other-fluency-disorders.csl
+++ b/japan-journal-of-stuttering-and-other-fluency-disorders.csl
@@ -1,0 +1,1695 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" page-range-format="expanded">
+  <info>
+    <!-- title xml:lang="en">Japan Journal of Stuttering and Other Fluency Disorders Style</title -->
+    <title xml:lang="ja">吃音・流暢性障害学研究</title>
+    <title-short>JJSFD</title-short>
+    <!-- summary xml:lang="en">This Citation Style Language (CSL) file is for the Japan Journal of Stuttering and Other Fluency Disorders (JJSFD), which is an official journal of the Japan Society of Stuttering and Other Fluency Disorders (www.jssfd.org). In order to fulfil the required sort order of Japanese author names in the transliterated alphabetic order, it introduces a variable "author-yomi" in the Extra field, which generates a validation error against CSL 1.0.2 strict. Because it is essential to have this variable in this CSL, the user should override the error alert in installing it into their bibliography managing program.</summary -->
+    <id>http://www.zotero.org/styles/japan-journal-of-stuttering-and-other-fluency-disorders</id>
+    <link href="http://www.zotero.org/styles/japan-journal-of-stuttering-and-other-fluency-disorders" rel="self"/>
+    <link href="http://www.zotero.org/styles/apa-6th-edition-no-ampersand" rel="template"/>
+    <link href="https://www.jssfd.org/gakkaishi.html" rel="documentation"/>
+    <author>
+      <name>Koichi Mori</name>
+    </author>
+    <category citation-format="author-date"/>
+    <category field="communications"/>
+    <category field="engineering"/>
+    <category field="humanities"/>
+    <category field="linguistics"/>
+    <category field="medicine"/>
+    <category field="psychology"/>
+    <category field="science"/>
+    <category field="social_science"/>
+    <category field="sociology"/>
+    <eissn>2760-1749</eissn>
+    <summary xml:lang="ja">この文献引用スタイルファイルは、日本吃音・流暢性障害学会の公式論文誌である吃音・流暢性障害学研究(JSSFD)のために、引用スタイル言語 (CSL) で書かれたものです。日本語の著者名のソート順については、Extra (その他)のフィールド内に"author-yomi: "という項目を作り、ローマ字化した著者名を";"で区切って記入すれば、それがソートに使われます（順序以外の引用文献の表示には影響しません）。"author-yomi"は元来のCSL 1.0.2 (strict) にはないキーワードのため、スタイル・マネージャーに登録する時に「有効なCSL引用スタイルファイルではない」という警告が出ますが、OKをクリックして登録してください。なお、多人数著者の論文等で省略に「et al.」が使われるか「他」が使われるかは、「エクスポート」の設定の「言語：」の設定を「English (US)」にするか「日本語」にするかによって決まります。文献ごとに自動的に切りかえることはできません。</summary>
+    <published>2025-05-02T09:00:00+09:00</published>
+    <updated>2025-08-13T19:00:00+09:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="en">
+    <terms>
+      <term name="et-al">et al.</term>
+      <term name="and"/>
+      <term name="at"/>
+      <term name="in"/>
+    </terms>
+  </locale>
+  <locale xml:lang="ja">
+    <!-- defaultのjaのlocaleでは「年」「月」「日」が表示されることがあるのを防止するための修正 -->
+    <date form="text">
+      <date-part name="year" suffix="-"/>
+      <date-part name="month" form="numeric-leading-zeros" suffix="-"/>
+      <date-part name="day" form="numeric-leading-zeros" suffix=""/>
+    </date>
+    <date form="numeric">
+      <date-part name="year" suffix="/"/>
+      <date-part name="month" form="numeric-leading-zeros" suffix="/"/>
+      <date-part name="day" form="numeric-leading-zeros" suffix=""/>
+    </date>
+    <terms>
+      <!-- defaultのjaのlocale設定では"et-al"が「ほか」になるので、「他」とする等の修正。 -->
+      <term name="et-al"> 他</term>
+      <term name="personal-communication">バーソナル・コミュニケーション</term>
+      <term name="podcast">ポドキャスト</term>
+      <term name="album">アルバム</term>
+      <term name="film">映画</term>
+      <term name="circa" form="short">ca.</term>
+      <term name="issue" form="long">
+        <single>号</single>
+        <multiple>号</multiple>
+      </term>
+      <term name="at"/>
+      <term name="in"/>
+      <term name="retrieved"/>
+      <term name="from"/>
+      <term name="and"/>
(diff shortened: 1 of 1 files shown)


KoichiM0:
I need help. I cannot understand the error message (lines 16 and 17). Where in the CSL it refers to? How should the CSL be changed? The CSL passes the validation test of 1.0.2 strict except one variable "author-yomi", which cannot be resolved because it is essential to the style I want to realize. (Validation with 1.0.1 generates more errors with some more variables that are validated by 1.0.2.) But the error message seems to me unrelated to those variable and pointing to something else. Can someone help me?

bwiernik:
Thank you for your submission. We can only accept styles that validate against the CSL schema, and it is not possible to create a variable that is not part of the schema like `author-yomi`.

Could you explain what the "author-yomi" behavior is supposed to be?

Note that Zotero's citation processor, citeproc-js, was written by a Japanese legal scholar and was specifically designed to handle complexities of Japanese formatting, so I'm not sure any special handling of author name ordering is needed. Can you explain what is incorrect when you simply using the `author` variable?

bwiernik:
@KoichiM0 could you answer the questions above?

KoichiM0:
Thank you for your question. In order to sort bibliography including Japanese papers in accordance to the sort rules that the journal JJSFD requires, "author-yomi" is necessary. Because the journal requires sorting Japanese papers in the alphabetical order of author names that should be transliterated in alphabet (or romanized) just for the sorting purpose. Since transliteration of Japanese names can't be uniquely defined (one to multiple correspondence), transliteration must be provided manually and that should be used for sorting. I can't improvise any way around this issue without setting up an extra field as "author-yome". This is why the validation fails against the CSL scheme 1.0.2 strict.  I wish I could find a way to solve this transliteration and sort issue without violating the scheme, but I can't. Is there any method that allows entry of transliteration and validates? Thank you for your attention.

> Thank you for your submission. We can only accept styles that validate against the CSL schema, and it is not possible to create a variable that is not part of the schema like `author-yomi`.
> 
> Could you explain what the "author-yomi" behavior is supposed to be?
> 
> Note that Zotero's citation processor, citeproc-js, was written by a Japanese legal scholar and was specifically designed to handle complexities of Japanese formatting, so I'm not sure any special handling of author name ordering is needed. Can you explain what is incorrect when you simply using the `author` variable?

KoichiM0:
Additionally, it is rather common to sort alphabetically in scientific journals published in Japan, though most other journals requires numbered citations in the order of appearances. Perhaps legal documents may use the numbered references. If you do not specify how to read Japanese names, which are usually written in ideograms (Kanji), the order of bibliography becomes that of the character codes, which does not sort Japanese names correctly even in the Japanese alphabet (Kana, or Japanese phonogram).

bwiernik:
citeproc-js includes functionality to transliterate Japanese text using Hepburn romanization. This is applied for titles, but I'm not sure if it is used for names. 

Can you give a specific example of how names are sorted when you actually sort on author with Zotero and what the correct sort order should be? 

In any event, we cannot accept a citation style in the repository that does not pass validation.

bwiernik:
Tagging @fbennett in case you are available and have any insight

KoichiM0:
Transliteration cannot be uniquely (or mechanically) determined for Japanese names. For instance, the same "角田" could be read either as "Kakuta" or "Tsunoda", depending on the person (or family) you are referring to. Transliteration of titles may be easier if you have a large dictionary. However, this again cannot be fully automated because some ideograms (Kanji) are read differently depending on the context.

I appreciate the better handling of Japanese names than other commercial bibliography database, especially the special treatment (or non-treatment) of initials (Japanese use full given names instead of initials). However, there still remains the problem of how to read and sort.

Here is what the sort order in the bibliography looks like if transliterated names are NOT used. Zotero most likely resort to the character codes when it sort Japanese names. It works just fine with English names, but not Japanese names written in the native forms. The order looks rather random to those who can read Japanese (the citations are from a paper written in Japanese, discarding English citations and tltles etc.):
伊藤絵美 (2021).
原由紀, 小林宏明, 坂田善政, 前新直志, 餅田亜希子, 村瀬忍, 安田菜穂 (2009).
坂田菩致 (2003).
坂野雄二, 東條光彦 (1986).
富里周太, 矢田康人, 白石紗衣, 和佐野浩一郎 (2020).
小澤恵美, 原由紀, 鈴木夏枝, 森山晴之, 大橋由紀江, 餅田亜希子, 坂田善政, et al. (2016).
嶋大樹, 川井智理, 柳原茉美佳, 熊野宏昭 (2016).
嶋大樹, 柳原茉美佳, 川井智理, 熊野宏昭 (2013).
川合紀宗 (2010).
朝倉聡, 井上誠士郎, 佐々木史, 佐々木幸哉, 北川信樹, 井上猛, 傳田健三, et al. (2002).
松崎朝樹 (2013).
森浩一 (2017).
清水秀美, 今栄国晴 (1981).
熊野宏昭 (2012).
角田航平, 坂田善政, 北條具仁, 酒井奈緒美, 森浩一 (2017).
酒井奈緒美, チュウシンイン, 森浩一, 餅田亜希子, 北條具仁, 坂田善政 (2014).
阿栄娜, 酒井奈緒美, 安啓一, 森浩一 (2018).
阿栄娜, 酒井奈緒美, 森浩一 (2015).


Below is the list of the same bibliography in the (correct) alphabetic order of transliterated author names, with the citation style with "author-yomi" added: 
阿栄娜, 酒井奈緒美, 森浩一 (2015).
阿栄娜, 酒井奈緒美, 安啓一, 森浩一 (2018).
朝倉聡, 井上誠士郎, 佐々木史, 佐々木幸哉, 北川信樹, 井上猛, 傳田健三, et al. (2002).
原由紀, 小林宏明, 坂田善政, 前新直志, 餅田亜希子, 村瀬忍, 安田菜穂 (2009).
伊藤絵美 (2021).
角田航平, 坂田善政, 北條具仁, 酒井奈緒美, 森浩一 (2017).
川合紀宗 (2010).
熊野宏昭 (2012).
松崎朝樹 (2013).
森浩一 (2017).
小澤恵美, 原由紀, 鈴木夏枝, 森山晴之, 大橋由紀江, 餅田亜希子, 坂田善政, et al. (2016).
酒井奈緒美, チュウシンイン, 森浩一, 餅田亜希子, 北條具仁, 坂田善政 (2014).
坂野雄二, 東條光彦 (1986).
坂田菩致 (2003).
嶋大樹, 川井智理, 柳原茉美佳, 熊野宏昭 (2016).
嶋大樹, 柳原茉美佳, 川井智理, 熊野宏昭 (2013).
清水秀美, 今栄国晴 (1981).
富里周太, 矢田康人, 白石紗衣, 和佐野浩一郎 (2020).

I'm hoping that Zotero (or CSL) would officially provides a field to store transliterated names.
Perhaps an extension could be created that would sort the bibliography by reading the author-yomi field in the Extra, thus exempting the CSL from the necessity of reading author-yomi. But I am not capable of writing an extension.

bwiernik:
CSL 1.0.2 does not currently support any special treatment of transliterated names. This is something we could consider for a future version, but it is not available now.

You could look into the Jurism fork of Zotero and whether the added functionality for transliteration fits your needs, but not that this software is not actively maintained any longer. https://juris-m.readthedocs.io/en/latest/

For the time being, for this style to be included in the repository, you would need to remove the author-yomi variable. 

You are of course welcome to host a style that doesn't strictly validate against CSL 1.0.2 on your own website.

fbennett:
The issue is as @koichiMO describes it. Proper sorting of Japanese
references is impossible without access to phonetic transliterations.

Jurism will indeed provide the needed support. No extra variable is
required in the CSL: Jurism can cast a supplementary field for the
transliteration, and it can be configured to use that field for sorting
purposes.

While I had stepped back from the project some time ago due to a personal
tragedy, another programmer with a workplace need for its functionality has
come forward to assist, and we have been working to get a 6.0.30 release
out the door.

All tests are passing against the code for the Jurism client and its
browser extensions. The only obstacles to release are in signing the code
and getting it approved for release through the various platforms. That
cumbersome work is in progress.

Once 6.0.30 is out the door, I'll be turning to a leapfrog update based on
Zotero 8.

I am aware of the view among Zotero devs that multilingual functionality
should be implemented as a Zotero plugin, rather than as a project such as
Jurism that is "intermittently maintained". That reservation conflates two
separate problems, so a few words about that.

Robust multilingual support requires substantial changes to the Zotero
database schema and the user interface. Technically it would be possible,
with sufficient programmer resources, to accomplish this by "monkey
patching" Zotero code in a plugin extension. That's a substantial task that
anyone with the inclination is welcome to undertake, but the result would
be a tool with the same complexity, limitations, and compatibility issues
as Jurism itself.

Most important, a plugin implementation would require regular updates to
maintain compatibility with specific releases of mainline Zotero. The issue
is thus less whether a plugin implementation would be more elegant, and
more one of the resources required for long-run maintenance. The former
comes into view as a possibility only if the latter are secured.

I do have ideas for how the project might be sustained over the longer
term, with eventual handoff to other maintainers. It's a long-term goal and
I will be working in that direction, independently of Zotero. But the first
priority will be to get 6.0.30 up and running, and then to target a version
8 update, as that is the path of least resistance with the resources
currently available.

As for the immediate style, it should work fine in Jurism if the non-CSL
variable is removed and it validates against the CSL schema.

Cheers,
FB

bwiernik:
Fantastic to hear from you @fbennett

KoichiM0:
Thank you for the additional input, @fbennett. Juris-M surely seams to have everything that is needed for multilingual citation. However, after I tried Juris-M v.6.0.22 (current release) for the mac for a few hours, I have to say that it is unusable at this point because it incorrectly switches the first and last names of Japanese authors when the data are imported from J-Stage archive through Jurism Connector 6.0.30.1 installed in Brave browser (Chrome lookalike). And there is no "Add by Identifier..." import menu as available in Zotero. Also worrisome is the compatibility of the online database between Zotero and Juris-M if one wants to share it. 
Now, back to Zotero7 I'm thinking of forging an existing validated variable, hopefully infrequently used, to assume the proxy role of holding transliterated creator names.

fbennett:
Good luck then.

KoichiM0:
The CSL is now modified so that it does not employ an unvalidated variable ('author-yomi'). Instead, it uses 'original-author' for the entry of transliterated creators, which makes the CSL validated (1.0.2 strict). It is syntactically correct, though semantically dubious. The choice was made in the hope that the variable is rarely used in most of the papers that the journal JJSFD publishes. This is meant to be a temporary solution until transliterated names of creators would be officially supported in Zotero. Currently, the user should be warned in the summary and otherwise about the possible confusion in the sorting order of bibliography if s/he wants to use the variable for its original purpose.

PR #7708, opened 2025-08-14
Update zeitschrift-fur-theologie-und-philosophie.csl

knorb:
Update according to revised author guidelines (as of 1 March 2024), and some minor improvements in the code.

### DIFF
--- a/zeitschrift-fur-theologie-und-philosophie.csl
+++ b/zeitschrift-fur-theologie-und-philosophie.csl
@@ -18,7 +18,7 @@
     <category field="philosophy"/>
     <issn>2709-8427</issn>
     <summary>English and German full note style for Zeitschrift für Theologie und Philosophie (ZTP).</summary>
-    <updated>2025-08-14T14:25:45+00:00</updated>
+    <updated>2025-08-15T12:30:00+00:00</updated>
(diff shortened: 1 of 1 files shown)


zeitschrift-fur-theologie-und-philosophie.csl
```diff
@@ -363,29 +374,22 @@
       </choose>
       <choose>
         <if type="webpage post-weblog article-newspaper article-magazine article-journal entry-dictionary entry-encyclopedia" match="any">
-          <text term="in" prefix=", " suffix=": "/>
-          <text variable="container-title" form="short"/>
+          <group>
+            <text term="in" prefix=", " suffix=": "/>
```
POBrien333: since you have a group now you can set the : as a group delimiter.

knorb: Thank you. That's done, see below.

POBrien333:
I made an improvement to the point-locators macro. please confirm it's working correctly and then we can merge.

knorb:
> I made an improvement to the point-locators macro. please confirm it's working correctly and then we can merge.

Thank you, it works correctly!

POBrien333:
Thank you. Style should be available from the repository in an hour or so.

PR #7711, opened 2025-08-15
MLA: Expand CSL support

adunning:
- Provide supplemental elements (MLA 5.106– 118), including support for reviews, multivolume and multi-part works, and journal special issues
- Add variants without URLs (MLA 5.84)
- Use `<label>` element to label variables to allow plural 'nos.' etc.
- Support uncertain dates
- Use em dashes in the bibliography (MLA 5.127 allows either dashes or hyphens but prefers the former)
- Provide `archive_collection`
- Allow `publisher-place` as fallback where there is no `publisher`
- Label `version` variable
- Sort attributes
- Add comments to reflect MLA container structure

### DIFF
--- a/modern-language-association.csl
+++ b/modern-language-association.csl
@@ -56,6 +56,7 @@
         <single>n</single>
         <multiple>nn</multiple>
       </term>
+      <term name="original-work-published">originally published</term>
       <term form="short" name="paragraph">
         <single>par.</single>
         <multiple>pars.</multiple>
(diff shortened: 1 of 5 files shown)


bwiernik:
Looks good other than the validation errors. Can you fix those?

bwiernik:
Looks good other than the validation errors. Can you fix those?

modern-language-association.csl
```diff
@@ -75,126 +247,510 @@
...
+        </if>
+        <else>
+          <text macro="label-version-capitalized"/>
+        </else>
+      </choose>
+    </group>
+  </macro>
+  <!-- 4.4. Number (MLA 5.51-53) -->
+  <macro name="container1-number">
+    <group delimiter=", ">
```
adunning, line 707: This group is the cause of the extra comma in the test citations, due to the usual `citeproc-ruby` bug, but it isn't an issue elsewhere.

adunning:
This version doesn't yet provide full coverage of the _MLA Handbook_ (the name types and item descriptions need more detailed work), but it covers the key missing variables, so I think it is good enough for now.

PR #7725, opened 2025-08-21
Update angewandte-chemie.csl

POBrien333:
via https://forums.zotero.org/discussion/126249/citation-format-error-angew-chem-style#latest

### DIFF
--- a/angewandte-chemie.csl
+++ b/angewandte-chemie.csl
@@ -21,7 +21,7 @@
     <category field="chemistry"/>
     <issn>1433-7851</issn>
     <eissn>1521-3773</eissn>
-    <summary>A style for Wiley-VCH's journal &quot;Angewandte Chemie International Edition&quot;
+    <summary>A style for Wiley-VCH's journal "Angewandte Chemie International Edition"
 This style has many limits due to csl constraints, most notably the inability to include pages/page ranges for books, the wrong square brackets for two items cited together, and the lack of support for citing 1a)... b)...c)....</summary>
     <updated>2025-08-21T12:42:21+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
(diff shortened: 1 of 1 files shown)


angewandte-chemie.csl
```diff
@@ -104,19 +113,18 @@ This style has many limits due to csl constraints, most notably the inability to
   <bibliography entry-spacing="0" second-field-align="flush">
     <layout suffix=".">
       <text variable="citation-number" prefix="[" suffix="]"/>
-      <group delimiter=", ">
+      <group>
         <text macro="author"/>
-        <text macro="title" quotes="true"/>
+        <text macro="title" prefix=", "/>
```
bwiernik, line 118: Can you make these prefixes throughout the bibliography into groups/delimiters instead ?

angewandte-chemie.csl
```diff
@@ -104,19 +113,18 @@ This style has many limits due to csl constraints, most notably the inability to
...
-            <group delimiter=", ">
-              <text variable="title" text-case="title" font-style="italic"/>
+            <group delimiter=", " prefix=", ">
               <text macro="publisher"/>
               <text macro="year-date"/>
             </group>
           </if>
           <else-if type="chapter paper-conference" match="any">
-            <group delimiter=", ">
+            <group delimiter=", " prefix=" ">
```
bwiernik, line 127: Just confirming that the two places the prefix is just a space with no comma are correct?

angewandte-chemie.csl
POBrien333: This is done like so on purpose because depending on the item type I need different outputs. For example, they want chapters like so, with spaces:
`I. Mares in Varieties of Capitalism. The Institutional Foundations of Comparative Advantage (Eds.: P.A. Hall, D. Soskice), Oxford University Press, New York, 2001, pp. 184–213.`

angewandte-chemie.csl
POBrien333: Yeah, this is to achieve the chapter output as shown in above comment.

adam3smith: I think group delimiters are often preferable, but there's nothing wrong w affixes if they are used carefully (e.g. not mixing suffix and prefix spaces). I'll merge this

PR #7737, opened 2025-08-23
Chicago: Fix treaty party delimiter

adunning:
Move `and` from a global attribute to individual `<name>` elements to allow for a different configuration with Bluebook treaty parties.


chicago-author-date-17th-edition.csl
```diff
@@ -3560,11 +3552,11 @@
           </group>
           <choose>
             <if variable="issued">
-              <!-- APA manual includes "rev." before the revision year, but this isn't part of the Bluebook rules. -->
               <text macro="date-issued-year"/>
             </if>
             <else>
-              <!-- Show proposal date for uncodified regulations. Assume date is entered literally ala "proposed May 23, 2016". -->
+              <!-- Show proposal date for uncodified regulations, using `status` for "proposed" or similar. -->
+              <text variable="status"/>
```
bwiernik: 'status' isn't a good fit here because the regulation would generally be implemented, but not yet codified

PR #7738, opened 2025-08-24
MHRA: Ensure preposition with album track

adunning:
Add a conditional to work around `citeproc-js` hiding the preposition in certain cases.

### DIFF
--- a/modern-humanities-research-association-author-date.csl
+++ b/modern-humanities-research-association-author-date.csl
@@ -2458,7 +2458,11 @@
       <choose>
         <if match="any" type="broadcast entry entry-dictionary entry-encyclopedia motion_picture"/>
         <else-if type="song" variable="title">
-          <text macro="label-chapter-number"/>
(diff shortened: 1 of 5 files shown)


bwiernik:
Can you describe the citeproc-js behavior you're working around?

adunning:
Without this extra conditional, the word 'from' simply doesn't appear if there isn't also a `chapter-number`. I think it's trying to be clever and hide it as if it were a label.

bwiernik:
Oh, okay, this is expected behavior. Terms are suppressed inside of groups if there is also a variable in the group and it is empty. This is one of the benefits of using a term vs hard coding `<text value="..."/>`. I had thought the behavior only applied to groups, but I guess it also is activated inside of macros (makes sense to me).

The approach here works — does it also need to be added to Chicago?

adunning:
Many thanks! This is only an oddity of MHRA.

PR #7748, opened 2025-08-26
Update american-medical-association.csl: improve preprints

### LINKED ISSUE #5442: Citing preprint articles in AMA
Is there currently any support for citing preprint articles in correct AMA style? Here is an excerpt from the AMA style guide with the correct basic format for preprints (from section 3.11.4.1):

> Preprints are another online method for publication in which a manuscript is uploaded by authors to a public server, without editing or formatting, and typically without peer review.<sup>9</sup> A preprint may be a predecessor to publication in a peer-reviewed journal; it is “archived” and citable. Preprint servers include arXiv.org, bioRxiv.org, MedRxiv, and many others. Preprints were initially used more often in the physical sciences than in medicine, but they are becoming more common in the biological sciences.<sup>10</sup> Preprints may have DOIs and can follow this citation format:
> 
> 1. 1\. Bloss CS, Wineinger NE, Peters M, et al. A prospective randomized trial examining health care utilization in individuals using multiple smartphone-enabled biosensors. *bioRxiv*. Preprint posted online October 28, 2015. doi:10.1101/029983

Issue #3585 suggests using a reference type of `article`, but I have not been able to use the available fields to recreate the "Preprint posted online [full date including day of month]" from the example above. Additionally, when using `article`, the archive name (e.g., bioRxiv) does not show in the rendered citation.

If preprints are not currently supported for the AMA CSL, could this be considered as a new enhancement/feature?

POBrien333:
Closes #5442

Let me know this is ok and then I'll adapt the other variants also.

### DIFF
--- a/american-medical-association.csl
+++ b/american-medical-association.csl
@@ -262,7 +262,7 @@
             <text variable="container-title" font-style="italic" form="short" strip-periods="true" suffix="."/>
             <choose>
               <if type="article" match="any">
-                <text variable="publisher" suffix="."/>
+                <text variable="publisher" font-style="italic" suffix="."/>
               </if>
             </choose>
             <group delimiter=";" prefix=" ">
@@ -274,7 +274,14 @@
                 </if>
                 <else>
                   <group delimiter=" ">
-                    <text value="Published online"/>
+                    <choose>
+                      <if type="article">
+                        <text value="Preprint posted online"/>
+                      </if>
+                      <else>
+                        <text value="Published online"/>
+                      </else>
+                    </choose>
                     <date form="text" date-parts="year-month-day" variable="issued"/>
                   </group>
                 </else>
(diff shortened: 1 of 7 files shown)


american-medical-association.csl
```diff
@@ -260,6 +260,11 @@
           <text macro="editor" prefix=" " suffix="."/>
           <group prefix=" " suffix=".">
             <text variable="container-title" font-style="italic" form="short" strip-periods="true" suffix="."/>
+            <choose>
+              <if type="article" match="any">
+                <text variable="publisher" suffix="."/>
```
adam3smith: This should be in italics, no?

bwiernik:
Also change line 277 to:

```
<choose>
  <if type="article">
    <text value="Preprint posted online"/>
  </if>
  <else>
    <text value="Published online"/>
  </else>
</choose>
```

american-medical-association.csl
POBrien333: True that. thanks

PR #7750, opened 2025-08-26
MHRA: `part-title` and `reviewed-genre` consistency

adunning:
- Simplify rendering of part numbers (not documented in MHRA)
- Render `part-title` in quotes where necessary, mirroring `title`: fixes cases such as a multi-part lecture
- Reposition `reviewed-genre` to normal order
- Provide styles without URLs for all variants
- Remove unused terms
- Flatten `<layout>` elements to avoid an additional macro call


bwiernik:
For the ibid and publisher-place variants, should we just have the no URL variants? Seems like any publisher calling for the more antiquated variants wouldn't want URL?

adunning:
I think it would be best to have both. I can name at least one publisher using MHRA with place of publication and URLs; and the guide itself recommends using both ibid. and URLs.

PR #7752, opened 2025-08-26
MLA: Improve `number` and event

adunning:
- Provide `publisher` as a possible author substitute with `software` and `authority` for `standard` (see MLA examples)
- Provide labels for `number`
- Render event information with additional types


bwiernik:
For the event information, that means it will be show for published conference proceedings and some letters (event-date = date sent). Is that intended?

adunning:
Good catch; I have fixed this.

PR #7760, opened 2025-08-27
Update kunstakademie-munster.csl

malte-frey:
Added Citation option for posistion="ibid"

### DIFF
--- a/kunstakademie-munster.csl
+++ b/kunstakademie-munster.csl
@@ -282,7 +282,10 @@
   <citation et-al-min="3" et-al-use-first="2" disambiguate-add-year-suffix="true">
     <layout suffix="." delimiter="; ">
       <choose>
-        <if position="subsequent">
+        <if position="ibid">
+          <text term="ibid" form="short">
+        </if>
+        <else-if position="subsequent">
           <text macro="author-short"/>
           <choose>
             <if type="broadcast" match="any">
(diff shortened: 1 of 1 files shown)


kunstakademie-munster.csl
```diff
@@ -282,7 +282,10 @@
   <citation et-al-min="3" et-al-use-first="2" disambiguate-add-year-suffix="true">
     <layout suffix="." delimiter="; ">
       <choose>
-        <if position="subsequent">
+        <if position="ibid">
+          <text term="ibid" form="short">
```
adam3smith:
```suggestion
          <text term="ibid" form="short"/>
```

adam3smith (changes requested):

POBrien333:
The way you implemented it right now is incomplete.
Test for ibid-with-locator first and add ibid+locator, then ibid, subsequent, else.

POBrien333:
Could you provide a new link to the style's guidelines, please?

POBrien333:
Thanks. However, I don't see anything about ibid in the style guide. Can you clarify?

malte-frey:
You are right. Unfortunately, this style guide is lacking severely in terms of accuracy. My dissertational supervisor is the dean of this university and wanted to implement ibid.

POBrien333:
In that case, I don't think we'd accept this PR. The guidelines is the one thing we can refer to as repository maintainers. If one Dean says one thing, then another will say another thing.
Of course you can use the style yourself for your specific work. Make sure to give the style a different name, link and self ID. That way you can have both styles installed.

malte-frey:
I understand. Thank you

PR #7767, opened 2025-09-01
Chicago: Improve URL handling

adunning:
Retain access dates in variants that do not print URLs. Fix punctuation with legal references including URLs.

### DIFF
--- a/chicago-author-date-17th-edition.csl
+++ b/chicago-author-date-17th-edition.csl
@@ -1093,7 +1093,7 @@
       </group>
       <text macro="source-medium-bib"/>
       <text macro="source-archive-bib"/>
-      <text macro="source-DOI-URL-bib"/>
+      <text macro="source-date-accessed-DOI-URL-bib"/>
     </group>
   </macro>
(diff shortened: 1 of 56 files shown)


bwiernik:
What's the reasoning behind access date if no url?

adunning:
_CMOS_ 13.6 is the relevant section allowing the title of the citation to linked and the URL removed; it doesn't say what to do with access dates, but it doesn't say that any other adaptations should be made to the citation. Chicago only shows the access date if no other date is available, so I now think that it makes more sense to rely on this filtering mechanism.

To take an [example from the quick citation guide](https://www.chicagomanualofstyle.org/tools_citationguide/citation-guide-1.html), I think that the access date is crucial information, whether or not the URL is printed:

> 1 . “About Yale: Yale Facts,” Yale University, accessed March 8, 2022.

By contrast the citation looks incomplete without either n.d. or an access date:

> 1 . “About Yale: Yale Facts,” Yale University.

bwiernik:
Thanks sounds good

PR #7774, opened 2025-09-03
Update the-journal-of-bone-and-joint-surgery.csl (make independent st…

POBrien333:
https://forums.zotero.org/discussion/126479/need-jbjs-compliant-reference-format


POBrien333:
@adam3smith 
I'll let you merge this if you agree this is the correct style.

bwiernik:
@POBrien333 what are the differences between the AMA template style and this one?

POBrien333:
The addition of the "month" for journal articles. (Publications show it)

PR #7779, opened 2025-09-06
Fix issue without volume in apa.csl

simvrh:
Issue still needs to be in brackets and non-italic even if no volume, please see:

https://apastyle.apa.org/style-grammar-guidelines/references/examples/journal-article-references#3

### DIFF
--- a/apa.csl
+++ b/apa.csl
@@ -1529,7 +1529,7 @@
             </group>
           </if>
           <else>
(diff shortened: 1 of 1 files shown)


adunning:
This is already fixed in #7510.

bwiernik:
Closing in favor of the bigger PR

PR #7785, opened 2025-09-10
infoclio.ch: add support for periodicals

nchachereau:
A small update to the infoclio.ch styles, for history students (and scholars) - originally submitted in #234 and last updated in #7191.

This update adds support for items of type `periodical`, allowing users to cite specific issues or even a journal as a whole. 

Furthermore, it enables users to respect the style guide more closely, by introducing a difference between articles that exist both in print and online (in which case the standard citation is followed by the keyword "Online:" before the URL), and those that exist only online (in which case the keyword is left out). Articles without indication of pages are assumed to exist only online.

As always, thank you for your continued work on maintaining the CSL styles! :pray: 

cc @janbaumanninfoclio

### DIFF
--- a/infoclio-de-kurzbelege.csl
+++ b/infoclio-de-kurzbelege.csl
@@ -472,13 +472,6 @@
       <if type="webpage post post-weblog dataset" match="any">
         <text macro="url-or-doi"/>
       </if>
-      <else-if type="periodical">
-        <choose>
-          <if variable="note">
-            <text macro="url-or-doi"/>
-          </if>
-        </choose>
-      </else-if>
       <else-if type="article-journal">
         <choose>
           <if variable="page" match="none">
@@ -490,20 +483,13 @@
   </macro>
   <macro name="url-non-web-documents">
     <choose>
-      <if type="periodical">
-        <choose>
-          <if variable="note" match="none">
-            <text macro="url-or-doi-with-online-term"/>
-          </if>
-        </choose>
-      </if>
-      <else-if type="article-journal">
+      <if type="article-journal">
         <choose>
           <if variable="page">
             <text macro="url-or-doi-with-online-term"/>
           </if>
         </choose>
-      </else-if>
+      </if>
       <else-if type="webpage post post-weblog dataset" match="none">
         <text macro="url-or-doi-with-online-term"/>
       </else-if>
--- a/infoclio-de-kurzbelege.csl
+++ b/infoclio-de-kurzbelege.csl
@@ -25,7 +25,7 @@
     <category citation-format="note"/>
     <category field="history"/>
     <category field="social_science"/>
-    <updated>2024-09-04T12:30:00+02:00</updated>
+    <updated>2025-09-10T14:15:00+02:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="de">
@@ -73,6 +73,7 @@
       <group delimiter=", ">
         <group delimiter=": ">
           <text macro="creator"/>
+          <text macro="periodical-info"/>
           <group delimiter=", ">
             <text macro="title"/>
             <text macro="scientific-editor"/>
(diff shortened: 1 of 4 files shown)


infoclio-de-kurzbelege.csl
```diff
@@ -444,31 +472,58 @@
       <if type="webpage post post-weblog dataset" match="any">
         <text macro="url-or-doi"/>
       </if>
+      <else-if type="periodical">
+        <choose>
+          <if variable="note">
```
bwiernik: What is the intent behind testing for `note`?

nchachereau: Sorry, I should have explained this in the PR. The goal is to allow users to make the distinction between print+online and online-only for periodicals. Unlike the missing page indication for journal articles, we needed a way for the user to manually indicate "this exists only online". We considered many other variables (e.g. `status`, `genre`, `medium`), but ultimately `note` seemed the safest solution (other variables are often used in other styles and the content of the variable would thus appear in their citations; we consider supporting `status` in the future for e.g. forthcoming publications).

Problems should also be limited by the fact that we only implement the `note` test for periodicals, which are unlikely to be a very common item in existing databases.

(Ideally, we would want to test not only for the presence of the variable, but for its content (e.g. `note: online`), but that is obviously not possible currently in CSL.)

bwiernik: Hmm, we don't like to use that sort of workaround. 

The best test would probably be to test for the presence of `number`, as online-only journals typically give articles numbers (you would also want to print `number` in this case).

What was the concern with testing for pages?

nchachereau: We do test for the absence of pages for items with type `article-journal` (see a few lines below). Testing for `note` is *only* for items with type `periodical` (specific issues or a journal). We are aware that it is a hack, but could not find a better solution.

bwiernik: I would suggest one of two options:

1. Print `medium` instead of `note` for periodicals--users in Zotero can enter "medium: Online" in Extra to show for online only journals. This is at least valid metadata and doesn't interfere with using that field for actual annotations or other data. 

2 Use the same `pages` check for periodicals. This would capture most online issues, as print issues typically have page ranges

janbaumanninfoclio: Dear Brenton, thank you very much for your suggested solutions. We had also discussed the solution with "medium".
The problem is that if you put medium: online in the extra field, it can look a bit strange in other styles, e.g., in APA 7th edition:
`Stress! (2014). [Online]. Zeithistorische Forschungen / Studies in Contemporary History, 11(3). https://zeithistorische-forschungen.de/3-2014`

Die The second solution you suggested does not work to distinguish between online journals and print journals. Also for print journals, the pages field is usually left blank.

Note that we are only talking about the rather rare case of citing a single issue of a journal or an entire journal series. For journal articles, the pages test works well. We therefore believe that our solution of testing for note: online has the fewest disadvantages and is the best solution for distinguishing between online journals and journals.

bwiernik: That is correct APA format for a journal issue without an editor or other author, though generally a more descriptive medium would be given like [Special issue]. In any event, including descriptive information in this format for unusual items in APA is not incorrect. 

Print issues of a journal typically have a page range, and most journals use continuous pagination across issues within a volume. Although the pages field could be left blank, it would be straightforward for a user to add the relevant page information to the item to make the "Online" tag go away. 

"note" is frequently not blank in real world data, as Zotero and Mendeley and other clients use their Extra/Note fields to store a variety of data. It's not a reliable workaround, and we can't accept a style with this sort of hack in the repository. 

Happy to merge this once you choose one of the options I described.

nchachereau: Thank you for the feedback. Understood, we went ahead with testing for the `medium` variable. Does this constitute an acceptable solution?

janbaumanninfoclio:
I wanted to ask whether our modified version with “medium” instead of “note” can now be accepted for the repository. Thank you very much.

janbaumanninfoclio:
@bwiernik, the pull request is still open. We would be very happy if the request could be completed soon.If there are any further changes required, please let us know. Thank you, Jan

nchachereau:
With the latest commit, we have given up on a special format for periodicals that are only online. We have thus completely given up on the hack initially submitted (and with which @bwiernik understandably took issue).

We hope that this new version can now be merged easily and soon :crossed_fingers: 
Thank you again for your work on the repo.

PR #7786, opened 2025-09-10
Update nature.csl (add DOI for non-published articles)

POBrien333:
https://forums.zotero.org/discussion/126665/citing-in-nature-style-with-doi#latest

### DIFF
--- a/nature.csl
+++ b/nature.csl
@@ -4,17 +4,21 @@
     <title>Nature</title>
     <id>http://www.zotero.org/styles/nature</id>
     <link href="http://www.zotero.org/styles/nature" rel="self"/>
-    <link href="http://www.nature.com/nature/authors/gta/index.html#a5.4" rel="documentation"/>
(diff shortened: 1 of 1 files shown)


nature.csl
```diff
@@ -69,7 +73,7 @@
               <text variable="genre" text-case="capitalize-first"/>
             </if>
             <else>
-              <text term="article" text-case="capitalize-first"/>
+              <text term="preprint" text-case="capitalize-first"/>
```
bwiernik, line 76: These are generally the same, so keeping article would be fine

bwiernik:
Looks good to me

adam3smith:
Only looking quickly, but did we lose the period at the end of the bib entry in the other 2 Nature styles?

POBrien333:
somehow nature is different from the other two versions. E.g.: [https://www.diffchecker.com/MEW9EpSO/](https://www.diffchecker.com/MEW9EpSO/)
I'll just remake those based on the new nature.csl I prepared yesterday.

Issue #6513, opened 2023-04-23
Requesting Style : World Journal of Pediatrics


NourKH2:
Hello, 
I'd like to request the style of the World Journal of Pediatrics. Could you help me? Thank you

POBrien333:
From the looks of it that just follows the standard Springer Vancouver style: https://www.zotero.org/styles?q=id%3Aspringer-basic-brackets&format=numeric&dependent=0

PR #7791, opened 2025-09-14
Update springer-vancouver-brackets.csl (incl. DOI)

POBrien333:
I noticed this while reviewing the style for another dependent style (see https://github.com/citation-style-language/styles/issues/6513)

I noticed that springer-vancouver.csl and springer-vancouver-brackets.csl are both styled like [1] in-text. Is this an error?

### DIFF
--- a/springer-vancouver-brackets.csl
+++ b/springer-vancouver-brackets.csl
@@ -18,7 +18,7 @@
     </contributor>
     <category citation-format="numeric"/>
     <category field="generic-base"/>
-    <summary>Based on the document &quot;Key Style Points 2.0 І December 2024&quot;.</summary>
+    <summary>Based on the document "Key Style Points 2.0 І December 2024".</summary>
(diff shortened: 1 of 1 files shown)


bwiernik:
Possibly an error. Is one supposed to be superscript?

POBrien333:
I honestly don't know. They guidelines don't show how citations in the text should be shown...
https://dam.springernature.com/file/1iVex7Gr46Q8xLzdF9RNhT/*/Key+Style+Points_Dec_24.pdf?authcred=Q29weVVSTDpDMHB5X1VSTA==

I'm thinking, since that style doesn't have dependents, just making it either `1` or `<sup>1</sup>` to differentiate it and see what we get back from users? Or just a dependent on springer-brackets, this style? That way we don't need to maintain 2 styles.

bwiernik:
Let's change the other one to superscript to differentiate. Other publishers' "regular" Vancouver are superscript

PR #7804, opened 2025-09-19
Update les-journees-de-la-recherche-porcine.csl

BenesIfip:
Hello, as an editor of Les Journées de la Recherche Porcine, we would like to commit this update of our Zotero style in order to have no more "et al" in the bibliography.
This is my first contribution in GitHub.
Thank you,
Claire

### DIFF
--- a/les-journees-de-la-recherche-porcine.csl
+++ b/les-journees-de-la-recherche-porcine.csl
@@ -127,7 +127,7 @@
       </group>
     </layout>
   </citation>
-  <bibliography hanging-indent="true" et-al-min="6" et-al-use-first="1" entry-spacing="0">
+  <bibliography hanging-indent="true" et-al-min="19" et-al-use-first="1" entry-spacing="0">
     <sort>
       <key macro="author"/>
       <key variable="title"/>


POBrien333:
Can youupdate the date in the info section?
You can actually also just completely remove those 2 et-al settings. It's cleaner and less ambiguous then.

POBrien333:
@BenesIfip 
2 things:
1. I see that you have some rather odd et-al rules for in-text citations. can you clarify what you actually want for this journal and I can tell you if it does that.
2. I see that you link to a .csl style for direct download
image

I would recommend that you use the following link instead: https://www.zotero.org/styles?q=id%3Ales-journees-de-la-recherche-porcine
That way the style always stays update, no matter if it receives PRs from you or from the CSL community or maintainers for future updates.

BenesIfip:
@POBrien333 
Thank you for your recommendations and help.
For in-text citations, we would like to : 
- write "_et al._" when there are 3 or more authors,
- write both names separated by " et " when there are only 2 authors.

Furthermore, in the bibliography we would prefer NOT to write the web link (for example : _> Consultable : https://www.nature.com/articles/s41597-019-0031-8 [Consulté le 31 mars 2025]._), except when needed (when pages are missing for example).

I haven't changed that because I am not sure whether I can do it without creating new problems. But if you can do it for me, that would be great.

POBrien333:
@BenesIfip 
Thanks for the clarification.
I've corrected the et-al rules and built in some logic for the URL.
It will print the DOI/URL for article-journal if no volume+page is available. It will print the URL for webpages/blogs/posts.

Please test it out and then we can merge.

BenesIfip:
@POBrien333 
Thanks a lot for et-al rules and URL/DOI rules, it works very well.
After testing it out, I realized 2 things:
- I should have added a point after "et al". So I have added it in "term name="et-al">et al.</term"
- It seems that there is a useless  " : " after URL, for example : 
> Consultable : https://dr.lib.iastate.edu/handle/20.500.12876/RwyqoG3w : [Consulté le 23 avril 2025].

So I have deleted " : " in the group delimiter just after "macro name="URL""

Is it correct ?
Will it still possible to merge our changes ?

POBrien333:
The "et al." is already the normal translation in fr-FR, as can be seen in our French locale. So that line can just be deleted. https://github.com/citation-style-language/locales/blob/master/locales-fr-FR.xml

I've fixed the " : " issue.

(Note though that you didn't make any changes in this PR, so I did them now)

BenesIfip:
Thank you again for your great help @POBrien333 (and sorry, I didn't realize that I wasn't in this PR when modifying the file.)

PR #7807, opened 2025-09-23
Create gdr-sciences-du-bois.csl

potguillaume:
Style for "GDR sciences du bois" conference: https://gdr-sciences-du-bois.hub.inrae.fr/

### DIFF
--- a/gdr-sciences-du-bois.csl
+++ b/gdr-sciences-du-bois.csl
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" default-locale="en-US">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" default-locale="fr-FR">
   <info>
-    <title>gdr sciences du bois</title>
-    <title-short>gdr</title-short>
+    <title>GDR Sciences du bois (Français)</title>
+    <title-short>GDR</title-short>
     <id>http://www.zotero.org/styles/gdr-sciences-du-bois</id>
     <link href="http://www.zotero.org/styles/gdr-sciences-du-bois" rel="self"/>
     <link href="http://www.zotero.org/styles/springer-basic-author-date-no-et-al-with-issue" rel="template"/>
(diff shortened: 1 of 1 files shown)


POBrien333:
Can you please check the following guide and adapt your style to contain all the necessary information? Specifically, the template style and a link to the style guidelines.
https://github.com/citation-style-language/styles/blob/master/STYLE_REQUIREMENTS.md

potguillaume:
@POBrien333 I think I have made the requested modifications, and commited the changes. Should I do something else?

gdr-sciences-du-bois.csl
```diff
@@ -0,0 +1,181 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="never" default-locale="en-US">
+  <info>
+    <title>gdr sciences du bois</title>
+    <title-short>gdr</title-short>
+    <id>http://www.zotero.org/styles/gdr-sciences-du-bois</id>
+    <link href="http://www.zotero.org/styles/gdr-sciences-du-bois" rel="self"/>
+    <link href="http://www.zotero.org/styles/springer-basic-author-date-no-et-al-with-issue" rel="template"/>
+    <link href="https://gdr-sciences-du-bois.hub.inrae.fr/manifestations/les-manifestations-du-gdr-bois" rel="documentation"/>
```
POBrien333: I currently don't see at that link where the style is described. Can you pinpoint this for me?

potguillaume: Actually the role of the documentation link is still unclear to me, on `[this new link ](https://gdrbois-2025.sciencesconf.org/data/pages/GDR_bois_2025_NOM_Prenom_Labo_1.docx)`, the instruction for author are given

POBrien333: Because we only accept documented styles and also want to check the style actually does what it is supposed to do.
Also, in the future when we get an error report, without guidelines we will not be able to check the error is an error and provide a fix.

potguillaume (commented):
This changes make the style not working with zotero, error of invalid type file.

POBrien333:
The validation pipeline shows no errors and downloading the style and installing it in Zotero works just fine (otherwise Zotero throws an error). I assume you're doing something wrong on your end.

PR #7808, opened 2025-09-23
Update associacao-brasileira-de-normas-tecnicas-ufrgs.csl (refactor)

POBrien333:
I noticed this style was used as a template for other ABNT variants and I was requesting the same changes. This PR fixes issues with this style, so we don't see them in the future when it is used as a template.

Main issues:
- Use of `value` instead of term/label
- Lack of use of groups
- repetition of elements that could be done via a macro

(nota bene, I've added a bunch of macros that are not shown as changes, oddly enough)

### DIFF
--- a/associacao-brasileira-de-normas-tecnicas-ufrgs.csl
+++ b/associacao-brasileira-de-normas-tecnicas-ufrgs.csl
@@ -820,7 +820,7 @@
                 <text variable="publisher" suffix=", "/>
                 <text macro="volume" suffix=", "/>
                 <text macro="number" suffix=", "/>
-                <text macro="page" suffix=", "/>
+                <text macro="pages" suffix=", "/>
                 <text macro="issued-legislation" suffix=". "/>
                 <text variable="genre" suffix=". "/>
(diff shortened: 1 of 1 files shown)


POBrien333:
apparently I messed up this morning.
I accidently pushed this straight to master, instead of making a PR. I was able to revert it and make this PR. But it seems like other PRs since say that there is an error in the abnt style. e.g. #7809 
Could you help me here, @bwiernik 
sorry for the issue.

POBrien333:
@bwiernik 
Would you be able to help me with this issue?
This style now pops up in new PRs giving an error.

POBrien333:
I was able to fix the issue.
I thought my "reversion" worked, but apparently my non-validation style update was still live and was giving these errors. Merging this PR fixed that and therefore also the issue. :)

bwiernik:
@POBrien333 okay great. Sorry I just moved and haven't been at computer in a few days. Let me know if you need anything else

POBrien333:
No worries. Thought you were travelling/moving. :)
Thanks

PR #7810, opened 2025-09-24
Update uniwersytet-gdanski-kulturoznawstwo.csl

kjklosowski:
Two changes here:
- Added `author` and `author-bibliography` in corresponding `type="software"` to reflect a situation where the software (or a video game) is authored by a programmer, or a group of programmers, instead of a sofware house/studio, as per internal request I received. In Zotero, a `"Programmer"` field is just `author` in CSL, so this is a fairly simple change, but will be reflected in rendering and sorting (these styles sort by authors and then titles if no author present).
- Slightly rephrased `summary` for clarity.

I'm the original author of the style (#7549, #7547).

Thanks!

### DIFF
--- a/uniwersytet-gdanski-kulturoznawstwo-autor-rok.csl
+++ b/uniwersytet-gdanski-kulturoznawstwo-autor-rok.csl
@@ -272,7 +272,7 @@
               <name delimiter-precedes-last="always" initialize="false" initialize-with=". " name-as-sort-order="all" sort-separator=" "/>
               <et-al font-style="normal"/>
             </names>
-            <names variable="producer" prefix="prod. ">
+            <names variable="producer">
               <name delimiter-precedes-last="always" initialize="false" initialize-with=". " name-as-sort-order="all" sort-separator=" "/>
               <et-al font-style="normal"/>
             </names>
--- a/uniwersytet-gdanski-kulturoznawstwo-autor-rok.csl
+++ b/uniwersytet-gdanski-kulturoznawstwo-autor-rok.csl
@@ -22,14 +22,6 @@
   <locale xml:lang="pl">
     <terms>
       <term name="et-al">i in.</term>
-      <term name="number" form="short"> <!-- The default-locale term is currently broken -->
-        <single>nr</single>
-        <multiple>nr.</multiple>
-      </term>
-      <term name="collection-number" form="short"> <!-- The default-locale term is currently broken -->
-        <single>nr</single>
-        <multiple>nr.</multiple>
-      </term>
     </terms>
   </locale>
   <macro name="title">
--- a/uniwersytet-gdanski-kulturoznawstwo-autor-rok.csl
+++ b/uniwersytet-gdanski-kulturoznawstwo-autor-rok.csl
@@ -16,13 +16,17 @@
     <category field="sociology"/>
     <category field="anthropology"/>
     <summary>A Polish in-text citation style based on Uni. of Gdansk Cultural Studies Department guidelines. Wide use range for cultural studies, digital humanities</summary>
-    <updated>2025-09-27T15:52:20+00:00</updated>
+    <updated>2025-09-27T16:51:48+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="pl">
     <terms>
       <term name="et-al">i in.</term>
       <term name="number" form="short"> <!-- The default-locale term is currently broken -->
+        <single>nr</single>
+        <multiple>nr.</multiple>
+      </term>
+      <term name="collection-number" form="short"> <!-- The default-locale term is currently broken -->
         <single>nr</single>
         <multiple>nr.</multiple>
       </term>
@@ -63,8 +67,12 @@
           <if match="any" variable="collection-title">
             <group delimiter=" ">
               <text variable="collection-title"/>
-              <text term="number" form="short"/>
-              <text variable="collection-number"/>
+              <choose>
+                <if match="any" variable="collection-number">
+                  <label variable="collection-number" form="short"/>
+                  <text variable="collection-number"/>
+                </if>
+              </choose>
             </group>
           </if>
         </choose>
@@ -197,8 +205,12 @@
               <if match="any" variable="collection-title">
                 <group delimiter=" ">
                   <text variable="collection-title"/>
-                  <text term="number" form="short"/>
-                  <text variable="collection-number"/>
+                  <choose>
+                    <if match="any" variable="collection-number">
+                      <label variable="collection-number" form="short"/>
+                      <text variable="collection-number"/>
+                    </if>
+                  </choose>
                 </group>
               </if>
             </choose>
@@ -226,7 +238,7 @@
             <choose>
               <if match="any" variable="number">
                 <group delimiter=" ">
-                  <text term="number" form="short"/>
+                  <label variable="number" form="short"/>
                   <text variable="number"/>
                 </group>
               </if>
@@ -329,8 +341,12 @@
               <if match="any" variable="collection-title">
                 <group delimiter=" ">
                   <text variable="collection-title"/>
-                  <text term="number" form="short"/>
-                  <text variable="collection-number"/>
+                  <choose>
+                    <if match="any" variable="collection-number">
+                      <label variable="collection-number" form="short"/>
+                      <text variable="collection-number"/>
+                    </if>
+                  </choose>
                 </group>
               </if>
             </choose>
--- a/uniwersytet-gdanski-kulturoznawstwo-autor-rok.csl
+++ b/uniwersytet-gdanski-kulturoznawstwo-autor-rok.csl
@@ -16,19 +16,20 @@
     <category field="sociology"/>
     <category field="anthropology"/>
     <summary>A Polish in-text citation style based on Uni. of Gdansk Cultural Studies Department guidelines. Wide use range for cultural studies, digital humanities</summary>
-    <updated>2025-09-27T18:39:48+00:00</updated>
+    <updated>2025-10-05T15:08:38+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="pl">
     <terms>
       <term name="et-al">i in.</term>
+      <term name="entry">hasło</term>
     </terms>
   </locale>
   <macro name="title">
     <text variable="title" text-case="capitalize-first" quotes="false" font-style="italic"/>
   </macro>
   <macro name="title-short">
-    <text variable="title" form="short" text-case="capitalize-first" font-style="normal" prefix="„" suffix="”"/>
+    <text variable="title" form="short" text-case="capitalize-first" quotes="false"/>
   </macro>
   <macro name="author">
     <names variable="author">
@@ -37,13 +38,13 @@
   </macro>
   <macro name="author-bibliography">
     <names variable="author">
-      <name delimiter-precedes-last="always" initialize="false" initialize-with=". " name-as-sort-order="all" sort-separator=" "/>
+      <name delimiter-precedes-last="always" initialize="false" name-as-sort-order="all" sort-separator=" "/>
     </names>
   </macro>
   <macro name="editor-translator-bibliography">
     <names variable="editor translator" delimiter=", ">
       <label form="short" suffix=" "/>
-      <name delimiter-precedes-last="always" initialize="false" initialize-with=". " name-as-sort-order="all" sort-separator=" "/>
+      <name delimiter-precedes-last="always" initialize="false" name-as-sort-order="all" sort-separator=" "/>
     </names>
   </macro>
   <macro name="container-bibliography">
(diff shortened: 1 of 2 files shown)


uniwersytet-gdanski-kulturoznawstwo-autor-rok.csl
```diff
@@ -15,8 +15,8 @@
     <category field="humanities"/>
     <category field="sociology"/>
     <category field="anthropology"/>
-    <summary>A Polish in-text citation style with wide use range for cultural studies, digital humanities. Based on Uni. of Gdansk guildelines</summary>
-    <updated>2025-08-10T17:26:29+00:00</updated>
+    <summary>A Polish in-text citation style based on Uni. of Gdansk Cultural Studies Department guidelines. Wide use range for cultural studies, digital humanities</summary>
+    <updated>2025-09-24T15:17:52+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="pl">
```
POBrien333, line 22: You can delete this whole "locale" section since the et al. term is already like this in our pl-PL locale: https://github.com/citation-style-language/locales/blob/master/locales-pl-PL.xml

POBrien333 (changes requested):
This style can be a lot more simplified:
1. you can just call author, title, [conditional for all item types]. 
That way any item types receives an author+title no matter what. Further bibliographic details can then be output via your conditional.
2. You have a lot of conditionals that test for a variable to then output said variable. These are all obsolete and you can remove those.

kjklosowski:
Hi @POBrien333! Thanks for looking at the code.

Yeah, the redundant if-statements are definietly overengineered, I regret ever doing it that way and we actually had a conversation about it during initial PR.

I have optimized the code greatly, removed all redundant if-statements, improved group-delimiters logic in some places. The only ifs that remained are there to guard against orphan term-labels showing up, when the user (for whatever reason) decides to leave a particular variable empty (look for example the usage of `edition` macro. I would prefer to avoid that, as this style can be used for many different applications. So the terms should remain guarded with an if.

As for locale, I've received yet another request to change these terms to their Polish counterparts. The documentation permits both (provided that one is used consistently!), but in the given Department the Polish ones are used much more frequently, hence the change.

I have not ventured into your first point `"you can just call author, title, [conditional for all item types]."`. I mean, it can be done, it would decrease the size of file, but it would not decrease the amount of times each object is called (in a given citation object). I know what you mean by that, I can play around with it if you want, but **for me** it would make the file less readable, require some additional prefix work (as some author-title-the rest are delimited by `,`, while other with a `space`). I'm personally not that into it.

The ifs were most important, and they are gone!

I have confirmed no substantial behavior change against a 5 page long diverse bibliography and citations!

uniwersytet-gdanski-kulturoznawstwo-przypis.csl
```diff
@@ -36,36 +36,34 @@
...
       <name delimiter-precedes-last="always" initialize-with=". "/>
-      <et-al font-style="italic"/>
+      <et-al font-style="normal"/>
     </names>
   </macro>
   <macro name="author-bibliography">
     <names variable="author">
       <name delimiter-precedes-last="always" initialize="false" initialize-with=". " name-as-sort-order="all" sort-separator=" "/>
-      <et-al font-style="italic"/>
+      <et-al font-style="normal"/>
```
kjklosowski: I think `font-style="normal"` can be removed also? I added it to italize the latin terms. Can you confirm that @POBrien333?

POBrien333: That whole line can be removed since nothing is "done" to the et-al.

POBrien333:
To remove the validation error:
Can you go here and click on "Sync fork"?
https://github.com/kjklosowski/styles/tree/patch-1

uniwersytet-gdanski-kulturoznawstwo-przypis.csl
```diff
@@ -36,36 +36,34 @@
...
       <text variable="container-title" quotes="false" font-style="italic"/>
-      <choose>
-        <if match="any" variable="editor translator">
-          <text macro="editor-translator"/>
-        </if>
-      </choose>
+      <text macro="editor-translator"/>
       <choose>
         <if match="any" variable="edition">
           <text macro="edition"/>
```
POBrien333: If you use this "edition" macro, you can get rid of all these conditionals:
`  <macro name="edition">
    <choose>
      <if is-numeric="edition">
        <group delimiter=" ">
          <number variable="edition" form="ordinal"/>
        </group>
      </if>
      <else>
        <text variable="edition"/>
      </else>
    </choose>
  </macro>`

uniwersytet-gdanski-kulturoznawstwo-przypis.csl
kjklosowski: Oh I thought it would impact et-al behavior. I will remove all of them

uniwersytet-gdanski-kulturoznawstwo-przypis.csl
```diff
@@ -15,15 +15,15 @@
...
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="pl">
     <terms>
-      <term name="ibid">ibid.</term>
-      <term name="et-al">et al.</term>
-      <term name="op-cit">op. cit.</term>
+      <term name="ibid">tamże</term>
+      <term name="et-al">i in.</term>
+      <term name="op-cit">dz. cyt.</term>
```
POBrien333, line 26: Our polish locale probably just still has the English term in it. Would you say that "dz. cit." should be the general translation?
https://github.com/citation-style-language/locales/blob/208128470ef09fba9fdb5bae5fb442a91e364668/locales-pl-PL.xml#L45

uniwersytet-gdanski-kulturoznawstwo-przypis.csl
kjklosowski: Oh I understand your logic. I can definietly include the if-statement in the macro and then remove it in actual citations. Will do it!

uniwersytet-gdanski-kulturoznawstwo-przypis.csl
kjklosowski: Actually the polish locale used to include the "i in." but we changed it to latin with Brenton, as that is generally more common in Poland, but in that style Polish counterparts could work as well. See: [Update locales-pl-PL.xml](https://github.com/citation-style-language/locales/pull/353).

I can just remove the local section and stick with default global Polish locale if that's what you prefer, it's not that important.

POBrien333: Nono, I'm just asking to make sure we don't duplicate anything. If you need it, absolutely go for it :)

uniwersytet-gdanski-kulturoznawstwo-autor-rok.csl
```diff
@@ -16,12 +16,16 @@
...
     <category field="anthropology"/>
     <summary>A Polish in-text citation style based on Uni. of Gdansk Cultural Studies Department guidelines. Wide use range for cultural studies, digital humanities</summary>
-    <updated>2025-09-27T13:06:45+00:00</updated>
+    <updated>2025-09-27T15:52:20+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="pl">
     <terms>
       <term name="et-al">i in.</term>
+      <term name="number" form="short"> <!-- The default-locale term is currently broken -->
```
POBrien333: Is this because you're using the Code/VisualEditor at https://editor.citationstyles.org/codeEditor/? 

You can ignore that and remove this. If you test it out inside the Zotero editor you'll see it works. The CSL Code/VisualEditor needs an update, that's why.

kjklosowski: I've been just writing a comment about this.

Actually, Zotero behaves the same way, so I'm still concerned about it.

Zotero defaults to English term, Visual Editor outputs nothing.

What could be the reason behind it?

POBrien333: You need to call `<label variable="number" form="short"/>`

(Note also that collection-number exists as a label that you could use where you call the collection-number variable.)

kjklosowski: Still the same behavior inside Zotero with label variable :/

EDIT: @POBrien333 Changed to label variables. Still the same thing. And collection-number behaves the same way.

POBrien333: Works fine for me for report which is the only item type where you use "number".
image

POBrien333:
It looks like you have changed the settings that we as admins can't change the code in your PR without making PRs on your branch. Can you change that?

uniwersytet-gdanski-kulturoznawstwo-autor-rok.csl
kjklosowski: After performing full Zotero reinstall, the issue is gone.

Sorry for that hustle, I have no idea what caused this erratic behaviour.

Thanks @POBrien333!

kjklosowski:
> It looks like you have changed the settings that we as admins can't change the code in your PR without making PRs on your branch. Can you change that?

Changed it. Thanks @POBrien333 for your assistance. The issue is now resolved after reinstalling Zotero!

I've removed these terms :)

uniwersytet-gdanski-kulturoznawstwo-przypis.csl
```diff
@@ -242,167 +185,101 @@
...
-                    </choose>
-                    <choose>
-                      <if match="any" variable="edition">
-                        <text variable="edition" prefix="wyd. " suffix=", "/>
-                      </if>
-                    </choose>
+                  <group delimiter=", ">
+                    <text macro="author"/>
+                    <text macro="title"/>
+                    <text variable="container-title" font-style="italic" prefix="[hasło w:] "/>
```
POBrien333: what does the "haslo w" mean? 
We try to use terms/labels where possible and not hardcode text like this.

uniwersytet-gdanski-kulturoznawstwo-przypis.csl
```diff
@@ -242,167 +185,101 @@
...
+                    <text macro="author"/>
+                    <text variable="container-title" font-style="italic"/>
                     <text macro="publisher-place-date"/>
-                    <choose>
-                      <if match="any" variable="locator">
-                        <text macro="locator"/>
-                      </if>
-                    </choose>
+                    <text macro="locator"/>
                     <text macro="title" prefix="[hasło: " suffix="]"/>
```
POBrien333: same as above

uniwersytet-gdanski-kulturoznawstwo-przypis.csl
```diff
@@ -555,117 +373,71 @@
...
+          <group delimiter=", ">
+            <text macro="author-bibliography"/>
+            <text macro="title"/>
             <choose>
-              <if match="any" variable="composer">
-                <names variable="composer" suffix=", ">
-                  <name delimiter-precedes-last="always" initialize="false" initialize-with=". " name-as-sort-order="all" sort-separator=" "/>
-                  <et-al font-style="italic"/>
-                </names>
+              <if match="any" variable="collection-title">
```
POBrien333, line 419: Can you explain what you are trying to do with this test? It seems a bit odd to me.

POBrien333 (changes requested):
Hi
I made some improvements. See how I created the "collection" macro that I call everywhere.
I also removed some more superfluous conditionals.
There is some remaining comments.
If you can, please then make these changes also in the other file. :)

uniwersytet-gdanski-kulturoznawstwo-przypis.csl
kjklosowski: haslo w means entry in (the dictionary). We could use the term but in polish locale it's still left untranslated, so this would need to get taken care of as well. (see: https://github.com/citation-style-language/locales/blob/master/locales-pl-PL.xml#L132). However, translating entry to 'hasło' would be insufficient, the 'w:' needs to be there as well, to this part would have to remain hardcoded? 'w:' should not be in the locale as well. Please let me know and I can make a PR there as well. But I think it can be done.

uniwersytet-gdanski-kulturoznawstwo-przypis.csl
kjklosowski: Actually all ifs you have deleted in your commit preceded terms and labels. It's the only change I'm against as it can cause the label being present when there is no actual number following it.

I actually had an issue where the book was a part of a collection that had a name but had no number, the 'nr' label was printed anyways, making this unusable. It looked like 'Author, Book title, publisher info, collection title, nr .'

I think that all labels should have these ifs before for that particular reason. Can you reconsider that and let me readd these? There would be only a couple of them. As this style can have many applications with many source types, I think it is important to mitigate this.

kjklosowski:
Hi @POBrien333! Thanks for your contribution to my style.
I've answered you in these discussions you opened :)

I will take care of the 'haslo' term and add changes to the other file! I will also fix one mistake in journal macro I've spotted

uniwersytet-gdanski-kulturoznawstwo-przypis.csl
kjklosowski: A leftover I missed in my deletions of redundant if statements :)

uniwersytet-gdanski-kulturoznawstwo-przypis.csl
POBrien333: As you see, I've put them in a group. A group is like a conditional and it will only print out the label if the variable is also present. Test it out and you'll see.

kjklosowski: Oh that's interesting. I completly misunderstood that. Thanks for clarifying! Then all ifs can be gone!

uniwersytet-gdanski-kulturoznawstwo-przypis.csl
POBrien333: Please:
1. provide a generic translation in a PR for the pl-PL locale (ideally you also add other untranslated terms to help the pl-PL community)
2. Add the "haslo w" in your locale overrides and then use the correct term in the style

kjklosowski: Gotcha!

kjklosowski:
@POBrien333. Files updated, I've found and deleted some more unnecessary things and fixed couple of bugs and discrepancies.

I've confirmed against 5 page long diverse bibliography: no unexpected behaviour change.

Sooner or later, I will delete 'entry' term, as I plan to work on the Polish locale soon and translate it :)

POBrien333:
I've merged the style now and it'll show up on zotero.org/styles within an hour or so. You can then use that link to share the style out.
If you spot any issues, of course you can make a new PR.

kjklosowski:
Thank you @POBrien333 for your dedication and help!

PR #7811, opened 2025-09-25
Chicago: Refine archives, multivolume works, inaccessible personal communications; additional variants

adunning:
When providing a physical locator (`column`, `folio`, `page`) with an item that has a full archival reference (at least `archive` and `archive_location`), assume that the locator is part of the archival item (as shown in CMOS18 14.127). The locator must always be labelled in this case.

Fix a regression introduced with the styles for the 18th edition wherein users could receive an archival reference formatted as a database reference (as in CMOS18 14.113). The styles now revert to following the CSL specification in always handling `archive_location` as a 'storage location', rather than attempting to use this as a database reference in some situations, which is now assumed to be stored in `number`. This corrects the display of items where the archive place is part of the archive name, as in 'Lambeth Palace Library, MS 51'.

In shortened notes, when citing multiple items with a `genre` only (such as untitled introductions by the same author), disambiguate using a `container-title` if necessary.

### DIFF
--- a/chicago-author-date-17th-edition.csl
+++ b/chicago-author-date-17th-edition.csl
@@ -255,15 +255,11 @@
             </else>
           </choose>
         </else-if>
+        <else-if is-numeric="locator" locator="line">
+          <label variable="locator"/>
+        </else-if>
         <else-if is-numeric="locator">
-          <choose>
-            <if locator="line">
-              <label variable="locator"/>
-            </if>
-            <else>
-              <label form="short" variable="locator"/>
-            </else>
-          </choose>
+          <label form="short" variable="locator"/>
         </else-if>
         <else-if locator="chapter line verse" match="any"/>
         <!-- a non-numeric canonical reference is identified by its formatting and does not need a label (CMOS18 14.143-54) -->
@@ -4012,14 +4008,11 @@
             </else-if>
           </choose>
         </else-if>
-        <else-if match="any" type="interview personal_communication">
-          <choose>
-            <if match="any" variable="archive archive-place container-title DOI number publisher references URL">
-              <!-- Personal communications only appear in the bibliography if the reader can retrieve them (CMOS18 14.13, 14.111) -->
-              <text macro="bibliography-author-date"/>
-            </if>
-          </choose>
+        <!-- Personal communications only appear in the bibliography if the reader can retrieve them (CMOS18 14.13, 14.111) -->
+        <else-if match="any" variable="archive archive-place container-title DOI number publisher references URL">
+          <text macro="bibliography-author-date"/>
         </else-if>
+        <else-if match="any" type="interview personal_communication"/>
         <else>
           <text macro="bibliography-author-date"/>
         </else>
--- a/chicago-author-date-access-dates.csl
+++ b/chicago-author-date-access-dates.csl
@@ -258,15 +258,11 @@
             </else>
           </choose>
         </else-if>
+        <else-if is-numeric="locator" locator="line">
+          <label variable="locator"/>
+        </else-if>
         <else-if is-numeric="locator">
-          <choose>
-            <if locator="line">
-              <label variable="locator"/>
-            </if>
-            <else>
-              <label form="short" variable="locator"/>
-            </else>
-          </choose>
+          <label form="short" variable="locator"/>
         </else-if>
         <else-if locator="chapter line verse" match="any"/>
         <!-- a non-numeric canonical reference is identified by its formatting and does not need a label (CMOS18 14.143-54) -->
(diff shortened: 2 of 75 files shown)


bwiernik:
> The styles now revert to following the CSL specification in always handling archive_location as a 'storage location', rather than attempting to use this as a database reference in some situations, which is now assumed to be stored in number. This corrects the display of items where the archive place is part of the archive name, as in 'Lambeth Palace Library, MS 51'.

In APA, we sometimes assume that `archive_location` is a database reference if I recall correctly. Would testing for `archive-place` be a clearer way to distinguish physical and digital archives?

adunning:
This is my fourth iteration on approaches to handling archival references:

1. Initially I was using the approach in the current APA styles; this misrenders many more complicated examples, as you will see from comparing renderings of the test library in <https://github.com/citation-style-language/styles/pull/7510#issuecomment-3334051640>.
2. I then attempted a revised version based on the presence of `archive_collection` or `archive-place`. This also led to problems reported on the forums, because in the example I cited, 'Lambeth Palace Library' is the full `archive` name and 'MS 51' is the the full `archive_location`; there is nothing more that can be added.
3. I tried to make an exception for types such as `manuscript`, which is reflected in the CMOS styles currently in the repository. This also has not worked in practice: for example, one can easily have an unpublished thesis which is only accessible in a particular library. For example, a thesis available as 'Cambridge University Library, MS Add. 123' would correctly use only `archive` and `archive_location`.
4. Having cycled through the above three methods, the only remaining approach that I can see is to follow the spec's direction to handle `archive_location` as part of an archival reference in all situations.

I haven't combed through the commit history, but my guess is that treating `archive_location` as a database reference derives from the need to record ProQuest numbers for a thesis, as in CMOS18 14.113 or APA 503. In practice, I've never seen anyone cite a thesis with this identifier. It was only necessary in the days before ProQuest had open URLs. In any case the `number` variable is a much better place for this, since this is firmly established as the correct variable for an identifier in a dataset.

Let me know if you can see another option that I have overlooked!

bwiernik:
What specific items are you referring to in (1)?

For (2), even if the place for 'Lambeth Palace Library' isn't shown, the archive still has a physical address that could be tested for.

adunning:
1. For examples of items not working in the published APA styles, see e.g. <https://www.zotero.org/groups/2205533/items/ESNIH8AM>, <https://www.zotero.org/groups/2205533/items/54V2UIU8>, <https://www.zotero.org/groups/2205533/items/RR2RS99S>, <https://www.zotero.org/groups/2205533/items/E2X568PN>, <https://www.zotero.org/groups/2205533/items/BBHDZPC7>, <https://www.zotero.org/groups/2205533/items/MGJXM6D3>
2. You could put 'Lambeth' in `archive-place`, but then you would end up with the nonsensical citation 'Lambeth, Lambeth Palace Library'.

bwiernik:
> You could put 'Lambeth' in archive-place, but then you would end up with the nonsensical citation 'Lambeth, Lambeth Palace Library'.

Personally, I don't find that nonsensical. The name of the library and the city it's in happen to contain the same word.

If the citation style calls for  archive-place to be rendered, I don't see any problem, showing it in this example.

adunning:
Several of these examples are drawn directly from the Chicago Manual. The `archive-place` is to be filled in as needed, and is not a mandatory field. It isn't even exposed in Zotero, so it's not reasonable to require users to use the `archive-place` variable. Most of all, this doesn't follow the CSL spec, which defines `archive_location` as 'Storage location within an archive (e.g. a box and folder number)'.

adunning:
It is also worth pointing out that where Zotero supports importing from archival catalogues, it appears only to cover `archive` and `archive_location`, e.g. <https://discovery.nationalarchives.gov.uk/details/r/C6203263>. I should emphasize that personally I would always include an `archive-place` in citations for consistency, but there are lots of examples in the published style guides that do not include one.

adunning:
@bwiernik I expect that you want to think through the archival citation problem carefully as it has implications for APA, but for Chicago this is merely reverting to the pre-July behaviour. The annotated bibliography styles are currently broken and I have included a fix here, if you are able to review it.

POBrien333:
Shall we merge this? The PR looks pretty mature by now and it'd be a nice xmas gift for authors. :)

adunning:
Yes, this is ready to go as far as I am concerned.

bwiernik:
Will get all of these reviewed and merged tomorrow or Jan 2. Thanks for your patience @adunning !

adunning:
Wonderful; thank you! I should be able to fix any mistakes fairly quickly up to Saturday; after that, it may be awhile before I can make revisions.

PR #7815, opened 2025-09-25
Update istanbul-bilgi-universitesi-lisansustu-programlar-enstitusu-hu…

korkmazaharun:
…kuk.csl

I've added some lines to fix certain errors I discovered afterwards.

These were:
1- The volume should've been displayed as "C:" instead of "c.:" 
2- The translator should've been displayed as "Çev.:" instead of "çev.:" in the subsequent citations. 
3- The short titles should've been displayed in the bibliography entries; currently, they aren't displayed.

### DIFF
--- a/istanbul-bilgi-universitesi-lisansustu-programlar-enstitusu-hukuk.csl
+++ b/istanbul-bilgi-universitesi-lisansustu-programlar-enstitusu-hukuk.csl
@@ -20,7 +20,6 @@
       <term name="editor" form="short">edt.</term>
       <term name="accessed">e.t.</term>
       <term name="issue" form="short">S</term>
-      <term name="volume" form="short">C</term>
       <term name="edition" form="long">bası</term>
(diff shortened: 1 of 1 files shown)


POBrien333:
Hi
Thank you for the PR.
I removed those locales in your original PR as they're already included in the tr-TR translation. If you need the capitalisation, please set it on where you call element via text-case=capitalize-first.

korkmazaharun:
Thank you for the comment.

You are right, but the short form of "volume" in the Turkish tr locale is "c."
If capitalized, it would be "C.", which is different from this style's requirement; it should be "C". But "çev." can be solved via the capitalization feature, you are right. 

Fixing it now.

POBrien333:
You can use strip-period=true then.

korkmazaharun:
> You can use strip-period=true then.

Done

POBrien333:
Thanks

PR #7822, opened 2025-09-28
Update ausonius-editions.csl

jgravier001:
After discussion with the Publisher, it appears that the publication guidelines require directly including the DOI or URL link for online references, without the [online] mention or the access date information.

In this update:
- We modified the update information for the steel sheets (l. 44).
- We deleted the macro for the [online] and [accessed] mentions, keeping only the DOI/URL information (l. 295).
- We adjusted the corresponding macro call accordingly (l. 342).

### DIFF
--- a/ausonius-editions.csl
+++ b/ausonius-editions.csl
@@ -41,7 +41,7 @@
     <summary>Style CSL "Ausonius éditions" revu et mis à jour en collaboration avec Ausonius Éditions (d'après le guide de l'auteur du 15 juin 2022), à partir du travail réalisé au départ par N. Monteix, S. Thomas, et les contributions postérieures. - CSL style  "Ausonius éditions" updated, with the collaboration of the Ausonius Editions team (according to the "Guide de l'auteur - 15 juin 2022"), and based on the first work realised by N. Monteix, S. Thomas, and the next contributions.</summary>
     <!-- Informations -->
     <published>2023-01-06T05:43:05+00:00</published>
-    <updated>2024-03-02T02:24:11+00:00</updated>
+    <updated>2025-09-28T09:37:19.544</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <!-- PARAMETRAGE LOCALE\TERMS -->
@@ -296,20 +296,7 @@
     <choose>
       <if variable="accessed" match="any">
         <choose>
-          <if variable="DOI URL" match="any">
-            <text term="online" form="short"/>
-            <text term="accessed" prefix=", "/>
-            <date variable="accessed" form="text" prefix=" "/>
-          </if>
-        </choose>
-      </if>
-    </choose>
-  </macro>
-  <macro name="bloc_en-ligne_lien">
-    <choose>
-      <if variable="accessed" match="any">
-        <choose>
-          <if variable="DOI">
+            <if variable="DOI">
             <text variable="DOI" prefix="DOI&#160;: "/>
           </if>
           <else>
@@ -351,9 +338,8 @@
         <text macro="bloc_auteur"/>
         <text macro="bloc_date-publication"/>
         <text macro="bloc_titre-ouvrage-contribution-article" prefix="&#160;: "/>
-        <text macro="bloc_en-ligne" prefix=". "/>
       </group>
-      <text macro="bloc_en-ligne_lien" prefix=" "/>
+      <text macro="bloc_en-ligne" prefix=" "/>
     </layout>
   </bibliography>
 </style>


jgravier001 (commented):
After discussion with the Publisher, it appears that the publication guidelines require directly including the DOI or URL link for online references, without the [online] mention or the access date information.

In this update:

We modified the update information for the steel sheets (l. 44).

We deleted the macro for the [online] and [accessed] mentions, keeping only the DOI/URL information (l. 295).

We adjusted the corresponding macro call accordingly (l. 342).

ausonius-editions.csl
POBrien333: Please delete this line.

POBrien333: I assume this link is a remnant from the style used as a template?

POBrien333, line 8: Can you link to the actual style guidelines?
If not available, could you upload it to this PR? (in the opening post)

POBrien333: Please use a label instead of hardcoding the "no"

POBrien333: Better to put these two into a group, setting the prefix=" " as a delimiter instead.

POBrien333: Use `<text term="in"/> instead of hardcoding this. 
Important, as your style is used in different languages.

POBrien333: this whole conditional would never be used since it is impossible to test for "true" as it's inside the conditional testing for chapter paper-conference.

POBrien333: Since you want this in two places, make it into a macro and just call the macro.

POBrien333: Why are you testing for these 2? This seems very unlikely and untypical.

POBrien333: Put both into a group and set the " " as a delimiter instead.

POBrien333:
Thank you for the update.
I've left a whole bunch of comments.
In general, it would be better to use more groups with delimiters and less affixes to improve the robustness of the style. :)

ausonius-editions.csl
POBrien333: "en ligne" is already the standard translation in fr-FR. We'd prefer if you set the capitalisation and the brackets where you call the term instead. So, please remove from here.

POBrien333: Already included in the locale and duplicates things: https://github.com/citation-style-language/locales/blob/master/locales-fr-FR.xml

POBrien333: Please check against each locale for duplicated terms: https://github.com/citation-style-language/locales/tree/master

jgravier001: Thank you for your comment.
The author guidelines are currently being updated and will soon be available at the URL mentioned in the CSL metadata.
For the moment, the adjustments were made at the request of the publisher, based on our email exchanges with the editorial team.

jgravier001: Thank you for your comment.
The link refers to documentation used by the publisher, particularly for journal abbreviations.
To be more precise, we can replace it with the direct link to the official list of abbreviations: "https://about.brepolis.net/aph-abreviations/"

jgravier001: Thank you. After checking, we have cleaned up the local terms and now rely on the standard locale terms whenever they match the publisher’s requirements.

jgravier001: Thank you for the suggestion. We’ll replace the hardcoded “n°” with the appropriate label element to ensure proper localization.

jgravier001: Thank you for the suggestion. We’ll group the two variables together with a delimiter to handle spacing more cleanly.

jgravier001: Thank you for your comment.
In this case, in is intentionally kept in Latin, as required by the publisher’s guidelines, and therefore does not vary according to language.

POBrien333: Ok. Let us know once these guidelines are online and then I can review the style once more.

ausonius-editions.csl
```diff
@@ -93,8 +93,14 @@
       <name and="text" name-as-sort-order="all" sort-separator=", " initialize-with=". " delimiter-precedes-last="never"/>
       <label prefix=", "/>
       <substitute>
-        <names variable="editorial-director"/>
-        <names variable="editor"/>
+        <names variable="editorial-director">
```
POBrien333: All of this is not necessary. The label from the main "names" will be used for the subsitute name variables. Please revert these changes.

ausonius-editions.csl
```diff
@@ -162,8 +174,12 @@
     <choose>
       <if type="report thesis" match="any">
         <text variable="genre" prefix=", "/>
-        <label variable="number" form="short" prefix=" "/>
-        <number variable="number"/>
+        <choose>
+          <if variable="number" match="any">
```
POBrien333: This conditional is superfluous.

PR #7824, opened 2025-09-29
Create circulation 10 et al.csl

mcj1:
Circulation style with the new et al required "list the first 10 authors only followed by "et al" for all references"


POBrien333:
As we already have that style on the repository, please "edit" the style, not add a new style to the repository. Please update the documentation link also.

 If you can, please do this inside this PR.

POBrien333:
continued in #7833

PR #7826, opened 2025-09-30

POBrien333:
Fixes https://github.com/citation-style-language/styles/issues/7112#issuecomment-3349333158


POBrien333:
@libora 
I've replaced a bunch of hardcoded terms, specifically the no-publisher/no-place terms. Could you please add translations for these in a PR in the cz-CZ locale? If you spot any other translations that are missing, it'd be great also, if you could add those.
https://github.com/citation-style-language/locales/blob/master/locales-cs-CZ.xml

PR #7828, opened 2025-09-30
Update eu-interinstitutional-style-2nd-edition.csl

LuisMach:
Revisions included in this 2nd edition:

rev22c-20250611 
-Included the bibliographic element original author in the elements to display when present in records from social media post document type (Zotero item types Forum Post and Blog Post) to include the name of the author of the original post in case of a repost reference. 
 
rev22a-20250527 
-Added a specific tag (annote:) to use in the Zotero 'Extra' field for the author social media identification. The added rules allow this identification to be the displayed in the list of the references, but exclude it from in-text citation. 

rev21i-20250526 
-Update rules for report document type to include series title.  -Added rules for additional document types: 
*Conference-paper not in event proceedings.  
 *Conference-paper published in conference proceedings (strictly speaking this type should be included in the 'Part of a work' type, but in Zotero it's preferable to use  the 'Conference Paper' item rather than the 'Book Section' item).  

rev20c-20250414  
-Update the reference and in-text citation for legal documents in a more standard way for legislation. Namely using the short title instead of author's name for in-text citation and include the container-title (the Code) in the reference. 

rev19c-20250326  
-Change the location of the optional internal reference number ('Call Number' Zotero field) to a more standard position before the hyperlink.  -Add the internal reference number ('call number' as an 'Extra' field) to the web page item type after the title, following the ISG example Reference to a web page. 
-Update the rules for documents that are part of a collection ('Series' Zotero field) to include the title of the collection into the reference.

### DIFF
--- a/eu-interinstitutional-style-1st-edition.csl
+++ b/eu-interinstitutional-style-1st-edition.csl
@@ -1,20 +1,19 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" and="text" delimiter-precedes-et-al="never" delimiter-precedes-last="never" initialize-with=". " page-range-format="expanded" demote-non-dropping-particle="never" default-locale="en-GB">
-  <!-- This style was edited with the Visual CSL Editor (https://editor.citationstyles.org/visualEditor/) -->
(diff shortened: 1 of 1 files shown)


LuisMach:
it was necessary to fix an error due to the renaming of a macro

POBrien333:
Superseded by #7829

LuisMach:
Thanks @POBrien333

PR #7829, opened 2025-09-30
Update eu-interinstitutional-style.csl

LuisMach:
rev22c-20250611 
Included the bibliographic element original author in the elements to display when present in records from social media post document type (Zotero item types Forum Post and Blog Post) to include the name of the author of the original post in case of a repost reference. 

rev22a-20250527 
Added a specific tag (annote:) to use in the Zotero 'Extra' field for the author social media identification. The added rules allow this identification to be the displayed in the list of the references, but exclude it from in-text citation. 

rev21i-20250526 
Update rules for report document type to include series title.  Added rules for additional document types: Conference-paper not in event proceedings and Conference-paper published in conference proceedings (strictly speaking this types should be included in the 'Part of a work' type, but in Zotero it's preferable to use  the 'Conference Paper' item rather than the 'Book Section' item).  

rev20c-20250414  
Update the reference and in-text citation for legal documents in a more standard way for legislation. Namely using the short title instead of author's name for in-text citation and include the container-title (the Code) in the reference. 

rev19c-20250326  
Change the location of the optional internal reference number ('Call Number' Zotero field) to a more standard position before the hyperlink.  Add the internal reference number ('call number' as an 'Extra' field) to the web page item type after the title, following the ISG example Reference to a web page.  Update the rules for documents that are part of a collection ('Series' Zotero field) to include the title of the collection into the reference.

### DIFF
--- a/eu-interinstitutional-style.csl
+++ b/eu-interinstitutional-style.csl
@@ -9,6 +9,9 @@
       <name>Luis Machado</name>
       <email>luis.oliveira-machado@ec.europa.eu</email>
     </author>
+    <contributor>
+      <name>Patrick O'Brien</name>
+    </contributor>
     <category citation-format="author-date"/>
     <category field="generic-base"/>
     <summary>European Union Interinstitutional style in author-data format, Eng 2nd Edition</summary>


LuisMach:
I understand the style ID error, I shouldn't change it, only the title (unfortunately I wasn't aware that I shouldn't include the edition number in the ID, is there any way to solve this or will it always have to stay as the 1st edition?)

But the other errors I don't understand, since those settings were the same in the previous file.

POBrien333:
Thanks for the submission. Please keep further edits in one PR. You can simply edit the code in this PR. If you don't know, please ask.

LuisMach:
I'll do, sorry for any inconvenience, beginner's mistake.

POBrien333:
@LuisMach 
Can you confirm that the 1st and 2nd edition thing is/was just for internal purposes and is not an official version like for say APA 7the edition?

LuisMach:
@POBrien333 
Actually, it's not just for internal control. The second edition has a difference in the number of authors presented in the in-text citations. I have to confess that I still don't fully understand how this system works. I would be grateful for any clarification if I'm not following all the procedures correctly.

POBrien333:
I think we're talking past each other.
On the page with the [guidelines](https://style-guide.europa.eu/en/home), nowhere is the word "edition" mentioned.
So, this style should not have 1st/2nd edition in its style name and filename. I think this was an oversight when this style was first accepted. 

@bwiernik @adam3smith 
Can we rename styles post hoc on the repository?

LuisMach:
@POBrien333 
You're right, we only have the mention that [the rules were "revised."](https://style-guide.europa.eu/en/content/-/isg/topic?identifier=5.9.4-bibliographic-references) What do you suggest we do?

adam3smith:
@POBrien333 we can change the title all we want. It's possible to change filenames, but we try to avoid that as much as possible

POBrien333:
Thank you.
I think in this case the filename should be `eu-interinstitutional-style.csl`, removing any hint to a version. 
What happens if a style gets their filename changed? Endusers get an error?

LuisMach:
@POBrien333 
Sorry to step in, but the language (Eng) and the format (author-date) should also be in the name because we will develop a CSL for other languages and in note system.

POBrien333:
This is about the filename only.

LuisMach:
@POBrien333 OK, but what should be the name of the file with the style in the note system? They can have the same name?

eu-interinstitutional-style.csl
```diff
@@ -246,8 +256,7 @@
         <group delimiter=", ">
           <group delimiter=" ">
             <text variable="genre"/>
-            <label variable="number" form="short" strip-periods="true"/>
-            <text variable="number"/>
+            <text variable="number" prefix="No "/>
```
POBrien333: Please use the correct label to achieve this output. We try not to hardcode text into affixes etc.

POBrien333: Please check the whole style for such instances, as there are a few.

eu-interinstitutional-style.csl
```diff
@@ -277,50 +286,65 @@
...
-    <names variable="translator contributor recipient" delimiter=", ">
+    <names variable="translator" delimiter=", ">
       <name delimiter-precedes-et-al="never" delimiter-precedes-last="never" initialize-with=". "/>
       <label form="short" text-case="lowercase" prefix=" (" suffix=")"/>
     </names>
   </macro>
-  <macro name="original-author">
+  <macro name="in">
+    <choose>
+      <if type="chapter" match="any">
```
POBrien333: you can just make this `if type="chapter paper conference" match="any"`. No need for two tests.

eu-interinstitutional-style.csl
```diff
@@ -332,89 +356,162 @@
...
-      <text term="in" text-case="lowercase"/>
-      <group delimiter=", ">
-        <names variable="editor" delimiter=", ">
-          <name delimiter-precedes-et-al="never" delimiter-precedes-last="never" initialize-with=". "/>
-          <label form="short" text-case="lowercase" prefix=" (" suffix=")"/>
-        </names>
+  <macro name="container-title">
+    <choose>
+      <if type="webpage" match="any">
+        <text variable="container-title" strip-periods="false" suffix=" website"/>
```
POBrien333: please use a term instead of adding it as a suffix

eu-interinstitutional-style.csl
```diff
@@ -332,89 +356,162 @@
...
+      <else>
+        <text variable="container-title" strip-periods="false" font-style="italic"/>
+      </else>
+    </choose>
+  </macro>
+  <macro name="extra-IDnum">
+    <choose>
+      <if type="webpage" match="none">
+        <choose>
+          <if match="any" variable="call-number">
```
POBrien333: this conditional is not necessary since it would only output it if it exists in the first place.

eu-interinstitutional-style.csl
LuisMach: I recognise that there are many similar cases but, unless I've missed it, the cases where I've added text are because the output generated doesn't match the style format. In some cases the difference is minimal (as in this case, between No and No.), I know. However, this is the formatting that is agreed between the various EU institutions. Believe me, it's not a personal choice, if it were, I would have simplified many of the particularities of the style.

eu-interinstitutional-style.csl
LuisMach: Thank you very much for pointing this out. When I corrected it, I realized that it makes more sense to use the editor's presence (or similar) as the criterion for applying the rule.

eu-interinstitutional-style.csl
LuisMach: This is one of the cases I mentioned above. The available term is 'webpage' and the style uses 'website'.

eu-interinstitutional-style.csl
LuisMach: The reason for excluding the document type webpage is for that type the same variable is use in a different location of the reference.

eu-interinstitutional-style.csl
POBrien333: Sure, but then you redefine that term in the locale, just after the info section to achieve "webpage"

eu-interinstitutional-style.csl
```diff
@@ -221,22 +235,18 @@
...
-          <text variable="number"/>
-        </group>
-      </else-if>
     </choose>
   </macro>
   <macro name="citation-locater">
-    <group delimiter=" ">
-      <label text-case="lowercase" variable="locator" form="short"/>
+    <group>
+      <label text-case="lowercase" suffix=" " variable="locator" form="short"/>
```
POBrien333: please set this sufix on the group, so make line 248 into `<group delimiter=" ">`

eu-interinstitutional-style.csl
POBrien333: I understand that CSL is all new to you.
Generally, we don't want to see any text in an affix or using "value" in a style.
In this case, using the `label variable="number" form="short"` would give "No." sure, but then you set `strip-periods="true"` and it'll remove the dot. 
See https://docs.citationstyles.org/en/stable/specification.html#strip-periods

eu-interinstitutional-style.csl
```diff
@@ -121,26 +137,24 @@
...
-          <group delimiter=" ">
-            <label variable="issue" text-case="capitalize-first"/>
-            <text variable="issue"/>
-          </group>
-        </group>
+      <if match="none" variable="collection-title">
+        <choose>
+          <if type="article-journal article-magazine article" match="any">
+            <group delimiter=", ">
+              <text variable="volume" strip-periods="false" prefix="Vol. "/>
```
POBrien333: please use a label

eu-interinstitutional-style.csl
```diff
@@ -121,26 +137,24 @@
...
+              <text variable="issue" prefix="Issue "/>
```
POBrien333: https://docs.citationstyles.org/en/stable/specification.html#strip-periods

eu-interinstitutional-style.csl
POBrien333: Yeah, I'm not talking about the test for "webpage". My comments usually are specific to the line the comment is on.
Here you're testing for call-number to output call-number. This test is superfluous, because it can only ever output call-number if it actually exists in the metadata. It's redundant.

eu-interinstitutional-style.csl
```diff
@@ -332,89 +356,162 @@
...
+    </choose>
   </macro>
   <macro name="preprint">
     <choose>
-      <if type="article" match="all" variable="genre">
-        <text variable="genre"/>
+      <if type="article" match="all" variable="genre" is-numeric="number">
+        <group delimiter=" ">
+          <text variable="genre"/>
+          <text variable="number" prefix="No "/>
```
POBrien333: use label

POBrien333:
@LuisMach 
As there was still some hardcoded text, I simplified the style and work on these issues. Please check everything. If you spot anything, let me know and I'll handle it.
(note, in case you're testing this in the CodeEditor/VisualEditor: it doesn't print out "No " correctly as it's out of date. Please use the Style editor in Zotero.)

Also, the test for the number being a number is usually not really an issue. Or do you have grounds for that?

LuisMach:
@POBrien333 
Thank you so much for your work and the warning. I'll check more carefully the style, but I already spot an issue looking the sample provided by the boot. In journal articles the style uses Issue and not Number. 
The reason for testing if the value is a number is because we don't want to display the term Issue in cases of, eg, the value is April. Also, in reports we don't want, eg, No Com(2022)56.

POBrien333:
Yeah, the bot here is based on citeproc-ruby which we are looking to replace soon and will also not have the correct output.

**Zotero's output is correct:**
Campbell, J. L. and Pedersen, O. K. (2007), ‘The Varieties of Capitalism and Hybrid Success’, Comparative Political Studies, Vol. 40, Issue 3, 1 March, pp. 303–323, MSGG-2024-003, https://doi.org/10.1088/1748-0221/7/03/P03012.

LuisMach:
@POBrien333 
I've tested the file in Zotero, and detected three things that need to be corrected (maybe some were due to my early changes).

1. The collection number is displayed like this: No., 5
And should be: No 5

2. In web-resources types, like webpage, post, or dataset, with other date-parts than year, these are being displayed twice.

3. Book-chapters and Conference-papers with container-title, but no editors, the 'in:' should be displayed before the container-title.

For this last issue, my workaround was to create this macro:
```
<macro name="in-no-editors">
    <choose>
      <if match="any" variable="author">
        <choose>
          <if match="none" variable="editor container-author collection-editor">
            <text term="in" suffix=": "/>
          </if>
        </choose>
      </if>
    </choose>
  </macro>
```
And add it in two places:
```
406       <else>
            <group delimiter=" ">
              <text macro="in-no-editors"/>
            </group>
408       </else>

410   </else-if>
      <else-if type="chapter" match="any">
        <group delimiter=" ">
          <text macro="in-no-editors"/>
          <text variable="container-title" strip-periods="false" font-style="italic"/>
        </group>
      </else-if>
411   <else>
```
Do you still prefer to make the code changes yourself, or do you want me to do it?
Thanks again.

POBrien333:
Have a look at the changes I made. We should be pretty much there now.

LuisMach:
@POBrien333 
Yes, it is almost there, but I spotted 3 more problems (😟 sorry)

1. The label for editors should be without the dot, like this: (eds)

2a. In journal-articles, the issued number is still displayed with the label 'Number' for non-number values, and shouldn't have any label. 
2b. And, for number values, now it doesn't display anything and should have the label 'Issue'

3. A conference-paper without container-title shouldn't have the 'in:'

LuisMach:
@POBrien333 
I'm terribly sorry, but one thing is still not right.
In journal-articles, the issued number, for number values, is now displayed with the label 'Number', and should have the label 'Issue'. 
Sorry

POBrien333:
image
image


I'm not sure how you are testing this, but this works as designed.

LuisMach:
@POBrien333 I copied the code to a CSL file and imported to Zotero and I got this output:
image

POBrien333:
Not sure why that would happen.
Can you test this out in a Word document?

LuisMach:
@POBrien333, it gave me the same result
image
That was the reason, in my version, I used a _term_ and not a _label_. And even with the _term_ I needed to define locally the intended output, otherwise I would get **Number** and not **Issue.**

POBrien333:
Again, works for me:
image

You have the latest version of Zotero?
You have nothing odd in the extra field? Can you share a screenshot of how this is entered in Zotero?

LuisMach:
I'm using the version 7.0.15 (64-bit)
image

POBrien333:
I'd start with updating Zotero since the last version is 7.0.24 I think

LuisMach:
@POBrien333 
Yes, it seems that is a version issue. I tried in my personal laptop with the current version (7.0.27) and it works as it should. (I'm working with my corporate laptop, and I've to request the new version from the IT department, the installation rights are limited.)

POBrien333:
Thanks. I've merged this now. If you spot any further issues you can of course open a new PR.

PR #7838, opened 2025-10-08
Update and rename journal-of-analytical-toxicology.csl to dependent/j…

POBrien333:
…ournal-of-analytical-toxicology.csl

https://forums.zotero.org/discussion/127254/style-error-journal-of-analytical-toxicology


POBrien333:
Hi @adam3smith 
Is this correct way of handling this situation where an independent style seizes to exist?

adam3smith:
So the independence style becomes a depend style here, yes? That's perfect, then. You do have to make some judgment calls about the template Styles as you've done here

POBrien333:
Great. Thank you

PR #4627, opened 2020-03-12
Create unified-style-sheet-for-linguistics-de-gruyter-literature.csl

fussballlinguist:
This is a modified version of the Unified Stylesheet for Linguistics that meets the requirements for German linguistics as communicated by the publisher de Gruyter. Most important changes are: 
- Publication date will appear in brackets (2019)
- The abbreviation for editors is "Hrsg."
- Access informations for websites and blog posts are in German

PR #7841, opened 2025-10-09
Add new CSL style: Linguistics Vanguard

Yihtsy:
This commit adds a new CSL style for the journal *Linguistics Vanguard*.

- Based on the *Journal of Historical Linguistics* template.  
- Adapted according to the official author guidelines of the journal.  
- Validated using the CSL validator (https://validator.citationstyles.org/).  

Please let me know if any further adjustments are needed.

### DIFF
--- a/linguistics-vanguard.csl
+++ b/linguistics-vanguard.csl
@@ -7,28 +7,31 @@
     <link href="http://www.zotero.org/styles/linguistics-vanguard" rel="self"/>
     <link href="http://www.zotero.org/styles/journal-of-historical-linguistics" rel="template"/>
     <link href="https://www.degruyterbrill.com/journal/key/lingvan/html#submit" rel="documentation"/>
+    <link href="file:///D:/Installs/LINGVAN_LINGVAN_Mouton_journal_stylesheet-1.pdf" rel="documentation"/>
     <author>
       <name>Tsy Yih</name>
       <email>yihtsy@outlook.com</email>
     </author>
+    <contributor>
+      <name>Patrick O'Brien</name>
+    </contributor>
     <category citation-format="author-date"/>
     <category field="linguistics"/>
     <eissn>2199-174X</eissn>
-    <updated>2025-09-17T12:00:00+00:00</updated>
+    <updated>2025-10-10T09:05:28+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="editor-translator">
-    <names variable="editor translator" prefix="(" suffix=")" delimiter=", ">
+    <names variable="editor translator" delimiter=", " prefix="(" suffix=")">
       <name and="text" initialize-with="" delimiter=", "/>
-      <et-al font-style="normal"/>
+      <et-al font-style="italic"/>
       <label form="short" prefix=", " text-case="capitalize-first"/>
     </names>
   </macro>
   <macro name="author">
     <names variable="author">
       <name delimiter-precedes-last="never" initialize="false" initialize-with="." name-as-sort-order="first" and="symbol"/>
-      <et-al font-style="normal"/>
-      <label form="short" prefix=" (" suffix=")" text-case="capitalize-first"/>
+      <label form="short" prefix=" (" suffix=")"/>
       <substitute>
         <names variable="editor"/>
         <names variable="translator"/>
@@ -39,7 +42,6 @@
   <macro name="author-short">
     <names variable="author">
       <name form="short"/>
-      <et-al font-style="normal"/>
       <substitute>
         <names variable="editor"/>
         <names variable="translator"/>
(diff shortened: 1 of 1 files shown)


POBrien333:
Thank you for the PR.
The guidelines mentioned "The format for reference entries should follow the “Unified style sheet for linguistics”" which we already have on the repository:
https://www.zotero.org/styles?q=Unified%20style%20sheet

Is this style you're proposing the same?

Yihtsy:
> Thank you for the PR. The guidelines mentioned "The format for reference entries should follow the “Unified style sheet for linguistics”" which we already have on the repository: https://www.zotero.org/styles?q=Unified%20style%20sheet
> 
> Is this style you're proposing the same?

@POBrien333  Thank you for pointing that out!
To be frank, I did only notice the “Unified Stylesheet for Linguistics (de Gruyter Literature)” before (which differs quite substantially) and hadn’t realized there was also a “Unified style sheet for linguistics.”

The latter does share some similarities with the style I proposed, but after a careful comparison I found a few key differences. For instance, Linguistics Vanguard uses “and” (not “&”) for two authors in in-text citations, applies et al. from three authors onward, and does not include DOI numbers at the end of journal references.

So, in short, the LingVan style I provided is indeed different from the “Unified style sheet for linguistics.”

linguistics-vanguard.csl
```diff
@@ -0,0 +1,149 @@
...
+    </author>
+    <category citation-format="author-date"/>
+    <category field="linguistics"/>
+    <eissn>2199-174X</eissn>
+    <updated>2025-09-17T12:00:00+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="en">
+    <terms>
+      <term name="by">ed. by</term>
```
POBrien333: Instead of using the short form and redefining the label here, you can just use the "verb-short" form which is exactly what you want: https://github.com/citation-style-language/locales/blob/67d6db54bb47c481d496bc3eab9ec0f477d32653/locales-en-GB.xml#L802

linguistics-vanguard.csl
```diff
@@ -0,0 +1,149 @@
...
+    <names variable="editor translator" prefix="(" suffix=")" delimiter=", ">
+      <name and="text" initialize-with="" delimiter=", "/>
+      <et-al font-style="italic"/>
+      <label form="short" prefix=", " text-case="capitalize-first"/>
+    </names>
+  </macro>
+  <macro name="author">
+    <names variable="author">
+      <name delimiter-precedes-last="never" initialize="false" initialize-with="." name-as-sort-order="first" and="symbol"/>
+      <et-al font-style="normal"/>
```
POBrien333: sometimes you have et al. in italics, sometimes not. Is that on purpose?
If you indeed want no styling here in the author macro, you can just delete this line.

Yihtsy: Thank you for pointing this out. You're absolutely right — the inconsistency comes from line 28, where <et-al font-style="italic"/> was mistakenly used. It should have been font-style="normal" to match the rest of the macros.

linguistics-vanguard.csl
Yihtsy: Thank you for pointing that out. I've deleted the whole <locale> block and tested on my own ref lists. It seems to be right. Please let me know if there's still other problems.

POBrien333:
Hi,
Thank you for checking these and for your explanation.
I'm a bit unsure about the "de Gruyter" style we already have.
When we accepted that PR https://github.com/citation-style-language/styles/pull/4627, we didn't have the guidelines at hand and I don't know which it could/would be. It simply could be that that style is actually the same you're proposing, but obviously with errors as maybe the style changed? Hard to say. 

I would be more inclined to call this style something like "Unified Stylesheet for Linguistics (de Gruyter Mouton)" and then making a dependnet style for this specific journal.

@adam3smith and @bwiernik 
Any input here?

Yihtsy:
> Hi, Thank you for checking these and for your explanation. I'm a bit unsure about the "de Gruyter" style we already have. When we accepted that PR #4627, we didn't have the guidelines at hand and I don't know which it could/would be. It simply could be that that style is actually the same you're proposing, but obviously with errors as maybe the style changed? Hard to say.
> 
> I would be more inclined to call this style something like "Unified Stylesheet for Linguistics (de Gruyter Mouton)" and then making a dependnet style for this specific journal.
> 
> @adam3smith and @bwiernik Any input here?

@POBrien333 Thank you for your comment and for clarifying this.
I think your assumption might be correct. My initial reason for creating the Linguistics Vanguard style was that I am currently preparing a paper for submission to that journal. My first attempt was actually to use the Unified Stylesheet for Linguistics (de Gruyter Literature), but I found that its format differs quite substantially from the current version required by Linguistics Vanguard.

So I started from the Journal of Historical Linguistics style, which was relatively close in format, and modified it accordingly. However, JHL is published by John Benjamins, not de Gruyter. In fact, the Unified Stylesheet for Linguistics you mentioned is indeed quite similar to the Linguistics Vanguard guidelines (except for some minor details), but I hadn’t noticed that style when I began—otherwise I might have used it as the base instead.

As far as I know, the Unified Stylesheet for Linguistics was proposed by the Linguistic Society of America. The original link seems to be no longer available, but an archived version can still be found here: https://langsci-press.org/unifiedstylesheet

Interestingly, I found that several de Gruyter journals explicitly state in their author guidelines that “The format for reference entries should follow the Unified Stylesheet for Linguistics.”

Would it therefore make sense to change the template of this style from JHL to the Unified Stylesheet for Linguistics?
(That said, Linguistics Vanguard still differs from that template in several details, so they are not entirely identical. I haven’t checked other de Gruyter journals yet.)

PR #7844, opened 2025-10-11
Update associacao-brasileira-de-normas-tecnicas-ufrgs-initials.csl

therloux:
Update according to main style (associacao-brasileira-de-normas-tecnicas-ufrgs.csl).

### DIFF
--- a/associacao-brasileira-de-normas-tecnicas-ufrgs-note-initials-with-ibid.csl
+++ b/associacao-brasileira-de-normas-tecnicas-ufrgs-note-initials-with-ibid.csl
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- XML Declaration: não é preciso alterar -->
-<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" class="note" demote-non-dropping-particle="never" page-range-format="expanded" default-locale="pt-BR">
+<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" class="note" demote-non-dropping-particle="never" default-locale="pt-BR">
   <info>
     <title>Universidade Federal do Rio Grande do Sul - ABNT (autoria abreviada, nota, com Ibid.) (Português - Brasil)</title>
     <title-short>UFRGS-ABNT</title-short>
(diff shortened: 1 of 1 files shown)


POBrien333:
@therloux 
Thank you for your PR. Can you explain what this PR is?

therloux:
The ABNT-UFRGS style has multiple variants, but these haven't always been kept up-to-date with the main style. This PR intends to update the initials variant as it hasn't been changed in two years.

The file exactly like the main style file (associacao-brasileira-de-normas-tecnicas-ufrgs.csl), except for the code for initials in names, which I've kept.

POBrien333:
Great. Thanks!
I made these changes recently as this style is used by a lot of people as a base and had a few issues that I wanted to remove. Great effort. Can you also make these changes for "associacao-brasileira-de-normas-tecnicas-ufrgs-note-initials-with-ibid.csl"?

therloux:
@POBrien333 I was just looking at the sample citations in https://github.com/citation-style-language/styles/pull/7844#issuecomment-3392671988 and I've just now realised some of the terms used do not conform with the ABNT style.

For instance, volume should be abbreviated to 'v.' instead of 'vol.' and all abbreviations should be singular (so no 'orgs.'). I've noticed removing the custom `term` tags in the file caused this. Because those are style-specific terms (they are not related to localisation), I think the tags should be kept.

Problem is, the tags have also been removed from the other styles (e.g. associacao-brasileira-de-normas-tecnicas-ufrgs.csl). Because this affects other files, I believe this should be discussed before merging this pull request or committing any files.

POBrien333:
Can you make these changes in the "normal" style as part of this PR?
Please use the `plural"=never"` where necessary.

The "v." should already be included as the pt-BR locale does this already: https://github.com/citation-style-language/locales/blob/master/locales-pt-BR.xml
That's why I took it out.
See the actual output in Zotero:
image

How are you testing this?

therloux:
I didn't find any issues in Zotero either, but I did in https://github.com/citation-style-language/styles/pull/7844#issuecomment-3392671988 and in https://editor.citationstyles.org.

At any rate, I would assume this is ABNT-specific terminology, much like the Latin words adopted by the standard (e.g. 'in' and 's.l.'). Should we still depend on the locale file? Isn't it style-agnostic?

POBrien333:
The Code/Visual Editor is out of date and the Ruby output pipeline also has errors. Zotero is the ground truth and what users actually use. Therefore, nothing to worry and the style can be left as is. (Minus actual errors, of course).
Note that you can avoid plural the way I already pointed out and dots can be removed with strip-peeiods. This is the preferred way.

therloux:
I've left out the unnecessary terms and updated associacao-brasileira-de-normas-tecnicas-ufrgs-note-initials-with-ibid.csl, as requested.

POBrien333:
Thanks a lot for your help with this. :)

PR #7850, opened 2025-10-16
Create institut-agro-dijon-2025.csl

mideus21:
This is the Zotero style used by the library of the Institut Agro Dijon.


POBrien333:
Thank you for your PR.
Can you go over points 8+10 in our requirements document and add this info into the style?
https://github.com/citation-style-language/styles/blob/master/STYLE_REQUIREMENTS.md

PR #7856, opened 2025-10-20
Update population.csl

nalimilan:
Remove quotes around titles and improve author and translator fields to better reflect the journal's style.

### DIFF
--- a/population.csl
+++ b/population.csl
@@ -10,9 +10,6 @@
       <name>Milan Bouchet-Valat</name>
       <email>nalimilan@club.fr</email>
     </author>
-    <contributor>
-      <name>Patrick O'Brien</name>
-    </contributor>
     <category citation-format="author-date"/>
     <category field="social_science"/>
     <category field="sociology"/>
@@ -22,6 +19,14 @@
     <updated>2025-10-20T15:13:29+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
+  <locale xml:lang="fr">
+    <terms>
+      <term name="editor" form="short">
+        <single>dir.</single>
+        <multiple>dir.</multiple>
+      </term>
+    </terms>
+  </locale>
   <macro name="editor">
     <names variable="editor" delimiter=" ">
       <name name-as-sort-order="all" sort-separator=" " delimiter=", " delimiter-precedes-last="always" initialize-with=". "/>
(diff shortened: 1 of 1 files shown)


POBrien333:
Thanks for the PR. Can you update the documentation link?

nalimilan:
Good catch. Fixed.

POBrien333:
@nalimilan 
I've just gone over the style and massively simplified it. Can you test it out? Then we can merge.

population.csl
```diff
@@ -36,8 +36,9 @@
       <name name-as-sort-order="all" sort-separator=" " delimiter=", " delimiter-precedes-last="always" initialize-with=". " font-variant="small-caps" font-weight="bold"/>
       <label form="short" text-case="lowercase" prefix=" (" suffix=")"/>
       <substitute>
-        <names variable="editorial-director"/>
         <names variable="editor"/>
+        <names variable="editorial-director"/>
+        <text variable="title"/>
```
nalimilan, line 47: I don't understand this. Is it standard practice? The journal's style doesn't mention that a reference could lack an author and an editor.

nalimilan:
Thanks. In general I'm not a fan of simplifying as it's hard to test all cases and it's enough for me to have a working style. I've found some issues:
- Missing comma before pages: Amossé T., Bouchet-Valat M., 2024, Dirigeants et professionnels de haut niveau : une définition statistique de l’« élite socioprofessionnelle », Population, 79(1)p. 41‑73. DOI: [10.3917/popu.2401.0041](https://doi.org/10.3917/popu.2401.0041)
- "éd." instead of "dir." (also for chapters): Aronson P., Gold S., 2007, Dual-Earner Couples Ritzer G. (éd.), Blackwell Encyclopedia of Sociology.
- Missing "in" before name of book: Becker G. S., 1991, Assortative Mating in Marriage Markets, A Treatise on the Family, Cambridge (Massachusetts), Harvard University Press.

population.csl
POBrien333: It's standard practise to have that as a fallback if a journal doesn't specifiy anything.

POBrien333:
I know actually compared the guidelines with actual typeset papers and there is many issues and the style doesn't match that at all. I should've checked that first.

nalimilan:
Indeed recent issues of the journal don't follow the guidelines for some reason. But for now I'm sticking to the guidelines since that's what authors need to use when submitting anyway. I'll update the style if guidelines change.

nalimilan:
Unfortunately there still are issues with examples in the previous comment.

POBrien333:
The preview pipeline here has issues. You'll need to test it out in Zotero...

nalimilan:
Ah OK. Thanks, looks good now.

PR #7865, opened 2025-10-25
Update dependent/bioethics.csl to bioethics.csl (new independent style)

POBrien333:
### Description
via https://forums.zotero.org/discussion/104163/style-request-bioethics-updated-with-full-reference-in-footnote-no-bibliography#latest

Creates a new independent style necessary for the journal.

### Checklist

### DIFF
--- a/bioethics.csl
+++ b/bioethics.csl
@@ -13,7 +13,7 @@
     <category field="anthropology"/>
     <issn>0269-9702</issn>
     <eissn>1467-8519</eissn>
-    <updated>2025-10-26T10:41:57+00:00</updated>
+    <updated>2025-10-26T10:46:24+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="cite-locator">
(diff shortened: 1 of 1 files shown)


adunning:
Note that titles should be in title case and initials should be spaced.

bioethics.csl
```diff
@@ -0,0 +1,166 @@
...
+        <names variable="editor"/>
+        <names variable="translator"/>
+        <text variable="title"/>
+      </substitute>
+    </names>
+  </macro>
+  <macro name="editor">
+    <names variable="editor">
+      <label form="short" plural="never" suffix=" "/>
+      <name initialize-with="."/>
```
adunning: Should be `<name initialize-with=". "/>`.

bioethics.csl
```diff
@@ -0,0 +1,166 @@
...
+      <text variable="publisher"/>
+      <date date-parts="year" form="text" variable="issued"/>
+    </group>
+  </macro>
+  <macro name="edition">
+    <choose>
+      <if is-numeric="edition">
+        <group delimiter=" ">
+          <number variable="edition" form="ordinal"/>
+          <text term="edition" form="short" strip-periods="false"/>
```
adunning: This should be a `<label>` rather than `<text>`.

bioethics.csl
```diff
@@ -0,0 +1,166 @@
...
+            <text macro="year-date"/>
+          </else-if>
+          <else-if type="webpage post post-weblog" match="any">
+            <text variable="container-title"/>
+          </else-if>
+          <else>
+            <text macro="title"/>
+            <group delimiter=": ">
+              <group delimiter=" ">
+                <text variable="container-title" font-style="italic" suffix=","/>
```
adunning: Remove `suffix=","` – there is only a space after the journal name.

I find this style interesting because it's essentially Chicago notes with initialized author names, reconfigured as a numeric style. If this is something that other journals do, it might be worth setting up a numeric Chicago variant?

POBrien333: Thanks for reviewing the style. :)

PR #7866, opened 2025-10-26
Dependent styles for Medieval Studies journals

adunning:
Dependent styles mostly for journals published by Brepols.


bwiernik:
How many journals does Brepol have? Should we add them to the /journals repo as well?

adunning:
@bwiernik Thank you for reminding me of that repository! The [Brepols website lists 55 journals](https://www.brepols.net/series?0=0&types=Journals); certainly larger than some of the other publishers on there. I'm separately working on tidying up CSL coverage of styles associated with Oxford University Press, and should probably look into using it.

PR #7871, opened 2025-10-29
Update deutsches-archaologisches-institut-romisch-germanische-kommiss…

POBrien333:
via https://forums.zotero.org/discussion/127730/problem-with-csl-substitute-does-not-work#latest


POBrien333:
@adam3smith 
This should work, but it doesn't work... Help me fix my brain block.

adam3smith:
Thanks!

PR #7878, opened 2025-11-04
New Hart's Rules, 2nd edition

### LINKED ISSUE #7105: New Hart's Rules
`new-harts-rules-the-oxford-style-guide.csl` remains a rough approximation of the target style, specified in chapter 18 of *New Hart’s Rules: The Oxford Style Guide*, 2nd ed. (Oxford, 2014). The current `csl` departs from specification in at least these details:

- Bibliography entries should be terminated with a full stop, as noted [here](https://forums.zotero.org/discussion/comment/338719/#Comment_338719).
- No em rules in bibliography entries: "In second and subsequent works by the same author do not replace the name with an em rule or rules: although this has been traditional print practice, it is not suitable for many types of electronic publication, so it is preferable to spell out the name in full for each entry" (18.1.3, p. 348).
- "When there are four or more authors, works in the humanities usually cite the first name followed by ‘and others’ or ‘et al.’" but "In some scientific journals [...] the policy can be to cite up to six authors before reducing the list to a single name and ‘et al.’ [or 'and others']" (18.2.2, p. 352). So, erring on the side of verbosity, set `et-al-min` to 7 (the current value) and `et-al-use-first` to 1. 
- DOIs should be supplied if available and prefixed *doi:* (18.8.5, p. 373).
- "It is not advisable to enclose a URL in angle brackets" (18.8.3, p. 372). Nor are access dates enclosed in square brackets (18.8.5, p. 373). For instance: "Hooper, Richard, ‘Lebanon’s Forgotten Space Programme’, *BBC News Magazine* (14 Nov. 2013), http://www.bbc.co.uk/news/magazine-24735423, accessed 14 Nov. 2013."
- "In books comprising the edited works of a number of authors, or a collection of documents, essays, congress reports, etc., the editor’s name appears first followed by *ed.* (standing for ‘editor’; plural *eds* or *eds.*) before the book title" (18.2.3, p. 353). This rule pertains also to a chapter within an edited collection. For instance: "Ashworth, A., ‘Belief, Intent, and Criminal Liability’, in J. Eekelaar and J. Bell, eds., *Oxford Essays in Jurisprudence*, 3rd ser. (1987), 6–25." And note the term *in* in such cases.
- Reviews also employ the term *in*, and the reviewed author precedes the title. For instance: "Ames-Lewis, F., review of Ronald Lightbown, *Mantegna* (Oxford, 1986), in *Renaissance Studies*, 1 (1987), 273–9." The specification says "the place of publication and date of the book reviewed are helpful but not mandatory," which is good, as I don't believe CSL can do that.
- "Editors of literary texts (or of another author’s papers) are cited after the title; in this case *ed.* (standing for ‘edited by’) remains unchanged even if there is more than one editor: Hume, David, *A Treatise of Human Nature*, ed. David Fate Norton and Mary J. Norton (Oxford, 2000)" (18.2.3, p. 353). A similar rule applies to translators: they are named after the title, preceded by the label *tr.* or *trans.* (not *trans. by*) (18.2.3, p. 354).
- Page ranges: "A span of numbers is often elided to the fewest figures possible" but "do not elide digits in (or ending with) the group 10 to 19" (11.1.4, p. 189). This behavior cannot be replicated by CSL, but setting `page-range-format` to `minimal-two` gets closer than `chicago`, and errs on the side of verbosity.

adunning:
Styles for *New Hart's Rules: The Oxford Style Guide*, 2nd edition, created by encoding the example citations from the official guide in the [Zotero Test Items Library] (http://zotero.org/groups/2205533/collections/499SCKUW) and adjusting the output to match as closely as possible.

The most noticeable changes include:

- removal of em rules in bibliographies;
- listing authors of a book before rather than after the title;
- correct positioning of `edition`;
- support for multivolume works;
- URLs without angle brackets.

The code draws on a combination of the APA, Chicago, and MHRA styles, aiming for complete support of the CSL specification with consistent handling of variables across these families.

The *New Hart's Rules* styles do not yet include support for legal types. They should ideally include an embedded version of OSCOLA, similar to the embedding of Bluebook in APA and Chicago.

There is scope for creating many more variants on these styles. Those included here focus on the guide's humanities variants, but it also recommends forms for scientific publishing.

Closes #7105.

### Checklist

### DIFF
--- a/new-harts-rules-author-date.csl
+++ b/new-harts-rules-author-date.csl
@@ -12,7 +12,7 @@
       <name>Andrew Dunning</name>
       <uri>https://orcid.org/0000-0003-0464-5036</uri>
     </author>
-    <category citation-format="note"/>
+    <category citation-format="author-date"/>
     <category field="anthropology"/>
     <category field="communications"/>
     <category field="generic-base"/>


bwiernik:
Thanks!

PR #7879, opened 2025-11-05
MLA: Use official name; clean up old editions; minor corrections

adunning:
### Description
In MLA, revert to displaying the `source` variable only with a DOI or URL. While there are technically situations in which one could display the title without a URI, this is also resulting in titles such as 'K10plus ISBN' appearing in Zotero bibliographies.

### Checklist


bwiernik:
Could you explain a bit what MLA's rules around citing the source/source database are?

adunning:
Container 2 is ideally meant to provide a full citation of any source through which one has accessed an original. This is best illustrated by the [citation examples in the _MLA Handbook_](https://doi.org/10.1632/KVQB7337). Right now we can handle something like this, URL formatting limitations aside ([encoded in Zotero](https://www.zotero.org/groups/2205533/items/24H4TKT6)):

> Woudhuysen, H. R. _Leicester’s Literary Patronage: A Study of the English Court, 1578–1582_. 1981. Oxford U, DPhil thesis. _Oxford University Research Archive_, ora.ox.ac.uk/objects/td:602337209. PDF download.

Container 2 doesn't, however, need to be a website. The different examples of citing a painting illustrates the idea:

> Viewed in person
> Bearden, Romare. _The Train_. 1975, Museum of Modern Art, New York City.
>
> Nonell, Isidre. _La azotea_. 1894–96, Museo Nacional Centro de Arte Reina Sofía, Madrid.
>
> Viewed online
> Bearden, Romare. The Train. 1975. **_MoMA_**, www.moma.org/collection/works/65232.
>
> Viewed online in a slide presentation
> Benton, Thomas Hart. _Instruments of Power from America Today_. 1930–31. **_The Met_**, 2012, Metropolitan Museum of Art, www.metmuseum.org/blogs/now-at-the-met/from-the-director/2012/benton/slideshow. Slide 1.
>
> Viewed in a book
> Velázquez, Diego. _An Old Woman Cooking Eggs_. Circa 1618, Scottish National Gallery. **_The Vanishing Velázquez: A Nineteenth-Century Bookseller’s Obsession with a Lost Masterpiece_**, by Laura Cumming, Scribner, 2016, p. 27.

I had removed the conditional requiring the presence of a URL to cover examples such as these:

> _E.T. the Extra-Terrestrial_. Directed by Steven Spielberg, Universal Studios, 1982. **_Netflix_ app**.
>
> _Fleabag_. Created by Phoebe Waller-Bridge, season 2, episode 3, BBC, 18 Mar. 2019. **_Amazon Prime Video_ app**.
>
> _Wild Planet: Virtual Tour_. Australian Museum. **_Google Arts and Culture_ app**, accessed 1 Aug. 2022.

The problem is finding a way to support such examples without spraying titles automatically added to `source` such as  _Library of Congress ISBN_ all over a bibliography.

bwiernik:
Sorry to be a little dense, but could you add bolding to the `source` elements for the examples?

adunning:
Yes, I've updated my comment.

modern-language-association-annotated-bibliography.csl
```diff
@@ -947,13 +947,14 @@
   <!-- 6. Container 2 (MLA 5.102) -->
   <macro name="container2">
     <group delimiter=", ">
-      <!-- Title of Container -->
-      <text font-style="italic" variable="source"/>
-      <!-- Location -->
       <choose>
-        <if variable="source">
+        <if match="none" variable="source"/>
```
bwiernik, line 923: This line seems odd to me, but maybe that's just MLA style--if an item has a DOI but no source/catalog value, shouldn't the DOI be shown?

bwiernik:
@adam3smith do you have thoughts here? 

Ideally we would want to support MLA's requirement to cite things like the service where an item was accessed, but the frequent inclusion of metadata-source versus full text-source in Zotero items does make this a bit messy.

Do you have a preference between continuing to always include `source` and leaving it to users to clear the Zotero library catalog field as needed vs. only including `source` for items with URIs?

(Should we also consider being more conservative about saving translator names to the Zotero library catalog field?)

modern-language-association-annotated-bibliography.csl
adunning: Yes, it will be shown, but within container 1 rather than 2. This conditional matches it.

adunning:
I have no opinion on the best approach; I assumed it would reduce complaints on the forum to revert to the earlier behaviour, but if you think there's scope for improving `source` data in translators, that could be even better.

bwiernik:
@adam3smith I'm going to merge for now. I think we should consider reducing the frequency of saving metadata source information to Zotero Library Catalog (as I understand it, saving it there was primarily for MLA style anyway). Dan's comment in https://github.com/zotero/zotero/issues/5680#issue-3763998834 seems like he would be in favor of that too.

A simple fix would be to just drop saving translator information for ISBN and PMID retrieval, as those are clearly not full-text sources. Do you have a more robust idea?

PR #7881, opened 2025-11-05
Remove broken MLA 6th edition notes style

adunning:
This style is not functioning correctly and it seems simplest to remove it. It appears to have been kept solely for the journal *Angelaki*, but this publication now uses standard MLA in-text citations.


adunning:
Cf. <https://forums.zotero.org/discussion/127821/>.

PR #7913, opened 2025-11-23
Add ORCID IDs for contributors

adunning:
Replace contributor email addresses with ORCID IDs where possible for stable identification and privacy. Add an ID only where there is a single exact match in the ORCID registry for the recorded name.

These changes have been made using a simple [Python script that looks up names in the ORCID public API](https://github.com/user-attachments/files/23696185/enrich_orcid.py) and spot-checked by hand.

### DIFF
--- a/arthropod-systematics-and-phylogeny.csl
+++ b/arthropod-systematics-and-phylogeny.csl
@@ -13,8 +13,8 @@
       <uri>https://orcid.org/0000-0001-7945-4205</uri>
     </author>
     <category citation-format="author-date"/>
-    <eissn>1864-8312</eissn>
     <issn>1863-7221</issn>
+    <eissn>1864-8312</eissn>
     <summary>Arthropod Systematics and Phylogeny Journals style</summary>
     <updated>2021-12-09T16:32:46+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
--- a/boletin-de-la-sociedad-geologica-mexicana.csl
+++ b/boletin-de-la-sociedad-geologica-mexicana.csl
@@ -76,7 +76,7 @@
         <group delimiter=" ">
           <text term="available at"/>
           <group delimiter=", ">
-            <text variable="URL" prefix="&lt;" suffix=">"/>
+            <text variable="URL" prefix="&lt;" suffix="&gt;"/>
             <text term="accessed" suffix=" "/>
           </group>
           <date form="text" variable="accessed"/>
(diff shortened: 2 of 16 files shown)


adunning:
@POBrien333 There are several people with your name in ORCID: let me know which one you are, if you would like me to add your ID.

bwiernik:
The point of email addresses is to permit users of styles to contact the creator with error reports and update requests. We don't require contributors to include an email address, but if they supply one, I am not sure I agree this is a privacy concern. The report on the Zotero forums was the first time I've heard such a concern in the 13 years I've been involved in CSL.

I don't know that we need to have author contact details in styles, but I think we need to make a policy change about not including email addresses in styles before making this change. @adam3smith what are your thoughts?

adunning:
My idea with this is to get ahead of potential concerns such as the one just raised on the forum where there is a good alternative. Although I wasn't intending to suggest a blanket policy of removing email addresses, I do not think that contributors will always understand the implications of listing this in a CSL file. I was receiving a lot of random requests for help several years ago when my address was in the Chicago and MHRA styles. I suspect also that many of the email addresses in the files have expired. By referring to ORCID IDs where possible (which has become a common practice for journals), users can list, update, and remove academic contact information centrally as life happens.

bwiernik:
What's our level of confidence with the ORCID matching?

adam3smith:
Yes, I agree that orcids would be excellent as permanent identifiers style authors, but I am worried about mismatching them. Just because the name only exists in orcid once doesn't mean it's an actual match. The only reliable match would be by email address, most people don't have theirs public on their orcid profiles.

adunning:
We have the advantage that most people create an ORCID to submit a journal article or grant proposal, which is also the usual reason for creating a CSL style. I have so far found only one false positive: a name that looks quite common but for which there happens to be only one ORCID. I have however come up with a fairly usable interface for selecting between multiple matches, if you think this information should only be added manually.

It doesn't appear that email addresses can be fetched from the public API, but often one can match the listed email address with the institutions listed in ORCID, and sometimes one can even directly match the journal style name with the publications list in ORCID.

One could also use the ORCID API to create a useful report on the diversity of CSL contributors in applying for funding and so forth.

bwiernik:
That's a good point on the co-occurrence of ORCID and CSL generation. Could you do a short write up of the various matching signals to help us evaluate the precision?

POBrien333:
Thanks for the efforts, Andrew.
I'm not 100% convinced just replacing all of the emails is the right thing though. Think of where the providers of a style from a university or journal WANT to be contacted if something is wrong with the style?
While I do understand that we receive the majority of style error complaints in the forum/here on GitHub and we take care of it, certain styles are managed and maintained very closely by certain organisations. MDPI comes to mind.
ORCID doesn't really have a "contact" functionality, no?

adunning:
I have come to the conclusion that this should only be done if verified by hand.

PR #7918, opened 2025-11-24
CSE Manual: Use official naming scheme

adunning:
Use the normal CSL `<title>` formatting to alphabetize the 8th and 9th editions in the repository listing. Use the official published title of the source as the primary name.


adunning:
@bwiernik I updated this to take #7907 into account. It looks as if the CI has run into a caching issue of some sort, and I am hoping that merging something else will fix it.

PR #7921, opened 2025-11-24
ACS: Use official name in `<title>`

adunning:
The full title of the source for this style is the *ACS Guide to Scholarly Communication*, but <https://pubs.acs.org/page/acsguide> refers to it as the *ACS Guide*. There are no longer official editions but changes continue to be made, which is why I am suggesting the inclusion of the date in the title. (It is not clear to me whether the revisions in 2025 need CSL changes.)

### DIFF
--- a/american-chemical-society.csl
+++ b/american-chemical-society.csl
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="expanded" default-locale="en-US">
   <info>
-    <title>ACS Guide 2022</title>
+    <title>ACS Guide 2022 revision</title>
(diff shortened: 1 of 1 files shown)


american-chemical-society.csl
```diff
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="expanded" default-locale="en-US">
   <info>
-    <title>American Chemical Society</title>
-    <title-short>ACS</title-short>
+    <title>ACS Guide 2022</title>
```
bwiernik:
```suggestion
    <title>ACS Guide 2022 revision</title>
```

bwiernik:
Thanks, let's refer to the years as "revision" in the style title for clarity

adunning:
Yes, that addition makes it more consistent with the usual CSL naming scheme; thank you!

PR #7924, opened 2025-11-24
ASA: Use official name in `<title>`

adunning:
The official website at <https://www.asanet.org/publications/journals/asa-style-guide/> refers to this as the *ASA Style Guide*.


bwiernik:
@adunning Let's keep both a link to the official documentation source and to a page that has freely accessible guidelines/examples like Purdue OWL if the official documentation isn't freely available

PR #7928, opened 2025-11-25
NLM/Vancouver: Use source titles and official terminology

adunning:
'Vancouver' is a shorthand for any numbered/numeric/author–number referencing system (see <https://en.wikipedia.org/wiki/Vancouver_system>), named after the Vancouver Group, which is now known as the International Committee of Medical Journal Editors (ICMJE). This adjusts the titles of the various styles in the repository labelled as Vancouver to distinguish between a) the official style recommended by the ICMJE, published in *Citing Medicine: The NLM Style Guide*; b) sources that mostly follow the ICMJE Recommendations but not NLM style; c) numeric styles that have no relationship to ICMJE/NLM.

Rename the NLM variants as citation-sequence, citation-name, and name-year, which is the official terminology at <https://www.ncbi.nlm.nih.gov/books/NBK7265/> (which follows the CSE Manual).

### DIFF
--- a/taylor-and-francis-vancouver-national-library-of-medicine.csl
+++ b/taylor-and-francis-vancouver-national-library-of-medicine.csl
@@ -71,10 +71,13 @@
       <if type="article-journal">
         <group delimiter=". ">
           <text variable="DOI" prefix="doi: "/>
-          <group delimiter=": ">
+          <group delimiter=" ">
             <text term="cited" text-case="capitalize-first"/>
-            <text term="in" prefix=" "/>
-            <text variable="PMID" prefix=": PMID: "/>
+            <text term="in"/>
+            <group delimiter=": "> 
+              <text value="PMID"/>
+              <text variable="PMID"/>
+            </group>
           </group>
         </group>
       </if>


bwiernik:
I'm not sure about this one, it seems like "Vancouver" is much more widely used to refer to NLM style than "NLM". What do you think @adam3smith ?

adam3smith:
Don't we have a dependent style that's NLM so both are covered/discoverable? I definitely want to keep Vancouver

adunning:
It certainly doesn't help that there are at least five different names for the style in circulation! The problem is that 'Vancouver' has come to mean any numeric citation system, in the same way as 'Harvard' is a shorthand for author–date. As an outsider to the field, it took years before I realized it is even possible to link Vancouver to an official style, unlike Harvard. You'll find lots of references to Vancouver that appear at first to be using the NLM style, but are actually following AMA or CSE, or just doing their own thing. This is why I feel that it would offer a lot of clarity to use the official names as the `<title>`, while keeping Vancouver a searchable term.

adunning:
@adam3smith I hadn't seen your comment when I posted mine: as a Canadian, I am always charmed to see 'Vancouver' at the end of the list of styles in Zotero. I wonder whether it is used as a generic term for numbered styles more in British English than it is in the USA? I just tried searching for 'Vancouver style', and the 13th result is the first one following the style published by the NLM; and the official source doesn't look to be anywhere in the top 50 results. A lot of users are likely to select Vancouver in Zotero without knowing where it is coming from, probably without realizing that there are multiple interpretations of it, and get the wrong style. By contrast, searching for 'Citing Medicine NLM' will lead directly to the official source and make it very easy to check what the style is meant to accomplish.

adunning:
@adam3smith I've modified this in an attempt to meet your desire for a distinctive Vancouver style, which I am taking as referring specifically to the Vancouver Group/ICMJE Recommendations.

bwiernik:
@adunning I think there may be a difference in usage of "Vancouver" to generically refer to numeric styles in the UK vs elsewhere. In US and Canada (and some UK) author guideline examples I looked through, I see, I am typically seeing either (a) explicit references to NLM, AMA, or CSE styles, with no use of "Vancouver" at all, or (b) where "Vancouver" is used, I see it mostly being used explicitly as a synonym for NLM style (e.g., by linking to the NLM page, calling it "NLM/Vancouver", or providing examples that align with NLM style).

I think a more apt comparison than "Harvard style" would be to APA style — a specific style that many publications claim to follow but often have minor variations/deviations. 

Some examples
1. In Pubmed's "Cite" tool provided on .[article pages](https://pubmed.ncbi.nlm.nih.gov/29688512/), they refer to the style as "NLM" (they also offer AMA, APA, and MLA). 
2. The [BMJ](https://www.bmj.com/about-bmj/resources-authors/house-style) refers to "Vancouver style" and links to the NLM guidelines. 
3. [McGill University](https://libraryguides.mcgill.ca/citations/vancouver) refers to "Vancouver style" 
4. Journal editors referring to "NLM/Vancouver" https://onlinelibrary.wiley.com/doi/abs/10.1002/leap.1425


The Wikipedia [article](https://en.wikipedia.org/wiki/Vancouver_system) you linked to seems to distinguish between "Vancouver system" (any numeric style) and "Vancouver style", by which it seems to mean *Citing Medicine*:

> the names "Vancouver system" or "Vancouver style" have existed since 1978. The latest version of the latter is [Citing Medicine](https://en.wikipedia.org/wiki/Citing_Medicine), per the "Style and Format" section of the [Recommendations for the Conduct, Reporting, Editing, and Publication of Scholarly Work in Medical Journals](https://en.wikipedia.org/wiki/Recommendations_for_the_Conduct,_Reporting,_Editing,_and_Publication_of_Scholarly_Work_in_Medical_Journals) published by the International Committee of Medical Journal Editors (ICMJE).


I see enough [examples](https://iscpc.si/wp-content/uploads/2024/12/Instructions-for-authors.pdf) of guidelines referring only to "Vancouver style" in isolation with examples that I think we need to keep "Vancouver" displaying prominently.

With that in mind, I suggest:

1. Include "Vancouver" in the visible style name for NLM style variations in parentheses
2. When derivative styles refer to "Vancouver" currently, say "NLM/Vancouver" rather than more obscure things like "IJCME recommendations" unless the guidelines explicitly say the other phrase.
3. Given the ubiquity of vancouver.csl in CSL clients, we should keep the filename and style ID for these styles rather than renaming. (@adam3smith I think this case becomes rather confusing; should we consider renaming this style ID in this case?)
4. The NLM style title in the PR is currently "Citing Medicine: The NLM Style Guide 2nd Edition". Is "NLM" prominent enough here, or should we rename to "NLM Citing Medicine 2nd Edition" or similar? @adam3smith 

Could you make those changes @adunning?

adunning:
@bwiernik I hope that I have found a way to cover all four of your points.

For the title: what about "NLM Style Guide (Vancouver): Citing Medicine 2nd edition"? This reverses the title as officially published, but I think it is justifiable for the reasons you give. It is also consistent with the recent change of the title of CSE from _Scientific Style and Format: The CSE Manual for Authors, Editors, and Publishers_ to _The CSE Manual: Scientific Style and Format for Authors, Editors, and Publishers_.

I have retained `vancouver.csl` as a dependent style for the _ICMJE Recommendations_, which allows for an NLM style that uses the official terminology in _Citing Medicine_ without deleting the Vancouver style.

I am aiming to have the file structure reflect the explanation of the relationship between the recommendation documents in the [_AMA Manual of Style_, 3.1](https://doi.org/10.1093/jama/9780190246556.003.0003). My understanding is that the _ICMJE Recommendations_ currently dictate that their reference style is that of _Citing Medicine_, but this hasn't always been the case in the past and theoretically ICMJE could in the future decide to recommend a different style manual.

bwiernik:
Okay great

adunning:
I simplified the title of `vancouver.csl` to 'ICMJE Recommendations (Vancouver)'. There is an article at <https://doi.org/10.1136/bmj.p2069> that discuses their formation. 'ICMJE Recommendations (Declaration of Vancouver)' might be more historically accurate but I think this is unnecessary.

bwiernik:
Is this ready to merge @adunning ?

adunning:
@bwiernik I was going to synchronize the variants to remove unintended differences between them, but the changes are significant enough that Git is losing track of them in renaming, so I won't make any further changes to this.

bwiernik:
Okay, feel free to make a new PR after merging if that makes sense

adunning:
I will do that; thank you!

adunning:
I fixed a minor bug here reported at <https://forums.zotero.org/discussion/128891/>.

adunning:
I now have a synchronized version of the NLM styles with the affixes cleaned up and several bugs fixed, but I am waiting to submit a pull request until the IDs are merged, to ensure that Git is able to track the file renaming:

https://github.com/citation-style-language/style-variant-builder/blob/main/templates/nlm-template.csl

bwiernik:
@adunning feel free to make your follow-up

bwiernik:
@dstillman It would be good to update the default style packaged in Zotero from http://www.zotero.org/styles/vancouver to http://www.zotero.org/styles/nlm-citation-sequence

dstillman:
Done, but not having it be a renaming of vancouver.csl is a problem. It means that existing installations will end up with both NLM and the now-dependent Vancouver style. Even if we deleted vancouver.csl on upgrade in Zotero clients, existing documents that pointed to it would cause it to be redownloaded (assuming all went well — e.g., there was an internet connection), so people would end up with both. I don't think this is a good solution.

dstillman:
If this can't be a renaming, I think we'll need to continue bundling the dependent style in Zotero.

dstillman:
Ugh, but that's not straightforward either, because we'd have to bundle NLM as a hidden dependency (as happens if you install a dependent style manually) or else Vancouver wouldn't work for a new offline installation of Zotero. This is kind of a mess.

dstillman:
These names are also kind of unreasonably long and unwieldy, and in Zotero these styles (here `nlm-citation-sequence.csl` and `nlm-citation-sequence-brackets.csl`) are indistinguishable unless you hover over to get a tooltip:

Screenshot 2026-01-25 at 11 10 44 PM

dstillman:
This is causing duplicate Vancouver/NLM styles even for a new Zotero installation. Zotero comes with the non-dependent Vancouver, and on the first update, it gets updated to a dependent style, which installs NLM, and I guess does so in a way that circumvents the normal process of independent-parent styles being installed as hidden styles. That's a bug, but…it would've been helpful to have been pinged on all of this before it went out.

If we're including "(Vancouver)" in the NLM name anyway, and "ICMJE Recommendations (Vancouver)" won't be recognizable to most users of "Vancouver", is there a reason we can't add `vancouver` to renamed-styles.json like all the other Vancouver variants?

dstillman:
This commit is in fact already resulting in duplicate Vancouver/NLM styles in all Zotero installations as soon as the style update comes through. If we make `vancouver` a renamed style, we can add code (in Zotero 8+) to delete the renamed style if both the renamed style and the new one are installed. I think that's our best option at this point.

bwiernik:
Let's just make it a renaming

bwiernik:
@dstillman  done in https://github.com/citation-style-language/styles/pull/8026

adunning:
Thank you for finding a solution to this! Continued in #8035.

PR #7931, opened 2025-11-27
Update zurcher-hochschule-fur-angewandte-wissenschaften-soziale-arbei…

buhc:
**Description:**
term: no date. Suffix only when there is a citation locator.


POBrien333:
I've been trying to understand what you want to achieve, but can't figure it out.

buhc:
The problem was that if no date was specified for an in-text citation in Word, an unnecessary suffix (comma, space) was always inserted. 

Incorrect: (author, n. d., )
Correct: (author, n. d.)

The comma is only necessary if page numbers are specified. 

Is there a better way to prevent this?

POBrien333:
Ahhhh. Ok. Thanks for clarifying.

PR #7935, opened 2025-11-30
Render Harvard as a format rather than a style name

adunning:
'Harvard' is a shorthand for author-date style, especially in British English. Ensure that this is consistently formatted as a format/system name rather than as the title of a specific style, set off with parentheses rather than a hyphen.


bwiernik:
@adunning On this one, Cite Them Right is installed by default in Zotero, Mendeley, and others, so let's not rename the style IDs and instead keep them as "harvard-cite-them-right". It's inconsistent, but we should avoid changing style IDs when possible, especially for very popular styles.

Could you adjust the style IDs?

bwiernik:
Elsevier does refer to the style as a "Harvard" style, so let's keep that in the title like the others 

https://supportcontent.elsevier.com/Support%20Hub/Elsa/31730_References%20_Elsevier_Style_Guide.pdf

adunning:
@bwiernik Good catches! I've taken care of both points.

PR #7888, opened 2025-11-08
Rework associacao-brasileira-de-normas-tecnicas.csl

POBrien333:
### Description
This style was and is used heavily as a template for other university-specific variations.
We have a whole bunch of open PRs all with the same issues. This PR reworked the style for us not to have review the same thing over and over again.

PR #7958, opened 2025-12-19
Improve ABTNs and make more robust

andreappel:
Continues https://github.com/citation-style-language/styles/pull/7888#issuecomment-3582674370

### DIFF
--- a/associacao-brasileira-de-normas-tecnicas-numerico.csl
+++ b/associacao-brasileira-de-normas-tecnicas-numerico.csl
@@ -393,9 +393,7 @@
             <group delimiter=". ">
               <text macro="container-contributors"/>
               <text macro="secondary-contributors"/>
-              <group delimiter=": ">
-                <text macro="container-title"/>
-              </group>
+              <text macro="container-title"/>
             </group>
             <text macro="locators"/>
             <text macro="event"/>


POBrien333:
I assume you just took my code and made some amendements, right? If so, please replicate these changes in the other 2 styles and then we can merge. Thanks :)

andreappel:
@POBrien333
I've just finished replicating the changes between ANBT note, numeric and author-date. I'm not sure if I submitted the right way.

POBrien333:
Thanks!

Somehow though something happened with the -note.csl variant as that's now a numeric style.

andreappel:
I've tried to track the error without success. But the procedures I've took to updated has been:
1. updating the numeric variant;
2. copying its entire code below the </info> and replicating to the note and author-date variants, only updating the data, time, and contributor within the <info>;
2.1 for the note variant, though, I've also updated the title, translating the word note to "nota", in Portuguese;

So I think it could be something below the info tag. I've not worked with the numeric and note variants before, so I'm not sure if they need some differentiation within layout, citation or bibliography.

Also, do you think it would be convenient changing the title "associacao-brasileira-de-normas-tecnicas-numerico.csl" to "associacao-brasileira-de-normas-tecnicas-numeric.csl", for consistency, and removing the <category field="generic-base"/> from all the three variants? I guess I've seen that being dropped from all the styles in some past pull-request.

POBrien333:
Oh, that's why I had asked "I assume you just took my code and made some amendements, right? If so, please replicate these changes in the other 2 styles and then we can merge."

Your approach makes numeric styles out of everything basically.
The better approach would've been to take each style from my PR and then making those small amendments in each one. In the end you only changed a few delimiters etc.

andreappel:
> Oh, that's why I had asked "I assume you just took my code and made some amendements, right? If so, please replicate these changes in the other 2 styles and then we can merge."
> 
> Your approach makes numeric styles out of everything basically. The better approach would've been to take each style from my PR and then making those small amendments in each one. In the end you only changed a few delimiters etc.

I guess I've been able to revert, keeping the adjustments I did. Please take a look.

All three don’t cover some specific ANBT types though. In the future I’ll try to shrink the <bibliography> part of the abnt-ibict style, which is broader and more adherent to ABNT types, and then I’ll propose using that to overwrite these ones.

POBrien333:
Thanks for your efforts. Merging this now. 
Let's see what complaints we get back ;)

p3palazzo:
Thanks for all the hard work, @POBrien333 ! I submitted a quick fix to two appearance issues.

PR #7959, opened 2025-12-20
Update Ecosistemas csl to match current journal requirements

Pakillo:
### Description
A few minor changes to match current journal requirements, following a petition from the journal editors. Most importantly, adding the DOI/URL/ISBN after each reference in the bibliography. Also, the year should appear within parentheses.

### Checklist

### DIFF
--- a/ecosistemas.csl
+++ b/ecosistemas.csl
@@ -55,19 +55,20 @@
         <text variable="DOI" prefix=" https://doi.org/"/>
       </if>
       <else-if type="webpage post post-weblog" match="any">
-        <group delimiter=" "> <group delimiter=": ">
-          <text term="available at" text-case="capitalize-first"/>
-          <text variable="URL"/>
+        <group delimiter=" ">
+          <group delimiter=": ">
+            <text term="available at" text-case="capitalize-first"/>
+            <text variable="URL"/>
+          </group>
+          <group prefix="[" suffix="]" delimiter=" ">
+            <text term="accessed" text-case="capitalize-first"/>
+            <date form="text" variable="accessed">
+              <date-part name="month"/>
+              <date-part name="day"/>
+              <date-part name="year"/>
+            </date>
+          </group>
         </group>
-        <group prefix="[" suffix="]" delimiter=" ">
-          <text term="accessed" text-case="capitalize-first"/>
-          <date form="text" variable="accessed">
-            <date-part name="month"/>
-            <date-part name="day"/>
-            <date-part name="year"/>
-          </date>
-        </group>
-</group>
       </else-if>
       <else-if type="book" match="all">
         <text variable="ISBN" prefix=" ISBN: "/>


ecosistemas.csl
```diff
@@ -180,6 +181,17 @@
         </else>
       </choose>
       <text prefix=" " macro="access" suffix="."/>
+      <choose>
```
POBrien333: Please move this logic into the "access" macro that is for this purpose. (do not include the ISBN into that. Call that only for booklike items)

ecosistemas.csl
```diff
@@ -1,24 +1,23 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="es-ES">
-  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" and="symbol" et-al-min="3" et-al-use-first="6" demote-non-dropping-particle="sort-only" default-locale="es-ES">
```
POBrien333: I'm not really following all your et-al rules you've now set everywhere possible and that don't match up at all. 
I'd suggest you only call them on <citation... and <bibliography... and remove them from here.

ecosistemas.csl
```diff
@@ -41,7 +40,7 @@
     </names>
   </macro>
   <macro name="author-short">
-    <names variable="author">
+    <names variable="author" suffix=",">
```
POBrien333: please add this delimiter wherever the macro is called.

ecosistemas.csl
Pakillo: Thanks. Sorry, these are remains from earlier attempts to modify rules. Will clean up

ecosistemas.csl
Pakillo: Sorry I'm unclear about this. Should I add `suffix=","` to L126 instead than here? Thanks

ecosistemas.csl
Pakillo: Ok, thanks, will do

Pakillo: Done

ecosistemas.csl
Pakillo: Done

ecosistemas.csl
POBrien333: yes

Pakillo: ok, thanks!

Pakillo:
Many thanks for the review. I believe everything is fixed now

ecosistemas.csl
```diff
@@ -52,7 +51,10 @@
   </macro>
   <macro name="access">
     <choose>
-      <if variable="URL" type="webpage">
+      <if match="any" variable="DOI">
+        <text variable="DOI" prefix=" https://doi.org/"/>
+      </if>
+      <else-if variable="URL" type="webpage">
         <text value="Disponible en:" suffix=" "/>
```
POBrien333: Please don't hardcode text, but call the adequate term.

ecosistemas.csl
```diff
@@ -52,7 +51,10 @@
   </macro>
   <macro name="access">
     <choose>
-      <if variable="URL" type="webpage">
+      <if match="any" variable="DOI">
+        <text variable="DOI" prefix=" https://doi.org/"/>
+      </if>
+      <else-if variable="URL" type="webpage">
```
POBrien333: We'd probably want to add post and post-weblog here into the test. The test for the URL here is not neccesary.

ecosistemas.csl
```diff
@@ -63,7 +65,10 @@
             <date-part name="year"/>
           </date>
         </group>
-      </if>
+      </else-if>
+      <else-if type="book" match="all" variable="ISBN">
```
POBrien333: The additional condition for ISBN is not required.

ecosistemas.csl
```diff
@@ -124,14 +129,16 @@
       </group>
     </layout>
   </citation>
-  <bibliography hanging-indent="true" et-al-min="9" et-al-use-first="7">
+  <bibliography et-al-min="8" et-al-use-first="6" et-al-use-last="true" hanging-indent="true">
     <sort>
-      <key macro="author"/>
+      <key macro="author" names-min="1" names-use-first="1" names-use-last="false"/>
```
POBrien333, line 139: Do you actually need the names-min/names-use-first?

ecosistemas.csl
POBrien333: hardcoded label. Please check your style and use the correct term/label for it.

Also, you heavily rely on affixes. Please use gorup delimiters instead.

ecosistemas.csl
Pakillo: Thanks. Changed to 
``
<else-if type="webpage post post-weblog" match="any">
``

ecosistemas.csl
Pakillo: Sorry, not following. What is the adequate term?

Would something like this work? (that is, removing the text value and using a prefix for the URL)

```
      <else-if type="webpage post post-weblog" match="any">
        <text variable="URL" prefix="Disponible en: "/>
``

File: ecosistemas.csl (hunk 6 above)
Comment (Pakillo) [reply ecosistemas.csl]: Ok. Changed to 
```
      <else-if type="book" match="all">
        <text variable="ISBN" prefix=" ISBN: "/>
      </else-if>
``

ecosistemas.csl
Pakillo: I think so... Otherwise the bibliography is not sorted properly when there are multiple references from the same first author in different years with different coauthors...

ecosistemas.csl
Pakillo: Sorry, not following. I don't know what I need to change here...

ecosistemas.csl
POBrien333: no, that's still hardcoding text, but now as a prefix.
`<text term="available at"/>`

https://docs.citationstyles.org/en/stable/specification.html#miscellaneous

Pakillo: ah got it, thanks!

ecosistemas.csl
Pakillo: ok, changed to 
```
            <group suffix="." delimiter=". ">
              <label variable="page" form="short"/>
              <text variable="page"/>
            </group>
``

Pakillo:
@POBrien333 many thanks for your review, and apologies for my inexperience with CSL...

I believe everything is fixed now :pray:

POBrien333:
Reviewing the style guidelines just now I release that this journal [specifies ](https://www.revistaecosistemas.net/index.php/ecosistemas/about/submissions) to use APA.
Does it actually differ from APA in any way that requires an independent style?
If not, we'd prefer to just make this a dependent style.

Pakillo:
> Reviewing the style guidelines just now I release that this journal [specifies ](https://www.revistaecosistemas.net/index.php/ecosistemas/about/submissions) to use APA. Does it actually differ from APA in any way that requires an independent style? If not, we'd prefer to just make this a dependent style.

Thank you. I'm in conversation with the editorial board to see how it differs from APA (if anything). I'll look into making a dependent style on APA.

PR #7976, opened 2026-01-05
New style: American Psychological Association 7th edition (Polish)

csPinKie:
### Description
This pull request adds a Polish localization of the APA 7th edition style. While based on the standard APA 7th edition, this version incorporates specific Polish linguistic requirements and academic conventions.

Locale: pl-PL

Documentation source: Guidelines provided by [liberilibri.pl](https://apa7.liberilibri.pl/), which is a widely recognized standard for APA 7 in Poland based on recent publications and differs quite a bit from the typical english APA7 due to inherent differences in how the Polish language works.

Key changes: 
-Localized terms for "editor" (red.), "edition" (wyd.), "page" (s.), and "no date" (b.d.).
-Adjusted "et al." to the Polish equivalent (i in.).
-Corrected punctuation and delimiters for Polish citations (e.g., using "i" instead of "&" where appropriate for the locale) which are the main difference as well as problem and not able to be implemented in parallel to other languages in one file.

### Checklist

### DIFF
--- a/apa-pl.csl
+++ b/apa-pl.csl
@@ -11,7 +11,7 @@
       <name>cs_pinkie</name>
     </author>
     <contributor>
-      <name></name>
+      <name/>
     </contributor>
     <category citation-format="author-date"/>
     <category field="psychology"/>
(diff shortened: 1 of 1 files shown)


POBrien333:
From the looks of it the main changes are certain terms and then the ampersand to text change. Would the apa-no-ampersand.csl fix the latter for you?
Then the localisations can be just be put into the locale overrides. We try to not have too many variations of the same thing if not absolutely necessary.

PR #7978, opened 2026-01-06
OSCOLA: Create label macros and simplify conditionals

adunning:
Make the code for OSCOLA simpler and more robust in preparation for creating an embedded version for _New Hart's Rules_ and MHRA, as well as a probable fifth edition this year:

- Introduce label macros for `edition`, `volume`, `locator`, and `first-reference-note-number`.
- Simplify the logic for legal case titles, date parentheses, publisher info, and locator formatting for greater consistency and maintainability.
- Sort XML attributes and variables.

### Checklist

### DIFF
--- a/oscola-no-ibid.csl
+++ b/oscola-no-ibid.csl
@@ -487,14 +487,14 @@
         <text variable="locator"/>
       </else-if>
       <else-if type="book" variable="issued">
-        <text macro="locator-basic"/>
+        <text macro="label-locator"/>
       </else-if>
       <else-if type="book" variable="publisher">
-        <text macro="locator-basic"/>
+        <text macro="label-locator"/>
       </else-if>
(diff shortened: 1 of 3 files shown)


oscola-no-ibid.csl
```diff
@@ -190,247 +241,203 @@
...
-            <text macro="issued-year" prefix="[" suffix="]"/>
-          </else-if>
-        </choose>
-      </if>
-      <else-if type="legislation bill" match="any">
+      <if match="any" type="bill legislation">
         <text macro="issued-year"/>
+      </if>
+      <else-if match="none" type="article-journal article-magazine legal_case"/>
+      <else-if variable="authority number">
```
bwiernik, line 296: Should this be `match="all"` (current) or `match="any"`?

adunning: This is correct; it matches a parallel condition in the `article-case-info` macro.

PR #7986, opened 2026-01-09
Chicago and friends: Use numeric form for year-only dates

adunning:
Replaces the `form="text"` attribute with `form="numeric"` for year-only `<date>` elements across the Chicago styles and its derivatives in CSL. This ensures that years are consistently rendered as numeric values across certain localizations, as reported at <https://forums.zotero.org/discussion/128581/>.

### Checklist

### DIFF
--- a/chicago-author-date-17th-edition.csl
+++ b/chicago-author-date-17th-edition.csl
@@ -2501,7 +2501,7 @@
         <else>
           <text variable="collection-title"/>
           <choose>
-            <if match="any" type="interview" variable="interviewer">
+            <if type="interview">
               <!-- publisher possible with `interview` (cf. CMOS18 14.110) -->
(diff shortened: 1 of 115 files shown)


chicago-author-date-17th-edition.csl
```diff
@@ -2501,7 +2501,7 @@
         <else>
           <text variable="collection-title"/>
           <choose>
-            <if match="any" type="interview" variable="interviewer">
+            <if type="interview">
```
bwiernik, line 2504: What's the intention behind this change?

adunning: This is specifically there to catch an interview type being treated as a serial item, which could be a broadcast. The check for an interviewer is incorrect.

PR #7997, opened 2026-01-12
Create dependent/in-monte-artium.csl

adunning:
Dependent style for _In Monte Artium_.

### Checklist

### DIFF
--- a/dependent/in-monte-artium.csl
+++ b/dependent/in-monte-artium.csl
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" default-locale="en-GB">
+<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0">
   <info>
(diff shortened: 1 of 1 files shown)


dependent/in-monte-artium.csl
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" version="1.0" default-locale="en-GB">
+  <info>
+    <title>In Monte Artium: Journal of the Royal Library of Belgium</title>
+    <title-short>IMA</title-short>
+    <id>http://www.zotero.org/styles/in-monte-artium</id>
```
POBrien333, line 6: I'm wondering if we should add the `-journal-of-the-royal-library-of-belgium` part?

adunning: It's a subtitle rather than part of the journal's main title; cf. its ISSN listing, <https://portal.issn.org/resource/issn/20313098>. Unless you were thinking of a way to accomodate additional styles for this journal in its other languages?

PR #8003, opened 2026-01-13
update norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl

StianOby:
### Description
Correctly implements the thesis item type, in accordance with the style documentation.

### DIFF
--- a/norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
+++ b/norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
@@ -146,7 +146,7 @@
   <macro name="retrieved-from">
     <choose>
       <if type="article-journal">
-        <group delimiter=" "  prefix=" (" suffix=")">
+        <group delimiter=" " prefix=" (" suffix=")">
           <text term="cited"/>
           <text term="from"/>
           <text variable="archive" text-case="capitalize-first"/>


POBrien333:
Thanks.
Would you be able to also add formatting for webpages? Currently the style doesn't do that.

Also, please update the date in the info section.

StianOby:
Thanks for catching the "updated" date. I have neglected that for the last couple of updates.

I will add the webpage item type in a separate pull request.

StianOby:
Nevermind, I ended up adding the web page item type to this pull request. Hope everything is on order now – I'm not terribly experienced with VSCode and Git.

norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
```diff
@@ -319,7 +319,23 @@
             </choose>
           </group>
         </else-if>
-        <else-if type="book thesis" match="any">
+        <else-if type="webpage" match="any">
+          <group suffix=".">
+            <choose>
+              <if variable="author" match="any">
+                <text suffix=", " macro="author-full"/>
```
bwiernik: Throughout all of your changes, can you please set spaces/punctuation with `<group delimiter=", ">`, etc. rather than as prefix/suffix? This helps avoid extraneous punctuation showing when variables are missing

bwiernik (commented):
Thanks that's exactly how to update the PR. 

Made a general style comment -- can you update your formatting to use groups/delimiters rather than prefix/suffix?

norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
StianOby: Will do. Should I just do this for the current changes? The use of prefix/suffix is endemic in this style, so I guess a thorough revision is necessary. But I am unsure whether it is best to fix it as part of this PR (which will then delay the new features, as I will need some time to learn the syntax properly and to actually go through the entire style) or whether to just fix the amended parts, merge into master, and then I'll make a new PR with the full revision later.

POBrien333: Yeah, as part of this PR would be appreciated.
I'd pull the author macro out of the conditionals as it doesn't need to be repeated for each item type.

StianOby: I have now updated the style, trying to use group delimiters wherever feasible. As a CSL novice, I might have overcomplicated things or underused the group delimiter feature.

PS: I did not understand what @POBrien333 meant by "pull the author macro out of the conditionals". Can you or anyone else explain?

POBrien333:
- removed  most localed overrides as they were duplications from the standard locale
- removed hardcoded terms
- simplified some delimiters
- set et-al rules more centrally

Please double check the et-al rules. Wasn't 100% what you want to achieve

norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
POBrien333: Can ignore. I hadn't fully looked at the style and it'll be too complex.

norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
```diff
@@ -180,7 +173,7 @@
         </group>
       </if>
       <else-if locator="section paragraph" match="any">
-        <text variable="locator" prefix=" §&#160;"/>
+        <text variable="locator" prefix=" § "/>
```
POBrien333: You can set `<label variable="locator" form="symbol"/>` instead of hardcoding

POBrien333:
With your merge into master you've undone some changes I had made. Or was that on purpose?

StianOby:
This is on purpose, as the amendments produced some errors (trying to point them out in the commit titles). I'm currently working on the style, but realize I might be merging/pushing too much and that this may be confusing. Sorry about that. I am still learning best practices. I'm not the original author of the style, just the maintainer, and before I have mostly done marginal improvements, no rework of this kind.

norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
StianOby: I didn't understand this comment. I think the purpose of the original HTML entity is to ensure that the § symbol and the number that follows stats on the same line (non-breaking space).

StianOby:
Now, after doing some adjustments to the changes made by @POBrien333, it seems as if all items are displayed as expected. Hopefully the style is ready to be merged. But see my reply to the review above, which I didn't fully understand.

PS: I really appreciate your help with improving the style. I am not very experienced with CSL, and having an expert look over and add improvements is really useful. I learned a few new tricks along the way. Thank you!

norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
```diff
@@ -208,35 +212,45 @@
...
+          <text variable="number"/>
+          <text prefix="(" suffix=")" font-style="italic" variable="title-short"/>
+        </group>
       </else-if>
       <else>
-        <text suffix="s dom av " macro="author-full"/>
-        <text macro="issued-full-date"/>
-        <text prefix=" (" suffix=")" variable="title-short"/>
+        <group delimiter=" ">
+          <text suffix="s dom av" macro="author-full"/>
```
POBrien333: Not sure what "s dom av" means. Please replace with a term or label.

norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
POBrien333: You can set that on a group with this:
non-breaking space: `&#160;`

POBrien333 (changes requested):

norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
StianOby: The "s" is the possessive in Norwegian (comparable to "'s" in English). "dom av" means "judgment of".

I have now made "dom av" a term. The possessive "s" I was a bit unsure about. So far it has been kept as a suffix, since adding it as a term would unnecessarily complicate the groups and delimiters here (IMHO). If that is not the right way to go about it, I will amend.

StianOby:
Seems as if I have misunderstood CSL terms. I though I could define a new term for a citation style. Turns out you cannot, and thus I need to amend my solution for "s dom av"...

StianOby:
That should be it, I think.

StianOby:
One more thing: the term for number (short) is currently being displayed as "no." rather than the correctly localized "nr."

I guess the reason is that this term was wrongly localized in the CSL locales/locales-nb-NO.xml file. I corrected this in a committ just over a day ago: https://github.com/citation-style-language/locales/commit/d62a59a7f8261178284fbd9740254ba022274900

Thus, I assume that the style should be left as is, and that the correct CSL locales file will be pulled into Zotero at some point. Does this happen with every point release of Zotero? If there is a long delay, we should consider re-adding the number (short) term to the citation style as a temporary workaround.

POBrien333:
"no." Is not a thing at all in Norwegian?

StianOby:
Correct. The Norwegian word for number is "nummer", and it is abbreviated "nr." We never use the "no." abbreviation for "nummer".

norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
POBrien333: Please replace the hardcoded text with a term/label if possible.

StianOby: Done.

PR #8005, opened 2026-01-16
Chicago: Update `webpage` to use `publisher` variable

adunning:
Use the `publisher` variable that Zotero now provides on `webpage` items, which corresponds closely to 'the owner or sponsor of the site' in CMOS 14.104. Creating a distinction between the website title and its publisher allows the style to match the Manual's examples more closely, such as the Google example in the Quick Guide, <https://www.zotero.org/groups/2205533/items/QVWG47MP>; and the AI examples, <https://www.zotero.org/groups/2205533/items/KUTRD58P> and <https://www.zotero.org/groups/2205533/items/DUJY83G4> (CMOS 14.112). Correct the rendering of a `part-title` on a web page.

### Checklist

### DIFF
--- a/chicago-author-date-17th-edition.csl
+++ b/chicago-author-date-17th-edition.csl
@@ -1510,16 +1510,25 @@
     <group delimiter=" ">
       <choose>
         <if type="chapter" variable="container-title genre">
-          <text form="short" variable="genre"/>
           <!-- untitled introductions: cf. review model, which includes a title (CMOS18 14.101) -->
-          <text macro="source-monographic-preposition-note"/>
-          <text font-style="italic" form="short" text-case="title" variable="container-title"/>
+          <text form="short" variable="genre"/>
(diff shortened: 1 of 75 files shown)


adunning:
@bwiernik @adam3smith @POBrien333 Would one of you please be kind enough to prioritize this for review? It's not a complex change, but the latest version of Zotero makes it rather urgent. Its addition of the Publisher field makes this the most obvious place to store what Chicago calls 'the owner or sponsor of the site', which is much better than cramming everything into `container-title`, but I hadn't taken account of this variable when I wrote the style. Right now, a correctly encoded `webpage` item will not use the field correctly. Parallel corrections are in #7990 and #7993.

bwiernik:
Thanks! Do you think it's a good idea to render `container-title` in the event that `publisher` is empty? I'm thinking no. What do you think?

PR #8006, opened 2026-01-16
Add nn support and extra documentation link (user's guide)

StianOby:
Added an extra documentation link, that points to a new page with information and guides for users of the style, authored by the current maintainer (me): https://obykanalen.no/zotero/

Added locale info for Norsk - Nynorsk (nn), which is a dialect of Norwegian - Bokmål (nb). Since the removal of hardcoded text in https://github.com/citation-style-language/styles/pull/8003 the style should now support both dialects through the general CSL locale files.

### DIFF
--- a/norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
+++ b/norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="expanded">
+  <!-- Polyglot; this citation style supports both Norwegian dialects, i.e. bokmål (nb-NO) and nynorsk (nn-NO) -->
   <info>
     <title>Norsk henvisningsstandard for rettsvitenskapelige tekster (Norsk - Bokmål | Norsk - Nynorsk)</title>
     <id>http://www.zotero.org/styles/norsk-henvisningsstandard-for-rettsvitenskapelige-tekster</id>


norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
```diff
@@ -26,6 +27,12 @@
       <term name="accessed">lest</term>
     </terms>
   </locale>
+  <locale xml:lang="nn">
```
POBrien333, line 31: The style, in its current state, will never use this locale.
In line 2 you have the language hardcoded to one single language. See in Zotero's document preferences where you can set the language. That's greyed out as it's set. You'd need to remove the default-locale and then add a polyglot line like for other styles.

POBrien333:
Note that on your new website, you can link to the style with this link: https://www.zotero.org/styles?q=id%3Anorsk-henvisningsstandard-for-rettsvitenskapelige-tekster

norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
StianOby: Thanks for spotting this!

Removed default locale in a new commit. Didn't find any information in the CSL specification or by searching about "a polyglot line". Checked e.g. the Chicago styles, which have both US and UK dialects, and couldn't find something that this style is lacking. But perhaps I am missing something?

StianOby:
(see comment to the review above)

norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
bwiernik: There is no way to specify multiple default locales--there can be one or none. So omitting the default locale is correct here

POBrien333: See e.g. https://github.com/citation-style-language/styles/blob/bf9c383af1c0e3417d4a95e007154e382c29f968/textual-cultures.csl#L3

StianOby: Will add this later today.

StianOby:
Polyglot line added. Should be ready to merge.

PR #8007, opened 2026-01-16
Improve journal article examples in `items.json`

adunning:
Fix the Fenner example in the sample citations, which has an article number rather than a page reference.

### DIFF
--- a/spec/sheldon/items.json
+++ b/spec/sheldon/items.json
@@ -1,54 +1,5 @@
 [
   {
-    "id": "hanckeVarietiesCapitalismConflict2007",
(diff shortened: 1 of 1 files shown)


bwiernik:
Can you add a journal article with a page as well?

adunning:
I've added another journal article that shows the standard volume/issue with pages. I looked for one that would exhibit more of a style's features: this one has more than three authors, is part of a special issue title with two editors, uses issue and date ranges, has a page reference that will show the style's `page-range-format`, and includes PMID/PMCID.

If it would be useful to replace any of the other citations with examples that use the newer variables such as `part-title`, there are plenty to hand in the Test Items Library.

bwiernik:
Awesome thank you! Feel free to add/replace other items in another PR if some good choices come to mind from the Test Items Library. It would be good to leave Campbell/Pedersen and Mares in place given their use in style requests.

PR #8019, opened 2026-01-20
Change et-al-use-first attribute value in BSSA CSL file

andreasplesch:
After pushback from journal review to increase number of authors, also see https://www.seismosoc.org/publications/bssa-submission-guidelines-2/#data for journal example of 10 use-first authors.

### Description
Increase et-al-use-first for biblio. from 1 to 10.

### Checklist

### DIFF
--- a/bulletin-of-the-seismological-society-of-america.csl
+++ b/bulletin-of-the-seismological-society-of-america.csl
@@ -17,12 +17,13 @@
     <category field="geology"/>
     <issn>0037-1106</issn>
     <eissn>1943-3573</eissn>
-    <updated>2018-08-14T15:15:20+00:00</updated>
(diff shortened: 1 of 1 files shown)


POBrien333:
Please change the date in the info section with today's date.
In the guidelines I see that the  et al. is styled in italics and the style currently doesn't do that. Can you add that?

andreasplesch:
Thanks for the quick response.

> Please change the date in the info section with today's date.

Updated with approximate timestamp since time of day also required.

> In the guidelines I see that the et al. is styled in italics and the style currently doesn't do that. Can you add that?

Following other examples, added et-al font-style in author macro which is only used in biblio. section. Perhaps there is a better solution for restricting italics to biblio. Tested with zotero.

POBrien333:
Thanks!

PR #8022, opened 2026-01-22
Update norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl

StianOby:
Added support for the "treaty" item type (via Zotero Extra fields for the time being)

The hardcoded terms on line 444 and 446 are terms that I could not find in the locale file, which are rather specific for citing treaties. They are, respectively, the Norwegian equivalents of "treaty between" (444) and "in force" (446).

PS: can terms that do not exist in the locale file at all be defined in the style-specific terms section? E.g. like this (see the last two term name lines):
```
  <locale xml:lang="nb">
    <terms>
      <term name="and others">et al.</term>
      <term name="accessed">lest</term>
      <term name="treaty between">traktat mellom</term>
      <term name="in force">i kraft</term>
    </terms>
  </locale>
```
If so, I would like to amend the CSL to set the terms like this before the pull request is merged.

### DIFF
--- a/norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
+++ b/norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
@@ -454,16 +454,16 @@
             </group>
             <group prefix="(" delimiter=", " suffix=")">
               <group delimiter=" ">
-                <text term="henceforth"/>
                 <!-- Hack (see <locale> for details) -->
+                <text term="henceforth"/>
                 <names variable="author">
                   <name delimiter-precedes-last="never" name-as-sort-order="first" et-al-min="99" and="text"/>
                   <et-al term="and others"/>
                 </names>
               </group>
               <group delimiter=" ">
-                <text term="in press"/>
                 <!-- Hack (see <locale> for details) -->
+                <text term="in press"/>
                 <!-- "issued" is mapped to "Date Enacted" in Zotero, but nevertheless used for the "in force" date in this style due to the above-mentioned mention of treaty in the CSL specification for "available-date". -->
                 <text macro="issued-full-date"/>
               </group>


StianOby:
I see this has been labled as "incomplete submission". Why is that?

bwiernik:
@POBrien333 do you have some comments waiting for submission?

POBrien333:
Because you said "If so, I would like to amend the CSL to set the terms like this before the pull request is merged." I interpreted this as "Work in progress" and was waiting for your green light.

norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
```diff
@@ -415,13 +428,37 @@
...
+              <!-- In the CSL specification, "Date Variables", it is stated under "available-date": "the date a treaty was made available for signing". In (legal) citation of treaties, and in the style in question, date enacted and date in force is used. Date enacted and the date a treaty is opened for signature is usually the same date, thus "available-date" is used for date enacted. -->
+              <text macro="available-full-date"/>
+            </group>
+            <group prefix="(" delimiter=", " suffix=")">
+              <names prefix="traktat mellom " variable="author">
+                <name delimiter-precedes-last="never" name-as-sort-order="first" et-al-min="99" and="text"/>
+                <et-al term="and others"/>
+              </names>
+              <!-- "issued" is mapped to "Date Enacted" in Zotero, but nevertheless used for the "in force" date in this style due to the above-mentioned mention of treaty in the CSL specification for "available-date". -->
+              <text prefix="i kraft " macro="issued-full-date"/>
```
POBrien333: Can you replace this with a term that fits?
https://docs.citationstyles.org/en/stable/specification.html#miscellaneous

As mentioned before, we try avoid hardcoded text.

StianOby: No terms in the spec fit. Therefore my initial question: can I add "custom" terms to the style term definitions? As suggested in my initial example?

Re-reading the relevant parts of the specification I saw this (emphasis added):

> Localization data, by default drawn from the “locales-xx-XX.xml” locale files, may be redefined or supplemented with cs:locale elements, which should be placed directly after the cs:info element.

I guess this means that I can have additional terms in the style CSL, which are not at all defined in the locale file? (Thus _supplementing_ the locale file.)

StianOby:
> Because you said "If so, I would like to amend the CSL to set the terms like this before the pull request is merged." I interpreted this as "Work in progress" and was waiting for your green light.

I was hoping that you or someone else could give a yes/no answer, and if the answer was "yes" I could update the pull request. if the answer is "no", then this should be ready to merge. See also my comment on your review just above.

norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
POBrien333: You can use a term like "in press" or something that matches kind of, but is otherwise not used, then redefine it in the locale.
You can NOT invent non-existing terms. The style will not validate against the XML schema and we'll not be able to accept the PR.

StianOby: Thanks for answering my question. I'll think about this a bit more before determining which locale term to use. I find it rather strange that it is better to "misuse" locale terms in order not to hardcode, though I see the benefits in cases where the terms are different in the two dialects that the style supports (which could easily have been the case).

FYI: I have also added these terms to the issue for new terms in the CSL Schema repo: https://github.com/citation-style-language/schema/issues/438#issuecomment-3786967642

POBrien333: In your case, YOU will want this as your style supports several languagse. 
WE want a style that uses best practises, because people use a given style as a template. if there is bad practises in the template style we then need to review/fix those issues now in 2 styles.

StianOby:
Revised the pull request now, to use terms (piggybacking on "in press" and "henceforth" for the two treaty-specific terms that were not available in the CSL spec).

This should be ready to merge now.

StianOby:
Now these errors confuse me... They seem to be errors related to other styles?

Perhaps the errors were introduced with this (I see no automated tests were run before merging that): https://github.com/citation-style-language/styles/pull/8027 ?
@bwiernik ?

StianOby:
Apparently all I needed was to sync my branch to pass the tests.

PR #8030, opened 2026-01-27
Update norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl: macro refactoring and some improvements for "treaty"

StianOby:
### Description
- refactored title macro and nor-legislation macro to make code more logical/less obscure.
- add hard spaces for relevant locators
- reordered locator conditionals and removed unneccesary conditional
- improve treaty citation (add options for other citation formats using the "Public Law Number"/"number" field)
- add support for "article" locator, including forcing long-form where appropriate (though this will only work in Zotero after this fix is pulled into zotero/zotero: https://github.com/Juris-M/citeproc-js/pull/263)

### Checklist

### DIFF
--- a/norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
+++ b/norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
@@ -203,10 +203,10 @@
       <else>
         <choose>
           <if locator="paragraph article-locator" match="any">
-              <group delimiter=" ">
-                <label variable="locator" form="long"/>
-                <text variable="locator"/>
-              </group>
+            <group delimiter=" ">
+              <label variable="locator" form="long"/>
+              <text variable="locator"/>
+            </group>
           </if>
           <else-if locator="page">
             <group delimiter="&#160;">


StianOby:
(sorry for the messy number of commits, I had some merge issues that seem to have cause some commits to duplicate...)

StianOby:
This should be ready to merge.

StianOby:
While waiting for review I added another very minor feature: support for journals with issue numbers, with page numbers resetting for each issue. (This is the case for some Norwegian law journals, including MarIus and PetrIus.)

StianOby:
@POBrien333 this is now ready for review (and hopefully merge).

norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
```diff
@@ -195,23 +184,49 @@
...
-      <if type="legislation bill" match="none">
-        <group delimiter=" ">
-          <label variable="locator" form="short"/>
-          <text variable="locator"/>
-        </group>
+      <if type="legislation regulation" match="any">
+        <choose>
+          <if locator="section paragraph" match="any">
+            <group delimiter="&#160;">
+              <text term="section" form="symbol"/>
```
POBrien333: Could you change this to a label and if need, redefine them in the locale?

norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
POBrien333: What does "s dom av" mean? Could we replace this with a term?

StianOby: We discussed this in the PR when this was introduced: https://github.com/citation-style-language/styles/pull/8003#discussion_r2693863741

tl;dr: there is no relevant lable available. "dom av " "judgment by ". And the "s" is the possessive s in Norwegian (similar to 's in English).

StianOby: Forgot to mention the important yet obvious bit: there is no CSL term for "judgment" or "of", and styles cannot define their own terms (as I have now learned). So, I sadly need to hardcode this bit.

StianOby: I could perhaps slightly improve this by using the term "by", since the Norwegian word for by is the same as the Norwegian term for of (="av"). I would still need to hardcode "s dom", though.

StianOby: I have committed a version with "av" no longer being hardcoded, and instead rendered by the CSL variable "by". This is a slight hack, but is probably acceptable due to Norwegian using "av" for both "of" and "by".

norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
```diff
@@ -195,23 +184,49 @@
...
-          <text term="section" form="symbol"/>
-          <text variable="locator"/>
-        </group>
-      </else-if>
       <else>
-        <group delimiter=" ">
-          <label variable="locator" form="short"/>
-          <text variable="locator"/>
-        </group>
+        <choose>
```
POBrien333, line 204: Can this whole conditional not be:
If paragraph or article locator
else?

norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
StianOby: Done.

norsk-henvisningsstandard-for-rettsvitenskapelige-tekster.csl
StianOby: Was able to simplify it a bit at least. Not sure whether I want non-breaking space for the longform paragraph/article-locator. Thus a separate check for them. But for page numbers there should be a non-breaking space. For others I am not sure, and thus keeping default space for now.

StianOby:
I think I have addressed all issues you identified, @POBrien333.

PR #8031, opened 2026-01-27
Create Portal: Libraries and the Academy

adunning:
Although the author guidelines state that the journal follows Chicago 17th edition, the most recent articles are following the 18th edition.

### Checklist


POBrien333:
Thanks

PR #8032, opened 2026-01-27
Journal of Historical Geography: update style

adunning:
Update journal to MHRA as cited on its 'Guide to authors'.

### Checklist


POBrien333:
Thanks.
Wouldn't it be better to update this here: https://github.com/citation-style-language/journals/blob/eea9baad745e9805152d9473f15635e82f5172b3/elsevier/_journals.tab#L2046

adunning:
Done in <https://github.com/citation-style-language/journals/pull/160>, but the version here is still better because it has a documentation link and both ISSNs.

PR #8035, opened 2026-01-27
NLM: Synchronize variants and reorganize

adunning:
Synchronize NLM variants to ensure consistent behaviour. Reorganize macros using the structure and terminology in *Citing Medicine*. Rewrite affixes as delimiters. Support `number`, `supplement-number`, `DOI`, `PMID`, `PMCID` variables.

### Checklist

### DIFF
--- a/nlm-name-year.csl
+++ b/nlm-name-year.csl
@@ -485,7 +485,7 @@
       <text variable="references"/>
     </group>
   </macro>
-  <citation collapse="year" disambiguate-add-year-suffix="true" et-al-min="3" et-al-use-first="1">
+  <citation collapse="year" disambiguate-add-givenname="true" disambiguate-add-names="true" disambiguate-add-year-suffix="true" et-al-min="3" et-al-use-first="1" givenname-disambiguation-rule="primary-name-with-initials">
     <sort>
       <key variable="issued"/>
       <key macro="author-short"/>


dstillman:
These are better names, but I worry that we're still going to end up with a huge amount of confusion in Zotero. From just the first day:

https://forums.zotero.org/discussion/129318/cant-download-vancouver-style
https://forums.zotero.org/discussion/129335/vancouver-style-is-missing-in-zotero-8
https://forums.zotero.org/discussion/129347/in-the-last-update-of-zotero-the-vancouver-its-missing

Maybe "NLM/Vancouver" will help, but "Vancouver" is already in there — it's shown in the screenshot in the third thread — and people are missing it.

I understand the reasons for these changes, but I just don't think it's realistic to not have a style beginning with "Vancouver". Having it be searchable, or elsewhere in the title, is only minimally helpful — people are going to scroll to the end of the style list, in Zotero or any other tool, and look for "Vancouver".

So while I think adding more precise descriptors to the title is fine, I would really advocate for swapping these and putting "Vancouver" first, or calling it something like "Vancouver (NLM Citing Medicine 2nd edition, citation-sequence)".

(I really wish we could have had a conversation about this on the CSL list before these changes all went out.)

adunning:
@dstillman Thank you for addressing this! I'm sorry that you weren't aware of the original proposal; there had been little discussion of it for two months, so it appeared that anyone who was going to comment on it had done so. By 'CSL list' I assume you mean the [CSL Discourse page](https://discourse.citationstyles.org). My understanding is that this is for discussion of CSL itself (i.e. the schema and tools), while development of individual styles happens entirely in this repository. For example, when I was reworking the Chicago styles, I wrote a post on Discourse to obtain opinions of the best way to present the different variants, but had no response. Would it be helpful in the future to tag you on particular types of changes to bring them to your attention?

The official name of the book from which these styles are derived is _Citing Medicine: The NLM Style Guide for Authors, Editors, and Publishers_, 2nd edition. Somehow, the style names need to reflect this. It must also be noted that the creators do not wish their style to be known as Vancouver: the word only occurs in the 1,690-page PDF of _Citing Medicine_ as a place of publication in the sample citations. The ICMJE Recommendations do not use this term anywhere.

The links you supplied are very helpful. In Zotero, both NLM and Vancouver are conveniently at the end of the default list, so I hope that this simplification of the name will make 'Vancouver' sufficiently visible; swapping NLM and Vancouver would not change the order of the style's appearance in the list.

It looks to me as if the main problem comes about when users search for Vancouver in the style repository, look through the long list of options alphabetically, and assume the style is missing because there is nothing under V. To address this, I have proposed adding a dependent style listed under 'Vancouver Declaration', which was the original name in 1978 for what was published as the _Uniform Requirements for Manuscripts_ in 1979, and has obviously persisted. There is a [helpful _BMJ_ article that summarizes this history and dissects the terminology](https://doi.org/10.1136/bmj.p2069). I hope that this will both allow the listing of NLM under its official name and ensure that users looking under the style's old name will find it.

dstillman:
> Would it be helpful in the future to tag you on particular types of changes to bring them to your attention?

I suggested the CSL list because this is a major change affecting all CSL tools, but realistically, I don't know that you would've gotten feedback from anyone else. But yes, we'd want to know ahead of time about any major changes to styles that are bundled with Zotero. Because of our special relationship to CSL and the use of the Zotero Forums for style support, we push CSL changes to users immediately, so changes like these are particularly impactful for us. I don't know that any other tools do that.

> It must also be noted that the creators do not wish their style to be known as Vancouver

I don't care about that. My concern is Zotero users. "Vancouver" has been a bundled style in Zotero for 15+ years, and it's is also the common term in many citation tools and university guides. (Incidentally, before that we appear to have had an "NLM" style. I don't remember the history there.)

> In Zotero, both NLM and Vancouver are conveniently at the end of the default list, so I hope that this simplification of the name will make 'Vancouver' sufficiently visible

Well, we can see. As I mentioned, "Vancouver" was already present in that person's screenshot. But yes, "NLM/Vancouver" is definitely better, so we can try that and see if we continue to get reports.

> To address this, I have proposed adding a dependent style listed under 'Vancouver Declaration'

I think that would just lead to more confusion. People looking for "Vancouver" expect to see "Vancouver" — they don't know or care about the history here. If we want to try to address the problem of people missing "NLM/Vancouver […]" in the default list and going to the styles page, I think we should have a style that's "Vancouver" with other details in parentheses — e.g., "Vancouver (NLM Citing Medicine 2nd edition, citation-sequence)". In Zotero, people would end up with both styles visible, since the NLM style is already bundled (and therefore won't be installed as a hidden dependency), but at least people could find what they're looking for.

adunning:
@dstillman Thanks for your ideas: I've added simpler Vancouver aliases.

dstillman:
Great, thank you.

PR #8037, opened 2026-01-28
Update et-al-min value in citation macro

docsez:
USW Harvard doesn't want 'et al' in in-text citations until four or more authors

### Description
(Briefly describe the style or changes you're proposing.)
Edited the line: <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true"> and changed the "3" to "4"

### Checklist
- [ ] Check that you've added a link to the style guidelines with `rel="documentation"`.  
- [ ] Check that you've added yourself as the `<author>` of the style or `<contributor>` for a style update. 
- [ ] Check that you've not changed line 1 of the style.

### DIFF
--- a/university-of-south-wales-harvard.csl
+++ b/university-of-south-wales-harvard.csl
@@ -164,7 +164,7 @@
   <macro name="patent">
     <text variable="number" suffix=" [Patent]."/>
   </macro>
-  <citation et-al-min="3" et-al-use-first="1" disambiguate-add-year-suffix="true">
+  <citation et-al-min="4" et-al-use-first="1" disambiguate-add-year-suffix="true">
     <sort>
       <key macro="year-date"/>
       <key variable="author"/>


adam3smith:
What is this based on? I'm seeing two works with three authors and et al for in text
 (i book and iTunes) in the Harvard guide

docsez:
> What is this based on? I'm seeing two works with three authors and et al for in text (i book and iTunes) in the Harvard guide

It's based on the University of South Wales Referencing Guide: https://www.southwales.ac.uk/media/university-of-south-wales/site-assets/documents/services/harvard-referencing-2023-en.pdf
On page 5 it's explicit about only using 'et al' if there are four or more authors.

adam3smith:
That's pretty vague though as for the in text format and see by contrast Roche et al. on page 18, which has three authors

docsez:
> That's pretty vague though as for the in text format and see by contrast Roche et al. on page 18, which has three authors

I don't think it's vague so much as internally contradictory (later in the document (on p18) in reference to iTunesU - in contrast with 'books' on p5 which I will be citing somewhat more often!).  p5 is very clear, to my mind - it's also clear to my tutors' minds (hence the desire to change it for everyone rather than just edit my own Zotero CSL).

adam3smith:
I think it's vague because page 5 doesn't actually include an example with three authors or a specific explanation of et al. _and_ because there are no corresponding rules or examples for other item types, including journal articles, which I assume you'll also cite frequently. I'd contact the library and ask them to clarify this in the guide -- if folks are expected to precisely follow citation guidelines, they can't have internal contradictions like this.

Anyway -- the Cite Them Right guide the list elsewhere on their website uses three authors before et al. so I guess we'll assume that's what they intend and switch this.

docsez:
> I think it's vague because page 5 doesn't actually include an example with three authors or a specific explanation of et al. _and_ because there are no corresponding rules or examples for other item types, including journal articles, which I assume you'll also cite frequently. I'd contact the library and ask them to clarify this in the guide -- if folks are expected to precisely follow citation guidelines, they can't have internal contradictions like this.
> 
> Anyway -- the Cite Them Right guide the list elsewhere on their website uses three authors before et al. so I guess we'll assume that's what they intend and switch this.

Thank you - I completely agree and will indeed get in touch with the library about their guidance (thank you for spotting the p18 inconsistency)

PR #8041, opened 2026-01-29
Chicago: Set off `container-author` from other contributors; always set `initialize-with`

adunning:
A `container-author` is rendered separately from other contributors in a bibliography (_CMOS_ 14.12):

> Mansfield, Harvey C., and Delba Winthrop. Introduction to _Democracy in America_, by Alexis de Tocqueville. Translated and edited by Harvey C. Mansfield and Delba Winthrop. University of Chicago Press, 2000.

This example is encoded at <https://www.zotero.org/groups/2205533/items/URSAL3JG>.

### Checklist


adam3smith:
Just as a sanity check, can you walk me through the logic of the implicit `match="any"` that Codepilot complains about in its review? Why is it not a problem that we're losing all the other container creators when there is a container-title?

adunning:
Copilot is out to lunch here because it thinks that the default is to match any rather than all. The effect here is merely to change the punctuation; the information displayed should be the same.

adam3smith:
Oh right -- can we make that explicit please? Both for readability and because I think citeproc-js can do weird things w the implicit match conditions

adunning:
I've relied on the default `match="all"` behaviour quite a lot in the codebase; it would take a lot of time to add the attribute wherever it is used. Would you mind if I leave this as it stands, at least for the time being?

The only inconsistent behaviour appears if there is more than one `type` specified (citeproc-js matches any of these in a way that is quite sensible but outside the spec), and I am already careful to avoid this.

adam3smith:
Right, that's what I remembered, wasn't sure if that extended to other situations. Yes, we can leave as is then.

adunning:
Thank you!

PR #8042, opened 2026-01-30
Update journal-of-the-royal-anthropological-institute.csl

POBrien333:
https://forums.zotero.org/discussion/129415/style-error-journal-of-the-royal-anthropological-institute#latest

- rework style
- add original-date


adam3smith:
Thanks!

PR #8057, opened 2026-02-06
OUP Journals: Rework SciMed using New Hart's Rules

adunning:
Rewrite the Oxford Journals SciMed styles using _New Hart's Rules_, which the style sheets indicate is their basis. This resolves some differences between the author–date and numbered variants. It also uses the new DOI format found in the 2024 style sheet and ensures consistent inclusion of this information. A report at <https://forums.zotero.org/discussion/129559> confirms this is the expectation of OUP editors. Adjust the title format to give the full organization name.

### Checklist

### DIFF
--- a/oxford-journals-scimed-author-date.csl
+++ b/oxford-journals-scimed-author-date.csl
@@ -3275,7 +3275,7 @@
     </choose>
   </macro>
   <!-- Citation -->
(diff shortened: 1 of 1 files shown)


adam3smith:
Edit: actually nevermind I missed part of the commit here. I'm still wondering if we can keep one of the ids like the original one?

adunning:
Yes, done.

adam3smith:
Thanks!

PR #8058, opened 2026-02-07
OSCOLA: Support DOI and use more global options

adunning:
Use a DOI if available; although the OSCOLA guide never refers to them, they give a more reliable permalink than the `URL` variable, and it is implied at <https://www.law.ox.ac.uk/research-subject-groups/publications/oscola-faqs> that they may be used in this way.

Set repeated name options globally to simplify code.

### Checklist


adam3smith:
Thanks!

PR #8059, opened 2026-02-07
Add publisher to `webpage` example item

adunning:
Include the publisher field for the "CSL search by example" item.


adam3smith:
Thanks!

PR #8060, opened 2026-02-07
SBL: Remove place of publication

adunning:
The publisher has released an amendment at <https://sblhs2.com/2025/12/05/publisher-locations/> instructing authors to remove the place of publication, in line with the 18th edition of Chicago.

Use titles as published, update documentation links, and remove first edition (from 1999).

### Checklist

### DIFF
--- a/renamed-styles.json
+++ b/renamed-styles.json
@@ -542,6 +542,7 @@
     "small-wiley": "small",
     "smartt": "biomed-central",
     "society-for-general-microbiology": "microbiology-society",
+    "society-of-biblical-literature-1st-edition-fullnote-bibliography": "society-of-biblical-literature-fullnote-bibliography",
     "soil-biology-biochemistry": "soil-biology-and-biochemistry",
     "sozialwissenschaften-teilmann": "sozialwissenschaften-heilmann",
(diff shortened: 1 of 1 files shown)


bwiernik:
Do you think we need to keep a version with publisher place around for slow adopting publishers and institutions? Or is that unnecessary?

adunning:
This style is merely a variant on Chicago, and the page to which I've linked above doesn't make it sound as if they want to support the old version as an option. My feeling is that if users want to diverge from official guidance, they will be better served by one of the other Chicago styles providing for the older format. Authors will undoubtedly tell us if they think otherwise.

mbquintana:
> This style is merely a variant on Chicago, and the page to which I've linked above doesn't make it sound as if they want to support the old version as an option. My feeling is that if users want to diverge from official guidance, they will be better served by one of the other Chicago styles providing for the older format. Authors will undoubtedly tell us if they think otherwise.

Coming from a regular user of this style, it would be very helpful if the older format was kept around. There are definitely still publishers who otherwise adhere to SBL but have not adopted this change yet, and the SBL style differs enough from CMOS that users will run into problems if they try to use one of the older CMOS styles instead.

PR #8073, opened 2026-02-13
APA and friends: Use `givenname-disambiguation-rule`

adunning:
APA 8.20 and CMOS 13.116 indicate that an author's initials should be added to an author–date citation wherever there is more than one person with the same surname, even if the dates are different. Reported at <https://forums.zotero.org/discussion/129753/>.

### Checklist

### DIFF
--- a/apa-annotated-bibliography.csl
+++ b/apa-annotated-bibliography.csl
@@ -2206,7 +2206,7 @@
     </group>
   </macro>
   <!-- Citation -->
-  <citation collapse="year" disambiguate-add-givenname="true" disambiguate-add-names="true" disambiguate-add-year-suffix="true" et-al-min="3" et-al-use-first="1">
+  <citation collapse="year" disambiguate-add-givenname="true" disambiguate-add-names="true" disambiguate-add-year-suffix="true" et-al-min="3" et-al-use-first="1" givenname-disambiguation-rule="primary-name-with-initials">
     <sort>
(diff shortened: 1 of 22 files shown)


zepinglee:
I think the correct rule for APA should be `primary-name-with-initials`. APA's disambiguation is only applied to the first name of each cite as described in <https://apastyle.apa.org/style-grammar-guidelines/citations/basic-principles/citing-authors-same-surname>.

Previously, the disambiguation rule for `apa.csl` was `primary-name-with-initials` before #7510 was merged.

https://github.com/citation-style-language/styles/blob/b00efc62ed14e15dabb31dcb35f486ff106d13fb/apa.csl#L2087

adam3smith:
Yup @zepinglee is right, and I believe that applies to Chicago too

adunning:
Thanks for catching this. Chicago 13.116 simply states, 'Where two or more works by different authors with the same last name are listed in a reference list, the text citation must include an initial (or two initials or a given name if necessary).' That means it needs to be either `all-names` or `primary-name`: it doesn't seem possible to tell which.

adam3smith:
I forgot why we set this to primary-name way back (I think still referencing the 16th edition, so maybe that had something more precise?), but `all-names` is so incredibly noisy, adding initials left and right, that it's basically never useful -- we got sooo many complaints when that was the only available CSL option (in 0.7, I think, maybe even in 1.0?)

adunning:
That seems sensible!

PR #8077, opened 2026-02-19
Add CSL for Journal of Vacuum Science & Technology A

thiasyoung:
## Description
Adds a new independent CSL style for Journal of Vacuum Science & Technology A, derived from the AIP style, with JVST A-specific bibliography behavior (omits journal article titles and issue numbers).

### Checklist
- [x] Added a template link in `<info>` with `rel="template"`.
- [x] Added documentation link(s) with `rel="documentation"`.
- [x] Added myself as `<author>`.
- [x] Used CSL labels/terms instead of hardcoded affixes where applicable.
- [x] Did not use `<text value="...">` unnecessarily.
- [x] Did not change line 1 of the style.

### DIFF
--- a/journal-of-vacuum-science-and-technology-a.csl
+++ b/journal-of-vacuum-science-and-technology-a.csl
@@ -14,10 +14,8 @@
     <category citation-format="numeric"/>
     <category field="physics"/>
     <category field="chemistry"/>
-    <issn>0734-2101</issn>
-    <eissn>1520-8559</eissn>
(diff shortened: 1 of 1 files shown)


POBrien333:
We already have styles for that journal. That's why the CI was complaining. The ISSN was used already. Just deleting the ISSN from your PR doesn't fix this. You'd need to make the changes to the already existing .csl file in the repository.
(https://www.zotero.org/styles?q=Vacuum%20Science)

PR #8089, opened 2026-03-03
Update sociologie.csl with contributor and DOI changes

pmerckle:
### Description
The list of corrections made is as follows:

- Repair of the display of DOIs and URLs for references, in accordance with the instructions to authors and what appears to be the practice in the bibliographies of certain articles in the journal. 
- Addition of a non-breaking space before the semicolon in in-text citations. 
- Addition of a “Speech” type, which was missing, to correctly format references to oral presentations at scientific events.

### DIFF
--- a/sociologie.csl
+++ b/sociologie.csl
@@ -229,8 +229,8 @@
               <choose>
                 <if type="article-journal bill book chapter graphic legal_case legislation motion_picture paper-conference report song thesis" match="none">
                   <date variable="issued">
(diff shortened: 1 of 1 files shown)


POBrien333:
Thanks for your PR @pmerckle 
Reviewing your PR I noticed that the style itself was in not ideal shape. I've gone over it (instead of pointing things out).
Please test out this new style and either make any fixes yourself or point them out.
I simplified the style a lot and made it more robust, but might've missed something. ;)

PR #8100, opened 2026-03-11
Create estonian-journal-of-archaeology.csl

POBrien333:
via emails


POBrien333:
@adam3smith 
I just noticed that this style has not appeared on the repository. How can we check this?

adam3smith:
@dstillman is the pipeline to Zotero Styles working?

PR #8102, opened 2026-03-14

jamesftierney:

### Description
This revision corrects several formatting errors in prior implementations and extends coverage to additional source types. Specific changes include: italicizing case names in subsequent short-form citations regardless of whether a title-short is stored, but not full first-occurrence citations; fixing the statute-citation macro's erroneous comma-delimited outer group (which produced 42 U.S.C. , § 1983) by switching to a space delimiter; correcting the et-al-min threshold from 4 to 3 per Bluebook R 30.2.3; eliminating a double-space artifact between the author group and source text by setting the outer citation group delimiter to an empty string; adding explicit bill/legislation and manuscript branches in the subsequent-citation block to prevent impermissible supra forms for statutes and procedural rules; adding a dedicated report branch for Restatements and Model Codes (title in small caps, § number from the Report Number field, institutional publisher in parenthetical); adding a webpage/post-weblog source branch with URL and "last visited" access macro; adding status-variable support in the article-journal branch for forthcoming articles; removing the <substitute> element from name-macro to prevent double-title output for authorless items; suppressing author output for legal_case, legislation, and bill types; removing a spurious <label form="verb-short"> from name-short-macro; suppressing URL and "last visited" output for case and statute citations; adding a legislation/bill branch to the container macro; and splitting bill and legislation into separate source branches so that Zotero Bill items render in Statutes at Large session-law format (title, Pub. L. No. number, § section, volume Stat. page (year)) while legislation items continue to use the codified-statute macro. 

### Checklist

### DIFF
--- a/bluebook-law-review.csl
+++ b/bluebook-law-review.csl
@@ -23,10 +23,14 @@
       <name>Shay Elbaum</name>
       <email>selbaum@law.harvard.edu</email>
     </contributor>
+    <contributor>
+      <name>James Tierney</name>
+      <email>jtierney1@illinoistech.edu</email>
+    </contributor>
     <category citation-format="note"/>
     <category field="law"/>
     <summary>The Bluebook legal citation style for law reviews.</summary>
-    <updated>2025-05-06T00:00:00+00:00</updated>
+    <updated>2026-03-14T00:00:00+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale>
@@ -42,17 +46,22 @@
       <label form="short" prefix=" "/>
     </names>
   </macro>
+  <macro name="consec-periodical-issuance">
+    <date variable="issued">
+      <date-part name="year"/>
+    </date>
+  </macro>
+  <macro name="periodical-issuance">
+    <date variable="issued">
+      <date-part name="month" form="short" suffix=" "/>
+      <date-part name="day" suffix=", "/>
+      <date-part name="year"/>
+    </date>
+  </macro>
   <macro name="author-short">
     <choose>
       <if type="legal_case">
-        <choose>
-          <if variable="title-short">
-            <text macro="name-short-macro" font-style="italic"/>
-          </if>
-          <else>
-            <text macro="name-short-macro"/>
-          </else>
-        </choose>
+        <text macro="name-short-macro" font-style="italic"/>
       </if>
       <else-if type="bill legislation" match="any">
         <text macro="name-macro"/>
@@ -75,17 +84,12 @@
   </macro>
   <macro name="author">
     <choose>
-      <if type="legal_case legislation" match="any">
-        <!-- legal cases do not display the author at the beginning of the line -->
-      </if>
-      <else-if type="bill" match="any">
-        <text macro="name-macro"/>
-      </else-if>
+      <if type="bill legal_case legislation" match="any"/>
       <else-if type="book graphic motion_picture report song" match="any">
-        <text macro="name-macro" font-variant="small-caps"/>
+        <text macro="name-macro" font-variant="small-caps" suffix=", "/>
       </else-if>
       <else>
-        <text macro="name-macro"/>
+        <text macro="name-macro" suffix=", "/>
       </else>
     </choose>
   </macro>
@@ -97,24 +101,18 @@
   </macro>
   <macro name="access">
     <choose>
-      <if type="legal_case legislation" match="any">
-        <!-- legal case citations generally do not use URL's -->
-      </if>
+      <if type="legal_case bill legislation article-journal" match="any"/>
       <else-if variable="URL">
-        <group delimiter=" ">
+        <group prefix=", " delimiter=" ">
           <text variable="URL"/>
-          <choose>
-            <if variable="issued" match="none">
-              <group delimiter=" " prefix="(" suffix=")">
-                <text value="last visited"/>
-                <date variable="accessed">
-                  <date-part name="month" form="short" suffix=" "/>
-                  <date-part name="day" suffix=", "/>
-                  <date-part name="year"/>
-                </date>
-              </group>
-            </if>
-          </choose>
+          <group delimiter=" " prefix="(" suffix=")">
+            <text value="last visited"/>
+            <date variable="accessed">
+              <date-part name="month" form="short" suffix=" " strip-periods="true"/>
+              <date-part name="day" suffix=", "/>
+              <date-part name="year"/>
+            </date>
+          </group>
         </group>
       </else-if>
     </choose>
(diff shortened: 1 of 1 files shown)


bluebook-law-review.csl
```diff
@@ -130,11 +128,59 @@
               <text macro="container"/>
               <text variable="page-first"/>
             </group>
-            <text variable="locator"/>
+            <choose>
```
adam3smith: revert this -- no need to test for a variable before calling it: an empty variable will just not render, same as your test

bluebook-law-review.csl
```diff
@@ -130,11 +128,59 @@
...
+        <text macro="statute-citation"/>
+      </else-if>
+      <else-if type="bill">
+        <group delimiter=", ">
+          <text variable="title"/>
+          <group delimiter=" ">
+            <text value="Pub. L. No."/>
+            <text variable="number"/>
+          </group>
+          <group prefix="§ ">
```
adam3smith: No need for a group around a single element -- remove the group, move the prefix directly to section

bluebook-law-review.csl
```diff
@@ -130,11 +128,59 @@
...
+              <text variable="page-first"/>
+              <text variable="locator"/>
+            </group>
+          </group>
+        </group>
+        <text macro="consec-periodical-issuance" prefix=" (" suffix=")"/>
+      </else-if>
+      <else-if type="book">
+        <text variable="title" text-case="title" font-variant="small-caps"/>
+        <text variable="locator" prefix=" "/>
```
adam3smith: BB would never put a locator label (like col., ch. etc.) in there? Locators are most commonly pages, but can be all sort of other pinpoints.

bluebook-law-review.csl
```diff
@@ -130,11 +128,59 @@
...
         </group>
       </if>
+      <else-if type="legislation" match="any">
+        <text macro="statute-citation"/>
+      </else-if>
+      <else-if type="bill">
+        <group delimiter=", ">
+          <text variable="title"/>
+          <group delimiter=" ">
+            <text value="Pub. L. No."/>
```
adam3smith: Won't this always put the Public Law Number on bills? My understanding would be that more often that not bills are cited without it and it gets included only when the bill has passed (but I could be wrong here)

bluebook-law-review.csl
```diff
@@ -144,59 +190,31 @@
...
-              <group delimiter=" ">
-                <text value="§"/>
-                <text variable="section"/>
-              </group>
-            </group>
-          </group>
-        </group>
-      </else-if>
-      <else-if type="article-newspaper article-magazine" match="any">
+      <else-if type="article-newspaper article-magazine thesis" match="any">
```
adam3smith: why are you adding thesis here? That seems like a very awkward fit with the periodical item types

bluebook-law-review.csl
```diff
@@ -144,59 +190,31 @@
...
-            <text macro="container"/>
-          </group>
-          <text macro="issuance"/>
+          <text macro="container"/>
+          <text macro="periodical-issuance"/>
           <group delimiter=" ">
             <text value="at"/>
             <text variable="page-first"/>
           </group>
+          <choose>
```
adam3smith: Same as above, this test isn't necessary.

bluebook-law-review.csl
```diff
@@ -207,17 +225,36 @@
...
-        <text variable="locator" prefix=" "/>
-        <text macro="issuance" prefix=" (" suffix=")"/>
+      <else-if type="webpage post-weblog" match="any">
+        <group delimiter=", ">
+          <text variable="title" text-case="title" font-style="italic"/>
+          <text variable="container-title"/>
+          <text macro="periodical-issuance"/>
+        </group>
+      </else-if>
+      <else-if type="report">
```
adam3smith, line 260: What exactly are you envisioning this for? Reports in Zotero/CSL are typically reports by government non/government organizations, very much not legal reporters

bluebook-law-review.csl
```diff
@@ -328,24 +398,44 @@
...
-                <group delimiter=" ">
-                  <text variable="volume"/>
-                  <text variable="container-title" form="short"/>
-                  <text macro="at_page"/>
+                <text macro="author-short"/>
+                <group>
+                  <text value="supra" font-style="italic" suffix=" "/>
+                  <text value="note" suffix=" "/>
+                  <text variable="first-reference-note-number"/>
+                  <group>
```
adam3smith: Same as above, this entire group isn't needed.

bluebook-law-review.csl
```diff
@@ -360,17 +450,38 @@
...
-                  <text value="note"/>
-                  <text variable="first-reference-note-number"/>
-                  <text macro="at_page"/>
+                  <choose>
+                    <if type="legal_case" match="any">
+                      <group>
+                        <text variable="volume" suffix=" "/>
+                        <text variable="container-title"/>
+                        <group>
+                          <text value="at" prefix=" " suffix=" "/>
```
adam3smith: We prefer using terms where available, so 
`<text term="at".../>` 
with everything else the same.

bluebook-law-review.csl
```diff
@@ -328,24 +398,44 @@
...
               <group delimiter=", ">
-                <text variable="title" form="short" font-style="italic"/>
-                <group delimiter=" ">
-                  <text variable="volume"/>
-                  <text variable="container-title" form="short"/>
-                  <text macro="at_page"/>
+                <text macro="author-short"/>
+                <group>
+                  <text value="supra" font-style="italic" suffix=" "/>
+                  <text value="note" suffix=" "/>
```
adam3smith: Also for note -- please use the term rather than the raw text value form.

adam3smith (changes requested):
This is great! -- I have a couple of questions that might affect actual output and a couple of requests to improve/simplify the code.

For reference for others and posterity: https://papers.ssrn.com/sol3/papers.cfm?abstract_id=6416679

Since you already have reference examples in the library, would you consider putting them into a Zotero group? We have a group where we have APA and Chicago Manual examples, it'd make sense to have BB in there, too.

bluebook-law-review.csl
jamesftierney: Thanks — I noted a couple other places where a term could be swapped in for a text value.

bluebook-law-review.csl
jamesftierney: This is correct. As we only want Pub. L. No. for session laws, that can be printed conditional on the number variable being populated (i.e., enter the session law number only if you want to cite the session law). If the number variable is absent, the next update will use the codified citation instead.

bluebook-law-review.csl
jamesftierney: Next version will implement this.

bluebook-law-review.csl
jamesftierney: Your objection is correct, and the next version adds a new version that correctly implements thesis in Bluebook form:
> Author, Title (year) (Type, Institution)

bluebook-law-review.csl
jamesftierney: This objection is also correct. The current report block was targeting C.F.R.-style regulatory citations — it uses § number prefix and (publisher year) with only a year. But that's wrong for most things report would actually hold in Zotero. While report is not a legal reporter, it is for government/institutional publications — working papers, agency reports, technical reports. That actually makes it a reasonable fit for Federal Register entries and similar administrative law materials, since those ARE government publications. There's no better CSL type for them. As the existing implementation was wrong for this use case, the next version keeps report but repurposes it specifically for a subset of administrative law materials since it's the best-fit CSL type. Any true C.F.R.-style citation that was previously using report should instead use legislation under the next version.

jamesftierney:
Per reviewer's request, added a public Zotero group library that shows conventions for data-entry here: https://www.zotero.org/groups/6479791/bluebook_law_review_style_sample_citations

jamesftierney:
Just checking on the status of this pull request in response to reviewer's requests for updates?

bluebook-law-review.csl
shelbaum, line 115: The currently active version of this style has \<choose\> tags around this block, making sure that the last visited date parenthetical only shows up if there's no other publication date provided, per Indigo Book R33.5 and Bluebook R18.2.2(c). Could that be restored to this version? As is I think this would incorrectly include (last visited) any time there is a URL.  
(PS, I just came across your paper documenting these updates.  Extremely useful, many thanks for improving the style!)

PR #8115, opened 2026-03-24
Add oscola-5.csl: Oxford Standard for Citation of Legal Authorities, 5th edition

veale:
Adds a new style for [OSCOLA 5 (2026)](https://www.law.ox.ac.uk/sites/default/files/2026-03/OSCOLA%205.pdf). Key changes from oscola.csl:
- Ibid abolished: all repeated citations use short-form author + note number
- Thesis titles italicised
- Newspaper articles: city of publication suppressed
- Podcasts/broadcasts: series name inside date parentheses
- Social media posts
- No trailing full stop on footnote citations

Author: Michael Veale. Original oscola.csl by Sebastian Karcher and Andrew Dunning (CC BY-SA 3.0).


StianOby:
@adunning: this should probably be included in the style-variant-builder repo?

POBrien333:
From what I understand, Andrew is already working on it: https://forums.zotero.org/discussion/130498/oscola-march-2026#latest

oscola-4th-edition.csl
```diff
@@ -0,0 +1,760 @@
...
+        <if is-numeric="volume">
+          <label form="short" strip-periods="true" variable="volume"/>
+        </if>
+      </choose>
+      <text variable="volume"/>
+    </group>
+  </macro>
+  <!--Authors and Persons-->
+  <macro name="author-note">
+    <!--for bills & hearing this should start with jurisdiction once available-->
```
adunning, line 90: There is now a `jurisdiction` variable.

adunning:
I have been looking at this: I'm working on an embedded version of OSCOLA in MHRA and New Hart's Rules (as Bluebook is with APA and Chicago). I won't be able to work on it seriously until the summer, so in the meantime it's very helpful to have others establishing a baseline.

oscola-5.csl
```diff
@@ -0,0 +1,760 @@
...
+  <!--
+    Updated to OSCOLA 5th edition by Michael Veale.
+    Original OSCOLA 4 style by Sebastian Karcher and Andrew Dunning.
+    Licensed under Creative Commons Attribution-ShareAlike 3.0.
+  -->
+  <info>
+    <title>OSCOLA 5th edition</title>
+    <title-short>Oxford University Standard for Citation of Legal Authorities</title-short>
+    <id>http://www.zotero.org/styles/oscola-5</id>
+    <link href="http://www.zotero.org/styles/oscola-5" rel="self"/>
```
adunning: This should replace the current `oscola.csl` rather than being added as a new file. The old files should be moved to `oscola-4th-edition.csl` and `oscola-no-ibid-4th-edition.csl`. I think `oscola-journal-abbreviations.csl` can be deleted and redirected in `renamed-styles.json`.

oscola-4th-edition.csl
```diff
@@ -0,0 +1,760 @@
...
+              </group>
+            </if>
+          </choose>
+          <group delimiter=" ">
+            <text variable="volume"/>
+            <choose>
+              <if type="legal_case" variable="container-title">
+                <text form="short" strip-periods="true" variable="container-title"/>
+              </if>
+              <else-if match="any" type="article-magazine article-newspaper">
```
adunning, line 438: To cover OSCOLA journal abbreviations, the old edition had a separate `oscola-journal-abbreviations.csl`, but I think this can be added by default in the new edition. I believe that adding this simplified condition would have the same effect:

```xml
              <else-if type="article-journal">
                <text form="short" strip-periods="true" variable="container-title"/>
              </else-if>
```

oscola-4th-edition.csl
veale: Ah, I didn't know this — this was a note leftover from the previous edition CSL file. I will update it accordingly.

veale:
> I have been looking at this: I'm working on an embedded version of OSCOLA in MHRA and New Hart's Rules (as Bluebook is with APA and Chicago). I won't be able to work on it seriously until the summer, so in the meantime it's very helpful to have others establishing a baseline.

This is interesting indeed - I am not sure however that MHRA or New Hart's Rules recommendations to use Oscola to cite legal texts will work so well with CSL given that a lot of what Oscola requires cannot be implemented in CSL at all because it requires checks of the content of the variable fields (cf Juris-M and CSL-M). I've been supporting an update of the biblatex OSCOLA package as well as making these edits, where the required logic is implemented, but as far as I understand it's not really possible in CSL, sadly.

In any case @adunning, I have made the changes in your comments! Thanks!

StianOby:
@veale : see the discussion here: https://github.com/citation-style-language/style-variant-builder/issues/10

Could you make the necessary adjustments to your PR so that DOIs are not outputted for e.g. books and journal articles? For the very rare occasions of e.g. books without ISBN it is possible for users to input a DOI in the URL field (or manually add it as a citation suffix).

Any amendment you make of this kind should be applied not only to the 5th edition, but the 4th edition as well.

adunning:
If you think this needs to be done retroactively for the 4th edition, I've proposed something in <https://github.com/citation-style-language/styles/pull/8147> – let's focus this discussion on OSCOLA 5.

alfonsoss:
I have a working file which also supports WTO, and GATT documents, actually based other changes on the edits in `oscola.csl`. What would be the best way to get them in this PR? Or should I open a PR to be merged on top of this one?

Example of changes I have:
Data entry for WTO/GATT, uses legal_case with:

- genre = document description, 
- authority = WTO/GATT, 
- number = "Doc No WT/DS161/AB/R (circulated 11 December 2000, adopted 10 January 2001)" 
  - (both dates encoded in the number field since CSL can only hold one date), 
- issued left blank
  - (year bracket auto-suppressed), 
- and locator type set to paragraph for [n]-format pinpoints.



This gives: *Korea – Measures...* (Appellate Body Report) WTO Doc No WT/DS161/AB/R (circulated 11 December 2000, adopted 10 January 2001) [161] — which matches the 4.3.4 format in the guide (and what we needed initially).

StianOby:
@alfonsoss : CSL can handle multiple dates, using the extra field. The "available-date" variable seems appropriate for "circulated".

From the [CSL spec:](https://docs.citationstyles.org/en/stable/specification.html) 

> available-date
>     Date the item was initially available (e.g. the online publication date of a journal article before its formal publication date; the date a treaty was made available for signing)

alfonsoss:
I will test tomorrow if the variable works for it then.

MikeN123:
While looking at this new style, I noticed that `container-title` is now added at the start of `legislation` (in `author-note`), since the `jurisdiction` changes. I'm not entirely sure how that fits in the OSCOLA standard? In par. 2.6.2. of OSCOLA jurisdiction is discussed as being added in brackets at the end.

There is also a remark on EU cases, noting that `volume`, `container-title` and `page` should be blank. But I think `authority` also needs to be blank, may be good to note that. Additionally, I think this still doesn't give a good way to add the case number at the start (as expected in OSCOLA), but that was already the case for OSCOLA 4.

[edit]
Some other things I noted while looking at OSCOLA 5, which could be of use to anyone working on this PR:
- It uses journal abbreviations (instead of letter the user choose)
- For EU cases it also uses square brackets for the paragraph numbers (previously it would ask for 'para'/'paras', but the CSL style has always been somewhat inconsistent, using square brackets for the initial citation, but para/paras for subsequent citations)

veale:
> @veale : see the discussion here: [citation-style-language/style-variant-builder#10](https://github.com/citation-style-language/style-variant-builder/issues/10)
> 
> Could you make the necessary adjustments to your PR so that DOIs are not outputted for e.g. books and journal articles? For the very rare occasions of e.g. books without ISBN it is possible for users to input a DOI in the URL field (or manually add it as a citation suffix).
> 
> Any amendment you make of this kind should be applied not only to the 5th edition, but the 4th edition as well.

Hello,
Apologies — term led to this being on the back burner.

OSCOLA is divided on this when put together with the limited functionality in Zotero/CSL. The rule in OSCOLA 5th around secondary sources is pragmatic, like the citation style: 'Always provide sufficient information to enable the reader to identify the secondary source'. In one reading, this means that we need to help users provide everything that they need. In another reading, one could focus on the reading of being parsimonious.

As one of the educational leaders of a major law school in the UK using OSCOLA, we have a policy of maximising links in light of generative AI and the need for source checking. This is a trend I have seen in journals and academic work more broadly.

OSCOLA 5th is divided on links for physical items. While it does not mention them in the cases of journals, it indicates that chapters of online books, such as some encyclopaedias, should be rendered as such:

Leslie Green, ‘Legal Positivism’, *The Stanford Encyclopedia of Philosophy* (Fall edn, 2009)
<https://perma.cc/9L4R-GYHX>

Journals are increasingly dropping any physical copies and this is important context.

I would be in favour of retaining maximal DOI support and having those who want minimalist citations not collect DOI information in Zotero, rather than have those who want it to add it manually to every suffix of every electronic-only journal. However I think this is a matter of usability of the style and the limitations of CSL. Therefore, I suggest, like the previous (ibid/no ibid) version, we have OSCOLA and OSCOLA (including DOIs). It is unusual, sure — but it is also extremely unusual for any style in 2026 to omit DOIs given the state of academic misconduct today.

That way, people of all persuasions can easily use Zotero.

veale:
> While looking at this new style, I noticed that `container-title` is now added at the start of `legislation` (in `author-note`), since the `jurisdiction` changes. I'm not entirely sure how that fits in the OSCOLA standard? In par. 2.6.2. of OSCOLA jurisdiction is discussed as being added in brackets at the end.

I have changed this too, in both my suggested files.

My suggested changes are in the PR now. GitHub's actions/checkout action looks like it has recently started refusing to pull contributor branches into the repo's own trusted CI environment, to avoid letting reqestors' code run with the repo's secrets, so the checks are not working -- i think that's what's going on at least.

PR #8131, opened 2026-04-02
SBL: volume-title variable support and revised handling of multivolume works

mbquintana:
### Description
Revisions made to several macros (collection-title, collection-title-note, container-title, locators-note, locators, pages) and <citation> / <bibliography> in order to accommodate the volume-title variable and allow for proper handling of multivolume works and chapters of multivolume works (see SBLHS 6.2.20–23), which previously were either impossible to produce correctly or required extensive work-arounds.

### Checklist

### DIFF
--- a/society-of-biblical-literature-fullnote-bibliography.csl
+++ b/society-of-biblical-literature-fullnote-bibliography.csl
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="chicago" default-locale="en-US">
   <info>
-    <title>SBL Handbook of Style 2nd edition (notes)</title>
-    <title-short>Society of Biblical Literature (SBLHS)</title-short>
-    <id>http://www.zotero.org/styles/society-of-biblical-literature-fullnote-bibliography</id>
+    <title>SBL Handbook of Style 2nd edition (notes)-xyz</title>
+    <title-short>Society of Biblical Literature (SBLHS)-xyz</title-short>
+    <id>http://www.zotero.org/styles/society-of-biblical-literature-fullnote-bibliography-xyz</id>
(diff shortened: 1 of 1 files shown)


adunning:
This is a good start, but you should be using group delimiters instead of prefixes and suffixes, and `<label>` elements rather than `<text>` when accompanying a variable. Printing volume information shouldn't rely on `collection-title`. You may also want to use `part-number` and `part-title`. For a more robust implementation, see my approach to declaring volume titles for Chicago, which is published in the author–date style (where it is a requirement rather than an option), but you can see how it would work for notes styles if you look at the `specific-title-first` macros, such as at <https://github.com/citation-style-language/style-variant-builder/blob/aaf19bf5a0bfa33423a7aa53a2c28bb8d2acfb65/templates/chicago-template.csl#L3485-L3536> and <https://github.com/citation-style-language/style-variant-builder/blob/aaf19bf5a0bfa33423a7aa53a2c28bb8d2acfb65/templates/chicago-template.csl#L1685-L1703>.

mbquintana:
@adunning I've made some edits addressing the issues with group delimiters and <label> elements. Let me know if there are any further changes needed here.

~~After testing the revised style against a representative set of real Zotero items, I decided not to make further changes to the collection-title conditions. Although the general caution against relying on collection-title for printing volume information is reasonable in principle, the current collection-title logic was not producing any actual errors in this set of examples, so I didn't want to risk disturbing output that was already correct by making changes—especially in multivolume chapter citations.~~

I also chose not to add support for part-number and part-title at this stage. The current coding produces the desired results for all the examples I could come up with on the basis of SBLHS by using only the volume, volume-title, number-of-volumes, and related fields, so integrating the part variables would have added complexity without solving a real problem demanded by SBLHS requirements. For now, at least, I judged it better to keep the implementation narrower and avoid introducing additional logic that doesn't seem relevant for SBLHS.

**EDIT:** I realized I did not interpret your comment "Printing volume information shouldn't rely on collection-title" correctly. I have made a few changes to the code that take care of this under `point-locators-subsequent`.

PR #8136, opened 2026-04-06
Update nature.csl (line spacing)

oncebo:
Change line spacing to 1 to match the real Nature citation style.

Improve numeric formatting in the bibliography to make it more compatible with Word processors, avoiding unwanted line breaks after the reference number.

### Description
(Briefly describe the style or changes you're proposing.)

### Checklist

### DIFF
--- a/nature.csl
+++ b/nature.csl
@@ -166,9 +166,9 @@
       <text variable="citation-number"/>
     </layout>
   </citation>
-  <bibliography et-al-min="6" et-al-use-first="1" second-field-align="flush" entry-spacing="0" line-spacing="2">
+  <bibliography et-al-min="6" et-al-use-first="1" second-field-align="flush" entry-spacing="0" line-spacing="1">
     <layout suffix=".">
-      <text variable="citation-number" suffix="."/>
+      <text variable="citation-number" suffix=". "/>
       <group delimiter=" ">
         <text macro="author" suffix="."/>
         <text macro="title" suffix="."/>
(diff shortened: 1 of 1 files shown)


POBrien333:
Thank you for your PR.
Can you explain why you suggest this change?
In the guidelines I see "Contributions should be double-spaced and written in English (spellings as in the [Oxford English Dictionary](http://www.oed.com/))."

oncebo:
Thank you.
I checked the published Nature papers, and the line spacing used there is not double‑spaced, so I changed it to match the actual format.
The other extra blank in the numeric format is an unknown bug in some text editors, which causes an unintended line break in the bibliography information. Adding an extra blank can help to fix this issue.

POBrien333:
I understand. Many publisher request a certain style though and then do the typesetting themselves. So, unless an editor at Nature has requested this I am not sure we'd accept this change.

subhramit:
Hi, is the double-space really intended?

> The other extra blank in the numeric format is an unknown bug in some text editors, which causes an unintended line break in the bibliography information. Adding an extra blank can help to fix this issue.

Shouldn't the text editors be the ones to fix that bug then? Is it right to change the CSL specification based on that?

cc @POBrien333

pluto-han:
> Hi, is the double-space really intended?
> 
> > The other extra blank in the numeric format is an unknown bug in some text editors, which causes an unintended line break in the bibliography information. Adding an extra blank can help to fix this issue.
> 
> Shouldn't the text editors be the ones to fix that bug then? Is it right to change the CSL specification based on that?
> 
> cc @POBrien333

+1 
On behalf of @JabRef team

cc @POBrien333 @adam3smith

PR #8155, opened 2026-04-21
Create vegetation-ecology-and-diversity.csl

LucianoDeBenedictis:
### Description
I have made a CSL for the journal "Vegetation Ecology and Diversity".

### Checklist

### DIFF
--- a/vegetation-ecology-and-diversity.csl
+++ b/vegetation-ecology-and-diversity.csl
@@ -1,0 +1,250 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <!-- This style was edited with the Visual CSL Editor (https://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>Vegetation Ecology and Diversity</title>
+    <id>http://www.zotero.org/styles/vegetation-ecology-and-diversity</id>
+    <link href="http://www.zotero.org/styles/vegetation-ecology-and-diversity" rel="self"/>
+    <link href="https://www.zotero.org/styles/pensoft-journals" rel="template"/>
+    <link href="https://zookeys.pensoft.net/about#CitationsandReferences" rel="documentation"/>
+    <author>
+      <name>Luciano Ludovico Maria De Benedictis</name>
+      <email>luciano.debenedictis@unicam.it</email>
+    </author>
+    <contributor>
+      <name>Brian Stucky</name>
+      <email>stuckyb@colorado.edu</email>
+    </contributor>
+    <contributor>
+      <name>Teodor Georgiev</name>
+      <email>t.georgiev@pensoft.net</email>
+    </contributor>
+    <category citation-format="author-date"/>
+    <issn>3033-1447</issn>
+    <eissn>3033-1447</eissn>
+    <issnl>3033-1447</issnl>
+    <summary>Zotero style for VED</summary>
+    <updated>2026-04-21T09:11:25+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="en-US">
+    <date form="text">
+      <date-part name="month" suffix=" "/>
+      <date-part name="day" suffix=", "/>
+      <date-part name="year"/>
+    </date>
+    <terms>
+      <term name="editor" form="short">
+        <single>ed.</single>
+        <multiple>eds</multiple>
+      </term>
+      <term name="et-al">…</term>
+    </terms>
+  </locale>
+  <macro name="editor">
+    <names variable="editor" delimiter=", ">
+      <name initialize-with="" name-as-sort-order="all" sort-separator=" "/>
+      <label form="short" prefix=" (" text-case="capitalize-first" suffix=")"/>
+    </names>
+  </macro>
+  <macro name="anon">
+    <text term="anonymous" form="short" text-case="capitalize-first" strip-periods="true"/>
+  </macro>
+  <macro name="author">
+    <names variable="author">
+      <name delimiter-precedes-last="never" initialize-with="" name-as-sort-order="all" sort-separator=" "/>
+      <et-al font-style="italic"/>
+      <label form="short" prefix=" (" text-case="capitalize-first" suffix=")"/>
+      <substitute>
+        <names variable="editor"/>
+        <text macro="anon"/>
+      </substitute>
+    </names>
+  </macro>
+  <macro name="author-short">
+    <names variable="author">
+      <name form="short" delimiter=" " and="text" delimiter-precedes-last="never" initialize-with=". "/>
+      <substitute>
+        <names variable="editor"/>
+        <names variable="translator"/>
+        <text macro="anon"/>
+      </substitute>
+    </names>
+  </macro>
+  <macro name="authorcount">
+    <names variable="author">
+      <name form="count"/>
+    </names>
+  </macro>
+  <macro name="access">
+    <choose>
+      <if type="legal_case" match="none">
+        <choose>
+          <if variable="DOI">
+            <group delimiter=" ">
+              <text variable="DOI" prefix="https://doi.org/"/>
+            </group>
+          </if>
+          <else-if variable="URL">
+            <group delimiter=" " suffix=".">
+              <text variable="URL" prefix="Available from: "/>
+              <group prefix="(" suffix=")">
+                <date variable="accessed" form="text"/>
+              </group>
+            </group>
+          </else-if>
+        </choose>
+      </if>
+    </choose>
+  </macro>
+  <macro name="title">
+    <text variable="title"/>
+  </macro>
+  <macro name="legal_case">
+    <group prefix=" " delimiter=" ">
+      <text variable="volume"/>
+      <text variable="container-title"/>
+    </group>
+    <text variable="authority" prefix=" (" suffix=")"/>
+  </macro>
+  <macro name="publisher">
+    <choose>
+      <if type="thesis" match="none">
+        <group delimiter=", ">
+          <text variable="publisher"/>
+          <text variable="publisher-place"/>
+        </group>
+        <text variable="genre" prefix=". "/>
+      </if>
+      <else>
+        <group delimiter=". ">
+          <text variable="genre"/>
+          <text variable="publisher"/>
+        </group>
+      </else>
+    </choose>
+  </macro>
+  <macro name="year-date">
+    <choose>
+      <if variable="issued">
+        <group>
+          <date variable="issued">
+            <date-part name="year"/>
+          </date>
+        </group>
+      </if>
+      <else>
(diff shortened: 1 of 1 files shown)


LucianoDeBenedictis:
Sorry, but this error message is a bit terse, and it seems just a matter of escaping characters. <https://www.zotero.org/styles/pensoft-journals> exists and is linked in the same form by other CSL of other journals.

POBrien333:
you made it into "https"

LucianoDeBenedictis:
> you made it into "https"

Well, that fixed it, thanks. Seems restrictive, since https is allowed for the documentation field, but I guess it doesn't matter, since most modern browsers should convert it to https.

Let me fix the inline citations: "..." should be used in the references but not in the citations instead of "et al."

LucianoDeBenedictis:
There. I ended up using the default behaviour of the "and others" term, which seems to correctly render as ellipses in all editors I have tried (Zotero, visual CSL editor, the bot here). It doesn't seem like the terms specification is flexible enough to define something custom more explicitly (to ensure that it always gets rendered as "...") so this will have to do.

vegetation-ecology-and-diversity.csl
```diff
@@ -0,0 +1,248 @@
...
+    <title>Vegetation Ecology and Diversity</title>
+    <id>http://www.zotero.org/styles/vegetation-ecology-and-diversity</id>
+    <link href="http://www.zotero.org/styles/vegetation-ecology-and-diversity" rel="self"/>
+    <link href="http://www.zotero.org/styles/pensoft-journals" rel="template"/>
+    <link href="https://ved.arphahub.com/about#CitationsandReferences" rel="documentation"/>
+    <author>
+      <name>Luciano Ludovico Maria De Benedictis</name>
+      <email>luciano.debenedictis@unicam.it</email>
+    </author>
+    <contributor>
```
POBrien333, line 15: I assume these are remnants from the template style. Remove if so

vegetation-ecology-and-diversity.csl
```diff
@@ -0,0 +1,248 @@
...
+      <name>Teodor Georgiev</name>
+      <email>t.georgiev@pensoft.net</email>
+    </contributor>
+    <category citation-format="author-date"/>
+    <issn>3033-1447</issn>
+    <summary>Zotero style for VED</summary>
+    <updated>2026-04-21T09:11:25+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  </info>
+  <locale xml:lang="en-US">
```
POBrien333: Remove the `-US`

vegetation-ecology-and-diversity.csl
```diff
@@ -0,0 +1,248 @@
...
+    <names variable="author">
+      <name form="count"/>
+    </names>
+  </macro>
+  <macro name="access">
+    <choose>
+      <if type="legal_case" match="none">
+        <choose>
+          <if variable="DOI">
+            <group delimiter=" ">
```
POBrien333, line 82: THis group does nothing. remove

vegetation-ecology-and-diversity.csl
```diff
@@ -0,0 +1,248 @@
...
+              <text variable="DOI" prefix="https://doi.org/"/>
+            </group>
+          </if>
+          <else-if variable="URL">
+            <group delimiter=" " suffix=".">
+              <text variable="URL" prefix="Available from: "/>
```
POBrien333, line 88: Replace such hardcoded text elements with the appropriate term

vegetation-ecology-and-diversity.csl
```diff
@@ -0,0 +1,248 @@
...
+          <text variable="genre"/>
+          <text variable="publisher"/>
+        </group>
+      </else>
+    </choose>
+  </macro>
+  <macro name="year-date">
+    <choose>
+      <if variable="issued">
+        <group>
```
POBrien333, line 128: This group does nothing. Please check the rest of your style for single-element groups and remove them.

vegetation-ecology-and-diversity.csl
```diff
@@ -0,0 +1,248 @@
...
+      <key variable="title"/>
+    </sort>
+    <layout suffix=" ">
+      <text macro="author" suffix=" "/>
+      <date variable="issued" prefix="(" suffix=")">
+        <date-part name="year"/>
+      </date>
+      <choose>
+        <if type="book" match="any">
+          <text macro="legal_case"/>
```
POBrien333, line 194: What does this do?

vegetation-ecology-and-diversity.csl
```diff
@@ -0,0 +1,248 @@
...
+          <group prefix=" " delimiter=" ">
+            <text macro="title" font-style="normal" suffix="."/>
```
POBrien333, line 196: remove the "normal"

vegetation-ecology-and-diversity.csl
```diff
@@ -0,0 +1,248 @@
...
+            <text macro="edition"/>
+            <text macro="editor" suffix="."/>
+          </group>
+          <group prefix=" " suffix="." delimiter=", ">
+            <text macro="publisher"/>
+            <text variable="number-of-pages" prefix=" " suffix=" pp"/>
```
POBrien333, line 202: replace the hardcoded text with the appropriate label. Check the rest of your style for such issues.

POBrien333 (changes requested):

vegetation-ecology-and-diversity.csl
LucianoDeBenedictis: I am the author. I reported the authors of the template style as contributors. Is this correct?

vegetation-ecology-and-diversity.csl
LucianoDeBenedictis: done for the next commit

vegetation-ecology-and-diversity.csl
LucianoDeBenedictis: For this and the following: these are all present in the [template Pensoft style](https://github.com/citation-style-language/styles/blob/master/pensoft-journals.csl), to which I made minimal edits, mostly to correctly use "..." for more than 6 authors in the bibliography. I will trust you and remove these lines if unnecessary, but then that should be done for the Pensoft style as well (which most journals of the publisher point to). I am not affiliated with the publisher so I won't correct that style myself, focusing only on this one.

POBrien333: Right, but we'd want this fixed.
In the future somebody will use this style as a template and then we need to fix it in that one.

I also don't know how old that style is, but we've gotten a bit more stringent with style quality.

LucianoDeBenedictis: Right, agreed. Maybe this could become the template for the updated version of that style! I will do what I can.

vegetation-ecology-and-diversity.csl
LucianoDeBenedictis: I'm not sure what to replace them with. Do you have any pointers? The [CSL 1.0.2 Specification](https://docs.citationstyles.org/en/stable/specification.html#group) shows a similar hardcoded solution befor the URL, just within a group.

vegetation-ecology-and-diversity.csl
```diff
@@ -0,0 +1,248 @@
...
+  <macro name="access">
+    <choose>
+      <if type="legal_case" match="none">
+        <choose>
+          <if variable="DOI">
+            <group delimiter=" ">
+              <text variable="DOI" prefix="https://doi.org/"/>
+            </group>
+          </if>
+          <else-if variable="URL">
```
LucianoDeBenedictis, line 86: @POBrien333 If I remove the group above (line 82), does it mean I can remove the if wrapping it? Then what would happen to this else-if?

vegetation-ecology-and-diversity.csl
LucianoDeBenedictis: As above, does this mean keeping the content inside the group, or not? If the latter, what happens to the if/else and choose statements?

vegetation-ecology-and-diversity.csl
LucianoDeBenedictis: I don't know any better than you do. Should I just remove this "legal_case" line?

vegetation-ecology-and-diversity.csl
LucianoDeBenedictis: I think in this case it should be something like <text variable="number-of-pages" prefix=" form="short"/> leaving the default plural="contextual". Not sure if prefix is necessary. I will do the same at line 228. Anything else of this sort?

vegetation-ecology-and-diversity.csl
POBrien333: No, why would you remove the conditional?
A group makes sense for more than one element. A group of one is not a group.

LucianoDeBenedictis: I understand this a bit better now. So it means stripping the group but keeping the statement inside, which is equivalent since the group is doing nothing, but `<text>` is.

vegetation-ecology-and-diversity.csl
```diff
@@ -0,0 +1,248 @@
...
+        </if>
+        <else-if type="chapter paper-conference" match="any">
+          <text macro="title" prefix=" " suffix="."/>
+          <group prefix=" In: " delimiter=" ">
+            <text macro="editor" suffix=","/>
+            <text variable="container-title" suffix="."/>
+            <text variable="collection-title" suffix="."/>
+            <group suffix=".">
+              <text macro="publisher"/>
+              <group delimiter=" " prefix=", " suffix=".">
```
LucianoDeBenedictis, line 213: this group is single element, but has some additional variables. move them to the text block?

vegetation-ecology-and-diversity.csl
```diff
@@ -0,0 +1,248 @@
...
+        </else-if>
+        <else>
+          <group prefix=" " delimiter=". " suffix=".">
+            <text macro="title"/>
+            <text macro="editor"/>
+          </group>
+          <group prefix=" " suffix=".">
+            <text variable="container-title"/>
+            <group prefix=" ">
+              <text variable="volume"/>
```
LucianoDeBenedictis, line 239: move prefix inside `<text>` and remove group?

vegetation-ecology-and-diversity.csl
POBrien333: As I say, if they're from the template, remove them.

vegetation-ecology-and-diversity.csl
POBrien333: No, it doesn't show that. It uses a `term`

vegetation-ecology-and-diversity.csl
LucianoDeBenedictis: ok, I will remove contributors and leave what's before, because those other things are updated. I appreciate your assistance, if you want to be listed as contributor or acknowledged in some other way let me know under which credentials.

vegetation-ecology-and-diversity.csl
LucianoDeBenedictis: ok. The closest terms are either "available at" or "retrieved". Should I use either of those, possibly with the appropriate suffix, in another `<text>` before the URL line inside the group, even if it might result in a slightly different wording?

PR #8159, opened 2026-04-28
Update suomen-laakarilehti.csl (documentation link)

shiyuwang-jamk:
### Description

The old documentation link to the Journal's style guide is no longer accessible.  I found a working link in Finnish.


POBrien333:
While we don't typically update such minor issues, I've accepted the PR.
We'd only really update something like this if there is an actual change to the style needed. Thanks :)

PR #8163, opened 2026-05-03
Update rmit-university-harvard.csl

nguynking:
### Description
Updates `rmit-university-harvard.csl` to better match the current RMIT Easy Cite Harvard guidance.

This revises formatting for journal articles, webpages, newspaper and magazine articles, books and chapters, conference papers, reports, and datasets. It also updates item-type-specific author/date handling, DOI/report number output, access date/URL/database formatting, no-author substitutions, and style metadata with contributor information.

### Checklist

### DIFF
--- a/rmit-university-harvard.csl
+++ b/rmit-university-harvard.csl
@@ -138,6 +138,21 @@
       </substitute>
     </names>
   </macro>
+  <macro name="author-short-newspaper">
+    <choose>
+      <if variable="author">
+        <names variable="author">
+          <name form="short" and="text" delimiter=", " delimiter-precedes-last="never" initialize-with=". "/>
+        </names>
+      </if>
+      <else>
+        <group delimiter=" ">
+          <text variable="container-title"/>
+          <text macro="issued-date-full"/>
+        </group>
+      </else>
+    </choose>
+  </macro>
   <macro name="access">
     <choose>
       <if variable="URL" type="article-newspaper document report speech webpage post post-weblog" match="any">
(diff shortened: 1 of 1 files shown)


nguynking:
Hi @POBrien333, could you please take a look at this PR? Is there anything I should update or address before it can be merged? Thank you!

rmit-university-harvard.csl
```diff
@@ -189,16 +598,82 @@
...
+          </group>
+        </else-if>
+        <else>
+          <group delimiter=", ">
+            <group delimiter=" ">
+              <text macro="author-short"/>
+              <text macro="year-date"/>
+            </group>
+            <group delimiter=" ">
+              <label variable="locator" form="short"/>
```
POBrien333, line 671: everywhere else you call just the locator and here now also the label.
In general, it probably is better to use something like this:
`    <macro name="cite-locator">
    <group delimiter=" ">
      <choose>
        <if match="none" locator="page">
        </if>
      </choose>
      <text variable="locator"/>
    </group>
  </macro>`

Also, take this out of the conditional and only call it once.

rmit-university-harvard.csl
```diff
@@ -77,6 +106,67 @@
       </substitute>
     </names>
   </macro>
+  <macro name="editor-book">
+    <names variable="editor" delimiter=", ">
+      <name name-as-sort-order="all" and="text" sort-separator=" " initialize-with="" delimiter=", " delimiter-precedes-last="never"/>
+      <label form="short" strip-periods="true" prefix=" (" suffix=")"/>
+    </names>
+  </macro>
+  <macro name="author-book">
```
POBrien333, line 115: Why did you create all these author-x macros?

PR #8165, opened 2026-05-05
Add a style for SBL short note only citations with bibliography.

dstark:
This style supports, e.g., edited volumes that don't want multiple 'first note' instances across chapters but still want SBL-style bibliographies.


dstark:
The CSL validates, and I'm unsure what might need to happen due to the failed check. If something does need to be adjusted so that passes, please just let me know. Thanks so much!

PR #8177, opened 2026-05-20
Update lluelles.csl to latest style guide (9th ed.)

davidlesieur:
### Description

This updates the lluelles.csl to the 9th edition of the style guide:

Didier LLUELLES et Josée RINGUETTE, _Guide des références pour la rédaction juridique_, 9e éd., Montréal, Les Éditions Thémis, 2017, en ligne : https://ssl.editionsthemis.com/livres/livre-4895-guide-des-references-pour-la-redaction-juridique-9e-ed-.html.

This style is widely used in Canada for citing French-language legal sources. 

Making the CSL compliant with the latest edition of the style guide should encourage law students to use a reference manager.

Some details:

- Since this brings important changes, this update keeps the previous version as lluelles-v7.csl, for a transition period, in order to minimize disruption for users.
- There is no basis for lluelles-no-ibid.csl in the style guide. We thus preserve that style for the same transition period, but we remove its dependent lluelles-lexisnexis.csl which, we believe, has very little or no use at all.

This work has been funded by Université de Montréal - Bibliothèque de droit.

### Checklist


POBrien333:
Please remove all the "v7" stuff from the links in the style.

davidlesieur:
Thanks for following up. Removed "v7" from the links, left it in the ids.

davidlesieur:
Please let me know if "v7" should also be removed from the ids. Thanks.

PR #8195, opened 2026-06-06
Update pontifical-biblical-institute.csl

erasmobarresi:

### Description
The issue is discussed here: https://forums.zotero.org/discussion/130546/style-error-pontifical-biblical-institute

Since then, I've come across and read the CSL primer on citationstyles.org, and I've realized that the CSL language is not too hard to understand! (though it takes time to understand a long CSL file like this one) Therefore, I propose this change.

The logic of the change is this: in a bibliography, the volume number should be followed by a comma but only if further information (page range or URL) follows in the same entry. The macro called "container-review" is excluded from this reasoning because it already generates a comma as prefix (prefix=", ").

### Checklist
- [ ] Check that you've added a link to the style you used as a template in the `<info>` block at the beginning of the file with `rel="template"`.  [not applicable, I think]
- [ ] Check that you've used the correct terms or labels instead of hardcoding into affixes (e.g., `<text variable="page" prefix="pp. "/>`). [I'm not sure what this means.]

### DIFF
--- a/pontifical-biblical-institute.csl
+++ b/pontifical-biblical-institute.csl
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" initialize-with-hyphen="true" page-range-format="expanded" demote-non-dropping-particle="sort-only" name-delimiter=" &#8211; " names-delimiter=" &#8211; " delimiter-precedes-last="always" et-al-min="4" et-al-use-first="1" name-form="long" initialize-with="." sort-separator=", ">
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" initialize-with-hyphen="true" page-range-format="expanded" demote-non-dropping-particle="sort-only" name-delimiter=" – " names-delimiter=" – " delimiter-precedes-last="always" et-al-min="4" et-al-use-first="1" name-form="long" initialize-with="." sort-separator=", ">
   <!-- Polyglot -->
   <info>
     <title>Pontifical Biblical Institute</title>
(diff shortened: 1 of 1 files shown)


POBrien333:
Would you be able to apply this logic into the macro instead of your current suggestion?

erasmobarresi:
@POBrien333 possibly my comment from June 9 was a bit hidden (I'm sorry, I'm quite new to GitHub)... is that latest code good?

POBrien333:
Honestly, it's rather confusing. Can you tell us, in words, what you want to achieve?

erasmobarresi:
The volume number should be followed by a comma (and a space) but only if further information follows. Currently, the volume number and the page range of a book section do not get separated by a comma but by a mere space (at least if no locator is set; weirdly, the comma appears in a footnote only if a locator is set)

POBrien333:
Thanks for explaining.
Can you check the style I just adapted?

erasmobarresi:
I'm sorry, your code generates this kind of output, with an extra comma at the end:

Audo, T., Dictionnaire de la langue chaldéenne (Mossoul 1897) II,.

My twelve extra lines of code are intended to make the comma appear only when some other information follows.

Due to other complexities in the code, more drastic changes would be necessary to use a delimiter, which would be ideal here. When editing, I was a bit frustrated with the "container-review" macro (which is invoked right between two groups but outside of any group!) and the other eighteen provisions for the "Conference Paper" item type, which I did not understand. But I've just discovered (see https://forums.zotero.org/discussion/20563/new-style-please-upload-pontifical-biblical-institute) that the author of the code used "Conference Paper" to mean either "Review" or "Preface, introduction, etc. written by someone other than the author". Also, the "Archive" field means "Subtitle" and "Location in Archive" means "Subtitle of the container"...

PR #8205, opened 2026-06-20
APA 7th: add cs:intext for narrative ("Author and Year") citations

dawidstone5:
## Summary

Adds a `cs:intext` element to **apa.csl** so the processor can render
**narrative / in-text citations** — the author in running text with the year
in parentheses, e.g. *Smith and Jones (2020)* — as a complement to the existing
parenthetical form *(Smith & Jones, 2020)*.

This is the style-side counterpart to citeproc's `composite` cluster mode, and
enables word-processor integrations (e.g. Zotero — see zotero/zotero#5966) to
offer an "Author in text" option for active-voice citations.

## What changed

- New `author-intext` macro: identical to `author-short`, except the name
  conjunction renders as the word **"and"** instead of the ampersand **"&"**,
  per APA 8.17. `et-al-min`/`et-al-use-first` are set on the name so that 3+
  author works collapse to *Smith et al.* in narrative form too.
- New `<intext>` element rendering `author-intext`. The processor pairs this
  author-only output with the suppress-author form `(2020)` to assemble the
  full narrative citation.

## Verification

Rendered against citeproc-js with the modified style:

| Authors | Parenthetical | Narrative |
|---|---|---|
| 1 | `(Smith, 2020)` | `Smith (2020)` |
| 2 | `(Smith & Jones, 2020)` | `Smith and Jones (2020)` |
| 3+ | `(Smith et al., 2020)` | `Smith et al. (2020)` |

## Notes / for discussion

- `author-intext` is currently a near-duplicate of `author-short` (one
  attribute differs). Happy to refactor toward a shared macro if maintainers
  prefer a DRYer approach.
- Marked as **draft**: the common author/editor item types are covered and
  verified; review welcome on long-tail types (broadcast, legal, personal
  communication) before merge.

### DIFF
--- a/apa.csl
+++ b/apa.csl
@@ -515,6 +515,96 @@
       </else>
     </choose>
   </macro>
+  <!-- Narrative (in-text) form of the author, identical to `author-short` but with
+       the conjunction rendered as the word "and" rather than the ampersand "&",
+       per APA 8.17 (e.g. "Smith and Jones (2020)" vs "(Smith & Jones, 2020)"). -->
+  <macro name="author-intext">
+    <choose>
+      <if match="any" type="bill hearing legal_case legislation regulation treaty">
+        <text macro="title-and-descriptions-short"/>
+      </if>
+      <else-if match="any" type="interview personal_communication">
+        <choose>
+          <!-- These variables indicate that the letter is retrievable by the reader. If not, use the APA in-text-only personal communication format. -->
+          <if match="any" variable="archive archive-place container-title DOI number publisher references URL">
+            <names variable="author">
+              <name and="text" form="short"/>
+              <substitute>
+                <text macro="title-and-descriptions-short"/>
+              </substitute>
+            </names>
+          </if>
+          <else>
+            <group delimiter=", ">
+              <names variable="author">
+                <name and="text"/>
+                <substitute>
+                  <text macro="title-and-descriptions-short"/>
+                </substitute>
+              </names>
+              <text term="personal-communication"/>
+            </group>
+          </else>
+        </choose>
+      </else-if>
+      <else>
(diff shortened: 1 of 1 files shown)


dawidstone5:
After looking at the CI failure, this PR can't be merged as-is, and the reason is structural rather than a bug in the style.

**Root cause:** `cs:intext` is a **CSL 1.0.2** element, but this repository validates every style against the **CSL 1.0.1** schema, and the `csl-ruby` parser used in CI has no model for `intext`. So `apa.csl` fails to parse, and because it is the independent parent of ~780 dependent styles, all of them then fail the "same citation-format as its independent-parent" check. The 785 failures all cascade from that single parse failure.

There is no CSL 1.0.1 construct that can render the author with an ampersand inside parentheses (`(Smith & Jones, 2020)`) but with "and" in narrative form (`Smith and Jones (2020)`) — `cs:intext` is the mechanism specifically designed for this, and it only exists in 1.0.2.

So this is the well-known ecosystem-level gap, not something a single-style PR can resolve. It's already tracked here:
- citation-style-language/csl-evolution#13 (in-text/"narrative" citation configuration)
- citation-style-language/schema#365 (additional citation modes)
- zotero/zotero#1580 (downstream feature request)

The style itself renders correctly under citeproc-js (`Smith (2020)` / `Smith and Jones (2020)` / `Smith et al. (2020)`), so it can serve as a concrete reference for what an APA `cs:intext` would look like once the repo adopts 1.0.2 validation. Closing for now to keep CI clean; happy to revisit if/when 1.0.2 narrative styles become acceptable here.

bwiernik:
Narrative citations are not yet fully supported in CSL. They will be added in CSL 1.1

PR #8241, opened 2026-07-20
Delete harvard-cranfield-university.csl

andyp-uk:
Removing obsolete Cranfield University Harvard style

### Description
Removing the Cranfield University (author-date/Harvard) style because this is no longer endorsed by Cranfield University as per current guidance at [library.cranfield.ac.uk/referencing](https://library.cranfield.ac.uk/referencing).


POBrien333:
Are they suggesting to just use APA7?
In that case we'd prefer making it a dependent style instead of just deleting it.

andyp-uk:
@POBrien333 in general our students should use APA7 Author-Date and NLM Numbered styles. This particular style is different from APA7 and should no longer be used by our students.

We find that some students are using the "Cranfield University (author-date/Harvard)" style after stumbling across it, because they mistakenly believe it has been endorsed by us. We would prefer to remove it, but if maintained it should be renamed to make it clear to users that its a legacy style, and the code should also be updated to match https://raw.githubusercontent.com/searlesr/custyles/master/cranfield-university-author-date.csl.

Thanks,

Andy Panayiotou
Library Content and Discovery Manager
Cranfield University

POBrien333:
I don't think you understood what I was saying.
We can point this style to APA7. That way anybody who wrongly uses this style or any future style looking for it will just be using APA7.

andyp-uk:
@POBrien333 I do understand what you're suggesting. However, we do not want the name of the old style to appear anywhere.

POBrien333:
@bwiernik 
What's your take here?

bwiernik:
Okay to delete it and then add to renamed-styles pointing to APA7

POBrien333:
Ah great, thanks @bwiernik !

@andyp-uk 
Can you also modify the renamed-styles.json in this PR here accordingly?

andyp-uk:
@POBrien333 have updated renamed-styles.json and this seemed straightforward but am now getting a failing check, which I do not understand.

PR #8248, opened 2026-07-28

cdholmes:
### Description
The copernicus-publications style was previously unable to create bibliography entries for software items.

Software items are closely related to Dataset items,  so the Dataset type is modified to accommodate both. For both Software and Dataset, include URL in bibliography if DOI is empty

### Checklist

### DIFF
--- a/copernicus-publications.csl
+++ b/copernicus-publications.csl
@@ -318,13 +318,13 @@
             <text macro="editor"/>
             <group prefix=", ">
               <text variable="container-title" form="short"/>
-                <group>
-                  <text variable="volume"/>
-                  <text macro="published-date"/>
-                </group>
-                <text variable="archive"/>
-                <text macro="pages"/>
-                <text macro="doi"/>
+              <group>
+                <text variable="volume"/>
+                <text macro="published-date"/>
+              </group>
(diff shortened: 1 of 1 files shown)


cdholmes:
Thanks for your work on this, @POBrien333.

I see that the pull request has failed a test, but it's not clear to me what the cause is. Please let me know if I can do something to help.

POBrien333:
@cdholmes 
I've made some more edits to the style. Please test out the output.

cdholmes:
@POBrien333 
Your simplification looks good to me, with one exception. "Anon" isn't used in the [Copernicus style specifications](https://publications.copernicus.org/for_authors/manuscript_preparation.html#references).

For materials without an author, the specification says...
- the in-text citation should use the editor or title in place of the author. (I suppose Anon is OK as a last resort if there somehow isn't an editor or title either, but that edge case is not mentioned)
- the reference should continue to the rest of the reference, leaving the author area blank. 

I think that I did both of these, but a quick check with your experience wouldn't hurt. If you are happy, it's ready to merge.

POBrien333:
Note that I didn't introduce the "anon" behaviour. I only made the style more robust and removed bad coding practises.

PR #8253, opened 2026-07-29
Create cleaner-environmental-system.csl

LouisFreboeuf:
### Description

This pull request adds a new CSL style for the **Cleaner Environmental Systems** journal (Elsevier).

The style implements the journal's numeric citation format, with references numbered in order of first appearance. It includes formatting for the principal reference types accepted by the journal, including journal articles, books, book chapters, standards, reports, theses, datasets, software, webpages, and preprints. Journal titles are rendered using their abbreviated forms when available, DOIs are formatted as `https://doi.org/...`, and URLs include access dates where appropriate.

The style has been tested in Zotero and matches the journal's reference guidelines as closely as possible. A small number of `<text value="...">` elements are intentionally retained where CSL does not provide an equivalent built-in term or label.

### Checklist

* [x] Check that you've added a link to the style you used as a template in the `<info>` block at the beginning of the file with `rel="template"`.
* [x] Check that you've added a link to the style guidelines with `rel="documentation"`.
* [x] Check that you've added yourself as the `<author>` of the style or `<contributor>` for a style update.
* [ ] Check that you've used the correct terms or labels instead of hardcoding into affixes.
* [ ] Check that you've not used `<text value="...">` if not absolutely necessary.
* [x] Check that you've not changed line 1 of the style.


cleaner-environmental-system.csl
```diff
@@ -0,0 +1,249 @@
...
+        <else-if type="book">
+          <group suffix=".">
+            <text macro="author" suffix=", "/>
+            <text variable="title" suffix=", "/>
+            <text macro="edition" suffix=", "/>
+            <text macro="publisher" suffix=", "/>
+            <text macro="year"/>
+            <choose>
+              <if variable="DOI">
+                <text value=". " suffix=""/>
```
POBrien333: This can be set on the macro you're calling directly

cleaner-environmental-system.csl
```diff
@@ -0,0 +1,249 @@
...
+            <text variable="title" suffix=", "/>
+            <group delimiter=": ">
+              <text term="in" text-case="sentence"/>
+              <text macro="editor" suffix=", "/>
+            </group>
+            <text variable="container-title" suffix=", "/>
+            <text macro="publisher" suffix=", "/>
+            <text macro="year" suffix=", "/>
+            <group delimiter=" ">
+              <text value="pp" suffix=" "/>
```
POBrien333: We have a label for pages. Please replace

cleaner-environmental-system.csl
```diff
@@ -0,0 +1,249 @@
...
+            <text variable="version" prefix="v" suffix=", "/>
+            <text macro="year" suffix=". "/>
+            <text macro="doi"/>
+          </group>
+        </else-if>
+        <else-if type="software">
+          <group>
+            <text macro="author" suffix=", "/>
+            <text variable="title" suffix=" "/>
+            <text value="[software]" suffix=", "/>
```
POBrien333: You can set this on the title variable above as a suffix instead.

cleaner-environmental-system.csl
```diff
@@ -0,0 +1,249 @@
...
+            <text variable="publisher" suffix=", "/>
+            <date variable="issued" form="text" suffix=". "/>
+            <text macro="doi"/>
+          </group>
+        </else-if>
+        <else-if type="report">
+          <group>
+          <text macro="author" suffix=", "/>
```
POBrien333: all these suffix commas would be better set on the group level via "delimiter". Please do this for your whole style.

cleaner-environmental-system.csl
```diff
@@ -0,0 +1,249 @@
...
+              <text term="accessed"/>
+              <date variable="accessed" form="text"/>
+            </group>
+          </group>
+        </else-if>
+        <else-if type="manuscript">
+          <group>
+            <text macro="author" suffix=", "/>
+            <text variable="title" suffix=" "/>
+            <text value="[preprint]" suffix=", "/>
```
POBrien333: The term preprint exists and should be used instead.

LouisFreboeuf: Dear reviewer,

I, and with the help of mistral.ai, have replied to your comments with the commit: _00e0963a392df2cebe43b5ffd392d29bdbc14169_

Hope that this now made the style valid !

Kind regards,

Louis Freboeuf

PR #8256, opened 2026-08-03

POBrien333:
Continues #8244 to test out release pipeline.


POBrien333:
@adam3smith 
You were right. This new PR partially succeeds.
I don't fully understand the error, but I'll go ahead and merge this one.
It's a minor change.

adam3smith:
So the expected behavior is that both the test on pull_request and the test on pull_request_target pass for you (because you have commit access) whereas only pull_request passes for others. That doesn't seem to happen, so currently this is still broken and we need to figure out why pull_request isn't working

PR #8261, opened 2026-08-04

m-ryan4:
Working edit of Lancaster Harvard style citation format. In line with format guidelines (https://www.lancaster.ac.uk/media/lancaster-university/content-assets/documents/sociology/practice-learning/Harvard__Lancaster_University_Library__Referencing_Guide_21-22.pdf) the style should not add publisher for journal articles.

### Description
Removed publisher from journal articles, as it should not be a part of this style (see https://www.lancaster.ac.uk/media/lancaster-university/content-assets/documents/sociology/practice-learning/Harvard__Lancaster_University_Library__Referencing_Guide_21-22.pdf)

### Checklist
- [ ] Check that you've added a link to the style guidelines with `rel="documentation"`.  
- [ ] Check that you've added yourself as the `<author>` of the style or `<contributor>` for a style update. 
- [ ] Check that you've not changed line 1 of the style.

### DIFF
--- a/lancaster-university-harvard.csl
+++ b/lancaster-university-harvard.csl
@@ -10,9 +10,12 @@
     <author>
       <name>Andy Hartland</name>
     </author>
+    <contributor>
+      <name>Patrick O'Brien</name>
+    </contributor>
     <category citation-format="author-date"/>
     <summary>The Harvard author-date style - adapted for Lancaster University</summary>
-    <updated>2026-01-25T22:16:24+00:00</updated>
+    <updated>2026-08-06T08:14:16+00:00</updated>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <macro name="accessed">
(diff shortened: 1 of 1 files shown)


POBrien333:
Can you explain in words why you are suggesting this change?

m-ryan4:
According to the Lancaster University Harvard style guidelines (https://www.lancaster.ac.uk/media/lancaster-university/content-assets/documents/sociology/practice-learning/Harvard__Lancaster_University_Library__Referencing_Guide_21-22.pdf), the publisher should not appear for journal articles. Yet, in many instances, "publisher" shows up in the bibliography. This is an error.

For example, Corvellec, H. et al. (2022) Critiques of the circular economy. Journal of industrial ecology, 26(2), 421–432. Wiley Online Library. The "Wiley Online Library" should not be there for journal articles (see page 10).

I have modified the style for myself and would be glad to make this correction for anyone else who uses this style. Hope this helps!

POBrien333:
thanks.
Just setting an empty `value=""` is not the right approach. I've fixed this.
(In the future, if you don't feel comfortable with the code, you can just open an "issue" and report the error. We'll then fix it)

m-ryan4:
Thank you for working on this. Unfortunately the publisher still appears to be showing for journal articles though. I guess my way of resolving the issue wasn't the right way, but it did seem to eliminate the publisher. Can you take another look?

PR #8263, opened 2026-08-04
Use merge ref for pull_request_target checkout

fasihhulisan:
### Summary

`actions/checkout@v6` refuses to check out the head of a fork from a `pull_request_target` workflow.

As a result, the workflow currently fails before the validation steps (`rake` and `sheldon`) are executed for pull requests opened from forks.

This updates the checkout step to use GitHub's pull request merge ref instead:

```
refs/pull/<PR-number>/merge
```

which is provided by the base repository and is permitted in `pull_request_target` workflows.


adam3smith:
I fixed this issue yesterday -- as you can see, all PRs today run tests successfully

PR #8265, opened 2026-08-06
Update european-journal-of-neuroscience.csl to APA dependent

POBrien333:
…uropean-journal-of-neuroscience.csl

https://forums.zotero.org/discussion/133086/edit-european-journal-of-neuroscience-style-to-include-doi#latest

This is quite a mess.
Guidelines and papers mismatch. Actual papers are also quite a mix of styles. I suggest just switching this to an APA dependent, since that is what the guidelines say. The journal also applies their style during typesetting and authors can hand in as they want.


POBrien333:
@adam3smith 
Something funky is going on here.
Another, unrelated style just appeared now.
Any clue?

PR #8286, opened 2026-08-18
Add Bibliotheca Hertziana - Max Planck Institute for Art History (Primo/Alma, English/German/Italian)

littlemy82:
## Summary

This adds three new citation styles derived from the existing 
`bibliotheca-hertziana-max-planck-institute-for-art-history.csl` (linked via 
the `template` element in each):

- `bibliotheca-hertziana-primo-eng.csl` (English)
- `bibliotheca-hertziana-primo-ger.csl` (German)
- `bibliotheca-hertziana-primo-ita.csl` (Italian)

They are optimized for use with Ex Libris Alma/Primo VE's built-in "Cite" 
button, which requires a single, standalone full citation per record rather 
than the note+short-title citation apparatus the original style supports.

Three separate language variants are needed because the institute's editorial 
guidelines specify that connector words ("ed." vs. "hg. v." vs. "a cura di") 
and punctuation conventions (e.g. the Oxford comma, used in English but not 
in German/Italian) follow the language of the *citing* publication, not the 
language of each individual cited item. This cannot be handled by per-item 
locale switching within a single style, since Alma's Cite button needs to 
select one fixed convention per rendered reference. We are submitting all 
three together in one PR since they share the same structure, rationale, and 
test suite, and differ only in locale-specific terms and punctuation rules.

## Why we believe this fits the repository's scope

We are aware that the "Criteria for Inclusion" generally discourages styles 
for narrow internal/API use. We believe this submission differs from that 
case:

- All three are based on the institute's official, publicly documented 
  editorial guidelines (linked via the `documentation` element), not on 
  undocumented internal conventions.
- Ex Libris Alma/Primo VE is used by a very large number of academic and 
  research libraries; these styles demonstrate how to adapt an existing 
  Zotero-oriented style for this platform's citation constraints, which may 
  be of practical use beyond this one institution.
- An equivalent style for this same institution (the template above) is 
  already accepted in this repository, establishing precedent for 
  institutional styles of this kind.

We are happy to discuss further, or to split/withdraw individual variants, 
if the maintainers feel any of them does not fit.

## Testing

All three styles were:
- Validated against the CSL 1.0.2 schema via https://validator.citationstyles.org/ 
  with no errors.
- Manually tested against the same 22 constructed test cases covering all 
  supported types (book, chapter, article-journal, review, entry-dictionary, 
  thesis, webpage, manuscript), derived directly from the worked examples in 
  the institute's editorial guidelines, including edge cases such as:
  - multi-volume works with and without a specific sub-volume title
  - exhibition catalogues vs. conference proceedings (different punctuation 
    depending on presence of an event date)
  - reviews with and without their own title
  - anonymous/no-author works
  - DOI vs. plain URL handling
  - locale-specific punctuation (Oxford comma, quotation mark styles)

## Known limitations (documented in each style's comments)

CSL cannot inspect or reformat free-text field content, which means:
- Publisher place names, multi-place "et al." notations, and genre/thesis-type 
  labels are passed through as-is from the underlying data and may need 
  manual correction depending on how the record is catalogued.
- Multi-year publication spans for multi-volume works (e.g. "1957-1959") 
  cannot be rendered alongside a separate specific-volume year, since CSL 
  only supports one date value per field.

These are noted directly in each style file for future maintainers.


adam3smith:
I'm afraid I don't agree that this meets criteria for inclusion: just because implementations (here in Alma/Primo) have limitations doesn't mean we should add custom styles for them. (I also think 6 citation styles of a single institute is going a bit overboard). If you need this style, you should talk to ExLibris about the ability to use locally hosted styles with the catalog (the Zotero API, e.g., allows for using styles outside CSL) but I'm afraid we're not going to take this.

We have to set limits for hosting styles at some point, and since you're arguing from precedent above, think about what accepting Primo specific variants of styles would mean as a precedent.

PR #8296, opened 2026-08-25
Create neue-politische-literatur.csl

adam3smith:
### Description
New style for NPL  (generated with https://github.com/adam3smith/csl-style-generation and human review)

### Checklist

### DIFF
--- a/neue-politische-literatur.csl
+++ b/neue-politische-literatur.csl
@@ -24,6 +24,7 @@
       <!-- NPL prints journal issues as "H." (Heft) and edited-by as "hrsg. v."; op-cit carries the back-reference to the note with the full citation, "wie Anm. 4". -->
       <term name="issue" form="short">H.</term>
       <term name="op-cit">wie Anm.</term>
+            <term name="accessed">Zugriff</term>
       <term name="editor" form="verb-short">hrsg. v.</term>
       <term name="editortranslator" form="verb-short">hrsg. u. übers. v.</term>
       <term name="translator" form="verb-short">übers. v.</term>
@@ -244,7 +245,10 @@
       </choose>
       <choose>
         <if variable="DOI URL" match="any">
-          <date variable="accessed" form="numeric" prefix="[" suffix="]"/>
+          <group delimiter=": " prefix="[" suffix="]">
+          <text term="accessed" text-case="capitalize-first"/>
+          <date variable="accessed" form="numeric"/>
+          </group>
         </if>
       </choose>
     </group>
--- a/neue-politische-literatur.csl
+++ b/neue-politische-literatur.csl
@@ -81,11 +81,11 @@
   </macro>
   <macro name="title">
     <choose>
-      <if variable="collection-title volume" match="all" type="book">
+      <if variable="volume-title volume" match="all" type="book">
         <!-- Mehrbändige Werke: "Reihentitel. Bd. 1: Titel des Bandes" -->
         <group delimiter=": ">
           <group delimiter=". ">
-            <text variable="collection-title"/>
+            <text variable="title"/>
             <group delimiter=" ">
               <label variable="volume" form="short"/>
               <number variable="volume"/>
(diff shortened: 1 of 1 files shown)


neue-politische-literatur.csl
```diff
@@ -0,0 +1,325 @@
...
+  <macro name="editor-secondary">
+    <!-- "hrsg. v. Hubertus Buchstein/Rainer Kühn": only next to a named author, and only where the editor is not already carrying the container. -->
+    <choose>
+      <if type="chapter paper-conference entry-encyclopedia entry-dictionary" match="none">
+        <choose>
+          <if variable="author" match="all">
+            <names variable="editor translator" delimiter=", ">
+              <label form="verb-short" suffix=" "/>
+              <name delimiter="/"/>
+              <et-al/>
```
POBrien333: This could be deleted, no?
Same in line 81

neue-politische-literatur.csl
```diff
@@ -0,0 +1,325 @@
...
+          </group>
+          <group delimiter=" ">
+            <text term="issue" form="short"/>
+            <text variable="issue"/>
+          </group>
+        </group>
+      </else-if>
+      <else-if type="article-newspaper post-weblog webpage" match="any">
+        <group delimiter=" ">
+          <text term="in" suffix=": "/>
```
POBrien333: I'd move this suffix into the delim above

neue-politische-literatur.csl
```diff
@@ -0,0 +1,325 @@
...
+        </group>
+      </else-if>
+    </choose>
+  </macro>
+  <macro name="issue-number">
+    <!-- Wochenzeitungen: "Nr. 28". The issue term is bound to "H." for journals here, so the newspaper sense uses the number term, which de-DE already spells "Nr.". -->
+    <choose>
+      <if type="article-newspaper" match="any">
+        <group delimiter=" ">
+          <text term="number" form="short"/>
```
POBrien333, line 143: this style (or the template I guess) has a lot of "text terms" instead of "label variable"

neue-politische-literatur.csl
```diff
@@ -0,0 +1,325 @@
...
+            <date-part name="year"/>
+          </date>
+        </if>
+        <else>
+          <text term="no date" form="short"/>
+        </else>
+      </choose>
+    </group>
+  </macro>
+  <macro name="edition">
```
POBrien333, line 175: Never seen this logic for edition. Oversight or on purpose?

POBrien333 (commented):
Some nitpicks :)

neue-politische-literatur.csl
```diff
@@ -0,0 +1,325 @@
...
+  </macro>
+  <macro name="title">
+    <choose>
+      <if variable="collection-title volume" match="all" type="book">
+        <!-- Mehrbändige Werke: "Reihentitel. Bd. 1: Titel des Bandes" -->
+        <group delimiter=": ">
+          <group delimiter=". ">
+            <text variable="collection-title"/>
+            <group delimiter=" ">
+              <text term="volume" form="short"/>
```
POBrien333: Checking new label-gun.yml.
change to `label variable`

POBrien333 (changes requested):

adam3smith:
Thanks for taking a look -- the labelling test failed, alas, let me see what we can do

neue-politische-literatur.csl
adam3smith: When not used as a count (i.e. needing to distinguish singular from plural) I don't see a benefit of `label plural="never"` and `term` and term is shorter. 
So I think "vol. 5" is actually fine as
```
<group delimiter=" ">
  <text variable="volume"/>
</group>
```

whereas "5 vols." requires

```
<group delimiter=" ">
  <number variable="volume"/>
</group>
```

label does also have an additional built-in conditional (label doesn't render if the corresponding variable is empty) that term doesn't have, but since that's also part of group, we rarely rely on it.

neue-politische-literatur.csl
adam3smith: On purpose -- the style handles numeric editions as a superscript to year above, but longer, free text edition ("überarbeitete Neuauflage" or the like) get rendered here as plain text. 
The is-numeric check isn't unusual, though splitting the macro up like this is (but so is the handling of edition in the style)

PR #8297, opened 2026-08-25

### LINKED ISSUE #8295: ACS Revision 2026
Based on the [newest revision](https://pubs.acs.org/books/book/4293/chapter/244189111/ACS-Style-Quick-Guide) for the ACS Style Quick Guide, there are two things that I noticed that will need to be updated.

First, there is an option to cite theses and dissertations with links; however, the CSL does not allow for the links to be shown on the bibliography (so I'm presuming it should be able to function similarly with the webpage type, except without an access date). As an example from the website:

Cable, M. L. Life in Extreme Environments: Lanthanide-Based Detection of Bacterial Spores and Other Sensor Design Pursuits. Ph.D. Dissertation, California Institute of Technology, 2010. http://resolver.caltech.edu/CaltechTHESIS:05102010-145436548

Second, I noticed that after links are posted in the bibliography, there should be no periods showing up. The above example highlights that there is no period after the reference, though if there is no link involved (such as in a print reference), a period does show up. I have posted the other two examples from the webpage here as well for your comparison:

Image

There is also no period shown at the end if the link has an access date associated with it:

Image

Hope this all makes sense! Please don't hesitate to let me know if you have any questions or comments!

POBrien333:
closes #8295

- Removes publisher-place for thesis
- removes trailing dot for DOI/URLs
- minor cleanup

@adunning and @bwiernik 
You recently made a change to the name of the style, but I'm not sure what the year number was based on. Should I update it to `2026`? I'm not confident since I only fixed some issues, but cannot guarentee there are others.

### DIFF
--- a/american-chemical-society.csl
+++ b/american-chemical-society.csl
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" demote-non-dropping-particle="sort-only" page-range-format="expanded" default-locale="en-US">
   <info>
-    <title>ACS Guide 2022 revision</title>
+    <title>ACS Guide 2026 revision</title>
     <title-short>American Chemical Society</title-short>
     <id>http://www.zotero.org/styles/american-chemical-society</id>
     <link href="http://www.zotero.org/styles/american-chemical-society" rel="self"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
> @adunning and @bwiernik
You recently made a change to the name of the style, but I'm not sure what the year number was based on. Should I update it to 2026? I'm not confident since I only fixed some issues, but cannot guarentee there are others.

Here's the relevant commit message: https://github.com/citation-style-language/styles/commit/7573fb9e3ae8185d1fe43f53c48cb84ece1c9fe1

I think we should update, yes; if we find other issues compared to the 2026 version of the style, we'll update accordingly

PR #8269, opened 2026-08-10
Add GitHub Action to label PR on review submission

POBrien333:
I'm suggesting this workflow so a PR automatically recieves the "waiting-for-contributor" label once a review was sent by us.

PR #8252, opened 2026-07-29
Add UIN Sunan Kalijaga - Thesis (Notes and Bibliography)

fasihhulisan:
## CSL Styles Pull Request Template
You're about to create a pull request to the CSL **styles** repository.  
If you haven't done so already, see <https://github.com/citation-style-language/styles/blob/master/CONTRIBUTING.md> for instructions on how to contribute, and <https://github.com/citation-style-language/styles/blob/master/STYLE_REQUIREMENTS.md> for the requirements a style must meet before acceptance.  
In addition, please fill out the pull request template below.

### Description

This pull request adds a new Citation Style Language (CSL) style for the **Graduate Thesis Writing Guidelines of UIN Sunan Kalijaga Yogyakarta (Indonesia)**.

The style is adapted from the official **Chicago Manual of Style (Notes and Bibliography)** and customized to comply with the university's thesis writing guidelines.

Main customizations include:

- Full Notes and Bibliography implementation following UIN Sunan Kalijaga guidelines.
- Indonesian citation conventions (`default-locale="id-ID"`).
- Customized formatting for:
  - Books
  - Book chapters
  - Journal articles
  - Conference papers
  - Working papers
  - Reports
  - Webpages
  - Theses
  - Other supported CSL item types
- Footnote formatting according to the official guideline.
- Bibliography formatting according to the official guideline.
- Correct handling of:
  - *Ibid.*
  - *Ibid.* with page numbers
  - Repeated citations separated by other references
- Customized multivolume book formatting.
- Customized punctuation and delimiter rules.
- Customized quotation and italic rules.
- Proper handling of Working Paper references as required by the UIN guideline.
- Updated CSL metadata to satisfy CSL Repository requirements.

The implementation has been validated using the official CSL Validator with no validation errors.

Documentation:
https://pps.uin-suka.ac.id/id/show/pengumuman/4991/pedoman-penulisan-tesis

Repository:
https://github.com/fasihhulisan/UIN-Sunan-Kalijaga-CSL

### Checklist
- [x] Check that you've added a link to the style you used as a template in the `<info>` block at the beginning of the file with `rel="template"`.  
- [x] Check that you've added a link to the style guidelines with `rel="documentation"`.  
- [x] Check that you've added yourself as the `<author>` of the style or `<contributor>` for a style update. 
- [x] Check that you've used the correct terms or labels instead of hardcoding into affixes (e.g., `<text variable="page" prefix="pp. "/>`).
[...]

PR #8225, opened 2026-07-09
Create dumbarton-oaks-pre-columbian.csl

ej2432:
## CSL Styles Pull Request Template
You're about to create a pull request to the CSL **styles** repository.  
If you haven't done so already, see <https://github.com/citation-style-language/styles/blob/master/CONTRIBUTING.md> for instructions on how to contribute, and <https://github.com/citation-style-language/styles/blob/master/STYLE_REQUIREMENTS.md> for the requirements a style must meet before acceptance.  
In addition, please fill out the pull request template below.

### Description
(Briefly describe the style or changes you're proposing.)

### Checklist
- [x] Check that you've added a link to the style you used as a template in the `<info>` block at the beginning of the file with `rel="template"`.  
- [x] Check that you've added a link to the style guidelines with `rel="documentation"`.  
- [x] Check that you've added yourself as the `<author>` of the style or `<contributor>` for a style update. 
- [x] Check that you've used the correct terms or labels instead of hardcoding into affixes (e.g., `<text variable="page" prefix="pp. "/>`).
- [x] Check that you've not used `<text value="...` if not absolutely necessary.
- [x] Check that you've not changed line 1 of the style.

Issue #8107, opened 2026-03-18
polish_CHAT.csl


Ondraszek13:
<?xml version="4.0" encoding="utf-8"?>
<style class="note" version="1.0" name-as-sort-order="all" default-locale="pl-PL" xmlns="http://purl.org/net/xbiblio/csl">
  <info>
    <title>Polish Legal (Polski) CHAT</title>
    <id>http://www.zotero.org/styles/polish-legal-polski-chat</id>
    <link href="http://www.zotero.org/styles/polish-legal-polski-chat" rel="self"/>
    <link href="https://github.com/citation-style-language/styles/pull/8107" rel="documentation"/>
    <author>
      <name>Aleksandra O-S</name>
      <email>ondraszek@gmail.com</email>
    </author>
    <category citation-format="note"/>
    <category field="law"/>
    <summary>A style for legal citations CHAT</summary>
    <updated>2026-03-18T13:20:29+00:00</updated>
    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
  </info>
  <locale xml:lang="pl">
    <terms>
      <term name="month-01" gender="masculine">stycznia</term>
      <term name="month-02" gender="masculine">lutego</term>
      <term name="month-03" gender="masculine">marca</term>
      <term name="month-04" gender="masculine">kwietnia</term>
      <term name="month-05" gender="masculine">maja</term>
      <term name="month-06" gender="masculine">czerwca</term>
      <term name="month-07" gender="masculine">lipca</term>
      <term name="month-08" gender="masculine">sierpnia</term>
      <term name="month-09" gender="masculine">września</term>
      <term name="month-10" gender="masculine">października</term>
      <term name="month-11" gender="masculine">listopada</term>
      <term name="month-12" gender="masculine">grudnia</term>
      <term name="section" form="short">teza</term>
    </terms>
  </locale>
  <macro name="editor-translator">
    <names variable="editor translator" delimiter=", ">
      <name initialize-with=". " delimiter=", " delimiter-precedes-last="always" sort-separator=" "/>
      <label form="short" prefix=" (" suffix=".)"/>
    </names>
  </macro>
  <macro name="author">
    <names variable="author" display="block">
      <name delimiter="," delimiter-precedes-last="always" initialize-with="." name-as-sort-order="last" sort-separator=" "/>
      <label form="short" prefix=" (" suffix=")"/>
      <substitute>
[...]

PR #8305, opened 2026-08-30
label-gun: fix the awaiting label, stop the reopens, handle reviews

adam3smith:
Three separate problems in `label-gun.yml`, plus the review handling that #8269 asked for.

### 1. `label.awaiting` has been a dead input since the v5 bump

e829310 bumped label-gun v4.0.11 → v5.0.0, which renamed every dotted input (`label.awaiting` → `label-awaiting`). Every run since logs:

```
##[warning]Unexpected input(s) 'label.awaiting', valid inputs are [... 'label-awaiting' ...]
  label.awaiting: waiting-for-response-from-contributor   <- ignored
  label-awaiting: awaiting-user-feedback                  <- the default, used
```

So the action has been applying its own `awaiting-user-feedback` default, not ours. That accounts for the grey, description-less `awaiting-user-feedback` label that appeared in the label list; the 12 issues/PRs carrying it have been sorted out separately and the label deleted.

### 2. `pull_request_target` can never work

a25eb6f6 added it, but label-gun's [`src/main.ts`](https://github.com/retorquere/label-gun/blob/v5.0.0/src/main.ts#L95) hard-rejects anything that isn't `issues` or `issue_comment`:

```js
if (!['issues', 'issue_comment'].includes(context.eventName)) {
  throw new Error(`Unexpected event ${context.eventName}`)
}
```

Every PR opened/edited/closed since has been a red cross (e.g. [run 33303759368](https://github.com/citation-style-language/styles/actions/runs/33303759368): `##[error]Unexpected event pull_request_target`). Even past that guard the payload has no `issue` key, so it would throw again. Removed — PR comments already reach the action as `issue_comment`, so nothing is lost.

### 3. The reopens

[`src/app.ts`](https://github.com/retorquere/label-gun/blob/v5.0.0/src/app.ts#L263-L288) reopens a closed thread down two paths, both in the `!sender.owner` branch:

- a non-maintainer comments on a closed issue/PR — unconditional, no config involved
- a non-maintainer closes an issue/PR themselves, unless it carries a `labels.reopened`/`labels.canClose` label or no maintainer ever participated

Neither is switchable off. `labels.exempt`/`labels.active` are the only kill switches and they disable the whole action.

It fires a lot. Over the last 60 issues and 60 PRs: #8252 reopened by `github-actions` five times, #7720 / #7910 / #7854 three or four times each, ~30 more across #7590–#8107. The other recent red run is this too — on #8225 it tried to reopen an **already-merged** PR and got `Validation Failed` back from the API.

Both paths are reachable only for a non-maintainer acting on an already-closed item, so this adds a step that resolves the sender's permission and gates the action on it:

```yaml
if: steps.sender.outputs.maintainer == 'true' || github.event.issue.state == 'open'
```

| | |
|---|---|
| contributor comments, item open | runs — clears the label, as now |
| contributor comments, item closed | **skipped — no reopen** |
| contributor closes their own PR | **skipped — no reopen** |
| maintainer closes | runs — strips the awaiting label |
| maintainer comments, either state | runs, as now |

The permission set the gate accepts (`admin|maintain|push|triage|write`) is copied from label-gun's own `Collaborators.isOwner` and read off the same REST field, so the gate cannot disagree with how the action classifies the sender. That call already happens inside the action under this token, so no permission change was needed.

One thing this gives up: a contributor commenting on a closed thread no longer gets the awaiting label stripped, because the action doesn't run at all. That only affects already-closed items.

### 4. Reviews — supersedes #8269

`pull_request_review` and `pull_request_review_comment` hit the same guard as (2), so label-gun never sees them. Two gaps followed: a maintainer review didn't set the label, and — the less obvious half — a contributor **replying to an inline review comment** fires `pull_request_review_comment`, not `issue_comment`, so it couldn't clear the label either.

Added as a separate `review` job, taking @POBrien333's `author_association` gate from #8269 and adding the reverse direction:

- maintainer review or inline comment → add the label, **except** on `approved` (an approval isn't waiting on the contributor). Easy to flip if you'd rather approvals cleared it.
- anyone else → remove it, mirroring what label-gun does for a non-maintainer comment

No checkout happens in this workflow, so unlike `sheldon.yaml` the write token is never exposed to pull request code.

Note this is chatty: a review with N inline comments fires N+1 runs. They're idempotent, and the alternative is dropping `pull_request_review_comment` and losing the contributor-reply half.

### Also

Dropped `closed` from the `issue_comment` types — not a valid activity type for that event (only `created`/`edited`/`deleted`), so it was a no-op.

### Upstream

(3) and (4) are workarounds for gaps that belong in the action rather than here. Both are worth raising with @retorquere — a switch to opt out of the reopen, and review-event support — and if they land, both the gate step and the `review` job can be deleted.

### DIFF
--- a/.github/workflows/label-gun.yml
+++ b/.github/workflows/label-gun.yml
@@ -3,22 +3,93 @@ name: 'Automated tagging for PRs and issues'
 on:
   issues:
     types: [opened, edited, closed]
-  pull_request_target:
-    types: [opened, edited, closed]
   issue_comment:
-    types: [created, edited, closed]
+    types: [created, edited]
+  # label-gun cannot handle these -- src/main.ts hard-rejects any event that
+  # isn't 'issues' or 'issue_comment' -- so they go to the 'review' job below.
+  pull_request_review:
+    types: [submitted]
+  pull_request_review_comment:
+    types: [created]
 
 permissions: {}
 
+env:
+  AWAITING: waiting-for-response-from-contributor
+
 jobs:
   label:
+    if: github.event_name == 'issues' || github.event_name == 'issue_comment'
+
     permissions:
       issues: write # to add label to issues (retorquere/label-gun)
       pull-requests: write # to add label, comment on pull request (retorquere/label-gun)
 
     runs-on: ubuntu-latest
     steps:
+      # label-gun reopens a closed issue/PR when a non-maintainer comments on it
+      # (src/app.ts, 'issue_comment.created' && issue.state === 'closed') or closes
+      # it themselves ('issues.closed'), and neither is switchable off in config.
+      # Both paths are reachable only for non-maintainers acting on a closed item,
+      # so resolve the sender's permission here and skip exactly that case.
+      #
+      # The accepted set below is label-gun's own (src/app.ts Collaborators.isOwner)
+      # read off the same REST field, so this gate cannot disagree with how the
+      # action itself classifies the sender.
+      - name: Is the sender a maintainer?
+        id: sender
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          SENDER: ${{ github.event.sender.login }}
+        # non-collaborators return 'none'/'read'; bot accounts and deleted users 404
+        run: |
+          permission=$(gh api "repos/$REPO/collaborators/$SENDER/permission" --jq .permission 2>/dev/null || echo none)
+          echo "$SENDER has $permission permission"
+          case "$permission" in
+            admin|maintain|push|triage|write) maintainer=true ;;
+            *)                                maintainer=false ;;
+          esac
+          echo "maintainer=$maintainer" >> "$GITHUB_OUTPUT"
+
       - uses: retorquere/label-gun@v5.0.0
+        # a maintainer never triggers a reopen, and neither does anyone acting on
+        # an item that is still open. Everything else -- contributor comments on a
+        # closed thread, contributor closing their own PR -- is dropped.
+        if: steps.sender.outputs.maintainer == 'true' || github.event.issue.state == 'open'
         with:
           token: ${{ github.token }}
-          label.awaiting: "waiting-for-response-from-contributor"
+          # v5 renamed the dotted inputs; 'label.awaiting' is silently ignored and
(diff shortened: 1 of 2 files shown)


.github/workflows/label-gun.yml
```diff
@@ -3,22 +3,93 @@ name: 'Automated tagging for PRs and issues'
...
       issues: write # to add label to issues (retorquere/label-gun)
       pull-requests: write # to add label, comment on pull request (retorquere/label-gun)
 
     runs-on: ubuntu-latest
     steps:
+      # label-gun reopens a closed issue/PR when a non-maintainer comments on it
+      # (src/app.ts, 'issue_comment.created' && issue.state === 'closed') or closes
+      # it themselves ('issues.closed'), and neither is switchable off in config.
+      # Both paths are reachable only for non-maintainers acting on a closed item,
+      # so resolve the sender's permission here and skip exactly that case.
```
retorquere: I am open to changing that.

adam3smith:
The PR review labels create an extra check in the merge box: label-on-review.yml shows up there and plain label-gun never did — that's not an oversight but a property of the trigger. Runs from issues/issue_comment attach to the default branch, so they're invisible on a PR, whereas anything triggered by pull_request_review attaches to the PR head and gets a check run, skipped jobs included, with no way to opt out. 

The only invisible alternative would have been a scheduled reconciler, since cron runs also attach to the default branch, and I've decided against that: the lag before the label appears, plus sweeping 250-odd open PRs every few minutes, both cost more than a check row labelled label does. So we accept one extra visible check. 

Edit: no longer relevant; we ended up using Cron after all.

POBrien333:
Wowser! Excellent. Thanks! :)

retorquere:
I have published 5.0.1, which makes the reopen behavior configurable (default off)

adam3smith:
@retorquere -- thanks! When I have a minute I'll test with 5.0.1 and remove the workaround
@POBrien333 -- note that the way the labels for the awaiting review tag on pull request reviews and comments work is via a check every 10-15mins. It wasn't easily possible to do this on review (complicated story about github permissions on pull requests and automation security), so keep that in mind if you don't see them appearing immediately. They check what the current label should be, hopefully that doesn't cause any unwanted behavior.

PR #8307, opened 2026-08-31
Update bibliotheca-hertziana-max-planck-institute-for-art-history.csl (minor fixes)

littlemy82:
### Description

Three small, independent bug fixes found while working with this style, 
each directly contradicting an explicit rule in the linked editorial 
guidelines:

1. The "anonymous" term was set to the Italian "s.a." in both the German 
   and English locale blocks (apparent copy-paste artifact). Corrected to 
   "o.V." (German) and "Anon." (English), matching the guidelines' explicit 
   examples ("[Anon.]" / "[o.V.]").
2. Type "review"/"review-book" items fell through to the generic fallback 
   branch, losing the container title, volume, issue, and pages. They now 
   share the existing article-journal branch, which already produces the 
   format shown in the guidelines' review examples.
3. The accessed-date month/day lacked leading zeros (e.g. "4.12.2024" 
   instead of "04.12.2024"), not matching the dd.mm.yyyy format the 
   guidelines specify and exemplify.

No structural or macro changes beyond these three fixes.

### Checklist


bibliotheca-hertziana-max-planck-institute-for-art-history.csl
```diff
@@ -64,7 +64,7 @@
         <multiple>Bde.</multiple>
       </term>
       <term name="page-range-delimiter">&#8211;</term>
-      <term name="anonymous">s.a.</term>
+      <term name="anonymous">o.V.</term>
```
POBrien333, line 67: You want "ohne Verfasser" for the Italian locale?

bibliotheca-hertziana-max-planck-institute-for-art-history.csl
POBrien333: Please change this date to today

bibliotheca-hertziana-max-planck-institute-for-art-history.csl
littlemy82: Good catch on double-checking - but this particular hunk is actually within 
the German locale block (`<locale xml:lang="de">`), not Italian. You can 
verify this directly: the "Bde." term right above it (plural of "Bd.", 
i.e. "Band") only appears in the German locale block in this file - the 
Italian locale's number-of-volumes term is "voll." instead. So "o.V." for 
"ohne Verfasser" is correct here. The Italian locale's "anonymous" term 
("s.a.") was already correct and untouched by this PR.

bibliotheca-hertziana-max-planck-institute-for-art-history.csl
littlemy82: Done - updated to today's date.

PR #8308, opened 2026-08-31
Update istanbul-universitesi-sosyal-bilimler-enstitusu.csl (et-al, author repetition, Roman-numeral volumes, two-column bibliography)

Ercaner1988:
This fixes several concrete discrepancies between `istanbul-universitesi-sosyal-bilimler-enstitusu.csl` and the citation manual it implements, İstanbul Üniversitesi Sosyal Bilimler Enstitüsü's *Tez Hazırlama Yönergesi* (thesis-preparation directive), verified with citeproc-js against the directive's own worked examples (Örnek 3 / Örnek 4, Madde 20–21).

## Bugs fixed

1. **`et-al` abbreviation** — rendered as `vd.`, the directive (Madde 21) requires `v.d.` (with a period after each letter).
2. **Repeated-author dash substitution** — the bibliography used `subsequent-author-substitute="———"` to replace a repeated author's name with a dash. Örnek 4 shows two different works by the same author (Bülent Tanör) each with the author's full name repeated, not dashed. Removed the substitution.
3. **Volume numbers** — for journal articles, books, encyclopedia entries, dictionary entries, and conference papers, volume numbers were rendered as plain Arabic numerals (`C. 9`). Madde 20-14-d states volume numbers must be given in **uppercase Roman numerals** (`C.II`, `C.XV`, `Vol.IX` per the directive's own examples). Fixed at every location in the style where a cited volume number is rendered.
   - This required switching `default-locale` from `tr-TR` to `en-US`, because citeproc-js derives the locale used for uppercase-casing (`text-case="uppercase"`) from the style's base locale, not from any per-element locale — with `tr-TR` active, lowercase `i` uppercases to dotted `İ`, breaking Roman numerals (`II` → `İİ`). All Turkish terms the style previously inherited from `<locale xml:lang="tr">`'s implicit tr-TR fallback are now explicitly defined in the `en-US` locale block (month names, ordinal suffixes, quote marks, contributor role labels, `no date`, etc.), plus the `en-US`-only `<date form="numeric">` and `punctuation-in-quote` locale settings that would otherwise silently change from Turkish to American conventions. This was validated with a broad regression test across many additional types (newspaper articles, interviews, reviews, edited books, book chapters) to confirm no other output changed.
4. **Bibliography layout** — restructured to the two-column, "Author: / hanging citation" layout shown in Örnek 4 (`second-field-align="flush"`), replacing the previous `hanging-indent` layout.
5. **Duplicate volume number** — pre-existing bug (present before this PR, on `entry-dictionary` and `paper-conference` types): the volume number was rendered twice in the same citation/bibliography entry (once from the generic title macro, once from the locator-specific macro). Fixed by excluding those two types from the generic volume macro, since their volume is already rendered by the dedicated locator macros.

## Validation

All changes were checked with citeproc-js against the directive's own footnote/bibliography examples (Hirsch, Kaboğlu, Homans, Tanör, Sartori, Bohannan), plus a broader smoke test across `article-newspaper`, `entry-encyclopedia`, `interview`, `chapter` (with `container-author`), edited `book`, `review`, `entry-dictionary`, and `paper-conference` items, run both before and after the change, to confirm the only differences are the ones described above (no regressions in unrelated types, e.g. `legal_case` citation formatting, which intentionally keeps Arabic reporter volume numbers, was left untouched).

No other functionality of the style (legal citations, archival sources, interviews, letters, etc.) was modified.

### DIFF
--- a/istanbul-universitesi-sosyal-bilimler-enstitusu.csl
+++ b/istanbul-universitesi-sosyal-bilimler-enstitusu.csl
@@ -1,5 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
-<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" delimiter-precedes-et-al="never" page-range-format="expanded" default-locale="tr-TR">
+<!--
+  NOT: default-locale bilincli olarak en-US birakildi. citeproc-js (Zotero),
+  buyuk harf donusumunu (text-case="uppercase") stilin ana diline gore yapar;
+  tr-TR ana dilde kucuk "i" harfi noktali buyuk harfe ("I" yerine "İ") donusur,
+  bu da Madde 20/14-d'nin istedigi "C.II" yerine "C.İİ" gibi hatali Romen
+  rakami cikmasina yol acar. Ana dili en-US tutup butun Turkce terimleri (ay
+  adlari, ad/soyad rolleri, siralama ekleri dahil) asagidaki locale blogunda
+  tanimlayinca hem dogru Romen rakamlari hem de tum Turkce kisaltmalar cikar.
+-->
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" delimiter-precedes-et-al="never" page-range-format="expanded" default-locale="en-US">
   <info>
     <title>İstanbul Üniversitesi Sosyal Bilimler Enstitüsü (Türkçe)</title>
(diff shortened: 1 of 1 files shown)


Ercaner1988 (commented):
Ufak tefek bazı eksiklikler vardı Ali onları düzelttim, kabul edersen makbule geçer.

PR #8327, opened 2026-09-17
Update saa.csl and american-antiquity.csl to CMOS dependent

POBrien333:
via emails.
Publishers wrote they are switching over to CMOS


POBrien333:
So, there is an issue with bioarchaeology-international.csl, chungara-revista-de-antropologia-chilena.csl and
journal-of-the-association-of-environmental-and-resource-economists.csl just as they mention this style as their template style. Can we just ignore this?

adam3smith:
@POBrien333 There's no perfect solution, but we can't just accept it no (the error will show up in every PR test). I typically just update the templates to the new independent style, though it's not 100% right. Alternatively you could add the author of the old style as a contributor to the styles that used it as a template.

